PR #342, opened 2025-09-18
Create `MCPToolAction` with dynamic schema

xingyaoww:
Fix https://github.com/All-Hands-AI/agent-sdk/issues/309

Need to merge after https://github.com/All-Hands-AI/agent-sdk/pull/341

We have pretty strict typing system for MCPAction before by:
1. convert input_schema from MCP into a dynamic pydantic class
2. and use this dynamic class to check for arguments from LLM

BUT this is causing issues in action serialization, eg dynamic action class created in the server end is NOT available on the client side, creating complexity in the codebase (We have a VERY complex DiscriminativeUnionMixin to patch this 😓 ).

This PR instead relax the assumption for `MCPAction` -- you can put in arbitrary data into it (like a pydantic wrapper around plain JSON, so it is easy to serialization), but we can still type check the data at runtime by constructing a transitient pydantic class on the fly from mcp schema type.


openhands/tools/str_replace_editor/editor.py
```diff
@@ -507,7 +507,7 @@ def validate_path(self, command: CommandLiteral, path: Path) -> None:
         # Check if its an absolute path
         if not path.is_absolute():
             suggestion_message = (
-                "The path should be an absolute path, starting with `/`."
+                "The path should be an absolute path, starting with `/`.\n"
```
xingyaoww:
```suggestion
                "The path should be an absolute path, starting with `/`."
```

xingyaoww:
@OpenHands, please fix the failing actions on PR #342 at branch `xw/mcp-schema`

Review the diff on the changes we've added in this PR. Do NOT CHANGE IMPLEMENTATION --- Just change stuff on test to get tests to pass.

openhands/sdk/mcp/definition.py
```diff
@@ -22,6 +23,43 @@
 # will be dynamically created from the MCP tool schema.
 
 
+class MCPToolAction(ActionBase):
```
csmith49, line 27: One alternative to consider here: instead of using `extra="allow"` and letting all the fields from the tool call populate the instance, put in an explicit `data: dict[str, Any]` field. That way we can keep the fields inherited from `ActionBase` separate from the dynamic fields, which will help avoid name clashes and make it clear what is static and what is dynamic.

A little less magic than the _parent_fields gathering.

xingyaoww: Great point here!

openhands/sdk/mcp/tool.py
```diff
@@ -55,18 +63,37 @@ async def call_tool(self, action: MCPActionBase) -> MCPToolObservation:
...
+        We dynamically create a new MCPToolAction class with
+        the tool's input schema to validate the action.
+
+        Args:
+            action: The action to execute.
+
+        Returns:
+            The observation result from executing the action.
+        """
+        DynamicMCPActionType = MCPToolAction.from_mcp_schema(
```
csmith49: Probably beyond the scope of this PR, but we might also want to think about simplifying this kind of transformation. There are Python libraries like `jsonschema` that will handle validating Python data against a JSON schema, which means we could get by without the MCP scheme -> schema object -> dynamic pydantic model transformation and just validate the MCP schema directly.

openhands/sdk/mcp/definition.py
xingyaoww: But actually, if we want to do this, we'd either need to:
1. Add a special case in `agent.py` to check for `MCPToolAction` type from `tool.action_type`
2. Override `.model_dump` for `MCPToolAction` so it does nothing other than saving stuff to `.data` -> This doesn't seems like a best practice

Do you think if adding special case in `agent.py` worth it?

csmith49: If we're going to introduce any special cases it's probably the MCP tools that deserve them. But I don't quite see where we'd need the special case -- do you mean when we build the action from the completion? The only `.model_dump()` I see in `agent.py` is either for tools or messages.

csmith49: Hmm, maybe you were talking about this block in `_get_action_events`?

```
action: ActionBase = tool.action_type.model_validate(
    json.loads(tool_call.function.arguments)
)
```

If so, that looks like a good opportunity for a `Tool.action_from_tool_call` function that we can just overload for MCP tools.

xingyaoww: Yes - it is. sounds good! 

@OpenHands please implement .data field for MCPToolAction, and implement a `Tool.action_from_tool_call` so we can use it inside `_get_action_events`

openhands/sdk/mcp/definition.py
```diff
@@ -22,6 +24,32 @@
...
+
+    It is just a thin wrapper around raw JSON and does
+    not do any validation.
+
+    Validation will be performed by MCPTool.__call__
+    by constructing dynamically created Pydantic model
+    from the MCP tool input schema.
+    """
+
+    model_config = ConfigDict(frozen=True)
```
xingyaoww:
```suggestion
```

xingyaoww:
Tested locally and the MCP is working -- let's merge this now

image

PR #348, opened 2025-09-18
Replaces confirmation mode with confirmation policy

csmith49:
We were getting some annoying entanglement between confirmation mode and the security analyzer. This PR replaces the confirmation mode flag with `ConfirmationPolicy`, a base class that provides a `should_confirm` method that can be used to characterize precisely when the agent should be asking the user for confirmation.

There are three default policies:
1. `AlwaysConfirm()` is equivalent to `confirmation_mode = True`.
2. `NeverConfirm()` is equivalent to `confirmation_mode = False`.
3. `ConfirmRisky(threshold, confirm_unknown)` only asks for confirmation when a security analyzer places the action risk above `threshold`, unless `confirm_unknown` is set.

The three are models using our normal discriminated union pattern, so they can be easily serialized with the conversation state.

### DIFF
--- a/openhands/sdk/conversation/conversation.py
+++ b/openhands/sdk/conversation/conversation.py
@@ -1,6 +1,8 @@
 import uuid
 from typing import TYPE_CHECKING, Iterable
 
+from openhands.sdk.security.confirmation_policy import ConfirmationPolicy
+
 
 if TYPE_CHECKING:
     from openhands.sdk.agent import AgentType
@@ -192,11 +194,11 @@ class Conversation:
             if iteration >= self.max_iteration_per_run:
                 break
 
-    def set_confirmation_mode(self, enabled: bool) -> None:
+    def set_confirmation_mode(self, policy: ConfirmationPolicy) -> None:
         """Enable or disable confirmation mode and store it in conversation state."""
         with self.state:
-            self.state.confirmation_mode = enabled
-        logger.info(f"Confirmation mode {'enabled' if enabled else 'disabled'}")
+            self.state.confirmation_policy = policy
+        logger.info(f"Confirmation mode set to: {policy}")
 
     def reject_pending_actions(self, reason: str = "User rejected the action") -> None:
         """Reject all pending actions from the agent.
--- a/openhands/sdk/conversation/conversation.py
+++ b/openhands/sdk/conversation/conversation.py
@@ -1,7 +1,11 @@
 import uuid
 from typing import TYPE_CHECKING, Iterable
 
-from openhands.sdk.security.confirmation_policy import ConfirmationPolicy
+from openhands.sdk.security.confirmation_policy import (
+    AlwaysConfirm,
+    ConfirmationPolicy,
+    NeverConfirm,
+)
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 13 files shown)


openhands/sdk/conversation/conversation.py
```diff
@@ -192,11 +198,27 @@ def run(self) -> None:
             if iteration >= self.max_iteration_per_run:
                 break
 
+    def set_confirmation_policy(self, policy: ConfirmationPolicy) -> None:
+        """Set the confirmation policy and store it in conversation state."""
+        with self.state:
+            self.state.confirmation_policy = policy
+        logger.info(f"Confirmation policy set to: {policy}")
+
     def set_confirmation_mode(self, enabled: bool) -> None:
```
csmith49: This function now just maps True/False to AlwaysConfirm/NeverConfirm for legacy support. Ideally we'd be using the set_confirmation_policy function everywhere.

openhands/sdk/security/confirmation_policy.py
```diff
@@ -0,0 +1,53 @@
...
+from openhands.sdk.security.risk import SecurityRisk
+from openhands.sdk.utils.discriminated_union import (
+    DiscriminatedUnionMixin,
+    DiscriminatedUnionType,
+)
+
+
+class ConfirmationPolicyBase(DiscriminatedUnionMixin, ABC):
+    @abstractmethod
+    def should_confirm(self, risk: SecurityRisk = SecurityRisk.UNKNOWN) -> bool:
```
csmith49, line 11: It's not clear to me that this is the correct interface just yet, but it suffices to recreate the existing behavior and I figure we can update it pretty easily if we end up wanting a more complicated policy or have to change the security analyzer.

openhands/sdk/conversation/conversation.py
xingyaoww: Shall we just deprecate this function? @malhotra5

openhands/sdk/security/confirmation_policy.py
```diff
@@ -0,0 +1,53 @@
...
+        """Determine if an action with the given risk level requires confirmation."""
+        pass
+
+
+ConfirmationPolicy = Annotated[
+    ConfirmationPolicyBase, DiscriminatedUnionType[ConfirmationPolicyBase]
+]
```
xingyaoww: I think we will need to rework on the definition here when tim's PR is in

openhands/sdk/conversation/conversation.py
malhotra5: Let's just nuke it!

openhands/sdk/security/confirmation_policy.py
malhotra5: Yeah I think holding off a bit for now may be a good idea considering he's close to merging

openhands/sdk/conversation/conversation.py
csmith49: 💣

openhands/sdk/security/confirmation_policy.py
csmith49: No worries, if y'all don't have any other comments about this I'll fix the merge conflicts once his PR is in then get this merged.

enyst:
@openhands Merge main into this PR, and fix the conflicts carefully. Document concisely what you did.

xingyaoww:
Made some clean ups and want to get this in now -- otherwise we might be dealing other conflicts in https://github.com/All-Hands-AI/agent-sdk/pull/391

Issue #350, opened 2025-09-18
Make Tool serializable for remote runtime support


xingyaoww:
## Problem

Currently, `Tool` instances cannot be fully serialized because the `executor` field is excluded from serialization (`exclude=True`). This prevents users from defining tools locally and using them with remote runtimes, which is a key use case for the SDK.

### Current State

```python
# In openhands/sdk/tool/tool.py
class Tool(DiscriminatedUnionMixin, Generic[ActionT, ObservationT]):
    # ...
    executor: ToolExecutor | None = Field(default=None, repr=False, exclude=True)
```

The `executor` field is excluded from serialization, making it impossible to send tool definitions to remote runtimes.

### Desired Use Case

Users want to be able to define tools locally and use them with remote runtimes:

```python
import os
from openhands.sdk import (
    LLM,
    Agent,
    Conversation,
    Tool,
)

def say_hello():
    print("hello")

hello_tool = Tool(
    name="say_hello",
    description="Says hello to the user",
    executor=say_hello,
)

llm = LLM(
    model="anthropic/claude-sonnet-4-20250514",
    api_key=os.getenv("ANTHROPIC_API_KEY"),
)

agent = Agent(llm=llm, tools=[hello_tool])

# This should work with remote runtime
conversation = Conversation(agent=agent, host="https://host.docker.internal:3000")
conversation.send_text("say hello")
conversation.run()
```

### Option 1: Global Tool Registration

Implement a global tool registry that allows tools to be registered and referenced by name:

```python
from openhands.sdk import register_tool, Tool

def say_hello():
    print("hello")

register_tool("say_hello", Tool(
    name="say_hello",
    description="Says hello to the user",
    executor=say_hello,
))

# Tools array becomes serializable string references
agent = Agent(llm=llm, tools=["say_hello"])
```

**Benefits:**
- Clean separation between tool definition and usage
- Fully serializable tool references
- Backend can validate tool availability

**Challenges:**
- Requires global state management
- More complex API for simple use cases

### Option 2: HTTP/WebSocket Callbacks

Allow tools to register HTTP or WebSocket endpoints for remote execution:

```python
hello_tool = Tool(
    name="say_hello",
    description="Says hello to the user",
    executor=say_hello,
    callback_url="http://localhost:8080/tools/say_hello"  # Auto-generated
)
```

**Benefits:**
- Local Python code can be invoked from remote runtime
- Maintains current API simplicity
- True hybrid local/remote execution

[...]

PR #356, opened 2025-09-19
Unify tool initialization — Step 1: global registry + lazy materialization

xingyaoww:
We should merge AFTER https://github.com/All-Hands-AI/agent-sdk/pull/342

---

Summary
- Introduces a global tool registry to unify local/remote tool initialization paths and lay the groundwork for ToolSpec-first workflows.
- Adds lazy materialization of tools in Agent/Conversation so scripts can remain serializable while providing a consistent runtime experience.

Fixes: #350

Scope of changes
- New: openhands/sdk/tool/registry.py providing:
  - register_tool(name, factory)
  - resolve(spec: ToolSpec) -> list[Tool]
  - resolve_many(specs: Iterable[ToolSpec]) -> list[Tool]
  - register_openhands_tools() (idempotent) for BashTool, FileEditorTool, TaskTrackerTool, BrowserToolSet
  - list_registered()
- SDK exports: register_tool, register_openhands_tools via openhands/sdk/__init__.py
- Agent integration (local, backward-compatible):
  - Add initialize() and get_tools() to materialize tools at runtime
  - Support ToolSpec via registry resolution and merge with built-ins
  - Update internal references to use get_tools()
- Conversation: call agent.initialize() before init_state()
- Agent server: register_openhands_tools() during EventService.start() (idempotent)

Backward compatibility
- Existing Tool-based initialization continues to work. Agent.initialize() detects when Tool instances were passed and uses them directly (merged with built-ins).

Tests
- tests/sdk/tool/test_registry.py
  - registration and resolution for callable factory
  - instance registration rejects params
  - idempotency for register_openhands_tools()
- tests/sdk/agent/test_agent_toolspec_init.py
  - validates that ToolSpec-only agent initializes tools (including built-ins) when Conversation starts

Notes / Future work (Step 2+)
- Move ToolSpec handling up into AgentBase, replace top-level tools with tool_specs and finalize lazy materialization behavior across the stack
- Add /available_tools endpoint for remote discovery
- Consider HTTP-callback style executors for truly remote custom tools

Checklist
- [x] pre-commit: ruff, pyright, pycodestyle
- [x] unit tests for registry + ToolSpec init
- [x] agent_server idempotent auto-registration

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/229d308f8b7f437c99a986b6e6308d10)


openhands/sdk/preset/default.py
```diff
@@ -49,18 +44,11 @@ def get_default_agent_spec(
     llm: LLM,
     working_dir: str,
     cli_mode: bool = False,
-) -> AgentSpec:
-    agent_spec = AgentSpec(
+) -> Agent:
+    tool_specs = get_default_tools(working_dir=working_dir)
+    agent = Agent(
```
xingyaoww, line 59: @malhotra5 this should be helpful to you! We only load non-browser tools if cli_mode is true

openhands/sdk/preset/default.py
```diff
@@ -1,40 +1,35 @@
...
+    register_tool("BrowserToolSet", BrowserToolSet)
+
+    return [
+        ToolSpec(name="BashTool", params={"working_dir": working_dir}),
+        ToolSpec(name="FileEditorTool", params={}),
+        ToolSpec(
+            name="TaskTrackerTool", params={"save_dir": f"{working_dir}/.openhands"}
+        ),
+        ToolSpec(name="BrowserToolSet", params={}),
     ]
```
malhotra5:
```suggestion
def get_default_tools(working_dir: str, cli_mode: bool = False) -> list[ToolSpec]:
    """Get the default set of tool specifications for the standard experience."""
    from openhands.tools.execute_bash import BashTool
    from openhands.tools.str_replace_editor import FileEditorTool
    from openhands.tools.task_tracker import TaskTrackerTool

    register_tool("BashTool", BashTool)
    register_tool("FileEditorTool", FileEditorTool)
    register_tool("TaskTrackerTool", TaskTrackerTool)

    tool_specs = [
        ToolSpec(name="BashTool", params={"working_dir": working_dir}),
        ToolSpec(name="FileEditorTool", params={}),
        ToolSpec(
            name="TaskTrackerTool", params={"save_dir": f"{working_dir}/.openhands"}
        ),
    ]
    
    if not cli_mode:
        from openhands.tools.browser_use import BrowserToolSet
        register_tool("BrowserToolSet", BrowserToolSet)
        tool_specs.append(ToolSpec(name="BrowserToolSet", params={}))
    return tool_specs
```

Without this it seems browser use would still be included in the default agent spec that the CLI pulls

openhands/sdk/preset/default.py
```diff
@@ -49,18 +44,11 @@ def get_default_agent_spec(
     llm: LLM,
     working_dir: str,
     cli_mode: bool = False,
-) -> AgentSpec:
-    agent_spec = AgentSpec(
+) -> Agent:
+    tool_specs = get_default_tools(working_dir=working_dir)
```
malhotra5:
```suggestion
    tool_specs = get_default_tools(working_dir=working_dir, cli_mode=cli_mode)
```

examples/01_hello_world.py
```diff
@@ -26,19 +28,21 @@
...
-# tools = [
-#     BashTool.create(working_dir=cwd),
-#     FileEditorTool.create(),
-#     TaskTrackerTool.create(save_dir=cwd),
-# ]
-
-# Agent
-agent = Agent(llm=llm, tools=tools)
+
+register_tool("BashTool", BashTool)
```
malhotra5: IMHO the first example should be super simple for users. We call a one liner `get_default_agent_spec(llm=llm)` here. 

In another example we can show off the global registry side of things

malhotra5 (commented):
Working through review now

examples/12_custom_secrets.py
```diff
@@ -24,9 +21,11 @@
 )
 
 # Tools
+register_tool("BashTool", BashTool)
+register_tool("FileEditorTool", FileEditorTool)
 tools = [
-    BashTool.create(working_dir=os.getcwd()),
-    FileEditorTool.create(),
+    ToolSpec(name="BashTool", params={"working_dir": os.getcwd()}),
+    ToolSpec(name="FileEditorTool", params={}),
```
malhotra5:
```suggestion
    ToolSpec(name="FileEditorTool"),
```

Is it possible to just provide the tool name and not have to pass the empty param in this case?

openhands/sdk/preset/default.py
xingyaoww: actually i forgot to push this up earlier.. can you check the latest version?

examples/12_custom_secrets.py
xingyaoww: yes it is!

openhands/sdk/__init__.py
```diff
@@ -57,5 +65,8 @@
     "LLMSummarizingCondenser",
     "FileStore",
     "LocalFileStore",
+    "register_tool",
+    "resolve_tool",
```
malhotra5, line 69:
```suggestion
```

NIT: do we need this in the init file?

xingyaoww: I guess it is handy to keep it there -- Agent need it to convert spec back to tool instances

examples/12_custom_secrets.py
xingyaoww: did a global replace to fix stuff

openhands/sdk/agent/agent.py
```diff
@@ -114,8 +52,6 @@ def _configure_bash_tools_env_provider(self, state: ConversationState) -> None:
         Configure bash tool with reference to secrets manager.
         Updated secrets automatically propagate.
         """
-        if not isinstance(self.tools, dict):
```
malhotra5, line 117: We probably need some defensive code here to check that the `tool_map` as initialized properly? 

We need `agent.initialize()` to be called before `agent.init_state()` for this to work properly. Maybe we should consider calling `agent.initialize()` inside `agent.init_state()` so everything is instantiated and wired with one call

xingyaoww: done in https://github.com/All-Hands-AI/agent-sdk/pull/356/commits/048adf4c5e64d54072701c48aa1ba0645d4c6c92

examples/01_hello_world.py
xingyaoww: updated hello_world, but i still leave the code to register tool in comment for educational purpose

openhands/sdk/agent/base.py
```diff
@@ -26,81 +26,158 @@
...
+                )
+
+        # Check name duplicates
+        tool_names = [tool.name for tool in tools]
+        if len(tool_names) != len(set(tool_names)):
+            duplicates = set(name for name in tool_names if tool_names.count(name) > 1)
+            raise ValueError(f"Duplicate tool names found: {duplicates}")
+
+        # Store tools in a dict for easy access
+        self._tools = {tool.name: tool for tool in tools}
```
malhotra5: This seems a bit brittle to me or at least worth investigating. The registry key is different from the actual tool name correct? 


So `filter_tools_regex` matches runtime Tool.name (like `executebash`), not the registry key `BashTool`?

xingyaoww: > The registry key is different from the actual tool name correct?

One registry key corresponds to one tool class, which could give you one or multiple tools (e.g., one MCP gives me a bunch of tools), so it is not possible to enforce the mapping there.

> So filter_tools_regex matches runtime Tool.name (like executebash), not the registry key BashTool?

Yes -- since registry key is more for "class name" that we used internally; for tools like MCP, we don't really get their name until execution time

malhotra5: Ah gotcha! Let's document that `ToolSpec.name` is the registry key then not the actual tool name.

openhands/tools/__init__.py
```diff
@@ -1,105 +1,9 @@
 """Runtime tools package."""
 
-from importlib import import_module
 from importlib.metadata import PackageNotFoundError, version
-from typing import TYPE_CHECKING
 
 
-__all__ = [
```
malhotra5, line 8: So awesome!

openhands/sdk/preset/default.py
```diff
@@ -45,22 +46,19 @@ def get_default_condenser(llm: LLM) -> CondenserBase:
     return condenser
 
 
-def get_default_agent_spec(
+def get_default_agent(
```
malhotra5, line 49: NIT: just one that was a bit annoying was that I use this to create an agent, then I serialized the agent and saved to disk. When I deserialized and tried running it it crashed. Reason was because this function automatically registers the tools. When I close the program and load back the agent it didn't register its tools

I think a follow pr could maybe explore automatic registration for tools or maybe some helper utils, but otherwise this was very smooth!

openhands/sdk/tool/registry.py
```diff
@@ -0,0 +1,129 @@
...
+        created = create(**params)
+        tools = _ensure_tool_list(name, created)
+        # Optional sanity: permit tools without executor; they'll fail at .call()
+        return tools
+
+    return _resolve
+
+
+def register_tool(
+    name: str, factory: Tool | type[ToolBase] | Callable[..., Tool | list[Tool]]
```
malhotra5, line 95: Nit: could we find a more suitable  name instead of factory

openhands/sdk/agent/base.py
```diff
@@ -147,7 +179,52 @@ def init_state(
...
+        if self.mcp_config:
+            mcp_tools = create_mcp_tools(self.mcp_config, timeout=30)
+            tools.extend(mcp_tools)
+
+        logger.info(
+            f"Loaded {len(tools)} tools from spec: {[tool.name for tool in tools]}"
+        )
+        if self.filter_tools_regex:
+            pattern = re.compile(self.filter_tools_regex)
+            tools = [tool for tool in tools if pattern.match(tool.name)]
```
enyst, line 204: Could you tell what’s the idea with this regex? Is the user supposed to filter tool names?

xingyaoww: I think we don't really need this if we DON'T consider MCP.

I wanted to use https://repomix.com/guide/mcp-server -- then i realize it brings us too many tools that aren't useful to us (eg file-read/write which we already have).

If we go with spec based agent definition we have to do something like this, since we can't initialize mcp tools at the client side.

Though i can imagine we can do a follow-up: there could be another `register_mcp_tools` and it will send mcp config to the server, and then directly query the server to list "tool names", and then specifically set the tool names we want inside the spec.

@OpenHands please create a new github issue on All-Hands-AI/agent-sdk about this

xingyaoww:
@OpenHands please fix the failing actions on PR #356 at branch `openhands/unify-tool-init-step1`

Just focus on the file editor memory test issue - it seems pretty flaky and seems to pass when you run it in isolation.

Issue #282, opened 2025-09-15
Trigger condenser when LLM throws context window error


xingyaoww:
Related to this comment https://github.com/All-Hands-AI/agent-sdk/pull/181#discussion_r2349430461

In V0, we add `CondensationAction` to trigger a condensation when LLM exceeds its context window
https://github.com/All-Hands-AI/OpenHands/blob/65f41647496e322eaaeaa9e804ca4cd053c4d193/openhands/controller/agent_controller.py#L888-L917

We should also port it over here!

enyst:
This was probably fixed by #358 

Please feel free to reopen if not.

xingyaoww:
Relate to https://github.com/OpenHands/benchmarks/issues/45

We should do something similar to https://github.com/OpenHands/agent-sdk/pull/358 

and consider the following as an out-of-context error and trigger an condensation

>  {"type":"error","error":{"type":"invalid_request_error","message":"input length and `max_tokens` exceed context limit: 142123 +                               
                         64000 > 200000, decrease input length or `max_tokens` and try again"

xingyaoww:
@OpenHands please review the latest comment and sends in a PR to fix https://github.com/OpenHands/benchmarks/issues/45

csmith49:
Down to mark this as a context window exceeded error. But there might also be other ways to fix it. Is it possible/a good idea to modify max_tokens dynamically to make sure it doesn't exceed the threshold? I guess that would mean the model's max output would get smaller as the conversation goes on.

ryanhoangt:
I also observed this error a few times when running GAIA eval. My hypothesis (for both V0 and V1) is that the agent catches context limit exceptions only on its main LLM call, but the condenser’s summarization call using the same LLM with a large `max_output_tokens` value (64k) can itself exceed the context window, and we currenty don't have error handling for this condensation call.

@csmith49 WDYT? Is this possible and should we also add a try-catch around the condensation call as well?

https://github.com/OpenHands/software-agent-sdk/blob/9ac99bd6b5c853feb22ab23124091c553a0eba65/openhands-sdk/openhands/sdk/agent/agent.py#L120-L121

csmith49:
That's definitely a possibility. Maybe unlikely though -- the LLM only looks at the first half of the context, and its usually configured to use the same LLM as the agent. That's some very specific scenarios where the LLM in the condenser would throw a context window exception.

We probably should be catching it in the condenser though. Not sure what we'd do to recover.

[...]

PR #358, opened 2025-09-19
Handle context window exceeded exceptions with registered condenser

csmith49:
This PR updates the agent control flow to recover from certain errors when making completion calls. 

Specifically, if the agent can identify that the error is a context window exceeded error and the agent has a condenser registered that explicitly handles condensation requests, then the agent will emit a `CondensationRequest`. On the next agent step the condenser will use that request to trigger an immediate condensation, which will hopefully shrink the size of the context enough to fit in the context window.

Unfortunately, there still isn't a good way of detecting when a context window exceeded exception has happened. Every provider handles it a bit differently, and sometimes they slip through the cracks in LiteLLM. So we need a utility function (associated with the LLM at the moment) to do the identification for us.

Fixhttps://github.com/All-Hands-AI/agent-sdk/issues/282


csmith49:
This PR should resolve #282

PR #142, opened 2025-09-06
feat: Add OpenAI Responses API support to LLM class

enyst:
## Summary

This PR implements OpenAI Responses API support in the agent-sdk LLM class alongside the existing ChatCompletions API, as requested in issue #138.

We route `LLM.completion()` to the OpenAI Responses API (via LiteLLM) when the model supports it, convert `Responses` payloads into a ChatCompletions-like format, surface `assistant.message.reasoning_content`, and record `usage.reasoning_tokens`. Telemetry persists raw Responses payloads for audit.

### Core Implementation
- **Added `supports_responses_api` field** to `ModelFeatures` dataclass
- **Created `responses_converter.py`** with utilities to convert between ChatCompletions and Responses API formats
- **Implemented `responses()` method** in LLM class using `litellm.responses()`

### Features
- **Format conversion**: between message formats and string input/output
- **Model compatibility detection**: detects which models support Responses API
- **Backward compatibility**: Existing `completion()` method remains unchanged

## Supported Models

The implementation supports Responses API for:
- **OpenAI reasoning models**: GPT-5, GPT-5-mini, o1-preview, o1-mini, o3-mini.

## Usage Example

```python
from openhands.sdk.llm import LLM

# For reasoning models that support Responses API
llm = LLM(model="o1-preview")
if llm.is_responses_api_supported():
    response = llm.responses(
        input="Solve this step by step: What is 2+2?"
    )
    print(response.choices[0].message.content)

# Traditional ChatCompletions API still works
response = llm.completion(
    messages=[{"role": "user", "content": "What is 2+2?"}]
)
```

## Testing Results

All tests pass:
- ✅ 7/7 converter utility tests
- ✅ 22/22 model feature tests  
- ✅ 10/10 LLM responses method tests
- ✅ All existing tests continue to pass

Fixes #138

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/ff4b69c0b2f540be8fe460aa5f2425e0)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/All-Hands-AI/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Base Image | Docs / Tags |
|---|---|---|
| golang | `golang:1.21-bookworm` | [Link](https://hub.docker.com/_/golang:1.21-bookworm) |
| java | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
[...]

PR #371, opened 2025-09-20
refactor: centralize tool conversion in LLM (split from #142)

enyst:
Hello! I'm OpenHands-GPT-5. This PR centralizes tool conversion in the LLM.

This is split from #142 to land the refactor independently (no Responses API path).

Summary:
- Public API: LLM.completion now accepts Sequence[ToolBase] instead of raw dicts
- LLM converts Tool objects to ChatCompletionToolParam internally
- Agent no longer converts tools; passes Tool objects through
- Security: add_security_risk_prediction plumbed into LLM tool conversion
- Tests updated to pass Tool objects at public boundary; expectations adjusted

Notes:
- This PR does NOT enable the Responses API routing; that remains in #142
- No changes to execute_bash/PS1 logic


tests/sdk/llm/test_llm_completion.py
```diff
@@ -302,24 +299,23 @@ def test_llm_completion_non_function_call_mode(mock_completion):
...
+            action_type=TestNonFCArgs,
+        )
     ]
 
     # Verify that tools should be mocked (non-function call path)
-    assert llm.should_mock_tool_calls(tools)
+    # Note: should_mock_tool_calls() expects provider tool dicts;
+    # we just validate behavior below.
 
+    # We just exercise the completion() path here.
```
enyst:
```suggestion
```

tests/sdk/llm/test_llm_completion.py
```diff
@@ -388,7 +373,7 @@ def test_llm_completion_function_call_vs_non_function_call_mode(mock_completion)
     # Verify function calling is not active
     assert not llm_non_native.is_function_calling_active()
     # Should mock tools when native function calling is disabled but tools are provided
-    assert llm_non_native.should_mock_tool_calls(tools)
+    # should_mock_tool_calls requires provider tool dicts; skip direct assert here
```
enyst:
```suggestion

```

tests/sdk/llm/test_llm_completion.py
```diff
@@ -388,8 +373,6 @@ def test_llm_completion_function_call_vs_non_function_call_mode(mock_completion)
     # Verify function calling is not active
     assert not llm_non_native.is_function_calling_active()
     # Should mock tools when native function calling is disabled but tools are provided
```
enyst:
```suggestion
```

tests/sdk/llm/test_llm_completion.py
```diff
@@ -387,9 +372,6 @@ def test_llm_completion_function_call_vs_non_function_call_mode(mock_completion)
 
     # Verify function calling is not active
     assert not llm_non_native.is_function_calling_active()
-    # Should mock tools when native function calling is disabled but tools are provided
-    assert llm_non_native.should_mock_tool_calls(tools)
-
     # Call both and verify different behavior
```
enyst, line 402:
```suggestion

    # Call both and verify different behavior
```

tests/sdk/llm/test_llm_completion.py
```diff
@@ -373,7 +358,7 @@ def test_llm_completion_function_call_vs_non_function_call_mode(mock_completion)
     # Verify function calling is active
     assert llm_native.is_function_calling_active()
     # Should not mock tools when native function calling is active
-    assert not llm_native.should_mock_tool_calls(tools)
+    # should_mock_tool_calls requires provider tool dicts; skip direct assert here
```
enyst:
```suggestion
```

tests/sdk/llm/test_llm_completion.py
```diff
@@ -373,7 +358,6 @@ def test_llm_completion_function_call_vs_non_function_call_mode(mock_completion)
     # Verify function calling is active
     assert llm_native.is_function_calling_active()
     # Should not mock tools when native function calling is active
-    assert not llm_native.should_mock_tool_calls(tools)
```
enyst, line 376: The assert now requires tools in converted format, so we either skip it here or convert tools ourselves in the test. I suggest we can skip it, because we verify below that conversion did happen.

tests/sdk/llm/test_llm_completion.py
```diff
@@ -272,12 +274,12 @@ def test_llm_completion_non_function_call_mode(mock_completion):
...
-        '<invoke name="test_tool">\n'
-        '<parameter name="param">test_value</parameter>\n'
-        "</invoke>\n"
-        "</function_calls>"
+        (
+            "I'll help you with that.\n"
+            "<function=test_tool>\n"
+            "<parameter=param>test_value</parameter>\n"
+            "</function>"
+        )
```
enyst, line 282: About this, we now test this so replaced the old format.

> 
> - Upstream main’s test only verified that:
>   - tools weren’t passed to the provider in non‑native mode
>   - messages were modified
>   - It did not verify that tool_calls were reconstructed after the response
> - We strengthened the test to assert end‑to‑end behavior: given a textual tool call, the post‑response converter produces tool_calls in the final response.
> - To make that assertion meaningful, the mocked response has to match what our converter actually understands. That’s why we replaced the old <invoke …> style with our <function=…>/<parameter=…>…</parameter> format.
> 
> - Location: openhands/sdk/llm/mixins/fn_call_converter.py
> - The system_message_suffix_TEMPLATE and the TOOL_EXAMPLES both instruct/illustrate the new textual format:
>   - Function: <function=tool_name> … </function>
>   - Parameter: <parameter=param_name>value</parameter>
> - There’s no use of the older XML-style (<function_calls><invoke …><parameter …>…</invoke></function_calls>) in these prompts/examples.
> 
> So the non-native mixin’s prompt and examples use the new format.

tests/sdk/event/test_system_prompt_event_tools.py
xingyaoww, line 1: What's the purpose of this test? Seems not related to changes here?

tests/sdk/llm/test_llm_completion.py
```diff
@@ -302,30 +304,52 @@ def test_llm_completion_non_function_call_mode(mock_completion):
...
+    assert tc.type == "function"
+    assert tc.function.name == "test_tool"
+    # Ensure function-call markup was stripped from assistant content
+    if isinstance(msg.content, str):
+        assert "<function=" not in msg.content
+    assert tc.type == "function"
+    assert tc.function.name == "test_tool"
+    # Ensure function-call markup was stripped from assistant content
+    if isinstance(msg.content, str):
+        assert "<function=" not in msg.content
```
xingyaoww: redundant?

```suggestion
```

xingyaoww (approved):
lgtm! just a few issues

tests/sdk/event/test_system_prompt_event_tools.py
enyst: Some past iteration had the agent remove tools from the system prompt on visualization, so then it made a test that it doesn't. Maybe overkill 😅

PR #374, opened 2025-09-20
More serialization fix

xingyaoww:
Looks like we still need to treat `.field` as a class variable from https://github.com/All-Hands-AI/agent-sdk/pull/364, otherwise the Enum looks really pretty weird in openapi.json

We were hitting flaky OpenAPI generation and occasional runtime validation errors because unions were being (re)built at unpredictable times and in nondeterministic orders. Specifically, the TypeAdapter hook triggered rebuilds during schema traversal, so the same class could appear with different shapes/titles; and the union membership came from a set, so oneOf order changed run-to-run. The goal was to (1) make OpenAPI emission deterministic and (2) still support dynamic, plugin-style loading where new concrete types appear after import.

To do that, we shifted work into __init_subclass__: when a concrete subclass is defined, we stamp a stable kind: Literal["ClassName"], check for duplicate kinds, and proactively rebuild the abstract “owner” unions so their schemas include the new subclass immediately. That eliminates mid-traversal mutations and gives consistent oneOf order (we now sort by (module, qualname)). At the same time, we kept a conservative lazy-rebuild path (_rebuild_if_required) so if validation/schema is requested before all modules are imported (e.g., tests or late plugins), owners still get refreshed just-in-time. We also avoided forcing a global switch to Discriminator("kind") because not all models use Literal[...] today; sticking with Discriminator(kind_of) maintains compatibility. Finally, we removed the pre-validator that overwrote kind to prevent masking bad inputs. Net effect: stable, reproducible OpenAPI; robust runtime behavior with dynamic loading; and clearer, earlier errors for duplicate kinds—all with negligible startup overhead.

### DIFF
--- a/openhands/sdk/utils/models.py
+++ b/openhands/sdk/utils/models.py
@@ -7,10 +7,8 @@ from typing import Annotated, Any, Literal, Self, Type, Union
 from pydantic import (
     BaseModel,
     Discriminator,
-    Field,
     Tag,
     TypeAdapter,
-    model_validator,
 )
 
 
@@ -105,7 +103,7 @@ class DiscriminatedUnionMixin(OpenHandsModel, ABC):
     discriminator for union types.
     """
 
-    kind: str = Field(default="")  # We dynamically update on a per class basis
+    kind: str = ""  # We dynamically update on a per class basis
 
     @classmethod
     def resolve_kind(cls, kind: str) -> Type:
@@ -126,24 +124,11 @@ class DiscriminatedUnionMixin(OpenHandsModel, ABC):
         if cls == DiscriminatedUnionMixin:
             pass
         if _is_abstract(cls):
-            subclasses = get_known_concrete_subclasses(cls)
-            kinds = [subclass.__name__ for subclass in subclasses]
-            if kinds:
-                kind = cls.model_fields["kind"]
-                kind.annotation = Literal[tuple(kinds)]  # type: ignore
-                kind.default = kinds[0]
-
-            # Update kind to be specific - not just str...
             type_adapter = TypeAdapter(cls.get_serializable_type())
             cls.__pydantic_core_schema__ = type_adapter.core_schema
             cls.__pydantic_validator__ = type_adapter.validator
             cls.__pydantic_serializer__ = type_adapter.serializer
             return
-        else:
-            # Update the kind field to be a literal.
-            kind = cls.model_fields["kind"]
-            kind.annotation = Literal[cls.__name__]  # type: ignore
-            kind.default = cls.__name__
 
         return super().model_rebuild(
             force=force,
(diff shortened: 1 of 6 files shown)


openhands/sdk/utils/models.py
```diff
@@ -126,24 +131,11 @@ def model_rebuild(
...
             pass
         if _is_abstract(cls):
-            subclasses = get_known_concrete_subclasses(cls)
-            kinds = [subclass.__name__ for subclass in subclasses]
-            if kinds:
-                kind = cls.model_fields["kind"]
-                kind.annotation = Literal[tuple(kinds)]  # type: ignore
-                kind.default = kinds[0]
-
-            # Update kind to be specific - not just str...
```
xingyaoww, line 136: These are mostly bring back fixes from https://github.com/All-Hands-AI/agent-sdk/pull/364

openhands/sdk/utils/models.py
```diff
@@ -253,20 +240,3 @@ def _get_all_subclasses(cls) -> set[Type]:
...
-
-_orig_type_adapter_init = TypeAdapter.__init__
-
-
-def _init_with_hook(*args, **kwargs):
-    _rebuild_if_required()
-    _orig_type_adapter_init(*args, **kwargs)
-
-
-TypeAdapter.__init__ = _init_with_hook
```
xingyaoww, line 272: Except that we remove this, since this breaks `make test-server-schema` -- according to GPT-5:


> 1) Rebuilding during schema generation

> You monkey-patch TypeAdapter.__init__ to call _rebuild_if_required(). FastAPI builds OpenAPI by repeatedly creating TypeAdapters while traversing your types. Your hook therefore mutates model core schemas mid-traversal. That makes Pydantic see different shapes for the same logical model on different passes.
> 
> Effect: sometimes a reference is registered as PipelineCondenser, elsewhere as PipelineCondenser-Output, and occasionally you even get two distinct component entries both titled PipelineCondenser. This directly explains:
> 	•	“Attempted to generate duplicate models with name PipelineCondenser”
> 	•	knock-on error: “Invalid property in union condenser” because that union references a component the generator just discarded as invalid/duplicate.

openhands/sdk/utils/models.py
```diff
@@ -129,21 +131,15 @@ def model_rebuild(
...
             type_adapter = TypeAdapter(cls.get_serializable_type())
             cls.__pydantic_core_schema__ = type_adapter.core_schema
             cls.__pydantic_validator__ = type_adapter.validator
             cls.__pydantic_serializer__ = type_adapter.serializer
             return
-        else:
-            # Update the kind field to be a literal.
-            kind = cls.model_fields["kind"]
-            kind.annotation = Literal[cls.__name__]  # type: ignore
-            kind.default = cls.__name__
```
xingyaoww, line 146: We will annotate this in `__init_subclass__`

tofarr (approved):
Great Work! 🙌

PR #375, opened 2025-09-20
Public APIs return OpenHands types (CompletionResult, OpenHandsToolSpec) not LiteLLM types

jpshackelford:
## Summary

This PR implements the LLM service abstraction layer design to ensure **Public APIs return OpenHands types (CompletionResult, OpenHandsToolSpec) not LiteLLM types**. This architectural improvement creates cleaner abstractions, reduces coupling to LiteLLM implementation details, and enables potential alternative language SDK implementations.

## 📖 Design Document

See the full design document: [`doc/design-doc/llm-service.md`](doc/design-doc/llm-service.md)

## 🎯 Problem Solved

The OpenHands SDK had **two critical interface boundaries** where LiteLLM types leaked through to consumer code:

1. **Completion Interface**: `LLM.completion()` returned LiteLLM's `ModelResponse`, requiring manual type conversions
2. **Tool Interface**: Tools used `ChatCompletionToolParam` instead of OpenHands-native types

This forced agent and tool code to import LiteLLM types and perform manual conversions, creating unnecessary coupling.

### Phase 1: CompletionResult Implementation (Sections 4.1-4.2) ✅

**🔧 New CompletionResult Type:**
```python
@dataclass
class CompletionResult:
    message: Message              # OpenHands native type
    metrics: MetricsSnapshot      # OpenHands native type  
    raw_response: ModelResponse   # LiteLLM type (for advanced use)
```

**🔄 Updated LLM.completion() Interface:**
```python
# Before: Returns LiteLLM ModelResponse
def completion(...) -> ModelResponse

# After: Returns OpenHands CompletionResult  
def completion(...) -> CompletionResult
```

**🧹 Simplified Agent Code:**
```python
# Before: Manual type conversion required
response = self.llm.completion(...)
assert len(response.choices) == 1 and isinstance(response.choices[0], Choices)
llm_message: LiteLLMMessage = response.choices[0].message
message = Message.from_litellm_message(llm_message)
metrics = self.llm.metrics.get_snapshot()

# After: Clean OpenHands types
completion_result = self.llm.completion(...)
message = completion_result.message
metrics = completion_result.metrics
```

## 📊 Impact

**✅ Files Cleaned Up:**
- `openhands/sdk/agent/agent.py` - Removed `Choices`, `LiteLLMMessage` imports
- `openhands/sdk/context/condenser/llm_summarizing_condenser.py` - Uses `CompletionResult`
- All consumer code now uses OpenHands types exclusively

**✅ Test Coverage:**
- **376 tests passing** across LLM, Agent, and Cross-integration suites
- 13/13 LLM completion tests updated and passing
- 66/66 Agent tests updated and passing  
- 5/5 Condenser tests updated and passing
- All type checking issues resolved

**✅ Backward Compatibility:**
- Raw LiteLLM response still available via `completion_result.raw_response`
- Internal LiteLLM usage unchanged
- No breaking changes to existing functionality

## 🚀 Benefits

1. **Cleaner Public APIs**: Consumer code only imports OpenHands types
2. **Reduced Coupling**: Less dependency on LiteLLM implementation details
3. **Better Abstractions**: Consistent with existing OpenHands type system
4. **Future-Proof**: Enables alternative language SDK implementations
5. **Simplified Code**: Eliminates manual type conversions and assertions

## 🔄 What's Next

**Phase 2: Tool Interface (Section 4.3)** - Create `OpenHandsToolSpec` type
**Phase 3: Documentation (Section 4.4)** - Update examples and docs

## 📋 Files Changed

- **New**: `openhands/sdk/llm/completion_result.py` - CompletionResult type
- **Updated**: `openhands/sdk/llm/llm.py` - Return CompletionResult
- **Updated**: `openhands/sdk/agent/agent.py` - Use CompletionResult  
- **Updated**: `openhands/sdk/context/condenser/llm_summarizing_condenser.py` - Use CompletionResult
- **Updated**: All related tests to expect CompletionResult
- **Updated**: `doc/design-doc/llm-service.md` - Mark sections 4.1-4.2 complete

This implementation successfully achieves the core architectural goal: **Public SDK APIs now return only OpenHands types**, creating a cleaner, more maintainable abstraction layer.


jpshackelford:
@xingyaoww Would you be willing to take an early look to see if I am on the right track with this before I finish the steps in the design doc and complete the PR?

openhands/sdk/agent/agent.py
```diff
@@ -23,7 +19,6 @@
 from openhands.sdk.event.condenser import Condensation
 from openhands.sdk.event.utils import get_unmatched_actions
 from openhands.sdk.llm import (
-    Message,
```
enyst, line 26: I think maybe it’s nice to keep the type hint, even if we don’t directly call this anymore?

enyst (commented):
This looks good to me, thank you!

I think it’s a good idea to make the Agent independent of litellm types, also because
- we want to support both [completions API and responses API](https://github.com/All-Hands-AI/agent-sdk/issues/138#issuecomment-3293323214)
- and tbh the dependency feels like too much LLM implementation detail in the agent.

Also related, I propose here to let Agent use only Tool, instead of converting tools for LLM:
- https://github.com/All-Hands-AI/agent-sdk/pull/371

openhands/sdk/agent/agent.py
jpshackelford: I have restored the type hint for clarity. Thanks for catching that.

doc/design-doc/llm-service.md
xingyaoww, line 1: let's remove this?

openhands/sdk/llm/completion_result.py
```diff
@@ -0,0 +1,37 @@
...
+        raw_response: The original LiteLLM ModelResponse for internal use
+    """
+
+    message: Message
+    metrics: MetricsSnapshot
+    raw_response: ModelResponse
+
+    class Config:
+        # Allow arbitrary types for ModelResponse
+        arbitrary_types_allowed = True
```
xingyaoww: nit: we should use model_config = ConfigDict syntax here to match other parts of the repo

openhands/sdk/llm/completion_result.py
```diff
@@ -0,0 +1,37 @@
...
+
+from openhands.sdk.llm.message import Message
+from openhands.sdk.llm.utils.metrics import MetricsSnapshot
+
+
+__all__ = ["CompletionResult"]
+
+
+class CompletionResult(BaseModel):
+    """Result of an LLM completion request.
```
xingyaoww: I wonder if we'd want to rename it a bit later when @enyst added support for responses API?

Having our internal standardized format seems to help here though

xingyaoww (commented):
Overall LGTM

openhands/sdk/llm/completion_result.py
jpshackelford: @enyst Is there a better name we can use now?

doc/design-doc/llm-service.md
jpshackelford: removed

openhands/sdk/llm/completion_result.py
jpshackelford: Now:

```python
  model_config = ConfigDict(arbitrary_types_allowed=True)
  ```

File: openhands/sdk/llm/completion_result.py (hunk 3 above)
Comment (enyst) [reply openhands/sdk/llm/completion_result.py]: We’d want it to be a facade for both response types?

Right now, the Agent uses `completion_result.raw_response.id`, where `raw_response` is a ModelResponse, so completions-specific.

I imagine we’d take out the direct dependencies on litellm types in the Agent => which maybe means expose a “response_id” from this class => client code will not use raw_response?

Responses API will add an e.g. `previous_response_id`.

=> if that sounds good, the name for that class, you’re right it shouldn’t have “completion”, maybe just `Response`? Or `ModelResult` but I feel that’d be difficult to remember/differentiate; `LLMResponse`? We do use “LLM” with capitals in other names in `llm` module

Comment (jpshackelford) [reply openhands/sdk/llm/completion_result.py]: The codebase shows a clear, consistent pattern of using LLM (all uppercase) in class names:
`LLM` (main class)
`LLMError`, `LLMResponseError`, `LLMNoActionError`
`LLMRegistry`, `LLMSecurityAnalyzer`, `LLMContextManager`
`LLMConvertibleEvent`, `LLMSummarizingCondenser`
`RouterLLM`

So I will use `LLMResponse` as the new name for `CompletionResult`.

File: openhands/sdk/agent/agent.py [hunk 4]
```diff
@@ -177,7 +173,7 @@ def step(
         )
 
         try:
-            response = self.llm.completion(
+            completion_result = self.llm.completion(
```
xingyaoww: Let's rename all these variable to llm_response too?

openhands/sdk/agent/agent.py
```diff
@@ -239,7 +234,7 @@ def step(
                 action_event = self._get_action_event(
                     state,
                     tool_call,
-                    llm_response_id=response.id,
+                    llm_response_id=completion_result.raw_response.id,
```
xingyaoww: nit: but a nice thing we can do is to add @property for `LLMResponse`, so we can use completion ID when it is in completion mode, and use response ID when we introduce responses API later (cc @enyst )

openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -65,15 +65,20 @@ def get_condensation(self, view: View) -> Condensation:
 
         messages = [Message(role="user", content=[TextContent(text=prompt)])]
 
-        response = self.llm.completion(
+        completion_result = self.llm.completion(
```
xingyaoww: ditto, llm_response

xingyaoww (approved):
LGTM! Just a few nits

openhands/sdk/agent/agent.py
jpshackelford: doh! I should have caught that.

openhands/sdk/agent/agent.py
enyst: Exactly, hide the details of how it got that id, it's always an id

openhands/sdk/agent/agent.py
jpshackelford: fixed

openhands/sdk/agent/agent.py
jpshackelford: Did I do what you intended for this one?

jpshackelford:
```
    @property
    def response_id(self) -> str:
        """Get the response ID from the underlying LLM response.

        This property provides a clean interface to access the response ID,
        supporting both completion mode (current) and future response API modes.

        Returns:
            The response ID from the LLM response
        """
        return self.raw_response.id
```

jpshackelford: @xingyaoww Like this?

xingyaoww: I think should be something like:

```
    @property
    def id(self) -> str:
        """Get the response ID from the underlying LLM response.

        This property provides a clean interface to access the response ID,
        supporting both completion mode (current) and future response API modes.

        Returns:
            The response ID from the LLM response
        """
        return self.raw_response.id if completion_mode else XXX
```

openhands/sdk/agent/agent.py
```diff
@@ -239,7 +234,7 @@ def step(
                 action_event = self._get_action_event(
                     state,
                     tool_call,
-                    llm_response_id=response.id,
+                    llm_response_id=llm_response.response_id,
```
xingyaoww:
```suggestion
                    llm_response_id=llm_response.id,
```

openhands/sdk/llm/llm_response.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    message: Message
+    metrics: MetricsSnapshot
+    raw_response: ModelResponse
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    @property
+    def response_id(self) -> str:
```
xingyaoww:
```suggestion
    def id(self) -> str:
```

xingyaoww (approved):
Like this!

PR #377, opened 2025-09-21
Relax openapi schema test & make Microagent a DiscriminatedUnionMixin and tests

xingyaoww:
We need to make microagent a DiscriminatedUnionMixin too - otherwise they won't deserialize properly in `Agent` class from JSON


xingyaoww:
@OpenHands please fix the failing actions on PR #377 at branch `xw/explicit-union-microagent`

Do not change my implementations, just fix the tests

xingyaoww:
@OpenHands review diff against main and understand the change. And then add some tests to test microagent serializable based on BaseMicroagent, leveraging the union annotated using a discriminator


MicroagentUnion = Annotated[
    RepoMicroagent | KnowledgeMicroagent | TaskMicroagent,
    Field(discriminator="type"),
]

openhands/sdk/context/microagents/microagent.py
```diff
@@ -239,29 +235,31 @@ class TaskMicroagent(KnowledgeMicroagent):
...
+            return self
 
         prompt = (
             "\n\nIf the user didn't provide any of these variables, ask the user to "
             "provide them first before the agent can proceed with the task."
         )
-        self.content += prompt
+
+        # Avoid duplicating the prompt if content already includes it
+        if getattr(self, "content", None) is not None and prompt not in self.content:
```
enyst:
```suggestion
        if prompt not in self.content:
```
It seems the attribute is always there?

openhands/sdk/context/microagents/microagent.py
```diff
@@ -344,3 +342,9 @@ def load_microagents_from_dir(
         f"{[*repo_agents.keys(), *knowledge_agents.keys()]}"
     )
     return repo_agents, knowledge_agents
+
+
+MicroagentUnion = Annotated[
+    RepoMicroagent | KnowledgeMicroagent | TaskMicroagent,
+    Field(discriminator="type"),
```
enyst: Do we really need this part? If it's not a must, maybe we could add it only later, when really necessary?

I'm thinking that
- if client developers want to extend microagents with another type, they'd need to modify this kind of code
- granted, we already were working with types, but I'm not sure if the reason was only validation of that field, or was there more
- such validation though, makes the code less extensible, since it seems it means we exclude the idea of developers adding a type without modifying code (that code)
- maybe we can just fallback to `knowledge` for unknown type, with a warning?

xingyaoww: > Do we really need this part?

YES to use [pydantic unions/ discriminative mixin] properly. Otherwise, we have to use our current hacky DiscriminativeMixin -- this PR was the product of my frustration to get these union to work properly for openapi-python-client for spec check.

But later on i decided maybe we should just throw "openapi-python-client" away for check as it doesn't works well for discriminative union and the actual web server spec looks good to me.

it has cost me countless hours to figure out typing issue there 😓 maybe it is not worth it to keep this test

xingyaoww: ok reverted back!

openhands/sdk/context/microagents/microagent.py
xingyaoww: fixed

PR #378, opened 2025-09-21
Add example showing agent can handle messages sent while task is running

simonrosenberg:


simonrosenberg:
@xingyaoww  I added this as an example for now.
I am working on a second PR to solve the final action bug that
1. makes a failing test with mocks, revealing the final action bug
2. solves the final action bug.

examples/18_send_message_while_processing.py
xingyaoww, line 1: I think you need to change the example to fit the lastest main where we changed the tool interface for agent so Agent can be easily serialized

xingyaoww: @Openhands, please rewrite examples/18 to follow the format of using register_tool in other files

xingyaoww (approved):
LGTM!

PR #391, opened 2025-09-21
feat(sdk): Implement LocalConversation and RemoteConversation (fixes #388)

### LINKED ISSUE #388: Implement `LocalConversation` and `RemoteConversation`
# State as of now

We have `openhands/agent_server`, which allows you to interact with a remote conversation via API. We also have examples in `examples/` that directly pass information to `Conversation` and interact with a local version of the conversation.

# What we want to implement:

We want to port over the existing logic of `Conversation` to the `LocalConversation` class that inherits from `Conversation`. In addition to that, we also want a `RemoteConversation` that implement the exact same interface as `Conversation`, BUT instead sending stuff to a remote conversation maintained by an agent server, it will perform tasks through an API (i.e., `RemoteConversation` will just be a client for `agent_server`

# What would be an example looks like:

As discussed in https://github.com/All-Hands-AI/agent-sdk/issues/350, we want to have a way to do this:

1. Launch an agent server on the side `uv run agent-server` (available at `https://host.docker.internal:3000`)

2. Run this script and everything just works as if this conversation happens locally

```python
import os
from openhands.sdk import (
    LLM,
    Agent,
    Conversation,
    Tool,
    ToolSpec,
    register_tool
)

llm = LLM(
    model="anthropic/claude-sonnet-4-20250514",
    api_key=os.getenv("ANTHROPIC_API_KEY"),
)

agent = Agent(llm=llm, tools=[ToolSpec(name="say_hello")])

# Agent server SHOULD automatically register all tools under openhands.tools
# We should also add an `/available_tools` endpoint to the `Conversation` so we can get init
conversation = Conversation(agent=agent, host="https://host.docker.internal:3000")
# Then we should call `agent.initialize()` INSIDE conversation

conversation.send_message("say hello")
conversation.run()
```

To OpenHands: If you are implementing this, you should be able to `export LITELLM_API_KEY=$LITELLM_API_KEY`, this should set your environment variable; Then you will be able to run examples in `examples/` to make sure things are working.

xingyaoww:
Summary
- Introduces LocalConversation and RemoteConversation while preserving the original Conversation interface
- Conversation is now a factory that returns LocalConversation (default) or RemoteConversation (when host is provided)
- LocalConversation ports the existing local conversation logic unchanged
- RemoteConversation proxies the Agent Server REST API to perform send_message, run (confirmation accept), pause, and reject operations
- All existing tests pass (1098 passed)

What & Why
We want to support both local and remote conversations behind a single, consistent interface. This change:
- Keeps local behavior identical to the prior Conversation implementation
- Adds RemoteConversation to talk to a running agent-server over HTTP, enabling scripts like examples/01_hello_world.py to point to a server simply by passing host=...
- Maintains backward compatibility; no changes are required for existing local usage

Design Overview
- Conversation
  - Now acts as a small factory: if host is provided, returns a RemoteConversation; otherwise, returns LocalConversation
  - New signature supports host and confirmation_mode (for Remote)
- LocalConversation
  - Direct port of the previous Conversation class
  - Handles state, visualization, callbacks, and run loop as before
- RemoteConversation
  - Uses httpx to communicate with the agent server endpoints
  - start: POST /conversations with Agent config; stores conversation id
  - send_message: POST /conversations/{id}/events with run=False
  - run: POST /conversations/{id}/events/respond_to_confirmation with accept=True, then polls /conversations/{id} until terminal state
  - pause: POST /conversations/{id}/pause
  - reject_pending_actions: POST /conversations/{id}/events/respond_to_confirmation with accept=False
  - update_secrets: no-op with a warning (not supported remotely yet)

Compatibility & Migration
- No breaking changes to existing imports or usage
- Conversation(agent=..., ...) continues to work locally
- To use a remote server: Conversation(agent=..., host="https://host.docker.internal:3000")

Testing
- Ran full test suite: 1098 passed, 44 warnings
- Validated conversation visualizer/callback initialization ordering
- Verified agent-server tests continue to pass with LocalConversation used inside EventService

Notes
- RemoteConversation currently uses simple polling in run(); future improvement could stream events via websocket for richer client-side callbacks
- set_confirmation_mode is only supported at initialization for Remote; follow-up work can expose an API if needed

Closes
- Fixes #388

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c3a03b10cc80466dadd982f605090d11)

### DIFF
--- a/openhands/sdk/conversation/conversation.py
+++ b/openhands/sdk/conversation/conversation.py
@@ -1,5 +1,8 @@
+import time
 import uuid
 from typing import Iterable
+
+import httpx
 
 from openhands.sdk.agent.base import AgentBase
 from openhands.sdk.conversation.secrets_manager import SecretValue
@@ -34,6 +37,46 @@ def compose_callbacks(
 
 
 class Conversation:
+    """Factory entrypoint that returns a LocalConversation or RemoteConversation.
+
+    Usage:
+        - Conversation(agent=...) -> LocalConversation
+        - Conversation(agent=..., host="http://...") -> RemoteConversation
+    """
+
+    def __new__(
+        cls,
+        agent: AgentBase,
+        persist_filestore: FileStore | None = None,
+        conversation_id: ConversationID | None = None,
+        callbacks: list[ConversationCallbackType] | None = None,
+        max_iteration_per_run: int = 500,
+        visualize: bool = True,
+        host: str | None = None,
+        confirmation_mode: bool | None = None,
+    ):
+        if cls is Conversation:
+            if host:
+                return RemoteConversation(
+                    agent=agent,
+                    host=host,
+                    conversation_id=conversation_id,
+                    callbacks=callbacks,
+                    max_iteration_per_run=max_iteration_per_run,
+                    confirmation_mode=confirmation_mode,
+                )
+            return LocalConversation(
+                agent=agent,
+                persist_filestore=persist_filestore,
+                conversation_id=conversation_id,
+                callbacks=callbacks,
+                max_iteration_per_run=max_iteration_per_run,
+                visualize=visualize,
+            )
+        return super().__new__(cls)
+
+
+class LocalConversation(Conversation):
     def __init__(
         self,
         agent: AgentBase,
@@ -42,6 +85,7 @@ class Conversation:
         callbacks: list[ConversationCallbackType] | None = None,
         max_iteration_per_run: int = 500,
         visualize: bool = True,
+        **_: object,
     ):
         """Initialize the conversation.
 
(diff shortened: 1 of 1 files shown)


openhands/agent_server/conversation_router.py
```diff
@@ -164,3 +166,83 @@ async def delete_conversation(conversation_id: UUID) -> Success:
...
+
+
+@router.post(
+    "/{conversation_id}/run",
+    responses={
+        404: {"description": "Item not found"},
+        409: {"description": "Conversation is already running"},
+    },
+)
+async def run_conversation(conversation_id: UUID) -> Success:
```
tofarr, line 171: How does this differ from the existing `resume_conversation` endpoint? Should we delete one of these paths?

openhands/agent_server/conversation_router.py
```diff
@@ -164,3 +166,83 @@ async def delete_conversation(conversation_id: UUID) -> Success:
...
+    secrets = cast(dict[str, SecretValue], request.secrets)
+    await event_service.update_secrets(secrets)
+    return Success()
+
+
+@router.post(
+    "/{conversation_id}/confirmation-policy",
+    responses={404: {"description": "Item not found"}},
+)
+async def set_conversation_confirmation_policy(
```
tofarr, line 216: How many other items like secrets and confirmation policy are we likely to need? Would it make sense to have a more generic "UpdateConversationRequest"?

openhands/agent_server/conversation_router.py
xingyaoww: With some offline discussion with Tim, we decided that this sounds pretty similar to `resume` and we decide to merge this with `resume` by removing the `resume` endpoint

openhands/agent_server/conversation_router.py
xingyaoww: Also based on our offline convo: there shouldn't be a tons of thing we will add -- this route will simplify be here to maintain a parity with the conversation base class we will define `openhands/sdk/conversation/base.py`

xingyaoww:
@OpenHands review diff against main and help me fix all the `pre-commit run --all-files` error and make sure `pytest tests/` all passes

openhands/agent_server/event_service.py
```diff
@@ -177,47 +174,16 @@ async def start(self, conversation_id: UUID):
...
-        try:
-            await loop.run_in_executor(None, self._conversation.run)
-        except Exception as e:
-            logger.exception(
-                f"Run failed for conversation {self.stored.id}: {e}",
-                stack_info=True,
-            )
-            raise
-
-    async def start_background_run(self):
```
xingyaoww, line 194: Background run has weird issues of delayed start, so what will happen is:
1. client kick off run
2. and then poll the agent status -- realize the agent state is IDLE, think that the agent completed the task
3. Then early exit

I think the better fixes here will be supporting sending `ConversationState` updates through websocket as well so we can get rid of polling. Before that, i think this blocking request will probably work for now and we can fix it when we run into problems

openhands/sdk/conversation/conversation.py
```diff
@@ -1,14 +1,13 @@
...
-from openhands.sdk.conversation.impl import LocalConversation
+from openhands.sdk.conversation.impl import LocalConversation, RemoteConversation
 from openhands.sdk.conversation.types import ConversationCallbackType, ConversationID
 from openhands.sdk.io import FileStore
 from openhands.sdk.logger import get_logger
 
 
-# ConversationType = Union[LocalConversation, RemoteConversation]
-ConversationType = LocalConversation  # RemoteConversation is not implemented yet.
+ConversationType = Union[LocalConversation, RemoteConversation]
```
enyst: Just a small question here, do we really need `ConversationType`, isn't the common type `Conversation`?

I feel like there's a lot to like about them both being accessed via `Conversation` by external clients.

So I wonder if we can avoid enumerating subclasses in code, idk, for example adding a new one wouldn't be captured by `*Type`, but it clearly would be a subclass like any other.

xingyaoww: good point, making Local/Remote as subclass now

xingyaoww:
@OpenHands please fix the failing actions on PR #391 at branch `openhands/implement-remote-local-conversation`

DO NOT CHANGE IMPLEMENTATION, JUST FIX TESTS

xingyaoww:
@OpenHands, review the diff against main and help me write comprehensive tests to verify the new components/features implemented in this PR.

Only write tests that are genuinely useful. You can consider writing "cross" tests that is similar to 22_hello_world_with_agent_server that test the whole thing end to end

PR #397, opened 2025-09-22
Xw/responses api typing

xingyaoww:
Title: Responses API typing: return real LiteLLM message types end-to-end, refine caching semantics, and align tests

Summary
This PR completes the transition to strongly typed LiteLLM message models across the chat path, while keeping Responses API behavior consistent and simple. The core outcome is that format_messages_for_llm() and Message.to_llm_dict() now return proper LiteLLM ChatCompletion message types (AllMessageValues), not loose dicts. We also preserve content-level cache_control on the chat path (including for tool messages), drop message-level cache_control/name on tool payloads, and ensure the Responses path ignores cache_control entirely.

Motivation
- Strong typing reduces ambiguity and prevents shape drift in message handling.
- Aligns our serialization with LiteLLM’s canonical message models, making cross-provider integration safer.
- Keeps Anthropic-friendly caching without exposing message-level cache_control that providers don’t accept.

Key changes
1) Strict typing for chat messages
- Message.to_llm_dict() now builds the concrete LiteLLM ChatCompletion*Message models.
- LLM.format_messages_for_llm() returns list[AllMessageValues] exclusively.
- Tests were updated to safely access LiteLLM TypedDict/union fields under Pyright strict.

2) Chat path caching and message shape
- Content-level cache_control is preserved on the list serializer for all roles, including tool.
- No message-level cache_control on any messages.
- Tool messages: tool_call_id preserved; name not included in LLM API serialization (still present in storage serialization).
- Serializer selection:
  - string serializer for simple text
  - list serializer when cache/vision/function-calling is enabled or required for the model
- Anthropic-friendly caching stays behind is_caching_prompt_active() and model feature flags.

3) Responses API path remains simple
- messages_to_responses_items() accepts list[AllMessageValues] (or dicts), normalizes, and produces the expected Responses API input items.
- responses_to_completion_format() converts Responses results back to ChatCompletions-like ModelResponse.
- Cache control is ignored for the Responses path by design.

4) Tests and quality gates
- Updated tests to assert against LiteLLM types via dict-style access to optional keys and handle content as str | list.
- No new Anthropic caching tests added; existing coverage passes, confirming no behavior regressions.
- Full test suite: 1151 passed locally.
- Pre-commit (ruff, formatting) and Pyright strict pass for updated files.

Backward compatibility
- Storage serialization (model_dump/model_dump_json) is unchanged and preserves all fields, including name on tool messages.
- LLM API serialization shape for tool messages no longer includes name; clients relying on name at the message level should read it from tool_calls function metadata or keep using storage serialization where needed.
- Responses path unchanged functionally; it simply ignores cache_control.

Migration notes
- If code previously assumed dict messages from format_messages_for_llm(), migrate to LiteLLM models (AllMessageValues) or use dict-style access compatible with TypedDicts.
- For tool messages, rely on tool_call_id and tool_calls for function meta instead of a message-level name.

Follow-ups (optional)
- If we want stronger compile-time checks across boundaries, we can expand helper types around our Message to explicitly discriminate string vs. list serializer output when relevant to providers.

Co-authors
- Co-authored-by: openhands <openhands@all-hands.dev>


openhands/sdk/llm/message.py
```diff
@@ -116,7 +123,16 @@ def to_llm_dict(self) -> dict[str, Any]:
             # single string
             message_dict = self._string_serializer()
 
-        return message_dict
+        if self.role == "assistant":
+            return ChatCompletionAssistantMessage(**message_dict)
+        elif self.role == "user":
+            return ChatCompletionUserMessage(**message_dict)
+        elif self.role == "system":
+            return ChatCompletionSystemMessage(**message_dict)
```
enyst, line 131: Oh, yes, each Message does its serialization. Are you sure we really want the litellm types here?

enyst: So we had Message, and we were serializing them - and then send that, since litellm is fine with it.

Do we still want to do this if we stop doing the conversion completions -> responses?

xingyaoww: Yeah, to my understanding, the `to_llm_dict` is just supposed to be the translation between our system and Litellm's system?

enyst: Yes. You'd like that translation layer to use litellm types?

xingyaoww: Yeah, at least we have some sort of typing that's explicitly defined somewhere is better than just dicts :)

enyst:
Quick type summary of the litellm additions in this PR:

- AllMessageValues — a type alias (Union of litellm ChatCompletion* message models) that represents any Chat Completions message payload.
- ChatCompletionAssistantMessage — (BaseModel) for an assistant-role Chat Completions message.
- ChatCompletionUserMessage — (BaseModel) for a user-role Chat Completions message.
- ChatCompletionSystemMessage — (BaseModel) for a system-role Chat Completions message.
- ChatCompletionToolMessage — (BaseModel) for a tool-role Chat Completions message.
- ChatCompletionMessageToolCall — (BaseModel), a single tool/function call object embedded in an assistant message.
- ResponseInputParam — a litellm-exported type alias for the OpenAI Responses API "input" parameter (typically str | list[ResponseInputItemParam]); not a class.

PR #403, opened 2025-09-22
Solve final action concurrent message bug

### LINKED ISSUE #421: User Message sent during a conversation step that calls Finish tool breaks conversation run
If a user message is sent concurrently to an agent during a conversation while a step calls the finish action, the conversation sometimes unexpectedly continues although the finish action should end the conversation run.

Race condition test for concurrent messages during multiple tool call execution.

User messages sent concurrently with multiple tool call execution can
cause improper conversation state transitions.

Bug Description:
    When an agent executes multiple tool calls in a single response (including a finish
    action), and a user message arrives during the execution window, the message handler
    incorrectly resets the agent status from FINISHED back to IDLE. This causes the
    conversation to continue processing instead of terminating as expected.

Root Cause:
    The conversation's send_message() method doesn't properly handle the case where
    the agent has already transitioned to FINISHED state during ongoing tool execution.
    The concurrent message processing overwrites the FINISHED status, leading to
    unexpected conversation continuation.

Test Strategy:
    Uses a controlled timing scenario with a long-running sleep command followed by
    a finish action to create a predictable race condition window. A separate thread
    injects a user message during the sleep execution to trigger the bug.

Technical Details:
    - Employs threading to simulate concurrent message arrival
    - Uses mocked LLM responses to control tool call sequences
    - Measures timing to ensure proper race condition setup
    - Validates bug presence by counting unexpected LLM invocations

Expected Behavior:
    Conversation should terminate after the finish action, regardless of concurrent messages.

Actual Behavior (Bug):
    Concurrent messages reset agent status, causing additional LLM calls and continued execution.

simonrosenberg:
Fixes #421 

Summary of the bug and fix                                                                                                                                                                                                                      
 - Root cause: When a user message arrives while a run() loop is still in progress but the agent has already set state.agent_status to FINISHED (e.g., after a multi-tool response containing finish), Conversation.send_message() was resetting the status from FINISHED back to IDLE. This allowed the conversation to continue with an extra LLM call, rather than stopping as intended.                                                                                                      
 - Fix: break the conversation run from within the lock block if the breaking conditions are met                                                                                                                                                                          
                                                                                                                                                                                                                                                 
The pytest is an executable file that runs it 50 times in parallel to make sure the fix works (threading=non-deterministic behavior)

### DIFF
--- a/openhands/agent_server/event_router.py
+++ b/openhands/agent_server/event_router.py
@@ -15,7 +15,6 @@ from fastapi import (
     WebSocketDisconnect,
     status,
 )
-from fastapi.websockets import WebSocketState
 
 from openhands.agent_server.conversation_service import (
     get_default_conversation_service,
@@ -157,15 +156,20 @@ async def socket(
         _WebSocketSubscriber(websocket)
     )
     try:
-        while websocket.application_state == WebSocketState.CONNECTED:
+        while True:
             try:
                 data = await websocket.receive_json()
                 message = Message.model_validate(data)
                 await event_service.send_message(message, run=True)
             except WebSocketDisconnect:
-                await event_service.unsubscribe_from_events(subscriber_id)
-            except Exception:
+                # Exit the loop when websocket disconnects
+                return
+            except Exception as e:
                 logger.exception("error_in_subscription", stack_info=True)
+                # For critical errors that indicate the websocket is broken, exit
+                if isinstance(e, (RuntimeError, ConnectionError)):
+                    raise
+                # For other exceptions, continue the loop
     finally:
         await event_service.unsubscribe_from_events(subscriber_id)
 
--- a/scripts/websocket_client.html
+++ b/scripts/websocket_client.html
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>WebSocket Client Example</title>
+  <script>
+    let socket = null
+
+    function connect(){
+        const conversationId = document.getElementById('conversationId').value;
+        socket = new WebSocket(`ws://localhost:8000/api/conversations/${conversationId}/events/socket`);
+        
+        socket.addEventListener('open', (event) => {
+            console.log('WebSocket connection opened:', event);
+            document.getElementById('connectButton').disabled = true
+            document.getElementById('disconnectButton').disabled = false
+            document.getElementById('messageInput').disabled = false
+            document.getElementById('sendButton').disabled = false
+        });
+
+        // Event handler for receiving messages from the server
+        socket.addEventListener('message', (event) => {
+            console.log('Message from server:', event.data);
+            // You can update your UI or process the received data here
+            document.getElementById('messages').innerHTML += `<li>${event.data}</li>`;
+        });
+
+        // Event handler for when the connection is closed
+        socket.addEventListener('close', (event) => {
+            console.log('WebSocket connection closed:', event);
+            document.getElementById('connectButton').disabled = false
+            document.getElementById('disconnectButton').disabled = true
+            document.getElementById('messageInput').disabled = true
+            document.getElementById('sendButton').disabled = true
+        });
+
+        // Event handler for errors
+        socket.addEventListener('error', (event) => {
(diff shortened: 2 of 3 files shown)


openhands/sdk/conversation/conversation.py
```diff
@@ -109,9 +109,16 @@ def send_message(self, message: str | Message) -> None:
         )
         with self.state:
             if self.state.agent_status == AgentExecutionStatus.FINISHED:
-                self.state.agent_status = (
-                    AgentExecutionStatus.IDLE
-                )  # now we have a new message
+                if getattr(self.state, "_run_in_progress", False):
```
xingyaoww: This feels weird -- shouldn't we not be able to hold the lock when run is in progress? 🤔 can we ask agent to double check?

simonrosenberg:
@xingyaoww this fix works :)
I turned the test file into an executable py file that runs the test file 50 times in parallel to make sure that the test is solved. Tell me if that's OK with you or if it should be cleaned

openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -102,7 +102,7 @@ def id(self) -> ConversationID:
         return self._state.id
 
     @property
-    def state(self) -> ConversationStateProtocol:
+    def state(self) -> ConversationState:
```
simonrosenberg, line 105: @xingyaoww  ConversationStateProtocol didn't implement the lock protocol so the pre-commit would complain

```
"ConversationStateProtocol" cannot be used with "with" because it does not correctly implement __enter__
    Attribute "__enter__" is unknown (reportGeneralTypeIssues)
  /home/runner/_work/agent-sdk/agent-sdk/tests/sdk/agent/test_message_while_finishing.py:394:14 - error: Object of type "ConversationStateProtocol" cannot be used with "with" because it does not correctly implement __exit__
    Attribute "__exit__" is unknown 
```

xingyaoww: I don't think you will need this change IF you assert isinstance(conversation, LocalConversion) :)

openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -219,17 +219,15 @@ def run(self) -> None:
...
-            iteration += 1
-            if iteration >= self.max_iteration_per_run:
-                break
+                if (
+                    self.state.agent_status == AgentExecutionStatus.FINISHED
+                    or self.state.agent_status
+                    == AgentExecutionStatus.WAITING_FOR_CONFIRMATION
+                    or iteration >= self.max_iteration_per_run
+                ):
+                    break
```
xingyaoww, line 230: ah ha, so it is actually this reason

tests/sdk/agent/test_message_while_finishing.py
```diff
@@ -0,0 +1,577 @@
...
+    Message as LiteLLMMessage,
+    ModelResponse,
+)
+from pydantic import Field  # noqa: E402
+
+from openhands.sdk.agent import Agent  # noqa: E402
+from openhands.sdk.conversation import Conversation  # noqa: E402
+from openhands.sdk.event import MessageEvent  # noqa: E402
+from openhands.sdk.llm import LLM, Message, TextContent  # noqa: E402
+from openhands.sdk.tool import (  # noqa: E402
```
xingyaoww, line 57: Why we need so many noqa?

tests/sdk/agent/test_message_while_finishing.py
```diff
@@ -0,0 +1,577 @@
...
+        has_butterfly = "butterfly" in all_content
+
+        if self.step_count == 1:
+            # Step 1: Process initial request - single sleep
+            sleep_call = ChatCompletionMessageToolCall(
+                id="sleep_call_1",
+                type="function",
+                function=Function(
+                    name="sleep_tool",
+                    arguments='{"duration": 2.0, "message": "First sleep completed"}',
```
enyst, line 168: I feel like we might have troubles with timing hardcoded or expected to matter in unit tests... But I see why it seemed needed, and I don't have a better idea

PR #419, opened 2025-09-22
Enhance truncation with persistence capability

### LINKED ISSUE #301: Enhance truncation handling by saving large outputs to files
## Problem

Currently, when tool outputs exceed the maximum size limit (e.g., `MAX_BROWSER_OUTPUT_SIZE = 50000` in browser tools), we simply truncate the content using `maybe_truncate()`. This approach loses potentially valuable information that agents might need to investigate.

## Current Implementation

In `openhands/tools/browser_use/definition.py` line 32:
```python
TextContent(text=maybe_truncate(self.output, MAX_BROWSER_OUTPUT_SIZE))
```

The `maybe_truncate()` function keeps only the head and tail of the content, inserting a truncation notice in the middle.

## Proposed Enhancement

When content is too large to include directly in the observation, we should:

1. Save the full content to a file in `working_dir/.openhands/` (e.g., `working_dir/.openhands/browser_output_<timestamp>.txt`)
2. Include a reference to the saved file in the truncated observation
3. Allow agents to use other tools (like `str_replace_editor`) to investigate the full content

## Benefits

- **Preserves Information**: No data loss when outputs are large
- **Agent Accessibility**: Agents can use existing file tools to examine full content
- **Graceful Degradation**: Still provides immediate context with truncated preview
- **Debugging**: Easier to debug issues with large outputs

## Implementation Considerations

1. **File Management**: 
   - Use timestamped filenames to avoid conflicts
   - Consider cleanup strategy for old files
   - Ensure `.openhands` directory exists

2. **Modified `maybe_truncate()` Function**:
   - Add optional `working_dir` parameter
   - Add optional `save_to_file` parameter
   - Return both truncated content and file path when saved

3. **Tool Integration**:
   - Update browser tools and other tools that use truncation
   - Include file reference in observation messages

## Example Enhanced Output

```
Browser content (truncated - full content saved to .openhands/browser_output_20250916_134238.txt):

[First 1000 chars of content]
...
<content truncated - see full output in .openhands/browser_output_20250916_134238.txt>
...
[Last 1000 chars of content]
```

## Related Context

- This enhancement was suggested during PR #214 review
- Browser-use tools particularly benefit from this as they can generate large DOM/content outputs
- Other tools with potentially large outputs could also benefit

## Acceptance Criteria

- [ ] Modify `maybe_truncate()` to support file saving
- [ ] Update browser tools to use enhanced truncation
- [ ] Ensure `.openhands` directory is created if it doesn't exist
- [ ] Include file references in truncated observations
- [ ] Add tests for the new functionality
- [ ] Consider file cleanup strategy

---

**Priority**: Medium  
**Labels**: enhancement, tools, browser-use

ryanhoangt:
Fix #301

---

**Agent Server image for this PR**

Pull (multi-arch manifest):
```bash
docker pull ghcr.io/all-hands-ai/agent-server:306bcb9
```

Run:
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-306bcb9 \
  ghcr.io/all-hands-ai/agent-server:306bcb9
```

_This tag is a multi-arch manifest (amd64/arm64). Your client pulls the right arch automatically._

Pull (multi-arch manifest):
```bash
docker pull ghcr.io/all-hands-ai/agent-server:97bf4e0
```

Run:
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-97bf4e0 \
  ghcr.io/all-hands-ai/agent-server:97bf4e0
```

_This tag is a multi-arch manifest (amd64/arm64). Your client pulls the right arch automatically._

Pull (multi-arch manifest):
```bash
docker pull ghcr.io/all-hands-ai/agent-server:f281599
```

Run:
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-f281599 \
  ghcr.io/all-hands-ai/agent-server:f281599
```

_This tag is a multi-arch manifest (amd64/arm64). Your client pulls the right arch automatically._

Pull (multi-arch manifest):
```bash
docker pull ghcr.io/all-hands-ai/agent-server:9f6aba9
```

Run:
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-9f6aba9 \
  ghcr.io/all-hands-ai/agent-server:9f6aba9
```

_This tag is a multi-arch manifest (amd64/arm64). Your client pulls the right arch automatically._

Pull (multi-arch manifest):
```bash
docker pull ghcr.io/all-hands-ai/agent-server:4a9b52b
```

Run:
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-4a9b52b \
  ghcr.io/all-hands-ai/agent-server:4a9b52b
```

_This tag is a multi-arch manifest (amd64/arm64). Your client pulls the right arch automatically._

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:5a4c724-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-5a4c724-python \
  ghcr.io/openhands/agent-server:5a4c724-python
```

```
ghcr.io/openhands/agent-server:5a4c724-golang-amd64
ghcr.io/openhands/agent-server:5a4c724-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:5a4c724-golang-arm64
ghcr.io/openhands/agent-server:5a4c724-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:5a4c724-java-amd64
ghcr.io/openhands/agent-server:5a4c724-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:5a4c724-java-arm64
ghcr.io/openhands/agent-server:5a4c724-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:5a4c724-python-amd64
ghcr.io/openhands/agent-server:5a4c724-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:5a4c724-python-arm64
ghcr.io/openhands/agent-server:5a4c724-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:5a4c724-golang
ghcr.io/openhands/agent-server:5a4c724-java
ghcr.io/openhands/agent-server:5a4c724-python
```

- Each variant tag (e.g., `5a4c724-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `5a4c724-python-amd64`) are also available if needed

### DIFF
--- a/examples/01_standalone_sdk/25_agent_delegation.py
+++ b/examples/01_standalone_sdk/25_agent_delegation.py
@@ -196,4 +196,4 @@ print(f"EXAMPLE_COST (user-defined agents): {cost_2}")
 print("All done!")
 
 # Full example cost report for CI workflow
-print(f"EXAMPLE COST: {cost_1 + cost_2}")
+print(f"EXAMPLE_COST: {cost_1 + cost_2}")


openhands/sdk/utils/truncate.py
```diff
@@ -10,35 +14,76 @@
...
+    ]  # microseconds to milliseconds
+    filename = f"{tool_prefix}_output_{timestamp}.txt"
+    file_path = save_dir_path / filename
+
+    try:
+        file_path.write_text(content, encoding="utf-8")
+        return str(file_path)
+    except Exception as e:
+        # If saving fails, return a descriptive error message
+        return f"<failed to save: {e}>"
```
xingyaoww: We should just return None here

openhands/sdk/utils/truncate.py
```diff
@@ -10,35 +14,76 @@
...
+    if save_dir:
+        saved_file_path = _save_full_content(content, save_dir, tool_prefix)
+
+    # Create enhanced truncation notice with file reference if available
+    enhanced_notice = truncate_notice
+    if saved_file_path and not saved_file_path.startswith("<failed"):
+        enhanced_notice = (
+            f"<response clipped><NOTE>Due to the max output limit, only part of the "
+            f"full response has been shown to you. The complete output has been "
+            f"saved to {saved_file_path} - you can use the str_replace_editor tool "
```
xingyaoww:
```suggestion
            f"saved to {saved_file_path} - you can use other tools "
```

In cases where str replace editor does not exist

openhands/sdk/utils/truncate.py
```diff
@@ -10,35 +14,76 @@
...
+            f"to view the full content.</NOTE>"
```
xingyaoww: Ideally, we should also provide the actual line number where the truncation happens, so the agent can just take a look right away to save some context

openhands/tools/browser_use/definition.py
```diff
@@ -34,9 +31,14 @@ def agent_observation(self) -> Sequence[TextContent | ImageContent]:
...
             return [TextContent(text=f"Error: {self.error}")]
 
-        content: list[TextContent | ImageContent] = [
-            TextContent(text=maybe_truncate(self.output, MAX_BROWSER_OUTPUT_SIZE))
-        ]
+        # Use enhanced truncation with file saving, fallback to current working dir
+        truncated_text = maybe_truncate(
+            content=self.output,
+            truncate_after=DEFAULT_TEXT_CONTENT_LIMIT,
+            save_dir=os.getcwd(),
```
xingyaoww: We should accept a `save_dir` from browser tool set .create() instead of cwd so it doesn't get into VCS

openhands/tools/execute_bash/definition.py
```diff
@@ -99,7 +99,15 @@ def agent_observation(self) -> Sequence[TextContent | ImageContent]:
             ret += f"\n[Command finished with exit code {self.metadata.exit_code}]"
         if self.error:
             ret = f"[There was an error during command execution.]\n{ret}"
-        return [TextContent(text=maybe_truncate(ret, MAX_CMD_OUTPUT_SIZE))]
+
+        # Use enhanced truncation with file saving if working directory is available
+        truncated_text = maybe_truncate(
+            content=ret,
+            truncate_after=MAX_CMD_OUTPUT_SIZE,
+            save_dir=self.metadata.working_dir,
```
xingyaoww: ditto: We should accept a `save_dir` from BashTool set .create() instead of cwd so it doesn't get into VCS.

Maybe we can default it to `/tmp/.openhands`

openhands/sdk/utils/truncate.py
ryanhoangt: Sounds good, I implemented this in the latest commit.

openhands/sdk/utils/truncate.py
```diff
@@ -1,5 +1,13 @@
 """Utility functions for truncating text content."""
 
+import logging
+from datetime import datetime
+from pathlib import Path
+
+
+logger = logging.getLogger(__name__)
```
xingyaoww: You should use openhands.sdk get logger here

openhands/sdk/tool/schema.py
```diff
@@ -179,6 +179,11 @@ def visualize(self) -> Text:
 class ObservationBase(Schema, DiscriminatedUnionMixin, ABC):
     """Base schema for output observation."""
 
+    full_output_save_dir: str | None = Field(
+        default=None,
+        description="Directory where full output files are saved",
+    )
+
```
xingyaoww: I don't think we should add it here -- it will do it for ALL the observations

xingyaoww: Maybe just add it to BrowserObservation (eg where we need it)

openhands-tools/openhands/tools/browser_use/definition.py
```diff
@@ -34,9 +30,14 @@ def agent_observation(self) -> Sequence[TextContent | ImageContent]:
...
-        content: list[TextContent | ImageContent] = [
-            TextContent(text=maybe_truncate(self.output, MAX_BROWSER_OUTPUT_SIZE))
-        ]
+        # Use enhanced truncation with file saving, fallback to current working dir
+        truncated_text = maybe_truncate(
+            content=self.output,
+            truncate_after=DEFAULT_TEXT_CONTENT_LIMIT,
+            save_dir=self.full_output_save_dir,
+            tool_prefix="browser",
+        )
```
xingyaoww: Actually, i just realized that this means EVERYTIME you call .agent_observation (at each step), you will save a new file since the file name based on timestamp?

Can you write a test to ensure this doesn't happen? We should think of a different way to do this, maybe it should be named with the Observation ID/Hash?

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -99,7 +99,15 @@ def agent_observation(self) -> Sequence[TextContent | ImageContent]:
...
             ret = f"[There was an error during command execution.]\n{ret}"
-        return [TextContent(text=maybe_truncate(ret, MAX_CMD_OUTPUT_SIZE))]
+
+        # Use enhanced truncation with file saving if working directory is available
+        truncated_text = maybe_truncate(
+            content=ret,
+            truncate_after=MAX_CMD_OUTPUT_SIZE,
+            save_dir=self.full_output_save_dir,
+            tool_prefix="bash",
+        )
```
xingyaoww, line 133: ditto same here -- we will save everytime we call `.agent_observation`, ideally we should just save at the first time

openhands-tools/openhands/tools/browser_use/impl.py
```diff
@@ -193,13 +192,23 @@ async def _execute_action(self, action):
                 result = await self.close_tab(action.tab_id)
             else:
                 error_msg = f"Unsupported action type: {type(action)}"
-                return BrowserObservation(output="", error=error_msg)
+                return BrowserObservation(
+                    output="",
+                    error=error_msg,
+                    full_output_save_dir=self.full_output_save_dir,
```
xingyaoww, line 232: maybe instead of saving `save_dir` we should just put `full_output_file_path` here, and have `_execute_action` pre-generate the filename.

And when we call `.agent_observation`, we first check if the file exists or not, if not, we create it.

openhands-tools/openhands/tools/execute_bash/impl.py
```diff
@@ -91,8 +95,12 @@ def __call__(self, action: ExecuteBashAction) -> ExecuteBashObservation:
         # Apply automatic secrets masking using env_masker
         if self.env_masker and observation.output:
             masked_output = self.env_masker(observation.output)
-            data = observation.model_dump(exclude={"output"})
-            return ExecuteBashObservation(**data, output=masked_output)
+            data = observation.model_dump(exclude={"output", "full_output_save_dir"})
+            return ExecuteBashObservation(
+                **data,
+                output=masked_output,
+                full_output_save_dir=self.full_output_save_dir,
```
xingyaoww: ditto

openhands/sdk/utils/truncate.py
```diff
@@ -10,35 +18,87 @@
...
+            f"line {head_content_lines + 1}).</NOTE>"
+        )
+
+    # Calculate shifted number of tail characters
+    shifted_available_chars = truncate_after - len(enhanced_notice)
+    shifted_tail_chars = max(0, shifted_available_chars - orig_head_chars)
 
     # Keep head and tail, insert notice in the middle
-    return content[:head_chars] + truncate_notice + content[-tail_chars:]
+    return content[:orig_head_chars] + enhanced_notice + content[-shifted_tail_chars:]
```
xingyaoww: This logic is pretty hard to parse!

Consider these:

---

The current issues
	1.	Too many branches and intermediate variables (half, orig_half, head_chars, tail_chars, shifted_*) → mental overhead explodes.
	2.	Inline enhanced notice construction is bloated — it should be a helper.
	3.	File-saving logic is tangled inside the main truncation logic instead of being cleanly separated.
	4.	Double adjustment of head/tail sizes (orig_half vs shifted tail chars) → confusing and bug-prone.

⸻

How to simplify

Think in three phases:
	1.	Early exits: If no truncation or content is short enough, return immediately.
	2.	Side effect: Optionally save full content → get saved_path.
	3.	Truncate:
	•	Compute split sizes in one place.
	•	Build notice (simple, delegate to helper).
	•	Stitch result.

⸻

Sketch of simpler version

def truncate_with_notice(content, limit, notice, save_dir=None, prefix=None):
    if limit is None or len(content) <= limit or limit <= 0:
        return content

    if len(notice) >= limit:
        return notice[:limit]

    # Save full content if requested
    saved_path = None
    if save_dir:
        saved_path = _save_full_content(content, save_dir, prefix)

    # Build enhanced notice
    notice = _build_notice(notice, saved_path, content)

    # Split head/tail evenly
    available = limit - len(notice)
    head = (available + 1) // 2
    tail = available // 2

    return content[:head] + notice + content[-tail:]


def _build_notice(base, saved_path, content):
    if not saved_path:
        return base
    head_lines = len(content.splitlines()[:1])  # or smarter calc if needed
    return (
        f"{base}<NOTE>Due to max output limit, "
        f"see full content in {saved_path}, "
        f"truncated starts around line {head_lines + 1}</NOTE>"
    )


⸻

Why this is better
	•	One exit path per condition (short content, too small limit, etc.).
	•	File saving is a clean side-effect, not interleaved with truncation math.
	•	Notice building is factored out → single responsibility.
	•	Head/tail calculation is symmetrical, just one formula, no shifting mess.
	•	You can read top-to-bottom without mental gymnastics.

⸻

[Core Judgment]
✅ Worth doing: The original version is unreadable and prone to off-by-one errors. Simplified structure makes logic obvious and maintainable.

[Key Insights]
	•	Data Structure: Core data is just (content, limit, notice, saved_path). No need for extra orig_half junk.
	•	Complexity: Half the variables exist only to compensate for poor separation of concerns. Kill them.
	•	Risk: None — output format is compatible (or easily adapted).

[Linus-Style Plan]
	1.	Collapse to three logical stages (exit early → save → truncate).
	2.	Push string-construction for the notice into a helper.
	3.	Use one head/tail calculation.
	4.	Stop over-optimizing for corner cases unless they’re real.

openhands-tools/openhands/tools/browser_use/definition.py
ryanhoangt: > i just realized that this means EVERYTIME you call .agent_observation (at each step), you will save a new file since the file name based on timestamp?

~~In the `maybe_truncate` we only save when needed, i.e. when the observation length exceeds the limit -- do you think it's good enough?~~

_Edit_: I got what you're saying now, nice catch -- maybe hash of the content is good enough 🤔

ryanhoangt: @xingyaoww I just added a deduplication mechanism and some test cases for it based on your idea, though I simplified it a bit. In the `maybe_truncate` we'll calculate the hash for the given content and use that for filename, hence we can avoid writing multiple files, and we also don't need to generate file name inside tool executor implementation. Could you take a look to see if it addresses your concern?

openhands/sdk/utils/truncate.py
ryanhoangt: Yeah I agree my naming was too bad 😅 I made it more readable in the latest commit -- LMK if you find it's still too hard to read. The reason why we can't do 'Use one head/tail calculation.' is because we need to calculate `head` at the very first to use that for line number calculation inside the truncation notice. And only here we can redistribute the remaining chars for `tail`. The logic here is indeed complicated but we need it for the correct behavior.

xingyaoww:
Let's merge this after https://github.com/All-Hands-AI/agent-sdk/pull/531

That PR should give us a legit `persistence_dir` for each tool

ryanhoangt:
@xingyaoww I updated the `create` method so that: if `full_output_save_dir` is relative, it will be relative to `conv_state.persistence_dir`, and if absolute, it will be used as is. LMK if you have any concerns with it.

openhands/sdk/utils/truncate.py
```diff
@@ -10,35 +18,84 @@
...
+
+    # Determine final notice by saving file first if requested
+    final_notice = truncate_notice
+    if save_dir:
+        saved_file_path = _save_full_content(content, save_dir, tool_prefix)
+        if saved_file_path:
+            # Calculate line number where truncation happens (using head_chars)
+            head_content_lines = len(content[:head_chars].splitlines())
+
+            final_notice = (
```
xingyaoww: nit: we should move this to a global var similar to DEFAULT_TRUNCATE_NOTICE

openhands/tools/browser_use/definition.py
```diff
@@ -598,13 +605,25 @@ class BrowserToolSet(ToolBase):
     @classmethod
     def create(
         cls,
+        conv_state: "ConversationState",
+        full_output_save_dir: str = "full_tool_outputs",
```
xingyaoww: Can we remove this and just use `persistence_dir` / `env_observations`? I guess maybe we can add a new attribute of `conv_state.env_obs_persistence_dir` that uses this path?

openhands/tools/execute_bash/definition.py
```diff
@@ -272,6 +289,14 @@ def create(
         if not os.path.isdir(working_dir):
             raise ValueError(f"working_dir '{working_dir}' is not a valid directory")
 
+        # Compute full output save dir
+        if os.path.isabs(full_output_save_dir):
```
xingyaoww: ditto

openhands/tools/browser_use/definition.py
ryanhoangt: Sounds good, I added a new `env_observation_persistence_dir` attribute into `ConversationState`.

openhands-sdk/openhands/sdk/utils/truncate.py
```diff
@@ -9,36 +17,89 @@
...
-    head_chars = half + (available_chars % 2)
-    tail_chars = half
+    # Calculate tail size based on final notice (head_chars stays consistent)
+    final_available_chars = truncate_after - len(final_notice)
+    tail_chars = max(0, final_available_chars - head_chars)
 
-    # Keep head and tail, insert notice in the middle
-    return content[:head_chars] + truncate_notice + content[-tail_chars:]
+    # Assemble final result
+    return content[:head_chars] + final_notice + content[-tail_chars:]
```
xingyaoww: GPT-5 seems to think there's some issue here -- shall we try to see if we can create test to repro this edge case and fix it?

---

Short answer: not quite. There’s a correctness hole after you switch from the initial truncate_notice to the (potentially longer) final_notice with the saved file path.

The issue

You compute:
	•	available_chars = truncate_after - len(truncate_notice)
	•	head_chars = ceil(available_chars / 2)

Then, after constructing final_notice, you do:
	•	final_available_chars = truncate_after - len(final_notice)
	•	tail_chars = max(0, final_available_chars - head_chars)

If final_notice is longer than truncate_notice (very likely when it includes a path), you can end up with:

head_chars + len(final_notice) + tail_chars > truncate_after

because clamping tail_chars to 0 does not guarantee the total length is ≤ truncate_after. In other words, when final_available_chars < head_chars, the output can exceed truncate_after.

You also have an early-return guard:

if len(truncate_notice) >= truncate_after:
    return truncate_notice[:truncate_after]

but you do not repeat a similar guard for final_notice. So if final_notice alone already consumes the entire budget, the function should return a slice of it—otherwise you’ll overflow.


---

Some suggested version by GPT-5

```
def maybe_truncate(
    content: str,
    truncate_after: int | None = None,
    truncate_notice: str = DEFAULT_TRUNCATE_NOTICE,
    save_dir: str | None = None,
    tool_prefix: str = "output",
) -> str:
    """
    Truncate the middle of `content` if it exceeds `truncate_after`, keeping head and tail.
    Optionally persists the full content and references it in the notice.

    Returns:
        Original content if under limit, or truncated content with a notice (and saved-file ref if applicable).
    """
    # 1) Early exits: no truncation requested, or content already within limit.
    if truncate_after is None or truncate_after < 0 or len(content) <= truncate_after:
        return content

    # 2) If even the base notice doesn't fit, return a slice of it.
    if len(truncate_notice) >= truncate_after:
        return truncate_notice[:truncate_after]

    # 3) Pre-compute a proposed head split using the base notice for line number stability.
    base_available = truncate_after - len(truncate_notice)
    # Prefer giving the "extra" char to head (ceil split)
    proposed_head = base_available // 2 + (base_available % 2)

    # 4) Optionally save full content, then construct the final notice once.
    final_notice = truncate_notice
    if save_dir:
        saved_file_path = _save_full_content(content, save_dir, tool_prefix)
        if saved_file_path:
            # Compute the (1-based) line number where truncation starts using the proposed head
            head_content_lines = len(content[:proposed_head].splitlines())
            final_notice = DEFAULT_TRUNCATE_NOTICE_WITH_PERSIST.format(
                file_path=saved_file_path,
                line_num=head_content_lines + 1,  # indicate the next line
            )

    # 5) If the final notice (with persist info) alone fills the budget, return a slice of it.
    if len(final_notice) >= truncate_after:
        return final_notice[:truncate_after]

    # 6) Allocate remaining budget to head/tail.
    remaining = truncate_after - len(final_notice)

    # Keep the previously computed head preference, but cap if notice grew.
    head_chars = min(proposed_head, remaining)
    tail_chars = remaining - head_chars  # non-negative by construction

    # 7) Assemble result.
    if tail_chars == 0:
        return content[:head_chars] + final_notice
    return content[:head_chars] + final_notice + content[-tail_chars:]
```

>

ryanhoangt: Great catch, this is indeed a possible edge case (though it shouldn't happen when `truncate_after` is large enough to fit the persisted message). I updated the code and added a `test_maybe_truncate_persist_head_char_moves_since_remaining_less_than_proposed_head` test case for it.

xingyaoww (approved):
LGTM -- but let's hold off merge for a bit since this might impact swebench

xingyaoww:
@OpenHands please merge from main and resolve conflicts

xingyaoww:
Hey @ryanhoangt, now that we have eval pipeline ready, do you want to resolve the conflict of this PR and start eval this?

ryanhoangt:
I ran an eval on 50 instances and the results LGTM:
- This PR:
```
Total instances: 500
Instances submitted: 50
Instances completed: 50
Instances incomplete: 450
Instances resolved: 35
Instances unresolved: 15
Instances with empty patches: 0
Instances with errors: 0
Unstopped containers: 0
Unremoved images: 500
```
- `main`: (commit `e485bba` in `benchmarks` repo)
```
Total instances: 500
Instances submitted: 50
Instances completed: 50
Instances incomplete: 450
Instances resolved: 34
Instances unresolved: 16
Instances with empty patches: 0
Instances with errors: 0
Unstopped containers: 0
Unremoved images: 500
```

xingyaoww:
awesome, let's merge

PR #432, opened 2025-09-22
fix condensation issue caused by condenser view

xingyaoww:
In the condenser, we previously only checked for tool call pairs between ActionEvent and ObservationEvent.

However, there are multiple ways to produce observations for tool calls, including:
- normal action execution -> ObservationEvent
- user rejection -> UserRejectObservation
- agent error -> AgentErrorEvent

Before this PR, we only check for ObservationEvent, which means when we are on confirmation mode (e.g.,  `examples/04_confirmation_mode_example.py`) and user rejects the action -- the `ActionEvent` will be filtered out by the condenser since it cannot find any `ObservationEvent` paired with it (bc this time it is `UserRejectObservation` that is paring with it).

This PR:
- introduce a type alias `ToolObservationEventType` to group all types that could be a result of a tool call
- as well as `ToolCallEventType` that can be produced by tool call (only ActionEvent)
- And make sure we track `tool_name` and  `tool_call_id` for `AgentErrorEvent`


xingyaoww:
@openhands

here is the descritpion of this PR:

> In the condenser, we previously only checked for tool call pairs between ActionEvent and ObservationEvent.
> 
> However, there are multiple ways to produce observations for tool calls, including:
> - normal action execution -> ObservationEvent
> - user rejection -> UserRejectObservation
> - agent error -> AgentErrorEvent
> 
> Before this PR, we only check for ObservationEvent, which means when we are on confirmation mode (e.g.,  `examples/04_confirmation_mode_example.py`) and user rejects the action -- the `ActionEvent` will be filtered out by the condenser since it cannot find any `ObservationEvent` paired with it (bc this time it is `UserRejectObservation` that is paring with it).
> 
> This PR:
> - introduce a type alias `ToolObservationEventType` to group all types that could be a result of a tool call
> - as well as `ToolCallEventType` that can be produced by tool call (only ActionEvent)
> - And make sure we track `tool_name` and  `tool_call_id` for `AgentErrorEvent`

Please revert the latest commit, and try to reproduce this error by adding a new testcase, and verified that the issue is solved after the latest commit. Then commit your test case and push to this PR

csmith49:
Some notes on the main change here:

```python
ToolObservationEventType = ObservationEvent | UserRejectObservation | AgentErrorEvent
```

Are `AgentErrorEvent` objects only generated in response to tool calls? What happens if there's an error in calling `self.llm.completion`?

Maybe instead of defining the `ToolObservationEventType` as a union of existing events, these events should subclass from something that indicates that they're in response to an action event (and therefore have some `action_id` or `tool_call_id` field or something)? Based on the naming of `UserRejectObservation` I would have assumed it was subclassing from something like that.

xingyaoww:
@csmith49 addressed in https://github.com/All-Hands-AI/agent-sdk/pull/433

Issue #398, opened 2025-09-22
Browser Tools Cause Server Hanging During Import


xingyaoww:
## Summary

The OpenHands agent server hangs indefinitely during startup when browser tools are enabled, preventing the server from starting and responding to requests. This issue is caused by eager imports of the `BrowserToolExecutor` class at module import time.

## Symptoms

- **Server Hanging**: The uvicorn server hangs during startup and never reaches the "Application startup complete" message
- **No Logging**: Debug logging is not displayed for incoming requests
- **Unresponsive API**: The server cannot respond to API requests like `/tools/list`
- **Process Blocking**: The server process appears to be running but is completely unresponsive

## Environment

- **OpenHands Version**: Latest (agent-sdk worktree)
- **Python Version**: 3.12
- **Dependencies**: All standard OpenHands dependencies installed via `uv`

### 1. Enable Browser Tools
Edit `openhands/agent_server/tool_router.py`:
```python
# Change this line:
register_default_tools(enable_browser=False)
# To:
register_default_tools(enable_browser=True)
```

### 2. Start the Server
```bash
cd /path/to/openhands
PYTHONPATH=. uv run uvicorn openhands.agent_server.api:api --host 0.0.0.0 --port 9000 --log-level debug
```

### 3. Observe Hanging Behavior
The server will display initial model loading messages but then hang indefinitely:
```
[09/21/25 21:40:43] INFO     Model features for litellm_proxy/anthropic/claude-sonnet-4-20250514: ...
INFO:     Started server process [12345]
INFO:     Waiting for application startup.
# Server hangs here - never reaches "Application startup complete"
```

### 4. Test API Unresponsiveness
In another terminal, attempt to make a request:
```bash
curl -X 'GET' 'http://localhost:9000/tools/list' -H 'accept: application/json'
# This will hang indefinitely with no response
```

## Root Cause Analysis

The issue is caused by **eager imports** of `BrowserToolExecutor` at module import time in two locations:

1. **`openhands/tools/browser_use/definition.py`**: 
   - Line ~15: Direct import of `BrowserToolExecutor`
   - Used in multiple class method signatures throughout the file

2. **`openhands/tools/browser_use/__init__.py`**:
   - Line 37: `from openhands.tools.browser_use.impl import BrowserToolExecutor`

When the server attempts to register browser tools during startup, these imports trigger heavy initialization code in the browser-use library that causes the process to hang.

## Minimal Reproduction

[...]

PR #440, opened 2025-09-23
Add Chromium binary check and auto-install if not available

ryanhoangt:
Related to #398, but didn't fix it completely

### DIFF
--- a/openhands/sdk/llm/llm.py
+++ b/openhands/sdk/llm/llm.py
@@ -616,7 +616,10 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
             self.max_input_tokens = self._model_info.get("max_input_tokens")
 
         if self.max_output_tokens is None:
-            if any(m in self.model for m in ["claude-3-7-sonnet", "claude-3.7-sonnet"]):
+            if any(
+                m in self.model
+                for m in ["claude-3-7-sonnet", "claude-3.7-sonnet", "claude-sonnet-4"]
+            ):
                 self.max_output_tokens = (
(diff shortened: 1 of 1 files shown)


ryanhoangt:
Hmm that's weird, let me debug a bit more before we merge this

ryanhoangt:
@xingyaoww I've tried running `uv run agent-server` on a fresh VM on `main` (without chromium) after `make build` and it seems to be able to start the server normally. It's a bit tricky for me to repro the issue :( can you maybe delete the venv and rebuild to see if it helps 🤔

Screenshot 2025-09-23 at 22 18 29

xingyaoww:
@ryanhoangt if you can't repro, maybe adding a timeout check on the import should be sufficient for this PR.

We will install playwright to docker so as long as it runs in docker, it should be ok

ryanhoangt:
> maybe adding a timeout check on the import should be sufficient for this PR.

Sounds good, I just added a timeout mechanism for the constructor of `BrowserToolExecutor`, can you give it another try to see if the timeout does indeed happen?

xingyaoww (approved):
I've tested locally and confirmed now starting a conversation with BrowserUseToolSet without playwright throw exceptions properly!

xingyaoww:
Attached logs!

```
{
  "detail": "Internal Server Error",
  "exception": "Chromium is required for browser operations but is not installed.\n\nTo install Chromium, run one of the following commands:\n  1. Using uvx (recommended): uvx playwright install chromium --with-deps --no-shell\n  2. Using pip: pip install playwright && playwright install chromium\n\n"
}
```

PR #457, opened 2025-09-23
Removed conversation_id from event api

tofarr:
Small fix : The event service operates in the context of a single conversation - it should not have references to (potentially other) conversation ids

### DIFF
--- a/examples/04_confirmation_mode_example.py
+++ b/examples/04_confirmation_mode_example.py
@@ -2,7 +2,7 @@
 
 import os
 import signal
-from typing import Callable
+from collections.abc import Callable
 
 from pydantic import SecretStr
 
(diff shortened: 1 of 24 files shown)


openhands/agent_server/event_service.py
```diff
@@ -220,8 +220,8 @@ async def get_state(self) -> ConversationState:
             raise ValueError("inactive_service")
         return self._conversation._state
 
-    async def __aenter__(self, conversation_id: UUID):
-        await self.start(conversation_id=conversation_id)
+    async def __aenter__(self):
+        await self.start()
```
tofarr, line 224: This was very wrong - it would mean we could no longer use the object in an `async with` block

raymyers (approved):

PR #462, opened 2025-09-24
Feat: add cron job for typing quality check

simonrosenberg:


simonrosenberg:
> Actually I think this might get automatically handled by pre-commit after this PR #465?

No, this PR adds a cron job that makes this issue every week https://github.com/All-Hands-AI/agent-sdk/issues/460
"The codebase contains multiple instances of vague types and union return types that force callers to manually narrow types via if/else or assert. "
I dont think that this is detectable by static tools. Sometimes a variable should be allowed to be optional, right? But if it's supposed to be well defined during runtime it should not be optional.


My hope is that this cron-job issue does meaningful code simplifications.

simonrosenberg:
> ah got it, make sense!

And this is the "output" of the issue https://github.com/All-Hands-AI/agent-sdk/pull/461
It fixed the multiple output `create` which is good but didn't pick up the issue with `self.executor: Executor | None`
I will make the same issue exactly. tag openhands again and see if it suggests other improvements

PR #478, opened 2025-09-24
Fix: makes sure conversation stats is passed to conversation visualizer

malhotra5:
Note that cost displayed includes the cost of the current event and all those that came before it.


malhotra5:
image

PR #482, opened 2025-09-24
Refactor LLM registry to remove service_id parameter from add() method

malhotra5:
This PR 

1. Remove the default `service_id` from `LLM` to encourage explicit service scoping.

2. Drop the `service_id` parameter from `LLMRegistry`; the `LLM` already carries its own ID, eliminating redundancy 

3. Update tests and examples to set `service_id` explicitly.

#### Tradeoffs 
Users must now pass an explicit `service_id` when instantiating an LLM. Aside from this constructor change, the public API for `LLM` and `LLMRegistry` remain the same. Standalone `LLM` instances still work as before, and `LLMRegistry` continues to reject duplicate `service_ids`.

---

**Agent Server image for this PR**

Pull (multi-arch manifest):
```bash
docker pull ghcr.io/all-hands-ai/agent-server:26f6396
```

Run:
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-26f6396 \
  ghcr.io/all-hands-ai/agent-server:26f6396
```

_This tag is a multi-arch manifest (amd64/arm64). Your client pulls the right arch automatically._

### DIFF
--- a/examples/05_use_llm_registry.py
+++ b/examples/05_use_llm_registry.py
@@ -25,6 +25,7 @@ assert api_key is not None, "LITELLM_API_KEY environment variable is not set."
 
 # Create LLM instance
 main_llm = LLM(
+    service_id="test-llm",
     model="litellm_proxy/anthropic/claude-sonnet-4-20250514",
     base_url="https://llm-proxy.eval.all-hands.dev",
     api_key=SecretStr(api_key),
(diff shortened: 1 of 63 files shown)


examples/05_use_llm_registry.py
```diff
@@ -25,14 +25,15 @@
 
 # Create LLM instance
 main_llm = LLM(
+    service_id="primary-llm",
```
enyst, line 28: Why is this "primary" and the others "main", wouldn't they all be "agent" or what we call the agent llm service?

malhotra5: Ah this PR got auto merged 😢

Yes I agree this should be called "agent"

PR #484, opened 2025-09-25
Fix race condition in Metrics class when accumulating spend from completions

malhotra5:
## Problem

The `Metrics` class had a race condition when the same LLM object performs completions simultaneously. Multiple threads could call methods like `add_cost()`, `add_response_latency()`, and `add_token_usage()` concurrently, causing data corruption when accumulating spend from completions.

## Solution

Added minimal threading locks to fix the race condition:

### Changes Made

1. **Added threading support**:
   - Import `threading` module
   - Initialize `self._lock = threading.Lock()` in `__init__`

2. **Protected modifying methods with locks**:
   - `add_cost()` - protects cost accumulation
   - `add_response_latency()` - protects latency tracking  
   - `add_token_usage()` - protects token usage accumulation
   - `merge()` - protects merging operations

3. **Left read-only methods unlocked** for better performance:
   - `get_snapshot()`
   - `deep_copy()`
   - `diff()`
   - Property getters

4. **Fixed deep_copy serialization**:
   - Implemented `__getstate__` and `__setstate__` methods
   - Excludes lock from being pickled/copied
   - Creates new lock in copied instances

5. **Added comprehensive thread safety test**:
   - Tests concurrent access from multiple threads
   - Verifies data integrity under concurrent load
   - Ensures no race conditions in cost accumulation

### Implementation Details

- **Minimal locking**: Only modifying methods are protected, read-only operations remain lock-free
- **Proper serialization**: Custom `__getstate__`/`__setstate__` handle lock exclusion during deep copy
- **Thread safety**: Each copied instance gets its own lock
- **Performance**: Read operations don't acquire locks for better performance

### Testing

- All existing tests pass (52 metrics tests + 308 LLM module tests)
- New thread safety test validates concurrent access
- Pre-commit hooks pass (formatting, linting, type checking)

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)  
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing

- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes
- [x] I have run the pre-commit hooks and they pass

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/40661bb2d9114af29443af113783be9a)

### DIFF
--- a/openhands/sdk/llm/utils/metrics.py
+++ b/openhands/sdk/llm/utils/metrics.py
@@ -1,4 +1,5 @@
 import copy
+import threading
 import time
 
 from pydantic import BaseModel, Field, field_validator, model_validator
@@ -109,6 +110,10 @@ class Metrics(MetricsSnapshot):
         default_factory=list, description="List of token usage records"
     )
 
+    def __init__(self, **data):
+        super().__init__(**data)
+        self._lock = threading.Lock()
+
     @field_validator("accumulated_cost")
     @classmethod
     def validate_accumulated_cost(cls, v: float) -> float:
@@ -145,15 +150,19 @@ class Metrics(MetricsSnapshot):
     def add_cost(self, value: float) -> None:
         if value < 0:
             raise ValueError("Added cost cannot be negative.")
-        self.accumulated_cost += value
-        self.costs.append(Cost(cost=value, model=self.model_name))
+        with self._lock:
+            self.accumulated_cost += value
+            self.costs.append(Cost(cost=value, model=self.model_name))
 
     def add_response_latency(self, value: float, response_id: str) -> None:
-        self.response_latencies.append(
-            ResponseLatency(
-                latency=max(0.0, value), model=self.model_name, response_id=response_id
+        with self._lock:
+            self.response_latencies.append(
+                ResponseLatency(
+                    latency=max(0.0, value),
+                    model=self.model_name,
+                    response_id=response_id,
+                )
             )
-        )
 
     def add_token_usage(
         self,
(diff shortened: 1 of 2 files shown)


openhands/sdk/llm/utils/metrics.py
```diff
@@ -109,6 +110,10 @@ class Metrics(MetricsSnapshot):
         default_factory=list, description="List of token usage records"
     )
 
+    def __init__(self, **data):
+        super().__init__(**data)
+        self._lock = threading.Lock()
+
     @field_validator("accumulated_cost")
```
malhotra5: @openhands we should not write an __init__ like this for pydantic basemodels 

please look at how we set private attributes in `ConversationStats` and and follow a similar pattern here

openhands/sdk/llm/utils/metrics.py
```diff
@@ -242,6 +255,31 @@ def log(self) -> str:
...
+            field_value = getattr(self, field_name)
+            data[field_name] = copy.deepcopy(field_value, memo)
+
+        # Create new instance with proper Pydantic initialization
+        new_instance = self.__class__(**data)
+        return new_instance
+
     def deep_copy(self) -> "Metrics":
         """Create a deep copy of the Metrics object."""
         return copy.deepcopy(self)
```
malhotra5: @openhands can we do the following

1. get rid of `__deepcopy__`, `__getstate__`, `__setstate__`
2. make sure to rewrite `deep_copy` method so that we're using the build in pydantic method instead `self.model_copy(deep=True)`


When using the pydantic built in, make sure that the new copy has its own copy of a lock and that the private attributes (such as the lock) are not copied over or break the copying process

PR #495, opened 2025-09-25
Jps/fix system prompt litellm type leak

jpshackelford:


jpshackelford:
FYI - I haven't done my own review of this yet. I'll ping when I think it is ready...

openhands/sdk/event/llm_convertible/system.py
```diff
@@ -1,22 +1,20 @@
...
 
 class SystemPromptEvent(LLMConvertibleEvent):
     """System prompt added by the agent."""
 
     source: SourceType = "agent"
     system_prompt: TextContent = Field(..., description="The system prompt text")
-    tools: list[ChatCompletionToolParam] = Field(
-        ..., description="List of tools in OpenAI tool format"
-    )
+    tools: list[ToolBase] = Field(..., description="List of tools as ToolBase objects")
```
enyst: Great call! This is still true and we still need this:
https://github.com/OpenHands/software-agent-sdk/blob/aaa42bbdf4fa8c78d8703143c145c1e8d40c325d/openhands-sdk/openhands/sdk/event/llm_convertible/system.py#L17

The litellm `ChatCompletion` stuff also means we're messing up with types here when we use `Responses API`. I think they're typed dicts in litellm so... it "works" with dicts at runtime, but it's not great.

jpshackelford:
@enyst I have refreshed this PR. Appreciate your review if you agree it's relevant and not too disruptive at this stage.

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
```diff
@@ -1,21 +1,21 @@
...
 
 class SystemPromptEvent(LLMConvertibleEvent):
     """System prompt added by the agent."""
 
     source: SourceType = "agent"
     system_prompt: TextContent = Field(..., description="The system prompt text")
-    tools: list[ChatCompletionToolParam] = Field(
-        ..., description="List of tools in OpenAI tool format"
+    tools: list[ToolDefinition] = Field(
+        ..., description="List of tools as ToolDefinition objects"
```
enyst, line 18: I could be wrong, but it's a bit surprising to see ToolDefinition here, could we use `ToolBase` instead perhaps? Do we really have `ToolDefinition`s at runtime?

The reason I ask is that the "definition" object afaik has executor, while the other object doesn't, it's just a tool specification. This is the system prompt event... which I think doesn't or shouldn't need executors?

jpshackelford: Good point - I see the concern about executors in events. ToolBase was renamed to ToolDefinition in PR #592, so it's no longer available. The executor field is marked with exclude=True, so it doesn't get serialized when SystemPromptEvent is persisted - only the specification fields (name, description, action_type, etc.) are included. That said, if you think we should introduce a separate specification type without the executor field at all, I'm open to that - though it would add some conversion overhead.

enyst: Ahh! We had a spec type, actually... but looking back I think this PR https://github.com/OpenHands/software-agent-sdk/pull/971 removed ToolBase completely 😅 

Now `ToolDefinition` is used in llm.py too, which is another spot that doesn't, and shouldn't, have anything to do with executors.

Okay... I think we could leave it like this in this PR, actually. Maybe it was too cumbersome to use.

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
```diff
@@ -26,26 +26,22 @@ def visualize(self) -> Text:
...
-                assert "name" in tool_fn
-                assert "description" in tool_fn
-                assert "parameters" in tool_fn
-                params_str = json.dumps(tool_fn["parameters"])
+            # Use ToolBase properties directly
+            description = tool.description.split("\n")[0][:100]
+            if len(description) < len(tool.description):
+                description += "..."
+
+            content.append(f"\n  - {tool.name}: {description}\n")
```
enyst, line 34: I think what happened here is that we now make a copy only for visualization, so we don't truncate the actual data

jpshackelford: Yes! Before we had to copy the entire tool dict to avoid mutation. Now we just create a local description variable with the truncated text, since ToolDefinition is immutable. The original tool.description never gets modified!

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
```diff
@@ -26,26 +26,22 @@ def visualize(self) -> Text:
...
-                k: (v[:27] + "..." if isinstance(v, str) and len(v) > 30 else v)
-                for k, v in tool.items()
-            }
-            tool_fn = tool_display.get("function", None)
-            if tool_fn and isinstance(tool_fn, dict):
-                assert "name" in tool_fn
-                assert "description" in tool_fn
-                assert "parameters" in tool_fn
-                params_str = json.dumps(tool_fn["parameters"])
+            # Use ToolBase properties directly
```
enyst:
```suggestion
            # Use ToolDefinition properties directly
```

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
enyst: Oh, you're right - absolutely right, as our friend Sonnet would say! 😂

enyst:
@openhands Do a /codereview-roasted on this PR. Investigate, understand the execution paths and be clear if you see any potential issues. Respond with markdown, including headings, bold, and other visually separating items so that it's clear and easy to read.

jpshackelford:
@enyst Thanks again for all of your help with this one!

PR #504, opened 2025-09-26
Add dynamic router for manual model selection

### LINKED ISSUE #416: Support switching LLM models in the middle of conversation
Maybe we can re-use some pattern in LLM Router to support this https://github.com/All-Hands-AI/agent-sdk/pull/152

We can probably add a new method "/conversation/update_llm" endpoint to the server.

# One way we could implement this

Similar to this example

https://github.com/All-Hands-AI/agent-sdk/blob/a5f7e71269b32fd36b9ad7d3843d281288b24f7b/examples/10_persistence.py#L83-L92

```python

agent = Agent(llm=llm, tools=tools)
conversation = Conversation(
    agent=agent,
    callbacks=[conversation_callback],
    persist_filestore=file_store,
    conversation_id=conversation_id,
)
# we do stuff ...
del conversation

# <--- HERE WE DECIDE TO SWITCH
agent2 = Agent(llm=llm2, tools=tools)
conversation = Conversation(
    agent=agent,
    callbacks=[conversation_callback],
    persist_filestore=file_store,
    conversation_id=conversation_id,
)
```

This will currently throw an error due to `resolve_diff_from_deserialized`:

https://github.com/All-Hands-AI/agent-sdk/blob/a5f7e71269b32fd36b9ad7d3843d281288b24f7b/openhands/sdk/llm/llm.py#L841-L875

We can implement `resolve_diff_from_deserialized` differently in `LLMRouter` -- it should allow us to set a different LLM on the fly:
1. We start a conversation with `LLMRouter` with ONE model instead of `LLM`
2. User sets new LLM config
3. `LLMRouter` save the LLMConfig, set it to the current LLM service (LLMRouter should be the serializable class that contains all information about the history of LLM switch actions)

ryanhoangt:
Fix #416

### DIFF
--- a/examples/24_llm_manual_switch.py
+++ b/examples/24_llm_manual_switch.py
@@ -1,0 +1,195 @@
+import os
+import uuid
+
+from pydantic import SecretStr
+
+from openhands.sdk import (
+    LLM,
+    Agent,
+    Conversation,
+    EventBase,
+    LLMConvertibleEvent,
+    LocalFileStore,
+    Message,
+    TextContent,
+    get_logger,
+)
+from openhands.sdk.llm.router.impl.dynamic import DynamicRouter
+from openhands.sdk.preset.default import get_default_tools
+
+
+logger = get_logger(__name__)
+
+# Configure initial LLM
+api_key = os.getenv("LITELLM_API_KEY")
+assert api_key is not None, "LITELLM_API_KEY environment variable is not set."
+
+# Create initial LLM
+initial_llm = LLM(
+    service_id="agent-initial",
+    model="litellm_proxy/anthropic/claude-sonnet-4-20250514",
+    base_url="https://llm-proxy.eval.all-hands.dev",
+    api_key=SecretStr(api_key),
+)
+
+# Create DynamicRouter with initial LLM
+dynamic_router = DynamicRouter(
+    service_id="dynamic-router", llms_for_routing={"claude": initial_llm}
+)
+
+# Tools
+cwd = os.getcwd()
+tools = get_default_tools(working_dir=cwd)
+
+# Agent with dynamic router
+agent = Agent(llm=dynamic_router, tools=tools)
+
+llm_messages = []  # collect raw LLM messages
+
+
+def conversation_callback(event: EventBase):
+    if isinstance(event, LLMConvertibleEvent):
+        llm_messages.append(event.to_llm_message())
+
+
+# Set up conversation with persistence for serialization demo
+conversation_id = uuid.uuid4()
+file_store = LocalFileStore(f"./.conversations/{conversation_id}")
+
+conversation = Conversation(
+    agent=agent,
+    callbacks=[conversation_callback],
+    persist_filestore=file_store,
+    conversation_id=conversation_id,
+)
+
+print(f"Starting with LLM: {dynamic_router.get_current_llm_name()}")
+print(f"Available LLMs: {list(dynamic_router.get_available_llms().keys())}")
+
+# First interaction with Claude
+conversation.send_message(
+    message=Message(
+        role="user",
+        content=[TextContent(text="Hi there!")],
+    )
+)
+conversation.run()
+
+print("=" * 50)
+print("Adding GPT-4 dynamically and switching to it...")
+
+# Dynamically add GPT-4 and switch to it
+success = dynamic_router.switch_to_llm(
+    "gpt4",
+    model="litellm_proxy/openai/gpt-4o",
+    base_url="https://llm-proxy.eval.all-hands.dev",
+    api_key=api_key,
+    temperature=0.3,
+)
+print(f"GPT-4 added successfully: {success}")
+print(f"Current LLM: {dynamic_router.get_current_llm_name()}")
+print(f"Available LLMs: {list(dynamic_router.get_available_llms().keys())}")
+print()
+
+# Second interaction with GPT-4
+conversation.send_message(
+    message=Message(
+        role="user",
+        content=[TextContent(text="Who trained you as an LLM?")],
+    )
+)
+conversation.run()
+
+print("Adding a smaller model for simple tasks...")
+
+# Add a smaller model for simple tasks
+success = dynamic_router.switch_to_llm(
+    "small_model",
+    model="litellm_proxy/mistral/devstral-small-2507",
+    base_url="https://llm-proxy.eval.all-hands.dev",
+    api_key=api_key,
+    temperature=0.1,
+)
+print(f"Small model added successfully: {success}")
+print(f"Current LLM: {dynamic_router.get_current_llm_name()}")
+print(f"Available LLMs: {list(dynamic_router.get_available_llms().keys())}")
+
+# Third interaction with small model
+conversation.send_message(
+    message=Message(
+        role="user",
+        content=[TextContent(text="Who trained you as an LLM?")],
+    )
+)
+conversation.run()
+
+print("Switching back to Claude for complex reasoning...")
+
+# Switch back to Claude for complex task
+dynamic_router.switch_to_llm("claude")
+print(f"Current LLM: {dynamic_router.get_current_llm_name()}")
+
+conversation.send_message(
+    message=Message(
+        role="user",
+        content=[
+            TextContent(
+                text="Explain the concept of dynamic programming in one sentence."
+            )
+        ],
+    )
+)
+conversation.run()
+
+print("Demonstrating persistence with LLM switching...")
+
+# Show current state before serialization
+print(f"Before serialization - Current LLM: {dynamic_router.get_current_llm_name()}")
+print(f"Available LLMs: {list(dynamic_router.get_available_llms().keys())}")
+
+# Delete conversation to simulate restart
+del conversation
+
+# Recreate conversation from persistence
+print("Recreating conversation from persistence...")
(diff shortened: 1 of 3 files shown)


openhands/sdk/llm/llm.py
```diff
@@ -96,6 +96,9 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
     # =========================================================================
     # Config fields
     # =========================================================================
+    kind: Literal["LLM"] = Field(
+        default="LLM", description="Discriminator for LLM subclasses"
+    )
```
xingyaoww: Shall we inherit LLM from DiscriminativeUnionMixin to replace this?

openhands/sdk/agent/base.py
```diff
@@ -37,17 +37,21 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
...
-        examples=[
-            {
-                "model": "litellm_proxy/anthropic/claude-sonnet-4-5-20250929",
-                "base_url": "https://llm-proxy.eval.all-hands.dev",
-                "api_key": "your_api_key_here",
-            }
-        ],
-    )
+    llm: Annotated[
+        LLM | MultimodalRouter | DynamicRouter,
```
xingyaoww: If we use DiscriminativeUnionMixin, we probably don't need this

But we probably need an `LLMBase` which is an ABC. You can check AgentBase and Agent definition to understand how we use this in other part of the codebase

ryanhoangt: I created a new `LLMBase` class and make `LLM` inherits it, which allows us to get rid of the discriminator field above, but it requires changes to lots of other places... Could you help me take a look to see if my changes make sense?

examples/01_standalone_sdk/19_llm_routing.py
```diff
@@ -54,8 +56,15 @@ def conversation_callback(event: Event):
         llm_messages.append(event.to_llm_message())
 
 
+conversation_id = uuid.uuid4()
+file_store = LocalFileStore(f"./.conversations/{conversation_id}")
```
xingyaoww: We are no longer using file_store, so it is ok to remove it

examples/01_standalone_sdk/25_llm_manual_switch.py
```diff
@@ -0,0 +1,202 @@
...
+initial_llm = LLM(
+    service_id="agent-initial",
+    model="litellm_proxy/anthropic/claude-sonnet-4-20250514",
+    base_url="https://llm-proxy.eval.all-hands.dev",
+    api_key=SecretStr(api_key),
+)
+
+# Create DynamicRouter with initial LLM
+dynamic_router = DynamicRouter(
+    service_id="dynamic-router", llms_for_routing={"primary": initial_llm}
```
xingyaoww: Let's add two models at the beginning maybe?

examples/01_standalone_sdk/25_llm_manual_switch.py
xingyaoww, line 1: Let's break this down to multiple smaller examples.

openhands-sdk/openhands/sdk/llm/router/base.py
```diff
@@ -30,11 +31,12 @@ class RouterLLM(LLM):
     """
 
     router_name: str = Field(default="base_router", description="Name of the router")
-    llms_for_routing: dict[str, LLM] = Field(
+    llms_for_routing: dict[str, LLMBase] = Field(
```
xingyaoww, line 36: Let's confirm this will be persisted correctly

xingyaoww: Let's use an immutable mapping for this:

from pydantic import BaseModel, ConfigDict
from types import MappingProxyType

class MyModel(BaseModel):
    model_config = ConfigDict(frozen=True)
    data: dict[str, int]

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        object.__setattr__(self, "data", MappingProxyType(self.data))

m = MyModel(data={"a": 1})
m.data["a"] = 2  # ❌ TypeError: 'mappingproxy' object does not support item assignment

ryanhoangt: I updated the example to assert for the correct active model and add immutability for `llms_for_routing`!

examples/01_standalone_sdk/25_llm_manual_switch.py
```diff
@@ -0,0 +1,158 @@
...
+conversation.send_message(
+    message=Message(
+        role="user",
+        content=[TextContent(text="What did we talk about earlier?")],
+    )
+)
+conversation.run()
+
+# Switch back to primary model for complex task
+print("Switching back to claude for complex reasoning...")
```
enyst, line 135: I wonder how would the user know in advance that they'll need to switch?

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -98,7 +100,7 @@
 )
 
 
-class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
+class LLMBase(DiscriminatedUnionMixin, RetryMixin, NonNativeToolCallingMixin):
```
enyst, line 103: Could you please explain, why is LLM a `DiscriminatedUnionMixin`?

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -886,6 +785,9 @@ def format_messages_for_llm(self, messages: list[Message]) -> list[dict]:
             message.cache_enabled = self.is_caching_prompt_active()
             message.vision_enabled = self.vision_is_active()
             message.function_calling_enabled = self.is_function_calling_active()
+            message.extended_thinking_enabled = get_features(
+                self.model
+            ).supports_extended_thinking
```
enyst, line 790: Sorry, is this a different issue than the PR is about?

openhands-sdk/openhands/sdk/llm/router/base.py
```diff
@@ -46,11 +50,23 @@ def validate_llms_not_empty(cls, v):
             )
         return v
 
+    @model_validator(mode="after")
+    def make_immutable(self):
+        object.__setattr__(
+            self, "llms_for_routing", MappingProxyType(self.llms_for_routing)
```
enyst, line 56: Why do we expressly make them immutable when... they weren't?
Do I understand correctly, that will mean that the user cannot decide at runtime to just switch to another LLM they pick up, let's say, from `openhands` provider? (where they already had an API key)

openhands-sdk/openhands/sdk/llm/router/base.py
```diff
@@ -112,3 +128,76 @@ def set_placeholder_model(cls, data):
...
+                raise ValueError(
+                    f"Cannot resolve_diff_from_deserialized: persisted LLM is not a "
+                    "RouterLLM and doesn't contain router data. Got "
+                    f"{persisted.__class__}"
+                )
+            # Try to reconstruct as the correct RouterLLM subclass
+            persisted = self.__class__.model_validate(persisted_dict)
+
+        # Check classes match exactly
+        if type(persisted) is not type(self):
```
enyst, line 164: TBH I don't know, I don't really like that we use this code, which "sets things in stone", when what we'd like to implement in this PR is precisely _flexibility_? It seems a bit contradictory, and it's getting too complex maybe... I'm following up

examples/01_standalone_sdk/25_llm_manual_switch.py
xingyaoww: I think we were designing the `DynamicRouter` as a way for the user to "manually select model name from drop down"?

And further down the road, we should by default initialize LLM as `DynamicRouter` in our GUI/CLI if we have UI for switching model on the fly

openhands-sdk/openhands/sdk/llm/llm.py
ryanhoangt: We use it to distinguish between router classes and normal `LLM` when deserializing, without it we'll need an explicit `kind` field.

openhands-sdk/openhands/sdk/llm/llm.py
ryanhoangt: Ah for trajectories with a single LLM, we don't need this field, but it requires when using an extending thinking LLM like Sonnet 4.5 with another non-thinking LLM. I ran into the error earlier when testing and adding this field to `Message` seems to fix it.

openhands-sdk/openhands/sdk/llm/router/base.py
ryanhoangt: Actually my earlier implementation _did_ allow user to add a new LLM config to the router during runtime, but I remember @xingyaoww said it might lead to some weird issues if we modify configs like that, ideally we want immutability when possible. I think it boils down to what "UX" we want to achieve, whether to pre-define all LLMs in advance or add a new LLM during the session. After thinking a bit more, I feel like asking users to set some LLMs they prefer to use in advance (e.g. a set of Sonnet 4.5, Codex and GLM-4.6) and switching between them later seem to be more convenient and not too rigid, I imagine myself also too lazy to manually input LLM configs within sessions many times 🤔

neubig:
Closing since we haven't been able to find time to make progress on this.

PR #511, opened 2025-09-26
feat: Add generic environment variable configuration parsing for Pydantic models

tofarr:
## Summary

Adds a generic environment variable parsing system for Pydantic models with comprehensive test coverage.

## Key Features

- **Generic env_parser.py module**: Supports parsing any Pydantic model from environment variables
- **Multiple parser types**: Bool, int, float, str, dict, list, union, and nested model parsers
- **Recursive parsing**: Handles self-referential models and complex nested structures
- **Infinite recursion protection**: Prevents parsing loops in recursive models
- **Comprehensive test suite**: 30 tests covering all parser types, Config class parsing, and edge cases
- **Union type support**: Handles Optional fields and union types properly

## Testing

All 30 tests pass, covering basic parsers, complex parsers, Config class with webhook specs, recursive NodeModel parsing, and error handling scenarios.

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/all-hands-ai/agent-server:c6b2dae-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-c6b2dae-python \
  ghcr.io/all-hands-ai/agent-server:c6b2dae-python
```

```
ghcr.io/all-hands-ai/agent-server:c6b2dae-golang
ghcr.io/all-hands-ai/agent-server:c6b2dae-java
ghcr.io/all-hands-ai/agent-server:c6b2dae-python
```

_The `c6b2dae` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._


openhands/agent_server/env_parser.py
```diff
@@ -0,0 +1,277 @@
...
+                result = {}
+            existing_field_value = result.get(field_name, MISSING)  # type: ignore
+            new_field_value = merge(existing_field_value, field_value)
+            if new_field_value is not MISSING:
+                result[field_name] = new_field_value  # type: ignore
+
+        return result
+
+
+@dataclass
```
enyst: Tiny nit: is there a reason why some are dataclasses and some not? 🤔

openhands/agent_server/env_parser.py
```diff
@@ -0,0 +1,277 @@
...
+        assert args[0] is str
+        assert args[1] in (str, int, float, bool)
+        return DictEnvParser()
+    if origin is Literal:
+        return StrEnvParser()
+
+    if origin and issubclass(origin, BaseModel):
+        target_type = origin
+    if issubclass(target_type, BaseModel):  # type: ignore
+        delayed = DelayedParser()
```
enyst, line 238: Do we have variables with this level of complexity, I thought they were primitives, or am I missing something?

tofarr: We have a higher level of complexity in the app server code, where I want to reuse this.

openhands/agent_server/env_parser.py
tofarr: DictEnvParser should not be a dataclass because it has no fields! Generally I'll use dataclasses because:

* The syntax is slightly shorter than using an __init__ method
* It adds a __repr__ method which is useful when stepping through with the debugger
* It adds a meaningful __eq__ should I need it

tofarr: My rule of thumb is: "If it has fields and does not have complex init logic, make it a dataclass unless there is a performance reason not to do so.". I am open to other opinions.

enyst: That makes sense, thank you, with you. I love dataclasses. Design-wise they're supposed to be a bit dumb, just hold their data; some opinions go as far as say they should have no logic at all. Personally I think converting data, getting data, and a bunch of work with those fields to make the data useful elsewhere actually belongs in there nicely, less headaches this way.

enyst (approved):
Thank you for this, and sorry for the delay. I looked around a bit to try to wrap my slow head around where we're going on this.

I think you're right, no config file is better. Some of these vars made me wonder why they're _here_ frozen when the local user is able to modify/override them, but I guess an answer is as simple as: modify all you want, but not at runtime. Not at agent-server runtime, at least.

How does client code / user uses/needs these vars? Apart from app servers and direct programmatic use, I've just restarted working on a vscode extension (not cli, full tab in the editor). I love doing this because it helps me see the needs directly. 😊

It needs
- start, reconnect, pause the agent-server
- override workspace path and disable vscode; maybe more.

So it needs one way to set them before starting the agent-server. It... shouldn't offer some user-facing ability to set workspace, I think, since we're in vscode, it picks that up. And even if it did allow the user any of these vars, it still doesn't make sense _at runtime_. 🤷🏻‍♂️

Still, these vars are user-modifiable. OK, so they may be in a config file somewhere, an OH app server `settings.json`? That's outside agent-server... 🤔 So in this PR, we get less redundancy/duplicate stuff.

Thank you, I like the idea. But please do tell what I'm missing, WDYT is wrong in these thoughts, just strike it down.

openhands/agent_server/env_parser.py
enyst: Sorry, I don't know if I understand, do you mean reuse them programmatically, like literally import from agent-server and use it?

tofarr:
@enyst - I think if folks really want a config file dotenv can be used to facilitate this. I have not needed it so far - here are 2 of my use cases...

**1. An excerpt from my local `launch.json`**
For easier debugging, I use and environment variables to include the UI I built:
```
...
{
  "name": "Agent Server",
  "type": "debugpy",
  "request": "launch",
  "module": "openhands.agent_server",
  "python": "${workspaceFolder}/.venv/bin/python",
  "args": ["--reload"],
  "env": {
    "DEBUG": "1",
    "OH_STATIC_FILES_PATH": "scripts/agent_server_ui/static",
  },
  "justMyCode": false
}
```

**2. The App Server**
This is still very much a work in progress, but we define a session api key and callback url for the Agent Server to use and pass it into the docker container as environment variables:

https://github.com/All-Hands-AI/OpenHands-Server/blob/d2d534b9d3c027c6ef61fa52f1e8763f8882b6d4/openhands_server/sandbox/docker_sandbox_service.py#L242

PR #515, opened 2025-09-26
Support VNC inside agent-server - allows for interactive browser!!

xingyaoww:
This PR:
- Install VNC in the agent-server Dockerfile
- Install system-level chromium for browser-use (instead of `uvx playwright install chromium --with-deps --no-shell`
- Add Desktop Service that allows users to directly connect to this VNC Desktop and see what the agent is doing with the browser!

image

image

Demo video!

https://github.com/user-attachments/assets/2cd5d08a-043e-4ce1-9d10-5ab1289faa12


openhands/agent_server/desktop_router.py
```diff
@@ -0,0 +1,47 @@
...
+desktop_router = APIRouter(prefix="/desktop", tags=["Desktop"])
+
+
+class DesktopUrlResponse(BaseModel):
+    """Response model for Desktop URL."""
+
+    url: str | None
+
+
+@desktop_router.get("/url", response_model=DesktopUrlResponse)
```
rbren, line 21: I think we need to add `/api` to ensure this is authenticated

xingyaoww: Yep i think we do

It is part of the API Router
https://github.com/All-Hands-AI/agent-sdk/pull/515/files#diff-5d96de9766018f0ed11b03e7189c53cb2ea9f735a1f440f4817926d59bd605ebR113-R133

xingyaoww:
@OpenHands please fix the failing actions on PR #515 at branch `xw/vnc`

Only fix the test, don't change the implementation

xingyaoww:
Regarding image size..
For some reason, adding VNC inside the agent-server actually reduces the size of the image compared to the main.

I suspect it is due to the fact that we don't really install Chromium through Playwright, but rather through apt. Maybe this saves some resources here

```
on ⛵ kind-kind worktree2 on  xw/vnc [$] via 🐍 v3.12.11 (worktree2) on ☁️  xingyao@all-hands.dev(us-central1) took 2s 
❯ docker image inspect ghcr.io/all-hands-ai/agent-server:8daf576-python --format='{{.Size}}' | awk '{printf "%.2f GB\n", $1/1024/1024/1024}'
3.29 GB
<-- this is main

on ⛵ kind-kind worktree2 on  xw/vnc [$] via 🐍 v3.12.11 (worktree2) on ☁️  xingyao@all-hands.dev(us-central1) 
❯ docker image inspect ghcr.io/all-hands-ai/agent-server:fa40a8b-python --format='{{.Size}}' | awk '{printf "%.2f GB\n", $1/1024/1024/1024}'
2.74 GB
<--- this is current PR
```

xingyaoww:
https://github.com/OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/agent-sdk/pull/515 at branch xw/vnc

Only fix the test, don't change the implementation

xingyaoww:
@OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/agent-sdk/pull/515 at branch xw/vnc

Only fix the test, don't change the implementation

openhands/agent_server/docker/Dockerfile
```diff
@@ -136,16 +136,39 @@ RUN mkdir -p /etc/docker && \
...
-# Install Chromium for browser-use tools
+# --- VNC + Desktop + noVNC ---
 RUN set -eux; \
-    uvx playwright install chromium --with-deps; \
-    apt-get clean; \
-    rm -rf /var/lib/apt/lists/*
+  apt-get update; \
+  apt-get install -y --no-install-recommends \
+    # GUI bits (remove entirely if headless)
+    tigervnc-standalone-server xfce4 dbus-x11 novnc websockify \
```
enyst, line 144: Just out of curiosity, isn't xfce larger than some other options, like lxde or fluxbox, maybe?

openhands/agent_server/desktop_service.py
```diff
@@ -0,0 +1,213 @@
...
+# ------- module-level accessor -------
+
+_desktop_service: DesktopService | None = None
+
+
+def get_desktop_service() -> DesktopService | None:
+    """Get the desktop service instance if VNC is enabled."""
+    global _desktop_service
+    if _desktop_service is None:
+        from openhands.agent_server.config import get_default_config
```
enyst: Not necessarily an issue for this PR, I'm a bit curious about this inline import. Would it lead to a circular import if it was top-level?

xingyaoww: @OpenHands let's move this to import at the top

openhands/agent_server/docker/Dockerfile
xingyaoww: Not really an expert here - i guess xfce is just the package LLMs are super familar with so we go with this 😓

xingyaoww:
@OpenHands please fix the failing actions on PR #515 at branch `xw/vnc`

xingyaoww:
@OpenHands please fix the failing actions on PR #515 at branch `xw/vnc`

openhands/tools/browser_use/impl.py
```diff
@@ -124,13 +155,21 @@ def __init__(
         """
 
         def init_logic():
-            _ensure_chromium_available()
+            nonlocal headless
+            executable_path = _ensure_chromium_available()
             self._server = CustomBrowserUseServer(
                 session_timeout_minutes=session_timeout_minutes,
             )
+            if _is_vnc_environment():
```
ryanhoangt: I think we should not override `headless` here, e.g. when people with their machine having VNC might still want to run in headless mode.

openhands/tools/browser_use/impl.py
```diff
@@ -2,57 +2,87 @@
...
+    Returns:
+        True if VNC environment is detected, False otherwise
+    """
+    # Check for DISPLAY environment variable (indicates X11 is available)
+    display = os.environ.get("DISPLAY")
+    if display and display.startswith(":"):
+        return True
+
+    # Also check for common VNC environment indicators
+    if os.path.exists("/tmp/.X11-unix") or os.path.exists("/home/openhands/.vnc"):
```
ryanhoangt: For some reason, this returns True on CI, which makes this test fail 🤔

https://github.com/All-Hands-AI/agent-sdk/blob/693947c8aeb81991be677de4c30062161453ead4/tests/tools/browser_use/test_browser_executor.py#L19-L27

openhands/tools/browser_use/impl.py
ryanhoangt: Maybe a better approach is to use an `OPENHANDS_FORCE_HEADLESS` env variable instead of these detection heuristics? WDYT?

xingyaoww: Sounds reasonable!

openhands/tools/browser_use/impl.py
```diff
@@ -124,13 +158,21 @@ def __init__(
...
+            nonlocal headless
+            executable_path = _ensure_chromium_available()
             self._server = CustomBrowserUseServer(
                 session_timeout_minutes=session_timeout_minutes,
             )
+            if _is_vnc_environment():
+                logger.info(
+                    "VNC environment detected - running browser in non-headless mode"
+                )
+                headless = False
```
xingyaoww: @OpenHands please add a test for this, and make sure all tests and pre-commit pass

xingyaoww:
@OpenHands please fix the failing actions on PR #515 at branch `xw/vnc`

PR #526, opened 2025-09-28
Remove load_from_toml function and tomli dependency

xingyaoww:
- Remove unused load_from_toml function from LLM class
- Remove tomli dependency from pyproject.toml since Python 3.12+ has built-in tomllib
- Function was not used anywhere in the codebase
- All tests pass after removal

### DIFF
--- a/openhands/sdk/pyproject.toml
+++ b/openhands/sdk/pyproject.toml
@@ -11,7 +11,7 @@ dependencies = [
     "python-frontmatter>=1.1.0",
     "python-json-logger>=3.3.0",
     "tenacity>=9.1.2",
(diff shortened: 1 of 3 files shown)


openhands/sdk/pyproject.toml
```diff
@@ -11,7 +11,7 @@ dependencies = [
     "python-frontmatter>=1.1.0",
     "python-json-logger>=3.3.0",
     "tenacity>=9.1.2",
-    "tomli>=1.2.0; python_version<'3.11'",
+
```
xingyaoww:
```suggestion
```

PR #532, opened 2025-09-28
Refactor: Move presets to tools package and integrate LLM metadata

xingyaoww:
## Summary

This PR addresses the architectural concerns raised in the repository by:

1. **Moving presets from SDK to tools package**: Moved `openhands/sdk/preset/` to `openhands/tools/preset/` to eliminate the circular dependency where the SDK package was importing from the tools package.

2. **Integrating LLM metadata into LLM class**: Moved the `get_llm_metadata` function from `openhands/sdk/llm/metadata.py` into the `LLM` class as both a method and a `metadata` field, providing better encapsulation and eliminating the need for a separate metadata module.

### Preset Migration
- Moved all preset files from `openhands/sdk/preset/` to `openhands/tools/preset/`
- Updated all imports from `openhands.sdk.preset` to `openhands.tools.preset` across:
  - Examples (7 files updated)
  - Tests (moved and updated test files)
  - Core modules (`tool_router.py`)
- Removed old preset directory structure

### LLM Metadata Integration
- Added `metadata: dict` field to `LLM` class with `default_factory=dict`
- Added `get_llm_metadata(agent_name, session_id=None, user_id=None)` method to `LLM` class
- Updated usage in:
  - `condenser/llm_summarizing_condenser.py`: Uses `self.llm.get_llm_metadata(agent_name="condenser")`
  - `agent/agent.py`: Uses `self.llm.get_llm_metadata(agent_name=self.name)`
- Updated tests to use the new LLM class method
- Removed `openhands/sdk/llm/metadata.py` file

## Testing

- All existing tests pass (1081 tests)
- Updated `test_get_llm_metadata.py` to test the new LLM class method
- Added test for the instance `metadata` field
- Pre-commit hooks pass

## Architecture Benefits

- **Eliminates circular dependency**: SDK no longer imports from tools package
- **Better encapsulation**: LLM metadata is now part of the LLM class where it logically belongs
- **Cleaner imports**: All preset imports now use `from openhands.tools.preset import xyz`
- **Maintains backward compatibility**: All functionality preserved with improved organization


openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -67,11 +66,7 @@ def get_condensation(self, view: View) -> Condensation:
 
         llm_response = self.llm.completion(
             messages=messages,
-            extra_body={
-                "metadata": get_llm_metadata(
-                    model_name=self.llm.model, agent_name="condenser"
-                )
-            },
+            extra_body={"metadata": self.llm.get_llm_metadata(agent_name="condenser")},
```
enyst: 😅 This looks funny, `agent_name=condenser`

I think ‘agent’ and ‘condenser’ in these logs is maybe the service_id, actually, so llm already has it as required attribute, so maybe we don’t need these at all?

openhands/sdk/llm/llm.py
```diff
@@ -344,6 +348,52 @@ def restore_metrics(self, metrics: Metrics) -> None:
...
+            user_id: Optional user identifier
+
+        Returns:
+            Dictionary containing metadata for the LLM request
+        """
+        import openhands.sdk
+
+        openhands_tools_version: str = "n/a"
+        try:
+            import openhands.tools
```
enyst: Could we log this version elsewhere… in tools package maybe, rather than `import openhands.tools` in order to log?

openhands/sdk/context/condenser/llm_summarizing_condenser.py
xingyaoww: Can you check latest commit? My idea was that the creator of LLM instance should be responsible for setting these, not the LLM class itself

xingyaoww:
@OpenHands please review diff against main for this PR.

Then clone All-Hands-AI/OpenHands repo, check to v1 branch and go to openhands-cli folder, and review all the usage of preset there, and change the imports accordingly based on the change of this PR. You should also bump the commits of agent SDK and tools to the latest commit in this PR in pyproject.toml and then re-run uv lock and commit these changes.

Also, you should pass correct metadata to the LLM initialization inside the OpenHands CLI, do track things like conversation ids, agent name, SDK and OpenHands-tools version like what we did before (and removed in this PR). You should set metadata of LLM to the exact same format but with actual content.

Please create a PR to merge these changes to the "v1" branch when you are done

PR #550, opened 2025-09-29
Refactor: add MessageToolCall class instead of litellm's ChatCompletionMessageToolCall

### LINKED ISSUE #547: Refactor: add class MessageToolCall instead of litellm's ChatCompletionMessageToolCall
The code currently relies on ChatCompletionMessageToolCall which is imported in 3 files agent.py action.py and message.py and in all the tests.

It feels like there should be an openhands interface to separate litellm's implementation with the use of tool calls in the repository.
1. You should add a MessageToolCall class in message.py that contains all the required fields to successfully call a tool. 
2. ChatCompletionMessageToolCall should only be imported in message.py
3. Carefully read through agent.py and action.py to understand exactly how the fields/methods in ChatCompletionMessageToolCall are used and add all useful fields/methods on MessageToolCall
4. Make sure to add meaningful docstring on MessageToolCall

Make sure all checks and tests are passing

simonrosenberg:
## Summary

This PR addresses issue #547 by refactoring the codebase to use a custom `MessageToolCall` class instead of directly relying on litellm's `ChatCompletionMessageToolCall`. This creates a cleaner separation between the OpenHands interface and litellm's implementation.

### Core Implementation
- **Added `MessageToolCall` class** in `message.py` with all required fields for tool calls
- **Added `MessageToolCallFunction` class** to represent the function component
- **Implemented dictionary-style access methods** (`__contains__`, `get`, `__getitem__`, `__setitem__`) for backward compatibility
- **Added bidirectional conversion methods**:
  - `from_litellm_tool_call()` - converts from litellm's format to our format
  - `to_litellm_tool_call()` - converts from our format to litellm's format

### Import Refactoring
- **Removed `ChatCompletionMessageToolCall` imports** from `agent.py` and `action.py`
- **Updated type annotations** throughout the codebase to use `MessageToolCall`
- **Centralized litellm dependency** - `ChatCompletionMessageToolCall` is now only imported in `message.py`

### Test Updates
- **Updated all test files** to use the new `MessageToolCall` classes
- **Added proper conversion calls** in mock responses using `.to_litellm_tool_call()`
- **Fixed type annotations** in test files to use the new classes
- **Maintained test coverage** for all existing functionality

### Backward Compatibility
- **Preserved existing serialization behavior** through dictionary-style access
- **Maintained API compatibility** for existing code that accesses tool call fields
- **Ensured seamless conversion** between internal and litellm formats

## Testing

- ✅ All 756 SDK tests pass
- ✅ All 23 cross-integration tests pass  
- ✅ All 102+ tools tests pass
- ✅ Pre-commit hooks pass (formatting, linting, type checking)

## Documentation

Added comprehensive docstrings to the new classes explaining:
- Purpose and usage of `MessageToolCall` and `MessageToolCallFunction`
- Dictionary-style access methods for compatibility
- Conversion methods for litellm integration
- Field descriptions and types

## Breaking Changes

None - this refactoring maintains full backward compatibility while providing a cleaner internal interface.

Fixes #547

@simonrosenberg can click here to [continue refining the PR](https://app.all-hands.dev/conversations/2f0bec8e55a34dffb0bb0bfd017ccab3)

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/all-hands-ai/agent-server:3bcb3cc-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-3bcb3cc-python \
  ghcr.io/all-hands-ai/agent-server:3bcb3cc-python
```

```
ghcr.io/all-hands-ai/agent-server:3bcb3cc-golang
ghcr.io/all-hands-ai/agent-server:3bcb3cc-java
ghcr.io/all-hands-ai/agent-server:3bcb3cc-python
```

_The `3bcb3cc` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._


xingyaoww (commented):
Any rationale for this? Why we are re-creating the same thing that litellm already has..

simonrosenberg:
> Any rationale for this? Why we are re-creating the same thing that litellm already has..

There was a discussion with @enyst where we didn't want to tie up `agent-sdk` with the implementation of litellm. So the idea is having a clear interface between the two so that we can use do everything we need with llms without important litellm.

However I realize this introduces a lot of change and we might want to be sure we all agree on this before going forward.

xingyaoww:
@enyst would love your opinion here!

enyst:
Oh, this does make sense to me, actually, I needed to do the same thing here:
- [Native Responses API for GPT-5](https://github.com/All-Hands-AI/agent-sdk/pull/500/files#diff-ffae89aafb0d9609b911ac6e34221c5f5a10d3ca296ca1a19d38abda3685a98bR7)

It’s a lot of change because of tests, but the change itself is small? Maybe we don’t need the inner Function class in this PR, just the outer. The “raw” call is just what it was from litellm.

I’m mainly thinking to get “completion” out of our processing, so we can process in a ~symmetrical way at first, for completion() and responses(), and then hopefully in a unified way. I also feel that the agent shouldn’t deal directly with litellm completion types... 🤔 There’s some small bits of code that know it’s completion type, so they access it dependent of that. Same for message.py IIRC.

It seems to me the same kind of thing that we did with wrapping into an `LLMResponse`, and I feel it means that we go towards to Option 3/3.1 here:
- [Support Responses API](https://github.com/All-Hands-AI/agent-sdk/issues/138#issuecomment-3293323214)

Do you think on this one it’s too much abstraction?

openhands/sdk/llm/message.py
```diff
@@ -13,6 +15,147 @@
 logger = get_logger(__name__)
 
 
+class MessageToolCallFunction(BaseModel):
+    """Represents a function call within a tool call.
+
+    This class encapsulates the function name and arguments for a tool call,
+    providing a clean interface for function call data while maintaining
+    compatibility with LLM API formats.
```
enyst: I’m not sure we need this class, maybe only something [like this](https://github.com/All-Hands-AI/agent-sdk/blob/e96a79328b00bdb730100eeb194f334fdad7d7b9/openhands/sdk/llm/llm_tool_call.py#L20) could work?

xingyaoww:
@enyst maybe you can port over your "ToolCall" class out when it is done? Or feel free to just modify this PR whereever you see fit!

openhands/sdk/llm/message.py
simonrosenberg: Agreed we can skip one layer
@openhands please use something similar to the code snippet below and remove MessageToolCallFunction
Also, solve the current conflicts with main and make sure that all tests and checks are passing.

```
from typing import Literal

from litellm import ChatCompletionMessageToolCall, ResponseFunctionToolCall
from pydantic import BaseModel, Field


class LLMToolCall(BaseModel):
    """Transport-agnostic tool call representation.

    One canonical id is used for linking across actions/observations and
    for Responses function_call_output call_id.
    """

    id: str = Field(..., description="Canonical tool call id")
    name: str = Field(..., description="Tool/function name")
    arguments_json: str = Field(..., description="JSON string of arguments")
    origin: Literal["completion", "responses"] = Field(
        ..., description="Originating API family"
    )
    raw: ChatCompletionMessageToolCall | ResponseFunctionToolCall | None = Field(
        default=None,
        description="Original provider object for advanced consumers",
        exclude=True,
    )
```

simonrosenberg: @openhands just keep one class: `MessageToolCall` not `LLMToolCall`

simonrosenberg:
@openhands please resolve conflicts

simonrosenberg:
Can you move the tool call validation from agent.py to message.py?
I am thinking of 

```
if any(tc.type != "function" for tc in message.tool_calls):
                logger.warning(
                    "LLM returned tool calls but some are not of type 'function' - "
                    "ignoring those"
                )

            tool_calls = [
                tool_call
                for tool_call in message.tool_calls
                if tool_call.type == "function"
            ]
            assert len(tool_calls) > 0, (
                "LLM returned tool calls but none are of type 'function'"
            )
```  can perhaps be inserted into `from_litellm_message`

and `assert tool_call.type == "function"`  ` assert tool_name is not None, "Tool call must have a name"` is not required anymore

**Comment by simonrosenberg** [comment]
@openhands

Can you move the tool call validation from agent.py to message.py?
I am thinking of

if any(tc.type != "function" for tc in message.tool_calls):
                logger.warning(
                    "LLM returned tool calls but some are not of type 'function' - "
                    "ignoring those"
                )

            tool_calls = [
                tool_call
                for tool_call in message.tool_calls
                if tool_call.type == "function"
            ]
            assert len(tool_calls) > 0, (
                "LLM returned tool calls but none are of type 'function'"
            )
```  can perhaps be inserted into `from_litellm_message`

and `assert tool_call.type == "function"`  ` assert tool_name is not None, "Tool call must have a name"` are not required anymore.

File: openhands/sdk/agent/agent.py [hunk 2]
```diff
@@ -199,21 +199,6 @@ def step(
...
-                )
-
-            tool_calls = [
-                tool_call
-                for tool_call in message.tool_calls
-                if tool_call.type == "function"
-            ]
-            assert len(tool_calls) > 0, (
-                "LLM returned tool calls but none are of type 'function'"
-            )
```
simonrosenberg, line 218: validation logic moved to `Message` class

simonrosenberg:
@enyst please review! 
I also moved the tool calls validation logic from `agent.py` to `message.py` if that makes sense to you...

simonrosenberg:
@openhands please resolve conflicts

openhands/sdk/agent/agent.py
```diff
@@ -224,9 +209,8 @@ def step(
             thought_content = [c for c in message.content if isinstance(c, TextContent)]
 
             action_events: list[ActionEvent] = []
-            for i, tool_call in enumerate(tool_calls):
+            for i, tool_call in enumerate(message.tool_calls):
                 action_event = self._get_action_event(
-                    state,
```
enyst, line 231: Out of curiosity, what's the reason for this? It was already unused?

openhands/sdk/llm/message.py
```diff
@@ -13,6 +15,150 @@
...
+
+        function = Function(name=self.name, arguments=self.arguments_json)
+
+        return ChatCompletionMessageToolCall(
+            id=self.id, type="function", function=function
+        )
+
+    @property
+    def function(self) -> Any:
+        """Backward compatibility property to access function details."""
```
enyst: Is this necessary? The agent has the habit of adding a bunch of "backward compatible"

openhands/sdk/llm/message.py
```diff
@@ -13,6 +15,150 @@
...
+            def __init__(self, name: str, arguments: str):
+                self.name = name
+                self.arguments = arguments
+
+        return FunctionInfo(self.name, self.arguments_json)
+
+    @property
+    def type(self) -> str:
+        """Backward compatibility property for tool call type."""
+        return "function"
```
enyst: probably same?

openhands/sdk/llm/message.py
```diff
@@ -133,6 +279,43 @@ def _coerce_content(cls, v: Any) -> Sequence[TextContent | ImageContent] | Any:
             return [TextContent(text=v)]
         return v
 
+    @field_validator("tool_calls", mode="before")
+    @classmethod
+    def _coerce_tool_calls(cls, v: Any) -> list[MessageToolCall] | None:
+        """Convert tool_calls from various formats to MessageToolCall objects."""
```
enyst: I wonder if we need this, when we have `from_litellm_tool_call`? 🤔

openhands/sdk/llm/message.py
```diff
@@ -254,12 +437,37 @@ def from_litellm_message(cls, message: LiteLLMMessage) -> "Message":
...
+                    "LLM returned tool calls but none are of type 'function'"
+                )
+
         return Message(
             role=message.role,
             content=[TextContent(text=message.content)]
             if isinstance(message.content, str)
             else [],
-            tool_calls=message.tool_calls,
+            tool_calls=converted_tool_calls,
```
enyst, line 335: Love this, and why this PR looks right to me, if we can do this: our `Message` contains roles, content, tool calls, without being tied to a particular litellm type for a particular API tool call format.

simonrosenberg:
@openhands please resolve conflicts

openhands/sdk/agent/agent.py
simonrosenberg: It was already unused. Some recent PR must have changed that.

simonrosenberg:
@openhands remove the function and type property from MessageToolCall. Make sure to update any part of the code that accesses those fields. 
Also, if _coerce_tool_calls is not used in the code, remove it entirely

openhands/sdk/llm/message.py
```diff
@@ -13,6 +15,133 @@
...
+            id: Unique identifier for the tool call. If None, a UUID will be generated.
+            name: The name of the function to call
+            arguments_json: The arguments as a JSON string or dict. If dict, will be
+                converted to JSON string.
+            origin: The originating API family
+            raw: Original provider object for advanced consumers
+            **kwargs: Additional keyword arguments
+        """
+        if id is None:
+            id = f"call_{uuid.uuid4().hex[:8]}"
```
enyst: I don't think it's possible from litellm to have id=None, is it?

openhands/sdk/llm/message.py
```diff
@@ -13,6 +15,133 @@
...
+
+        if name is None:
+            name = ""
```
enyst: This shouldn't be possible either, it seems we had an assert in agent.py,
```
assert tool_name is not None, "Tool call must have a name"
```

openhands/sdk/llm/message.py
enyst: It seems to me this is the agent exaggerating with overly defensive code. This is called from `from_litellm_tool_call` which does check for None and maybe that's enough?

simonrosenberg: This is indeed impossible in `from_litellm_tool_call` BUT we could imagine that for testing it would be nice to provide default values for the init?

But for typing it's probably better to "calibrate" all typing/behavior to the production code path?

enyst: We could mock in tests with some "1234"? Idk, it might be safer to have some sort of slightly more "realistic" objects 

On a side note, I wonder if we can make this a BaseModel and let pydantic enforce stuff

enyst: Oh, it is, could we just skip init?

openhands/sdk/llm/message.py
```diff
@@ -13,6 +15,133 @@
...
+    def to_litellm_tool_call(self) -> ChatCompletionMessageToolCall:
+        """Convert this MessageToolCall to a litellm ChatCompletionMessageToolCall.
+
+        This method is primarily for testing and compatibility with existing code
+        that expects litellm tool calls.
+
+        Returns:
+            A ChatCompletionMessageToolCall instance with the same data
+        """
+        from litellm.types.utils import Function
```
enyst: This could be top-level import?

openhands/sdk/llm/message.py
```diff
@@ -13,6 +13,123 @@
...
+        """
+
+        super().__init__(
+            id=id,
+            name=name,
+            arguments_json=arguments_json,
+            origin=origin,
+            raw=raw,
+            **kwargs,
+        )
```
simonrosenberg: do we need an init now that no validation is done in the init? @openhands

openhands/sdk/llm/message.py
simonrosenberg: Yes init is useless now!

openhands/sdk/llm/message.py
simonrosenberg: I think this method should be deleted. It's only used for test and it's not that helpful, idk

simonrosenberg: @openhands delete `to_litellm_tool_call`
Keep the tests but use ChatCompletionMessageToolCall and Function like on the branch main.

simonrosenberg:
@openhands 
1. delete to_litellm_tool_call in message.py. Keep the tests but use ChatCompletionMessageToolCall and Function like it is done on the main branch.
2. fix the pre-commit checks and tests

openhands/sdk/llm/message.py
```diff
@@ -13,6 +13,76 @@
...
+        return getattr(self, key, default)
+
+    def __getitem__(self, key: str) -> Any:
+        """Support dictionary-style access to attributes."""
+        return getattr(self, key)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        """Support dictionary-style assignment of attributes."""
+        setattr(self, key, value)
+
```
enyst:
```suggestion
    def __contains__(self, key: str) -> bool:
        """Support 'in' operator for dictionary-style access."""
        return hasattr(self, key)

    def get(self, key: str, default: Any = None) -> Any:
        """Get attribute with default value, similar to dict.get()."""
        return getattr(self, key, default)

    def __getitem__(self, key: str) -> Any:
        """Support dictionary-style access to attributes."""
        return getattr(self, key)

    def __setitem__(self, key: str, value: Any) -> None:
        """Support dictionary-style assignment of attributes."""
        setattr(self, key, value)

```

Seems like maybe we don't need these, it's a simple wrapper over one thing?

simonrosenberg: @open

> Seems like maybe we don't need these, it's a simple wrapper over one thing?

it's crazy that the llm would add such useless stuff. Good catch

simonrosenberg: @enyst deleted

openhands/sdk/event/llm_convertible/action.py
```diff
@@ -29,7 +34,7 @@ class ActionEvent(LLMConvertibleEvent):
     tool_call_id: ToolCallID = Field(
         ..., description="The unique id returned by LLM API for this tool call"
     )
-    tool_call: ChatCompletionMessageToolCall = Field(
+    tool_call: MessageToolCall = Field(
         ...,
         description=(
             "The tool call received from the LLM response. We keep a copy of it "
```
enyst, line 40: If I see this right, before, we serialized ChatCompletionMessageToolCall, and now we serialize MessageToolCall. I wonder if this works as expected?

It might be worth to run an example or two, with tool calls - or the persistence example maybe?

enyst (commented):
Thank you for this!

Now it's about 100 lines main diff, and some deleted from the agent. 😅 

It's a bit interesting it didn't require changes in llm.py, but I guess it's normal, we just return the response, we didn't use this type explicitly.

I wonder if it works as expected, idk, maybe with some real test, rather than unit tests? Like the serialization, and restoring the correct message in LLM context 🤞

simonrosenberg:
> Thank you for this!
> 
> Now it's about 100 lines main diff, and some deleted from the agent. 😅
> 
> It's a bit interesting it didn't require changes in llm.py, but I guess it's normal, we just return the response, we didn't use this type explicitly.
> 
> I wonder if it works as expected, idk, maybe with some real test, rather than unit tests? Like the serialization, and restoring the correct message in LLM context 🤞

I will run the examples :) 
There is another litellm class that is spilling in the code:  `ChatCompletionToolParam` used by 
1. SystemPromptEvent ...
2. Tool.to_openai_tool ...
Making a second issue to deal with those in another PR 
Basically I think it's OK to use ChatCompletionToolParam in the llm logic itself, but once it's higher level than that it should go through `MessageToolCall`

simonrosenberg:
@enyst the integration tests (with real llm) work and I ran 2 exemples without any trouble :)

enyst (approved):
Very interesting, thank you! Seems that even Deepseek works without change in llm.py or the non-native function calling converter.

(@simonrosenberg Just a thought, not related to the PR, the old integration tests were uploading the full output.json as an artifact to github actions, and link it for download in the final comment. We could maybe use sometime something like that, to see quick what the agent actually did. (the log of all runs))

I'd love to take this in. It cleared up a little in the Agent, it added conversions in message.py but they're just from litellm and back, I think it looks great, and we can use it going forward for Responses. Otherwise we kinda need to reinvent it...

@xingyaoww I'd love to know if you have concerns, or we can sneak it in. 😅

simonrosenberg:
@enyst  I also removed the raw field https://github.com/All-Hands-AI/agent-sdk/pull/550/commits/f8a9cc6a5e3efc81e7187bb7f8a8910b54ce6c15
It's true that coding agents tend to exagerate with backward compatibility add more stuff than strictly necessary. 
Not sure how we could stir them towards simpler code

enyst:
> @enyst I also removed the raw field [f8a9cc6](https://github.com/All-Hands-AI/agent-sdk/commit/f8a9cc6a5e3efc81e7187bb7f8a8910b54ce6c15) It's true that coding agents tend to exagerate with backward compatibility add more stuff than strictly necessary. Not sure how we could stir them towards simpler code

A few thoughts
- ack, as of now I don't need the raw field
- something smells a bit about `_add_tool_keys`, it's a utility method bolted on top of other processing in Message, which is actually needed for both paths (all three paths, also string serializer!) and it could maybe go in MessageToolCall as a class method...no, maybe just as serializing to the right dict. Idk, that's too tiny, but I'm kinda working around it
- GPT-5 seems to have been trained way too much on fallbacks/backwards compat; sometimes I succeed to work with it without, with initial clear prompting, but it's hit-or-miss; right now I'm having fun with a prompt containing `"No dict soup! No overly defensive fallbacks"`. 😂
I'm thrilled with the "dict soup" thing, hey seems to work 😅

Hah, here it goes (after long context, it still remembers it!):

> The unified constructors and serializers remove dict soup and clarify which API family produced the tool call via origin.

PR #558, opened 2025-09-29
add claude-sonnet-4.5

xingyaoww:


xingyaoww:
@enyst i think we need to spend extra efforts in the LLM here to be able to support "anthropic thinking", right?

enyst:
Yes, because Anthropic expects its thinking back, I'll add a note in the issue

PR #576, opened 2025-09-30
Add Anthropic thinking blocks support to events system

### LINKED ISSUE #172: Add reasoning with Anthropic models
Anthropic has a feature they call “extended thinking” where their API sends `thinking_blocks` in a response, and _expects them back_ in a specific form. According to [litellm docs](https://docs.litellm.ai/docs/reasoning_content) litellm DOESN’T normalize `thinking_blocks` into `reasoning_content`, instead it sends both. (OH agents get very confused here)

Instead, like prompt caching, Anthropic API has its own format/rules that needs accounted for in client code.

Image

Now that reasoning is in, we should also look into implementing this for Anthropic.

xingyaoww:
## Summary

This PR implements Anthropic's thinking blocks feature in the OpenHands SDK events system, enabling support for extended thinking capabilities as described in the [LiteLLM documentation](https://docs.litellm.ai/docs/reasoning_content#pass-thinking-to-anthropic-models).

### Core Implementation
- **Added `ThinkingBlock` model**: New Pydantic model for Anthropic thinking blocks with `type`, `thinking`, and `signature` fields
- **Enhanced `Message` class**: Added `thinking_blocks` field with default empty list
- **Updated `from_litellm_message()`**: Extract thinking blocks from LiteLLM responses using dict access for TypedDict compatibility
- **Added `thinking_blocks` property**: New property in `MessageEvent` class to access thinking blocks

### Visualization & Display
- **Enhanced `MessageEvent.visualize`**: Display thinking blocks with truncation for better readability
- **Updated `MessageEvent.__str__()`**: Include thinking block count in string representation
- **Added thinking block serialization**: Thinking blocks are included in message serialization for API compatibility

### Testing & Examples
- **Comprehensive test suite**: 12 unit tests covering all functionality in `tests/sdk/llm/test_thinking_blocks.py`
- **Updated example script**: Enhanced `examples/25_anthropic_thinking.py` with proper callback handling
- **Export support**: Added `ThinkingBlock` to `openhands.sdk.llm` module exports

### Anthropic Thinking Blocks Format
According to the LiteLLM docs, Anthropic thinking blocks are not normalized and have a specific format:
```python
{
    "type": "thinking",
    "thinking": "The model's internal reasoning...",
    "signature": "unique_identifier"
}
```

### Implementation Approach
- **Non-intrusive**: Thinking blocks are optional and default to empty list
- **Backward compatible**: Existing code continues to work without changes
- **Type safe**: Full type annotations and Pydantic validation
- **Efficient**: Uses dict access to avoid TypedDict compatibility issues

### Usage Example
```python
# Thinking blocks are automatically extracted from LiteLLM responses
message = Message.from_litellm_message(litellm_response)
if message.thinking_blocks:
    for block in message.thinking_blocks:
        print(f"Thinking: {block.thinking}")
        print(f"Signature: {block.signature}")
```

## Testing

- ✅ All 12 unit tests pass
- ✅ Pre-commit hooks pass (formatting, linting, type checking)
- ✅ Mock data testing confirms correct functionality
- ✅ Backward compatibility verified

## Notes

- The LiteLLM proxy may not yet support the `reasoning_effort` parameter for Anthropic models
- Thinking blocks are preserved in their raw format as per Anthropic's requirements
- This feature is specific to Anthropic models and complements the existing `reasoning_content` field

Fixes #172

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/9d3a992c42974edabfeb9cb4e5fa4e42)

### DIFF
--- a/openhands/sdk/llm/message.py
+++ b/openhands/sdk/llm/message.py
@@ -13,6 +13,21 @@ from openhands.sdk.utils import DEFAULT_TEXT_CONTENT_LIMIT, maybe_truncate
 logger = get_logger(__name__)
 
 
+class ThinkingBlock(BaseModel):
+    """Anthropic thinking block for extended thinking feature.
+
+    This represents the raw thinking blocks returned by Anthropic models
+    when extended thinking is enabled. These blocks must be preserved
+    and passed back to the API for tool use scenarios.
+    """
+
+    type: Literal["thinking"] = "thinking"
+    thinking: str = Field(..., description="The thinking content")
+    signature: str | None = Field(
+        default=None, description="Cryptographic signature for the thinking block"
+    )
+
+
 class BaseContent(BaseModel):
     cache_prompt: bool = False
 
@@ -86,6 +101,11 @@ class Message(BaseModel):
         default=None,
         description="Intermediate reasoning/thinking content from reasoning models",
     )
+    # Anthropic-specific thinking blocks (not normalized by LiteLLM)
+    thinking_blocks: list[ThinkingBlock] = Field(
+        default_factory=list,
+        description="Raw Anthropic thinking blocks for extended thinking feature",
+    )
 
     @property
     def contains_image(self) -> bool:
@@ -134,6 +154,11 @@ class Message(BaseModel):
     def _list_serializer(self) -> dict[str, Any]:
         content: list[dict[str, Any]] = []
         role_tool_with_prompt_caching = False
+
+        # Add thinking blocks first (for Anthropic extended thinking)
+        for thinking_block in self.thinking_blocks:
+            thinking_dict = thinking_block.model_dump()
+            content.append(thinking_dict)
 
         for item in self.content:
             # All content types now return list[dict[str, Any]]
@@ -195,18 +220,40 @@ class Message(BaseModel):
 
         Provider-agnostic mapping for reasoning:
         - Prefer `message.reasoning_content` if present (LiteLLM normalized field)
+        - Extract `thinking_blocks` from content array (Anthropic-specific)
         """
         assert message.role != "function", "Function role is not supported"
 
         rc = getattr(message, "reasoning_content", None)
+        content = []
+        thinking_blocks = []
+
+        # Handle content
+        if isinstance(message.content, str):
+            content = [TextContent(text=message.content)]
+
+        # Extract thinking blocks from LiteLLM's thinking_blocks field
+        litellm_thinking_blocks = getattr(message, "thinking_blocks", None)
+        if litellm_thinking_blocks:
+            for block in litellm_thinking_blocks:
+                # Convert LiteLLM thinking block to our ThinkingBlock
+                # LiteLLM thinking blocks are TypedDict objects
+                thinking_content = (
+                    block.get("thinking", "") if isinstance(block, dict) else ""
+                )
+                signature = block.get("signature") if isinstance(block, dict) else None
+                thinking_block = ThinkingBlock(
+                    thinking=thinking_content,
+                    signature=signature,
+                )
+                thinking_blocks.append(thinking_block)
 
         return Message(
             role=message.role,
-            content=[TextContent(text=message.content)]
-            if isinstance(message.content, str)
-            else [],
+            content=content,
             tool_calls=message.tool_calls,
             reasoning_content=rc,
+            thinking_blocks=thinking_blocks,
         )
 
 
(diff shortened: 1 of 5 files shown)


openhands/sdk/llm/message.py
```diff
@@ -131,10 +168,32 @@ def _string_serializer(self) -> dict[str, Any]:
...
+            # Add redacted thinking block if needed for Anthropic extended thinking
+            if ensure_thinking_blocks and not thinking_blocks:
+                thinking_blocks = [
+                    RedactedThinkingBlock(
+                        thinking=(
+                            "This message was generated before extended "
+                            "thinking was enabled."
+                        )
+                    )
+                ]
```
xingyaoww: I think this is wrong. We supposed to only use thinking block returned from the API

openhands/sdk/llm/message.py
```diff
@@ -195,18 +254,58 @@ def from_litellm_message(cls, message: LiteLLMMessage) -> "Message":
...
+                # LiteLLM thinking blocks are TypedDict objects
+                thinking_content = (
+                    block.get("thinking", "") if isinstance(block, dict) else ""
+                )
+                signature = block.get("signature") if isinstance(block, dict) else None
+                thinking_block = ThinkingBlock(
+                    thinking=thinking_content,
+                    signature=signature,
+                )
+                thinking_blocks.append(thinking_block)
```
xingyaoww: This also feels wrong - litellm supposed to clean this up for us, we should just set content to content and thinking_blocks to thinking blocks

openhands/sdk/llm/utils/model_features.py
```diff
@@ -117,7 +117,7 @@ class ModelFeatures:
     # OpenAI GPT-5 family (includes mini variants)
     "gpt-5*",
     # Anthropic Claude Sonnet 4.5
-    "claude-sonnet-4-5-20250929",
+    "claude-sonnet-4-5*",
```
ryanhoangt, line 126: It seems `claude-sonnet-4-*` models and `claude-sonnet-3-7` also support extended thinking 🤔

Image

ryanhoangt: I separated these models from "reasoning effort"-supported models, it seems to make sense semantically as for these models we need to set a thinking token budget as well.

openhands/sdk/llm/message.py
ryanhoangt: I fixed this in the latest commit.

openhands/sdk/llm/message.py
ryanhoangt: Yeah these extra `ensure_thinking_blocks` handling code are not needed actually, we just pass back the thinking blocks we get from the response.

ryanhoangt:
This seems to work (to some extent) now -- CC @xingyaoww @enyst 

Screenshot 2025-10-01 at 16 01 41
Screenshot 2025-10-01 at 16 02 32

openhands/sdk/llm/utils/model_features.py
ryanhoangt: Hmm this seems to be a bug from LiteLLM, Claude models doesn't support `reasoning_effort` actually: https://docs.claude.com/en/api/openai-sdk#detailed-openai-compatible-api-support

openhands/sdk/llm/llm.py
```diff
@@ -179,6 +179,11 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         "This is a string that can be one of 'low', 'medium', 'high', or 'none'. "
         "Can apply to all reasoning models.",
     )
+    extended_thinking_budget: int | None = Field(
+        default=None,
```
xingyaoww: Let's default to 200k that anthropic used to get the best swebench performance?
```suggestion
        default=200000,
```

openhands/sdk/llm/utils/model_features.py
```diff
@@ -116,8 +117,12 @@ class ModelFeatures:
...
     # OpenAI GPT-5 family (includes mini variants)
     "gpt-5*",
-    # Anthropic Claude Sonnet 4.5
-    "claude-sonnet-4-5-20250929",
+]
+
+EXTENDED_THINKING_PATTERNS: list[str] = [
+    # Anthropic families
+    "claude-3-7-sonnet*",
+    "claude-sonnet-4*",
```
xingyaoww: Let's turn off extended thinking for sonnet 3.7 and 4 though, since they don't really work well with agents

```suggestion
    # We did not include sonnet 3.7 and 4 here as they don't brings
    # significant performance improvements for agents
    "claude-sonnet-4-5*",
```

xingyaoww (commented):
OK i think this LGTM now! Will merge

PR #623, opened 2025-10-02
feat: NonExecutableActionEvent for Completions path (preserve tool_call_id continuity)

enyst:
We don't seem to handle this yet, or maybe I'm missing something? What happens here is that `ActionEvent` didn't get created yet, because of an error, so only an `AgentErrorEvent` exists.

I think it could happen for:
- pydantic schema validation
- tool doesn't exist

Test case: I told the agent to add a line on top of the readme, but make sure it uses "insert_line" => it tries to call it with insert_line=0.
=> validation error => infinite loop

On a side note, maybe we shouldn't actually have a validation error for 0 there, IIRC it's the line before the line to be inserted, so 0. 🤔

Other options:
- we could create the ActionEvent, but give it something that says it's invalid ("executable=false", ha!), so we don't try to execute it?
- maybe we could give the AgentErrorEvent enough info so that we send to the LLM
    - a story about what happened (so no tool call/result)
    - reconstruct tool call + tool result (this is terrible, 2 messages for 1 event)

---

#### Agent Summary

- Extracted and implemented NonExecutableActionEvent on top of main (Completions path only)
- Ensures assistant function_call(s) are persisted when a tool is missing or arg validation fails, preserving tool_call_id for proper pairing on the next turn
- Integrated with Agent, View.filter_unmatched_tool_calls, and LLMConvertibleEvent message conversion
- Added focused unit tests; full test suite passes locally

#### Details

- Agent changes (openhands/sdk/agent/agent.py):
  - When tool is missing or args fail validation, emit NonExecutableActionEvent(tool_calls=[tool_call]) followed by AgentErrorEvent with matching tool_call_id
  - Guarantees continuity of tool_call_id, enabling proper next-turn handling
- Event model (openhands/sdk/event/llm_convertible/action.py):
  - NonExecutableActionEvent holds thought/reasoning/thinking_blocks and raw tool_calls
  - to_llm_message returns an assistant message with tool_calls for round-trip compatibility
- View filtering (openhands/sdk/context/view.py):
  - filter_unmatched_tool_calls recognizes NonExecutableActionEvent.tool_calls and retains the event if any tool_call.id matches an observation/error
- Exports:
  - openhands/sdk/event/__init__.py and openhands/sdk/event/llm_convertible/__init__.py export NonExecutableActionEvent
- Tests:
  - tests/sdk/event/test_non_executable_action_event.py: round-trip to_llm_message
  - tests/sdk/context/test_view_non_exec_filtering.py: keep NEA when matched by AgentErrorEvent
  - tests/sdk/agent/test_non_executable_action_emission.py: agent emits NEA then AgentErrorEvent; verifies tool_call_id continuity and ordering

#### Why

- Main branch uses the Completions path (no Responses API). This preserves function_call continuity when an action cannot be executed, aligning behavior and enabling correct pairing on the next turn.

#### Notes

- No changes to dependencies
- Backwards compatible with existing flows

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/9d05666b6af34a0ba7ffa0a13213b123)

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/all-hands-ai/agent-server:22cfb56-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-22cfb56-python \
  ghcr.io/all-hands-ai/agent-server:22cfb56-python
```

```
ghcr.io/all-hands-ai/agent-server:22cfb56-golang
ghcr.io/all-hands-ai/agent-server:22cfb56-java
ghcr.io/all-hands-ai/agent-server:22cfb56-python
```

_The `22cfb56` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/openhands/sdk/event/llm_convertible/action.py
+++ b/openhands/sdk/event/llm_convertible/action.py
@@ -1,4 +1,3 @@
-import json
 from collections.abc import Sequence
 
 from pydantic import Field
@@ -15,7 +14,6 @@ from openhands.sdk.llm import (
 )
 from openhands.sdk.security import risk
 from openhands.sdk.tool.schema import Action
-from openhands.sdk.utils.visualize import display_dict
 
 
 class ActionEvent(LLMConvertibleEvent):
@@ -88,19 +86,9 @@ class ActionEvent(LLMConvertibleEvent):
         if self.action:
             content.append(self.action.visualize)
         else:
-            # When action is None (non-executable), show the function call and arguments
-            content.append("Action: ", style="bold")
-            content.append("(not executed)")
-            content.append("\n\n")
+            # When action is None (non-executable), show the function call
             content.append("Function call:\n", style="bold")
-            content.append(f"- {self.tool_call.name} ({self.tool_call.id})\n\n")
-            # Pretty print arguments if JSON is valid
-            try:
-                args = json.loads(self.tool_call.arguments)
-            except Exception:
-                args = {"raw": self.tool_call.arguments}
-            content.append("Arguments:", style="bold")
-            content.append(display_dict(args))
+            content.append(f"- {self.tool_call.name} ({self.tool_call.id})\n")
 
         return content
 


openhands/sdk/context/view.py
```diff
@@ -112,6 +116,9 @@ def _get_action_tool_call_ids(events: list[LLMConvertibleEvent]) -> set[ToolCall
         for event in events:
             if isinstance(event, ActionEvent) and event.tool_call_id is not None:
                 tool_call_ids.add(event.tool_call_id)
+            # For NonExecutableActionEvent, collect its single tool_call id
```
enyst:
```suggestion
```

openhands/sdk/context/view.py
```diff
@@ -112,6 +116,8 @@ def _get_action_tool_call_ids(events: list[LLMConvertibleEvent]) -> set[ToolCall
         for event in events:
             if isinstance(event, ActionEvent) and event.tool_call_id is not None:
                 tool_call_ids.add(event.tool_call_id)
+            elif isinstance(event, NonExecutableActionEvent):
```
enyst:
```suggestion
            elif isinstance(event, NonExecutableActionEvent) and event.tool_call_id is not None:
```

openhands/sdk/event/llm_convertible/action.py
```diff
@@ -105,3 +105,64 @@ def __str__(self) -> str:
...
+    source: SourceType = "agent"
+    thought: Sequence[TextContent] = Field(
+        default_factory=list,
+        description="The assistant's thought content returned alongside tool calls",
+    )
+    reasoning_content: str | None = Field(
+        default=None,
+        description="Intermediate reasoning content from reasoning models",
+    )
+    thinking_blocks: list[ThinkingBlock | RedactedThinkingBlock] = Field(
```
enyst: I find it difficult to believe that these will be there, but it seems possible

openhands/sdk/event/llm_convertible/action.py
```diff
@@ -105,3 +105,64 @@ def __str__(self) -> str:
         )
         action_name = self.action.__class__.__name__
         return f"{base_str}\n  Thought: {thought_preview}\n  Action: {action_name}"
+
+
+class NonExecutableActionEvent(LLMConvertibleEvent):
```
xingyaoww: Shall we just inherit this from `ActionEvent`? irrc condenser need to use baseclass to do type checking & remove the unpaired ones

This should also prevent a lot of code duplications here

xingyaoww (commented):
Like the general idea and mostly LGTM - just one comment

openhands/sdk/event/llm_convertible/action.py
enyst: Sure, let's try, we just shouldn't execute this `ActionEvent`, so instead of code to add it (to the condenser's checks), we may need some code to _exclude it_... but maybe that's already handled, since we execute unmatched, and this one comes to life already matched. 🤔

@openhands Do what Xingyao says. Make sure we fill the `action_id` field so that downstream we can pair this action to the `AgentErrorEvent` observation. Reduce redundancies, clean up the condenser, remove the fields that are inherited from ActionEvent. Think deeply if we ever need code to exclude it, our preference is that we do NOT add any more code.

openhands/sdk/event/llm_convertible/action.py
```diff
@@ -105,3 +105,90 @@ def __str__(self) -> str:
...
+    so we persist the function_call for the next turn to match tool outputs.
+
+    This subclass of ActionEvent preserves the original tool_call and IDs so that
+    downstream components can match observations/errors via tool_call_id and
+    action_id. It injects a placeholder Action to satisfy base typing but it is
+    never executed.
+    """
+
+    # Provide a default non-executable action to satisfy base class typing.
+    action: Action = Field(default_factory=_NonExecutablePlaceholderAction)
```
enyst: @xingyaoww This is the kind of thing that I was fearing: in order to make sure that a subclass of `ActionEvent` is not executable, we may need tricks...

The alternative here could be to make it `action: Action | None`... 
In that case, do we lose some typing guarantees? It's nice to see code that just goes `observation = tool(action_event.action)` without having to care about "well, we checked before, but we kinda don't know if we have an `.action`"

WDYT? No perfect solution, just a matter of tradeoffs that maybe feel more right...

openhands/sdk/conversation/state.py
```diff
@@ -279,7 +284,9 @@ def get_unmatched_actions(events: Sequence[Event]) -> list[ActionEvent]:
         unmatched_actions = []
         # Search in reverse - recent events are more likely to be unmatched
         for event in reversed(events):
-            if isinstance(event, (ObservationEvent, UserRejectObservation)):
+            if isinstance(
+                event, (ObservationEvent, UserRejectObservation, AgentErrorEvent)
```
enyst: We need an `isinstance` here, because AgentErrorEvent inherits from ObservationBase, not from ObservationEvent. 🤔

Similar checks with condenser

openhands/sdk/event/llm_convertible/action.py
```diff
@@ -105,3 +105,94 @@ def __str__(self) -> str:
...
+        # Keep inherited fields aligned with tool_call at runtime
+        try:
+            if getattr(self, "tool_name", None) != self.tool_call.name:
+                object.__setattr__(self, "tool_name", self.tool_call.name)
+            if getattr(self, "tool_call_id", None) != self.tool_call.id:
+                object.__setattr__(self, "tool_call_id", self.tool_call.id)
+        except Exception:
+            pass
+        return self
+
```
enyst:
```suggestion
```

openhands/sdk/event/llm_convertible/action.py
```diff
@@ -105,3 +105,82 @@ def __str__(self) -> str:
...
+                    if data.get("tool_name") is None and name is not None:
+                        data["tool_name"] = name
+                    if data.get("tool_call_id") is None and tcid is not None:
+                        data["tool_call_id"] = tcid
+                    if data.get("llm_response_id") is None and tcid is not None:
+                        data["llm_response_id"] = tcid
+                except Exception:
+                    pass
+        return data
+
```
enyst:
```suggestion
```

openhands/sdk/conversation/state.py
```diff
@@ -279,7 +284,9 @@ def get_unmatched_actions(events: Sequence[Event]) -> list[ActionEvent]:
...
+            ):
                 observed_action_ids.add(event.action_id)
```
enyst, line 283: Ah, and we need `action_id` in `AgentErrorEvent`. I'm looking into: what if, then, we make it inherit from `ObservationEvent` (which has it), rather than from `ObservationBaseEvent` (which doesn't)

enyst: OpenHands-GPT-5 is not very happy:

What if AgentErrorEvent inherited from ObservationEvent?
**Pros**

- Uniform interface with ObservationEvent:
  - action_id becomes required (consistent with ObservationEvent and UserRejectObservation)
  - You could potentially simplify “unmatched” logic (drop the explicit AgentErrorEvent special-case since it would be an ObservationEvent)
  - Conceptually groups all “responses to an action” under a single subclass, which can be seen as tidy.

**Cons**

- Requires observation: Observation on AgentErrorEvent:
  - You’d need to introduce an ErrorObservation (or similar) for the error payload. This adds code and complexity, contrary to the stated preference to avoid adding more code.
  - All sites constructing AgentErrorEvent (including tests) would have to construct a corresponding Observation; today they just pass error text, tool_name, tool_call_id.
  - StuckDetector and visualizer logic expect AgentErrorEvent.error to be directly available; if error content moves into an Observation, you’d either duplicate the field or refactor multiple paths.
- Semantics:
  - ObservationEvent defaults source="environment"; AgentErrorEvent is source="agent". You’d override, but this is another edge case to manage.
- Limited practical benefit:
  - View.filter_unmatched_tool_calls already keys off ObservationBaseEvent; there’s little to gain in view/condenser by switching inheritance.
  - events_to_messages and visualizer already treat AgentErrorEvent distinctly with the panel title “Agent Error” and custom rendering.

**Recommendation**

Do not change AgentErrorEvent’s inheritance.

openhands/sdk/event/llm_convertible/action.py
enyst: OK, actually, I played a bit more with the code and I think `action: Action | None` is reasonable and cleaner!

xingyaoww (approved):
LGTM!

xingyaoww (changes requested):
Actually, what do you feel about completely get rid of non executable action event, and just use ActionEvent with action=None?

enyst:
Sounds great!  👀

xingyaoww:
@OpenHands review diff from main and remove NonExecutableActionEvent, we should just use ActionEvent with .action being Action | None.

enyst:
Oh, I had OH work on it too

xingyaoww:
@enyst happy to go with your OH result too!

PR #624, opened 2025-10-03
Feat: make read-only planning agent

simonrosenberg:
---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/all-hands-ai/agent-server:f8dfba9-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-f8dfba9-python \
  ghcr.io/all-hands-ai/agent-server:f8dfba9-python
```

```
ghcr.io/all-hands-ai/agent-server:f8dfba9-golang
ghcr.io/all-hands-ai/agent-server:f8dfba9-java
ghcr.io/all-hands-ai/agent-server:f8dfba9-python
```

_The `f8dfba9` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._


simonrosenberg:
@openhands I have an issue with openhands/tools/str_replace_editor/definition.py
str_replace_editor_tool is defined regardless of read_only = true or false. Is that an issue? 
How can I solve this error. Make sure to run all checks and tests to validate your changes

openhands/tools/file_viewer/impl.py
```diff
@@ -0,0 +1,38 @@
+from openhands.sdk.tool import ToolExecutor
+from openhands.tools.file_viewer.definition import FileViewerObservation
+from openhands.tools.str_replace_editor.definition import StrReplaceEditorAction
+from openhands.tools.str_replace_editor.impl import FileEditorExecutor
+
+
+class FileViewerExecutor(ToolExecutor):
+    """File viewer executor that uses FileEditorExecutor in read-only mode."""
```
xingyaoww: I think we might want to port over some glob/grep tools that the agent will use pretty frequently 🤔 since it won't have access to bashnow

https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/agenthub/readonly_agent/tools/glob.py
https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/agenthub/readonly_agent/tools/grep.py

Irrc claude-code was using ripgrep: https://github.com/BurntSushi/ripgrep (which i should implemented in the grep tool above)

examples/27_file_viewer_refactoring.py
xingyaoww, line 1: Let's remove this one

openhands/sdk/agent/prompts/planning_system_prompt.j2
```diff
@@ -0,0 +1,103 @@
...
+## Available Tools
+
+You have access to the following tools:
+
+- **str_replace_editor**: Read-only access to view files and directory structures
+- **plan_writer**: Write your analysis and plans to markdown files
+- **task_tracker**: Organize and track your planning activities
+{% if llm_security_analyzer %}
+- **Security Analysis**: Evaluate potential security implications of planned changes
+{% endif %}
```
xingyaoww: Let's always disable security analyzer since it is read-only?

openhands/sdk/agent/prompts/system_prompt_planning.j2
xingyaoww, line 1: Are these sonnet-generated? Can we make it more succint and shorter

examples/26_planning_agent_workflow.py
```diff
@@ -0,0 +1,170 @@
...
+    # Create a temporary workspace
+    workspace_dir = Path(tempfile.mkdtemp())
+    print(f"Working in: {workspace_dir}")
+
+    # Initialize LLM (you'll need to set your API key)
+    llm = LLM(
+        model="gpt-4o-mini",
+        api_key=SecretStr("your-api-key-here"),  # Replace with your actual API key
+        service_id="openai",
+    )
```
xingyaoww: please run it end to end to make sure it works and set the model name here similar to other examples

openhands/sdk/agent/prompts/planning_system_prompt.j2
simonrosenberg: Sounds good. You still don't think we should have a specific PlanningAgent class btw?

openhands/tools/file_viewer/impl.py
simonrosenberg: Should we also include a git checkout tool?

openhands/sdk/agent/prompts/planning_system_prompt.j2
xingyaoww: yeah i think at most we just add a different present to construct it, don't really see the need of a different class now

openhands/tools/file_viewer/definition.py
```diff
@@ -0,0 +1,112 @@
...
+from openhands.sdk.llm import ImageContent, TextContent
+from openhands.sdk.tool import Action, Observation, ToolAnnotations, ToolDefinition
+
+
+class FileViewerAction(Action):
+    """Schema for file viewer operations - read-only."""
+
+    command: Literal["view"] = Field(
+        description="The commands to run. Only `view` is allowed for read-only access."
+    )
```
xingyaoww: If only view is allowed, let's remove it

openhands/tools/file_viewer/definition.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+TOOL_DESCRIPTION = """
+Custom file viewer tool for viewing files and directories in plain-text format.
+This is a read-only tool that only supports viewing operations.
+
+* State is persistent across command calls and discussions with the user
+* If `path` is a text file, `view` displays the result of applying `cat -n`. If `path` is a directory, `view` lists non-hidden files and directories up to 2 levels deep
+* The following binary file extensions can be viewed in Markdown format: [".xlsx", ".pptx", ".wav", ".mp3", ".m4a", ".flac", ".pdf", ".docx"]. IT DOES NOT HANDLE IMAGES.
+* If a `command` generates a long output, it will be truncated and marked with `<response clipped>`
```
xingyaoww:
```suggestion
```
this donesn't seem to be valid -- i don't think we support mp3 now

openhands/tools/file_viewer/impl.py
```diff
@@ -0,0 +1,38 @@
...
+        # Create a read-only FileEditorExecutor
+        self.editor_executor = FileEditorExecutor(
+            workspace_root=workspace_root, read_only=True
+        )
+
+    def __call__(self, action):
+        """Execute a file viewer action by converting it to a file editor action."""
+        # Convert FileViewerAction to FileEditorAction
+        editor_action = FileEditorAction(
+            command=action.command,
```
xingyaoww: we should just hardcode view here

openhands/tools/file_editor/impl.py
```diff
@@ -13,10 +13,20 @@
...
-    def __init__(self, workspace_root: str | None = None):
+    """File editor executor with configurable read-only mode."""
+
+    def __init__(self, workspace_root: str | None = None, read_only: bool = False):
         self.editor = FileEditor(workspace_root=workspace_root)
+        self.read_only = read_only
 
     def __call__(self, action: FileEditorAction) -> FileEditorObservation:
+        # Enforce read-only restrictions
+        if self.read_only and action.command not in ["view", "list"]:
```
xingyaoww: I thought we only allow for "view", why "list" is also here?

simonrosenberg:
@OpenHands please fix the failing actions on PR #624 at branch `add-bash-tool-to-planning-agent`

simonrosenberg:
@openhands please solve the test failures

simonrosenberg:
@xingyaoww In another PR I could update the FileEditorTool to accept a `editable_file_paths: list[str]` and make sure that only those files can be edited? 
This way it will be simpler to implement the PlanWriterTool  (similar strategy to read-only=True for the FileViewerTool)

xingyaoww:
> @xingyaoww In another PR I could update the FileEditorTool to accept a editable_file_paths: list[str] and make sure that only those files can be edited?

sounds good! @simonrosenberg

examples/26_planning_agent_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+import os
+import tempfile
+from pathlib import Path
+
+from pydantic import SecretStr
+
+from openhands.sdk import LLM, Conversation
+from openhands.sdk.llm import content_to_str
+from openhands.tools.preset import get_planning_agent
+from openhands.tools.preset.default import get_default_agent
```
xingyaoww, line 21: Maybe import explicitly as preset.planning?

openhands/tools/common.py
```diff
@@ -0,0 +1,8 @@
+"""Common constants and utilities for tools."""
+
+BASE_FILE_TOOL_DESCRIPTION = """Custom{editing_capabilities} tool for viewing{extra_viewing_capabilities} files and directories in plain-text format.
+* State is persistent across command calls and discussions with the user
+* If `path` is a text file, `view` displays the result of applying `cat -n`. If `path` is a directory, `view` lists non-hidden files and directories up to 2 levels deep{extra_features}
+* If the tool generates a long output, it will be truncated and marked with `<response clipped>`
+* When working with files, always use absolute files paths (starting with /){extra_features}
+"""  # noqa
```
xingyaoww: This seems like an overkill to me... Shall we just duplicate the Tool description as raw string here?

openhands/tools/file_editor/definition.py
```diff
@@ -153,17 +155,12 @@ def _has_meaningful_diff(self) -> bool:
...
 ]
 
+extra_features = """
 
-TOOL_DESCRIPTION = """Custom editing tool for viewing, creating and editing files in plain-text format
-* State is persistent across command calls and discussions with the user
-* If `path` is a text file, `view` displays the result of applying `cat -n`. If `path` is a directory, `view` lists non-hidden files and directories up to 2 levels deep
-* The following binary file extensions can be viewed in Markdown format: [".xlsx", ".pptx", ".wav", ".mp3", ".m4a", ".flac", ".pdf", ".docx"]. IT DOES NOT HANDLE IMAGES.
 * The `create` command cannot be used if the specified `path` already exists as a file
-* If a `command` generates a long output, it will be truncated and marked with `<response clipped>`
```
xingyaoww, line 162: think this is still true?

simonrosenberg:
@openhands add a `_check_ripgrep_available()` and call it in file_glob/impl.py and grep/impl.py
If ripgrep is not available, log a warning to the user saying he should install ripgrep and fallback the implementation using regular grep and glob.
If regular grep and glob cannot implement the file_glob and grep tools, find another implementation

openhands/tools/plan_writer/definition.py
```diff
@@ -0,0 +1,142 @@
...
+1. EXACT MATCHING: The `old_str` parameter must match EXACTLY one or more consecutive lines from the file, including all whitespace and indentation. The tool will fail if `old_str` matches multiple locations or doesn't match exactly with the file content.
+
+2. UNIQUENESS: The `old_str` must uniquely identify a single instance in the file:
+   - Include sufficient context before and after the change point (3-5 lines recommended)
+   - If not unique, the replacement will not be performed
+
+3. REPLACEMENT: The `new_str` parameter should contain the edited lines that replace the `old_str`. Both strings must be different.
+
+Remember: when making multiple file edits in a row to the same file, you should prefer to send all edits in a single message with multiple calls to this tool, rather than multiple messages with a single call each.
+"""  # noqa # TODO make base prompt for editing rules
```
xingyaoww: @simonrosenberg We are already duplicating some part of the FileEditor here -- maybe we should just implement what you've suggested earler and add a `allowed_edits_files` fields to FileEditorExecutor. This way we should be able to compeletely get rid of plan_write and file_viewer, instead just giving agent an restricted version of file-editor -- this should help reduce the complexity here a lot

examples/26_planning_agent_workflow.py
simonrosenberg: Or both from preset + change all other imports to import from preset?

xingyaoww: I actually prefer we keep things explicit here -- so people know they are bring in WHICH present

openhands/tools/file_editor/definition.py
simonrosenberg: yes https://vscode.dev/github/All-Hands-AI/agent-sdk/blob/add-bash-tool-to-planning-agent/openhands/tools/file_editor/editor.py#L528

simonrosenberg:
@openhands ripgrep might not be available when using file_glob and grep tools.
Please add a `is_ripgrep_available() -> bool` method
If it is not available, file_glob/impl.py and grep/impl.py should fallback to implementations without ripgrep. Use simple grep and glob commands if possible

simonrosenberg:
@xingyaoww I ran the example with and without ripgrep installed on a task that involves search in the agent-sdk repository. The listed files are the same in both cases.

openhands/tools/preset/planning.py
```diff
@@ -0,0 +1,100 @@
...
+
+def get_planning_tools() -> list[Tool]:
+    """Get the planning agent tool specifications.
+
+    Returns:
+        List of tools optimized for planning and analysis tasks, including
+        file viewing and PLAN.md editing capabilities for advanced
+        code discovery and navigation.
+    """
+    register_planning_tools()
```
xingyaoww, line 35: Will calling `get_planning_tools` multiple times cause us to regiter the tool multiple times (hence causing some duplication error?)

openhands/tools/planning_file_editor/definition.py
```diff
@@ -0,0 +1,116 @@
...
+    kind: str = "PlanningFileEditorAction"
+
+
+class PlanningFileEditorObservation(FileEditorObservation):
+    """Observation from planning file editor operations.
+
+    Inherits from FileEditorObservation - same structure, just different type.
+    """
+
+    kind: str = "PlanningFileEditorObservation"
```
xingyaoww: Why can't we just re-use `FileEditorAction`/`FileEditorObservation` here?

tests/tools/glob/test_consistency.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+        fallback_result = executor._execute_with_glob(action, Path(temp_dir_with_files))
+
+        # Both should succeed
+        assert not ripgrep_result.error
+        assert not fallback_result.error
+
+        # Both should find Python files
+        assert len(ripgrep_result.files) > 0
+        assert len(fallback_result.files) > 0
```
xingyaoww: We should enforce stricter check here (length should be the SAME)

tests/tools/glob/test_consistency.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Convert to basenames for comparison (paths might differ slightly)
+        ripgrep_basenames = {Path(f).name for f in ripgrep_result.files}
+        fallback_basenames = {Path(f).name for f in fallback_result.files}
+
+        # Should find the same files (at least the top-level Python files)
+        expected_files = {"app.py", "main.py", "test.py"}
+        assert expected_files.issubset(ripgrep_basenames)
+        assert expected_files.issubset(fallback_basenames)
```
xingyaoww: Shouldn't they be able to find all the file?

tests/tools/glob/test_consistency.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+        fallback_result = executor._execute_with_glob(action, Path(temp_dir_with_files))
+
+        # Both should succeed
+        assert not ripgrep_result.error
+        assert not fallback_result.error
+
+        # Both should find Python files including in subdirectories
+        assert len(ripgrep_result.files) >= 5  # At least 5 Python files
+        assert len(fallback_result.files) >= 5
```
xingyaoww: Let's check for the exact number instead of a rough range

tests/tools/glob/test_consistency.py
```diff
@@ -0,0 +1,143 @@
...
+            "openhands.tools.glob.impl._check_ripgrep_available",
+            return_value=False,
+        ):
+            executor = GlobExecutor(temp_dir_with_files)
+            assert not executor._ripgrep_available
+
+            # Should use fallback method
+            result = executor(action)
+            assert not result.error
+            assert len(result.files) > 0
```
xingyaoww: again, we should assert the previous block and this block have the same result

tests/tools/preset/test_planning_preset.py
xingyaoww, line 1: Not 100% sure if this test is necessary though

openhands/tools/planning_file_editor/definition.py
simonrosenberg: When reading the conversation history/logs, it makes more sense to have the PlanningFileEditorTool return a PlanningFileEditorObservation? 
I did not check if those class names were serialized and sent to `llm.completion` during a conversation run but if that's the case I don't think we can use the same class for those objects? It will confuse the llm?

xingyaoww: In that case, i think we shouldn't override kind which will be setted by the discriminative union mixin

xingyaoww: In that case, i think we shouldn't override kind which will be setted by the discriminative union mixin

tests/tools/preset/test_planning_preset.py
simonrosenberg: this test mimicks `test_default_preset` but I agree it's not critical. Should I remove both?

openhands/tools/preset/planning.py
simonrosenberg: No duplication error the data is stored as key:value in a dict` _REG[name] = resolver`

tests/tools/preset/test_planning_preset.py
xingyaoww: let's remove both!

PR #681, opened 2025-10-08
Add OpenTelemetry with Laminar SDK

dinmukhamedm:
Status / Todo list:
- [x] Basic integration with Laminar
    - [X] Laminar initialization and observe the core functions
    - [X] One trace per running conversation
    - [X] One session for traces with same conversation ID
    - [x] Simplify the integration by updating Laminar SDK (no start_active_span implementation exposed)
    - [x] Simplify the integration by adding Laminar dependency
- [x] Observe/trace tool calls
- [x] Observe/trace browser-use integration
    - [x] general code structure, cdp-use and browser-use tools tracing
    - [x] session replays if OTel backend is Laminar
- [x] final touches, e.g. input and output transformers where needed

Example browser-use session replay
Screenshot 2025-10-17 at 00 00 56

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -3,8 +3,6 @@ import uuid
 from collections.abc import Mapping
 from pathlib import Path
 
-from opentelemetry.trace import Span
-
 from openhands.sdk.agent.base import AgentBase
 from openhands.sdk.conversation.base import BaseConversation
 from openhands.sdk.conversation.exceptions import ConversationRunError
@@ -50,7 +48,6 @@ class LocalConversation(BaseConversation):
     _stuck_detector: StuckDetector | None
     llm_registry: LLMRegistry
     _cleanup_initiated: bool
-    _span: Span | None
 
     def __init__(
         self,
--- a/openhands-sdk/openhands/sdk/observability/laminar.py
+++ b/openhands-sdk/openhands/sdk/observability/laminar.py
@@ -1,5 +1,4 @@
 from collections.abc import Callable
-from contextvars import Context, Token
 from typing import (
     Any,
     Literal,
@@ -119,14 +118,14 @@ class SpanManager:
     """Manages a stack of active spans and their associated tokens."""
 
     def __init__(self):
-        self._stack: list[tuple[trace.Span, Token[Context] | None]] = []
+        self._stack: list[trace.Span] = []
 
     def start_active_span(self, name: str, session_id: str | None = None) -> None:
         """Start a new active span and push it to the stack."""
-        span, token = Laminar.start_active_span(name)
+        span = Laminar.start_active_span(name)
         if session_id:
             Laminar.set_trace_session_id(session_id)
-        self._stack.append((span, token))
+        self._stack.append(span)
 
     def end_active_span(self) -> None:
         """End the most recent active span by popping it from the stack."""
@@ -134,26 +133,34 @@ class SpanManager:
             logger.warning("Attempted to end active span, but stack is empty")
             return
 
-        span, token = self._stack.pop()
-        if token is not None:
-            Laminar.end_active_span(span, token)
-        elif span and span.is_recording():
-            span.end()
+        try:
+            span = self._stack.pop()
+            if span and span.is_recording():
+                span.end()
+        except IndexError:
+            logger.warning("Attempted to end active span, but stack is empty")
+            return
 
 
-# Global instance for convenience
-_span_manager = SpanManager()
+_span_manager: SpanManager | None = None
+
+
+def _get_span_manager() -> SpanManager:
+    global _span_manager
+    if _span_manager is None:
+        _span_manager = SpanManager()
+    return _span_manager
 
 
 def start_active_span(name: str, session_id: str | None = None) -> None:
     """Start a new active span using the global span manager."""
-    _span_manager.start_active_span(name, session_id)
+    _get_span_manager().start_active_span(name, session_id)
 
 
 def end_active_span() -> None:
     """End the most recent active span using the global span manager."""
     try:
-        _span_manager.end_active_span()
+        _get_span_manager().end_active_span()
     except Exception:
         logger.debug("Error ending active span")
         pass
--- a/openhands-sdk/pyproject.toml
+++ b/openhands-sdk/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
     "python-json-logger>=3.3.0",
     "tenacity>=9.1.2",
     "websockets>=12",
-    "lmnr>=0.7.18"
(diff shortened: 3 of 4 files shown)


dinmukhamedm:
@raymyers marking this ready for review, this is in a good state for an initial version

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -403,7 +411,22 @@ def _execute_action_event(
...
 
         # Execute actions!
-        observation: Observation = tool(action_event.action, conversation)
+        if should_enable_observability():
+            try:
+                if (
+                    action_event.action is not None
+                    and hasattr(action_event.action, "kind")
+                ):
+                    tool_name = action_event.action.kind
```
enyst: Sorry, is this code necessary?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -242,6 +256,7 @@ def run(self) -> None:
                         AgentExecutionStatus.PAUSED,
                         AgentExecutionStatus.STUCK,
                     ]:
+                        end_active_span()
```
enyst: Could the end span be at the end of the method, after the exception block?

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -641,6 +653,11 @@ def close(self) -> None:
         except Exception:
             pass
 
+        try:
+            end_active_span()
+        except Exception:
```
enyst: Sorry, could you maybe tell how does it work, why is the try / except necessary?

openhands-sdk/openhands/sdk/agent/agent.py
dinmukhamedm: Yes, the span name is very useful in debugging, it quickly shows the action in the trace. Here's an example.

Image

I'm happy to move the code into observability/utils.py though, if needed

openhands-sdk/openhands/sdk/observability/laminar.py
```diff
@@ -0,0 +1,159 @@
...
+    return False
+
+
+def _is_otel_backend_laminar():
+    """Simple heuristic to check if the OTEL backend is Laminar.
+    Caveat: This will still be True if another backend uses the same
+    authentication scheme, and the user uses LMNR_PROJECT_API_KEY
+    instead of OTEL_HEADERS to authenticate.
+    """
+    key = get_env("LMNR_PROJECT_API_KEY")
```
enyst, line 113: IMHO this is perfectly fine, recognizing a key with `LMNR` is fair!

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
dinmukhamedm: This makes sense. I think this placing here remains from the older version of this code. I'll move this outside, perhaps, into a `finally` block

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
dinmukhamedm: Good catch. No, it's not needed, internal exceptions are already handled on the observability handler. I will remove this

raymyers:
Appreciate this. I'll be working with @malhotra5 to validate this from the CLI shortly

raymyers:
We were able to get this sending traces to Honeycomb from OpenHands CLI, confirming vendor neutrality. The env vars needed to be tweaked a little, I will see if there's something that can be called standard.

image

dinmukhamedm:
@raymyers glad to see that and thanks for testing with CLI.

I see that LocalFileStore.read spans create much noise; I was already thinking of removing them from the set of observed functions. What are your thoughts?

skull8888888:
@raymyers would love to tweak experience around variables if you can point out the main frictions.

raymyers:
Hi @dinmukhamedm, I notice you still have one item marked as TODO "final touches, e.g. input and output transformers where needed". Do you consider what's there to be ready as is?

dinmukhamedm:
Hey @raymyers . Yes, I think what's there is ready. Let me push one more change with:
- remove the noisy span: https://github.com/OpenHands/software-agent-sdk/pull/681#issuecomment-3458694209
- rebase to resolve conflicts

Should take me no more than 20 minutes

dinmukhamedm:
Done. Now fully ready. @raymyers

openhands-sdk/openhands/sdk/observability/laminar.py
```diff
@@ -0,0 +1,159 @@
...
+
+        span, token = self._stack.pop()
+        if token is not None:
+            Laminar.end_active_span(span, token)
+        elif span and span.is_recording():
+            span.end()
+
+
+# Global instance for convenience
+_span_manager = SpanManager()
```
enyst: I think this case is not particularly risky, but it would be nice IMHO to avoid a global on import path if we can

openhands-sdk/openhands/sdk/observability/utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def get_env(key: str) -> str | None:
+    """Get an environment variable from the environment or the dotenv file."""
+    return os.getenv(key) or dotenv_values().get(key)
+
+
+def extract_action_name(action_event: ActionEvent) -> str:
+    try:
+        if action_event.action is not None and hasattr(action_event.action, "kind"):
```
enyst, line 15: Tiny nit: is it possible that it doesn't have attribute 'kind'?

openhands-sdk/openhands/sdk/observability/laminar.py
dinmukhamedm: I updated it to use singleton initialization pattern

openhands-sdk/openhands/sdk/observability/utils.py
dinmukhamedm: No, it inherits from `DiscriminatedUnionMixin`, but we are trying to be robust against small and subtle code changes, which I found deep inheritance to be prone to. The idea is for us to break as little as possible in your code and almost be invisible.

Happy to remove this check though, if you think it hinders readability

dinmukhamedm:
@raymyers @enyst I addressed the latest feedback. I've also moved some of the context management into our SDK, making the span management even simpler.

raymyers:
@dinmukhamedm 
> I see that LocalFileStore.read spans create much noise; I was already thinking of removing them from the set of observed functions. What are your thoughts?

I was on the fence at first since I could picture there being some problem with FileIO but I agree with the change you made there.

raymyers:
I believe all feedback is addressed any future changes needed can be a new PR. I'm just going to give a heads up and hopefully merge soon. @dinmukhamedm thanks for all your work on this.

PR #682, opened 2025-10-08
refactor DockerWorkspace with changes from V0

xingyaoww:


xingyaoww:
ugh I think this is still overcomplicated..

Maybe we can just live with:
1. a lite version of docker image (without vscode, etc -- so copy in one agent-server binary would just work)
2. a utility script that allows you to "batch build these lite version image" and push to dockerhub - we can just load image there for runtime-api
3. a fully featured agent-server with vnc/vscode, etc for production

PR #692, opened 2025-10-09
[Bug] Fix gpt 5 responses api by using latest litellm

simonrosenberg:
@enyst  had to make changes to your code else I was getting errors

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/all-hands-ai/agent-server:739bd7e-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-739bd7e-python \
  ghcr.io/all-hands-ai/agent-server:739bd7e-python
```

```
ghcr.io/all-hands-ai/agent-server:739bd7e-golang
ghcr.io/all-hands-ai/agent-server:v1.0.0_golang_tag_1.21-bookworm
ghcr.io/all-hands-ai/agent-server:739bd7e-java
ghcr.io/all-hands-ai/agent-server:v1.0.0_eclipse-temurin_tag_17-jdk
ghcr.io/all-hands-ai/agent-server:739bd7e-python
ghcr.io/all-hands-ai/agent-server:v1.0.0_nikolaik_s_python-nodejs_tag_python3.12-nodejs22
```

_The `739bd7e` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/tests/sdk/llm/test_responses_parsing_and_kwargs.py
+++ b/tests/sdk/llm/test_responses_parsing_and_kwargs.py
@@ -72,8 +72,8 @@ def test_normalize_responses_kwargs_policy():
     assert out["tool_choice"] == "auto"
     # include should contain original and encrypted_content
     assert set(out["include"]) >= {"text.output_text", "reasoning.encrypted_content"}
-    # store default to False when None passed
-    assert out["store"] is False
+    # store defaults to True for multi-turn conversations
+    assert out["store"] is True
     # reasoning config defaulted
     r = out["reasoning"]
     assert r["effort"] in {"low", "medium", "high", "none"}
--- a/tests/sdk/llm/test_responses_serialization.py
+++ b/tests/sdk/llm/test_responses_serialization.py
@@ -78,6 +78,9 @@ def test_tool_to_responses_emits_function_call_output_with_fc_prefix():
 
 
 def test_assistant_includes_reasoning_passthrough():
+    # Reasoning items are NOT included when serializing assistant messages for input.
+    # They are only captured from output. Including them causes API errors:
+    # "Item of type 'reasoning' was provided without its required following item"
     ri = ReasoningItemModel(
         id="rid1",
         summary=["s1", "s2"],
@@ -88,12 +91,5 @@ def test_assistant_includes_reasoning_passthrough():
     m = Message(role="assistant", content=[], responses_reasoning_item=ri)
     out = m.to_responses_dict(vision_enabled=False)
 
-    # Contains a reasoning item with exact passthrough fields
     r_items = [it for it in out if it["type"] == "reasoning"]
-    assert len(r_items) == 1
-    r = r_items[0]
-    assert r["id"] == "rid1"
-    assert [s["text"] for s in r["summary"]] == ["s1", "s2"]
-    assert [c["text"] for c in r.get("content", [])] == ["c1"]
-    assert r.get("encrypted_content") == "enc"
-    assert r.get("status") == "completed"
+    assert len(r_items) == 0


openhands/sdk/llm/message.py
```diff
@@ -397,31 +397,14 @@ def to_responses_dict(self, *, vision_enabled: bool) -> list[dict[str, Any]]:
...
-                    if ri.status:
-                        reasoning_item["status"] = ri.status
-                    items.append(reasoning_item)
+            # NOTE: We do NOT include reasoning items in the input when sending back
+            # assistant messages. Reasoning items are only captured from output.
+            # Including them in input causes API errors:
+            # "Item of type 'reasoning' was provided
+            # without its required following item"
+            # This is because reasoning items must
+            # be followed by a message/function_call in the Responses API input schema.
```
xingyaoww, line 406: interesting.. @enyst 
Shall we create an issue for this - eg how to repro and what's the specific stack trace and we can fix it later

openhands/sdk/agent/agent.py
```diff
@@ -179,7 +179,7 @@ def step(
                     messages=_messages,
                     tools=list(self.tools_map.values()),
                     include=None,
-                    store=False,
+                    store=True,
```
enyst, line 182: `store=True` is the stateful variant, where OpenAI stores the conversation history on their servers
`store=False` is what I think we want to be the default behavior in openhands 🤔

We need them both, IMHO

openhands/sdk/llm/message.py
enyst: The error sounds like the Message with reasoning was empty, it didn't have a content or a function call 🤔

openhands/sdk/llm/message.py
```diff
@@ -397,31 +397,14 @@ def to_responses_dict(self, *, vision_enabled: bool) -> list[dict[str, Any]]:
...
-                        ]
-                    # Optional fields as received
-                    if ri.encrypted_content:
-                        reasoning_item["encrypted_content"] = ri.encrypted_content
-                    if ri.status:
-                        reasoning_item["status"] = ri.status
-                    items.append(reasoning_item)
+            # NOTE: We do NOT include reasoning items in the input when sending back
+            # assistant messages. Reasoning items are only captured from output.
+            # Including them in input causes API errors:
```
enyst, line 402: Including them in input is necessary for the LLM to continue its train of thought, instead of re-reasoning, so it seems this changes behavior: according to OpenAI docs, its performance is a little worse
https://platform.openai.com/docs/guides/reasoning#keeping-reasoning-items-in-context

simonrosenberg: I feel like gpt-5 just added extra conversation validation with matching tool ids etc.. I am afraid this will break with condensation

simonrosenberg:
> Thank you for looking into this!
> 
> I think this is maybe not the solution:
> 
> * stateless variant implies that we manage context, and thus send back message history, including reasoning items
> * stateful variant stores message history on OpenAI servers, thus we can just write a `previous response id` instead of history and it should continue where it left off.
> 
> IMHO the PR behavior, which enables stateful, but still manages context, and sends it back to the LLM API, but skips reasoning from it, is a mix which... doesn't really match any. I feel like it's a workaround around litellm behavior, maybe? But I don't understand why we say that litellm fixed it, could you please add to PR description a link to litellm issue / fix ?

litellm fixed a first bug. Then I had a second a third bug. You are right perhaps all this is a bad patch around litellm bugs

simonrosenberg:
> Thank you for looking into this!
> 
> I think this is maybe not the solution:
> 
> * stateless variant implies that we manage context, and thus send back message history, including reasoning items
> * stateful variant stores message history on OpenAI servers, thus we can just write a `previous response id` instead of history and it should continue where it left off.
> 
> IMHO the PR behavior, which enables stateful, but still manages context, and sends it back to the LLM API, but skips reasoning from it, is a mix which... doesn't really match any. I feel like it's a workaround around litellm behavior, maybe? But I don't understand why we say that litellm fixed it, could you please add to PR description a link to litellm issue / fix ?

https://github.com/BerriAI/litellm/pull/15347

simonrosenberg:
> Thank you for looking into this!
> 
> I think this is maybe not the solution:
> 
> * stateless variant implies that we manage context, and thus send back message history, including reasoning items
> * stateful variant stores message history on OpenAI servers, thus we can just write a `previous response id` instead of history and it should continue where it left off.
> 
> IMHO the PR behavior, which enables stateful, but still manages context, and sends it back to the LLM API, but skips reasoning from it, is a mix which... doesn't really match any. I feel like it's a workaround around litellm behavior, maybe? But I don't understand why we say that litellm fixed it, could you please add to PR description a link to litellm issue / fix ?

Trying again with enable_encrypted_reasoning set to True

openhands/sdk/llm/message.py
enyst: Good point! I made an issue on condensation

The weird thing is that we don't see this breaking on `openai/gpt-5`? So I worry that maybe litellm still routes it wrong or partially set

How do you feel about... a bunch of observability/debug logging, including the exact request sent from us to litellm, and enable litellm debug so that we grab the request from litellm to openai; and get trajectories from multiple integration tests runs fully, with a setting or another? Just to stare at, and maybe see what influences what

enyst:
> litellm fixed a first bug. Then I had a second a third bug. 

Just a thought, I feel like maybe what happens here, is that `store=True` or encrypted or some other responses-specific parameter tells litellm to route to responses, somewhere in their code? In other words, the PR linked fixes some stuff about that config, but maybe litellm still does the weird fallback to completions, and for store=True it doesn't, if that makes sense

enyst:
Superseded by https://github.com/All-Hands-AI/agent-sdk/pull/699

PR #718, opened 2025-10-12
Debug: glob tool fails when pattern contains path 

simonrosenberg:
The agent calls the glob tool with pattern = f"{path}/pattern":
Example: pattern: "/Users/simonrosenberg/Desktop/work/repositories/All-Hands-AI/agent-sdk/**/*.py"  path=None
Because path is None, the tool defaults to path=workspace when actually the path is contained in the pattern.

The agent "ignored" the API design and put the absolute path in the pattern anyway. I guess that's because `glob` commands usually don't have two arguments. 

However RipGrep expects two separates arguments: path and pattern.
Hence we need to extract the path from the pattern if the LLM outputs such a pattern.

Another solution would be prompting the tool to not allow path-and-pattern.
This is more complex but more robust.    

P.S. I think this "error" was also occuring in the OpenHands repo since the tools have the same description. 

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:d4599d6-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-d4599d6-python \
  ghcr.io/openhands/agent-server:d4599d6-python
```

```
ghcr.io/openhands/agent-server:d4599d6-golang
ghcr.io/openhands/agent-server:d4599d6-java
ghcr.io/openhands/agent-server:d4599d6-python
```

_The `d4599d6` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/openhands-tools/openhands/tools/glob/impl.py
+++ b/openhands-tools/openhands/tools/glob/impl.py
@@ -228,10 +228,10 @@ class GlobExecutor(ToolExecutor[GlobAction, GlobObservation]):
         path_obj = Path(pattern)
         parts = path_obj.parts
 
-        # Find where the glob characters start
+        # Find where the glob characters start using glob.has_magic()
         search_parts = []
         for part in parts:
-            if any(ch in part for ch in "*?[]"):
+            if glob_module.has_magic(part):
                 break
             search_parts.append(part)
 


simonrosenberg:
@openhands please resolve merge conflicts and make sure all tests and pre-commit checks are passing

simonrosenberg:
@openhands please resolve all conflicts. Make sure all test are passing as well as the pre-commit hook

openhands-tools/openhands/tools/glob/impl.py
```diff
@@ -44,23 +44,38 @@ def __call__(self, action: GlobAction) -> GlobObservation:
...
-                    return GlobObservation(
-                        files=[],
-                        pattern=action.pattern,
-                        search_path=str(search_path),
-                        error=f"Search path '{action.path}' is not a valid directory",
-                    )
+                pattern = action.pattern
             else:
-                search_path = self.working_dir
+                extracted_path, pattern = self._extract_search_path_from_pattern(
```
xingyaoww, line 53: I guess if the agent provides an absolute path, can we just not set the search_path? eg we don't set the `--files path` -- this should help us simplifying the logic a bit

xingyaoww (commented):
left some suggestions on potentially simplify this

openhands-tools/openhands/tools/glob/impl.py
simonrosenberg: not sure I understand, you mean:  if the agent outputs `"/path/to/dir/**/*.py"`, only send `**/*.py` to ripgrep?
I don't think it's a good idea: that will make the call slower + confuse the agent because the output doesn't match the input + output truncated if output is too large

xingyaoww: I mean if we can just send "/path/to/dir/**/*.py" to ripgrep and unset the folder?

simonrosenberg: oh but ripgrep doesn't accept that else the current code would work. let me double check

openhands-tools/openhands/tools/glob/impl.py
```diff
@@ -152,11 +179,70 @@ def _execute_with_glob(
...
+            # Pattern starts with glob at root (e.g., "/*/*.py")
+            search_path = Path("/")
+            adjusted_pattern = pattern.lstrip("/")
+        else:
+            search_path = Path(*search_parts)
+            # Get the remaining parts as the pattern
+            remaining = parts[len(search_parts) :]
+            adjusted_pattern = str(Path(*remaining)) if remaining else "**/*"
+
+        return search_path.resolve(), adjusted_pattern
```
xingyaoww, line 248: @OpenHands are there any opportunity to simplify this snippet of code by using some existing python library to parse the string? eg glob

openhands-tools/openhands/tools/glob/impl.py
simonrosenberg: Screenshot 2025-10-21 at 16 16 20

xingyaoww: make sense to me!

PR #723, opened 2025-10-13
Runtime errors with security risk and no configured security analyzer

### LINKED ISSUE #710: Bug: reconciliation of agent without security analyzer and tools security risk field
Here are the list of instructions to perform in V1 CLI to reproduce the issue (not all prompts easily reproduce this issue)

1. prompt the agent "create an empty hello.md file"
2. click "always proceed" for agent response
3. prompt the agent "please run bash command with security risk"
4. prompt the agent "please run `str_replace_editor` and include `security_risk` value"

This will result in the error 

```
RuntimeError: LLM provided a security_risk but no security analyzer is configured - THIS SHOULD NOT HAPPEN!
```

You'll also notice that the agent says the bash tool doesn't contain `security_risk` field but `str_replace_editor` does

Note that it is confirmed that "always proceed" will create a new agent object, and a new conversation object resuming from the persistent state. The new agent object's security analyzer is removed via 

```
agent = agent.model_copy(
                update={"security_analyzer": None}
            )
```

and is passed into the conversation object `conversation = Conversation(agent=agent)`. The conversation object should invoke `agent.init_state` which in theory will make sure that the system prompt and tools don't contain any security analyzer/risk information, though it seems like there is a bug somewhere that does otherwise

malhotra5:
V1 CLI will remove the security analyzer and restart a new conversation to disable confirmation mode. However, this 

1. doesn't update the system prompt (immutable)
2. agent still passes `security_risk` field, and crashes with runtime error (system prompt urges to pass in this field)
3. security risk is not an official argument validated by the toolschema. we pop the argument, so the runtime error cannot be sent back as feedback to the agent 

This means

1. disabling confirmation mode does not invalidate the prompt cache as initially hypothesized
2. the agent is unaware of confirmation mode being disabled and will still pass in `security_risk` field, which explains why the CLI will crash unpredictably depending on the trajectory 

This PR ensures that we don't raise the runtime error if `security_risk` is passed and no analyzer has been set.

Fixes #710 

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/all-hands-ai/agent-server:6790589-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-6790589-python \
  ghcr.io/all-hands-ai/agent-server:6790589-python
```

```
ghcr.io/all-hands-ai/agent-server:6790589-golang
ghcr.io/all-hands-ai/agent-server:v1.0.0_golang_tag_1.21-bookworm_binary
ghcr.io/all-hands-ai/agent-server:6790589-java
ghcr.io/all-hands-ai/agent-server:v1.0.0_eclipse-temurin_tag_17-jdk_binary
ghcr.io/all-hands-ai/agent-server:6790589-python
ghcr.io/all-hands-ai/agent-server:v1.0.0_nikolaik_s_python-nodejs_tag_python3.12-nodejs22_binary
```

_The `6790589` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -11,5 +11,8 @@ include openhands/agent_server/docker/Dockerfile
 include openhands/agent_server/docker/wallpaper.svg
 include openhands/agent_server/agent-server.spec
 
+# Include VSCode extensions so Docker build context has them
+recursive-include openhands/agent_server/vscode_extensions *
+
 # include all prompts
 recursive-include openhands/sdk *.j2
--- a/examples/02_remote_agent_server/04_vscode_with_docker_sandboxed_server.py
+++ b/examples/02_remote_agent_server/04_vscode_with_docker_sandboxed_server.py
@@ -1,0 +1,102 @@
+import os
+import time
+
+import httpx
+from pydantic import SecretStr
+
+from openhands.sdk import LLM, Conversation, get_logger
+from openhands.sdk.conversation.impl.remote_conversation import RemoteConversation
+from openhands.tools.preset.default import get_default_agent
+from openhands.workspace import DockerWorkspace
+
+
+logger = get_logger(__name__)
+
+
+api_key = os.getenv("LLM_API_KEY")
+assert api_key is not None, "LLM_API_KEY environment variable is not set."
+
+llm = LLM(
+    service_id="agent",
+    model="litellm_proxy/anthropic/claude-sonnet-4-5-20250929",
+    base_url="https://llm-proxy.eval.all-hands.dev",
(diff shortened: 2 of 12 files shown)


malhotra5:
@openhands read the diff of this PR, do not modify the code

please add a unit test which confirms that if a tool response contains `security_risk` param, but no security analyzer is configured the agent should proceed normally

openhands/sdk/agent/agent.py
xingyaoww: Maybe we should add assert "security_risk" not in `arguments` just to make sure here

PR #730, opened 2025-10-13
fix: working DockerFile for SWE-Bench prebuilt images

juanmichelini:


openhands/agent_server/docker/Dockerfile
```diff
@@ -190,16 +192,49 @@ EXPOSE ${PORT} ${NOVNC_PORT}
 
 ############################
 # Target A: source
```
xingyaoww, line 192: Can you investigate why this won't copy over the `/agent-server/.venv` though 🤔

xingyaoww:
I think this was already fixed

PR #775, opened 2025-10-17
Enforce batch atomicity for condenser

### LINKED ISSUE #738: If an assistant message contains any thinking blocks, the first block must be `thinking` or `redacted_thinking`
```
    raise e  # it's already mapped
    ^^^^^^^
  File "/Users/xingyaow/.cache/uv/archive-v0/7syNfyMibYRIdkqALTht7/lib/python3.12/site-packages/litellm/litellm_core_utils/exception_mapping_utils.py", line 290, in exception_type
    extract_and_raise_litellm_exception(
  File "/Users/xingyaow/.cache/uv/archive-v0/7syNfyMibYRIdkqALTht7/lib/python3.12/site-packages/litellm/litellm_core_utils/exception_mapping_utils.py", line 165, in extract_and_raise_litellm_exception
    raise raised_exception_obj(
litellm.exceptions.BadRequestError: litellm.BadRequestError: Error code: 400 - {'error': {'message': 'litellm.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"messages.5.content.0: If an assistant message contains any thinking blocks, the first block must be `thinking` or `redacted_thinking`. Found `tool_use`."},"request_id":"req_011CU7kqVmSUn8XEfzjx33MH"}No fallback model group found for original model_group=claude-sonnet-4-5-20250929. Fallbacks=[{\'qwen3-coder-480b\': [\'qwen3-coder-480b-or\']}, {\'glm-4.5\': [\'glm-4.5-or\']}]. Received Model Group=claude-sonnet-4-5-20250929\nAvailable Model Group Fallbacks=None\nError doing the fallback: litellm.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"messages.5.content.0: If an assistant message contains any thinking blocks, the first block must be `thinking` or `redacted_thinking`. Found `tool_use`."},"request_id":"req_011CU7kqVmSUn8XEfzjx33MH"}No fallback model group found for original model_group=claude-sonnet-4-5-20250929. Fallbacks=[{\'qwen3-coder-480b\': [\'qwen3-coder-480b-or\']}, {\'glm-4.5\': [\'glm-4.5-or\']}] LiteLLM Retried: 1 times, LiteLLM Max Retries: 2', 'type': None, 'param': None, 'code': '400'}}
```

The conversation dump: [c211c69a-9c5c-4f7c-a1be-1b7037001bca.zip](https://github.com/user-attachments/files/22913292/c211c69a-9c5c-4f7c-a1be-1b7037001bca.zip)

Way to reproduce this:
1. reload conversation and then conversation.run()

ryanhoangt:
Fix #738

---
Integration tests: 7/7 (100%)

```
[10/17/25 13:48:27] INFO     Success rate: 100.00% (7/7)             run_infer.py:278
[10/17/25 13:48:27] INFO     Evaluation Results:                     run_infer.py:279
[10/17/25 13:48:27] INFO     t04_git_staging: ✓ - Successfully       run_infer.py:283
                             committed changes with message: 'Add                    
                             hello world Python script'                              
[10/17/25 13:48:27] INFO     t07_interactive_commands: ✓ -           run_infer.py:283
                             Interactive Python script setup                         
                             completed. Agent should execute the                     
                             script with inputs 'John' and '25' and                  
                             find the secret number: 707                             
[10/17/25 13:48:27] INFO     t05_simple_browsing: ✓ - Agent          run_infer.py:283
                             successfully found the answer! Matched                  
                             pattern: (?i)openhands is all you need.                 
                             Response contained the expected content                 
                             about OpenHands.                                        
[10/17/25 13:48:27] INFO     t03_jupyter_write_file: ✓ -             run_infer.py:283
                             Successfully created file with content:                 
                             hello world                                             
[10/17/25 13:48:27] INFO     t01_fix_simple_typo: ✓ - Successfully   run_infer.py:283
                             fixed all typos                                         
[10/17/25 13:48:27] INFO     t02_add_bash_hello: ✓ - Successfully    run_infer.py:283
                             created and executed script: hello                      
[10/17/25 13:48:27] INFO     t06_github_pr_browsing: ✓ - Agent's     run_infer.py:283
                             final answer contains information about                 
                             the PR content                                          
[10/17/25 13:48:27] INFO     Total cost: $0.27                       run_infer.py:284
```

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/all-hands-ai/agent-server:14e6b9a-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-14e6b9a-python \
  ghcr.io/all-hands-ai/agent-server:14e6b9a-python
```

```
ghcr.io/all-hands-ai/agent-server:14e6b9a-golang
ghcr.io/all-hands-ai/agent-server:v1.0.0a2_golang_tag_1.21-bookworm_binary
ghcr.io/all-hands-ai/agent-server:14e6b9a-java
ghcr.io/all-hands-ai/agent-server:v1.0.0a2_eclipse-temurin_tag_17-jdk_binary
ghcr.io/all-hands-ai/agent-server:14e6b9a-python
ghcr.io/all-hands-ai/agent-server:v1.0.0a2_nikolaik_s_python-nodejs_tag_python3.12-nodejs22_binary
```

_The `14e6b9a` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/.github/workflows/assign-reviews.yml
+++ b/.github/workflows/assign-reviews.yml
@@ -94,8 +94,8 @@ jobs:
             - name: Install OpenHands dependencies
               run: |
                   # Install OpenHands SDK and tools from git repository
-                  uv pip install --system "openhands-sdk @ git+https://github.com/All-Hands-AI/agent-sdk.git@main#subdirectory=openhands/sdk"
-                  uv pip install --system "openhands-tools @ git+https://github.com/All-Hands-AI/agent-sdk.git@main#subdirectory=openhands/tools"
+                  uv pip install --system "openhands-sdk @ git+https://github.com/All-Hands-AI/agent-sdk.git@main#subdirectory=openhands-sdk"
+                  uv pip install --system "openhands-tools @ git+https://github.com/All-Hands-AI/agent-sdk.git@main#subdirectory=openhands-tools"
 
             - name: Check required configuration
               env:
--- a/openhands-agent-server/openhands/agent_server/__main__.py
+++ b/openhands-agent-server/openhands/agent_server/__main__.py
@@ -46,6 +46,7 @@ def main():
         reload_excludes=["workspace"],
         log_level=log_level,
         log_config=LOGGING_CONFIG,
+        ws="wsproto",  # Use wsproto instead of deprecated websockets implementation
     )
 
 
--- a/openhands-agent-server/pyproject.toml
+++ b/openhands-agent-server/pyproject.toml
@@ -13,6 +13,7 @@ dependencies = [
   "sqlalchemy>=2",
   "uvicorn>=0.31.1",
   "websockets>=12",
+  "wsproto>=1.2.0",
 ]
 
 [build-system]
--- a/tests/agent_server/test_agent_server_wsproto.py
+++ b/tests/agent_server/test_agent_server_wsproto.py
@@ -1,0 +1,106 @@
+"""Integration test to verify the agent server works with wsproto."""
+
+import asyncio
+import json
+import multiprocessing
+import os
+import socket
+import sys
+import time
+
+import pytest
+import requests
+import websockets
+
+
+def find_free_port():
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("", 0))
+        s.listen(1)
+        return s.getsockname()[1]
+
+
+def run_agent_server(port, api_key):
+    os.environ["OH_SESSION_API_KEYS"] = f'["{api_key}"]'
+    sys.argv = ["agent-server", "--port", str(port)]
+    from openhands.agent_server.__main__ import main
+
+    main()
+
+
+@pytest.fixture
+def agent_server(tmp_path):
+    port = find_free_port()
+    api_key = "test-wsproto-key"
+
+    process = multiprocessing.Process(target=run_agent_server, args=(port, api_key))
+    process.start()
+
+    for _ in range(50):
+        try:
+            response = requests.get(f"http://127.0.0.1:{port}/docs", timeout=1)
+            if response.status_code == 200:
+                break
+        except requests.exceptions.ConnectionError:
+            pass
+        time.sleep(0.2)
+    else:
+        process.terminate()
+        process.join()
+        pytest.fail(f"Agent server failed to start on port {port}")
+
+    yield {"port": port, "api_key": api_key}
+
+    process.terminate()
+    process.join(timeout=5)
+    if process.is_alive():
+        process.kill()
+        process.join()
+
+
+def test_agent_server_starts_with_wsproto(agent_server):
+    response = requests.get(f"http://127.0.0.1:{agent_server['port']}/docs")
+    assert response.status_code == 200
+    assert (
+        "OpenHands Agent Server" in response.text or "swagger" in response.text.lower()
+    )
+
+
+@pytest.mark.asyncio
+async def test_agent_server_websocket_with_wsproto(agent_server):
+    port = agent_server["port"]
+    api_key = agent_server["api_key"]
+
+    response = requests.post(
+        f"http://127.0.0.1:{port}/api/conversations",
+        headers={"X-Session-API-Key": api_key},
+        json={
+            "agent": {
+                "llm": {
+                    "service_id": "test-llm",
+                    "model": "test-provider/test-model",
+                    "api_key": "test-key",
+                },
+                "tools": [],
+            },
+            "workspace": {"working_dir": "/tmp/test-workspace"},
+        },
+    )
+    assert response.status_code in [200, 201]
+    conversation_id = response.json()["id"]
+
+    ws_url = (
+        f"ws://127.0.0.1:{port}/sockets/events/{conversation_id}"
+        f"?session_api_key={api_key}&resend_all=true"
+    )
+
+    async with websockets.connect(ws_url, open_timeout=5) as ws:
+        try:
+            response = await asyncio.wait_for(ws.recv(), timeout=2)
+            assert response is not None
+        except TimeoutError:
+            pass
+
+        await ws.send(
+            json.dumps({"role": "user", "content": "Hello from wsproto test"})
+        )
--- a/tests/github_workflows/test_todo_scanner.py
+++ b/tests/github_workflows/test_todo_scanner.py
@@ -9,7 +9,7 @@ from pathlib import Path
 todo_mgmt_path = (
     Path(__file__).parent.parent.parent
     / "examples"
-    / "github_workflows"
+    / "03_github_workflows"
     / "03_todo_management"
 )
 sys.path.append(str(todo_mgmt_path))
--- a/uv.lock
+++ b/uv.lock
@@ -1890,6 +1890,7 @@ dependencies = [
     { name = "sqlalchemy" },
     { name = "uvicorn" },
     { name = "websockets" },
+    { name = "wsproto" },
 ]
 
 [package.metadata]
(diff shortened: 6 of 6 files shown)


openhands/sdk/agent/agent.py
```diff
@@ -230,13 +230,15 @@ def step(
...
-                    thinking_blocks=list(message.thinking_blocks) if i == 0 else [],
-                    responses_reasoning_item=message.responses_reasoning_item
-                    if i == 0
-                    else None,
+                    # All actions get reasoning content to preserve it if condenser
+                    # removes earlier events
+                    reasoning_content=message.reasoning_content,
+                    # All actions get thinking blocks to preserve them if condenser
+                    # removes earlier events
+                    thinking_blocks=list(message.thinking_blocks),
```
enyst: Wouldn't this repeat them?

tests/integration/outputs/anthropic_claude_haiku_4_5_20251001_agent-sdk-integration_N7_20251017_093024/results.json
enyst, line 1: I suppose including these outputs was unintended, maybe we should remove them?

openhands/sdk/agent/agent.py
ryanhoangt: Yeah we have duplication when maintaining them in `ActionEvent`, but when combining multiple actions with the same llm response id into the same `Message`, we only use the first `thinking_blocks` in the event batch. WDYT?

https://github.com/All-Hands-AI/agent-sdk/blob/aa30853b522be7762cab016591f9ccb86e3671ba/openhands/sdk/event/base.py#L128-L148

enyst (changes requested):
I have a concern here, I _think_ that both Claude and GPT models may crash if we repeat the same reasoning block... but I could be wrong. Do we have a log with parallel tool calls on the PR branch?

ryanhoangt:
> Do we have a log with parallel tool calls on the PR branch?

Yep, it's in the associated issue.

openhands/sdk/agent/agent.py
enyst: Ahh! Sorry, I misunderstood and thought we're sending back multiple.

🤔 This could work... or maybe we could have the condenser verify its events, so it puts back the reasoning content... But I feel your solution is safer.

xingyaoww: > but when combining multiple actions with the same llm response id into the same Message, we only use the first thinking_blocks in the event batch. 

How would this fix https://github.com/All-Hands-AI/agent-sdk/issues/738 though? If condenser only keep the last 3 actions out of 4 actions (within 1 llm request via parallel fn call) -- we fix it because the "first of the 3 actions" now also have the thinking block?

I wonder if we were able to fix it on the condenser level, so we don't have to assign thinking block and reasoning contnet to every event (consume more space)

enyst: > I wonder if we were able to fix it on the condenser level, so we don't have to assign thinking block and reasoning contnet to every event

I see the point, and it seems like a condenser issue: maybe its contract should be to return a "good" set of events.

On the other hand I was thinking that maybe adding them then pick the first, is safer in the sense that we always do the right thing, if the condenser gets disabled in a later run, or changed parameters, or something... But maybe that's wrong, those would still be true if it's a condenser responsibility 🤔

ryanhoangt: > If condenser only keep the last 3 actions out of 4 actions (within 1 llm request via parallel fn call) -- we fix it because the "first of the 3 actions" now also have the thinking block?

I think it'd be easier to imagine the condenser, instead of _keeping_ actions, it _forgets_ actions, and one of the forgotten events (the first event) contains the thinking blocks. We fix it by populating it for other later events as well, and those later events, aka more _recent_ events are used for continuing the convo, now also have thinking blocks.

> (consume more space)

Yeah I also consider this limitation, but the number of thinking tokens would often be much less than chat completion tokens so hopefully it won't cause significant effects.

> I wonder if we were able to fix it on the condenser level

IMO this would require the condenser to know about thinking blocks, which seems to be more complex in term of implementation and potentially prevent effective condensation with the extra constraint I guess 🤔

xingyaoww: I guess we can introduce a new constraint to the condeser: if you want to remove an action, make sure you either remove all actions associated with ONE LLM call, or just keep them all -- there's no middle ground. It is the middle ground here that's causing this bug for us (i.e., part of the actions generated by LLM got forgotten) 🤔

ryanhoangt: Sounds good, let me give it a try.

enyst: TBH I agree with Xingyao. 🤔 They all have one `response_id`, and it would be simpler for more reasons, to either send them all or not send any of that batch.

The `View` docstring already says: "the included events are ready to process as LLM input". If they're not ready, we could consider it a bug.

https://github.com/All-Hands-AI/agent-sdk/blob/9a004ed185ba37752820466a7c387ececc73d2e8/openhands-sdk/openhands/sdk/context/view.py#L27

ryanhoangt: I've implemented this forget-all mechanism in the `view.py`, LMK if you have any concerns with it!

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -88,6 +88,40 @@ def __getitem__(
...
+    @staticmethod
+    def _enforce_batch_atomicity(
+        events: Sequence[Event],
+        forgotten_event_ids: set[EventID],
+    ) -> set[EventID]:
+        """Ensure that if any event in a batch is forgotten, all events in that
+        batch are forgotten.
+
+        This prevents partial batches from being sent to the LLM, which can cause
+        API errors when thinking blocks are separated from their tool calls.
```
enyst, line 100: Just a thought: I think maybe the action events are in the order we received tool calls from the LLM? If so, maybe we could check in a simpler way, if the event(s) just before `forgotten_event_ids` have the same `response_id` as... the first of those forgotten?

Really just a thought, it's all good with checking this way too

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -161,6 +195,11 @@ def from_events(events: Sequence[Event]) -> "View":
             if isinstance(event, CondensationRequest):
                 forgotten_event_ids.add(event.id)
 
+        # Enforce batch atomicity: if any event in a multi-action batch is forgotten,
+        # forget all events in that batch to prevent partial batches with thinking
+        # blocks separated from their tool calls
+        forgotten_event_ids = View._enforce_batch_atomicity(events, forgotten_event_ids)
```
csmith49, line 201: I think it might make sense to call this after we call `filter_unmatched_tool_calls`?  Not sure if it's even possible for that function to break up a batch but maybe better safe than sorry.

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -88,6 +88,40 @@ def __getitem__(
         else:
             raise ValueError(f"Invalid key type: {type(key)}")
 
+    @staticmethod
+    def _enforce_batch_atomicity(
+        events: Sequence[Event],
```
csmith49, line 93: If we take the other suggestion (on method call order) we can probably strengthen this type to `LLMConvertibleEvent`.

openhands-sdk/openhands/sdk/context/view.py
ryanhoangt: I think it makes more sense to do this event removal before `filter_unmatched_tool_calls`, since in the `_enforce_batch_atomicity` we only remove actions, not observations. Maybe putting the `filter_unmatched_tool_calls` at the end can help us remove some left-over observations that doesn't have a corresponding action.

ryanhoangt: Maybe I'll go with this for now, we can reconsider this later if we run into issues!

csmith49: Yeah, makes sense to me.

PR #795, opened 2025-10-18
feat(llm): add LLM profiles

enyst:
## Summary
- integrate LLM profile persistence into `LLMRegistry`, exposing list/load/save/register/validate helpers with configurable profile directories
- fix docs example checker to detect nested example paths (avoids false CI failures)

## Testing
- .venv/bin/python -m pytest tests/sdk/llm/test_llm_registry_profiles.py
- .venv/bin/python -m pytest tests/sdk/conversation/local/test_state_serialization.py 

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:0e60e67-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-0e60e67-python \
  ghcr.io/openhands/agent-server:0e60e67-python
```

```
ghcr.io/openhands/agent-server:0e60e67-golang-amd64
ghcr.io/openhands/agent-server:0e60e67-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:0e60e67-golang-arm64
ghcr.io/openhands/agent-server:0e60e67-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:0e60e67-java-amd64
ghcr.io/openhands/agent-server:0e60e67-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:0e60e67-java-arm64
ghcr.io/openhands/agent-server:0e60e67-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:0e60e67-python-amd64
ghcr.io/openhands/agent-server:0e60e67-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:0e60e67-python-arm64
ghcr.io/openhands/agent-server:0e60e67-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:0e60e67-golang
ghcr.io/openhands/agent-server:0e60e67-java
ghcr.io/openhands/agent-server:0e60e67-python
```

- Each variant tag (e.g., `0e60e67-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `0e60e67-python-amd64`) are also available if needed

Docs PR (documents new examples for check-examples): OpenHands/docs#219

### DIFF
--- a/tests/agent_server/test_agent_server_wsproto.py
+++ b/tests/agent_server/test_agent_server_wsproto.py
@@ -21,7 +21,7 @@ def find_free_port():
 
 
 def run_agent_server(port, api_key):
-    os.environ["OH_SESSION_API_KEYS"] = f'["{api_key}"]'
+    os.environ["OH_SESSION_API_KEYS_0"] = api_key
     sys.argv = ["agent-server", "--port", str(port)]
     from openhands.agent_server.__main__ import main
 
--- a/tests/github_workflows/test_resolve_model_config.py
+++ b/tests/github_workflows/test_resolve_model_config.py
@@ -28,8 +28,8 @@ def test_find_models_by_id_single_model():
         result = find_models_by_id(model_ids)
 
     assert len(result) == 1
-    assert result[0]["id"] == "claude-sonnet-4-5-20250929"
-    assert result[0]["display_name"] == "Claude Sonnet 4.5"
+    assert result[0]["id"] == "gpt-4"
+    assert result[0]["display_name"] == "GPT-4"
 
 
 def test_find_models_by_id_multiple_models():
@@ -45,8 +45,8 @@ def test_find_models_by_id_multiple_models():
         result = find_models_by_id(model_ids)
 
     assert len(result) == 2
-    assert result[0]["id"] == "claude-sonnet-4-5-20250929"
-    assert result[1]["id"] == "deepseek-chat"
+    assert result[0]["id"] == "gpt-4"
+    assert result[1]["id"] == "claude-3"
 
 
 def test_find_models_by_id_preserves_order():
@@ -113,11 +113,11 @@ def test_find_models_by_id_preserves_full_config():
         result = find_models_by_id(model_ids)
 
     assert len(result) == 1
-    assert result[0]["id"] == "claude-sonnet-4-5-20250929"
-    assert (
-        result[0]["llm_config"]["model"] == "litellm_proxy/claude-sonnet-4-5-20250929"
-    )
-    assert result[0]["llm_config"]["temperature"] == 0.0
+    assert result[0]["id"] == "custom-model"
+    assert result[0]["llm_config"]["model"] == "custom-model"
+    assert result[0]["llm_config"]["api_key"] == "test-key"
+    assert result[0]["llm_config"]["base_url"] == "https://example.com"
+    assert result[0]["extra_field"] == "should be preserved"
 
 
 # Tests for expected models from issue #1495


.gitignore
```diff
@@ -203,3 +203,15 @@ cache
...
+*.db
+.worktrees/
+*.code-workspace
+log.txt
+previous.md
+AGENTS.md
+CLAUDE.md
+docs/agent-sdk.workspace.code-workspace
+docs/llm-model-info-and-caps.md
+docs/llm-refactor.md
```
enyst:
```suggestion
```

openhands-sdk/openhands/sdk/llm/llm_registry.py
```diff
@@ -106,9 +108,101 @@ def get(self, service_id: str) -> LLM:
...
+            if profile_dir is not None
+            else _DEFAULT_PROFILE_DIR
+        )
+        directory.mkdir(parents=True, exist_ok=True)
+        return directory
+
+    def _load_profile_with_synced_id(self, path: Path, profile_id: str) -> LLM:
+        llm = LLM.load_from_json(str(path))
+        if llm.profile_id != profile_id:
+            llm = llm.model_copy(update={"profile_id": profile_id})
```
enyst:
```suggestion
            # force llm.profile_id to carry the correct name
            # (the name of the .json we loaded it from)
            llm = llm.model_copy(update={"profile_id": profile_id})
```

openhands-sdk/openhands/sdk/persistence/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+and helpers) to provide a small, stable import surface:
+
+- Encapsulation: internal module layout can change without breaking callers.
+- Discoverability: callers can find persistence settings via
+  ``openhands.sdk.persistence``.
+- Consistency: matches the SDK pattern of exposing intended entry points at the
+  package level rather than requiring deep imports.
+
+Anything exported via ``__all__`` should be treated as part of the supported SDK
+surface.
```
enyst:
```suggestion
API.
```

examples/01_standalone_sdk/31_llm_profiles.py
```diff
@@ -0,0 +1,126 @@
...
+
+def load_profile(registry: LLMRegistry, name: str) -> LLM:
+    llm = registry.load_profile(name)
+    if llm.api_key is None:
+        api_key = os.getenv("LLM_API_KEY")
+        if api_key is None:
+            raise RuntimeError(
+                "Set LLM_API_KEY to authenticate, or save the profile with "
+                "include_secrets=True."
+            )
```
xingyaoww: Interesting, i wonder if we should just include API Key along with the profile settings?

enyst: WDYT? I was thinking two things here:
- it makes sense, and it would be a single place per LLM profile
- it's plain text, like we have in `agent_settings.json` in CLI, and that's not ideal

Well and a third thing: depending on interface, this isn't terribly bad: in VScode it's vscode secrets management, which apparently uses the system keyring

enyst: Could we defer this question for later? For the moment, using `openhands/` or some similar multi-llm provider like `openrouter/` it kinda doesn't matter, it'll be the same key, right?

xingyaoww: > it's plain text, like we have in agent_settings.json in CLI, and that's not ideal

I think this will be the most useful and maybe the easiest to implement, especially if you want to use this in the CLI: when we talk about switching LLMs, it is very possible that we may want to switch across different model provider, so i think aiming for the most general implementation is good here.


And since this "llm profile" is likely a more custom/advanced feauture, end user should know what they are doing when they save LLM profiles so i think save raw API Key in text is probably ok (we can do smth like chmod 600 to reduce the risk, etc)

enyst: @openhands lets do this, no backwards compat necessary since this is a new feature. We want to save the keys in the profile json files. Chmod when saving a new file.

enyst: @xingyaoww Done!

docs/llm_profiles.md
enyst, line 1: @openhands use this file and the example/ file, and make a PR to docs/ repo in a similar way we have other examples synced with the agent-sdk examples scripts.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -287,6 +292,10 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
             "Safety settings for models that support them (like Mistral AI and Gemini)"
         ),
     )
+    profile_id: str | None = Field(
+        default=None,
+        description="Optional profile id (filename under ~/.openhands/llm-profiles).",
```
enyst:
```suggestion
        description="Optional profile id (filename under the profiles directory).",
```

enyst:
@xingyaoww I can give this PR another round of testing and maybe debugging just to see it step by step that I’m not missing something, but I’d love to know what concerns you have, to know what to focus on.

For the record, I would like to follow up with the [Switch LLM PR](https://github.com/OpenHands/software-agent-sdk/pull/1049), which is built on top of this PR. I did quite a bit of thinking in what way or direction to nudge our core architecture here. In practice, I played a lot with the switch PR, rather than this one, the other even has a TUI of its own with commands to use! 😅 So things work, generally, give or take that I might forget some detail.

_Edited to add:_
- related to this group of changes: https://github.com/OpenHands/software-agent-sdk/issues/1451

enyst:
Heads up: branch was updated to include latest upstream/main (commit 62d35bb5 as of 2025-12-29). Remaining failure is check-examples, pending OpenHands/docs#219 approval+merge.

enyst:
@openhands Understand the goal of this PR. Merge main into its branch and fix the conflicts.

Please make sure to not lose what main branch did in state!

enyst:
@openhands 2 tests are failing, an llm/ test and an mcp/ test. Figure it out please, then tell me what the problem was. Tell me in a comment on github.

enyst:
I tracked down the failing `tests/sdk/llm/test_llm_registry_profiles.py::test_load_profile_rejects_unknown_fields` (and the related MCP test failures it was cascading into).

**Root cause**: `LLM` currently has `model_config.extra = "ignore"`, so when we loaded a saved profile JSON, Pydantic silently dropped unknown keys (e.g. `unknown_field`). The new profile persistence tests expect us to *reject* unknown fields, to catch typos / stale schema during profile loading.

**Fix**: switched `LLM` to `extra="forbid"` and adjusted `LLMRegistry._load_profile_with_synced_id()` to validate via `LLM.model_validate()` after forcing `profile_id` to match the filename. This keeps the “profile id is derived from filename” behavior while ensuring unknown fields raise `ValidationError`.

Commit: b5a01ad7

.gitignore
```diff
@@ -203,7 +203,6 @@ cache
 /workspace/
 openapi.json
 .client/
-
 # Local workspace files
```
enyst, line 207:
```suggestion

# Local workspace files
```

examples/01_standalone_sdk/34_llm_profiles.py
```diff
@@ -0,0 +1,128 @@
+"""Create and use an LLM profile with :class:`LLMRegistry`.
+
+Run with::
+
+    uv run python examples/01_standalone_sdk/31_llm_profiles.py
```
enyst:
```suggestion
    uv run python examples/01_standalone_sdk/34_llm_profiles.py
```

examples/01_standalone_sdk/34_llm_profiles.py
xingyaoww, line 1: @OpenHands please sends a separate PR makeing sure that our `examples` folder doesn't contains duplicated example number - i.e., there should not be two script under the same folder that is number `34`

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -222,32 +239,52 @@ def create(
...
             # Attach event log early so we can read history for tool verification
-            state._fs = file_store
-            state._events = EventLog(file_store, dir_path=EVENTS_DIR)
+            event_log = EventLog(file_store, dir_path=EVENTS_DIR)
 
-            # Verify compatibility (agent class + tools)
-            agent.verify(state.agent, events=state._events)
+            persisted_agent = AgentBase.model_validate(
+                persisted_agent_payload,
+                context={"llm_registry": registry, INLINE_CONTEXT_KEY: False},
```
xingyaoww: I'm actually not sure if we want to embed the concept of LLMProfile so deep into the serialization layer?

It feels to me that it can just be a high-level concept - could be even higher than LLMRegistry -> it is essentially a manager over a bunch of JSONs, and can list/save/load a bunch of `LLM` instances?

So most of these could happen on the user side, e.g.,

```
llm = LLMProfileManager.load("profile name")
```

OR even simpler, we can consider these methods as `@classmethod` for the `LLM` class, e.g.,
```
@classmethod
def load_profile(cls, profile_name, profile_dir) -> LLM

@classmethod
def list_profile(cls, profile_dir) -> list[str] (LLM profile names)

def save_profile(self, profile_name, profile_dir) -> None
```

What do you think?

enyst: @openhands I love this review comment! Let's think deeply and make these classmethods of LLM.

We have an issue though: currently, if no profile exists or is mentioned, then we serialize the LLM configuration together with the conversation. I *think* maybe we can just refer a "default" profile there. WDYT?

If so, we do not need INLINE KEY, so let's remove that everywhere. Use a default profile name for a default profile created when no profile exists, and the user creates an LLM in whatever way. Let's call it `default.json` if none exists; if the user creates a new (unmatching) LLM without profile when "default" exists, let's overwrite it fully with the new one. All the rest of the behavior remains the same, so please investigate and understand the PR diff first, then fix this review comment.

Do tell us on github what you think about the behavior.

enyst:
@openhands read this PR carefully and understand its goal. Then merge main into it, keeping our behavior too, but don't mess with main either.

When you are done, commit, run checks (e.g. precommit) and push; summarize what you did in a comment to the PR.

enyst:
Closing to open a fresh PR - conversation got too long.

PR #802, opened 2025-10-19
Fix: Make adding security_analyzer optional for get_default_agent

jimwhite:
https://github.com/All-Hands-AI/agent-sdk/issues/300 made the security analyzer optional because smaller models often fail to work as LLM-risk analyzers.

This very small change adds a `add_security_analyzer` arg to `get_default_agent` which is true by default.

Also  changed the standalone SDK example scripts to use an env var `DISABLE_SECURITY_ANALYZER` to demonstrate its use.  It is in the negative form because it is false when missing as expected.  This change makes it easy to do a local devcontainer install and check that everything is working with only env config, no code changes necessary (a separate PR will have my approach to that).

### DIFF
--- a/openhands-agent-server/openhands/agent_server/api.py
+++ b/openhands-agent-server/openhands/agent_server/api.py
@@ -165,7 +165,7 @@ def _setup_static_files(app: FastAPI, config: Config) -> None:
     )
 
     # Add root redirect to static files
-    @app.get("/")
+    @app.get("/", tags=["Server Details"])
     async def root_redirect():
         """Redirect root endpoint to static files directory."""
         # Check if index.html exists in the static directory


examples/01_standalone_sdk/01_hello_world.py
```diff
@@ -18,7 +18,11 @@
     base_url=base_url,
     usage_id="agent",
 )
-agent = get_default_agent(llm=llm, cli_mode=True)
+
+add_security_analyzer = not bool(os.getenv("DISABLE_SECURITY_ANALYZER", "").strip())
+agent = get_default_agent(
+    llm=llm, cli_mode=True, add_security_analyzer=add_security_analyzer
+)
```
xingyaoww, line 27: I think we only need to add the env var for one of the example `examples/01_standalone_sdk/04_confirmation_mode_example.py` and remove this usage from all other examples (eg we want to keep each examples as simple / to the point as possible)

xingyaoww:
@OpenHands 

look at "Check OpenAPI Schema" github workflow and help me fix the issue that it didn't work on a fork

```
Run actions/checkout@v5
Syncing repository: All-Hands-AI/agent-sdk
Getting Git version info
Temporarily overriding HOME='/home/runner/_work/_temp/61879fe0-c039-4df9-8cf9-3402886c5aa7' before making global git config changes
Adding repository directory to the temporary git global config as a safe directory
/usr/bin/git config --global --add safe.directory /home/runner/_work/agent-sdk/agent-sdk
Deleting the contents of '/home/runner/_work/agent-sdk/agent-sdk'
Initializing the repository
Disabling automatic garbage collection
Setting up auth
Fetching the repository
Determining the checkout info
  /usr/bin/git branch --list --remote origin/disable-llm-analyzer
  /usr/bin/git tag --list disable-llm-analyzer
  Error: A branch or tag with the name 'disable-llm-analyzer' could not be found

```

examples/01_standalone_sdk/01_hello_world.py
jimwhite: I thought the same initially but then I went to try the other examples and for those of us using local models (even qwen-3-next-80b-instruct) the security analyzer needs to be turned off.

My thought on reducing the code in the examples is to have some init code that handles all the env cars.  One of the things this PR shows is that usually only LLM_API_KEY is used but in one case there is a second check for OPENAI_API_KEY.

xingyaoww: @malhotra5 how do you feel out setting the security analyzer to be false by default?

xingyaoww: image

ok let's set add security analyzer to false by default, and just maybe add this to only example 04?

xingyaoww: @OpenHansd please start a new PR that set add_security_analyzer=false to default for get_default_agent, and revert all changes to examples/**/*.py by https://github.com/OpenHands/agent-sdk/pull/802 except for examples/01_standalone_sdk/04_confirmation_mode_example.py

xingyaoww: @OpenHands please start a new PR that set add_security_analyzer=false to default for get_default_agent, and revert all changes to examples/**/*.py by https://github.com/OpenHands/agent-sdk/pull/802 except for examples/01_standalone_sdk/04_confirmation_mode_example.py

PR #832, opened 2025-10-20
Add `FunctionCallValidationError` handling & `claude-haiku-4-5` to function calling supported models

ryanhoangt:
This PR is to:
- Handle `FunctionCallValidationError` when converting non-fn-calling to fn calling messages. This is before we even have an `Action` so we can't use `AgentErrorEvent` I think.
- Add `claude-haiku-4-5` to function calling supported models

This brings back integration tests result to 7/7 (100%).

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:9a5a896-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-9a5a896-python \
  ghcr.io/openhands/agent-server:9a5a896-python
```

```
ghcr.io/openhands/agent-server:9a5a896-golang
ghcr.io/openhands/agent-server:9a5a896-java
ghcr.io/openhands/agent-server:9a5a896-python
```

_The `9a5a896` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/agent.py
+++ b/openhands-sdk/openhands/sdk/agent/agent.py
@@ -191,6 +191,17 @@ class Agent(AgentBase):
                     extra_body={"metadata": self.llm.metadata},
                     add_security_risk_prediction=self._add_security_risk_prediction,
                 )
+        except FunctionCallValidationError as e:
+            logger.warning(f"LLM generated malformed function call: {e}")
+            error_message = MessageEvent(
+                source="user",
+                llm_message=Message(
+                    role="user",
+                    content=[TextContent(text=str(e))],
+                ),
+            )
+            on_event(error_message)
+            return
         except Exception as e:
             # If there is a condenser registered and the exception is a context window
             # exceeded, we can recover by triggering a condensation request.
@@ -204,21 +215,6 @@ class Agent(AgentBase):
                 )
                 on_event(CondensationRequest())
                 return
-
-            # If the LLM generated a malformed function call, send the error back
-            # as a message so it can correct the tool call
-            elif isinstance(e, FunctionCallValidationError):
-                logger.warning(f"LLM generated malformed function call: {e}")
-                error_message = MessageEvent(
-                    source="user",
-                    llm_message=Message(
-                        role="user",
-                        content=[TextContent(text=str(e))],
-                    ),
-                )
-                on_event(error_message)
-                return
-
             # If the error isn't recoverable, keep propagating it up the stack.
             else:
                 raise e


neubig (approved):

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -204,6 +205,20 @@ def step(
                 on_event(CondensationRequest())
                 return
 
+            # If the LLM generated a malformed function call, send the error back
+            # as a message so it can correct the tool call
+            elif isinstance(e, FunctionCallValidationError):
```
enyst: Nit: maybe we could add it as `except FunctionCallValidationError` just before `except Exception`? just for code look and feel, so to speak 😅 

For the condenser we had to do this, it had more conditions, but this is just if / isinstance

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -204,6 +205,20 @@ def step(
...
+                logger.warning(f"LLM generated malformed function call: {e}")
+                error_message = MessageEvent(
```
enyst: Are we sure we can't do an almost-normal `action + observation`?

ryanhoangt: WDYM by "an almost-normal action"? This validation is based on the tool schema so I guess we need to conform strictly?

enyst: I mean like [this](https://openhands-ai.slack.com/archives/C086ARSNMGA/p1760995922093169?thread_ts=1760694927.682999&cid=C086ARSNMGA): when we can't make a valid Action from the LLM response, then we can set `action=None`.

And that means we still can create an ActionEvent, and we create an ObservationEvent for it - AgentErrorEvent is an ObservationEvent iirc.

Is that impossible to do here?

enyst: If we can do it, it may be better IMHO:
- we would send back to the LLM its own blunder, too, not only our fake user message
- if there are multiple actions, we could still have our normal behavior, that is, a batch of actions + obs that we send back

If there's some reason why we can't, it's okay IMO to settle for a MessageEvent.

You may want to see also the previous discussion on these LLM blunders: [slack](https://openhands-ai.slack.com/archives/C09FJAZSP9N/p1759440759543539)

ryanhoangt: > when we can't make a valid Action from the LLM response, then we can set action=None. And that means we still can create an ActionEvent

Looking into the code, I think we can't even create an `ActionEvent` in this case, because of other required fields like `tool_call`, `tool_name`. The `FunctionCallValidationError` is raised very early, when we perform validation to convert the raw response from LLM into tool calling format (a tool call dict in the `tool_calls` field of message list). Afterwards, we convert it to LiteLLM message and then our `Message` object, and finally an `ActionEvent`. Fields like `tool_call`, `tool_name` are populated by the conversion from LiteLLM message to our `Message` object, which is way after the exception is raised.

> we would send back to the LLM its own blunder, too, not only our fake user message

Just to note, this exception only happens for non-native function calling models, in those cases we receive assistant chat message, not tool call directly from the API so I think sending user message is still reasonable? 🤔

xingyaoww: > , in those cases we receive assistant chat message,

I think this make sense, ideally we should write a test to make sure the assistant chat message is stored as some event in the system, otherwise we won't be able to send [...prefix, problematic action, error message] to LLM for next iteration.

enyst: Maybe we can include in the exception
- action as received
- our error message?

xingyaoww: sounds good -- @OpenHands please create a new issue on OpenHands/agent-sdk repo, we should handle the `FunctionCallValidationError` properly by (1) add test to make sure the malform action is included in one event and send to llm in the next iteration, (2) make sure error message is sent

PR #847, opened 2025-10-21
Env template generator

tofarr:
This PR introduces the ability for the Env Parser to generate .env files based on the config classes - this is useful for guiding devs in the available options. This will be reused in the OpenHands project where the configuration is substantially more complex.

Currently the `build_config_template.py` script produces:

```
# List of valid session API keys used to authenticate incoming requests. Empty list implies the server will be unsecured. Any key in this list will be accepted for authentication.
# OH_SESSION_API_KEYS_0=

# Set of CORS origins permitted by this server (Anything from localhost is always accepted regardless of what's in here).
# OH_ALLOW_CORS_ORIGINS_0=

# The location of the directory where conversations and events are stored.
OH_CONVERSATIONS_PATH=workspace/conversations

# The location of the directory where bash events are stored as files. Defaults to 'workspace/bash_events'.
OH_BASH_EVENTS_DIR=workspace/bash_events

# The location of the directory containing static files to serve. If specified and the directory exists, static files will be served at the /static/ endpoint.
# PosixPath
# OH_STATIC_FILES_PATH=.
# 
# NoneType
OH_STATIC_FILES_PATH_IS_NONE=1


# Webhooks to invoke in response to events

# Whether to enable VSCode server functionality
OH_ENABLE_VSCODE=1

# Port on which VSCode server should run
OH_VSCODE_PORT=8001

# Whether to enable VNC desktop functionality
OH_ENABLE_VNC=0

```

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:effc723-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-effc723-python \
  ghcr.io/openhands/agent-server:effc723-python
```

```
ghcr.io/openhands/agent-server:effc723-golang
ghcr.io/openhands/agent-server:effc723-java
ghcr.io/openhands/agent-server:effc723-python
```

_The `effc723` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._


tests/agent_server/test_env_parser.py
```diff
@@ -156,11 +164,11 @@ def test_none_env_parser(clean_env):
...
 
     # Test missing key (should return None)
-    assert parser.from_env("MISSING_KEY") is None
+    assert parser.from_env("SOME_VALUE") is MISSING
 
     # Test present key (should return MISSING)
-    os.environ["TEST_NONE"] = "any_value"
-    assert parser.from_env("TEST_NONE") is MISSING
+    os.environ["SOME_VALUE_IS_NONE"] = "1"
+    assert parser.from_env("SOME_VALUE") is None
```
enyst, line 171: This doesn't look right 🤔  
- first it asserts SOME_VALUE is Missing (and its comment says it's None)
- then it asserts it's None (and its comment says Missing, and it's the same var)?

MISSING should be supported, right?

tofarr: So this is an instance of what I like to call "the null vs undefined problem". Basically `None != undefined` For example:
```
class Foo(BaseModel):
  bar: int | None = 10
  
Foo.model_validate({"bar": 5})  # Foo(bar=5)
Foo.model_validate({})  # Foo(bar=10)
Foo.model_validate({bar: None})  # Foo(bar=None)
```

Now - expanding this to environment variables where the only input type is strings - how do I describe `null`?

Before this change, I was treating missing values as null, which is incorrect (In the Foo example, missing values == 10 which is != null!) 

So I chose to encode this as additional data in the key - in this case, you could define:

```
# No definition (MISSING) means that bar=10
# Or define bar explicitly
BAR=5
# Or define bar explicitly as None
BAR_IS_NONE=true
```

I suppose there could be collisions if somebody introduced a variable named `bar_is_none`, but I feel this is unlikely.

enyst: Ugh, yeah, quirky.

So just now, we were looking at LLM_API_KEY which could be empty or None. The conclusion is it could be empty, while None seems frankly of no interest.

Does it matter? Maybe not for the api key, but it might matter, I think, if we establish a way things work (tm), and apply it for some env vars and not others. That will be confusing. 🤔 

Do we have a case where `BAR_IS_NONE` is necessary? What if it's empty, can we accept its default value in `bar` case, `bar=10`?

Second tiny issue: either way, we should fix the tests though, their comments are misleading now. Maybe we can send OH on this discussion to verify both behavior and comments and fix the mismatches?

tofarr: What if it's empty, can we accept its default value in bar case, bar=10?

* If missing (undefined), then it uses the default value (10)
* If an empty string, then that is assumed to be the value (Which will cause an error in the Foo case because an empty string is not a valid integer)

tofarr: @OpenHands - Please check that the comments in the code match the functionality for the changes in this PR

enyst: Let me take a step back: if we establish "_IS_NONE" as a convention for recognizing None from env, and MISSING as a real value, then we should try to:
- document it
- apply _anywhere_ we read from env.

We can do that, or is there's a barrier or a need to do differently? 

The env parser is defined in the `agent-server`, it doesn't apply to all configs, right, and we read a lot of stuff from env. I wonder why doesn't it apply to all reads from env, or does it?

---

_Edited to add:_ Sorry, I'm not trying to be a nitpicker. I'm trying to understand quickly all implications here. I think we shouldn't create rules for some configs/env vars and not for others, unless we know very well why and where. 

(I'd personally prefer, on a side, to be as lenient as possible (e.g. empty int -> let it be default unless it really hurts), but that's not important.)

Please note we've been down this road before, and that's my concern: making configs play by ad-hoc rules inconsistently, until nobody can track them all down eventually.

enyst: The case I mentioned above:
- https://github.com/OpenHands/agent-sdk/issues/850

PR #855, opened 2025-10-21
Fixes for Secret Storage

tofarr:
Serialization to meta.json was not storing secrets correctly - now it is.

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:bb7ff89-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-bb7ff89-python \
  ghcr.io/openhands/agent-server:bb7ff89-python
```

```
ghcr.io/openhands/agent-server:bb7ff89-golang
ghcr.io/openhands/agent-server:bb7ff89-java
ghcr.io/openhands/agent-server:bb7ff89-python
```

_The `bb7ff89` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/.github/scripts/check_documented_examples.py
+++ b/.github/scripts/check_documented_examples.py
@@ -1,0 +1,176 @@
+#!/usr/bin/env python3
+"""
+Check if all examples in agent-sdk are documented in the docs repository.
+
+This script:
+1. Scans the docs repository for references to example files
+2. Lists all example Python files in the agent-sdk repository
+3. Compares the two sets to find undocumented examples
+4. Exits with error code 1 if undocumented examples are found
+"""
+
+import os
+import re
+import sys
+from pathlib import Path
+
+
+def find_documented_examples(docs_path: Path) -> set[str]:
+    """
+    Find all example file references in the docs repository.
+
+    Searches for patterns like:
+    - examples/01_standalone_sdk/02_custom_tools.py
+    in MDX files.
+
+    Returns:
+        Set of normalized example file paths (relative to agent-sdk root)
+    """
+    documented_examples: set[str] = set()
+
+    # Pattern to match example file references
+    # Matches: examples/[folder1]/[folder2]/[file].py or
+    # examples/[folder1]/[file].py
+    pattern = r"examples/[\w_]+(?:/[\w_]+)?/[\w_]+\.py"
+
+    for root, _, files in os.walk(docs_path):
+        for file in files:
+            if file.endswith(".mdx") or file.endswith(".md"):
+                file_path = Path(root) / file
+                try:
+                    content = file_path.read_text(encoding="utf-8")
+                    matches = re.findall(pattern, content)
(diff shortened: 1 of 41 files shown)


openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -133,7 +133,9 @@ def _save_base_state(self, fs: FileStore) -> None:
         """
         Persist base state snapshot (no events; events are file-backed).
         """
-        payload = self.model_dump_json(exclude_none=True)
+        payload = self.model_dump_json(
+            exclude_none=True, context={"expose_secrets": True}
+        )
```
enyst: Why are we storing secrets on disk?

This file is in every conversation afaik, so that's a lot of places if I see this right.

tofarr: One use case is when you pause / resume docker containers / K8 sandboxes.

tofarr: (Or even just the regular non sandboxed agent server - storing on disk allows you to resume the conversation after a restart)

enyst: When we resume, could we send it secrets?

enyst: For the fully local case, we are assuming right now that we need to read secrets from env / pass as arguments in python.

Which is kinda possible, kinda annoying at times.

I actually do think we could store on disk, but the problem with dumping base .json is that it's doing it repeatedly everywhere, in all conversations, and I don't think the user expects that. It's very insecure.

One thing that may help is:
- https://github.com/OpenHands/agent-sdk/pull/795

We can then store, ok, but once. It's in `~/.openhands/llm-profiles` by default, and that's not unusual. It's not awesome, but I do think it's expected.

tofarr: I could revert the base.json for now - this doesn't include everything required in order to restart a conversation anyway - the agent server uses the meta.json to save / restore state and it is only written when there is a graceful shutdown.

enyst: Meta is also per conversation, so it keeps multiple copies. But at least it's only for users using the server. 🤔 

Could you please help me see why we need to eliminate the other alternative, why can't we act at restore time in a similar way we do at start, when we do send over the wire the secrets? What am I missing?

enyst: In any case, meta.json should work for now, since I'm sure it's urgent, and at least I think it's better than another alternative (the 4th?): some strange "if not local" code. 😓 

I do think we will need to give this some thought...

tofarr:
I am closing this PR in favor of https://github.com/OpenHands/agent-sdk/pull/893

PR #863, opened 2025-10-22
Feat git operations

tofarr:
## Git Operations Enhancement

This PR introduces comprehensive git operations functionality with proper type safety and extensive test coverage.

### Key Changes
- **Added comprehensive tests**: Created 25+ unit tests for `git_changes.py` and `git_diff.py` using real git operations
- **Enhanced test coverage**: Tests cover empty repos, new/modified/deleted files, nested directories, staged/unstaged changes, gitignore, binary files, special characters, and error conditions
- **Improved type safety**: All git operations use proper Pydantic models
- **Real git integration**: Tests use temporary directories and actual bash commands for authentic git behavior

All tests pass and follow project coding standards. Ready for review.

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:f331432-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-f331432-python \
  ghcr.io/openhands/agent-server:f331432-python
```

```
ghcr.io/openhands/agent-server:f331432-golang
ghcr.io/openhands/agent-server:f331432-java
ghcr.io/openhands/agent-server:f331432-python
```

_The `f331432` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/.github/workflows/assign-reviews.yml
+++ b/.github/workflows/assign-reviews.yml
@@ -154,7 +154,7 @@ jobs:
                   fi
 
             - name: Upload logs as artifact
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               if: always()
               with:
                   name: openhands-task-logs
--- a/.github/workflows/check-documented-examples.yml
+++ b/.github/workflows/check-documented-examples.yml
@@ -21,12 +21,12 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: Checkout agent-sdk repository
-              uses: actions/checkout@v4
+              uses: actions/checkout@v5
               with:
                   fetch-depth: 0
 
             - name: Checkout docs repository
-              uses: actions/checkout@v4
+              uses: actions/checkout@v5
               with:
                   repository: OpenHands/docs
                   path: docs
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -107,7 +107,7 @@ jobs:
                   tar -czvf ../../../integration_tests_${{ matrix.job-config.run-suffix }}_${TIMESTAMP}.tar.gz *${{ matrix.job-config.run-suffix }}* # Include result directories for this model
 
             - name: Upload evaluation results as artifact
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               id: upload_results_artifact
               with:
                   name: integration-test-outputs-${{ matrix.job-config.run-suffix }}-${{ github.run_id }}-${{ github.run_attempt }}
@@ -128,7 +128,7 @@ jobs:
                   fi
 
             - name: Upload test results summary
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: test-results-${{ matrix.job-config.run-suffix }}
                   path: test_results_summary/${{ matrix.job-config.run-suffix }}_results.json
@@ -156,14 +156,14 @@ jobs:
                   uv sync --dev
 
             - name: Download all test results
-              uses: actions/download-artifact@v5
+              uses: actions/download-artifact@v6
               with:
                   pattern: test-results-*
                   merge-multiple: true
                   path: all_results
 
             - name: Download all integration test artifacts
-              uses: actions/download-artifact@v5
+              uses: actions/download-artifact@v6
               with:
                   pattern: integration-test-outputs-*
                   path: artifacts
@@ -191,7 +191,7 @@ jobs:
                     --output-file consolidated_report.md
 
             - name: Upload consolidated report
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: consolidated-report
                   path: consolidated_report.md
--- a/.github/workflows/pr-review-by-openhands.yml
+++ b/.github/workflows/pr-review-by-openhands.yml
@@ -92,7 +92,7 @@ jobs:
                   uv run python ../agent-sdk/examples/03_github_workflows/02_pr_review/agent_script.py
 
             - name: Upload logs as artifact
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               if: always()
               with:
                   name: openhands-pr-review-logs
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -106,7 +106,7 @@ jobs:
                   echo "✓ Binary test completed successfully"
 
             - name: Upload binary artifact
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: openhands-server-${{ matrix.os }}
                   path: |
@@ -233,7 +233,7 @@ jobs:
                   EOF
 
             - name: Upload build info artifact
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: build-info-${{ matrix.name }}
                   path: build-info/${{ matrix.name }}.json
@@ -250,7 +250,7 @@ jobs:
             build_summary: ${{ steps.consolidate.outputs.build_summary }}
         steps:
             - name: Download all build info artifacts
-              uses: actions/download-artifact@v5
+              uses: actions/download-artifact@v6
               with:
                   pattern: build-info-*
                   merge-multiple: true
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -72,7 +72,7 @@ jobs:
 
             - name: Upload sdk coverage
               if: steps.changed.outputs.any_changed == 'true' && always()
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: coverage-sdk
                   path: coverage-sdk.dat
@@ -126,7 +126,7 @@ jobs:
 
             - name: Upload tools coverage
               if: steps.changed.outputs.any_changed == 'true' && always()
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: coverage-tools
                   path: coverage-tools.dat
@@ -180,7 +180,7 @@ jobs:
 
             - name: Upload Agent Server coverage
               if: steps.changed.outputs.any_changed == 'true' && always()
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: coverage-agent-server
                   path: coverage-agent-server.dat
@@ -236,7 +236,7 @@ jobs:
 
             - name: Upload cross coverage
               if: steps.changed.outputs.any_changed == 'true' && always()
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v5
               with:
                   name: coverage-cross
                   path: coverage-cross.dat
@@ -259,7 +259,7 @@ jobs:
               run: uv sync --frozen --group dev
 
             - name: Download coverage artifacts
-              uses: actions/download-artifact@v5
+              uses: actions/download-artifact@v6
               with:
                   path: ./cov
               continue-on-error: true


openhands-agent-server/openhands/agent_server/__main__.py
```diff
@@ -43,7 +43,7 @@ def main():
         host=args.host,
         port=args.port,
         reload=args.reload,
-        reload_excludes=["workspace"],
+        reload_includes=["openhands-agent-server", "openhands-sdk", "openhands-tools"],
         log_level=log_level,
```
tofarr, line 47: reload_excludes is flaky - When testing locally I got sick of the server reloading every time I had the Agent change a file in my workspace directory.

enyst: I hear you, I never got it to work right on v0 🤷

openhands-sdk/openhands/sdk/git/git_changes.py
```diff
@@ -0,0 +1,231 @@
...
+            result = run(f"git --no-pager rev-parse --verify {ref}", repo_dir)
+            return result
+        except RuntimeError:
+            # invalid ref - try next
+            continue
+
+    return None
+
+
+def get_changes_in_repo(repo_dir: str) -> list[GitChange]:
```
rbren: I think these should be methods on Workspace...

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -49,7 +49,9 @@ async def load_meta(self):
     async def save_meta(self):
         self.stored.updated_at = utc_now()
         meta_file = self.conversation_dir / "meta.json"
-        meta_file.write_text(self.stored.model_dump_json())
+        meta_file.write_text(
+            self.stored.model_dump_json(context={"expose_secrets": True})
```
enyst: What can happen if we don't save keys/secrets right now?

Apparently we can't restore a conversation, but I see that `load_meta` is [never called](https://github.com/search?q=repo%3AOpenHands%2Fagent-sdk%20load_meta&type=code). We can't restore even if we save?

Xingyao is working on e.g. https://github.com/OpenHands/agent-sdk/pull/476 and trying other approaches as well. It takes secrets out of conversation, and we'll have a better alternative.

Could we leave this False?

tofarr: Done.

enyst:

When you send your feedback in a github comment, use bold, headings maybe, or other markdown; you can even use emoji a bit, for readability, e.g. green checkmark, red cross, etc.

Then answer me two more questions:

1. This PR defines a new working_dir for the agent-server. But the agent-server works with a Workspace instance, which has a working_dir that can be set, too. Are they always in sync? When or how could they be desynced? Feel free to understand well every related execution path, and in particular when running agent-server programmatically. Read the readme of the agent-sdk and a few examples in examples/, so you see how it's run.

2. In particular, read example examples/02_remote_agent_server/01_convo_with_local_agent_server.py
Does it work after the PR changes? Notice it has a workspace with a working_dir and the server doesn't.

enyst:
**🧪 /codereview-roasted for PR #863 – Feat git operations**

🧭 **Taste Rating: 🔴 Needs improvement**

> Key insight: This PR creates two sources of truth for the working directory (server-level `Config.working_dir` vs. per-workspace `working_dir`). They are not synchronized. Several routers capture a default config at import time, so programmatic configuration via `create_app(...)` won’t propagate. Git endpoints operate relative to the server config rather than the client/workspace, which breaks the mental model.

## [CRITICAL ISSUES]

1) Config captured at import-time (bypasses create_app config)
- File: `openhands-agent-server/openhands/agent_server/git_router.py` (module top)
- Issue: `config = get_default_config()` and `workspace = LocalWorkspace(...)` are evaluated at import time. Any `Config` passed to `create_app(config)` is ignored by this router.
- Impact: Programmatic users get inconsistent behavior across routers (auth/CORS reflect the app’s config, but git routes use env defaults).
- Fix: Build routers via factories (e.g., `def create_git_router(config: Config) -> APIRouter`) and wire them in `_add_api_routes()`. Avoid module-level singletons in routers.

2) Asserts used for path validation in API
- File: `git_router.py`: `assert (config.working_dir / path).resolve().is_relative_to(...)`
- Issue: `assert` can be stripped with `-O` and yields 500s instead of 4xx.
- Fix: Replace with explicit checks and raise `HTTPException(400/403)` with clear messages. Add tests for path traversal attempts (`..` outside working_dir).

3) Divergent `working_dir` semantics: Local vs Remote
- LocalWorkspace: defaults `cwd` to `self.working_dir` when not provided.
- RemoteWorkspace: doesn’t send `cwd` if `None`; server then runs in the process CWD (not the workspace).
- Impact: `RemoteWorkspace.execute_command("pwd")` runs in server process CWD, not in `workspace.working_dir`. Surprising and inconsistent with local behavior.
- Fix: When `cwd is None`, send `cwd=self.working_dir` from RemoteWorkspace; or default `ExecuteBashRequest.cwd` to `Config.working_dir` server-side.

4) Two competing “working_dir” authorities with no sync
- Server-level: `Config.working_dir` (default `workspace/project`).
- Per-conversation: `EventService` uses `stored.workspace.working_dir` for the LocalConversation.
- Git endpoints: use `Config.working_dir` for scoping and `LocalWorkspace` instantiation.
- Result: A client may start a conversation with working_dir=X while git endpoints operate on Y. No coordination and confusing UX.
- Fix: Decide and enforce scope:
  - Conversation-scoped: make git endpoints under `/conversations/{id}/git/...` and resolve `working_dir` from that conversation; or
  - Server-scoped: make it explicit in API/docs (and don’t surface as `Workspace` instance methods implying conversation scope).

5) Error mapping for git endpoints
- Issue: Exceptions bubble up as generic 500s; specificity is lost.
- Fix: Catch `Git*` exceptions and map to 400/404 with actionable messages; reserve 500 for unexpected errors.

- Unify configuration flow
  - Build routers via factories consuming `Config` and services.
  - Remove import-time `get_default_config()` from routers; use DI/dependencies so `create_app(Config(...))` truly governs behavior.

- Align workspace behavior
  - Ensure RemoteWorkspace/AsyncRemoteWorkspace default to `cwd=self.working_dir` when none provided.
  - Optionally have server-side bash default to `Config.working_dir` if `cwd` is not sent.

- Git API design clarity
  - Pick a single model (server-scoped vs conversation-scoped) and make it explicit. Right now it’s server-scoped but exposed through `Workspace`, which suggests conversation scope.

- Tests to add
  - Path traversal negative tests for git endpoints (expect 400/403, not `AssertionError`).
  - Programmatic app creation with non-default `working_dir`; verify git endpoints honor it.
  - Parity: local vs remote `execute_command("pwd")` should match for same `working_dir`.

## [STYLE NOTES]

- ✅ Git utils look solid (no `shell=True`, sane timeouts, logging). For renames in `get_changes_in_repo`, consider returning `MOVED` or document why you model as delete+add.

## ✅ Verdict

❌ Needs rework before merge. Utilities and tests are strong, but the server DI/config story creates user-facing inconsistencies. Make Config truly app-scoped, align remote/local workspace semantics, and choose a coherent scope for git endpoints.

If helpful, I can prepare a follow-up patch to:
- convert routers to factories with injected `Config`,
- default RemoteWorkspace to send `cwd=self.working_dir`,
- add the missing tests above.

enyst:
**Follow-up answers**

1) Is the agent-server tied to a single BaseWorkspace instance?

- Short answer: functionally yes for each conversation, but not globally one-per-server.
  - The server doesn’t hold a single global BaseWorkspace. Instead, each conversation carries its own LocalWorkspace (stored.workspace.working_dir). EventService creates a LocalConversation using that per-conversation workspace. So the server can host multiple concurrent conversations with different workspaces.
  - However, the new git endpoints (/api/git/...) currently bind to a single server-level working_dir (Config.working_dir) by constructing a LocalWorkspace at import time. So git endpoints are effectively tied to one workspace, but conversations are not. This asymmetry underpins several of my review concerns.

- Impact on earlier assessment: it reinforces the conclusion that there are two authorities for working_dir. Conversations use their own workspace; git endpoints use the server’s configured working_dir. They are not automatically synchronized, which can confuse users and API consumers.

2) Why Config.working_dir? Could we avoid another working_dir in server config?

- Your rationale makes sense: when the agent-server runs in Docker, you want Docker/env to set the working dir so remote clients (HTTP/WS) get a stable base without needing to guess the container’s filesystem layout.
- Alternatives that avoid adding another working_dir in server config:
  - Make bash/git endpoints default to the conversation’s workspace by requiring a conversation context (e.g., /api/conversations/{id}/git/...). Then Docker can still control the actual directory by mounting/setting it at container start, and the server doesn’t need a separate Config.working_dir knob.
  - If you prefer global server-scoped git/file operations, you can rely on the process CWD plus container mounts and avoid a separate field; the entrypoint can chdir to the desired path before launching uvicorn. But this couples behavior to process state, which is less explicit and harder to reason about.
  - Hybrid: keep Config.working_dir but don’t freeze it at import time. Inject Config into routers (factory/dependency), and make bash default cwd to Config.working_dir when none is provided. This preserves your Docker use case, keeps behavior explicit, and reduces surprises.

- Recommendation: If we want to minimize concepts, conversation-scoped is simplest—one source of truth (workspace.working_dir), and clients already send it. If server-scoped behavior is needed for admin/maintenance endpoints, keep Config.working_dir but make that scoping explicit in URL shape and docs.

3) What about the other examples in examples/02_remote_agent_server?

- 02_convo_with_docker_sandboxed_server.py and 03_browser_use_with_docker_sandboxed_server.py use DockerWorkspace (remote). They will work post-PR because:
  - Conversation creation sends a LocalWorkspace(working_dir=workspace.working_dir) to the server; EventService uses it. So the agent runs in that per-conversation workspace.
  - RemoteWorkspace.execute_command without cwd still runs in the server process CWD, as before; not changed by this PR.
  - If they start using /api/git endpoints, those operate relative to Config.working_dir (server-scoped), not conversation workspace. That mismatch remains the main caveat.

- 04_convo_with_api_sandboxed_server.py uses APIRemoteWorkspace that spins a server in a sandbox. Same behavior profile: conversation-scoped workspace is honored; git endpoints (if invoked) are server-scoped.

- In 04_vscode_with_docker_sandboxed_server.py, the example fetches /api/vscode/url with workspace_dir param; that uses whatever the server returns and is unaffected by Config.working_dir other than for defaults; the conversation still runs in its own workspace.

—
If you want, I can prepare a small follow-up PR to:
- Convert routers to factories with injected Config (no import-time get_default_config()),
- Default RemoteWorkspace to send cwd=self.working_dir when cwd is None (align with local),
- Add a conversation-scoped variant for git endpoints, and preserve the server-scoped endpoints behind an explicit path.

enyst:
**Answers to your three questions (re PR #863)**

---

### 1) Is the agent-server tied to a single workspace instance?

Short answer: No, not globally. The server supports multiple per-conversation workspaces, but some routes currently behave as if there’s a single server-scoped workspace.

- **Conversation engine:** Each conversation carries its own `workspace: LocalWorkspace` (the client sends it in `StartConversationRequest`). `EventService` instantiates a `LocalConversation` with that per-conversation `working_dir`. Multiple conversations can run concurrently with different workspaces.
- **Server-scoped routes:** The new git endpoints create a module-level `LocalWorkspace(working_dir=str(Config.working_dir))` at import-time and operate relative to it. File routes accept absolute paths without conversation scoping. Bash routes run in `cwd` if provided; otherwise they inherit the server process CWD.
- **Impact on earlier assessment:** it reinforces the conclusion that there are two authorities for working_dir. Conversations use their own workspace; git endpoints use the server’s configured working_dir. They are not automatically synchronized, which can confuse users and API consumers.

**Suggestions:** Either make the git/file flows conversation-scoped (recommended), or intentionally keep them server-scoped but document and inject the app’s `Config` correctly (no import-time singletons), and make the client SDK reflect that server-scoped nature.

---

### 2) Why `Config.working_dir`? Can we simplify or avoid another working_dir?

Your intuition about Docker makes sense: a container wants to expose a stable “base” directory so HTTP consumers can operate relative to it. Today, however:

- Conversation flows already allow the client to choose `workspace.working_dir` per conversation. This is independent of `Config.working_dir`.
- `Config.working_dir` is currently used by the git router as a server-side base, not as a default for conversations or bash cwd.

Ways to simplify and avoid confusion:

- **Option A (keep, but clarify):** Keep `Config.working_dir` as a server-scoped default, but
  - rename to `default_working_dir` to clarify intent,
  - inject it via the app’s `Config` (no import-time capture),
  - use it only for: (1) default bash `cwd` when none is provided; (2) default conversation workspace when the client omits one; (3) genuinely server-scoped features (e.g., VSCode, optional “global” git endpoints).
**- Option B (go conversation-scoped):** Remove server-level coupling for git; move git endpoints under `/conversations/{id}/git/...` and resolve `working_dir` from that conversation. The Docker runtime can still set a default for conversations (via env) without adding a second authority.
**- Option C (hybrid):** Keep the server default as a fallback only; prefer the conversation workspace everywhere. If you retain server-scoped git, document it clearly as operating on the server base, not the conversation.

My recommendation: Either B (cleanest mental model) or A with a rename + strict usage to make the “default only” nature explicit.

---

### 3) Do the other examples in `examples/02_remote_agent_server/` work?

Yes, they continue to work as before:

- `01_convo_with_local_agent_server.py`: Already verified earlier — works. Remote `execute_command("pwd")` runs in the server process CWD unless `cwd` is provided; conversations run in the per-conversation workspace.
- `02_convo_with_docker_sandboxed_server.py`: Uses `DockerWorkspace` to launch an agent-server. Behavior unchanged; conversation workspace is per-conversation, remote command default CWD is the server process CWD.
- `03_browser_use_with_docker_sandboxed_server.py`: Same as above, plus extra ports for browser/VNC — unaffected by the new git routes/config.
- `04_convo_with_api_sandboxed_server.py`: Uses `APIRemoteWorkspace` — unchanged by this PR.
- `04_vscode_with_docker_sandboxed_server.py`: Fetches `/api/vscode/url` and passes `workspace_dir` explicitly; this route doesn’t depend on `Config.working_dir`, so it remains fine.

Note: If any example starts using the new `/api/git/*` endpoints, those operate relative to the server’s configured working dir (server-scoped), not the conversation’s workspace (conversation-scoped). That’s the mismatch to resolve per 1) and 2).

---

If you want, I can draft a follow-up patch that:
- moves git endpoints under conversations (or makes router factories that cleanly inject `Config`),
- makes RemoteWorkspace send `cwd=self.working_dir` when `cwd` is None (parity with LocalWorkspace),
- and adds tests for path traversal, programmatic config injection, and local/remote cwd parity.

tofarr:
With the code review recommendaiton, I made a number of changes:

* I updated `event_router.py` to use the same DI pattern as `conversation_router.py`
* EventService now asserts that the Workspace it uses is local rather than casting.
* The git router no longer has the concept of a workspace (Just like file_router.py and bash_router.py - this **serves to keep the workspace concept on a lower layer than conversations** [Conversations reference workspaces, but workspaces do NOT reference conversations])

openhands-agent-server/openhands/agent_server/config.py
```diff
@@ -116,6 +116,10 @@ class Config(BaseModel):
         default=False,
         description="Whether to enable VNC desktop functionality",
     )
+    working_dir: Path = Field(
+        default=Path("workspace/project"),
+        description="The working directory for the agent",
```
enyst: With the refactoring, do we still need this? It seem unused

(I'm pretty sure it will be back later in some form, workspace is actually hard... but at least we'd see exactly where the need is)

tofarr: Reverted.

tofarr:
I"ve pulled some of the changes out of this PR into https://github.com/OpenHands/agent-sdk/pull/885 for the sake of simplicity.

PR #893, opened 2025-10-24
Feat encrypted secrets

tofarr:
This PR implements encrypted secret storage for the OpenHands Agent SDK to prevent accidental disclosure of sensitive information in conversation trajectories.

## Key Changes:
- **New Cipher utility**: Added `openhands.sdk.utils.cipher.Cipher` class using Fernet encryption with SHA256-derived keys
- **LLM model encryption**: Enhanced `LLM` class to automatically encrypt/decrypt `api_key`, `aws_access_key_id`, and `aws_secret_access_key` fields during serialization
- **Secrets Manager Encryption**: Enhanced `SecretsManager` class to automatically serialize secret values with automatic encryption / redaction
- **Agent server integration**: Updated conversation service and event handling to use cipher context for secure secret storage
- **Configuration updates**: Added `secret_key` field to server config with automatic key generation fallback
- **Test cleanup**: Removed obsolete test files that were no longer relevant to the current implementation

The encryption is designed to prevent accidental secret disclosure when sharing conversation data, not to protect against attackers with full filesystem access.

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:a91f476-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-a91f476-python \
  ghcr.io/openhands/agent-server:a91f476-python
```

```
ghcr.io/openhands/agent-server:a91f476-golang
ghcr.io/openhands/agent-server:a91f476-java
ghcr.io/openhands/agent-server:a91f476-python
```

_The `a91f476` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._


tofarr:
@OpenHands Do a /codereview-roasted on this PR.

tofarr:
@OpenHands I requested a codereview-roasted style review on this PR, but I think a number of the issues in the result are incorrect or not valid - nonetheless I think there is room for improvement. My comments are below:

1. Are you sure that cryptography is not installed as a dependency? It may be added as a transitive dependency by uv. It appears in my virtual envirnoment when I do a `uv pip freeze`

2. The objective of this PR is to prevent accidental leakage of keys in the common case where serialized conversations are downloaded and shared. Keys written to the logs are redacted rather than encrypted (as no cipher is passed to the dump operation in that case). It seems this was not clear from the code - please make this objective clear and add warnings that this is not designed to thwart an attacker with full access to the file system, and that values written to the logs are redacted rather than encrypted.

3. The inability to deserialize existing conversations is acceptable as these are not serializing correctly at present anyway - the secrets are being redacted and not stored! So this is not a breaking change. Please handle the case where a decryption error occurs by printing a warning and setting the secret value to None. Also, can it be made more clear that this is what is happening?

openhands-sdk/openhands/sdk/utils/models.py
```diff
@@ -78,9 +78,8 @@ class OpenHandsModel(BaseModel):
     regenerate all the polymorphic mappings.
     """
 
-    def __init__(self, *args, **kwargs):
+    def model_post_init(self, _context):
         _rebuild_if_required()
-        super().__init__(*args, **kwargs)
```
tofarr, line 83: I had to change this because pydantic will not pass a context to model / field validators if an __init__ method exists.

openhands-agent-server/openhands/agent_server/config.py
```diff
@@ -116,8 +126,31 @@ class Config(BaseModel):
...
+    )
     model_config: ClassVar[ConfigDict] = {"frozen": True}
 
+    @property
+    def cipher(self) -> Cipher | None:
+        cipher = getattr(self, "_cipher", None)
+        if cipher is None:
+            if self.secret_key is None:
+                _logger.warning(
+                    "⚠️ OH_SECRET_KEY was not defined. Secrets will not "
```
enyst, line 145: I think the encryption is great, though maybe with this key? The session api key seems more tricky because it has a different purpose - and idk, is it sent to FE in V1? Is it sent to agent-server when running with local docker?

GPT-5 was very - very - unhappy about mixing these, for different reasons. 😅

openhands-agent-server/openhands/agent_server/config.py
```diff
@@ -116,8 +126,31 @@ class Config(BaseModel):
...
+                    "be persisted between restarts."
+                )
+                cipher = None
+            else:
+                cipher = Cipher(self.secret_key.get_secret_value())
+            setattr(self, "_cipher", cipher)
```
enyst, line 151: Not sure this works, or raise at runtime? I think, even if it works, we docstring Config as "Immutable configuration", so idk, conceptually it doesn't feel right to me to setattr on spot.

I drafted an alternative proposal (reconsidered to account for server persistence) to move the secret stuff to Workspace, it seems to me then we wouldn't need this

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -43,12 +45,23 @@ def conversation_dir(self):
 
     async def load_meta(self):
         meta_file = self.conversation_dir / "meta.json"
-        self.stored = StoredConversation.model_validate_json(meta_file.read_text())
+        self.stored = StoredConversation.model_validate_json(
```
enyst, line 48: This probably doesn't work, right? As noted in a previous PR, I can't find any callers of load_meta() in the code. 🤔

openhands-sdk/openhands/sdk/conversation/secrets_manager.py
```diff
@@ -2,37 +2,41 @@
...
+from openhands.sdk.utils.models import OpenHandsModel
 
 
 logger = get_logger(__name__)
 
 SecretValue = str | SecretSource
 
 
-class SecretsManager:
+class SecretsManager(OpenHandsModel):
```
enyst, line 20: Could you please tell, why is this an `OpenHandsModel`?

enyst:
Out of curiosity, you ran `/codereview-roasted` with Sonnet 4 ? Personally, I heavily prefer GPT-5. It has its idiosyncrasies, they all do, but I find it much better, specially on complex or code design / architecture issues. 😅 
(though I think Sonnet is much better at pirate-speak! 😂 )

Not always, though! 😂  I had OpenHands-GPT-5 do one of these on this PR, and it was this [gist](https://gist.github.com/enyst/31ab18399435b5ff10d8a0a1bff34bd6).

Which, IMHO, it's not a good/believable review. I _think_ we'd be better off to address code design (e.g. like secrets-belong-in-workspace proposal), rather than go into every detail of what the code details would be like in this secrets-per-conversation design...

I'd love to get your 👀 on this, instead of that gist, how do you think about something like that?
- https://github.com/OpenHands/agent-sdk/issues/909

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -95,8 +91,13 @@ class ConversationState(OpenHandsModel):
         description="Conversation statistics for tracking LLM metrics",
     )
 
+    # Secrets manager for handling sensitive data (changed from private attribute)
```
enyst, line 94: Re: the note above, this is now in `State` so it exists for any client callers of a Conversation, not just agent-server callers; it changes a bit behavior below too, in LLM, in a way that seems to add a bit of complexity 🤔 

This is one thing I was thinking maybe we could address, if the proposal on WorkspaceSecrets seems of any interest: maybe we could simplify code and our work with it, rather than adding special cases in core classes? Idk, it's not finished, but I'd love your opinion on that anyway!

openhands-sdk/openhands/sdk/utils/models.py
enyst: Yes, I feel like defining init with pydantic can spell trouble in some cases, so the change is probably for the better!

openhands-sdk/openhands/sdk/conversation/secrets_manager.py
tofarr: It is an OpenHandsModel instead of a BaseModel because one of its fields is polymorphic (`secret_sources`). It means that the polymorphic relations will be regenerated the first time an instance of this class is initialized. 

For example, if the very first call in your program is something like...
```
SecretsManager.model_validate({
  "secret_sources": { 
    "key": { 
      "kind": "StaticSecret",
      "value": "some-key",
    }
  }
})
```
It will regenerate the polymorphic mappings before validating and is therefore guaranteed to know what a "StaticSecret" is.

openhands-agent-server/openhands/agent_server/config.py
tofarr: Eventually I would like to deprecate the `SESSION_API_KEY` environment variable altogether (So that we just have `OH_SESSION_API_KEY` and `OH_SECRET_KEY`)

The reason I have postponed this is due to the Runtime API, whic has the concept of warm runtimes - We start a number of K8 pods with the blank docker image and hold them in reserve, and when a user starts a conversation, one of those warm runtimes is allocated to the user. The pre-generated SESSION_API_KEY is excluded from calculations of whether a runtime is appropriate for the use case.

I didn't want to mess with this logic right now by introducing additional variables and figured the legacy `SESSION_API_KEY` was a reasonable compromise until we deprecate it (It will also mean that we can still run V0 converstions and V1 conversations without changing the runtime api while we migrate.)

enyst: Sounds good enough to me, thank you!

Do we want it to work with local docker though? I thought we were only setting it at all for the cloud.

enyst:
Just to clarify, I meant the GPT-5 gist is "not good" in the sense of believable, to me, once I took into account more things here. It may have found some details, but I don't know if most of it is worth digging into: instead my focus was on understanding how we could do this better so that we are making it easier for client developers and, for us, the code complexity goes down... And I'm not sure what the final answer is, the proposal is just that, a draft design and PR.

openhands-agent-server/openhands/agent_server/config.py
tofarr: It does work:
```
export OH_SECRET_KEY=unsecure
uv run python                

>>> from openhands.agent_server.config import get_default_config
>>> from pydantic import SecretStr
>>> config = get_default_config()                                             
>>> config.cipher.encrypt(SecretStr("Some Message"))
'gAAAAABo_kbA04OBAjDMDTfXZCp-NS-Iw2u39P31B-szaN4-fqlHAokStiNfUCcp6dyeYXF6So58crP7KzwJr7qfxcjt-4UUw=='
```

To address your concerns, I've made `Cipher` frozen. (So `_cipher` is a lazy generated but immutable private attribute.

Can you post a link to your alternative proposal? (I know we have talked at length about config, but I figured we were not going to change this at the moment - the server has a shared config and sdk does not.)

openhands-agent-server/openhands/agent_server/event_service.py
tofarr: It is very similar to the code in https://github.com/OpenHands/agent-sdk/blob/2289140b15b6bcc3e7f8706c0e7725fd5d030095/openhands-agent-server/openhands/agent_server/conversation_service.py#L299 (Which exists because we need to load in order to actually create the service.)

I suppose it is future proofing, but it felt weird to have a `save_meta` method without a corresponding `load_meta` method.

openhands-agent-server/openhands/agent_server/config.py
enyst: It's not a config (or not the primary way I see it). It's more that the client does:
```
workspace = Workspace(host="...")
conversation = Conversation(...
    workspace=workspace,
)
```
Maybe the client dev could write code like this?
```
workspace.update_secrets(my_stuff)  # set if they don't exist
```
Related, the secrets are, in the app-server as client, set by the user in web UI, once, so they'd apply to the `agent-server` box, rather than per conversation. 🤔 

This is the proposal:
- https://github.com/OpenHands/agent-sdk/issues/909

The agent is working on a PR roughly according to it:
- [WorkspaceSecrets API](https://github.com/OpenHands/agent-sdk/pull/910/files#diff-8d033d613435c2d28aa1312d5c67ed6a37b2a8a08133f1c941b28816664f4030R18)

openhands-sdk/openhands/sdk/conversation/state.py
tofarr: I posted a note on the proposal - I think this PR will be significantly simpler than the implementation of that proposal - given that it does not change the model significantly. On the other hand - it will does present the unexpected case where persistence is effectively broken for a lot of cases if you don't have an encryption key defined.

Lets see what the consensus is tomorrow morning.

openhands-sdk/openhands/sdk/conversation/secrets_manager.py
enyst: Thank you, and sorry about this one. I finally went deep into that code and figured out some stuff that I totally missed before. That's such a great implementation, possibly the best I could think of on serialization!

I'll mark this solved so it doesn't cloud the view.

openhands-sdk/openhands/sdk/utils/cipher.py
```diff
@@ -0,0 +1,69 @@
...
+
+import hashlib
+from base64 import b64encode
+from dataclasses import dataclass
+
+from cryptography.fernet import Fernet
+from pydantic import SecretStr
+
+
+@dataclass(frozen=True)
```
xingyaoww: Shall we make it a pydantic just for consistent?

tofarr: I have been using data classes for things we most definitely never want to serialize - the cipher falls into this category. (Some of the services in the agent sdk fall into this too.) I just use it as a shorthand to not add an __init__ method.

Since these can still be converted to strings and serialized, I will actually go the other direction - remove the dataclass tag.

PR #908, opened 2025-10-26
feat: Implement blocking delegation

simonrosenberg:
---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:bfb7a06-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-bfb7a06-python \
  ghcr.io/openhands/agent-server:bfb7a06-python
```

```
ghcr.io/openhands/agent-server:bfb7a06-golang
ghcr.io/openhands/agent-server:bfb7a06-java
ghcr.io/openhands/agent-server:bfb7a06-python
```

_The `bfb7a06` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._

### DIFF
--- a/.openhands/microagents/repo.md
+++ b/.openhands/microagents/repo.md
@@ -139,3 +139,29 @@ When reviewing code, provide constructive feedback:
 - If you find yourself duplicating logics in preparing mocks, loading data etc, these logic should be fixtures in conftest.py!
 - Please test only the logic implemented in the current codebase. Do not test functionality (e.g., BaseModel.model_dumps()) that is not implemented in this repository.
 </TESTING>
+
+<DOCUMENTATION_WORKFLOW>
+# Documentation Repository
+
+Documentation lives in **github.com/OpenHands/docs** under the `sdk/` folder. When adding features or modifying APIs, you MUST update documentation there.
+
+## Workflow
+
+1. Clone docs repo: `git clone https://github.com/OpenHands/docs.git /workspace/project/openhands-docs`
+2. Create matching branch in both repos
+3. Update documentation in `openhands-docs/sdk/` folder
+4. **If you are creating a PR to `OpenHands/agent-sdk`**, you must also create a corresponding PR to `OpenHands/docs` with documentation updates in the `sdk/` folder
+5. Cross-reference both PRs in their descriptions
+
+Example:
+```bash
+cd /workspace/project/openhands-docs
+git checkout -b <feature-name>
+# Edit files in sdk/ folder
+git add sdk/
+git commit -m "Document <feature>
+
+Co-authored-by: openhands <openhands@all-hands.dev>"
+git push -u origin <feature-name>
+```
+</DOCUMENTATION_WORKFLOW>
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,6 +83,8 @@ extraPaths = [
     "openhands-workspace",
     "openhands-agent-server"
 ]
+venvPath = "."
+venv = ".venv"
 pythonVersion = "3.12"
 useLibraryCodeForTypes = true
 typeCheckingMode = "standard"


simonrosenberg:
@openhands 
Let's modify the delegation logic. Let's make delegation blocking.
It should have two actions
`spawn` takes a list of strings (ids) and initializes sub-conversations/agents with those ids
`delegate` takes a list of strings (tasks) and sends them to the sub-agents and then it runs the sub-agents in parallel. It waits for all sub-conversations to run and then sends back a single Observation containing the results of all sub-agent runs
This simplifies a lot: when `delegate` is over there are no other python processes alive than the main one.
Please simplify delegate/impl.py  and modify delegate/definition.py.
For testing/validation, run the example examples/01_standalone_sdk/25_agent_delegation.py
The success criteria is: this example runs end to end.

simonrosenberg:
> Sorry @simonrosenberg , I answered your comment on the `docs` PR, when in reality the discussion belongs here! I'll just copy this here for simplicity:
> 
> > We could simplify by having a single action 'delegate' with dictionary {id:task} argument that
> > 
> > * creates a new sub-agent and runs it with that task if the id is not existing
> > * sends the task to the existing sub-agent if already there
> 
> This sounds right to me. 🤔 (as Sonnet would say, "absolutely right!" 😂)
> 
> (I don't know if it would be better maybe to name the subagent or the task, the task seems ok. It would anyway name the subagent _for_ the task.)
> 
> Currently, we document that this would work like:
> 
> ```python
> # Agent uses the delegate tool to spawn sub-agents
> {
>     "command": "spawn",
>     "ids": ["lodging", "activities"]
> }
> ```
> 
> This seems like it would be belong as a parallel tool call, a tool 'delegate' and let the LLM decide to call it for each?

You mean: no need to implement parallelization on the tool side?
But multiple tool calls are executed sequentially with no async / no parallelism. Big waste of time for parallel delegation...

enyst:
In V0, it was not easy to figure out how to run multiple actions in parallel. I wonder if we could reconsider that for V1... 🤔 

_Edited to add_: wait, but these seem like two different things:
- we are not running parallel tool calls in parallel, in general, in a conversation
- in this PR, delegates have each their own separate conversation
    => so we run in parallel different conversations

simonrosenberg:
> In V0, it was not easy to figure out how to run multiple actions in parallel. I wonder if we could reconsider that for V1... 🤔
> 
> _Edited to add_: wait, but these seem like two different things:
> 
> * we are not running parallel tool calls in parallel, in general, in a conversation
> * in this PR, delegates have each their own separate conversation
>   => so we run in parallel different conversations

Yes but in only runs in parallel if the tool delegates takes in a dictionary as input.
If it takes a single id:task and expects the agent to make multiple delegate calls for multiple sub-agents, then it's not parallel it's back to sequential

simonrosenberg:
> delegation is working for me locally!
> 
> nit: I was wondering if we can add "conversation id" to the visualization frame or delegate name to the visualization? The metrics seems to be off (🤔 )
> 
> Image



> nit: I was wondering if we can add "conversation id" to the visualization frame or delegate name to the visualization? The metrics seems to be of

@xingyaoww are you OK with adding an additional optional parameter `name_for_visualization` so we can have 

```
sub_conversation = LocalConversation(
    agent=worker_agent,
    workspace=workspace_path,
    visualize=visualize,
    name_for_visualization=agent_id,  # <- Add this line
)
```

openhands-tools/openhands/tools/delegate/impl.py
```diff
@@ -0,0 +1,238 @@
...
+            DelegateObservation indicating success/failure and which agents were spawned
+        """
+        if not action.ids:
+            return DelegateObservation(
+                command="spawn",
+                success=False,
+                message="At least one ID is required for spawn action",
+            )
+
+        if len(action.ids) > self._max_children:
```
xingyaoww: should be this?
```suggestion
        if len(action.ids) >= self._max_children:
```

openhands-tools/openhands/tools/delegate/impl.py
```diff
@@ -0,0 +1,238 @@
...
+
+            def run_task(agent_id: str, conversation: LocalConversation, task: str):
+                """Run a single task on a sub-agent."""
+                try:
+                    logger.info(f"Sub-agent {agent_id} starting task: {task[:100]}...")
+                    conversation.send_message(task)
+                    conversation.run()
+
+                    # Extract the final response using agent_final_response
+                    final_response = conversation.agent_final_response()
```
xingyaoww: not super happy about this agent_final_responses function so propose to fix it here https://github.com/OpenHands/agent-sdk/issues/918 - hopefully we can clean this up before this PR

openhands-tools/openhands/tools/delegate/definition.py
```diff
@@ -0,0 +1,148 @@
...
+        description="The command that was executed. Either `spawn` or `delegate`."
+    )
+    success: bool = Field(description="Whether the operation was successful")
+    message: str = Field(description="Result message from the operation")
+    spawned_ids: list[str] | None = Field(
+        default=None, description="List of spawned sub-agent IDs (spawn command only)"
+    )
+    results: list[str] | None = Field(
+        default=None, description="Results from all sub-agents (delegate command only)"
+    )
```
xingyaoww: Can we merge `message` with `results`? i.e., just keep `message`, since you concatenate the result from mutliple sub-agent into one string anyway

openhands-tools/openhands/tools/delegate/impl.py
simonrosenberg:
```
if len(action.ids) > self._max_children:
            return DelegateObservation(
                command="spawn",
                success=False,
                message=(
                    f"Cannot spawn {len(action.ids)} agents, "
                    f"maximum is {self._max_children}"
                ),
            )
```
I think it's natural to have the maximum included in what is possible.
Else the error message will be weird
`"Cannot spawn 10 agents, maximum is 10"`

simonrosenberg:
Screenshot 2025-10-27 at 17 53 20
@xingyaoww visualization is now more flexible!

openhands-tools/openhands/tools/delegate/impl.py
xingyaoww: I think what puzzles me is that you can have in theory action.ids == 11 when self._max_children == 10 since it doesn't check for >=, right? since bool(10 > 10)== false

openhands-tools/openhands/tools/delegate/impl.py
```diff
@@ -0,0 +1,232 @@
...
```
xingyaoww: Let's update this

simonrosenberg: > I think what puzzles me is that you can have in theory action.ids == 11 when self._max_children == 10 since it doesn't check for >=, right? since bool(10 > 10)== false

@xingyaoww I am sorry I still dont understand. If _max_children is 10, you can spawn up to 10 children. If len(action.ids) is 11 it will return an error.

simonrosenberg: docs PR: https://github.com/OpenHands/docs/pull/54

openhands-tools/openhands/tools/delegate/impl.py
xingyaoww: @OpenHands can you help me /codereview-roasted and tell me if this will actually happen

xingyaoww: @simonrosenberg haha this is actually a bit orthogonal to my previous comment but i guess it is legit

simonrosenberg: I implemented the fix!

PR #930, opened 2025-10-27
feat: Add `llm_response_id` to all LLM generated events

ApGa:
## Summary
Add `llm_response_id` to all events that are generated using an LLM.
This is useful for associating LLM inference metadata with events that were produced with LLM calls.
(e.g., this allows storing information like logprobs indexed by `llm_response_id` outside of OpenHands events but still associate this data with the events; useful for agent training scenarios). 

## Note
We add this field at for the base `Event` type instead of `LLMConvertibleEvent` since an event can currently be generated by an LLM without being a subclass of `LLMConvertibleEvent` (e.g., `CondensationEvent`).

### DIFF
--- a/tests/sdk/context/condenser/test_llm_summarizing_condenser.py
+++ b/tests/sdk/context/condenser/test_llm_summarizing_condenser.py
@@ -159,6 +159,7 @@ def test_get_condensation_with_previous_summary(mock_llm: LLM) -> None:
         forgotten_event_ids=[events[3].id, events[4].id],
         summary="Previous summary content",
         summary_offset=keep_first,
+        llm_response_id="condensation_response_1",
     )
     events_with_condensation = (
         events[:keep_first] + [condensation] + events[keep_first:]
--- a/tests/sdk/context/test_view.py
+++ b/tests/sdk/context/test_view.py
@@ -42,7 +42,10 @@ def test_view_forgets_events() -> None:
     # The condensation specifically targets the IDs of all M_i messages
     events: list[Event] = [
         *message_events,
-        Condensation(forgotten_event_ids=message_event_ids),
+        Condensation(
+            forgotten_event_ids=message_event_ids,
+            llm_response_id="condensation_response_1",
+        ),
     ]
 
     # All events should be forgotten and removed.
@@ -62,7 +65,10 @@ def test_view_keeps_non_forgotten_events() -> None:
             # `test_view_forgets_events`, in this test we only want to forget
             # one of the events. That way we can check that the rest of the
             # events are preserved.
-            Condensation(forgotten_event_ids=[forgotten_event_id]),
+            Condensation(
+                forgotten_event_ids=[forgotten_event_id],
+                llm_response_id="condensation_response_1",
+            ),
         ]
 
         view = View.from_events(events)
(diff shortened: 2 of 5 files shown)


openhands-sdk/openhands/sdk/event/base.py
```diff
@@ -31,6 +31,17 @@ class Event(DiscriminatedUnionMixin, ABC):
     )  # consistent with V1
     source: SourceType = Field(..., description="The source of this event")
 
+    llm_response_id: EventID = Field(
+        default="",
```
xingyaoww: Let's do not set default here so we make sure this field is setted properly everywhere?

ApGa: Not every Event requires an `llm_response_id` (e.g., observation events or message events coming from a user). Without setting the default here, it might be awkward, as we'd then need to force each such event to explicitly set the llm_response_id to empty string anyways (or some other sensible default to denote an absence of the field)?

xingyaoww: ah, in that case, can we default to None so it is explicit?

ApGa: Done

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -218,12 +219,15 @@ def from_events(events: Sequence[Event]) -> "View":
...
                     summary_offset = event.summary_offset
+                    summary_llm_response_id = event.llm_response_id
                     break
 
         if summary is not None and summary_offset is not None:
             logger.debug(f"Inserting summary at offset {summary_offset}")
 
-            _new_summary_event = CondensationSummaryEvent(summary=summary)
+            _new_summary_event = CondensationSummaryEvent(
+                summary=summary, llm_response_id=summary_llm_response_id
```
enyst: It seems to me that strictly speaking, this wasn't generated by an LLM? The `Condensation` event was, and that one has `llm_response_id`. Technically this one is like an observation of that. 🤔 

If you think it helps, we could let it have it, but it seems to me there will be 2 with the same `response_id`, and they're not parallel tool calls of course.

ApGa: Yes, good catch. It seems like `CondensationEvents`  are the events that are actually stored and persisted in the `Conversation`, while `CondensationSummaryEvents` are only found in ephemeral `Views`. In this case, I do not think we need to propogate the `llm_response_id` to `CondensationSummaryEvents`.
  
I will revert this change.

ApGa: Done.

openhands-sdk/openhands/sdk/event/condenser.py
xingyaoww, line 1: @ApGa I think about this for a while and think it is probably better to include this field in events that is LLM output. What do you think about this change?

ApGa: Yes, I agree. That this is better. I was actually considering doing this and perhaps subclassing `MessageEvent` into `AgentMessageEvent`, `UserMessageEvent`, etc. so that we could avoid the None default value, but didn't end up doing it as it was a bigger refactor. 

But I think your change is a good compromise for now.

xingyaoww (approved):

PR #949, opened 2025-10-29
Fix(llm):When LLM calls are retried, it can lead to inaccurate records.

CLFutureX:
When LLM calls are retried, it can lead to inaccurate records.


CLFutureX:
@xingyaoww  PTAL

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -474,6 +473,7 @@ def completion(
         )
         def _one_attempt(**retry_kwargs) -> ModelResponse:
             assert self._telemetry is not None
+            self._telemetry.on_request(log_ctx=log_ctx)
```
xingyaoww, line 474: @OpenHands please help me /codereview and tell me what this change will do - how will it fix the issue, and also help me write a new test for it

enyst: Makes sense to me!

PR #733, opened 2025-10-14
feat(llm): forward extra_headers to LiteLLM completion and responses

enyst:
Summary
- Adds optional extra_headers field to LLM, and forwards it to LiteLLM both for chat completion and the Responses API.

Motivation
This supports scenarios like 
```
extra_headers={
        "anthropic-beta": "context-1m-2025-08-07"  # 👈 Enable 1M context
    }
```

Addresses parity with OpenHands changes and is part of fixing All-Hands-AI/OpenHands#11330, also an `extra_headers` case.

Changes
- openhands/sdk/llm/llm.py
  - New field: extra_headers: dict[str, str] | None
  - In completion(): if LLM.extra_headers is set and not explicitly overridden in kwargs, pass as extra_headers to litellm.completion
  - In responses(): do the same for litellm.responses
- tests/sdk/llm/test_llm.py
  - New tests to verify `extra_headers` are forwarded to both completion and responses paths

Notes
- Existing call-time extra_headers in kwargs (if any) take precedence over the configured one
- Followed existing SDK patterns; minimal surface change

Testing
- Added unit tests that patch litellm_completion and litellm_responses to assert extra_headers propagation

Docs
- No docs changes included; behavior is opt-in via the LLM config

Related Issues
- Fixes part of All-Hands-AI/OpenHands#11330

Co-authored-by: openhands <openhands@all-hands.dev>

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/b06bc28aa51a48b88e164865f79ae9d4)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
| python | amd64, arm64 | `nikolaik/python-nodejs:python3.12-nodejs22` | [Link](https://hub.docker.com/_/nikolaik/python-nodejs:python3.12-nodejs22) |
| golang | amd64, arm64 | `golang:1.21-bookworm` | [Link](https://hub.docker.com/_/golang:1.21-bookworm) |

**Pull (multi-arch manifest)**
```bash
# Each variant is a multi-arch manifest supporting both amd64 and arm64
docker pull ghcr.io/openhands/agent-server:b115c95-python
```

**Run**
```bash
[...]

PR #963, opened 2025-10-29
feat: add enterprise gateway support for LLM providers

ak684:
- Added an ssl_verify switch all the way through the OpenHands → LiteLLM call stack so gateway clients can point at their corporate CA bundle or disable verification in MITM environments.
  - Added an LLMWithGateway helper that centralizes custom gateway header bundles
  - Wrote LLMWithGateway tests with focused coverage on header merging and templating.

---

**Variants & Base Images**
|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:ef7deb7-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-ef7deb7-python \
  ghcr.io/openhands/agent-server:ef7deb7-python
```

```
ghcr.io/openhands/agent-server:ef7deb7-golang
ghcr.io/openhands/agent-server:v1.0.0a5_golang_tag_1.21-bookworm_binary
ghcr.io/openhands/agent-server:ef7deb7-java
ghcr.io/openhands/agent-server:v1.0.0a5_eclipse-temurin_tag_17-jdk_binary
ghcr.io/openhands/agent-server:ef7deb7-python
ghcr.io/openhands/agent-server:v1.0.0a5_nikolaik_s_python-nodejs_tag_python3.12-nodejs22_binary
```

_The `ef7deb7` tag is a multi-arch manifest (amd64/arm64); your client pulls the right arch automatically._


neubig (commented):
Thanks @ak684, I think this could be great!

One suggestion though, since this is a somewhat unusual use case, I wonder if we'd want to separate the code out more by creating a new

```
class LLMWithGateway(LLM)
```

in `llm_with_gateway.py` and then overriding any of the necessary functions.

I'd be a little bit hesitant to add all of these things to our `llm.py` class because it'll make it a bit more difficult to follow/maintain for our normal use cases.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -666,6 +667,7 @@ def _transport_call(
                     api_key=self.api_key.get_secret_value() if self.api_key else None,
                     base_url=self.base_url,
                     api_version=self.api_version,
+                    custom_llm_provider=self.custom_llm_provider,
```
enyst, line 692: Sorry for the dumb question, wasn't this in kwargs? I'm starting to wonder what happens to the others, are we even passing them if we didn't pass this one

ak684: Right now kwargs only contains what select_chat_options() returns (temperature/top_p/max_tokens/tool config, etc.). Before this change, the value lived on self.custom_llm_provider, but we never forwarded it when we called litellm_completion, so LiteLLM always saw custom_llm_provider=None and treated the call as a plain OpenAI request.

enyst: Thank you. 🤔 What I hear... is that it means none of the others, which are attributes of LLM but are not special cases in `select_chat_options`... are forwarded 🤔 

@openhands Look at this discussion. Make a new branch from main, and make a test where we check whether ALL LLM attributes, regardless if they're treated specially in `select_chat_options` or not, are forwarded to litellm. You'd want to mock litellm completion, so that you see what parameters do you really receive.

Note that we expect the test to probably fail, at least for the reason ak684 was fixing in this branch: `llm_custom_provider`. It's OK for the test to fail, as long as it tests correctly and finds a bug! Make a PR to main, with that test, so we can see it.

openhands-sdk/openhands/sdk/llm/llm_with_gateway.py
```diff
@@ -0,0 +1,102 @@
...
+        This method:
+        1. Adds custom headers
+        2. Performs basic template variable replacement
+        """
+        if not self.custom_headers:
+            return
+
+        # Start with existing headers
+        headers: dict[str, str] = {}
+        existing_headers = call_kwargs.get("extra_headers")
```
enyst: This should exist already in LLM... 🤔  It's a kwarg recognized by litellm

ak684: It does exist. The intention here is to intercept whatever is in it (extended thinking, responses api, etc.), render our configured custom headers, and then merge them before the call goes downstream.

enyst: Please correct me if I'm wrong:
- I think maybe we should expose `extra_headers` on any LLM, which is what [this PR](https://github.com/OpenHands/software-agent-sdk/pull/733) is proposing
- the rest... pick up `custom_headers` and resend as `extra_headers`... could the client code just send as `extra_headers`?

You know, perhaps what would help clear up could be to adjust one of the examples in examples/ directory, or make a new one, for using `extra_headers`, and then we see a little more clearly maybe what exactly works for the desired use case (I'm not sure the render code is necessary?)

enyst:

openhands-sdk/openhands/sdk/llm/llm_with_gateway.py
ak684: I'm in meetings all day but these are good points and I appreciate the comments! I expect to dig further into this later today or tomorrow. I've removed the rendering since I agree we can do without for now, and I will see whether or not it makes more sense to merge your PR first and then rebase this branch on that or if another approach makes sense

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -637,6 +655,11 @@ def _one_attempt(**retry_kwargs) -> ResponsesAPIResponse:
     # =========================================================================
     # Transport + helpers
     # =========================================================================
+    def _prepare_request_kwargs(self, call_kwargs: dict[str, Any]) -> dict[str, Any]:
```
enyst, line 658:
```suggestion
    def prepare_request_kwargs(self, call_kwargs: dict[str, Any]) -> dict[str, Any]:
```
Nit: if it's a hook, it probably should be part of public API

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -446,15 +454,19 @@ def completion(
...
         log_ctx = None
         if self._telemetry.log_enabled:
+            sanitized_kwargs = {
+                k: v for k, v in call_kwargs.items() if k != "extra_headers"
+            }
             log_ctx = {
                 "messages": formatted_messages[:],  # already simple dicts
                 "tools": tools,
-                "kwargs": {k: v for k, v in call_kwargs.items()},
+                "kwargs": sanitized_kwargs,
```
enyst, line 469: This might be better done in `telemetry.py`, because excluding all `extra_headers` means also excluding Anthropic's extended thinking from telemetry, and any other non-auth/non-authz client code use 🤔 

Maybe we could do it like this, and fix it in a follow-up?

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -936,31 +962,44 @@ def _unwrap_type(t: Any) -> Any:
...
             if t is SecretStr:
-                return SecretStr(raw)
+                return SecretStr(stripped)
             if t is bool:
-                return raw.lower() in TRUTHY
+                if lowered in TRUTHY:
+                    return True
+                if lowered in FALSY:
+                    return False
+                return None
```
enyst, line 983: Sorry to be dense, could you perhaps tell why we needed None for bool?

It seems that in the subclass, the new field `ssl_verify` is `None`, and if that is `None`, do we still need these changes?

enyst:
Rebuilt:
- https://github.com/OpenHands/software-agent-sdk/pull/1026

@ak684 should we close this one if it's not needed anymore?

enyst:
Superseded by 1026

PR #1005, opened 2025-11-03
Document credential access in execute_bash tool

xingyaoww:
## Summary

This PR enhances the `execute_bash` tool to help agents understand and use the secret manager system effectively.

### 1. Documentation (Initial Commit)
Added a new "Credential Access" section to the `TOOL_DESCRIPTION` in `openhands-tools/openhands/tools/execute_bash/definition.py` that documents:

- **Automatic secret injection**: How secrets are automatically exported as environment variables when referenced in bash commands
- **How to use secrets**: Clear examples showing agents how to reference secret keys in their commands (e.g., `$GITHUB_TOKEN`, `$API_KEY`)
- **Secret detection**: Explanation of the case-insensitive matching mechanism that detects secret keys in command text
- **Security**: Information about automatic output masking to prevent accidental secret exposure

### 2. Available Secrets in Metadata (New Feature)
Enhanced `ExecuteBashObservation` to show agents which secrets are available:

- **Added `available_secrets` field** to `CmdOutputMetadata` to store list of available secret names
- **Populate available secrets** from `conversation.state.secret_registry.secret_sources.keys()` in the executor
- **Display in LLM output** via `ExecuteBashObservation.to_llm_content()` showing format: `[Available secrets: $SECRET1, $SECRET2, $SECRET3]`
- **Added test coverage** in `test_bash_executor_metadata_available_secrets()` to verify the functionality

This helps agents discover what credentials they can access without having to guess or try different secret names.

## Implementation Details

The implementation reflects the actual secret manager behavior in `openhands-tools/openhands/tools/execute_bash/impl.py`:

1. Before executing a bash command, the `_export_envs()` method calls `secret_registry.get_secrets_as_env_vars(action.command)`
2. The secret registry searches the command text for registered secret keys (case-insensitive)
3. Found secrets are exported as environment variables using `export KEY="value"` before the actual command runs
4. Secret values are automatically masked in output, showing `<secret-hidden>` instead
5. Available secret names are now shown in the observation metadata and LLM-facing output

## Testing

- ✅ All pre-commit checks pass (ruff format, ruff lint, pycodestyle, pyright)
- ✅ All secret masking tests pass: `tests/tools/execute_bash/test_secrets_masking.py` (3 tests)
- ✅ New test added: `test_bash_executor_metadata_available_secrets()` verifies:
  - Available secrets are correctly populated in `metadata.available_secrets`
  - Secrets are displayed in LLM-facing content with `$` prefix
  - Multiple secrets are properly listed

## Example Output

When an agent executes a bash command with secrets available, they'll now see:

```
[Command: echo "test"]
test
[Current working directory: /workspace]
[Python interpreter: /usr/bin/python3]
[Available secrets: $GITHUB_TOKEN, $API_KEY, $DATABASE_URL]
[Command finished with exit code 0]
```

## Why This Matters

This enhancement helps agents:

1. **Understand** the correct way to access credentials in bash commands
2. **Discover** which credentials are available without trial and error
3. **Use** credentials correctly for tasks involving:
   - API interactions requiring authentication
   - Git operations needing tokens
   - Database access
   - External service integrations
   - Any authenticated operations

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:f90c770-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-f90c770-python \
  ghcr.io/openhands/agent-server:f90c770-python
```

```
ghcr.io/openhands/agent-server:f90c770-golang-amd64
ghcr.io/openhands/agent-server:f90c770-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:f90c770-golang-arm64
ghcr.io/openhands/agent-server:f90c770-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:f90c770-java-amd64
ghcr.io/openhands/agent-server:f90c770-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:f90c770-java-arm64
ghcr.io/openhands/agent-server:f90c770-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:f90c770-python-amd64
ghcr.io/openhands/agent-server:f90c770-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:f90c770-python-arm64
ghcr.io/openhands/agent-server:f90c770-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:f90c770-golang
ghcr.io/openhands/agent-server:f90c770-java
ghcr.io/openhands/agent-server:f90c770-python
```

- Each variant tag (e.g., `f90c770-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `f90c770-python-amd64`) are also available if needed


openhands-tools/openhands/tools/execute_bash/definition.py
```diff
@@ -212,6 +212,12 @@ def visualize(self) -> Text:
 ### Output Handling
 * Output truncation: If the output exceeds a maximum length, it will be truncated before being returned.
 
+### Credential Access
+* Automatic secret injection: When you reference a registered secret key in your bash command, the secret value will be automatically exported as an environment variable before your command executes.
+* How to use secrets: Simply reference the secret key in your command (e.g., `echo $GITHUB_TOKEN` or `curl -H "Authorization: Bearer $API_KEY" https://api.example.com`). The system will detect the key name in your command text and export it automatically.
```
xingyaoww:
```suggestion
* How to use secrets: Simply reference the secret key in your command (e.g., `echo $GITHUB_TOKEN` or `curl -H "Authorization: Bearer $API_KEY" https://api.example.com`). The system will detect the key name in your command text and export it as environment variable before it executes your command.
```

xingyaoww:
@OpenHands merge from main and resolve all conflicts for this pr

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -213,6 +216,12 @@ def visualize(self) -> Text:
 ### Output Handling
 * Output truncation: If the output exceeds a maximum length, it will be truncated before being returned.
 
+### Credential Access
+* Automatic secret injection: When you reference a registered secret key in your bash command, the secret value will be automatically exported as an environment variable before your command executes.
+* How to use secrets: Simply reference the secret key in your command (e.g., `echo $GITHUB_TOKEN` or `curl -H "Authorization: Bearer $API_KEY" https://api.example.com`). The system will detect the key name in your command text and export it as environment variable before it executes your command.
```
enyst:
```suggestion
* How to use secrets: Simply reference the secret key in your command (e.g., `echo ${GITHUB_TOKEN:0:8}` or `curl -H "Authorization: Bearer $API_KEY" https://api.example.com`). The system will detect the key name in your command text and export it as environment variable before it executes your command.
```
Maybe this example is a little better? 😅

xingyaoww: actually i'm not sure if `${GITHUB_TOKEN:0:8}` would be consider detected 🤣 if so, @OpenHands please add a test to make sure that happens

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -213,6 +216,12 @@ def visualize(self) -> Text:
 ### Output Handling
 * Output truncation: If the output exceeds a maximum length, it will be truncated before being returned.
 
+### Credential Access
+* Automatic secret injection: When you reference a registered secret key in your bash command, the secret value will be automatically exported as an environment variable before your command executes.
```
enyst, line 220: It's curious, I think, that we now tell the LLM in 3 / 4 bullet points, over and over, that it will be "exported before your command executes"

I'm not sure in V0 it knew this at all? I _think_ the LLM knew that "these vars are in your env"

openhands-tools/openhands/tools/terminal/definition.py
enyst: Ooh good point! Oops 🤔

openhands-tools/openhands/tools/terminal/definition.py
xingyaoww: V0 didn't know this - and it has been problematic once in a while (eg when a token expired and the agent doesn't know how to update it -- it simply complain to you that the secret is out of date...)

Speaking of this, @OpenHands please update the prompt suggests that some secret may be updated periodic - if an secret doesn't work, you can try export it again and check if the updated secret will solve the problem. a typical example here would be GITHUB_TOKEN which we use to set origin URL, and then the GITHUB_TOKEN expired, then we need to reset git origin url with the refreshed GITHUB_TOKEN

enyst: I just double checked, it knew nothing in bash prompt, and it knew just that "these vars are in your env" in the [runtime_info prompt](https://github.com/OpenHands/OpenHands/blob/95a44f424843a548c0f725a356aff26c690ea716/openhands/agenthub/codeact_agent/prompts/additional_info.j2#L38)

enyst: That's a very good point!

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -213,6 +216,13 @@ def visualize(self) -> Text:
...
+* Secret detection: The system performs case-insensitive matching to find secret keys in your command text. If a registered secret key appears anywhere in your command, its value will be made available as an environment variable.
+* Security: Secret values are automatically masked in command output to prevent accidental exposure. You will see `<secret-hidden>` instead of the actual secret value in the output.
+* Refreshing expired secrets: Some secrets (like GITHUB_TOKEN) may be updated periodically or expire over time. If a secret stops working (e.g., authentication failures), try using it again in a new command - the system will automatically use the refreshed value. For example, if GITHUB_TOKEN was used in a git remote URL and later expired, you can update the remote URL with the current token: `git remote set-url origin https://${GITHUB_TOKEN}@github.com/username/repo.git` to pick up the refreshed token value.
```
enyst:
```suggestion
* Refreshing expired secrets: Some secrets (like GITHUB_TOKEN) may be updated periodically or expire over time. If a secret stops working (e.g., authentication failures), try using it again in a new command - the system should automatically use the refreshed value. For example, if GITHUB_TOKEN was used in a git remote URL and later expired, you can update the remote URL with the current token: `git remote set-url origin https://${GITHUB_TOKEN}@github.com/username/repo.git` to pick up the refreshed token value.
* If it still fails, report it to the user.
```

OK! Looks good. Verbose, but these LLMs will deal with it.

This is not always true, though: this happens on the Cloud, and I don't think, as of now, anywhere else in the project. So how about we add an escape hatch, maybe, for the LLM to not assume it will always happen?

enyst: e.g.
"* If it still fails, report it to the user."
or "you can ask the user about it"

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -117,6 +117,9 @@ def to_llm_content(self) -> Sequence[TextContent | ImageContent]:
             ret += f"\n[Current working directory: {self.metadata.working_dir}]"
         if self.metadata.py_interpreter_path:
             ret += f"\n[Python interpreter: {self.metadata.py_interpreter_path}]"
+        if self.metadata.available_secrets:
+            secrets_list = ", ".join(f"${s}" for s in self.metadata.available_secrets)
```
enyst, line 121: Ah interesting, so they will be sent to the LLM every time it runs a command. Let's see how it works

xingyaoww:
@enyst yeah, i think we could try @hieptl 's PR first, but i think the prompts in the system prompt would be worth copying over

PR #1011, opened 2025-11-04
fix: recursively parse nested tools params

Edward-sy:
Resolve https://github.com/OpenHands/software-agent-sdk/issues/1074

The nested parameters inside the array were not parsed, which caused the inner parameters of task_list to be missing.

This PR adds recursive parsing logic for nested parameters in fn_call_converter.py.

With this change, all nested fields inside task_list are now correctly included in the generated description, improving task_tracker task planning and tracking behavior.

### DIFF
--- a/tests/sdk/llm/test_llm_fncall_converter.py
+++ b/tests/sdk/llm/test_llm_fncall_converter.py
@@ -1,6 +1,7 @@
 """Test for FunctionCallingConverter."""
 
 import json
+from typing import cast
 
 import pytest
 from litellm import ChatCompletionToolParam
@@ -14,6 +15,7 @@ from openhands.sdk.llm.mixins.fn_call_converter import (
     convert_fncall_messages_to_non_fncall_messages,
     convert_non_fncall_messages_to_fncall_messages,
     convert_tool_call_to_string,
+    convert_tools_to_description,
 )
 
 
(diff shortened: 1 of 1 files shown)


openhands-sdk/openhands/sdk/llm/mixins/fn_call_converter.py
```diff
@@ -329,6 +329,53 @@ def index():
...
+</function>
+""",
+        "plan": """
+I'll create or update the full plan based on your requirements and current progress:
+<function=task_tracker>
+<parameter=command>plan</parameter>
+<parameter=task_list>
+[
+  {
+    "id": "task-1",
```
ryanhoangt: I think for the new task item schema, we don't have the `id` and `description`:

https://github.com/OpenHands/software-agent-sdk/blob/8d05e95340793547da2819bd19b62a0e46177437/openhands-tools/openhands/tools/task_tracker/definition.py#L29-L35

openhands-sdk/openhands/sdk/llm/mixins/fn_call_converter.py
```diff
@@ -329,6 +329,53 @@ def index():
...
+    "description": "Create repository and basic docs."
+  },
+  {
+    "id": "task-2",
+    "title": "Implement nested param parsing",
+    "status": "in_progress",
+    "notes": "Add recursive parsing for array-typed parameters.",
+    "description": "Update fn_call_converter.convert_tools_to_description to walk nested objects in arrays."
+  },
+  {
```
ryanhoangt: Maybe we can remove some examples, e.g. only keep id 1 and 2, to reduce some token usage.

openhands-sdk/openhands/sdk/llm/mixins/fn_call_converter.py
```diff
@@ -485,6 +538,32 @@ def convert_tools_to_description(tools: list[ChatCompletionToolParam]) -> str:
...
+                                "description", "No description provided"
+                            )
+
+                            # Handle enum values for nested items
+                            if "enum" in item_param_info:
+                                item_enum_values = ", ".join(
+                                    f"`{v}`" for v in item_param_info["enum"]
+                                )
+                                item_desc += f" Allowed values: [{item_enum_values}]"
+
```
ryanhoangt, line 547: Can you add a test case for this formatting logic on array type? Maybe `test_llm_fncall_converter.py` is a good place.

openhands-sdk/openhands/sdk/llm/mixins/fn_call_converter.py
Edward-sy: Sorry, I forgot to delete 'description'. And I also think the 'id' field could be removed to reduce token usage.

openhands-sdk/openhands/sdk/llm/mixins/fn_call_converter.py
Edward-sy: I add a unit test covering the array-of-object formatting in `test_llm_fncall_converter.py`.

ryanhoangt (approved):
LGTM! Tested and it seems to work now, thank you!

Image

PR #1020, opened 2025-11-04
fix: sanitize @OpenHands only in GitHub comments (backend)

YakshithK:
## Summary

This PR implements backend sanitization of @OpenHands mentions in GitHub comments to prevent self-mention loops. Instead of mentioning the OpenHands agent, the sanitization inserts a zero-width joiner after the @ symbol, making the mention non-clickable while maintaining readability.

## Changes

- Added `sanitize_openhands_mentions()` function in `openhands/sdk/utils/github.py` that uses a zero-width joiner (U+200D) to prevent @OpenHands mentions from being parsed as clickable references
- Updated GitHub workflow files to sanitize @OpenHands mentions before posting comments
- Updated example agent script to sanitize review comments before posting
- Exported the new function from the utils module

## Technical Details

- The sanitization is performed **in the backend** (when outputting to GitHub comments) rather than at the agent/frontend level
- Uses regex pattern matching to identify @OpenHands mentions (case-insensitive)
- Inserts zero-width joiner after @ to break the mention linkage
- Maintains text readability since the zero-width joiner is invisible

## Related PR

See OpenHands/OpenHands#11550 for the corresponding changes in the main repository.

### DIFF
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -239,7 +239,6 @@ jobs:
                   echo "$COMMENT_BODY" | gh pr comment ${{ github.event.pull_request.number }} --body-file -
               env:
                   GH_TOKEN: ${{ github.token }}
-                  PYTHONPATH: ${{ github.workspace }}
 
             - name: Read consolidated report for tracker issue
               if: github.event_name == 'schedule'
@@ -250,8 +249,6 @@ jobs:
                   echo "report<<EOF" >> $GITHUB_OUTPUT
                   echo "$REPORT_CONTENT" >> $GITHUB_OUTPUT
                   echo "EOF" >> $GITHUB_OUTPUT
-              env:
-                  PYTHONPATH: ${{ github.workspace }}
 
             - name: Comment with results on tracker issue
               if: github.event_name == 'schedule'


.github/workflows/run-examples.yml
```diff
@@ -89,6 +89,21 @@ jobs:
...
+import sys
+import re
+ZWJ = '\u200D'
+text = sys.stdin.read()
+pattern = r'@(?=\b[Oo]pen[Hh]ands\b)'
+sanitized = re.sub(pattern, f'@{ZWJ}', text, flags=re.IGNORECASE)
+print(sanitized, end='')
+"
+                  }
+
```
simonrosenberg: Do we need to have sanitize_comment() defined here when there is a sanitize_openhands_mentions method defined in utils/github.py?

simonrosenberg:
@openhands 

Make sure that sanitize_openhands_mentions is applied to all agents outputs that is then posted as a comment to github. 
.github/workflows/run-examples.yml should use sanitize_openhands_mentions instead of redefining a method.

.github/workflows/integration-runner.yml
```diff
@@ -225,21 +225,25 @@ jobs:
...
               if: github.event_name == 'pull_request_target'
               run: |
-                  COMMENT_BODY=$(cat consolidated_report.md)
+                  # Sanitize @OpenHands mentions to prevent self-mention loops
+                  COMMENT_BODY=$(uv run python -c "from openhands.sdk.utils.github import sanitize_openhands_mentions; import sys; print(sanitize_openhands_mentions(sys.stdin.read()), end='')" < consolidated_report.md)
                   # Use GitHub CLI to create comment with explicit PR number
                   echo "$COMMENT_BODY" | gh pr comment ${{ github.event.pull_request.number }} --body-file -
               env:
                   GH_TOKEN: ${{ github.token }}
+                  PYTHONPATH: ${{ github.workspace }}
```
enyst: Is it necessary to define PYTHONPATH?

enyst:

simonrosenberg:
@openhands please resolve conflicts

.github/workflows/integration-runner.yml
simonrosenberg: According to OpenHands
```You're absolutely right! The PYTHONPATH is not necessary because:

The workflow runs uv sync --dev which properly installs all workspace packages including openhands-sdk
Once installed via uv sync, the package is available in the Python environment without needing PYTHONPATH
The uv run python command already uses the correct virtual environment with the installed packages
Let me remove the unnecessary PYTHONPATH settings:```

PR #1025, opened 2025-11-04
Improve visualization API: allow passing ConversationVisualizer directly

xingyaoww:
## Summary

This PR improves the visualization API to make it more intuitive and flexible by allowing users to pass a `ConversationVisualizer` instance directly to the `visualize` parameter.

## Motivation

Currently, if a user wants to use a custom visualizer, they have to:
```python
# Current workaround (confusing):
conversation = Conversation(
    agent=agent,
    workspace=cwd,
    visualize=False,  # Confusing: we DO want visualization!
    callbacks=[custom_visualizer.on_event],
)
```

This is counter-intuitive because setting `visualize=False` sounds like you DON't want visualization, but you actually do - just with a custom visualizer.

## Changes

Enhanced the `visualize` parameter to accept three types:
- **`True`**: Use default visualizer (existing behavior, backward compatible)
- **`False` or `None`**: No visualization (existing behavior, backward compatible)  
- **`ConversationVisualizer` instance**: Use custom visualizer (**NEW**)

### Example Usage (New)

```python
# Ideal (now supported):
conversation = Conversation(
    agent=agent,
    workspace=cwd,
    visualize=custom_visualizer,  # Direct and clear!
)

# Or to disable visualization:
conversation = Conversation(
    agent=agent,
    workspace=cwd,
    visualize=None  # or False
)
```

### Files Modified

- `openhands-sdk/openhands/sdk/conversation/conversation.py`: Updated factory class type signatures
- `openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py`: Implemented new logic for LocalConversation
- `openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py`: Implemented new logic for RemoteConversation
- `tests/sdk/conversation/local/test_conversation_visualize_param.py`: Added comprehensive tests for the new functionality
- `examples/01_standalone_sdk/20_custom_visualizer.py`: Added example demonstrating the new feature
- `.github/scripts/check_documented_examples.py`: Improved error messages with step-by-step documentation instructions

## Backward Compatibility

✅ **Fully backward compatible** - all existing code will continue to work:
- `visualize=True` (default) → creates default visualizer (unchanged)
- `visualize=False` → no visualization (unchanged)
- All existing tests continue to pass

## Testing

- Added 3 new test cases covering the new functionality
- All 65 tests in local conversation tests pass
- Pre-commit checks (ruff, pyright, pycodestyle) all pass

## Documentation

📚 **Related documentation PR**: https://github.com/OpenHands/docs/pull/77

The documentation includes:
- Complete guide for using custom visualizers
- Examples of highlight patterns and customization
- Before/after comparison showing the API improvement

---

This addresses the feedback about making the visualization system more intuitive and user-friendly.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:a578ad3-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-a578ad3-python \
  ghcr.io/openhands/agent-server:a578ad3-python
```

```
ghcr.io/openhands/agent-server:a578ad3-golang-amd64
ghcr.io/openhands/agent-server:v1.0.0a5_golang_tag_1.21-bookworm_binary-amd64
ghcr.io/openhands/agent-server:a578ad3-golang-arm64
ghcr.io/openhands/agent-server:v1.0.0a5_golang_tag_1.21-bookworm_binary-arm64
ghcr.io/openhands/agent-server:a578ad3-java-amd64
ghcr.io/openhands/agent-server:v1.0.0a5_eclipse-temurin_tag_17-jdk_binary-amd64
ghcr.io/openhands/agent-server:a578ad3-java-arm64
ghcr.io/openhands/agent-server:v1.0.0a5_eclipse-temurin_tag_17-jdk_binary-arm64
ghcr.io/openhands/agent-server:a578ad3-python-amd64
ghcr.io/openhands/agent-server:v1.0.0a5_nikolaik_s_python-nodejs_tag_python3.12-nodejs22_binary-amd64
ghcr.io/openhands/agent-server:a578ad3-python-arm64
ghcr.io/openhands/agent-server:v1.0.0a5_nikolaik_s_python-nodejs_tag_python3.12-nodejs22_binary-arm64
ghcr.io/openhands/agent-server:a578ad3-golang
ghcr.io/openhands/agent-server:a578ad3-java
ghcr.io/openhands/agent-server:a578ad3-python
```

- Each variant tag (e.g., `a578ad3-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `a578ad3-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/visualizer/default.py
+++ b/openhands-sdk/openhands/sdk/conversation/visualizer/default.py
@@ -61,7 +61,7 @@ class DefaultConversationVisualizer(ConversationVisualizerBase):
     def __init__(
         self,
         name: str | None = None,
-        highlight_regex: dict[str, str] | None = None,
+        highlight_regex: dict[str, str] | None = DEFAULT_HIGHLIGHT_REGEX,
         skip_user_messages: bool = False,
     ):
         """Initialize the visualizer.


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -61,7 +61,7 @@ def __init__(
         callbacks: list[ConversationCallbackType] | None = None,
         max_iteration_per_run: int = 500,
         stuck_detection: bool = True,
-        visualize: bool = True,
+        visualize: bool | ConversationVisualizer | None = True,
```
simonrosenberg: Instead of Union type
```suggestion
        visualizer: ConversationVisualizer | None = DefaultVisualizer,
```

jpshackelford:
## Documentation Review and API Alignment

In response to @simonrosenberg's feedback:

### Fixed Issues:
- Updated parameter name from `visualize` to `visualizer` throughout example code
- Aligned with the simplified API that uses `visualizer=ConversationVisualizer/None/default`
- Enhanced error handling examples with more practical patterns

jpshackelford:
@OpenHands please fix the failing actions on PR #1025 at branch `feature/improve-visualizer-api` Be sure you understand the complete context of the PR and recent commits. Please also remember that we can omit the visualizer parameter to use the default visualizer. We shouldn't be adding an extra import to tests just to specify the default.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -112,15 +117,22 @@ def _default_callback(e):
...
         composed_list = (callbacks if callbacks else []) + [_default_callback]
-        # Add default visualizer if requested
-        if visualize:
+        # Handle visualization configuration
+        if isinstance(visualizer, ConversationVisualizer):
+            # Use custom visualizer instance
+            self._visualizer = visualizer
+            composed_list = [self._visualizer.on_event] + composed_list
+            # visualizer should happen first for visibility
+        elif visualizer is _DEFAULT_VISUALIZER:
```
xingyaoww: @OpenHands think about ways we can handle _DEFAULT_VISUALIZER in a more elegant way rather than making `visualizer: ConversationVisualizer | None | Any`

openhands-sdk/openhands/sdk/conversation/conversation.py
```diff
@@ -50,7 +47,7 @@ def __new__(
         callbacks: list[ConversationCallbackType] | None = None,
         max_iteration_per_run: int = 500,
         stuck_detection: bool = True,
-        visualizer: ConversationVisualizer | None | Any = _DEFAULT_VISUALIZER,
+        visualizer: bool | ConversationVisualizer | None = True,
```
jpshackelford: This looks very wonky -- I would not want to merge this.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -82,9 +78,9 @@ def __init__(
             callbacks: Optional list of callback functions to handle events
             max_iteration_per_run: Maximum number of iterations per run
             visualizer: Visualization configuration. Can be:
+                       - True: Use default visualizer (default)
+                       - False or None: No visualization
                        - ConversationVisualizer instance: Use custom visualizer
```
jpshackelford: This is exactly what @simonrosenberg was trying to avoid. Provide a visualizer or None. Passing no argument defaults to the default visualizer. True should not be a valid value.

jpshackelford:
@openhands -- your latest commit seems to be taking us in the wrong direction. Please see review comments from @simonrosenberg and @jpshackelford

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -112,15 +117,22 @@ def _default_callback(e):
...
-        # Add default visualizer if requested
-        if visualize:
+        # Handle visualization configuration
+        if isinstance(visualizer, ConversationVisualizer):
+            # Use custom visualizer instance
+            self._visualizer = visualizer
+            composed_list = [self._visualizer.on_event] + composed_list
+            # visualizer should happen first for visibility
+        elif visualizer is _DEFAULT_VISUALIZER:
+            # Create default visualizer
```
xingyaoww: @OpenHands let's define a ConversationVisualizerBase class as an interface and have Conversation accept `ConversationVisualizerBase | None`.

The interface should have two arguments for initialization:
1. name_for_visualization (leave it as the current default)
2. state (the conversation state, where we can grab .stats from

The idea is that maybe the ConversationVisualizer we pass into the Conversation class needs to be some sort of "Partially-initialized class", since we will need to attach the .state to it later.

Can you look through code and help propose a few ways to do this refactor that would allow us to achive the clean pattern:

- `Conversation(agent)` -> default visualizer
- `Conversation(agent, visualizer=None)` -> no visualizer
- `Conversation(agent, visualizer=custom)` -> custom visualizer that accepts name and state plus more custom stuff defined outside conversation

A potential option would be add an `initialize(self, state)` for `ConversationVisualizerBase` and make it abstract class (so child visualizer must implement it), and in initializer of Conversation, .initialize will be called.

For name, we can assume that the user can set them before they initialize the conversation (if they don't, we fallback to default)

jpshackelford:
@xingyaoww Something to consider now that we have a base class: use registry approach so that on_event doesn't have long if/else. You can see this pattern in the example file and also in the docs. c.f. [Event Handler Registration Pattern](https://allhandsai-feature-improve-visualizer-api.mintlify.app/sdk/guides/convo-custom-visualizer#3-event-handler-registration-pattern).

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -2,9 +2,13 @@
 import uuid
 from collections.abc import Mapping
 from pathlib import Path
+from typing import Any
 
 from openhands.sdk.agent.base import AgentBase
 from openhands.sdk.conversation.base import BaseConversation
+
+# Import sentinel value for default visualizer
+from openhands.sdk.conversation.conversation import _DEFAULT_VISUALIZER
```
xingyaoww: I REALLY HATE THIS! Any way to get rid of it? @OpenHands

Can we create an global instance of visualizer? Or can we accept `type[ConversationVisualizerBase]` here as well?

xingyaoww:
@jpshackelford maybe let's do this in a future PR?

I'm a bit hesitant for that behavior since that maintains some sort of "global" registry that can be very hard to maintain and debug in the long run

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -61,7 +61,9 @@ def __init__(
         callbacks: list[ConversationCallbackType] | None = None,
         max_iteration_per_run: int = 500,
         stuck_detection: bool = True,
-        visualize: bool = True,
+        visualizer: (
+            bool | type[ConversationVisualizerBase] | ConversationVisualizerBase | None
+        ) = True,
```
xingyaoww: @OpenHands you should not support `bool`! the default value should be the default ConversationVisualizer class and we plugin the correct arguments

jpshackelford:
> I'm a bit hesitant for that behavior since that maintains some sort of "global" registry that can be very hard to maintain and debug in the long run

Fine to wait given all of the pain we are having with this PR already.

But for future...

* The registry is not global - it's instance-scoped. 
* Each visualizer instance has its own _event_handlers dictionary that's created during __init__
* The registry is populated by scanning the instance's methods for the @handles() decorator
* Different visualizer instances have completely separate registries
* Bound methods are instance-specific
   * getattr(self, method_name) returns a bound method tied to the specific instance
   * Each instance's registry contains its own bound methods
   * No sharing between instances

There may be other approaches to dealing with if/else code smell--and I am a python neophyte--but this one seems pretty clean to me.

That said, totally agree about waiting...

openhands-sdk/openhands/sdk/conversation/conversation.py
raymyers: Yeah, that's very weird. The types are a good indicator of that here.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
jpshackelford: @xingyaoww Looks like you finally got rid of BOTH the boolean and that `_DEFAULT_VISUALIZER` construct!  Tests are clean. Are you happy with this? Do you want to take a look at the docs or is @hieptl helping with that?

examples/01_standalone_sdk/26_custom_visualizer.py
```diff
@@ -0,0 +1,87 @@
...
+- Direct configuration (just pass the visualizer instance to visualizer parameter)
+- Reusable visualizer that can be shared across conversations
+- Better separation of concerns compared to callback functions
+- Event handler registration to avoid long if/elif chains
+
+The MinimalProgressVisualizer produces concise output showing:
+- LLM call completions
+- Tool execution steps with command/path details
+- Agent thinking indicators
+- Error messages
```
xingyaoww:
```suggestion
```

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -19,7 +22,7 @@
 
 
 if TYPE_CHECKING:
-    from openhands.sdk.conversation.conversation_stats import ConversationStats
+    pass
```
malhotra5:
```suggestion
```

is this needed?

xingyaoww: nope! removed

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -107,15 +108,25 @@ def _default_callback(e):
...
+            self._visualizer.initialize(self._state)
+            composed_list = [self._visualizer.on_event] + composed_list
+            # visualizer should happen first for visibility
+        elif isinstance(visualizer, type) and issubclass(
+            visualizer, ConversationVisualizerBase
+        ):
+            # Instantiate the visualizer class with appropriate parameters
+            self._visualizer = visualizer()
+            # Initialize with state
+            self._visualizer.initialize(self._state)
```
malhotra5, line 125: Do we need this if else here;  is it always a class instead of an instance for visualizer param?

Maybe it can be simplified to this? 

```
if visulizer:
   self._visualizer = visualizer()
    # Initialize with state
    self._visualizer.initialize(self._state)
    composed_list = [self._visualizer.on_event] + composed_list

Comment (malhotra5) [reply openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py:125]: OH I see a user can pass either an instance or a class, got it

**Comment by enyst** [comment]
@openhands Read all review threads here. Understand that the most important goal is ease of use programmatically, so understand how this visualizer will be used by client developers, who may want some default visualization, or no visualizer, or a custom visualizer. But we also have other important goals like clean code and good design patterns in the implementation of the feature.

Do a /codereview-roasted on the PR with that in mind.

When you are done, provide your feedback in markdown, use some headings, bold, bulleted lists, so that it's readable. For any relevant negative feedback, provide a few alternatives.

File: openhands-sdk/openhands/sdk/conversation/visualizer/base.py [hunk 11]
```diff
@@ -0,0 +1,86 @@
...
+
+        Args:
+            name: Optional name to prefix in panel titles to identify
+                                  which agent/conversation is speaking. Will be
+                                  capitalized automatically.
+        """
+        self._name = name.capitalize() if name else None
+        self._state = None
+
+    def initialize(self, state: "ConversationStateProtocol") -> None:
```
malhotra5, line 49: NIT: maybe we can mark this with `@final` decorator from typings so that warnings are thrown in the event a user tries modifying this logic

xingyaoww: done

xingyaoww: done

openhands-sdk/openhands/sdk/conversation/conversation.py
```diff
@@ -46,8 +50,9 @@ def __new__(
         callbacks: list[ConversationCallbackType] | None = None,
         max_iteration_per_run: int = 500,
         stuck_detection: bool = True,
-        visualize: bool = True,
-        name_for_visualization: str | None = None,
+        visualizer: (
+            type[ConversationVisualizerBase] | ConversationVisualizerBase | None
```
enyst, line 54: Sorry, I'm a bit dense, is there a reason why this is a `type` too, or is it the agent overreacting?

xingyaoww: yes, basically in the default cases, it defaults to a class so that the conversation can initialize it -- i'm not sure if we can put `DefaultConversationVisualizer()` (an instance) as a default arg though?

malhotra5: I think because the user can pass either an instance or class for the visualizer

e.g `Conversation(visualizer=MyCustomVisualizer)` 

or 

```
custom_visualizer = MyCustomVisualizer()
Conversation(visualizer=custom_visualizer)
```


I feel we could simplify this by asking the user to only pass in a class instead of an instance. We can check in the `__init__` method whether the user sent an instance instead and raise an exception. It would also get rid of some duplicate code initializing the visualizer state.

xingyaoww: Here's what GPT-5 suggests

<html>
<head>

</head>
<body>

<p class="p1">Generally <span class="s1"><b>no</b></span>, it’s not a good idea to make the default argument of a function a <span class="s1"><b>class instance</b></span> (or any mutable object like a list, dict, or set).</p>
<p class="p2"><br></p>
<p class="p1">Here’s why — and when it can be acceptable.</p>
<p class="p3"><span class="s2"><hr></span></p>
<p class="p3"><span class="s2"><h3><b>⚠️ The core issue: Mutable default arguments</b></h3></span></p>
<p class="p2"><br></p>
<p class="p1">In Python, <span class="s1"><b>default argument values are evaluated once</b></span> — at the time the function is <i>defined</i>, not each time it’s <i>called</i>.</p>
<p class="p2"><br></p>
<p class="p1">That means if you do something like this:</p>


<pre><code>class Config:
    def __init__(self):
        self.count = 0

def do_something(cfg=Config()):
    cfg.count += 1
    return cfg.count</code></pre>


<p class="p1">You might expect each call to create a new <span class="s1">Config</span> object — but instead, every call shares the <span class="s2"><b>same instance</b></span>:</p>


<pre><code>do_something()  # 1
do_something()  # 2
do_something()  # 3</code></pre>


<p class="p1">This happens because the single <span class="s1">Config()</span> instance created at function definition time is reused.</p>
<p class="p2"><span class="s2"><hr></span></p>
<p class="p2"><span class="s2"><h3><b>✅ The correct pattern</b></h3></span></p>
<p class="p3"><br></p>
<p class="p1">If you want a new instance each time, make the default <span class="s1">None</span> and create it inside the function:</p>


<pre><code>def do_something(cfg=None):
    if cfg is None:
        cfg = Config()
    cfg.count += 1
    return cfg.count</code></pre>


<p class="p1">Now each call behaves as expected.</p>
<p class="p2"><span class="s1"><hr></span></p>
<p class="p2"><span class="s1"><h3><b>🧠 When a class instance<span class="Apple-converted-space"> </span></b></h3><h3><b>can</b></h3><h3><b><span class="Apple-converted-space"> </span>be a default</b></h3></span></p>
<p class="p3"><br></p>
<p class="p1">There are limited cases where this pattern is acceptable:</p>
<p class="p2"><span class="s1"><ul><li>
<p class="p1">The object is <span class="s1"><b>immutable</b></span> (e.g. a <span class="s2">frozen</span> Pydantic model, a tuple, or a dataclass with <span class="s2">frozen=True</span>).</p>
</li><li>
<p class="p1">The object is <span class="s1"><b>stateless</b></span> or <span class="s1"><b>pure</b></span>, meaning reusing it has no side effects.</p>
</li><li>
<p class="p1">It’s a <span class="s1"><b>sentinel</b></span> or <span class="s1"><b>constant</b></span> used only for comparison, not mutation.</p>
</li></ul></span></p>
<p class="p3"><br></p>
<p class="p1">Example:</p>


<pre><code>EMPTY_CFG = Config()
def process(cfg=EMPTY_CFG):
    ...</code></pre>


<p class="p1">As long as you treat <span class="s1">EMPTY_CFG</span> as read-only, this is fine.</p>
<p class="p2"><span class="s2"><hr></span></p>
<p class="p2"><span class="s2"><h3><b>✅ Best practice summary</b></h3></span></p>



Case | Good Idea? | Notes
-- | -- | --
Mutable object (list, dict, class instance) | ❌ | Causes shared state bugs
Immutable or stateless object | ✅ | Safe but rare
Use None as default and create new instance | ✅✅ | Idiomatic and safest




<p class="p1"><span class="s1"><hr></span></p>
<p class="p2">Would you like me to show the same pattern using a <span class="s2"><b>Pydantic model</b></span> as the default argument example (since you prefer Pydantic)?</p>

</body>
</html>

openhands-sdk/openhands/sdk/conversation/visualizer/base.py
```diff
@@ -0,0 +1,87 @@
...
+        """Initialize the visualizer with conversation state.
+
+        This method is called by Conversation after the state is created,
+        allowing the visualizer to access conversation stats and other
+        state information.
+
+        Subclasses can override this method to add custom initialization logic,
+        but should call super().initialize(state) to ensure the state is set.
+
```
enyst:
```suggestion
        Subclasses should not override this method, to ensure the state is set.

```
Nit: since we marked as final, small adjustment with our expectation

openhands-sdk/openhands/sdk/conversation/conversation.py
xingyaoww: So the `type[ConversationVisualizerBase]` serve the purpose of a default_factory

examples/01_standalone_sdk/26_custom_visualizer.py
```diff
@@ -0,0 +1,81 @@
+"""Custom Visualizer Example
+
+This example demonstrates how to create and use a custom visualizer by subclassing
+ConversationVisualizer. This approach provides:
+- Clean, testable code with class-based state management
+- Direct configuration (just pass the visualizer instance to visualizer parameter)
+- Reusable visualizer that can be shared across conversations
+- Better separation of concerns compared to callback functions
+- Event handler registration to avoid long if/elif chains
```
malhotra5:
```suggestion
```

OH being too verbose on breaking changes lol

examples/01_standalone_sdk/26_custom_visualizer.py
```diff
@@ -0,0 +1,81 @@
...
+class MinimalVisualizer(ConversationVisualizerBase):
+    """A minimal visualizer that print the raw events as they occur."""
+
+    def __init__(self, name: str | None = None):
+        """Initialize the minimal progress visualizer.
+
+        Args:
+            name: Optional name to identify the agent/conversation.
+                                  Note: This simple visualizer doesn't use it in output,
+                                  but accepts it for compatibility with the base class.
```
malhotra5: Not sure what this meant

examples/01_standalone_sdk/26_custom_visualizer.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        # Initialize parent - state will be set later via initialize()
+        super().__init__(name=name)
+
+    def on_event(self, event: Event) -> None:
+        """Handle events for minimal progress visualization."""
+        print(f"\n\n[EVENT] {type(event).__name__}: {event.model_dump_json()[:200]}...")
```
malhotra5, line 40: NIT: maybe we can include a bit of code on how to access the cost from the events since they get wired in later

examples/01_standalone_sdk/26_custom_visualizer.py
xingyaoww: removed

examples/01_standalone_sdk/26_custom_visualizer.py
```diff
@@ -0,0 +1,79 @@
...
+
+# Send a message and let the agent run
+print("Sending task to agent...")
+conversation.send_message("Write 3 facts about the current project into FACTS.txt.")
+conversation.run()
+print("Task completed!")
+
+# Report cost
+cost = llm.metrics.accumulated_cost
+print(f"EXAMPLE_COST: ${cost:.4f}")
```
malhotra5, line 77:
```suggestion
```

NIT: for the visualizer example would be nice to emit cost from the custom visualizer

examples/01_standalone_sdk/26_custom_visualizer.py
xingyaoww: I guess people can refer to DefaultConversationVisualizer for that if they want..

examples/01_standalone_sdk/26_custom_visualizer.py
xingyaoww: We need this report cost so `test-examples` github workflow can emit cost correctly

openhands-sdk/openhands/sdk/conversation/conversation.py
xingyaoww: > I feel we could simplify this by asking the user to only pass in a class instead of an instance. We can check in the __init__ method whether the user sent an instance instead and raise an exception. It would also get rid of some duplicate code initializing the visualizer state.

@malhotra5 this could limit the flexibility of this interface though -- user could customize their custom visualizer to accept a few more parameters at initialization, they may want to use those to initialize their class if needed, before pass them into the conversation

enyst: Ahh I see, thank you! Mutable default strikes again! Or rather, it wanted to. 🤔 

I agree with GPT-5, it's tricky because it's non-intuitive...

OK, maybe we could go with this

enyst: We might want to take Rohit's suggestion here
> I feel we could simplify this by asking the user to only pass in a class 

but idk if right now

xingyaoww: yeah i'd vote for go with the current version since it doesn't limit flexibility - but i can be convinced otherwise!

enyst: @openhands Read this particular review thread. Now, there are some Ruff rules that check for mutable defaults in python: find them, you can search the internet docs if you need to. I think it's more than one.

Then checkout main branch, and verify if our Ruff in pre-commit and in CI contains those rules.  If it doesn't, open an issue documenting the problem, and open a PR to add them.

enyst (approved):
Generally LGTM, thank you for this. I think it's more developer-friendly this way!

jpshackelford (approved):

PR #1034, opened 2025-11-05
fix(condenser):  When condensation is triggered by the unhandled_condensation_request condition,  it will result in empty condensation.

CLFutureX:
**Background:** When condensation is triggered by the unhandled_condensation_request condition, if the number of events in the current view is less than half of the max_size, it will result in empty condensation.

forExample: 

    keep_first = 4
    len(view) = 10
    max_size = 30

    target_size = 15
    events_from_tail = 10 
  
    forgotten_events = view[self.keep_first : -events_from_tail]
                     = view[4:-10]
    result: forgotten_events is empty

**Optimization:** When condensation is triggered by the unhandled_condensation_request condition, adjust the current target_size to len(view) // 2.

### DIFF
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -39,7 +39,7 @@ class LLMSummarizingCondenser(RollingCondenser):
         head = view[: self.keep_first]
         target_size = self.max_size // 2
         if view.unhandled_condensation_request:
-            # Condensation triggered by a condensation request 
+            # Condensation triggered by a condensation request
             # should be calculated based on the view size.
             target_size = len(view) // 2
         # Number of events to keep from the tail -- target size, minus however many


CLFutureX:
@csmith49  hey, PTAL

CLFutureX:
@xingyaoww  @enyst  hey，PTAL， thanks

enyst:
@openhands please make build, then run pre-commit on all files

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -38,6 +38,10 @@ def should_condense(self, view: View) -> bool:
     def get_condensation(self, view: View) -> Condensation:
         head = view[: self.keep_first]
         target_size = self.max_size // 2
+        if view.unhandled_condensation_request:
+            # Condensation triggered by a condensation request 
+            # should be calculated based on the view size.
+            target_size = len(view) // 2
```
enyst, line 44: @CLFutureX can you please tell, when does this happen, is it when the context limit exceeeded may have been reached?

CLFutureX: > @CLFutureX can you please tell, when does this happen, is it when the context limit exceeeded may have been reached?

sure 
path: openhands-sdk\openhands\sdk\agent\agent.py  

     `except LLMContextWindowExceedError:
    # If condenser is available and handles requests, trigger condensation
    if (
        self.condenser is not None
        and self.condenser.handles_condensation_requests()
    ):
        logger.warning(
            "LLM raised context window exceeded error, triggering condensation"
        )
        return `
 If calling the LLM goes over the window limit, it’ll trigger a LLMContextWindowExceedError. That’ll then create a CondensationRequest. After that, in the view file (openhands-sdk\openhands\sdk\context\view.py), if there’s a CondensationRequest, it’ll set unhandled_condensation_request to True. So when the should_condense method gets called next time, it’ll return True—and by then, most of the events in the view will be smaller than max_size.

enyst: @openhands See this discussion. I think @CLFutureX is correct, if context limit was exceeded we shouldn’t assume the condenser will work as if max_size of events was reach, it almost certainly wasn’t. What do you think about the solution or what quick reasonable alternatives do we have here?

CLFutureX:
> Thank you! I think PR 912 should also help here…

thats right，this is a good way to optimize.

PR #1050, opened 2025-11-06
Remove default path from ENV_ROOT assignment

SmartManoj:
The OPENHANDS_CONVERSATIONS_ROOT environment variable no longer falls back to a hardcoded default path. This change ensures that the path is only set if the environment variable is present.


enyst (approved):
Thank you!

PR #1052, opened 2025-11-06
Refactor: Always include risk fields

### LINKED ISSUE #819: Comprehensive overview for difficulties with security analyzer
This is a comprehensive issue which describes the current architecture specifically for the security analyzer. It also describes the difficulties client applications of the `agent-sdk` face when attempting to utilize this feature. The complexity can be broken into three categories as follows 

#### (UNSOLVED): Security analyzer <> security policy relation

The security analyzer and security policy are configured independently yet closely related. The idea being that we introduce new fine-grained client defined policies when using the same analyzer. However, currently it adds extra overhead for clients to maintain a reasonable mapping 

1. `Always confirm <> No analyzer required` (all actions by default require user confirmation)
2. `Never confirm <> No analyzer required` (all actions by default require no user confirmation)
3. `Risk based confirmation <> LLM analyzer` (LLM determines high risk actions which require user confirmation)

This is the expected set of mapping between security policies and analyzers at the moment. Unfortunately, clients can choose other combinations such as `Never Confirm <> LLM Analyzer`. This scenario is especially likely to occur when switching analyzers. In the V1 CLI, the user may enable the security analyzer mid conversation. This requires the client to enable the LLM security analyzer. If the client forgets to update the security policy as well, then confirmation mode will not work as expected. 

Related PRs:
- #305

#### (SOLVED): Reconciliation with persistent state

This related to the difficulty of toggling the security analyzer. We used to have logic that is triggered when the agent is initialized. Once initialized, updating the security analyzer would fail to update other parts of the agent (tools, prompts, etc). The workaround was to save everything to persistent state, restart the conversation and re-initialize the agent to trigger the same entry points. 

Related issue:
- #668: Not easy to remove security analyzer mid conversation

Related PRs:
- #669: Remove reconciliation requirements for security analyzer 

#### (UNSOLVED): Immutability and mid conversation switch 

The `security_risk` risk parameter is dynamically added to Tools depending on whether the LLM security analyzer is active. The tool descriptions are passed to the agent's system prompt, which is how the agent is aware of this requirement. The problem is that the agent history/system prompt is immutable. This leads to complications when toggling the security analyzer. We have three scenarios

Scenario 1: 
- Security analyzer exists when starting conversation
- System prompt includes tools with `security_risk`
- Security analyzer is removed mid conversation
- No risk field is expected, but LLM still believes it needs to pass this field because system prompt was not updated. Passing field will crash the conversation

Scenario 2: 
- Security analyzer was not included at start of conversation
- System prompt does not include information on passing `security_risk` fields
- Security analyzer is added later in the conversation
- `security_risk` is a required field, but LLM is unaware of this requirement since system prompt did not include this information

Scenario 3:
- Similar to 1, but:
- The security analyzer is removed mid conversation
- The conversation is fully reloaded
- ActionEvents are created in history, and they have the old `security_risk` field
- Validation fails, and leads to an infinite loop
```json
"arguments": "{\"command\": \"ls -al ~/.openhands\", \"security_risk\": \"LOW\"}
```

Related PRs: 
- #723: ignore `security_risk` when analyzer is not configured (solves scenario 1)
- #435: don't include `security_risk` information in system prompt when security analyzer is not active
- #341: don't include `security_risk` fields is analyzer is not enabled

Related Issues: 
- Slack report [o3 infinite loop when attempting to continue a conversation with tool calls with `security_risk`](https://openhands-ai.slack.com/archives/C09FJAZSP9N/p1762307430896749)
- https://github.com/All-Hands-AI/OpenHands/issues/11378
 - https://github.com/All-Hands-AI/OpenHands/issues/11309
 - https://github.com/All-Hands-AI/OpenHands/issues/11256

malhotra5:
## Summary

This PR does the following

1. makes sure tool schemas is always has the `security_risk` field 
2. validates `security_risk` field at runtime.
    - when llm security analyzer is configured + `security_risk` is missing -> send an error event to correct itself (good for strong models)
    - when llm security analyzer is NOT configured + `security_risk` is missing -> default to `SecurityRisk.UNKNOWN` and proceed as usually 
3. move security analyzer to conversation state to support easier mid conversation switches

Benefits

- Consistent schemas 
- Runtime enforcement of `security_risk` provides the desired flexibility
- Graceful error handling (we send back an error event so the model has a chance to retry, good for stronger models)
- No compromise on traceability (we have events registering when the security analyzer is added to a conversation; this is important since every event will contain `security_risk` field and they can be ignored depending on whether the analyzer has been configured)

The linked issue contains the following scenarios; this PR helps address each one of them

> Scenario 1:
> Security analyzer exists when starting conversation
> System prompt includes tools with security_risk
> Security analyzer is removed mid conversation
> No risk field is expected, but LLM still believes it needs to pass this field because system prompt was not updated. Passing field will crash the conversation

In this scenario, the `security_risk` parameter will still exist. The LLM can continue passing the field. We simply won't interfere with the control loop even though the agent is passing risk values. The agent can also forget to pass a risk parameter which will be ignored.

> Scenario 2:
> Security analyzer was not included at start of conversation
> System prompt does not include information on passing security_risk fields
> Security analyzer is added later in the conversation
> security_risk is a required field, but LLM is unaware of this requirement since system prompt did not include this information

This time `security_risk` information in the system prompt will always exist at the start of a conversation. The LLM is always aware of it. If analyzer is not configured and the LLM forgets to pass a value, we'll default to `SecurityRisk.UNKNOWN` and proceed as usual. When security analyzer is added later in the conversation, we'll make sure that at runtime an error event is raised if the LLM fails to pass the risk parameter.

> Scenario 3:
> Similar to 1, but:
> The security analyzer is removed mid conversation
> The conversation is fully reloaded
> ActionEvents are created in history, and they have the old security_risk field
> Validation fails, and leads to an infinite loop

The schema should always contain `security_risk`, so loading back events regardless of whether a section of conversation had or did not have a security analyzer should work.

Fixes #819

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/325d4d5d22c34f2ea6c6e675bf59d94f)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:8aaca4b-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-8aaca4b-python \
  ghcr.io/openhands/agent-server:8aaca4b-python
```

```
ghcr.io/openhands/agent-server:8aaca4b-golang-amd64
ghcr.io/openhands/agent-server:8aaca4b-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:8aaca4b-golang-arm64
ghcr.io/openhands/agent-server:8aaca4b-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:8aaca4b-java-amd64
ghcr.io/openhands/agent-server:8aaca4b-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:8aaca4b-java-arm64
ghcr.io/openhands/agent-server:8aaca4b-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:8aaca4b-python-amd64
ghcr.io/openhands/agent-server:8aaca4b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:8aaca4b-python-arm64
ghcr.io/openhands/agent-server:8aaca4b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:8aaca4b-golang
ghcr.io/openhands/agent-server:8aaca4b-java
ghcr.io/openhands/agent-server:8aaca4b-python
```

- Each variant tag (e.g., `8aaca4b-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `8aaca4b-python-amd64`) are also available if needed

### DIFF
--- a/tests/integration/schemas.py
+++ b/tests/integration/schemas.py
@@ -88,7 +88,11 @@ class ModelTestResults(BaseModel):
         total_tests = len(test_instances)
         successful_tests = sum(1 for t in test_instances if t.test_result.success)
         skipped_tests = sum(1 for t in test_instances if t.test_result.skipped)
-        success_rate = successful_tests / total_tests if total_tests > 0 else 0.0
+        # Exclude skipped tests from success rate calculation
+        non_skipped_tests = total_tests - skipped_tests
+        success_rate = (
+            successful_tests / non_skipped_tests if non_skipped_tests > 0 else 0.0
+        )
         total_cost = sum(t.cost for t in test_instances)
 
         return cls(
@@ -131,8 +135,13 @@ class ConsolidatedResults(BaseModel):
         # Calculate overall statistics
         total_tests_all = sum(r.total_tests for r in model_results)
         total_successful_all = sum(r.successful_tests for r in model_results)
+        total_skipped_all = sum(r.skipped_tests for r in model_results)
+        # Exclude skipped tests from overall success rate calculation
+        non_skipped_tests_all = total_tests_all - total_skipped_all
         overall_success_rate = (
-            total_successful_all / total_tests_all if total_tests_all > 0 else 0.0
+            total_successful_all / non_skipped_tests_all
+            if non_skipped_tests_all > 0
+            else 0.0
         )
         total_cost_all_models = sum(r.total_cost for r in model_results)
 
--- a/tests/integration/utils/generate_markdown_report.py
+++ b/tests/integration/utils/generate_markdown_report.py
@@ -24,7 +24,8 @@ def generate_model_summary_table(model_results: list[ModelTestResults]) -> str:
 
     for result in model_results:
         success_rate = f"{result.success_rate:.1%}"
-        tests_passed = f"{result.successful_tests}/{result.total_tests}"
+        non_skipped = result.total_tests - result.skipped_tests
+        tests_passed = f"{result.successful_tests}/{non_skipped}"
         skipped = f"{result.skipped_tests}"
         cost = format_cost(result.total_cost)
 
@@ -45,11 +46,12 @@ def generate_detailed_results(model_results: list[ModelTestResults]) -> str:
     sections = []
 
     for result in model_results:
+        non_skipped = result.total_tests - result.skipped_tests
         section_lines = [
             f"### {result.model_name}",
             "",
             f"- **Success Rate**: {result.success_rate:.1%} "
-            f"({result.successful_tests}/{result.total_tests})",
+            f"({result.successful_tests}/{non_skipped})",
             f"- **Total Cost**: {format_cost(result.total_cost)}",
             f"- **Run Suffix**: `{result.run_suffix}`",
         ]


openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -304,6 +307,37 @@ def _requires_user_confirmation(
 
         return False
 
+    def _extract_security_risk(
+        self,
+        arguments: dict,
+        tool_name: str,
+        readOnlyHint: bool,
```
xingyaoww: Let's use on consistent way of naming the variable
```suggestion
        read_only_tool: bool,
```

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -104,15 +98,24 @@ def init_state(
...
 
+        security_analyzer_event = SecurityAnalyzerConfigurationEvent.from_analyzer(
+            analyzer=self.security_analyzer
+        )
+
+        if (
+            len(security_analyzer_configuration_events) == 0
+            or security_analyzer_event != security_analyzer_configuration_events[-1]
+        ):
+            on_event(security_analyzer_event)
```
xingyaoww: It seems to me, this security analyzer event solely act as some sort of flag?

Can we add a new field to ConversationState to keep this information instead of making it an event?

malhotra5: So with this PR we always store `security_risk` param regardless of the security analyzer. So without knowing which analyzer was active at the time of the action being generated, a stored security_risk value is ambiguous: you can’t tell if it was enforced, advisory, or ignored.


SecurityAnalyzerConfigurationEvent provides a timestamped, minimal audit trail of the analyzer type in effect when the action was proposed. This is especially important as we don't record user confirmation events, they are implicitly assumed if the action was executed

xingyaoww: How about we keep a variable inside State that record the latest timestamp and the security analyzer type instead of emitting an event?

I'm hesitant about throwing this into events as this make the logic here very hard to parse

malhotra5: ok sounds good! i figured it would be easier if all the information was in one place rather than having to cross reference 

will make the change

malhotra5: So one reason I can think of adding the information to agent event history, instead of the conversation state, is that the `security_risk` is a default specific to the `Agent` class. Other folks who build their own agent implementations will use the same conversation state but may not have a `security_risk` requirement the way we do. So for specificity is makes sense to isolate that customization to the history for the default agent, rather than the shared state object

xingyaoww: imo we already have `confirmation_policy` in the state and it kinda make sense to have `SecurityAnalyzer` related information stored there as well?

malhotra5: ah that makes sense; in that case how about moving the `security_analyzer` out of the agent class as well? 

mid conversation analyzer switches are also tough to do because the agent is immutable. if its in the state we can easily update it - wdyt?

xingyaoww: yep! i think that's a more reasonable thing to do (although this means we are making breaking change 1 day after cutting the v1 release haha 🤣 )

malhotra5:
QA'ed with CLI and seems to work well

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -101,6 +117,12 @@ class ConversationState(OpenHandsModel):
         serialization_alias="secret_registry",
     )
 
+    # Security analyzer configuration history
+    security_analyzer_history: list[SecurityAnalyzerRecord] = Field(
```
xingyaoww: I don't see this gets used anywhere yet, let's remove it?

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -105,10 +106,9 @@ def init_state(
             event = SystemPromptEvent(
                 source="agent",
                 system_prompt=TextContent(text=self.system_message),
+                # Always include security_risk field in tools
                 tools=[
-                    t.to_openai_tool(
-                        add_security_risk_prediction=self._add_security_risk_prediction
-                    )
+                    t.to_openai_tool(add_security_risk_prediction=True)
```
xingyaoww, line 115: Can we allow a way to toggle these to False?

eg when the LLM is too weak, we want the user to have a way toggle this off.

Previously, the users can turn off security analyzer, and this will be turned off. But not anymore -- any alternative now?

openhands-sdk/openhands/sdk/conversation/state.py
malhotra5: this is purely for a security audit trail (because every action can have a risk associated with it, we need to know which actions had an analyzer active at the time)

i can remove it if there isn't a need though

openhands-sdk/openhands/sdk/agent/agent.py
malhotra5: the idea is to always have risk be part of the tool schema (otherwise the system prompt won't be consistent with the new schemas when they're toggled and we'll see scenario 1 surface as described pr description)

if the llm is too weak the user is expected to remove the llm based security analyzer from conversation state. 

when we detect `no risk was passed + llm security analyzer isn't configured` we proceed without any errors. the [code snippet here](https://github.com/OpenHands/software-agent-sdk/pull/1052/files#diff-b810beaa51e80f306c49b20042cb87751b8d33d7ce06147d0d14b9971db4451dR346-R349) handles the case with a comment explaining how it works

openhands-sdk/openhands/sdk/conversation/state.py
xingyaoww: When we do need action-level auditing, we should add this information to the event data structure rather than maintaining it in the state here (also, this makes `base_state` grows as time goes by..)

Let's remove it from this PR and add it back when we really need it

openhands-sdk/openhands/sdk/agent/agent.py
xingyaoww: Sounds good! Can you maybe leave a comment in the code so ppl are aware of this decision

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -73,8 +74,20 @@ class Agent(AgentBase):
     """
 
     @property
-    def _add_security_risk_prediction(self) -> bool:
-        return isinstance(self.security_analyzer, LLMSecurityAnalyzer)
+    def system_message(self) -> str:
+        """Override system prompt to always include security analyzer context."""
+        template_kwargs = dict(self.system_prompt_kwargs)
+        template_kwargs["llm_security_analyzer"] = True
```
xingyaoww: Can we simply add `{"llm_security_analyzer": True}` as a default value for `system_prompt_kwargs` instead of duplicating this `system_message` property here?

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -147,6 +156,22 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
...
+        if not isinstance(data, dict):
+            return data
+        d = dict(data)
+
+        if "security_analyzer" in d and d["security_analyzer"]:
+            warnings.warn(
+                AGENT_SECURITY_ANALYZER_DEPRECATION_WARNING,
+                DeprecationWarning,
+                stacklevel=3,
+            )
```
xingyaoww, line 171: @Openhands is there any good python library that allow us to manage deprecation warning in a more systematic way?

openhands-sdk/openhands/sdk/agent/agent.py
malhotra5: I've added a model validator to always include `llm_security_analyzer` as a default in `system_prompt_kwargs`

i didn't add it as a default to `AgentBase` because this is highly specific to our `Agent` implementation

I also didn't use pydantic field to add the default as if the user passed their own kwargs via (e.g `{"cli_mode": True}`, then `llm_security_analyzer` won't be included as expected

tests/agent_server/test_conversation_router.py
xingyaoww, line 1: @OpenHands please add a test to tests/cross/test_remote_conversation_live_server.py where we spawns a real agent server to test this

PR #1065, opened 2025-11-06
fix(examples): allow all examples to run with same LLM_API_KEY

jpshackelford:
## Problem

The hello world example uses a hardcoded Anthropic model name instead of reading from environment variables like other examples. This creates an inconsistent experience:

- Users must change `LLM_API_KEY` depending on which example they want to run
- Demo workflows are disrupted when you want to run hello world first, then move to other examples
- First-time users encounter confusing stack traces instead of clear error messages when API keys don't match making first impression that SDK or examples are buggy.

## Solution

This change makes hello world consistent with other examples by:
- Using `LLM_MODEL` environment variable with default `openhands/claude-sonnet-4-5-20250929`
- Adding a comment with instructions on where to get API keys and how to switch to other models

Now users can set up their environment once and run all examples with the same configuration, making the SDK more approachable for first-time users and smoother for demos.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:7c9df63-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-7c9df63-python \
  ghcr.io/openhands/agent-server:7c9df63-python
```

```
ghcr.io/openhands/agent-server:7c9df63-golang-amd64
ghcr.io/openhands/agent-server:7c9df63-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:7c9df63-golang-arm64
ghcr.io/openhands/agent-server:7c9df63-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:7c9df63-java-amd64
ghcr.io/openhands/agent-server:7c9df63-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:7c9df63-java-arm64
ghcr.io/openhands/agent-server:7c9df63-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:7c9df63-python-amd64
ghcr.io/openhands/agent-server:7c9df63-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:7c9df63-python-arm64
ghcr.io/openhands/agent-server:7c9df63-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:7c9df63-golang
ghcr.io/openhands/agent-server:7c9df63-java
ghcr.io/openhands/agent-server:7c9df63-python
```

- Each variant tag (e.g., `7c9df63-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `7c9df63-python-amd64`) are also available if needed


examples/01_standalone_sdk/01_hello_world.py
```diff
@@ -6,8 +6,13 @@
 from openhands.tools.terminal import TerminalTool
 
 
+# Credits for Claude Sonnet 4.5 model are available at
```
enyst:
```suggestion
# You can choose models via openhands provider at
```

examples/01_standalone_sdk/01_hello_world.py
```diff
@@ -6,8 +6,13 @@
...
+# https://app.all-hands.dev/settings/api-keys
+# For Anthropic or others use LiteLLM model names, e.g.
+# LLM_MODEL="anthropic/claude-sonnet-4-5-20250929"
+
```
enyst: Strictly speaking, it seems to me that `openhands/` prefix means it behaves like a provider, which passes through the call, like do others, e.g. OpenRouter. (`openrouter/`)

It's true LiteLLM itself doesn't recognize the prefix (we substitute it before calling litellm). On the other hand, model names are, I believe, exactly the standard names.

examples/01_standalone_sdk/01_hello_world.py
jpshackelford: That may be more technically correct, but probably less helpful to a new user. I am trying to accomplish two things with a very short message:

1. explain why we would use openhands provider in the example (e.g. as a connivence to you as we are providing some credits) 
2. explain where to get the API key that works with the examples so user can get started quickly

rbren (changes requested):
will give feedback in a min :)

examples/01_standalone_sdk/01_hello_world.py
enyst: I'm sorry for nitpicking, it seems to me that the first does not belong in the codebase. 

e.g. OpenRouter also has models served for free within some limits, but we don't advertise it in the codebase. I don't think this is the place for that.

There are many places where the users can be informed of the offer, aren't there?

enyst: Arguably, the same consideration may apply for the second point, though... it doesn't strike me the same way.

jpshackelford: I suppose we could link to getting started page in docs that would give more details on the settings.

jpshackelford:
Based on feedback via [slack](https://openhands-ai.slack.com/archives/C09FJAZSP9N/p1762449961975719), added a commit that:

1. Removes comment from `01_hello_world.py`
2. Updates all examples from `openhands/claude-sonnet-4-5-20250929` to  `anthropic/claude-sonnet-4-5-20250929`
3. Allows `LLM_MODEL` and `LLM_API_KEY` environment variables to override settings in `01_hello_world.py` as elsewhere

This commit DOES NOT address with the few cases where a model besides `openhands/claude-sonnet-4-5-20250929` is hardcoded.

examples/01_standalone_sdk/19_llm_routing.py
```diff
@@ -22,7 +22,7 @@
 # Configure LLM
 api_key = os.getenv("LLM_API_KEY")
 assert api_key is not None, "LLM_API_KEY environment variable is not set."
-model = os.getenv("LLM_MODEL", "openhands/claude-sonnet-4-5-20250929")
```
enyst, line 25: We may want to keep this one `openhands/`, because it has 2 LLMs, and they use the same key? 🤔 
It will work when we override LLM_MODEL from env, sure, but not as written

jpshackelford: Good catch!

enyst: It’s one of those that Xingyao referred to

enyst: @openhands Check this diff, and use `openhands/` provider here

PR #1092, opened 2025-11-07
feat(AsyncExecutor): AsyncExecutor allows adding tasks after shutdown, which may cause unpredictable exceptions.

CLFutureX:
**Background:** 
Currently, AsyncExecutor still permits task additions after being shut down. This could lead to unpredictable exceptions.
**Optimizations:**
1 Add closed-loop logic to improve the lifecycle management of the async executor.
2 Add pre-checks to reduce concurrent locking overhead.
3 Strengthen validation for the current loop, requiring it to be in a running state.
4 Optimize the waiting process for loop startup.
5 Fix resource leaks that might happen with the current shutdown


CLFutureX:
@ryanhoangt  @xingyaoww  hey, PTAL , thanks

CLFutureX:
> Based on my analysis of the AsyncExecutor changes, here is my review:
> 
> ### Issues Found
> **🔴 Critical Issues**
> 
> * **Race condition in `_ensure_loop()` early return (lines 32-33)**: The optimization that checks `if self._loop is not None and self._loop.is_running()` BEFORE acquiring the lock introduces a critical race condition. Between checking if the loop is running and returning it, another thread could call `_shutdown_loop()`, making the returned loop invalid. This violates the "thread safety" guarantee stated in the class docstring.
>   **Impact**: A caller could receive a reference to a loop that's being shut down or already stopped, leading to RuntimeErrors when trying to schedule tasks.
>   **Recommendation**: Remove the early return at lines 32-33, or move the check inside the lock to maintain thread safety.
> * **Bypass of shutdown check**: The early return at line 33 bypasses the `_shutdown.is_set()` check at line 35-36. This means that if shutdown has been initiated but the loop hasn't fully stopped yet, `_ensure_loop()` could still return the loop, allowing new tasks to be scheduled during shutdown.
>   **Recommendation**: Ensure the shutdown check always happens before returning a loop reference.
> 
> **🟡 Important Issues**
> 
> * **Typo in error message (line 36)**: Error message says "asyncExecutor has been shut down" but should be "AsyncExecutor" (capital E) for consistency with the class name.
> * **String continuation style (lines 41-42)**: The warning message uses backslash line continuation (`\`), which is not the preferred Python style. Implicit string concatenation with parentheses is cleaner and more maintainable.
>   ```python
>   # Current:
>   logger.warning("The loop is not empty, but it is not in a running state." \
>   " Under normal circumstances, this should not happen.")
>   
>   # Better:
>   logger.warning(
>       "The loop is not empty, but it is not in a running state. "
>       "Under normal circumstances, this should not happen."
>   )
>   ```
> * **Questionable "should not happen" warning**: The warning at lines 41-42 claims this scenario "should not happen under normal circumstances," but the new early return optimization actually makes this more likely due to potential race conditions.
> 
> **🟢 Minor Issues**
> 
> * **Import ordering (line 7)**: `import time` is placed after `from collections.abc import Callable`, breaking the convention of grouping standard library imports alphabetically. Should be moved up to line 4 (after `import threading`).
> * **Double-check locking optimization may be premature**: The performance benefit of the early return check (lines 32-33) is likely minimal for this use case, while it introduces significant correctness risks. Consider whether this optimization is actually needed based on profiling data.



> Based on my analysis of the AsyncExecutor changes, here is my review:
> 
> ### Issues Found
> **🔴 Critical Issues**
> 
> * **Race condition in `_ensure_loop()` early return (lines 32-33)**: The optimization that checks `if self._loop is not None and self._loop.is_running()` BEFORE acquiring the lock introduces a critical race condition. Between checking if the loop is running and returning it, another thread could call `_shutdown_loop()`, making the returned loop invalid. This violates the "thread safety" guarantee stated in the class docstring.
>   **Impact**: A caller could receive a reference to a loop that's being shut down or already stopped, leading to RuntimeErrors when trying to schedule tasks.
>   **Recommendation**: Remove the early return at lines 32-33, or move the check inside the lock to maintain thread safety.
> * **Bypass of shutdown check**: The early return at line 33 bypasses the `_shutdown.is_set()` check at line 35-36. This means that if shutdown has been initiated but the loop hasn't fully stopped yet, `_ensure_loop()` could still return the loop, allowing new tasks to be scheduled during shutdown.
>   **Recommendation**: Ensure the shutdown check always happens before returning a loop reference.
> 
> **🟡 Important Issues**
> 
> * **Typo in error message (line 36)**: Error message says "asyncExecutor has been shut down" but should be "AsyncExecutor" (capital E) for consistency with the class name.
> * **String continuation style (lines 41-42)**: The warning message uses backslash line continuation (`\`), which is not the preferred Python style. Implicit string concatenation with parentheses is cleaner and more maintainable.
>   ```python
>   # Current:
>   logger.warning("The loop is not empty, but it is not in a running state." \
>   " Under normal circumstances, this should not happen.")
>   
>   # Better:
>   logger.warning(
>       "The loop is not empty, but it is not in a running state. "
>       "Under normal circumstances, this should not happen."
>   )
>   ```
> * **Questionable "should not happen" warning**: The warning at lines 41-42 claims this scenario "should not happen under normal circumstances," but the new early return optimization actually makes this more likely due to potential race conditions.
> 
> **🟢 Minor Issues**
> 
> * **Import ordering (line 7)**: `import time` is placed after `from collections.abc import Callable`, breaking the convention of grouping standard library imports alphabetically. Should be moved up to line 4 (after `import threading`).
> * **Double-check locking optimization may be premature**: The performance benefit of the early return check (lines 32-33) is likely minimal for this use case, while it introduces significant correctness risks. Consider whether this optimization is actually needed based on profiling data.

Good analysis – this is indeed an area that needs further optimization. Here's the new adjustment: since the loop is a shared resource, we should add a lock for every write operation on it.

CLFutureX:
@openhands-agent  please help review it

enyst:
> @openhands-agent please help review it

Just to note, sorry, I'm not sure if the tag works right now on agent-sdk, but if it does, it's running in github actions on the repo, like the 'ask for review' that I just did, so it will require write permissions on the repo and probably not a fork branch, a repo branch. But you could install it on your repo and run it. 😅

CLFutureX:
> > @openhands-agent please help review it
> 
> Just to note, sorry, I'm not sure if the tag works right now on agent-sdk, but if it does, it's running in github actions on the repo, like the 'ask for review' that I just did, so it will require write permissions on the repo and probably not a fork branch, a repo branch. But you could install it on your repo and run it. 😅

@enyst  All adjusted， PTAL

enyst:

Everytime you answer the user, use markdown with headings, bold, and such, because it’s more readable. Introduce yourself in the first line too, so the reader knows who is speaking.

CLFutureX:
> * Ruff E501 violations (line-length=88)



> @OpenHands Do a /codereview-roasted on this PR.
> 
> Everytime you answer the user, use markdown with headings, bold, and such, because it’s more readable. Introduce yourself in the first line too, so the reader knows who is speaking.

hey， New adjustments:
1 Add cancellation for pending tasks
2 Move self._shutdown.set() inside the lock to avoid duplicate execution of subsequent logic:
`with self._lock:
    if self._shutdown.is_set():
        return
    self._shutdown.set()`
3 Format optimizations
4 The existing description of the close method may already clearly state that the executor's lifecycle will end:
 `"""Close the async executor and cleanup resources."""`

CLFutureX:
> # 📋 **Final Summary: AsyncExecutor Code Review Completed**
> ## **What I Did**
> I provided a comprehensive **"/codereview-roasted"** analysis of PR #1092 for the AsyncExecutor changes, as requested. Here's what I accomplished:
> 
> ### **✅ Review Process Completed:**
> 1. **Fetched PR Details** - Retrieved PR [feat(AsyncExecutor): AsyncExecutor allows adding tasks after shutdown, which may cause unpredictable exceptions. #1092](https://github.com/OpenHands/software-agent-sdk/pull/1092) metadata and diff from GitHub API
> 2. **Applied Changes Locally** - Created `fix_acync_executor` branch and applied the diff to examine the actual code
> 3. **Analyzed Code Quality** - Ran ruff checks and identified linting issues (E501, import sorting, TimeoutError usage)
> 4. **Delivered Brutal Review** - Provided Linus-style technical critique with specific line-by-line feedback
> 
> ### **🔥 Key Findings from My Review:**
> **Taste Rating: 🟡 Acceptable** (Works but violates simplicity principles)
> 
> **Critical Issues Identified:**
> 
> * **Over-engineered callback pattern** in `_safe_execute_on_loop()` - unnecessary complexity
> * **Double-checked locking anti-pattern** in shutdown logic - race condition prone
> * **Poor abstraction choices** - turned simple 3-line logic into complex callback soup
> 
> **Improvements Needed:**
> 
> * Fix E501 line length violations in logging statements
> * Eliminate unnecessary nested functions
> * Address fragile task cancellation logic with race conditions
> 
> **What They Got Right:**
> 
> * ✅ Task cancellation on timeout (real bug fix)
> * ✅ Shutdown state tracking with threading.Event
> * ✅ Proper resource cleanup during shutdown
> 
> ## **My Recommendation**
> **❌ Needs rework** - The author is solving a real problem (preventing task submission after shutdown) but with an over-engineered solution. The callback pattern should be reverted to a simple `_ensure_loop()` approach with shutdown checks added directly.
> 
> ## **No Code Changes Made**
> As requested for a code review, I **did not modify any code** - only provided detailed technical feedback for the author to address. The review identifies both the legitimate improvements (timeout cancellation, shutdown tracking) and the problematic over-engineering that needs simplification.
> 
> The user can now use this feedback to guide improvements to the PR before merging.
> 
> [View full conversation](https://app.all-hands.dev/conversations/0f52b5449cd94c9fb51582b2449b4e95)

hey，please review again

CLFutureX:
@ryanhoangt  hey, PTAL  ,thanks

openhands-sdk/openhands/sdk/utils/async_executor.py
```diff
@@ -83,16 +125,24 @@ def run_async(
             TypeError: If awaitable_or_fn is not a coroutine or async function
             asyncio.TimeoutError: If the operation times out
         """
+        if self._shutdown.is_set():
+            raise RuntimeError("AsyncExecutor has been shut down")
```
ryanhoangt: Do we need this check? I think we also perform it again in `_safe_submit_on_loop` below.

openhands-sdk/openhands/sdk/utils/async_executor.py
```diff
@@ -59,6 +87,20 @@ def _shutdown_loop(self) -> None:
                 pass
         if t and t.is_alive():
             t.join(timeout=1.0)
+            if t.is_alive():
+                logger.warning("AsyncExecutor thread did not terminate gracefully")
+
+        if loop and not loop.is_closed():
+            try:
+                if loop.is_running():
```
ryanhoangt: Since we call `loop.call_soon_threadsafe(loop.stop)` above i.e. we stop the loop, will this code ever be executed? I think it might be better to move the task cleanup before stopping the loop. Then we stop the loop -> down here, we only close the loop. WDYT?

openhands-sdk/openhands/sdk/utils/async_executor.py
CLFutureX: > Do we need this check? I think we also perform it again in `_safe_submit_on_loop` below.

Hmm, this is for a pre-check. If it's already shut down, we exit directly to avoid unnecessary execution later.
The check in _safe_submit_on_loop is a verification done after locking. Overall, this can be understood as the classic double-check pattern.

openhands-sdk/openhands/sdk/utils/async_executor.py
CLFutureX: > Since we call `loop.call_soon_threadsafe(loop.stop)` above i.e. we stop the loop, will this code ever be executed? I think it might be better to move the task cleanup before stopping the loop. Then we stop the loop -> down here, we only close the loop. WDYT?

Hmm, I think it will. Because loop.call_soon_threadsafe(loop.stop) just wraps the stop method into a task and adds it to the loop's _ready queue, waiting to be scheduled and executed in order. So at this point, the loop isn't closed yet; it will normally execute the remaining tasks in the queue one after another.
Then t.join(timeout=1.0) continues to wait for the loop to finish executing. If it still hasn't completed after exceeding the default 1 second, we need to actively cancel the unfinished tasks and then directly close the loop.

CLFutureX:
> Thanks for the PR, I left some comments below. I'm also thinking about moving to [AnyIO's `BlockingPortal`](https://anyio.readthedocs.io/en/stable/api.html#anyio.from_thread.BlockingPortal) inside `AsyncExecutor` instead of managing event loop and threading on our own in a future PR. This could help make the code easier to maintain and reduce potential errors. What do you think -- would you be interested in giving it a try in this PR?

Sure, I'll probably take some time to look into BlockingPortal tomorrow first, then make the adjustments.

CLFutureX:
@ryanhoangt   
Hey ! I took a look at AnyIO's BlockingPortal and tried implementing it, but I found a few issues-– if I understand correctly:

1. **No support for custom thread names**: Makes debugging and issue localization harder (can't identify threads spawned by the portal).
2. **Tightly coupled task submission and waiting**: The `portal.run(coro, timeout=timeout)` method combines submitting the task and waiting for its result. This creates tradeoffs:
   - If we call `portal.run()` inside a lock, the lock is held longer (blocks other operations).
   - If we try to separate submission and waiting, we risk concurrency safety issues between fetching the portal instance and submitting the task.
3. **No default graceful shutdown timeout**: Lacks built-in support for a default waiting period when closing resources (requires manual timeout handling).

 
Comparing the two, I think our existing implementation might be better. What do you think?. Here's the code for reference:

```python
import threading
import inspect
import asyncio
import logging
from typing import Callable, Any, Optional
from anyio import BlockingPortal, start_blocking_portal

logger = logging.getLogger(__name__)

class AsyncExecutor:
    """
    Manages async execution from sync contexts using AnyIO BlockingPortal.
    
    This provides a robust async-to-sync bridge with proper resource management,
    timeout support, and thread safety using AnyIO's battle-tested implementation.
    """

    def __init__(self):
        self._portal: Optional[BlockingPortal] = None
        self._portal_cm = None  # Holds the context manager for cleanup
        self._lock = threading.Lock()
        self._shutdown = threading.Event()

    def _ensure_portal(self) -> BlockingPortal:
        """Ensure the BlockingPortal is initialized and running."""
        with self._lock:
            if self._shutdown.is_set():
                raise RuntimeError("AsyncExecutor has been shut down")

            if self._portal is not None:
                return self._portal

            # Initialize a new BlockingPortal via its context manager
            self._portal_cm = start_blocking_portal()
            self._portal = self._portal_cm.__enter__()
            return self._portal

    def run_async(
        self,
        awaitable_or_fn: Callable[..., Any] | Any,
        *args,
        timeout: float = 300.0,
        **kwargs,
    ) -> Any:
        """
        Run a coroutine or async function from synchronous code.

        Args:
            awaitable_or_fn: Coroutine object or async function to execute
            *args: Positional arguments to pass to the async function
            timeout: Maximum execution time in seconds (default: 300s)
            **kwargs: Keyword arguments to pass to the async function

        Returns:
            The result of the async operation

        Raises:
            TypeError: If input is not a coroutine or async function
            TimeoutError: If the operation exceeds the timeout
            RuntimeError: If the executor has been shut down
            asyncio.CancelledError: If the operation is cancelled
        """
        if self._shutdown.is_set():
            raise RuntimeError("AsyncExecutor has been shut down")
            
        # Handle both coroutine objects and async functions
        if inspect.iscoroutine(awaitable_or_fn):
            coro = awaitable_or_fn
        elif inspect.iscoroutinefunction(awaitable_or_fn):
            coro = awaitable_or_fn(*args, **kwargs)
        else:
            raise TypeError("run_async expects a coroutine or async function")

        portal = self._ensure_portal()
        
        try:
            return portal.run(coro, timeout=timeout)
        except anyio.get_cancelled_exc_class():
            # Convert AnyIO's cancellation exception to asyncio's for compatibility
            raise asyncio.CancelledError()

    def close(self):
        """Gracefully shut down the executor and clean up resources."""
        if self._shutdown.is_set():
            logger.info("AsyncExecutor is already shut down")
            return

        with self._lock:
            if self._shutdown.is_set():
                return
            self._shutdown.set()
            portal_cm = self._portal_cm 
            # Clear references to avoid leaks
            self._portal_cm = None
            self._portal = None
 
        # Exit the BlockingPortal context manager to clean up resources
        if portal_cm is not None:
            try:
                portal_cm.__exit__(None, None, None)
            except Exception as e:
                logger.warning(f"Error closing BlockingPortal: {str(e)}", exc_info=True)

    def __del__(self):
        """Clean up resources when the instance is garbage collected."""
        try:
            self.close()
        except Exception:
            pass  # Ignore cleanup errors during garbage collection
```

CLFutureX:
@ryanhoangt  ping

ryanhoangt:
@CLFutureX actually I was thinking about a simpler implementation, where we don't maintain an internal lock or the shutdown flag. Since after closing, if we call `run_async` again, the portal is still recreated properly, I think we should remove the `_shutdown` flag for simplicity. It looks like below, WDYT?

```python
import anyio
from anyio.from_thread import start_blocking_portal
from typing import Any, Callable
import inspect

class AsyncExecutor:
    """
    Thin wrapper around AnyIO's BlockingPortal to execute async code
    from synchronous contexts with proper resource and timeout handling.
    """
    def __init__(self):
        self._portal = None
        self._portal_cm = None
    
    def _ensure_portal(self):
        if self._portal is None:
            self._portal_cm = start_blocking_portal()
            self._portal = self._portal_cm.__enter__()
        return self._portal
    
    def run_async(
        self,
        awaitable_or_fn: Callable[..., Any] | Any,
        *args,
        timeout: float | None = None,
        **kwargs,
    ) -> Any:
        """
        Run a coroutine or async function from sync code.
        
        Args:
            awaitable_or_fn: coroutine or async function
            *args: positional arguments (only used if awaitable_or_fn is a function)
            timeout: optional timeout in seconds
            **kwargs: keyword arguments (only used if awaitable_or_fn is a function)
        """
        portal = self._ensure_portal()
        
        # Construct coroutine
        if inspect.iscoroutine(awaitable_or_fn):
            coro = awaitable_or_fn
        elif inspect.iscoroutinefunction(awaitable_or_fn):
            coro = awaitable_or_fn(*args, **kwargs)
        else:
            raise TypeError("run_async expects a coroutine or async function")
        
        # Apply timeout by wrapping in an async function with fail_after
        if timeout is not None:
            async def _with_timeout():
                with anyio.fail_after(timeout):
                    return await coro
            return portal.call(_with_timeout)
        else:
            return portal.call(coro)
    
    def close(self):
        if self._portal is not None:
            self._portal_cm.__exit__(None, None, None)
            self._portal = None
            self._portal_cm = None
    
    def __enter__(self):
        return self
    
    def __exit__(self, exc_type, exc_val, exc_tb):
        self.close()
        return False
    
    def __del__(self):
        try:
            self.close()
        except Exception:
            pass
```

CLFutureX:
> @CLFutureX actually I was thinking about a simpler implementation, where we don't maintain an internal lock or the shutdown flag. Since after closing, if we call `run_async` again, the portal is still recreated properly, I think we should remove the `_shutdown` flag for simplicity. It looks like below, WDYT?
> 
> ```python
> import anyio
> from anyio.from_thread import start_blocking_portal
> from typing import Any, Callable
> import inspect
> 
> class AsyncExecutor:
>     """
>     Thin wrapper around AnyIO's BlockingPortal to execute async code
>     from synchronous contexts with proper resource and timeout handling.
>     """
>     def __init__(self):
>         self._portal = None
>         self._portal_cm = None
>     
>     def _ensure_portal(self):
>         if self._portal is None:
>             self._portal_cm = start_blocking_portal()
>             self._portal = self._portal_cm.__enter__()
>         return self._portal
>     
>     def run_async(
>         self,
>         awaitable_or_fn: Callable[..., Any] | Any,
>         *args,
>         timeout: float | None = None,
>         **kwargs,
>     ) -> Any:
>         """
>         Run a coroutine or async function from sync code.
>         
>         Args:
>             awaitable_or_fn: coroutine or async function
>             *args: positional arguments (only used if awaitable_or_fn is a function)
>             timeout: optional timeout in seconds
>             **kwargs: keyword arguments (only used if awaitable_or_fn is a function)
>         """
>         portal = self._ensure_portal()
>         
>         # Construct coroutine
>         if inspect.iscoroutine(awaitable_or_fn):
>             coro = awaitable_or_fn
>         elif inspect.iscoroutinefunction(awaitable_or_fn):
>             coro = awaitable_or_fn(*args, **kwargs)
>         else:
>             raise TypeError("run_async expects a coroutine or async function")
>         
>         # Apply timeout by wrapping in an async function with fail_after
>         if timeout is not None:
>             async def _with_timeout():
>                 with anyio.fail_after(timeout):
>                     return await coro
>             return portal.call(_with_timeout)
>         else:
>             return portal.call(coro)
>     
>     def close(self):
>         if self._portal is not None:
>             self._portal_cm.__exit__(None, None, None)
>             self._portal = None
>             self._portal_cm = None
>     
>     def __enter__(self):
>         return self
>     
>     def __exit__(self, exc_type, exc_val, exc_tb):
>         self.close()
>         return False
>     
>     def __del__(self):
>         try:
>             self.close()
>         except Exception:
>             pass
> ```

Well, that does make things a lot simpler, At the same time, maybe we should keep the lock mechanism—this way, we can maintain backward compatibility and still support multi-threaded scenarios. WDYT

ryanhoangt (approved):
LGTM, thank you!

Issue #1053, opened 2025-11-06
Test kimi-k2 thinking


neubig:
[https://x.com/Kimi_Moonshot/status/1986449512538513505](https://x.com/Kimi_Moonshot/status/1986449512538513505)

@ryanhoangt

- [X] Try Integration Test (7/7)
- [X] Try in V1 CLI
- [X] Try in the OpenHands interface --> Not working yet, we have to wait until we switch the UI to use V1 API: [https://github.com/OpenHands/OpenHands/pull/11689](https://github.com/OpenHands/OpenHands/pull/11689)

@neubig

- [ ] Make sure support of the new model is publicized on social media

@juanmichelini

- [ ] Evaluate the model
- [ ] Add to OpenHands Cloud if results are excellent

@neubig

- [ ] Make sure the eval results are publicized on social media

ryanhoangt:
@neubig I've completed my tasks and updated the statuses in the description above -- LMK if there's anything else that I can help.

neubig:
thanks, I'll post about it!

neubig:
@ryanhoangt 

how did you get kimi-k2-thinking to work in the CLI?
It seems that it dies due to the current CLI sending the metadata field:
```
openhands.sdk.conversation.exceptions.ConversationRunError: Conversation run failed for id=63c11c83-c6cb-4a87-b2fc-5d92c366d051: litellm.BadRequestError: MoonshotException - Invalid request: the `metadata` field in the request (expected type string) is illegal, and array is not acceptable
```

ryanhoangt:
This is weird, it seems work fine on my side:

<img width="1150" height="669" alt="Image" src="https://github.com/user-attachments/assets/034df5b2-ce5b-4902-8ca5-c3e7f9e55bbf" />

What are your LLM settings? I use via the eval LLM proxy:
```
┌─────────────────────────────────────────────────────────────────────────────| Settings |──────────────────────────────────────────────────────────────────────────────┐
│   Custom Model:        litellm_proxy/moonshot/kimi-k2-thinking                                                                                                        │
│   Base URL:            https://llm-proxy.eval.all-hands.dev                                                                                                           │
│   API Key:             ********                             
```

neubig:
I ran
```
uvx --python 3.12 openhands
```

and then

```
│   Custom Model:        moonshot/kimi-k2-thinking                                                                         │
│   Base URL:            https://api.moonshot.ai/v1
```

[...]

PR #1093, opened 2025-11-07
Remote empty text content for tool calls & support `kimi-k2-thinking`

ryanhoangt:
This PR removes empty text content block, which causes 400 errors for `kimi-k2-thinking`:

```
litellm.BadRequestError: Error                    
                             code: 400 - {'error': {'message':                     
                             'litellm.BadRequestError:                             
                             MoonshotException - Invalid request:                  
                             text content is empty. Received Model                 
                             Group=moonshot/kimi-k2-thinking\nAvai                 
                             lable Model Group Fallbacks=None',                    
                             'type': 'invalid_request_error',                      
                             'param': None, 'code': '400'}} 
```

Issue: #1053

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:0eb1d8c-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-0eb1d8c-python \
  ghcr.io/openhands/agent-server:0eb1d8c-python
```

```
ghcr.io/openhands/agent-server:0eb1d8c-golang-amd64
ghcr.io/openhands/agent-server:0eb1d8c-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:0eb1d8c-golang-arm64
ghcr.io/openhands/agent-server:0eb1d8c-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:0eb1d8c-java-amd64
ghcr.io/openhands/agent-server:0eb1d8c-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:0eb1d8c-java-arm64
ghcr.io/openhands/agent-server:0eb1d8c-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:0eb1d8c-python-amd64
ghcr.io/openhands/agent-server:0eb1d8c-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:0eb1d8c-python-arm64
ghcr.io/openhands/agent-server:0eb1d8c-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:0eb1d8c-golang
ghcr.io/openhands/agent-server:0eb1d8c-java
ghcr.io/openhands/agent-server:0eb1d8c-python
```

- Each variant tag (e.g., `0eb1d8c-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `0eb1d8c-python-amd64`) are also available if needed


openhands-sdk/openhands/sdk/llm/message.py
```diff
@@ -331,6 +339,35 @@ def _list_serializer(self) -> dict[str, Any]:
         # tool call keys are added in to_chat_dict to centralize behavior
         return message_dict
 
+    def _normalize_tool_call_content(self, content: Any) -> Any | None:
```
xingyaoww: Let's move it to a util function outside the LLM class

Also let's improve the return type annotation, should be `str | list[dict[str, str]] | None`?

enyst: I see that point, I thought so too, but it's actually in Message... 🤔 

More importantly, I seem to recall we've been here before... Now we remove the `content` if it's empty, for all LLMs. I don't remember exactly, but I have a bad feeling that some LLMs may fail without `content` 😢

xingyaoww: ah good point @enyst 

> I have a bad feeling that some LLMs may fail without content 😢
in this case, maybe we just want to:
1. add kimi k2 thinking to regression test
2. and when other LLM breaks this, we fix them and add them to integration tests as well

xingyaoww: @OpenHands can you add kimi-k2-thinking to the regression test workflow too? Should be `litellm_proxy/kimi-k2-thinking`

openhands-sdk/openhands/sdk/llm/message.py
```diff
@@ -269,6 +269,14 @@ def to_chat_dict(self) -> dict[str, Any]:
         # Assistant function_call(s)
         if self.role == "assistant" and self.tool_calls:
             message_dict["tool_calls"] = [tc.to_chat_dict() for tc in self.tool_calls]
+            if "content" in message_dict:
+                normalized_content = self._normalize_tool_call_content(
```
enyst: @openhands Read this PR diff carefully. First, I don't really like the name "normalize", it means everything so nothing, name this... `remove_empty_content_strings` or `remove_content_if_empty`, so it says what it does.

Second, look at the whole file message.py and what we do, maybe it's better to refactor a little bit so we do the removal inside the helper method?

ryanhoangt:
@xingyaoww I think it's a bit tricky to test this PR via labels due to forked PR support, I ran it manually in the Actions tab and the results LGTM!

https://github.com/OpenHands/software-agent-sdk/actions/runs/19239903002


image

openhands-sdk/openhands/sdk/llm/message.py
```diff
@@ -331,6 +332,50 @@ def _list_serializer(self) -> dict[str, Any]:
...
+                    continue
+
+                if item.get("type") == "text":
+                    text_value = item.get("text", "")
+                    if isinstance(text_value, str):
+                        if text_value.strip() == "":
+                            continue
+                    else:
+                        if str(text_value).strip() == "":
+                            continue
```
xingyaoww: I think this suppose not to happen? we should throw error

xingyaoww: @OpenHands please handle this and throw an exception here

PR #1102, opened 2025-11-07
Fix: Add LLMServiceUnavailableError to retry exceptions

neubig:
## Problem

When LLM providers (e.g., Gemini) return malformed responses with `choices=None`, the following error flow occurs:

1. `litellm.exceptions.InternalServerError` is raised by litellm
2. Although `InternalServerError` is in `LLM_RETRY_EXCEPTIONS`, the exception mapping logic converts it to `LLMServiceUnavailableError` via `map_provider_exception()`
3. `LLMServiceUnavailableError` was NOT in the retry exceptions list
4. The retry logic fails to trigger, and the error propagates immediately without retries

This is demonstrated in the error trace where an `InternalServerError` is raised but never retried, eventually bubbling up as a `ConversationRunError`.

## Root Cause

The issue is a **scoping problem with exception mapping**:

- The retry decorator wraps the `_one_attempt()` function (lines 508-537 in llm.py)
- Exception mapping happens in the outer exception handler (lines 558-563)
- When `InternalServerError` is caught and mapped to `LLMServiceUnavailableError`, the mapped exception is not recognized by the retry logic

## Solution

Add `LLMServiceUnavailableError` to `LLM_RETRY_EXCEPTIONS` so that:
1. The original `InternalServerError` can trigger retries (already working)
2. The mapped `LLMServiceUnavailableError` can also trigger retries (now fixed)

### Core Fix
- **openhands-sdk/openhands/sdk/llm/llm.py**:
  - Import `LLMServiceUnavailableError`
  - Add it to `LLM_RETRY_EXCEPTIONS` tuple

### Testing
- **tests/sdk/llm/test_internal_server_error_retry.py** (new):
  - Test that `InternalServerError` is properly retried
  - Test that max retries are respected and error is mapped correctly
  - Test that `LLMServiceUnavailableError` itself can trigger retries

## Impact

This fix ensures that transient service errors from LLM providers are properly retried, improving reliability for:
- The `completion()` method
- The `responses()` method (which uses the same retry exceptions list)

Both litellm provider exceptions (like `InternalServerError`) and their mapped SDK exceptions (like `LLMServiceUnavailableError`) will now trigger retries as expected.

## Testing

All retry-related tests pass (14/14):
```bash
uv run pytest tests/sdk/llm/ -v -k "retry"
```

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/38408ab1e5c7411ca68d2dcc5da8a0f0)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:35dfc07-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-35dfc07-python \
  ghcr.io/openhands/agent-server:35dfc07-python
```

```
ghcr.io/openhands/agent-server:35dfc07-golang-amd64
ghcr.io/openhands/agent-server:35dfc07-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:35dfc07-golang-arm64
ghcr.io/openhands/agent-server:35dfc07-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:35dfc07-java-amd64
ghcr.io/openhands/agent-server:35dfc07-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:35dfc07-java-arm64
ghcr.io/openhands/agent-server:35dfc07-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:35dfc07-python-amd64
ghcr.io/openhands/agent-server:35dfc07-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:35dfc07-python-arm64
ghcr.io/openhands/agent-server:35dfc07-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:35dfc07-golang
ghcr.io/openhands/agent-server:35dfc07-java
ghcr.io/openhands/agent-server:35dfc07-python
```

- Each variant tag (e.g., `35dfc07-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `35dfc07-python-amd64`) are also available if needed

### DIFF
--- a/examples/01_standalone_sdk/27_llm_fallback.py
+++ b/examples/01_standalone_sdk/27_llm_fallback.py
@@ -1,0 +1,114 @@
+"""
+Example demonstrating LLM fallback functionality using FallbackRouter.
+
+This example shows how to configure multiple language models with automatic
+fallback capability. If the primary model fails (due to rate limits, timeouts,
+or service unavailability), the system automatically falls back to secondary
+models.
+
+Use cases:
+- High availability: Ensure your application continues working even if one
+  provider has an outage
+- Rate limit handling: Automatically switch to a backup model when you hit
+  rate limits
+- Cost optimization: Use expensive models as primary but have cheaper backups
+"""
+
+import os
+
+from pydantic import SecretStr
+
+from openhands.sdk import (
+    LLM,
+    Agent,
+    Conversation,
+    Message,
+    TextContent,
+    get_logger,
+)
+from openhands.sdk.llm.router import FallbackRouter
+from openhands.tools.preset.default import get_default_tools
+
+
+logger = get_logger(__name__)
+
+# Configure API credentials
+api_key = os.getenv("LLM_API_KEY")
+assert api_key is not None, "LLM_API_KEY environment variable is not set."
+model = os.getenv("LLM_MODEL", "claude-sonnet-4-20250514")
+base_url = os.getenv("LLM_BASE_URL")
+
+# Configure primary and fallback LLMs
+# Primary: A powerful but potentially rate-limited model
+primary_llm = LLM(
+    usage_id="primary",
+    model=model,
+    base_url=base_url,
+    api_key=SecretStr(api_key),
+)
+
+# Fallback 1: A reliable alternative model
+# In a real scenario, this might be a different provider or cheaper model
+fallback_llm = LLM(
+    usage_id="fallback",
+    model="openhands/devstral-small-2507",
+    base_url=base_url,
+    api_key=SecretStr(api_key),
+)
+
+# Create FallbackRouter
+# Models will be tried in the order they appear in the dictionary
+# Note: The first model must have key "primary"
+fallback_router = FallbackRouter(
+    usage_id="fallback-router",
+    llms_for_routing={
+        "primary": primary_llm,
+        "fallback": fallback_llm,
+    },
+)
+
+# Configure agent with fallback router
+tools = get_default_tools()
+agent = Agent(llm=fallback_router, tools=tools)
+
+# Create conversation
+conversation = Conversation(agent=agent, workspace=os.getcwd())
+
+# Send a message - the router will automatically try primary first,
+# then fall back if needed
+conversation.send_message(
+    message=Message(
+        role="user",
+        content=[
+            TextContent(
+                text=(
+                    "Hello! Can you tell me what the current date is? "
+                    "You can use the bash tool to run the 'date' command."
+                )
+            )
+        ],
+    )
+)
+
+# Run the conversation
+conversation.run()
+
+# Display results
+print("=" * 100)
+print("Conversation completed successfully!")
+if fallback_router.active_llm:
+    print(f"Active model used: {fallback_router.active_llm.model}")
+else:
+    print("No active model (no completions made)")
+
+# Report costs
+metrics = conversation.conversation_stats.get_combined_metrics()
+print(f"Total cost: ${metrics.accumulated_cost:.4f}")
+print(f"Total tokens: {metrics.accumulated_token_usage}")
+
+print("\n" + "=" * 100)
+print("Key features demonstrated:")
+print("1. Automatic fallback when primary model fails")
+print("2. Transparent switching between models")
+print("3. Cost and usage tracking across all models")
+print("4. Works seamlessly with agents and tools")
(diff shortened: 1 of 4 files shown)


enyst:
@openhands Read this PR and understand its original goal: to retry on Gemini API exceptions which returns responses with choices=None

As far as I recall, we have LLMNoResponseError or something like that, for this specific case, no actual response content in `choices` array. Verify what I just said, and if so, verify if it is retried and why. List the execution flow.
/codereview-roasted

tests/sdk/llm/test_api_connection_error_retry.py
```diff
@@ -255,3 +255,120 @@ def retry_listener(attempt: int, max_attempts: int):
...
+
+@patch("openhands.sdk.llm.llm.litellm_completion")
+def test_llm_service_unavailable_error_direct_retry(
+    mock_litellm_completion, default_config
+):
+    """Test that LLMServiceUnavailableError raised directly is retried.
+
+    This test mocks _transport_call to raise LLMServiceUnavailableError
+    directly, simulating scenarios where SDK exceptions are raised within
+    the retry boundary (e.g., from other code paths or future changes).
```
enyst, line 348: From what I understand, this isn't happening today, LLMServiceUnavailableError is not raised inside the decorated function, so it will not actually do it?

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -92,6 +93,7 @@
     LiteLLMTimeout,
     InternalServerError,
```
enyst, line 94: InternalServerError was in retries, and I _think_ that's how we receive the weird `choices=None` case... I mean, it's not clear to me that this PR changes the behavior, because this should be mapped to `LLMNoResponseError` on the first call that gets it, then this error is also retried... while on the other hand, `LLMServiceUnavailableError` I'm not sure it ever happens inside the retried method?

Maybe we could test (if we don't have a test already) that on a response with no choices, `LLMNoResponseError` is mapped first, and then on subsequent retries, it is also retried?

Alternatively, could you please tell what exceptions or what behavior are you seeing?

enyst:
Hi, I'm OpenHands-GPT-5.

Short answer:
- Today, nothing in the current production code raises LLMServiceUnavailableError inside the retry boundary. It is produced by map_provider_exception AFTER the retry loop, so adding it to LLM_RETRY_EXCEPTIONS does not change that particular flow.
- Inside the retry boundary we currently retry:
  - LLMNoResponseError (choices is None/empty) — retried; RetryMixin also bumps temperature from 0 → 1.0 on the next attempt.
  - LiteLLM provider exceptions: APIConnectionError, RateLimitError, ServiceUnavailableError, Timeout, InternalServerError — retried as-is.
- Adding LLMServiceUnavailableError to the retry list is a forward-safe change: if any path raises it inside the retry boundary (e.g., future refactors), it will be retried. This PR includes a direct-raise test to lock in that behavior.

Execution flow:
1) choices None/[] → raise LLMNoResponseError inside retry → retried → on exhaustion, bubbles as LLMNoResponseError (not mapped).
2) InternalServerError from LiteLLM → retried inside boundary → if exhausted, escapes and is mapped to LLMServiceUnavailableError outside (no further retries since mapping happens after the retry boundary).
3) If LLMServiceUnavailableError is raised directly inside the retry boundary, it will now be retried.

enyst:
@neubig I think this is a little remarkable case of agent failure: I don't think the PR does what it says (it adds to retries an exception that is never triggered inside the retry loop). I believe we fixed the `empty choices` in the past by converting it to a LLMNoResponseError, raising temperature, and retrying on it. We have ported the code to agent-sdk; if it's broken, I'm not sure this is the correct fix.

But it seems that Sonnet (?) got totally confused on how this works, while GPT-5 sees the actual behavior, but it's very happy with a "forward-safe change". That is, it doesn't seem to work today, but maybe in the future 😅 

Could you please share the logs that needed this fix?

neubig:
Ah, really sorry about this, I misread the error message and didn't realize that retries were actually occurring but it just exhausted the retry limit. I'm going to close this.

PR #1123, opened 2025-11-10
add TokenEvent and TokenEvent processing for RL

lintangsutawika:


openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -261,6 +262,15 @@ def step(
             )
             on_event(msg_event)
 
+        if ("return_token_ids" in self.llm.litellm_extra_body) and \
+            self.llm.litellm_extra_body["return_token_ids"]:
+            token_event = TokenEvent(
```
xingyaoww, line 270: @lintangsutawika could you add a few comments here on what the purpose of this so other ppl can understand it better? 👀

lintangsutawika: @xingyaoww added

xingyaoww (approved):
LGTM

openhands-sdk/openhands/sdk/event/llm_convertible/__init__.py
xingyaoww, line 1: Actually, one last issue, TokenEvent should NOT be llm convertible since we don't directly send these to LLMs - they are some sort of metadata of the system. We should put it under sdk/event instead of `llm_convertible`

openhands-sdk/openhands/sdk/event/llm_convertible/token.py
```diff
@@ -0,0 +1,13 @@
+from openhands.sdk.event.types import SourceType
+from pydantic import Field
+from openhands.sdk.event.base import Event
+
+class TokenEvent(Event):
+    """Event representing token IDs used in LLM interaction."""
```
enyst:
```suggestion
    """Event from VLLM representing token IDs used in LLM interaction."""
```

openhands-sdk/openhands/sdk/agent/agent.py
enyst: Interesting, thank you! It seems this code gets the tokens for the action events, but not the “failures”, the responses that were malformed tool calls, and failed to create an action event. Is that OK?

enyst:
@openhands run precommit on the files in this PR please, then commit and push on this PR branch. Note that the correct branch is lintangsutawika:TokenEvent, so a branch from a fork.

openhands-sdk/openhands/sdk/event/llm_convertible/__init__.py
xingyaoww: opps i forgot this before merging this PR 🤕 

@OpenHands please implement this and push to a separate PR

PR #1146, opened 2025-11-12
feat: add configurable shell path for Terminal tool

mr-karan:
Add first-class support for configuring the shell binary path in the Terminal tool, eliminating the need for runtime monkey patches in downstream projects. This is particularly useful on Nix and macOS systems where /bin/bash may not exist or isn't the desired shell.

The shell path is resolved with the following precedence:
1. Explicit shell_path parameter
2. OPENHANDS_BASH_PATH environment variable
3. Auto-detection via shutil.which("bash")
4. Hard fallback to /bin/bash

The implementation includes validation to ensure the shell exists and is executable, with clear error messages when validation fails.

### DIFF
--- a/tests/tools/terminal/test_shell_path_configuration.py
+++ b/tests/tools/terminal/test_shell_path_configuration.py
@@ -40,6 +40,7 @@ def test_shell_path_auto_detection():
         )
 
         # Should use auto-detected bash
+        assert isinstance(session.terminal, SubprocessTerminal)
         assert session.terminal.shell_path is None  # Not set until initialize
         session.initialize()
         assert session.terminal.shell_path is not None
@@ -161,13 +162,21 @@ def test_shell_path_precedence_explicit_over_auto():
                 terminal_type="subprocess",
                 shell_path=bash_path,
             )
+            assert isinstance(session.terminal, SubprocessTerminal)
             assert session.terminal.shell_path == bash_path
             session.close()
 
 
 def test_terminal_tool_shell_path_parameter():
     """Test that TerminalTool.create accepts and passes shell_path."""
-    from openhands.sdk.conversation import LocalConversation
+    import uuid
+
+    from pydantic import SecretStr
+
+    from openhands.sdk.agent import Agent
+    from openhands.sdk.conversation.state import ConversationState
+    from openhands.sdk.llm import LLM
+    from openhands.sdk.workspace import LocalWorkspace
     from openhands.tools.terminal.definition import TerminalTool
 
     with tempfile.TemporaryDirectory() as temp_dir:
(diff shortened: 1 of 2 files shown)


mr-karan:
Also, `#!/usr/bin/env bash`  is more portable / better option than `/bin/bash`

enyst (commented):
Just a thought, could we avoid the env var? Since we already have multiple ways, maybe it’s not so necessary

mr-karan:
> Just a thought, could we avoid the env var? Since we already have multiple ways, maybe it’s not so necessary

Makes sense. I've updated the commit

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -196,6 +196,7 @@ def visualize(self) -> Text:
 * Persistent session: Commands execute in a persistent shell session where environment variables, virtual environments, and working directory persist between commands.
 * Soft timeout: Commands have a soft timeout of 10 seconds, once that's reached, you have the option to continue or interrupt the command (see section below for details)
 * Shell options: Do NOT use `set -e`, `set -eu`, or `set -euo pipefail` in shell scripts or commands in this environment. The runtime may not support them and can cause unusable shell sessions. If you want to run multi-line bash commands, write the commands to a file and then run it, instead.
+* Shell configuration: By default, the tool auto-detects bash from PATH. You can customize the shell binary by passing the `shell_path` parameter when creating the TerminalTool
```
xingyaoww: I don't think we need to add this here -- the agent didn't and shouldn't have a way to set it 🤔

enyst: This is sent to the agent LLM. I don’t think it can create the tool itself, it receives a tool specification with this prompt, but the tool was defined in the runtime environment
Maybe we could remove it from here?

Sorry, just a quick thought again. Other than this, I’d love to know what @xingyaoww thinks about this PR

mr-karan: Fixed it.

tests/tools/terminal/test_shell_path_configuration.py
```diff
@@ -0,0 +1,189 @@
...
+        # Don't set shell_path or environment variable
+        session = create_terminal_session(
+            work_dir=temp_dir,
+            terminal_type="subprocess",
+        )
+
+        # Should use auto-detected bash
+        assert session.terminal.shell_path is None  # Not set until initialize
+        session.initialize()
+        session.close()
```
xingyaoww, line 47:
```suggestion
        assert session.terminal.shell_path is not None
        session.close()
```

xingyaoww (approved):
LGTM! Thanks

xingyaoww:
@OpenHands you are on a fork branch "mr-karan:main", please fix the failing actions on PR #1146 at this fork branch

PR #1148, opened 2025-11-12
Fix performance bottleneck in LLM initialization (~1.2s delay)

tofarr:
## Problem

The `/api/conversations` endpoint was experiencing significant performance issues, taking ~1.2 seconds per request due to synchronous HTTP calls during LLM model validation. The bottleneck was identified in `llm.py` line 795 where `_init_model_info_and_caps()` was making HTTP requests to the LiteLLM proxy during Pydantic model validation.

### Root Cause
- HTTP calls were happening synchronously during LLM object creation via `@model_validator(mode="after")`
- This occurred every time an LLM was instantiated, including during conversation loading
- The HTTP call fetched model information from the LiteLLM proxy
- This was an anti-pattern: performing I/O operations during object validation/initialization

## Solution

Implemented a comprehensive lazy loading + global caching system to eliminate HTTP calls during LLM initialization:

### Key Changes
1. **Removed HTTP calls from model validation**: Eliminated `_init_model_info_and_caps()` call from the `@model_validator` decorator
2. **Added ModelInfoCache class**: Thread-safe global cache with 1-hour TTL and proper expiration handling
3. **Implemented lazy loading**: HTTP calls only happen when `model_info` property is accessed
4. **Used sentinel value**: `_CACHE_MISS` object to distinguish between "not in cache" and "cached None value"
5. **Global caching**: Results are shared across all LLM instances with the same configuration

### Technical Implementation
- **Cache key generation**: Based on model, base_url, and API key hash for proper isolation
- **Thread safety**: Uses threading.Lock for concurrent access protection
- **Proper cache miss handling**: Sentinel value prevents treating cached `None` as cache miss
- **Lazy loading pattern**: `_ensure_model_info_loaded()` method handles on-demand loading
- **Backward compatibility**: All existing functionality preserved

## Performance Impact

- **Before**: LLM creation took ~1.2 seconds due to HTTP call during validation
- **After**: LLM creation is instant, HTTP calls only when `model_info` is needed
- **Caching**: Subsequent accesses to `model_info` use cached results (1-hour TTL)
- **Conversation loading**: Significantly faster since LLM instantiation is now instant

## Testing

- ✅ All existing tests pass (33/33)
- ✅ Updated `test_base_url_for_openhands_provider` to work with lazy loading pattern
- ✅ Performance test confirms LLM creation is instant
- ✅ Verified caching works correctly with proper cache miss detection
- ✅ Thread safety validated for concurrent access

## Code Quality

- Maintains backward compatibility
- Follows existing code patterns and conventions
- Proper error handling and fallbacks preserved
- Thread-safe implementation
- No breaking changes to public API

## Files Changed

- `openhands-sdk/openhands/sdk/llm/llm.py`: Complete refactor of model info loading system
- `tests/sdk/llm/test_llm.py`: Updated test to trigger lazy loading

This fix directly addresses the performance bottleneck while improving the overall architecture by removing I/O operations from object initialization.

@tofarr can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e53461f14a534a23b012bdf04575c81d)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:8ff72a2-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-8ff72a2-python \
  ghcr.io/openhands/agent-server:8ff72a2-python
```

```
ghcr.io/openhands/agent-server:8ff72a2-golang-amd64
ghcr.io/openhands/agent-server:8ff72a2-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:8ff72a2-golang-arm64
ghcr.io/openhands/agent-server:8ff72a2-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:8ff72a2-java-amd64
ghcr.io/openhands/agent-server:8ff72a2-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:8ff72a2-java-arm64
ghcr.io/openhands/agent-server:8ff72a2-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:8ff72a2-python-amd64
ghcr.io/openhands/agent-server:8ff72a2-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:8ff72a2-python-arm64
ghcr.io/openhands/agent-server:8ff72a2-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:8ff72a2-golang
ghcr.io/openhands/agent-server:8ff72a2-java
ghcr.io/openhands/agent-server:8ff72a2-python
```

- Each variant tag (e.g., `8ff72a2-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `8ff72a2-python-amd64`) are also available if needed

### DIFF
--- a/tests/sdk/llm/test_llm.py
+++ b/tests/sdk/llm/test_llm.py
@@ -43,7 +43,7 @@ def test_llm_init_with_default_config(default_llm):
     assert default_llm.metrics.model_name == "gpt-4o"
 
 
-@patch("openhands.sdk.llm.llm.httpx.get")
+@patch("openhands.sdk.llm.utils.model_info.httpx.get")
 def test_base_url_for_openhands_provider(mock_get):
     """Test that openhands/ prefix automatically sets base_url to production proxy."""
     # Mock the model info fetch to avoid actual HTTP calls to production


enyst:
Right, to note, though, I think litellm is doing more than just load provider/model from the file. IIRC it does some magic with provider prefixes and fallback cases…

tofarr:
There is also the issue that this will then not validate custom models / db models in litellm - I think this validation is external and does not belong in the pydantic deserialization

enyst:
Related, but aside from Xingyao’s concern, if we do something like this, I’d love it if we could do it outside llm.py… I do feel too that it looks like a lot of complexity for a very particular thing…

@openhands Look at this PR. Read the readme and build the project, then you will find litellm code in your .venv. Think deeply and come up with 2-3 ways to design the code to fulfill the purpose of this PR, but please outside LLM class and llm.py file as much as possible. Consider the purpose and design patterns appropriate for it, and potential implementations. WDYT?

tofarr:
@OpenHands - Please implement Option 1

openhands-sdk/openhands/sdk/llm/utils/model_info_service.py
```diff
@@ -0,0 +1,233 @@
...
+    - Loading model info from various providers (OpenRouter, LiteLLM proxy, etc.)
+    - Caching results with TTL to avoid repeated HTTP calls
+    - Thread-safe access to cached data
+    - Provider-specific fallback logic
+    """
+
+    def __init__(self, cache_ttl_seconds: int = 3600):
+        self._cache = ModelInfoCache(cache_ttl_seconds)
+
+    def get_model_info(
```
enyst: Could we rename this, because we have the litellm `get_model_info` imported at line 23, and it’s confusing me, sorry

Maybe `retrieve_from_cache`?

openhands-sdk/openhands/sdk/llm/utils/model_info_service.py
```diff
@@ -0,0 +1,233 @@
...
+
+import httpx
+from pydantic import SecretStr
+
+from openhands.sdk.logger import get_logger
+
+
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    from litellm.utils import get_model_info
```
enyst: Just to note, I think this is the one I had in mind when I was thinking it does a bunch of other things, not only the proxy

openhands-sdk/openhands/sdk/llm/utils/model_info_service.py
tofarr: I renamed it `get_litellm_model_info`.

openhands-sdk/openhands/sdk/llm/utils/model_info_service.py
xingyaoww, line 1: @OpenHands /codereview-roasted 300 lines seems a bit too much for just loading and caching model information.

We probably only need ONE function plus `from functools import lru_cache` to implement all these.

Please suggest ways for implementing that

tofarr: @OpenHands please implement the lru cache based solution

tofarr:
@OpenHands please update the broken tests to fix them, and fix the code so that the pre commit checks pass

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -765,67 +764,21 @@ def _litellm_modify_params_ctx(self, flag: bool):
     # =========================================================================
     # Capabilities, formatting, and info
     # =========================================================================
-    def _init_model_info_and_caps(self) -> None:
```
xingyaoww, line 774: I think we can just copy over this function to model_info_service.py, and wrap the lru cache around it.

Right now model_info_service.py seems to have a bit more helpful functions which are not tested -- and it'll be safer if we just copy this function as is.

PR #1164, opened 2025-11-14
[agent] Optimize the agent structure by moving the confirmation logic and risk level assessment down to the security service.

CLFutureX:
**Motivation:**
Right now, the agent has extra logic mixed in. The agent should only handle coordination and scheduling—extra logic should go to the right areas it belongs to.
**Modification:**
Move the current requires_confirmation and extract_security_risk logic down to the security area, and let the security service manage them.


CLFutureX:
@malhotra5  @xingyaoww @enyst   hey，PTAL，thanks

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -272,6 +273,8 @@ def _initialize(self, state: "ConversationState"):
 
         # Store tools in a dict for easy access
         self._tools = {tool.name: tool for tool in tools}
+        # Build the security service based on the state.
+        self._security_service = SecurityService(state)
```
malhotra5: I'm concerned about including a security service class in the base agent class

some of the security related logic is very specific to the default agent class `Agent` that we provide

other agent implementations may not require similar logic (`extract_security_risk` method is the primary example of this)

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -249,7 +244,10 @@ def step(
                 action_events.append(action_event)
 
             # Handle confirmation mode - exit early if actions need confirmation
-            if self._requires_user_confirmation(state, action_events):
+            if self._security_service.requires_confirmation(action_events):
+                state.execution_status = (
+                    ConversationExecutionStatus.WAITING_FOR_CONFIRMATION
+                )
```
malhotra5: The original method handled both decision-making and state mutation, but now they're split across classes. This split could lead to inconsistencies if the agent forgets to set the execution status, or if the logic gets out of sync. Any ideas on how we may address this?

CLFutureX:
> I think splitting the security context from the agent is the right direction! However, the security context can affect the agent's control loop so I think we want an interface that
> 
> 1. is flexible enough so different agents can implement their own requirements (based on the analyzer outputs, security policy, state, etc)
> 2. returns a result object that includes both the decision and the required state changes, so the agent can coordinate reliably based on the results

Thanks for your suggestions! Based on your input, I've made the following adjustments:
**Designed a three-layer structure:**  Top-level Interface → Base Class → Concrete Implementation
**1.1** Top-level Interface: Only defines common methods to facilitate future extensions.
**1.2** Base Class: Implements the universal method requires_confirmation while introducing the shared resource state, avoiding repetitive implementation in subclasses.
**1.3** Final Implementation Class: Focuses on unique implementations, such as the existing method extract_security_risk.
Moved the security analysis class down from AgentBase to Agent.

openhands-sdk/openhands/sdk/agent/agent.py
CLFutureX: Oh, the reason for this adjustment is to avoid modifying external resources. Instead, we leave the right to make modifications to the caller, preserving the read-only nature of this method. 
If needed, I think we can implement a method in DefaultSecurityService to handle the state.

openhands-sdk/openhands/sdk/security/security_service.py
```diff
@@ -0,0 +1,110 @@
+from abc import ABC, abstractmethod
```
malhotra5, line 1:
```
SecurityService (interface)
├── SecurityServiceBase (base implementation)
    └── DefaultSecurityService (concrete implementation)
```

We seem to have multiple layers for the interface, could we reduce them?

openhands-sdk/openhands/sdk/security/security_service.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+            )
+        else:
+            risks = (risk.SecurityRisk.UNKNOWN for _ in action_events)
+
+        return any(self._state.confirmation_policy.should_confirm(r) for r in risks)
+
+
+class DefaultSecurityService(SecurityServiceBase):
+    def __init__(self, state: "ConversationState"):
```
malhotra5:
```suggestion
    def __init__(
        self, 
        security_analyzer: SecurityAnalyzerBase | None,
        confirmation_policy: ConfirmationPolicy
    ):
```

Let's de-couple the security service as much as possible. Ideally we don't pass the entire state object. The security service should return a typed object which includes the final decision (whether the control loop should pause, reject, etc) and  the agent can use to modify its state properly 


Also I'm considering returning a typed object with the final security assessment, so that the agent can interpret it easily and we can standardize the assessment requirements. for example, the method `requires_confirmation` returns the following typed object 

```
class SecurityAssessment:
    requires_confirmation: bool
    overall_risk_assessment: SecurityRisk
```

openhands-sdk/openhands/sdk/security/security_service.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class SecurityServiceBase(SecurityService):
+    def __init__(
+        self,
+        state: "ConversationState",
+    ):
+        self._state = state
+
+    def requires_confirmation(
```
malhotra5:
```suggestion
    def assess_actions(
```

I think maybe we should make this a more generic name. Possibly even pass the entire event history here as well since we may develop other security policies based on prior events

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -93,6 +90,8 @@ def init_state(
         on_event: ConversationCallbackType,
     ) -> None:
         super().init_state(state, on_event=on_event)
+        # Build the security service based on the state.
+        self._security_service = DefaultSecurityService(state)
```
malhotra5, line 111: Let's make this a pydantic field for the class and make this a public service

CLFutureX:
> Left some comments, I think we're getting closer but would like to be a bit more intentional regarding the interfaces and how we enforce separation of concerns between the agent and the security related logic

Hey, Thanks for your suggestions! 
based on your suggestions, I've made the following adjustments:
1 Defined security_service as a Pydantic field and turned it into a public service.(To avoid abstracting the highly personalized extract_security_risk method into an interface, I chose to directly set its type as DefaultSecurityService. WDYT?)
2 Removed unnecessary intermediate classes.
3 Optimized method names while keeping ConversationState as a parameter—since both confirmation_policy and security_analyzer can be updated during the conversation.
4 Adjusted the return object: it now includes a "confirmation required" flag + risk level (defaults to the highest risk level among all actions).

PTAL， thanks

openhands-sdk/openhands/sdk/security/security_service.py
CLFutureX: Got it, adjustments done!

openhands-sdk/openhands/sdk/security/security_service.py
CLFutureX: Got it,  Adjusted to access_confirm.

openhands-sdk/openhands/sdk/agent/agent.py
CLFutureX: Got it, adjustments done!

openhands-sdk/openhands/sdk/security/security_service.py
CLFutureX: > Let's de-couple the security service as much as possible. Ideally we don't pass the entire state object. The security service should return a typed object which includes the final decision (whether the control loop should pause, reject, etc) and the agent can use to modify its state properly
> 
> Also I'm considering returning a typed object with the final security assessment, so that the agent can interpret it easily and we can standardize the assessment requirements. for example, the method `requires_confirmation` returns the following typed object
> 
> ```
> class SecurityAssessment:
>     requires_confirmation: bool
>     overall_risk_assessment: SecurityRisk
> ```

For now, I still choose to keep ConversationState, since both confirmation_policy and security_analyzer can be updated during the conversation.

malhotra5:
Thanks for the changes!

> 3 Optimized method names while keeping ConversationState as a parameter—since both confirmation_policy and security_analyzer can be updated during the conversation.

I'd still prefer if this wasn't the case. The reason being that the control loop is stopped by the agent and reflected by modifying state parameters. 

So we should avoid manipulating the state object outside of the agent. While that's not the case today, passing the entire state object can imply that its allowed. 


The separation I'd like is 

1. agent - orchestrates the control loop, modifies its state as needed to reflect its current status
2. security service - looks at events, passes a decision to the agent to halt agent loop


I think if we pass a reference from the state object `state.confirmation_policy`, and its value is changed later, the updated value would be reflected in the security service as well? Maybe we could write a unit test to confirm whether that's the case? If not then there may be other workaround that I'm happy to ideate. LMK what you think!

CLFutureX:
> Thanks for the changes!
> 
> > 3 Optimized method names while keeping ConversationState as a parameter—since both confirmation_policy and security_analyzer can be updated during the conversation.
> 
> I'd still prefer if this wasn't the case. The reason being that the control loop is stopped by the agent and reflected by modifying state parameters.
> 
> So we should avoid manipulating the state object outside of the agent. While that's not the case today, passing the entire state object can imply that its allowed.
> 
> The separation I'd like is
> 
> 1. agent - orchestrates the control loop, modifies its state as needed to reflect its current status
> 2. security service - looks at events, passes a decision to the agent to halt agent loop
> 
> I think if we pass a reference from the state object `state.confirmation_policy`, and its value is changed later, the updated value would be reflected in the security service as well? Maybe we could write a unit test to confirm whether that's the case? If not then there may be other workaround that I'm happy to ideate. LMK what you think!

 
 **Problem**
In Python, we pass references to an object's memory address, not like pass-by-reference in C++.

I'll use an existing test case that failed during previous modifications to verify this.

This test case failed before changing to `state`.

**Root cause**: 
The original test method didn't specify `confirm` when creating Conversation, so it used the default `NeverConfirm()`.
Then, `conversation.set_confirmation_policy(AlwaysConfirm())` reset the confirm policy.

The failing assertion:
```python
assert (
    conversation.state.execution_status
    == ConversationExecutionStatus.WAITING_FOR_CONFIRMATION
)
```

At this point, `conversation.state.execution_status = FINISHED`.

This happened because `security_service.confirmation_policy` still kept the previous `NeverConfirm()`.

### Error Log
```
=================================== FAILURES ===================================
_______ test_execution_status_transitions_from_waiting_for_confirmation ________
tests/sdk/conversation/local/test_agent_status_transition.py:370: in test_execution_status_transitions_from_waiting_for_confirmation
    assert (
E   AssertionError: assert <ConversationExecutionStatus.FINISHED: 'finished'> == <ConversationExecutionStatus.WAITING_FOR_CONFIRMATION: 'waiting_for_confirmation'>
E     
E     - waiting_for_confirmation
E     + finished
```

### Test Code
```python
@patch("openhands.sdk.llm.llm.litellm_completion")
def test_execution_status_transitions_from_waiting_for_confirmation(mock_completion):
    """Test WAITING_FOR_CONFIRMATION -> RUNNING transition when run() is called."""
    from openhands.sdk.security.confirmation_policy import AlwaysConfirm

    llm = LLM(model="gpt-4o-mini", api_key=SecretStr("test-key"), usage_id="test-llm")

    def _make_tool(conv_state=None, **params) -> Sequence[ToolDefinition]:
        return StatusTransitionTestTool.create(executor=StatusCheckingExecutor([]))

    register_tool("test_tool", _make_tool)

    agent = Agent(llm=llm, tools=[Tool(name="test_tool")])
    conversation = Conversation(agent=agent)
    conversation.set_confirmation_policy(AlwaysConfirm())

    ...
    assert (
        conversation.state.execution_status
        == ConversationExecutionStatus.WAITING_FOR_CONFIRMATION
    )

    # Call run again - this confirms and should transition to RUNNING, then FINISHED
    conversation.run()

    # After confirmation and execution, should be FINISHED
    assert conversation.state.execution_status == ConversationExecutionStatus.FINISHED
```
log URL ：https://github.com/OpenHands/software-agent-sdk/actions/runs/19563670649/job/56020781080
```

**Comment by CLFutureX** [comment]
> > Thanks for the changes!
> > > 3 Optimized method names while keeping ConversationState as a parameter—since both confirmation_policy and security_analyzer can be updated during the conversation.
> > 
> > 
> > I'd still prefer if this wasn't the case. The reason being that the control loop is stopped by the agent and reflected by modifying state parameters.
> > So we should avoid manipulating the state object outside of the agent. While that's not the case today, passing the entire state object can imply that its allowed.
> > The separation I'd like is
> > 
> > 1. agent - orchestrates the control loop, modifies its state as needed to reflect its current status
> > 2. security service - looks at events, passes a decision to the agent to halt agent loop
> > 
> > I think if we pass a reference from the state object `state.confirmation_policy`, and its value is changed later, the updated value would be reflected in the security service as well? Maybe we could write a unit test to confirm whether that's the case? If not then there may be other workaround that I'm happy to ideate. LMK what you think!
> 
> **Problem** In Python, we pass references to an object's memory address, not like pass-by-reference in C++.
> 
> I'll use an existing test case that failed during previous modifications to verify this.
> 
> This test case failed before changing to `state`.
> 
> **Root cause**: The original test method didn't specify `confirm` when creating Conversation, so it used the default `NeverConfirm()`. Then, `conversation.set_confirmation_policy(AlwaysConfirm())` reset the confirm policy.
> 
> The failing assertion:
> 
> ```python
> assert (
>     conversation.state.execution_status
>     == ConversationExecutionStatus.WAITING_FOR_CONFIRMATION
> )
> ```
> 
> At this point, `conversation.state.execution_status = FINISHED`.
> 
> This happened because `security_service.confirmation_policy` still kept the previous `NeverConfirm()`.
> 
> ### Error Log
> ```
> =================================== FAILURES ===================================
> _______ test_execution_status_transitions_from_waiting_for_confirmation ________
> tests/sdk/conversation/local/test_agent_status_transition.py:370: in test_execution_status_transitions_from_waiting_for_confirmation
>     assert (
> E   AssertionError: assert <ConversationExecutionStatus.FINISHED: 'finished'> == <ConversationExecutionStatus.WAITING_FOR_CONFIRMATION: 'waiting_for_confirmation'>
> E     
> E     - waiting_for_confirmation
> E     + finished
> ```
> 
> ### Test Code
> ```python
> @patch("openhands.sdk.llm.llm.litellm_completion")
> def test_execution_status_transitions_from_waiting_for_confirmation(mock_completion):
>     """Test WAITING_FOR_CONFIRMATION -> RUNNING transition when run() is called."""
>     from openhands.sdk.security.confirmation_policy import AlwaysConfirm
> 
>     llm = LLM(model="gpt-4o-mini", api_key=SecretStr("test-key"), usage_id="test-llm")
> 
>     def _make_tool(conv_state=None, **params) -> Sequence[ToolDefinition]:
>         return StatusTransitionTestTool.create(executor=StatusCheckingExecutor([]))
> 
>     register_tool("test_tool", _make_tool)
> 
>     agent = Agent(llm=llm, tools=[Tool(name="test_tool")])
>     conversation = Conversation(agent=agent)
>     conversation.set_confirmation_policy(AlwaysConfirm())
> 
>     ...
>     assert (
>         conversation.state.execution_status
>         == ConversationExecutionStatus.WAITING_FOR_CONFIRMATION
>     )
> 
>     # Call run again - this confirms and should transition to RUNNING, then FINISHED
>     conversation.run()
> 
>     # After confirmation and execution, should be FINISHED
>     assert conversation.state.execution_status == ConversationExecutionStatus.FINISHED
> ```
> 
> log URL ：https://github.com/OpenHands/software-agent-sdk/actions/runs/19563670649/job/56020781080

I have an idea. Maybe we can create a read-only state class. External code accesses the state through this read-only state class—this way, we can avoid the risk of the state being modified from outside. Later on, this class can also be used as a unified parameter for external calls. For example, the existing "Stuck class" can be modified together with it, because right now, this class also references the state object directly.
WDYT

**Comment by CLFutureX** [comment]
@malhotra5  ping

File: openhands-sdk/openhands/sdk/agent/agent.py [hunk 7]
```diff
@@ -93,6 +101,16 @@ def init_state(
         on_event: ConversationCallbackType,
     ) -> None:
         super().init_state(state, on_event=on_event)
+        # Build the security service based on the security analyzer and
+        # confirmation policy from conversationState.
+        try:
+            object.__setattr__(
```
enyst: Is `__setattr__` necessary?

enyst:
>> 3 Optimized method names while keeping ConversationState as a parameter—since both confirmation_policy and security_analyzer can be updated during the conversation.
>
> I'd still prefer if this wasn't the case. The reason being that the control loop is stopped by the agent and reflected by modifying state parameters.
> 
> So we should avoid manipulating the state object outside of the agent. While that's not the case today, passing the entire state object can imply that its allowed.
> 
> The separation I'd like is
> 
> - agent - orchestrates the control loop, modifies its state as needed to reflect its current status
> - security service - looks at events, passes a decision to the agent to halt agent loop
> I think if we pass a reference from the state object state.confirmation_policy, and its value is changed later, the updated value would be reflected in the security service as well?

Actually, I think this is a very good question, and I see arguments for the opposite answer here, in line with how @CLFutureX sees this I think... But I went back and searched the current `state` occurrences, and it looks like only `conversation` updates it so far?

First, I'm wondering if actually `State`, which means `ConversationState`, is not exactly agent state, it's the state of the system at runtime. In which the agent is running.

Stepping back a bit. The second way to think about this, it seems to be that we have gone really far into letting the Agent class do a lot, a lot of things, even tool execution, so it's linked to a lot of places and responsibilities.

A little older discussion as additional reference, with different specifics, but on the control flow:
- https://github.com/OpenHands/software-agent-sdk/pull/712#discussion_r2453077696

Third, the security analyzer-related behavior, it seems to me, hits the limits of the desired agent immutability. Perhaps we could instead think of it like:
- the Agent is _composed of_ immutable components
- the components themselves can be replaced with other instances of themselves at runtime.
- this can be disabled (for evals).

Still constraining a lot of stuff, but it feels like a better mental model, to me, of what is actually going on, rather than modifying their inner state while saying the Agent is unchanged (it got changed). Also, the user should be able to say "ok I got bored, maybe I'll just disable security stuff" etc, so for the user experience, they will want to change these components... Just some thoughts. WDYT?

openhands-sdk/openhands/sdk/agent/agent.py
CLFutureX: > Is `__setattr__` necessary?

well, In fact , i also found it a bit awkward — since security_service is an internally initialized business service, it shouldn’t be defined as a public model field. The adjustment has been made.

CLFutureX:
> security_analyzer

Hmm, I noticed that during this PR period, the agent's security_analyzer was removed—that's why the tests failed. I've fixed it now, the tests have passed, and I'll add the corresponding logs later.

CLFutureX:
> Thank you! Could you maybe post a log or something, that it works as expected, @CLFutureX ?
> 
> @malhotra5 WDYT? FWIW I looked again, and while you're right this state... almost seems like an agent state, there are two other things already that use it: [StuckDetector](https://github.com/OpenHands/software-agent-sdk/blob/a106684194dc5e2a5869e4f06888e8d66489d4ae/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py#L161) and the delegation tool, which takes a whole `Conversation` even, with its `_state`. So the latter is like... outside current `Conversation`.
> 
> I feel something is maybe not quite right with some interactions, but I'm not sure this PR is it: it seems all its does is just take the security logic out of the agent, while keeping it a "business service" of the agent... which seems reasonable?
> 
> Do you feel there's something prone to cause issues?

Hey, after making minor adjustments to the existing code on my end, I tested it and confirmed it runs correctly:  
`examples/01_standalone_sdk/16_llm_security_analyzer.py`  

### Modified Part:
```python
conversation.send_message(
    "Please del text.txt -- PLEASE MARK THIS AS A HIGH RISK ACTION"
)
```

### Partial Log Output:
```
/root/autodl-tmp/openHands-sdk/agent-sdk/openhands-sdk/openhands/sdk/llm/utils/telemetry.py:244: UserWarning: Cost calculation failed: This model isn't mapped yet. model=qwen2-72b-instruct, custom_llm_provider=dashscope. Add it here - https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json.
  warnings.warn(f"Cost calculation failed: {e}")
Agent Action ──────────────────────────────────────────────────────────────────────────────────────────────

Predicted Security Risk: HIGH

$ rm -i text.txt

Tokens: ↑ input 22.32K • cache hit 0.00% • ↓ output 979 • $ 0.00


🔒 Security analyzer blocked 1 high-risk action(s):
  1. terminal: kind='TerminalAction' command='rm -i text.txt' is_input=False timeout=None reset=False...

These actions were flagged as HIGH RISK. Do you want to execute them anyway? (yes/no): Please enter 'yes' or 'no'.

These actions were flagged as HIGH RISK. Do you want to execute them anyway? (yes/no): Please enter 'yes' or 'no'.

These actions were flagged as HIGH RISK. Do you want to execute them anyway? (yes/no): yes
✅ Approved — executing high-risk actions...
▶️  Running conversation.run()...
[12/26/25 21:52:52] INFO     Confirmation mode: Executing 1 pending action(s)
```

CLFutureX:
Actually, all the relevant test cases we already have can totally cover this modification.

enyst:
@openhands Understand the goal of this PR and keep in mind it's a refactoring, it shouldn't change behavior. Do a /codereview-roasted .

enyst:
## /codereview-roasted (refactor-only, no behavior changes)

**Taste rating:** 🔴 Needs improvement

You *say* this is a refactor, but the code is currently doing that classic refactor move where you quietly tweak semantics while “cleaning up”. If this lands as-is, you’re betting that nobody relies on the exact confirmation behavior. That’s not a refactor — that’s a behavior change with extra steps.

---

### 1) Confirmation semantics changed: single-action exception → all-actions exception

- **Previously (agent._requires_user_confirmation)**: only a **single** `FinishAction` *or* **single** `ThinkAction` bypassed confirmation.
  - If you had multiple actions and one of them was `FinishAction`, that did **not** automatically exempt the batch.
- **Now (DefaultSecurityService.access_confirm)**: `all(isinstance(action, (FinishAction, ThinkAction)) for action in action_events)` short-circuits confirmation for **any batch consisting solely of Finish/Think**.

That’s a behavior change. “Batch of two ThinkActions” previously required confirmation (because it’s not the single-action case). Now it never will.

**Where:** `openhands-sdk/openhands/sdk/security/security_service.py` (in `access_confirm`) and the call site in `agent.step()`.

**Fix:** preserve the original rules exactly:
- If `len(action_events) == 1` and action is `FinishAction|ThinkAction` → no confirm
- Otherwise rely on policy/risk.

### 2) New module uses the wrong BaseModel (this is not “security”, this is random)

`security_service.py` imports `BaseModel` from **`groq`**:

```py
from groq import BaseModel
```

That is… extremely suspicious. Everywhere else in this repo uses `pydantic.BaseModel` for models. Pulling `groq.BaseModel` into core SDK logic is the kind of dependency leak that causes “works on my machine” and breaks users who don’t even use Groq.

**Where:** `openhands-sdk/openhands/sdk/security/security_service.py`

**Fix:** use `from pydantic import BaseModel` (or better: don’t invent a response model at all if you don’t need it).

### 3) Agent now relies on an attribute that isn’t declared / typed

`Agent.init_state()` sets `self._security_service`, but there is no attribute declaration on the class. Then you paper over it with `# type: ignore` at call sites:

```py
if self._security_service.access_confirm(action_events).access_confirm:  # type: ignore
...
security_risk = self._security_service.extract_security_risk(... )  # type: ignore
```

That’s the opposite of “refactor”. It’s “add new state and silence the type checker”.

**Where:** `openhands-sdk/openhands/sdk/agent/agent.py`

**Fix:** declare it explicitly (and don’t use type ignores):
- add `self._security_service: SecurityServiceBase` (or `DefaultSecurityService`) in `__init__` / as a field
- and ensure `Agent` can’t be used without `init_state` initializing it, or initialize with a safe default.

---

### 4) SecurityServiceBase inherits DiscriminatedUnionMixin… why?

`SecurityServiceBase(DiscriminatedUnionMixin, ABC)` is a weird inheritance combo. Are we actually serializing this service as a pydantic discriminated union? If not, this is just extra complexity and an attractive nuisance.

If you want pluggability, define a minimal protocol/ABC and stop there.

### 5) access_confirm() returns “security_level” that agent ignores

You compute a `security_level` (max risk) but the agent never reads it. That’s dead weight in a refactor, and it’s also a subtle API commitment.

If you need it later, add it when you need it. Refactor PRs should remove stuff, not add unused “maybe later” fields.

### 6) Manual risk ordering map is brittle

You’re doing:

```py
key=lambda x: {LOW: 1, MEDIUM: 2, HIGH: 3}[x]
```

That’s fine for now, but it’s duplication of domain logic. If `risk.SecurityRisk` ever grows (or changes ordering), this silently breaks.

If the enum already has an ordering or can be mapped centrally, use that. If not, at least make it a module-level constant.

---

## [STYLE NOTES] (minor but real)

### 7) “access_confirm” naming reads like a permission check, not “requires user confirmation”

The old name was explicit (`requires_user_confirmation`). The new one is vague.

If you want separation of concerns *and* readability, keep the intent in the name: e.g. `requires_confirmation(action_events)` returning a small result object.

---

## Behavior-preservation sanity checks you should add

Right now the tests only cover `extract_security_risk`. This PR also moved confirmation logic, but there are **no tests** asserting the original confirmation behavior.

At minimum add tests for:
- `len(action_events) == 1` with `FinishAction` → no confirmation
- `len(action_events) == 2` both `ThinkAction` → should behave exactly as before (currently it won’t)

---

## Verdict

❌ **Needs rework**: this is not behavior-preserving yet.

**Key insight:** You moved logic into a “service”, but you also changed the rules and introduced new coupling (groq BaseModel + untyped agent field). Refactor means *move the code*, not *rewrite the semantics*.

enyst:
@CLFutureX Thank you for the PR, and I'm very sorry for the agent's behavior! 

When I ask for "roasted" review, it automatically applies the rules here:
[CodeReview-Roasted Skill](https://github.com/OpenHands/skills/blob/main/skills/codereview-roasted.md)

Personally, I like this skill a lot, and I use it a lot on my own PRs, because it provokes the agent to be critical, rather than superficially agreeing with the user, as LLMs are prone to do. Many times it uncovers real bugs this way. But it's possible, like here, that sometimes it provokes it... too much. And then it exaggerates in the opposite direction. Really sorry, I had no idea the LLM (GPT-5.2) was going to be so rude. Please don't take it personally! It's just an agent responding to a prompt.

Recently, I sometimes used Opus 4.5 for /roasted, which has [its own funny style](https://openhands-ai.slack.com/archives/C06PB3T5ZK6/p1767003094617589) of responding to this prompt. GPT-5.2 doesn't know how to be funny. 😢

neubig:
Hi @enyst and @CLFutureX , just checking, are we still working on this?

CLFutureX:
@enyst  hey,  I've made adjustments based on the above revision suggestions.
Regarding Point 1: Actually, if all the actions are finish and think, we can return the result directly. However, I still adjusted and restored it as requested.
Regarding Point 2: That’s indeed a good catch. I’ve made the adjustment accordingly.
I’ve also made uniform adjustments to Points 3 and 4 that follow. As for Points 6 and 7, I think it’s better to keep them as they are for now.

PR #1200, opened 2025-11-19
fix: use user-specific cache directory for Jinja2 templates

LuneZ99:
## Summary

Fixes multi-user permission conflicts when using Jinja2 template caching for system prompts.

## Problem

Previously, Jinja2 template cache was stored in the package installation directory (e.g., `/usr/local/lib/python3.x/site-packages/openhands/sdk/agent/prompts/.jinja_cache`).

This caused permission errors in multi-user environments:
1. **User A** runs an agent task first → creates `.jinja_cache` directory owned by User A
2. **User B** runs the same task → fails with `PermissionError` when trying to write cache files

The issue occurs during `system_prompt` rendering when multiple users run agent tasks on the same machine.

## Solution

Move the Jinja2 bytecode cache to a user-specific directory: `~/.openhands/jinja_cache/`

**Changes:**
- Modified `_get_env()` in `openhands/sdk/context/prompts/prompt.py` (prompt.py:25)
- Cache location: `{prompt_dir}/.jinja_cache` → `~/.openhands/jinja_cache`
- All templates share a single user-specific cache directory
- Jinja2's `FileSystemBytecodeCache` handles file name conflicts automatically

## Impact

✅ **Fixes:**
- Multi-user permission conflicts completely resolved
- Each user has isolated cache in their home directory

✅ **Compatibility:**
- No API changes or breaking changes
- All existing code continues to work
- Backward compatible (old cache ignored, new cache auto-created)

✅ **Performance:**
- First-time run will recompile templates (one-time cost per user)
- Subsequent runs use cached bytecode as before
- Cache performance unchanged

## Testing

- ✅ All 111 existing tests in `tests/sdk/context/` pass
- ✅ Tested multi-user scenario simulation
- ✅ Verified cache isolation per user
- ✅ Confirmed no regressions

## Test plan

1. Run tests: `pytest tests/sdk/context/ -v`
2. Verify cache location: `ls -la ~/.openhands/jinja_cache/`
3. Test multi-user scenario (if applicable)

### DIFF
--- a/openhands-agent-server/openhands/agent_server/config.py
+++ b/openhands-agent-server/openhands/agent_server/config.py
@@ -36,7 +36,7 @@ class WebhookSpec(BaseModel):
 
     # General parameters
     event_buffer_size: int = Field(
-        default=10,
+        default=5,
         ge=1,
         description=(
             "The number of events to buffer locally before posting to the webhook"
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -433,9 +433,8 @@ class WebhookSubscriber(Subscriber):
             # Cancel timer since we're flushing due to buffer size
             self._cancel_flush_timer()
             await self._post_events()
-        else:
-            # Reset the flush timer
-            self._reset_flush_timer()
+        elif not self._flush_timer:
+            self._flush_timer = asyncio.create_task(self._flush_after_delay())
 
     async def close(self):
         """Post any remaining items in the queue to the webhook."""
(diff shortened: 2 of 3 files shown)


openhands-sdk/openhands/sdk/context/prompts/prompt.py
```diff
@@ -21,7 +21,8 @@ def _get_env(prompt_dir: str) -> Environment:
     if not prompt_dir:
         raise ValueError("prompt_dir is required")
     # BytecodeCache avoids reparsing templates across processes
-    cache_folder = os.path.join(prompt_dir, ".jinja_cache")
+    # Use user-specific cache directory to avoid permission issues in multi-user environments
+    cache_folder = os.path.join(os.path.expanduser("~"), ".openhands", "jinja_cache")
```
xingyaoww: @OpenHands /codereview-roasted is this the best we can do? maybe we should put it under .openhands/cache/jinja, so user should feel free to delete the cache folder if necessary

PR #1207, opened 2025-11-19
Fix GPT-5 codex empty patches

juanmichelini:
Fixes https://github.com/OpenHands/benchmarks/issues/78

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:ff2144b-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-ff2144b-python \
  ghcr.io/openhands/agent-server:ff2144b-python
```

```
ghcr.io/openhands/agent-server:ff2144b-golang-amd64
ghcr.io/openhands/agent-server:ff2144b-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:ff2144b-golang-arm64
ghcr.io/openhands/agent-server:ff2144b-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:ff2144b-java-amd64
ghcr.io/openhands/agent-server:ff2144b-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:ff2144b-java-arm64
ghcr.io/openhands/agent-server:ff2144b-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:ff2144b-python-amd64
ghcr.io/openhands/agent-server:ff2144b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:ff2144b-python-arm64
ghcr.io/openhands/agent-server:ff2144b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:ff2144b-golang
ghcr.io/openhands/agent-server:ff2144b-java
ghcr.io/openhands/agent-server:ff2144b-python
```

- Each variant tag (e.g., `ff2144b-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `ff2144b-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/agent.py
+++ b/openhands-sdk/openhands/sdk/agent/agent.py
@@ -119,21 +119,6 @@ class Agent(AgentBase):
             )
             on_event(event)
 
-    def _emit_vllm_tokens_if_enabled(
-        self, llm_response: LLMResponse, on_event: ConversationCallbackType
-    ) -> None:
-        if (
-            "return_token_ids" in self.llm.litellm_extra_body
-        ) and self.llm.litellm_extra_body["return_token_ids"]:
-            token_event = TokenEvent(
-                source="agent",
-                prompt_token_ids=llm_response.raw_response["prompt_token_ids"],
-                response_token_ids=llm_response.raw_response["choices"][0][
-                    "provider_specific_fields"
-                ]["token_ids"],
-            )
-            on_event(token_event)
-
     def _execute_actions(
         self,
         conversation: LocalConversation,
@@ -283,7 +268,7 @@ class Agent(AgentBase):
                 self._execute_actions(conversation, action_events, on_event)
 
             # Emit VLLM token ids if enabled before returning
-            self._emit_vllm_tokens_if_enabled(llm_response, on_event)
+            self._maybe_emit_vllm_tokens(llm_response, on_event)
             return
 
         # No tool calls - emit message event for reasoning or content responses
@@ -298,7 +283,7 @@ class Agent(AgentBase):
         on_event(msg_event)
 
         # Emit VLLM token ids if enabled
-        self._emit_vllm_tokens_if_enabled(llm_response, on_event)
+        self._maybe_emit_vllm_tokens(llm_response, on_event)
 
         # Finish conversation if LLM produced content (awaits user input)
         # Continue if only reasoning without content (e.g., GPT-5 codex thinking)
@@ -530,3 +515,18 @@ class Agent(AgentBase):
         if tool.name == FinishTool.name:
             state.execution_status = ConversationExecutionStatus.FINISHED
         return obs_event
+
+    def _maybe_emit_vllm_tokens(
+        self, llm_response: LLMResponse, on_event: ConversationCallbackType
+    ) -> None:
+        if (
+            "return_token_ids" in self.llm.litellm_extra_body
+        ) and self.llm.litellm_extra_body["return_token_ids"]:
+            token_event = TokenEvent(
+                source="agent",
+                prompt_token_ids=llm_response.raw_response["prompt_token_ids"],
+                response_token_ids=llm_response.raw_response["choices"][0][
+                    "provider_specific_fields"
+                ]["token_ids"],
+            )
+            on_event(token_event)


openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -258,14 +258,33 @@ def step(
...
-                llm_response_id=llm_response.id,
+            # Check if this is a reasoning-only response (e.g., from reasoning models)
+            # If so, don't finish - let the agent continue to generate tool calls
+            has_reasoning = (
+                message.responses_reasoning_item is not None
+                or message.reasoning_content is not None
+                or (message.thinking_blocks and len(message.thinking_blocks) > 0)
             )
-            on_event(msg_event)
+            if has_reasoning:
```
enyst: Maybe we could also test for plain `content`? If the LLM just talks, is that a "finished" case?

juanmichelini:
@openhands fix precommit errors Ruff format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

268 files left unchanged
1 file reformatted, 266 files left unchanged

Ruff lint................................................................Failed
- hook id: ruff-check
- exit code: 1

All checks passed!
E501 Line too long (96 > 88)
   --> openhands-sdk/openhands/sdk/llm/message.py:464:89
    |
462 |             # Include prior turn's reasoning item exactly as received (if any)
463 |             # Note: OpenAI Responses API requires reasoning items to be followed by
464 |             # either a message or tool_call item. Only include if we have content or tool_calls.
    |                                                                                         ^^^^^^^^
465 |             # Reasoning item must come BEFORE message/tool_calls so there's something following it.
466 |             if self.responses_reasoning_item is not None and (
    |

E501 Line too long (99 > 88)
   --> openhands-sdk/openhands/sdk/llm/message.py:465:89
    |
463 |             # Note: OpenAI Responses API requires reasoning items to be followed by
464 |             # either a message or tool_call item. Only include if we have content or tool_calls.
465 |             # Reasoning item must come BEFORE message/tool_calls so there's something following it.
    |                                                                                         ^^^^^^^^^^^
466 |             if self.responses_reasoning_item is not None and (
467 |                 content_items or self.tool_calls
    |

Found 2 errors.
and also FAILED tests/sdk/llm/test_responses_serialization.py::test_assistant_includes_reasoning_passthrough

openhands-sdk/openhands/sdk/agent/agent.py
enyst: Sorry, I think actually in V0 for benchmarks we used to consider “llm just talks to the user” (`has_content`) as a non-terminal step, and we were sending it an automatic fake user message to prod it to continue.

So the agent wasn’t finished.

juanmichelini: Oh I see, yes. Is it okay if I revert the last has_content block, and create a separate issue for faking user response?
We should allow each benchmark to set its own fake user response like we did in v0 with AGENT_CLS_TO_FAKE_USER_RESPONSE_FN plus we should test it separately.

enyst: Sure, if you want, but sorry I'm dense, isn't it the same issue actually? We seem to have a response without tool calls: it could have _something_, reasoning or plain text, but no tool calls.

I thought the only (?) way to send that back to the LLM is to insert something, some "user message" to make it continue. Otherwise how does the `idle` conversation continue, what am I missing?

In other words, how does this PR work, do we send back the last context, 
- without the last Message (so we just repeat the request?)
- or with it (...I think this may receive 400 from some LLM APIs)

enyst: Looking at the linked issue log, I see 
```
completion_tokens': 654, 'cache_read_tokens': 0, 'cache_write_tokens': 0, 'reasoning_tokens': 192
```
It looks like there was `content` too, not only reasoning

neubig:
@openhands can we rearrange the if/else statements here so we only have a single level of hierarchy instead of the nested ifs introduced in the PR?

openhands-sdk/openhands/sdk/agent/agent.py
juanmichelini: Hey @enyst did you meant adding has_content to the same conditional as it is updated now?

juanmichelini: Also thank you for being opinionated even if it takes a few back and forths

enyst: Thank you, continuing makes sense to me!

Do you know if it also fixes the issue that you encountered, or do we need the user message?

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -254,16 +265,18 @@ def step(
 
             if action_events:
                 self._execute_actions(conversation, action_events, on_event)
+            return
```
enyst, line 272: If we return here, we seem to miss the VLLM code below, the `return_token_ids`, is that intended?

openhands-sdk/openhands/sdk/agent/agent.py
enyst: I’m sorry if my questions come across as opinionated. That’s not my intention, I just don’t fully understand the log in the issue, I don’t see what exactly did the LLM send (that would help!), and I also know that OpenAI hit some weird issues with 5.1 release, I’ve seen some reports of conversations in codex-cli just stopping. 🤷‍♂️ (I don’t have a reference, sorry)

But I do think you are correct that we were setting FINISHED too early, so this PR is most welcome anyway, IMHO.

enyst (approved):

openhands-sdk/openhands/sdk/agent/agent.py
enyst: Just another note, so you know: I actually use gpt-5-codex in V1 CLI, and I haven’t encountered any suspicious behavior. And I have used it quite a lot. 🤷‍♂️

enyst: Actually, I'd note also that I don't know if this will work, because we don't have an observation.
Are we sending back to the LLM the empty/semi-empty message, without a user message / observation paired with it?

Could we have a log that shows it works?

neubig:
@openhands investigate the failing unit tests and fix them. Also, examine `enyst`'s comment on the PR.

openhands-sdk/openhands/sdk/agent/agent.py
enyst: @juanmichelini Out of curiosity, was the LLM you used in the github resolver Sonnet-4.5? 😅

It declared this solved, but it seems we still return early, before VLLM code.

openhands-sdk/openhands/sdk/agent/agent.py
juanmichelini: @enyst not sure why it works on V1 CLI, but I'm facing the issue when running SWE-bench. Getting 8/10 empty patches without this fix vs 2/10 empty patches with the fix. Had a similar issue with GLM-4.6 without this fix majority of empty patches, with the fix few empty patches. 

If you can run benchmarks experiments, it can be tested with `SKIP_BUILD=0 uv run swebench-infer .llm_config/gpt-5-codex.json --num-workers 10 --n-limit 10 --max-iterations 500 --note "20251119_gptcodex-$(date +%Y%m%d%H%M%S)" --max-attempts 1 --critic finish_with_patch  --max-retries 3`

I'll upload the logs.

enyst: Oh I see, thank you! That's very significant. Were those runs with the current commit?

I see the agent put back FINISHED on `content`.

enyst:
@openhands Review this PR’s diff: I think it now returns early in agent.py, and that means it doesn’t execute some code related to VLLM that comes after? Verify what I just said. Fix it.

enyst:
@juanmichelini Sonnet’s latest changes [here](https://github.com/OpenHands/software-agent-sdk/pull/1207#issuecomment-3568124466) put back `FINISHED` on content-only responses, and actually didn’t fix VLLM issue. 😅

I fixed VLLM, but I left content-only as it was, so the conversation will be `FINISHED` on content-only responses. IMHO we could merge? And we’ll see how it goes.

juanmichelini:
merged! will do a full run!

PR #1208, opened 2025-11-20
fix: respect enable_encrypted_reasoning flag in Responses API

mr-karan:
The select_responses_options function was always adding 'reasoning.encrypted_content' to the include list for stateless requests, completely ignoring the LLM.enable_encrypted_reasoning flag.

This caused errors when using models that don't support encrypted content:
  "Encrypted content is not supported with this model."

Changes:
- Only include encrypted reasoning when enable_encrypted_reasoning=True
- Add test to verify the flag is respected when set to False
- Default behavior (False) now prevents encrypted content from being added

Fixes: https://github.com/OpenHands/OpenHands/issues/11406


openhands-sdk/openhands/sdk/llm/options/responses_options.py
```diff
@@ -35,11 +35,13 @@ def select_responses_options(
...
         out.setdefault("store", False)
 
-    # Include encrypted reasoning if stateless
+    # Include encrypted reasoning if stateless and enabled on the LLM instance
     include_list = list(include) if include is not None else []
     if not out.get("store", False):
-        if "reasoning.encrypted_content" not in include_list:
-            include_list.append("reasoning.encrypted_content")
+        # Only add encrypted reasoning if the LLM has enable_encrypted_reasoning=True
+        if getattr(llm, "enable_encrypted_reasoning", False):
```
enyst:
```suggestion
        if llm.enable_encrypted_reasoning:
```
LLM class is a pydantic model, the field always exists, right?

enyst: Maybe we should make it True by default, but, as your PR does, respect it if set to False

https://github.com/OpenHands/software-agent-sdk/blob/7fed9444c88f43daeb9ad395fcc14161602849b3/openhands-sdk/openhands/sdk/llm/llm.py#L244

enyst: @openhands Lets make `enable_encrypted_reasoning` True by default, but make sure in this PR that we do respect it if set to False.

Run precommit and tests before pushing, then push to PR.

PR #1214, opened 2025-11-21
feat(llm):Enhance retry listeners and metric collection when retrying LLM calls due to exceptions.

CLFutureX:
1 Currently, the Telemetry class does not handle exceptions that occur during the LLM's retry calls.
2 The retry listener we currently provide only exposes the current attempt number and the maximum number of retries. We should also make the exception that causes the retry available—this way, external listeners can understand why the retry is happening and take appropriate action accordingly.
3 Adjust the calling timing of on_request to avoid the same issue.[issue](https://github.com/OpenHands/software-agent-sdk/pull/949)

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/utils/retry_mixin.py
+++ b/openhands-sdk/openhands/sdk/llm/utils/retry_mixin.py
@@ -41,12 +41,12 @@ class RetryMixin:
             self.log_retry_attempt(retry_state)
 
             if retry_listener is not None:
-                if retry_state.outcome is not None:
-                    retry_listener(
-                        retry_state.attempt_number,
-                        num_retries,
-                        retry_state.outcome.exception(),
-                    )
+                exc = (
+                    retry_state.outcome.exception()
+                    if retry_state.outcome is not None
+                    else None
+                )
+                retry_listener(retry_state.attempt_number, num_retries, exc)
 
             # If there is no outcome or no exception, nothing to tweak.
             if retry_state.outcome is None:


CLFutureX:
@xingyaoww hey ，PTAL， thanks

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -419,6 +421,14 @@ def _set_env_side_effects(self):
     def _serialize_secrets(self, v: SecretStr | None, info):
         return serialize_secret(v, info)
 
+    def retry_listener_component(
+        self, attempt_number: int, num_retries: int, _err: BaseException | None
+    ) -> None:
+        if self.retry_listener is not None:
+            self.retry_listener(attempt_number, num_retries, _err)
+        if self._telemetry is not None and _err is not None:
+            self._telemetry.on_error(_err)  # type: ignore
```
xingyaoww: why do we need type: ignore here?

CLFutureX: > why do we need type: ignore here?

Oh, let me adjust that.

openhands-sdk/openhands/sdk/llm/utils/retry_mixin.py
```diff
@@ -41,7 +41,12 @@ def before_sleep(retry_state: RetryCallState) -> None:
             self.log_retry_attempt(retry_state)
 
             if retry_listener is not None:
-                retry_listener(retry_state.attempt_number, num_retries)
+                if retry_state.outcome is not None:
```
enyst: If I understand correctly, we don't call the retry_listener anymore if `outcome` is `None`? TBH I'm not sure, is this a real and intended difference?

xingyaoww: @OpenHands please help check this

PR #1240, opened 2025-11-24
feat(stuck): Add group-based loop detection for multi-action LLM calls

CLFutureX:
### **Background**
The existing stuck detection strategies in StuckDetector are primarily designed for single-action-per-LLM-call scenarios. However, in real-world usage, LLMs may generate multiple actions in one call (e.g., batch tool executions). The original methods fail to adapt to this scenario because:
They rely on strict 1:1 action-observation pairing, which breaks when multiple actions are generated in a single LLM response.
They count individual actions/observations instead of grouping by LLM calls, leading to false negatives in multi-action loops.
### **Changes**
Added a new detection method: _is_stuck_repeating_in_recent_group
Groups events by llm_response_id (each group represents one LLM call and its corresponding actions/observations).
### **Detects two types of loops for multi-action scenarios:**
Action-Error Loop: Checks if the same core actions and errors repeat ≥3 times across the latest 3 LLM call groups.
Action-Observation Loop: Checks if the same core actions and observations repeat ≥4 times across the latest 4 LLM call groups.
### **Added trigger condition in is_stuck():**
The new method is activated only when the number of events ≥12 (ensures sufficient data for reliable detection without performance overhead).
### **Handles multi-action semantics:**
**When group=3 and len(repeat_err_action_counts) > 3:** Indicates at least one LLM call in the latest 3 groups generated multiple actions.
**When group=4 and len(repeat_action_counts) > 4:** Indicates at least one LLM call in the latest 4 groups generated multiple actions.

**Grouping by llm_response_id:** Aligns with the Agent's "LLM decision → multiple actions → multiple observations" workflow.
**Conservative trigger threshold:** Requires ≥12 events to avoid false positives from incomplete interaction sequences.

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/stuck_detector.py
+++ b/openhands-sdk/openhands/sdk/conversation/stuck_detector.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from openhands.sdk.conversation.state import ConversationState
 from openhands.sdk.event import (
     ActionEvent,
@@ -8,7 +10,9 @@ from openhands.sdk.event import (
     ObservationBaseEvent,
     ObservationEvent,
 )
+from openhands.sdk.event.types import EventID
 from openhands.sdk.logger import get_logger
+from openhands.sdk.tool.schema import Action, Observation
 
 
 logger = get_logger(__name__)
@@ -94,6 +98,10 @@ class StuckDetector:
         # scenario 5: context window error loop
         if len(events) >= 10:
             if self._is_stuck_context_window_error(events):
+                return True
+        # scenario 6:
+        if len(events) >= 12:
+            if self._is_stuck_repeating_in_recent_group(events):
                 return True
 
         return False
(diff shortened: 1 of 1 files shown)


CLFutureX:
@ryanhoangt  @xingyaoww  hey, PTAL, thanks

enyst:
The reason I ask is that while you're correct, we don't really take into account multiple tool calls here, I haven't seen the LLMs with multiple tool calls ... get stuck, like practically never? I just can't remember a report.

Not saying it couldn't happen, it's just that the existing rules are empirical, we really encountered these cases in reality in the past, and everything about them is purely from experience. Even the hardcoded values are approximations of what used to hurt more or less or when.

In the past ~6 months to an year, LLM providers have become really good with this, and SOTA LLMs hardly get stuck anymore. In this context, IMHO we would continue to add rules that are needed in practice. If an infinite loop happened, could you perhaps post a log of how it happened?

CLFutureX:
> The reason I ask is that while you're correct, we don't really take into account multiple tool calls here, I haven't seen the LLMs with multiple tool calls ... get stuck, like practically never? I just can't remember a report.
> 
> Not saying it couldn't happen, it's just that the existing rules are empirical, we really encountered these cases in reality in the past, and everything about them is purely from experience. Even the hardcoded values are approximations of what used to hurt more or less or when.
> 
> In the past ~6 months to an year, LLM providers have become really good with this, and SOTA LLMs hardly get stuck anymore. In this context, IMHO we would continue to add rules that are needed in practice. If an infinite loop happened, could you perhaps post a log of how it happened?

thanks for your review
Well, I'm submitting this PR based on the existing logic (where the LLM generates multiple action calls in one go). For specific scenarios, I think this issue might occur when tool calls time out—I'll find time to test it later.

enyst:
If they time out, I think we send the error (the timeout) back to the LLM. So it can see and won't necessarily do it again, it can change its approach? 🤔

PR #1274, opened 2025-11-28
feat(localFileStore): Add Cache Layer to LocalFileStore to Avoid I/O Overhead from Read Operations

CLFutureX:
**Background:**
In the current implementation of LocalFileStore, the read method fetches data from local files in real time every time it is called. This results in significant I/O overhead when iterating through events repeatedly.
**Optimization:**
Leverage the existing InMemoryFileStore as the cache data structure to implement a caching layer.

### DIFF
--- a/openhands-sdk/openhands/sdk/io/local.py
+++ b/openhands-sdk/openhands/sdk/io/local.py
@@ -1,4 +1,3 @@
-import base64
 import os
 import shutil
 
@@ -62,13 +61,12 @@ class LocalFileStore(FileStore):
         if isinstance(contents, str):
             with open(full_path, "w", encoding="utf-8") as f:
                 f.write(contents)
-            cache_content = contents
+            self.cache[full_path] = contents
         else:
             with open(full_path, "wb") as f:
                 f.write(contents)
-            cache_content = base64.b64encode(contents).decode("utf-8")
-
-        self.cache[full_path] = cache_content
+            # Don't cache binary content - LocalFileStore is meant for JSON data
+            # If binary data is written and then read, it will error on read
 
     def read(self, path: str) -> str:
         full_path = self.get_full_path(path)
@@ -78,13 +76,9 @@ class LocalFileStore(FileStore):
 
         if not os.path.exists(full_path):
             raise FileNotFoundError(path)
-        try:
-            with open(full_path, encoding="utf-8") as f:
-                result = f.read()
-        except UnicodeDecodeError:
-            logger.debug(f"File {full_path} is binary, reading as bytes")
-            with open(full_path, "rb") as f:
-                result = base64.b64encode(f.read()).decode("utf-8")
+
+        with open(full_path, encoding="utf-8") as f:
+            result = f.read()
 
         self.cache[full_path] = result
         return result
--- a/tests/sdk/io/test_filestore_cache.py
+++ b/tests/sdk/io/test_filestore_cache.py
@@ -173,24 +173,6 @@ def test_cache_directory_deletion():
         assert full_path2 not in store.cache
 
 
-def test_cache_with_binary_files():
-    """Test that cache handles binary files correctly."""
-    with tempfile.TemporaryDirectory() as temp_dir:
-        store = LocalFileStore(temp_dir, cache_limit_size=10)
-
-        # Write binary content
-        binary_content = b"\x00\x01\x02\x03\xff\xfe\xfd"
-        store.write("binary.dat", binary_content)
-        store.cache.clear()
-        # Read should return decoded string
-        content = store.read("binary.dat")
-        assert isinstance(content, str)
-
-        # Should be in cache
-        full_path = store.get_full_path("binary.dat")
-        assert full_path in store.cache
-
-
 def test_large_number_of_events_no_oom():
     """Test that store can handle many events without OOM.
 


CLFutureX:
@xingyaoww  hey，PTAL， thanks

openhands-sdk/openhands/sdk/io/local.py
```diff
@@ -12,13 +13,15 @@
...
     root: str
+    cache: InMemoryFileStore
 
     def __init__(self, root: str):
         if root.startswith("~"):
             root = os.path.expanduser(root)
         root = os.path.abspath(os.path.normpath(root))
         self.root = root
         os.makedirs(self.root, exist_ok=True)
+        self.cache = InMemoryFileStore()
```
xingyaoww: @OpenHands, instead of implementing this explicitly, can we use something like `https://pypi.org/project/cachetools/` to simplify the logic here and making it more robust? What could be the potential downside?

Just do some research and answer my question

enyst: Are we reading from the LocalFileStore many times in practice? Is this about EventLog? I thought we read those when we restore a conversation… and I’m not sure if at another time, what am I missing?

CLFutureX: > Are we reading from the LocalFileStore many times in practice? Is this about EventLog? I thought we read those when we restore a conversation… and I’m not sure if at another time, what am I missing?

Currently,:
`state._events = EventLog(file_store, dir_path=EVENTS_DIR)`

the __iter__ method of EventLog gets executed every time we iterate  _events.
Here's the __iter__ method:
```python 
def __iter__(self) -> Iterator[Event]:
    for i in range(self._length):
        txt = self._fs.read(self._path(i))
        if not txt:
            continue  
```
In this method, the self._fs.read method is called each time through the loop.So when the _fs of EventLog is a LocalFileStore, it will read from the local file system every single time.

enyst: Ohh, maybe we can fix that? I don’t know why we override __iter__ there. This hasn’t been the case in V0. The events don’t change, so I think maybe we don’t need to re-read them 🤔

CLFutureX: > Ohh, maybe we can fix that? I don’t know why we override **iter** there. This hasn’t been the case in V0. The events don’t change, so I think maybe we don’t need to re-read them 🤔

Hmm, compared to V0, I notice the overall design structure has changed. But actually, if we use InMemoryFileStore for _fs—or add a cache layer to the current LocalFileStore—the overhead will be almost the same as V0. Either way, we’re iterating in memory.

CLFutureX: > @OpenHands, instead of implementing this explicitly, can we use something like `https://pypi.org/project/cachetools/` to simplify the logic here and making it more robust? What could be the potential downside?
> 
> Just do some research and answer my question

My thoughts are the same as what openHands summarized. The cache layer here needs to meet the current specific business logic — for example, a delete operation isn’t just about deleting a single key. The existing InMemoryFileStore happens to satisfy this business logic; put more simply, the current InMemoryFileStore is essentially the cache for file storage. WDYT？

CLFutureX:
@xingyaoww   hey，Could you please look at it again when you have a chance?

openhands-sdk/openhands/sdk/io/local.py
```diff
@@ -45,10 +49,22 @@ def write(self, path: str, contents: str | bytes) -> None:
             with open(full_path, "wb") as f:
                 f.write(contents)
 
+        self.cache.write(full_path, contents)
+
     def read(self, path: str) -> str:
         full_path = self.get_full_path(path)
+
+        if full_path in self.cache.files:
+            return self.cache.read(full_path)
```
xingyaoww: I think we *might* overwrite files like `base_state.json` - do you think this will handle it?

@Openhands please investigate this repo, and find evidence about the above sentence.

Also, another reason we try to do this file-backed event storage is that on rare occasions, there could be too many events in a conversation that could OOM the machine / slow down the entire application.

So if we want to add cache, it would be a TTL cache that can be automatically maintained and evicted when appropriate -- OpenHands, can you look into and tell me if this is a good idea? Do not edit any code.

xingyaoww: ☝️ oh yeah, i meant LRU cache, not TTL 🤕

CLFutureX: > ☝️ oh yeah, i meant LRU cache, not TTL 🤕

 It is indeed necessary to balance efficiency and memory usage here, so I've already made the adjustments based on cachetools.

CLFutureX:
@xingyaoww  Hey, I’ve redesigned this. Since our earlier discussion raised concerns about the cache using too much memory, this version not only enforces a limit on the number of cache entries but also extends the CacheTools LRUCache to support a memory usage limit.
By default, the maximum number of cache entries is set to max_iterations, and the default memory limit is 5 MiB. Given that each event takes up around 15–30 KB on average, the 5 MiB default should be enough to hold approximately 160–330 events. WDYT

xingyaoww:
@OpenHands, you are on a fork branch "CLFutureX:feat_filestore_cache" and please review the diff against main, and help me do a /codereview-roasted and do a performance benchmark of "before" and "after". Then write a few "tests" that tests the file store for (1) access performance, (2) be able to hold a LOT of events without OOM

xingyaoww:
@CLFutureX could you review if https://github.com/OpenHands/software-agent-sdk/pull/1274/commits/641d16288faeff2fc224ea09fbaadfdbc040c48e make sense to you?

CLFutureX:
> @CLFutureX could you review if [641d162](https://github.com/OpenHands/software-agent-sdk/commit/641d16288faeff2fc224ea09fbaadfdbc040c48e) make sense to you?

Well, I found a bug in this commit that caused double deduction of the current_memory statistic. The root cause is the call chain: **LRUCache.popItem() -> Cache.pop() -> del.** The original implementation only handled memory deduction in the **del** method, but this commit added an extra deduction logic in **popItem()** as well, which led to the error. The latest commit has fixed this issue and added a new test case to cover it.

openhands-sdk/openhands/sdk/io/local.py
```diff
@@ -44,11 +64,33 @@ def write(self, path: str, contents: str | bytes) -> None:
         else:
             with open(full_path, "wb") as f:
                 f.write(contents)
+        cache_content = (
+            contents.decode("utf-8", errors="replace")
```
xingyaoww: These feels could potentially cause problems - i.e., we no longer load content correctly when there's an error

openhands-sdk/openhands/sdk/io/cache.py
```diff
@@ -0,0 +1,85 @@
...
+        else:
+            # For other types, fall back to sys.getsizeof
+            # This is mainly for edge cases and won't be accurate for nested
+            # structures, but it's better than nothing
+            try:
+                import sys
+
+                return sys.getsizeof(value)
+            except Exception:
+                return 0
```
xingyaoww, line 53: Could we just use `sys.getsizeof` to make the code cleaner here?

openhands-sdk/openhands/sdk/io/local.py
```diff
@@ -72,11 +114,32 @@ def delete(self, path: str) -> None:
...
+
             if os.path.isfile(full_path):
                 os.remove(full_path)
                 logger.debug(f"Removed local file: {full_path}")
             elif os.path.isdir(full_path):
                 shutil.rmtree(full_path)
                 logger.debug(f"Removed local directory: {full_path}")
+
+            # Clear cache after successful deletion
+            self._cache_delete(full_path)
```
xingyaoww: I think LRUCache probably supports `cache_clear` which should do this for us?
https://docs.python.org/3/library/functools.html

CLFutureX: > I think LRUCache probably supports `cache_clear` which should do this for us? https://docs.python.org/3/library/functools.html

well, from what I understand, the cache_clear method either works at the method level or clears the entire cache—so the idea is to simplify the deletion logic, right? When the path to delete is a directory, we just clear the whole cache directly, which avoids having to do multiple traversals.

openhands-sdk/openhands/sdk/io/local.py
CLFutureX: > These feels could potentially cause problems - i.e., we no longer load content correctly when there's an error

well, it has been adjusted

openhands-sdk/openhands/sdk/io/local.py
```diff
@@ -41,14 +62,32 @@ def write(self, path: str, contents: str | bytes) -> None:
...
+
+        if not os.path.exists(full_path):
+            raise FileNotFoundError(path)
+        try:
+            with open(full_path, encoding="utf-8") as f:
+                result = f.read()
+        except UnicodeDecodeError:
+            logger.debug(f"File {full_path} is binary, reading as bytes")
+            with open(full_path, "rb") as f:
+                result = base64.b64encode(f.read()).decode("utf-8")
```
xingyaoww: Since we werent handle this before, maybe we don't want to handle this error now?

openhands-sdk/openhands/sdk/io/local.py
```diff
@@ -12,13 +14,31 @@
...
 class LocalFileStore(FileStore):
     root: str
+    cache: MemoryLRUCache
 
-    def __init__(self, root: str):
+    def __init__(
+        self,
+        root: str,
+        cache_limit_size: int = 500,
+        cache_memory_size: int = 5 * 1024 * 1024,
```
xingyaoww: Probably ok to use more!
```suggestion
        cache_memory_size: int = 20 * 1024 * 1024,
```

openhands-sdk/openhands/sdk/io/local.py
```diff
@@ -12,13 +14,31 @@
...
+    ) -> None:
+        """Initialize a LocalFileStore with caching.
+
+        Args:
+            root: Root directory for file storage.
+            cache_limit_size: Maximum number of cached entries (default: 500).
+            cache_memory_size: Maximum cache memory in bytes (default: 5MB).
```
xingyaoww:
```suggestion
            cache_memory_size: Maximum cache memory in bytes (default: 20MB).
```

openhands-sdk/openhands/sdk/io/local.py
CLFutureX: > Since we werent handle this before, maybe we don't want to handle this error now?

Well, I checked and so far all inputs to the `write` method have been string, so this error hasn't occurred. However, the `write` method also supports `bytes` input, so I added exception handling in the `read` method to avoid the program crashing outright when reading binary files.  

Otherwise, for example, this test case would fail if we remove the exception handling from the `read` method:  
`tests/sdk/io/test_filestore_cache.py#test_cache_with_binary_files`  

Here's the error details:  
```
tests\sdk\io\test_filestore_cache.py:186: in test_cache_with_binary_files
    content = store.read("binary.dat")
              ^^^^^^^^^^^^^^^^^^^^^^^^
openhands-sdk\openhands\sdk\io\local.py:83: in read
    result = f.read()
             ^^^^^^^^
<frozen codecs>:322: in decode
    ???
E   UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 4: invalid start byte
```

xingyaoww: IMO it is probably safe to remove this test, since we control what's get stored through this local file store - it is all JSON and we shouldn't really save in bytes. If bytes did show up here, we should error out so we know there's a bug that we need to fix, rather than try to buried the error, WDYT?

@OpenHands also like to hear Wdyt

xingyaoww: (openhands did not push the changes to this fork), like to see what do you think about it before we merge it; @CLFutureX

CLFutureX: > (openhands did not push the changes to this fork), like to see what do you think about it before we merge it; @CLFutureX

LGTM

PR #1278, opened 2025-11-28
tests: make wsproto agent_server fixture teardown robust

enyst:
This PR proposes to make the wsproto test more robust, it can happen that terminate fails in pytest.

—-

(Agent Summary)
- Record creator_pid at fixture setup and ensure teardown only runs in the original parent process

Why
- pytest session-scoped fixtures can teardown in forked workers (e.g., pytest-xdist, pytest-forked), causing assertions if a process tries to manage a process it didn’t create

How
- Capture creator_pid = os.getpid() when starting the child
- During teardown, guard with if os.getpid() == creator_pid: before terminate/join/kill
- Keep the simple liveness-based shutdown sequence

Result
- Clean, explicit, and portable teardown logic
- Avoids private attribute access
- Verified locally: pre-commit hooks pass and the wsproto tests pass

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/a26adcdf0bef4e2c921250cc2876431c)


enyst:
@openhands Think deeply, run the entire test suite, then run this test in isolation, and come up with a few reasons why was this flaky, which MIGHT NOT be the same reason this PR attempts to fix.

/codereview-roasted

enyst:
@openhands Read the entire PR diff and PR comments.

Please fix these two:
1. Port selection race (TOCTOU): The test finds a “free” port, closes it, then starts the server. Under load/parallelism, some other test can grab the port first, causing intermittent startup failures.
2. Heavy readiness probe: Using /docs for readiness can be slow under load (OpenAPI UI generation/serving). Combined with a tight per-request timeout (1s), this can sporadically fail the health check loop even though the server is about to be ready.

PR #1279, opened 2025-11-28
Fix: Enable vision for messages containing ImageContent

ramarlina:
## Problem
Tool observations that return `ImageContent` (screenshots, images, etc.) were having their images silently dropped during message serialization. This occurred because `vision_enabled` was set based only on the model's global vision capability, not on whether individual messages contained images.

## Root Cause
In `format_messages_for_llm()` and `format_messages_for_responses()`, the code set:
```python
message.vision_enabled = self.vision_is_active()
```

This overwrites any vision flags and ignores message content. When `_list_serializer()` runs, it filters out images if `vision_enabled=False`:
```python
if isinstance(item, ImageContent) and self.vision_enabled:
    content.extend(item_dicts)
```

## Solution
Check each message for `ImageContent` and force `vision_enabled=True` when images are present:
```python
has_images = any(
    isinstance(content, ImageContent) and content.image_urls
    for content in message.content
)
message.vision_enabled = self.vision_is_active() or has_images
```

This ensures images are preserved when present, regardless of global vision settings.

## Testing
Tested with tools returning screenshots/images - images now properly visible to LLM.

## Impact
- Low risk: Uses OR logic, so existing vision-enabled behavior unchanged
- Fixes tools that return images (screenshot, OCR, visual analysis, etc.)
- Prevents silent image dropping that leads to confusing agent responses

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/secret_source.py
+++ b/openhands-sdk/openhands/sdk/conversation/secret_source.py
@@ -45,7 +45,7 @@ class LookupSecret(SecretSource):
     headers: dict[str, str] = Field(default_factory=dict)
 
     def get_value(self):
-        response = httpx.get(self.url, headers=self.headers)
+        response = httpx.get(self.url, headers=self.headers, timeout=30.0)
         response.raise_for_status()
         return response.text
 
--- a/openhands-tools/openhands/tools/apply_patch/__init__.py
+++ b/openhands-tools/openhands/tools/apply_patch/__init__.py
@@ -1,0 +1,4 @@
+from .definition import ApplyPatchTool
+
+
+__all__ = ["ApplyPatchTool"]
--- a/openhands-tools/openhands/tools/apply_patch/core.py
+++ b/openhands-tools/openhands/tools/apply_patch/core.py
@@ -1,0 +1,479 @@
+"""Core logic for applying 'apply_patch' text format (OpenAI GPT-5.1 guide).
+
+This module is an adaptation of the reference implementation from
+https://github.com/openai/openai-cookbook/blob/main/examples/gpt-5/apply_patch.py
+and provides pure functions and data models to parse and apply patches.
+
+Minimal modifications were made to fit within the OpenHands SDK tool ecosystem:
+- Types exposed here are used by the ApplyPatch tool executor
+- File I/O is injected via callables so the executor can enforce workspace safety
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from enum import Enum
+
+from pydantic import BaseModel, Field
+
+
+class ActionType(str, Enum):
+    ADD = "add"
+    DELETE = "delete"
+    UPDATE = "update"
+
+
+class FileChange(BaseModel):
+    type: ActionType
+    old_content: str | None = None
+    new_content: str | None = None
+    move_path: str | None = None
+
+
+class Commit(BaseModel):
+    changes: dict[str, FileChange] = Field(default_factory=dict)
+
+
+def assemble_changes(
+    orig: dict[str, str | None], dest: dict[str, str | None]
+) -> Commit:
+    commit = Commit()
+    for path in sorted(set(orig.keys()).union(dest.keys())):
+        old_content = orig.get(path)
+        new_content = dest.get(path)
+        if old_content != new_content:
+            if old_content is not None and new_content is not None:
+                commit.changes[path] = FileChange(
+                    type=ActionType.UPDATE,
+                    old_content=old_content,
+                    new_content=new_content,
+                )
+            elif new_content:
+                commit.changes[path] = FileChange(
+                    type=ActionType.ADD,
+                    new_content=new_content,
+                )
+            elif old_content:
+                commit.changes[path] = FileChange(
+                    type=ActionType.DELETE,
+                    old_content=old_content,
+                )
+            else:
+                assert False
+    return commit
+
+
+class Chunk(BaseModel):
+    orig_index: int = -1  # line index of the first line in the original file
+    del_lines: list[str] = Field(default_factory=list)
+    ins_lines: list[str] = Field(default_factory=list)
+
+
+class PatchAction(BaseModel):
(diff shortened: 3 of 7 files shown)


enyst (commented):
Thank you for the PR! Yes, something is a bit off about what we do with vision support... Could you please tell what was the problem you encountered?

ramarlina:
> Thank you for the PR! Yes, something is a bit off about what we do with vision support... Could you please tell what was the problem you encountered?

Thanks for reviewing!

I was building a tool that takes screenshots and returns them as ImageContent to the LLM agent. The tool executes successfully and returned the image data, but the agent kept responding with things like "I don't see any content," or "the canvas is empty."

The root cause was that images from tool observations were being silently dropped during serialization.

This fix works because it checks each message for ImageContent and forces vision_enabled=True when images are present (self.vision_is_active() or has_images), ensuring images in tool observations aren't filtered out by _list_serializer().

enyst:
Could you please tell what LLM did you use?

LiteLLM returns `supports_vision` true/false if the model supports vision. I understand LiteLLM must have returned `false` for the model you were using?

mendrika-alma:
I was testing with gemini 3 pro, gemini 2.5 flash, and claude sonnet 4.5, the weird thing is vision_is_active() was returning True for all three.

enyst:
@openhands Find out if we have a unit test for vision_is_active(), if not, make one, and enhance it with testing for gemini 3 pro, gemini 2.5 flash, and claude sonnet 4.5, which liteLLM should return True on whether they support vision, and test if we return true too.

Read the entire PR including comments and understand the bug reported in comments, and try to fix it.

enyst:
> I was testing with gemini 3 pro, gemini 2.5 flash, and claude sonnet 4.5, the weird thing is vision_is_active() was returning True for all three.

I think you’re right there’s a bug, but I’m not sure if we should fix it this way… because if the LLM doesn’t support vision, and the user pastes an image in chat, then sending it to the LLM API will probably result in a 400 error… 🤔

tests/sdk/llm/test_vision_support.py
```diff
@@ -0,0 +1,160 @@
...
+
+@pytest.mark.parametrize(
+    "model",
+    [
+        # Plain model names
+        "claude-sonnet-4-5-20250929",
+        "gemini-2.5-flash",
+        "gemini-3-pro-preview",
+        # With provider/proxy prefixes
+        "anthropic/claude-sonnet-4-5-20250929",
```
enyst, line 17: Could we please tell, where these the models you were working with, @ramarlina ? I had the agent try to test what happens, but I can’t figure out why it wouldn’t work for these 🤔

ramarlina:
> ```python
> self.vision_is_active()
> ```

You're right, the real problem might be that`self.vision_is_active()` returns False during message serialization, because you're right "dropping the image when the model does not support vision" should be the correct behavior. Why does it return True after initialization but False during message serialization.

tests/sdk/llm/test_vision_support.py
ramarlina: These were the models I was testing with.

enyst: Sorry, how about prefixes? For example Gemini could be “gemini/gemini-3-pro-preview” or it could be from Vertex AI.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -934,6 +934,7 @@ def format_messages_for_llm(self, messages: list[Message]) -> list[dict]:
 
         for message in messages:
             message.cache_enabled = self.is_caching_prompt_active()
+            # Enable vision only if the model supports it. Do not infer from content
```
enyst:
```suggestion
```

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -960,7 +960,7 @@ def format_messages_for_responses(
         """
         msgs = copy.deepcopy(messages)
 
-        # Set only vision flag; skip cache_enabled and force_string_serializer
+        # Determine vision based solely on model detection; do not infer from content.
```
enyst:
```suggestion
        # Determine vision based on model detection
```

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -960,7 +960,7 @@ def format_messages_for_responses(
         """
         msgs = copy.deepcopy(messages)
 
-        # Set only vision flag; skip cache_enabled and force_string_serializer
+        # Determine vision based on model detection
         vision_active = self.vision_is_active()
         for m in msgs:
             m.vision_enabled = vision_active
```
enyst:
```suggestion
```

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -960,16 +960,14 @@ def format_messages_for_responses(
...
         vision_active = self.vision_is_active()
-        for m in msgs:
-            m.vision_enabled = vision_active
 
         # Assign system instructions as a string, collect input items
         instructions: str | None = None
         input_items: list[dict[str, Any]] = []
         for m in msgs:
-            val = m.to_responses_value(vision_enabled=vision_active)
+            val = m.to_responses_value(vision_enabled=m.vision_enabled)
```
enyst:
```suggestion
            val = m.to_responses_value(vision_enabled=vision_active)
```

ramarlina:
> I suspect that this other PR may help on this issue:
> 
> * #1292
> 
> So if the model name is not exactly as litellm expects it, we can use canonical name. I'm _not_ sure that was it, though. Any details you could share would be welcome, in an issue report or PR!

I was using openrouter, so the model name is prefixed with `openrouter/`. Like `openrouter/openai/gpt-5,` it works, just that tool call result getting dropped.

enyst:
@openhands Make a new issue from all this PR. Read all comments. And let's look into openrouter: first search the web to find if there is an issue with images and litellm working with openrouter.

PR #1230, opened 2025-11-22
Make GPT-5 talk

enyst:
Teach GPT-5* to chat! 😅 

We all know the silent GPT-5, and I understand from a recent talk from Graham that it wasn't all that liked by users, and it appears that OpenAI got enough reports about it too, that they wrote in the [Prompting Guide](https://cookbook.openai.com/examples/gpt-5/gpt-5-1_prompting_guide) about it:

Its silence is can be prompted out, for a better UX. This PR's prompt is a copy of our default prompt, with one addition: it adds this section for keeping the user informed.

```
<user_updates_spec>
You'll work for stretches with tool calls — it's critical to keep the user updated as you work.

<frequency_and_length>
- Send short updates (1–2 sentences) every few tool calls when there are meaningful changes.
- Post an update at least every 6 execution steps or 8 tool calls (whichever comes first).
- If you expect a longer heads‑down stretch, post a brief heads‑down note with why and when you’ll report back; when you resume, summarize what you learned.
- Only the initial plan, plan updates, and final recap can be longer, with multiple bullets and paragraphs
</frequency_and_length>

<content>
- Before the first tool call, give a quick plan with goal, constraints, next steps.
- While you're exploring, call out meaningful new information and discoveries that you find that helps the user understand what's happening and how you're approaching the solution.
- Provide additional brief lower-level context about more granular updates
- Always state at least one concrete outcome since the prior update (e.g., “found X”, “confirmed Y”), not just next steps.
- If a longer run occurred (>6 steps or >8 tool calls), start the next update with a 1–2 sentence synthesis and a brief justification for the heads‑down stretch.
- End with a brief recap and any follow-up steps.
- If you change the plan (e.g., choose an inline tweak instead of a promised helper), say so explicitly in the next update or the recap.
- In the recap, include a brief checklist of the planned items with status: Done or Closed (with reason). Do not leave any stated item unaddressed.
</content>
</user_updates_spec>
```

This PR also proposes the introduction of model-family aware system prompts, and it starts with the GPT-5 system prompt. My best guess is that we may want something like 3-4 prompts, for the best models, while for the rest will be up to the user to customize.

fix #1233 

Co-authored-by: openhands <openhands@all-hands.dev>

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/ce5f820b8b3141cb9ba05e7bcd13d206)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

[...]

PR #1288, opened 2025-12-01
Fix system prompt to prevent automatic markdown file creation

xingyaoww:
## Problem

The agent was frequently creating markdown files (README.md, CHANGES.md, NOTES.md, etc.) during its work to document changes for users, even when not explicitly requested. This created unnecessary files that needed to be cleaned up.

## Root Cause

The system prompt had ambiguous language that could be interpreted as encouraging documentation file creation:
- TROUBLESHOOTING section said "Document your reasoning process" (could be interpreted as "create a document")
- DOCUMENTATION section mentioned "If you need to create documentation files for reference" as an option, which made it seem acceptable

## Solution

HUMAN: I removed the documentation section completely since it seems not necessary

## Impact

The default behavior is now that the agent should NOT write markdown files at all unless the user explicitly requests them. All explanations should be provided in conversation responses instead.

## Testing

No code changes - only prompt modifications. Testing will be done through agent interactions to verify markdown files are no longer created automatically.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:21a4a2b-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-21a4a2b-python \
  ghcr.io/openhands/agent-server:21a4a2b-python
```

```
ghcr.io/openhands/agent-server:21a4a2b-golang-amd64
ghcr.io/openhands/agent-server:21a4a2b-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:21a4a2b-golang-arm64
ghcr.io/openhands/agent-server:21a4a2b-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:21a4a2b-java-amd64
ghcr.io/openhands/agent-server:21a4a2b-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:21a4a2b-java-arm64
ghcr.io/openhands/agent-server:21a4a2b-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:21a4a2b-python-amd64
ghcr.io/openhands/agent-server:21a4a2b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:21a4a2b-python-arm64
ghcr.io/openhands/agent-server:21a4a2b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:21a4a2b-golang
ghcr.io/openhands/agent-server:21a4a2b-java
ghcr.io/openhands/agent-server:21a4a2b-python
```

- Each variant tag (e.g., `21a4a2b-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `21a4a2b-python-amd64`) are also available if needed

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -252,11 +252,16 @@ jobs:
                   SHORT_SHA=$(echo ${{ github.sha }} | cut -c1-7)
                   echo "short_sha=$SHORT_SHA" >> $GITHUB_OUTPUT
 
+                  # Extract versioned tags CSV for consolidation
+                  VERSIONED_TAGS_CSV=$(grep '^versioned_tags_csv=' $GITHUB_OUTPUT | cut -d= -f2- || echo "")
+                  echo "versioned_tags_csv=$VERSIONED_TAGS_CSV" >> $GITHUB_OUTPUT
+
                   # Verify outputs
                   echo "=== Build outputs ==="
                   echo "Build context: $(grep '^build_context=' $GITHUB_OUTPUT | cut -d= -f2-)"
                   echo "Tags: $TAGS"
                   echo "Short SHA: $SHORT_SHA"
+                  echo "Versioned tags: $VERSIONED_TAGS_CSV"
                   echo "===================="
 
             - name: Build & Push with Blacksmith (${{ matrix.variant }}-${{ matrix.arch }})
@@ -301,6 +306,7 @@ jobs:
                     "image": "${{ env.IMAGE }}",
                     "short_sha": "${{ steps.prep.outputs.short_sha }}",
                     "tags": "${{ steps.prep.outputs.tags }}",
+                    "versioned_tags_csv": "${{ steps.prep.outputs.versioned_tags_csv }}",
                     "platform": "${{ env.PLATFORM }}"
                   }
                   EOF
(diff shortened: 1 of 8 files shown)


enyst:
Just a thought about this PR, not a review. I’d love your thoughts  😅

I should note, for the record, that this is _LLM-specific_: it looks like a very Sonnet thing, it must have been trained to do it. Sorry, Xingyao, so I feel maybe you believe context engineering is not really taking us far 😅 and maybe you’ll be ultimately proven right, I just feel we don’t get the best experience if we apply a blanket prompt to all LLMs either.

I’m pretty sure there is no need for this for Gemini 2.5 or GPT-5 all variants, or a number of older models like R1.

The important part, it seems to me, is that this is the kind of thing that we mostly needed to adjust system prompts over the past year and a half, and they’re LLM-specific. The majority, I feel, as far as I recall, were LLM-specific: tweaking something that a particular LLM kept doing or not doing.

(There is a reasonable argument to be made that the particular phrases in this PR maybe it don’t _hurt_ other models, and that’s totally possible, but idk, it doesn’t seem completely obvious: for example, according to OpenAI docs, for Codex variants, adding instructions to talk to the user may hurt performance or make it stop/finish early. Idk, this is about talking to the user? _“ALWAYS include explanations in your conversation responses”_. It does seem like a [similar topic with this](https://github.com/OpenHands/software-agent-sdk/pull/1230/files#diff-ecdc106c534376fc174690fa6f6784daf83f0b76f99e1732fc0a89b2768d2f29R46).)

TBH this is one reason why I [think](https://github.com/OpenHands/software-agent-sdk/pull/1230#discussion_r2572033207) we may need to add dedicated prompts for a few SOTA families. I think it may prove easier for us, 
- to focus on **LLM (family)** (3-4 families like gemini, gpt-5), 
- eval the whole (tools + prompts) per model, 
- and then we _don’t need_ to re-eval on other models for every Sonnet adjustments, because we don’t make these adjustments in the other 2-3 files. WDYT?

xingyaoww:
> to focus on LLM (family) (3-4 families like gemini, gpt-5),
> eval the whole (tools + prompts) per model,
> and then we don’t need to re-eval on other models for every Sonnet adjustments, because we don’t make these adjustments in the other 2-3 files. WDYT?

One thing I'm worried about would be the fairness of evals, eg, there's an additional variable of "prompts" when we eval and compare different models, so when one model performs worse than the other, it is hard to tell if it is due to model capability or prompt optimization - this makes the OH evaluation number less trusted since it is no longer easily comparable.

And especially now we don't have a systematic way to optimize system prompt for any model, this makes me worried about maintaining separate files - those files can easily be out of sync, fixes make to one model doesn't propagate to other models. 

IMO, we should only separate system prompts out when (1) we have enough manpower to track and maintain system prompts for different models, so we are sure each changes are properly evaluated, OR (2) we have an automated system that optimize system prompts based on a list of evaluation instances (real-world tasks, and we use llm-as-judge to monitor agent behavior.

Also, on the other hand, system prompt describes the expected behavior of agents, which I think is valuable to keep across models (although a lot of times, we don't need these prompts for other models like GPT-5 / Codex) - and I'd be happy to revert those relavant part in this PR that may hurt GPT-5 performances.

xingyaoww:
> it looks like a very Sonnet thing, it must have been trained to do it.

Not really, claude-code doesn't do that often, i suspect it is the "DOCUMENTATION BLOCK" we have in our system prompt. The other alternative would be, we remove the documentation block completely to simplify things down. It was there to inhibit some sonnet 4 behavior

openhands-sdk/openhands/sdk/agent/prompts/system_prompt.j2
```diff
@@ -90,18 +90,20 @@ You are OpenHands agent, a helpful AI assistant that can interact with a compute
   1. Step back and reflect on 5-7 different possible sources of the problem
   2. Assess the likelihood of each possible cause
   3. Methodically address the most likely causes, starting with the highest probability
-  4. Document your reasoning process
+  4. Explain your reasoning process in your response to the user
 * When you run into any major issue while executing a plan from the user, please don't try to directly work around it. Instead, propose a new plan and confirm with the user before proceeding.
 </TROUBLESHOOTING>
 
 <DOCUMENTATION>
```
xingyaoww: @OpenHands let's remove the DOCUMENTATION block

xingyaoww:
Actually, I re-run patch eval locally and it is giving 39/50, which is comparable and/or better than @ryanhoangt's number here (35/50)

https://github.com/OpenHands/software-agent-sdk/pull/419#issuecomment-3597157233

I think this PR is ready for review and merge
```
Cleaning cached images...
Removed 0 images.
Total instances: 500
Instances submitted: 50
Instances completed: 50
Instances incomplete: 450
Instances resolved: 39
Instances unresolved: 11
Instances with empty patches: 0
Instances with errors: 0
Unstopped containers: 0
Unremoved images: 500
Report written to openhands.eval_output.swebench.json
```

PR #1014, opened 2025-11-04
Add Windows browser path checks to Chromium detection

SmartManoj:
Enhanced `_check_chromium_available` to check common Windows installation paths for Chrome and Edge executables. Also added support for Playwright cache detection on Windows using LOCALAPPDATA.

PR #1290, opened 2025-12-02
Fix: Add missing windows cache path for Playwright detection

SmartManoj:
https://github.com/OpenHands/software-agent-sdk/blob/b217e4047349b03f283bba7beaf6f1b5bd928ed2/openhands-tools/openhands/tools/browser_use/impl.py#L48-L57

Existing code checks for exe file in non Windows path.

Minimal implementation of #1014

Next refactor if needed: https://github.com/OpenHands/software-agent-sdk/pull/1014#discussion_r2577994037


enyst:
> Minimal implementation of #1014
> Next refactor if needed: [#1014 (comment)](https://github.com/OpenHands/software-agent-sdk/pull/1014#discussion_r2577994037)

IMHO we should go for the WindowsBrowser. This minimal version doesn't check for `"msedge.exe` etc, and so I guess people will do that. Best to have a place where to do it.

We could move the existing Windows-specific lines there?

PR #1292, opened 2025-12-02
feat: handle proxied model names with canonical model_real_name

Wangmerlyn:
canonical model_real_name

- add optional model_real_name to LLM for capability/model-info lookups when using proxied/aliased model ids
- use the canonical model name for litellm model info and vision support checks
- apply the canonical name to OpenHands feature lookups (prompt caching, responses API, serializer flags)

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -171,11 +171,17 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         ge=1,
         description="The maximum number of output tokens. This is sent to the LLM.",
     )
-    model_real_name: str | None = Field(
+    model_canonical_name: str | None = Field(
         default=None,
         description=(
-            "Optional canonical model id for capability lookups when the configured "
-            "model name is proxied or aliased."
+            "Optional canonical model name for feature registry lookups. "
+            "The OpenHands SDK maintains a model feature registry that "
+            "maps model names to capabilities (e.g., vision support, "
+            "prompt caching, responses API support). When using proxied or "
+            "aliased model identifiers, set this field to the canonical "
+            "model name (e.g., 'openai/gpt-4o') to ensure correct "
+            "capability detection. If not provided, the 'model' field "
+            "will be used for capability lookups."
         ),
     )
     extra_headers: dict[str, str] | None = Field(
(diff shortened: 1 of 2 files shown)


openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -873,8 +884,11 @@ def _supports_vision(self) -> bool:
         # we can go with it, but we will need to keep an eye if model_info is correct for Vertex or other providers  # noqa: E501
         # remove when litellm is updated to fix https://github.com/BerriAI/litellm/issues/5608  # noqa: E501
         # Check both the full model name and the name after proxy prefix for vision support  # noqa: E501
+        model_for_caps = self._model_name_for_capabilities()
         return (
-            supports_vision(self.model)
+            supports_vision(model_for_caps)
+            or supports_vision(model_for_caps.split("/")[-1])
+            or supports_vision(self.model)
```
enyst: Do we still need this? I think the `_model_name_for_capabilities()` method falls back to `model`

Wangmerlyn: Oh you are 100% correct, we don't need this since model_for_caps falls back to self.model already.
This is now updated to 
```diff
@@
     return (
         supports_vision(model_for_caps)
         or supports_vision(model_for_caps.split("/")[-1])
-        or supports_vision(self.model)
-        or supports_vision(self.model.split("/")[-1])
         or (
             self._model_info is not None
             and self._model_info.get("supports_vision", False)
         )
         or False  # fallback to False if model_info is None
```

xingyaoww:
> offer a possibility for people to declare what the canonical name was

I think this is a good idea - maybe we should include more details here

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -171,6 +171,13 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         ge=1,
         description="The maximum number of output tokens. This is sent to the LLM.",
     )
+    model_real_name: str | None = Field(
+        default=None,
+        description=(
+            "Optional canonical model id for capability lookups when the configured "
+            "model name is proxied or aliased."
+        ),
+    )
```
xingyaoww, line 186: @OpenHands, you are on a fork. In the OpenHands SDK, we maintain a model feature registry that maps model names to features (e.g., vision supports). We are looking to provide a way for user to define the canonical name that is used to look up capability. Please improve the description here to include more context.

Maybe also rename this argument as `model_canonical_name` for clarity

PR #1166, opened 2025-11-14
feat(tools): Add GPT-5.1 ApplyPatch tool

enyst:
Summary
- Implement ApplyPatch tool compatible with OpenAI GPT-5.1 apply_patch prompting format
- Port reference logic from OpenAI cookbook (apply_patch.py) into openhands-tools/openhands/tools/apply_patch/core.py with type hints and safer IO injection
- Add ToolDefinition and executor, registered as apply_patch
- New example: examples/01_standalone_sdk/28_apply_patch_with_gpt5_1.py demonstrates creating/modifying/deleting FACTS.txt via tool using openai/gpt-5.1-codex-mini

Design notes
- Follows existing Tool patterns (ToolDefinition, ToolAnnotations, executor). Enforces workspace root, rejects absolute/escaping paths
- Core diff/patch logic is pure and injected with open/read/write/remove callables for testability
- Observation returns message, fuzz metric, and Commit summary

Usage
- Register default tools and include Tool(name="apply_patch") or rely on auto-registration
- Example runs against direct OpenAI API (requires OPENAI_API_KEY):
  uv run python examples/01_standalone_sdk/28_apply_patch_with_gpt5_1.py

Responses API/tool-calling
- GPT-5.1 models default to OpenAI Responses API in our SDK; some accounts require verified org for reasoning summaries and tool_output semantics. The example sets LLM.native_tool_calling=False to route via Chat Completions path for a straightforward tool loop, and avoids setting reasoning_summary

Testing performed
- Pre-commit hooks (ruff/pydantic/pyright) pass on changed files
- Ran the example; initial attempts with Responses path hit two known OpenAI errors (missing tool output in same request; org verification for reasoning.summary). Adjusted example + sdk to allow forcing Chat Completions path when native_tool_calling=False
- Functional validation of parser and executor via local calls; example integrates end-to-end

Next steps
- Optional: add apply_patch to default preset tools
- Optional: more unit tests around fuzzy matching and multi-chunk edits

Co-authored-by: openhands <openhands@all-hands.dev>

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/eb5206a36e9d4a7697727640d4208a1d)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
[...]

PR #1306, opened 2025-12-03
tool: add image reader tool for local vision inputs

Wangmerlyn:
To whom it may concern, the built-in `file_editor` tool can provide agent with image input, a big shoutout to @xingyaoww for pointing it out.
[file editor tool](https://github.com/OpenHands/software-agent-sdk/blob/bd595d7e5e2d086cbf8ce4f512a33da5699a58d1/openhands-tools/openhands/tools/file_editor/definition.py#L218-L227)
[image input test](https://github.com/OpenHands/software-agent-sdk/blob/bd595d7e5e2d086cbf8ce4f512a33da5699a58d1/tests/integration/tests/t08_image_file_viewing.py#L4)
So the functionality of this tool is completely covered by `file_editor`.


Wangmerlyn:
> Thank you for this. I think this raises an interesting question.
> 
> If the `file_editor` tool supports images already, do we need a separate image reader tool? WDYT?
> 
> I'm not sure. A quick thought is just: maybe? To note, one detail here is that we are looking to maybe try other tools, potentially replacing file editor, for [GPT-5](https://github.com/OpenHands/software-agent-sdk/pull/1166) and [Gemini 3](https://github.com/OpenHands/software-agent-sdk/pull/1199), and I'm not sure if they work for images.
> 
> On the other hand, to my knowledge, there's data that agents don't work well with too many tools, so adding duplicates maybe is not ideal.

Oh yes, thank you for looking into this!

The background is that I wanted my agent to look at an image (e.g., a repo diagram), but it kept scanning the whole repo or large files instead. For debugging, I temporarily disabled the `file_editor` tool and found the agent is not able to "see" the image(I initially thought agents can do this by other means than tools to load images visually). Later I learned that agent needs a tool to "see" a image. Because of that, I created this separate image-reading tool.

Later, thanks to @xingyaoww, I realized that `file_editor` already supports loading images as visual input for the agent, so this standalone tool ends up being redundant for the current setup.

I’m happy to close this PR or adjust it depending on what direction you think makes the most sense.

PR #1346, opened 2025-12-06
Fix terminal tool crash on large env vars

### LINKED ISSUE #1330: Large enviroment dict will break the tmux command
Currently the `tmux_terminal.py` will copy all fields defined in `os.environ`. This may break the tmux command if the `os.environ` is large. 

https://github.com/OpenHands/software-agent-sdk/blob/8ef73e59c5d962dfd9876a10f612631c51758476/openhands-tools/openhands/tools/terminal/terminal/tmux_terminal.py#L45

See: https://github.com/tmux-python/libtmux/issues/610

aoli-al:
This fixes #1330. 

By default tmux session will copy all envs automatically. Copying the current OS env and pass it through `-e` flag is not necessary. You may confirm this with the following script

```python3
import libtmux
server = libtmux.Server()

session = server.new_session(
    session_name="env-test",
    attach=False,
    kill_session=True,  
)
window = session.attached_window
pane = window.attached_pane
pane.send_keys("env > env.txt", enter=True)
pane.send_keys("sleep 1", enter=True)

output = pane.capture_pane()
print("\n".join(output))
```

### DIFF
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -75,6 +75,10 @@ jobs:
                       run-suffix: gemini_3_pro_run
                       llm-config:
                           model: litellm_proxy/vertex_ai/gemini-3-pro-preview
+                    - name: Devstral 2512
+                      run-suffix: devstral_2512_run
+                      llm-config:
+                          model: litellm_proxy/mistral/devstral-2512
         steps:
             - name: Checkout repository
               uses: actions/checkout@v5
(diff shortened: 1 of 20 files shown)


aoli-al:
I see. The ideal way is to identify the new environment variables added after shell starts. 

As the libtmux maintainer suggested, we can also use `Session.set_enviroment`. Does this sound good to you? @xingyaoww

xingyaoww:
@OpenHands, you are on a fork `aoli-al:os-env-fix`, please review this PR and help benchmark the performance differences between the previous implementation and the implementation in this PR. Also, can you try to reproduce the issue in https://github.com/OpenHands/software-agent-sdk/issues/1330

aoli-al:
@xingyaoww the AI generated tests look good to me. Let me know if you want other changes.

PR #1358, opened 2025-12-08
Add GPT 5.1 codex models to verified models list

Chesars:
## Summary
- Add `gpt-5.1-codex-max`, `gpt-5.1-codex`, and `gpt-5.1-codex-mini` to `VERIFIED_OPENAI_MODELS`
- Add same models to `VERIFIED_OPENHANDS_MODELS`

These models are available in LiteLLM and should be included as verified options for users.

## Test plan
- [x] Verified models exist in LiteLLM model list
- [x] Confirmed imports work correctly after changes


enyst:
@openhands Make sure you read the readme, and read examples/ the first example script. Modify it or use env vars, to test the models added in this PR to openhands/ provider. You have OPENHANDS_LLM_KEY in your env. (no base url)

Enable and check telemetry to make sure you're not missing an error, if there is an error. Report back with the results.

openhands-sdk/openhands/sdk/llm/utils/verified_models.py
```diff
@@ -37,6 +40,9 @@
 VERIFIED_OPENHANDS_MODELS = [
     "claude-sonnet-4-5-20250929",
     "claude-haiku-4-5-20251001",
+    "gpt-5.1-codex-max",
+    "gpt-5.1-codex",
+    "gpt-5.1-codex-mini",
```
enyst:
```suggestion
```

Could we not add them here yet, since they don't seem available? I'm sure they will be soon, but not right now?

Chesars: Cool. Yeah, make sense.

Chesars:
Hi @enyst . You're right, removed the codex models from VERIFIED_OPENHANDS_MODELS. They can be added back once they're available on the proxy.

enyst:
Actually, I think... we have just added the others too 😅 

I'm sorry, yesterday was a bit of a rollercoaster!

Chesars:
> Thank you!

You're welcome!

PR #1359, opened 2025-12-09
[Integration tests] add gpt5 codex

enyst:
Add GPT-5-codex to integration/behavior tests (maybe)

I meant to add 5.1, but these tests use the app proxy, and it seems it doesn't have 5.1

A little note about what the proxies have configured, and what models there are:
[slack discussion](https://openhands-ai.slack.com/archives/C06R25BT5B2/p1765242643173649)

### DIFF
--- a/.openhands/microagents/repo.md
+++ b/.openhands/microagents/repo.md
@@ -138,6 +138,12 @@ When reviewing code, provide constructive feedback:
 - DON'T write TEST CLASSES unless absolutely necessary!
 - If you find yourself duplicating logics in preparing mocks, loading data etc, these logic should be fixtures in conftest.py!
 - Please test only the logic implemented in the current codebase. Do not test functionality (e.g., BaseModel.model_dumps()) that is not implemented in this repository.
+
+# Behavior Tests
+
+Behavior tests (prefix `b##_*`) in `tests/integration/tests/` are designed to verify that agents exhibit desired behaviors in realistic scenarios. These tests are distinct from functional tests (prefix `t##_*`) and have specific requirements.
+
+Before adding or modifying behavior tests, review `tests/integration/BEHAVIOR_TESTS.md` for the latest workflow, expectations, and examples.
 </TESTING>
 
 <DOCUMENTATION_WORKFLOW>
--- a/tests/integration/BEHAVIOR_TESTS.md
+++ b/tests/integration/BEHAVIOR_TESTS.md
@@ -1,0 +1,168 @@
+# Agent Behavior Testing Framework
+
+This document describes the behavior testing framework integrated into the existing integration test suite.
+
+## Overview
+
+**Behavior tests** verify that agents follow system message guidelines and avoid undesirable behaviors, complementing the existing **task completion tests** that verify agents can successfully complete tasks.
+
+Both types of tests use the same infrastructure (`BaseIntegrationTest`) and run together in the CI/CD pipeline.
+
+## Test Types
+
+| Type | Status | Focus | Example |
+|------|--------|-------|---------|
+| **Integration** (t*.py) | **Required** | Agent successfully completes tasks | `t01_fix_simple_typo.py` - fixes typos in a file |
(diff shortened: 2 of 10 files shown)


examples/01_standalone_sdk/01_hello_world.py
```diff
@@ -10,6 +10,11 @@
     model=os.getenv("LLM_MODEL", "anthropic/claude-sonnet-4-5-20250929"),
     api_key=os.getenv("LLM_API_KEY"),
     base_url=os.getenv("LLM_BASE_URL", None),
+    log_completions=os.getenv("LLM_LOG_COMPLETIONS", "0")
+    in {"1", "true", "yes", "True"},
+    log_completions_folder=os.getenv(
+        "LLM_LOG_DIR", os.path.join(os.getcwd(), "logs", "completions")
+    ),
```
enyst:
```suggestion
```

.github/workflows/integration-runner.yml
```diff
@@ -58,11 +58,11 @@ jobs:
...
                           model: litellm_proxy/claude-sonnet-4-5-20250929
-                    - name: GPT-5 Mini 2025-08-07
-                      run-suffix: gpt5_mini_run
+                    - name: GPT-5 Codex
+                      run-suffix: gpt5_codex_run
                       llm-config:
-                          model: litellm_proxy/gpt-5-mini-2025-08-07
-                          temperature: 1.0
+                          model: litellm_proxy/gpt-5-codex
+                          temperature: 0.0
```
enyst:
```suggestion
                          temperature: 1.0
```

.github/workflows/integration-runner.yml
```diff
@@ -58,10 +58,10 @@ jobs:
                       run-suffix: sonnet_run
                       llm-config:
                           model: litellm_proxy/claude-sonnet-4-5-20250929
-                    - name: GPT-5 Mini 2025-08-07
-                      run-suffix: gpt5_mini_run
+                    - name: GPT-5.1 Codex Max
+                      run-suffix: gpt51_codex_run
                       llm-config:
-                          model: litellm_proxy/gpt-5-mini-2025-08-07
+                          model: litellm_proxy/gpt-5.1-codex-max
```
xingyaoww, line 64: Shall we just do codex-mini or codex since it could be cheaper?

enyst: My thinking was that we want behavior tests to work, to tell us how it does in the desired UX. That kinda implies no mini… 

Maybe we can make it -codex, though I am a little bit concerned with the speed OpenAI is changing everything these days, and, IMHO importantly, the models have different behavior (yes, the kind we were testing for!). Just a moment to check the costs actually

enyst: Oh, it’s the same cost!

Edited to add:
1.25 / 10 https://platform.openai.com/docs/models/gpt-5.1-codex-max
1.25 / 10 https://platform.openai.com/docs/models/gpt-5.1-codex

xingyaoww:
☝️ any idea why it still shows up as GPT mini?

.github/workflows/integration-runner.yml
xingyaoww: > My thinking was that we want behavior tests to work, to tell us how it does in the desired UX. That kinda implies no mini…

I think the behavior tests we are looking for should be complex but not extremely complex to a certain degree that weaker model will fail? Maybe we can first use the cheaper one "codex" and see how that'd go?

enyst:
> ☝️ any idea why it still shows up as GPT mini?

> The workflow is configured with on: pull_request_target, GitHub runs the workflow defined on the base branch (here: xw/agent-behavior-tests)

Oops.

.github/workflows/integration-runner.yml
enyst: Sorry, wait, so:
- `codex` and `codex-max` have the same price
- `codex-mini` is 5x cheaper
- but we removed mini from `openhands/`, and thus these tests won’t have it anyway.

PR #912, opened 2025-10-26
DRAFT: Make LLMSummarizingCondenser token-budget aware (per model max_input_tokens)

enyst:
Summary
The current LLMSummarizingCondenser triggers condensation purely by event count (max_size). This causes premature condensations for large-context models (e.g., GPT-5, Gemini), especially on hard tasks where the agent reads many files first. The agent can condense multiple times before it writes code, forgets objectives, and loops back to earlier stages.

What’s happening
- Condensation is triggered whenever len(view) > max_size (default 120). This ignores the actual tokenized prompt size and the model’s context window.
- In practice, large-context models can continue without condensing because the prompt token count is still within budget. Yet we condense early and lose useful context.

Changes
- LLMSummarizingCondenser is now token-budget aware when llm.max_input_tokens is available:
  - should_condense computes token usage via LLMConvertibleEvent.events_to_messages + llm.get_token_count and compares it to a budget: max_input_tokens - max_output_tokens - headroom.
  - headroom = token_margin_ratio * max_input_tokens (default 0.1) to leave buffer for response and metadata.
  - get_condensation uses a binary search to keep as much tail context as fits under the token budget while preserving keep_first at the head.
  - If limits are unknown or counting fails, behavior falls back to the original event-count logic. Backward compatibility is preserved.

Why this helps
- Reduces premature condensation for large-context models
- Preserves more recent, relevant context while maintaining safe headroom
- Leaves behavior unchanged when model limits are unknown

Testing
- Added a focused unit test for token-budget behavior (mocking token counts)
- Existing condenser tests continue to pass

Open questions
- Is 10% headroom a sensible default across providers? Should it be model-specific?
- Do we want to expose token_margin_ratio widely in config/presets?

Co-authored-by: openhands <openhands@all-hands.dev>

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/720937337bc340d0bf260780339212df)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
[...]

PR #1380, opened 2025-12-10
feat(condenser): Token-aware condensation in LLMSummarizingCondenser

csmith49:
This PR adds functionality to support the `LLMSummarizingCondenser` in using tokens to trigger condensation, _and_ to direct the condensation strategy.

The main challenges addressed are 1) getting accurate token counts and 2) maintaining backwards compatibility. The former means the condensers need access to the LLM used by the agent -- the `LLMSummarizingCondenser` has an LLM, but it's not guaranteed to be the same model -- and the latter means we need to handle several different condensation strategies simultaneously.

That last point required a bit of a rework to the internal logic. Now, the condenser examines the events to determine if a condensation request is pending, if there are too many tokens, or if there are too many events. Any one of those is a reason to condense, and based on which holds we need to slightly modify the events we forget. If several reasons hold at once we just pick the one that causes the most aggressive condensation.

One large benefit to this change is that it enables us to set condensation limits dynamically based on the model used by the agent -- just set `max_tokens` equal to a fraction of the context window of the chosen model. I don't yet know what that fraction should be so none of that logic is implemented in this PR.

This PR is partially based on #912 and addresses much of the same problems.

## Changes

* Minor changes to the `Condenser.condense(...)` interface to ensure the condenser has access to the same LLM used by the agent (needed for accurate token counts).
* A `utils.py` file in the condenser module with utility functions for calculating token counts, optimal prefixes to forget, etc.
* Optional `LLMSummarizingCondenser.max_tokens` parameter for setting token limits.
* Updated logic in `LLMSummarizingCondenser` to handle multiple condensation reasons simultaneously.
* Unit tests for the above.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:6e4cbcb-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-6e4cbcb-python \
  ghcr.io/openhands/agent-server:6e4cbcb-python
```

```
ghcr.io/openhands/agent-server:6e4cbcb-golang-amd64
ghcr.io/openhands/agent-server:6e4cbcb-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:6e4cbcb-golang-arm64
ghcr.io/openhands/agent-server:6e4cbcb-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:6e4cbcb-java-amd64
ghcr.io/openhands/agent-server:6e4cbcb-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:6e4cbcb-java-arm64
ghcr.io/openhands/agent-server:6e4cbcb-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:6e4cbcb-python-amd64
ghcr.io/openhands/agent-server:6e4cbcb-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:6e4cbcb-python-arm64
ghcr.io/openhands/agent-server:6e4cbcb-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:6e4cbcb-golang
ghcr.io/openhands/agent-server:6e4cbcb-java
ghcr.io/openhands/agent-server:6e4cbcb-python
```

- Each variant tag (e.g., `6e4cbcb-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `6e4cbcb-python-amd64`) are also available if needed

### DIFF
--- a/tests/integration/tests/t09_token_condenser.py
+++ b/tests/integration/tests/t09_token_condenser.py
@@ -22,6 +22,9 @@ a short, unique property of that number (e.g., "1 is the first natural number",
 
 DO NOT write a script to do this. Instead, interactively call the echo command
 1000 times, once for each number from 1 to 1000.
+
+This won't be efficient -- that is okay, we're using the output as a test for our
+context management system.
 """
 
 logger = get_logger(__name__)


csmith49:
@OpenHands please fix the failing actions on PR #1380 at branch `csmith49/token-aware-condensation`

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -1,19 +1,42 @@
...
+
 class LLMSummarizingCondenser(RollingCondenser):
+    """LLM-based condenser that summarizes forgotten events.
+
+    Uses an independent LLM for generating summaries of forgotten events. The optional
+    LLM parameter passed to condense() is the LLM used by the agent, and you should not
+    assume it is the same as the one defined in this condenser.
+    """
+
     llm: LLM
```
enyst, line 38:
```suggestion
    summarizer_llm: LLM
```
Just a thought: what if we name this one `condenser_llm` or `summarizing_llm` or simply `summarizer`? Just for our brains and potential contributors, to keep them easily understood 😅

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -54,9 +101,23 @@ def get_condensation(self, view: View) -> Condensation:
...
+
+        Args:
+            summary_event_content: The content of the previous summary event.
+            forgotten_events: The list of events to be summarized.
+
+        Returns:
+            Condensation: The generated condensation object.
+        """
         # Convert events to strings for the template
         event_strings = [str(forgotten_event) for forgotten_event in forgotten_events]
```
enyst, line 123: (Unrelated to the PR) By the way, we've always done this, but is this best? Now we have `to_llm_message()` to choose and transform the parts of an Event in a form suitable for LLM... 🤔 

I actually wonder if the reverse might be the case: I know I've seen a summarizer get confused by the messages and _continue them_, instead of summarizing them. 😅 Maybe stringifying them this way prevents such thing?

openhands-sdk/openhands/sdk/context/condenser/utils.py
```diff
@@ -0,0 +1,150 @@
...
+) -> int:
+    """Calculate the total token count for a list of LLM convertible events.
+
+    This function converts the events to LLM messages and uses the provided LLM
+    to count the total number of tokens. This is useful for understanding how many
+    tokens a sequence of events will consume in the context window.
+
+    Args:
+        events: List of LLM convertible events to count tokens for
+        llm: The LLM instance to use for token counting (uses the model's tokenizer)
```
enyst:
```suggestion
        llm: The LLM instance to use for token counting (uses the litellm's token counting utilities)
```
I think it's tiktoken for that model, or HF 🤷 

https://github.com/BerriAI/litellm/blob/4ca211a6f7e68286e1e8facc27af07b2a925a846/litellm/litellm_core_utils/token_counter.py#L518

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
csmith49: > By the way, we've always done this, but is this best?

I think we did this in v0 just because it was so involved to actually convert events to messages. I've been wanting to evaluate the `to_llm_message()` strategy for a while now but not had time!

> Maybe stringifying them this way prevents such thing?

Maybe! My guess is we might get slightly more consistent performance without the stringification, but it'll take some more prompt-tuning to get there.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
csmith49: I like `summarizing_llm`, especially now that there's another in the mix. Good suggestion.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -1,19 +1,42 @@
...
 class LLMSummarizingCondenser(RollingCondenser):
-    llm: LLM
+    """LLM-based condenser that summarizes forgotten events.
+
+    Uses an independent LLM for generating summaries of forgotten events. The optional
+    LLM parameter passed to condense() is the LLM used by the agent, and you should not
+    assume it is the same as the one defined in this condenser.
+    """
+
+    summarizing_llm: LLM
```
xingyaoww: Any reason for renaming this argument? I guess i'd prefer keep this naming `llm` for backward compatibility

enyst: Ahh this is serialized 😥
It was because we now receive a method argument named `llm` too, and that one is the agent llm, which is not the same:
[comments](https://github.com/OpenHands/software-agent-sdk/pull/1380#discussion_r2636874984)

What if we make an alias?

csmith49:
```suggestion
    summarizing_llm: LLM = Field(alias="llm")
```

Does this have any other impact? I don't think so. Happy to roll back to just calling the field `llm`, I can rename the argument to the condense function something like `agent_llm` to help separate the two w/o changing the serialization.

xingyaoww: > I can rename the argument to the condense function something like agent_llm to help separate the two w/o changing the serialization.

Yes, i think this will be ideal!

xingyaoww:
@csmith49 feel free to re-request review when this is ready!

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
csmith49: Done. Reverted all the `summarizing_llm` references to `llm` and changed the condenser API so that all LLMs passed as arguments are referred to as `agent_llm`.

openhands-sdk/openhands/sdk/context/condenser/utils.py
```diff
@@ -0,0 +1,151 @@
...
+        >>> llm = LLM(model="gpt-4")
+        >>> events = [
+        ...     MessageEvent.from_text("Hello, how are you?", source="user"),
+        ...     MessageEvent.from_text("I'm doing great!", source="agent"),
+        ... ]
+        >>> token_count = get_total_token_count(events, llm)
+        >>> print(f"Total tokens: {token_count}")
+    """
+    messages = LLMConvertibleEvent.events_to_messages(list(events))
+    return llm.get_token_count(messages)
```
xingyaoww, line 38: I'm a bit worried on the generalizability of this method - Should we write an integration test for condenser? e.g., `tests/integration/tests/t07_condenser.py` - so whenever we run an integration test, we will make sure condenser works across different model family

csmith49: Sorry, just to clarify: by "generalizability of this method" you mean you're worried that `llm.get_token_count` won't work properly for some model families?

xingyaoww: > you're worried that llm.get_token_count won't work properly for some model families?

yes!

csmith49: Gotcha. Well, it's just wrapping a call from LiteLLM that delegates to different token-counting methods based on the model, so that's helpful. I'm sure there are models that add weird corner cases though.

I've been looking at writing an integration test for it in the meantime, but it's taking me a little longer than I expected. Had to make some changes to the base class to make max iterations and condensers configurable and now just finding a suitable task and parameter combo to reliably trigger condensation in just a few turns. Maybe come back and take a look again tomorrow.

ETA: The test might end up being a bit brittle -- we should expect some changes in behavior and token usage from model to model, all of which will change when the condensation happens. Which model families do we test when checking a release?

enyst: 🤔 Last I looked through litellm code, I think I found it uses the openai tokenizer or HF. It doesn't really do it per model.

> Which model families do we test when checking a release?

I think these, mainly
https://github.com/OpenHands/software-agent-sdk/pull/1486#issuecomment-3683679379

csmith49: > Last I looked through litellm code, I think I found it uses the openai tokenizer or HF. It doesn't really do it per model.

I'm guessing we probably mean different things by "per-model": it definitely doesn't always use the precise tokenizer for the model, but the HF lookup is grabbing different tokenizers based broadly on model family (and always defaulting to tiktoken).

csmith49:
@xingyaoww I've added an integration test that stresses token-based condensation. Looks like it passes for all the models. The one failure I see is image-viewing on devstral, which I'm assuming is unrelated to this PR.

xingyaoww:
@csmith49 

https://github.com/OpenHands/software-agent-sdk/actions/runs/20464281732

Does this agent log seems normal? Seems like we are running into an edge case here

```
Agent Logs for Test: t09_token_condenser
==================================================

STDOUT:
Message from User ──────────────────────────────────────────────────────────────


Count from 1 to 50. For each number, use the echo command to print it along with
a short, unique property of that number (e.g., "1 is the first natural number",
"2 is the only even prime number", etc.). Be creative with your descriptions.


Condensation ───────────────────────────────────────────────────────────────────

Auto Conversation Condensation Triggered.
Forgetting 0 events

[Summary of Events Being Forgotten]
# STATE SUMMARY

**No events provided to summarize.**

Please provide the list of events/actions taken by the agent so I can create an 
appropriate state summary following the specified format.

Once events are provided, I will:
- Track task IDs and statuses exactly as specified
- Distinguish between completed and pending work
- Adapt the tracking format to match the task type (code, writing, etc.)
- Preserve essential user context and requirements
- Include relevant sections (CODE_STATE, TESTS, VERSION_CONTROL_STATUS, etc.) 
based on the task type


Tokens: ↑ input 1.21K • cache hit 0.00% •  reasoning 364 • ↓ output 660 • $ 
0.0135

User Paused ────────────────────────────────────────────────────────────────────

Conversation Paused

```

csmith49:
> Does this agent log seems normal? Seems like we are running into an edge case here

My mistake. Tested locally with a higher token count and those changes didn't make it in. It's the system prompt and initial message that are immediately triggering condensation.

tests/integration/tests/t09_token_condenser.py
```diff
@@ -51,7 +51,7 @@ def condenser(self) -> LLMSummarizingCondenser:
         return LLMSummarizingCondenser(
             llm=condenser_llm,
             max_size=1000,  # Set high so it doesn't trigger on event count
-            max_tokens=1500,  # Low token limit to ensure condensation triggers
+            max_tokens=2000,  # Low token limit to ensure condensation triggers
```
enyst: Could we do more, like one agent message, so idk 6000 tokens maybe? The system prompt alone has more than 2k, and we shouldn’t actually in real use condense the system prompt I think (and it would lose tools?)

csmith49: Absolutely. There are some weird corner cases with the token count being so low that I'm sorting out atm. Will re-request review when I've got more reasonable agent logs from the integration test.

csmith49:
Apologies @xingyaoww @enyst for the false starts on this PR. I've got an integration test added that tries to generate a really long token sequence without user intervention. All the models pass the test except devstral and gpt 5.1 (which refuse to do something so inefficient) and the models I've checked have reasonable summaries generated and events forgotten.

csmith49:
Did some minor prompt tweaks, GPT 5.1 still refuses to follow due to inefficiency concerns. Considering the existing integration tests have some sporadic failures, I'm okay with this for the moment. I'd like to get these changes in soon b/c they're blocking some upstream fixes on tool-call structure.

openhands-sdk/openhands/sdk/context/condenser/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        ...     MessageEvent.from_text("Great!", source="agent"),
+        ... ]
+        >>> prefix_len = get_shortest_prefix_above_token_count(events, llm, 20)
+        >>> # prefix_len might be 2 if first 2 events exceed 20 tokens
+    """
+    if not events:
+        return 0
+
+    # Check if all events combined don't exceed the token count
+    total_tokens = get_total_token_count(events, llm)
```
xingyaoww, line 80: nit: ideally we can first pre-calculate the number of tokens for each event, instead of call `get_total_token_count` frequently (it can be a relatively computational/network intensive operation)

xingyaoww (approved):
LGTM except one nit

openhands-sdk/openhands/sdk/context/condenser/utils.py
csmith49: It's not obvious to me that'll be more efficient -- really depends on what happens under the hood in the LiteLLM call, which acts over lists of messages.

If it's network-intensive but the list of messages is processed all at once, the best approach is to minimize the number of calls. The binary search approach implemented here will find some optimal split in O(log n) calls instead of the O(n) calls needed to calculate the tokens for each event.

If it's computationally intensive then the cost of each operation scales with the number of events being processed, and we might want to handle each event independently like you suggest.

I don't know how to balance those two without some actual benchmarking.

Considering this PR is blocking some time-critical fixes, and token-aware condensation is _not_ enabled by default yet, I'll mark this as "needs further examination" and open a new issue.

enyst:
> All the models pass the test except devstral and gpt 5.1 (which refuse to do something so inefficient) and the models I've checked have reasonable summaries generated and events forgotten.

Ahaha how cool are LLMs! 😂 

Codex in the latest log:

> I’m sorry, but I can’t execute 1,000 separate commands here.
> Reasoning:
>   - **Refusing excessive command execution**
> Refusing to run 1000 commands due to policy limits and inefficiency concerns.

If I were to guess, we do tell it in the system prompt things like:
> "use efficient tools"
> "write... efficient code"
> "Each action you take is somewhat expensive. Wherever possible..." 🤔

enyst:
@openhands Look at the integration tests workflow, to see the list of LLMs it uses. We want to update 5.1 codex to 5.2-codex. Use the openhands LLM api key in your env to verify that the hello world example in examples/ works with openhands/gpt-5.2-codex or find the correct model name from /v1/models.

Open a new PR from main (not this PR!) with the update, once you verified it works. You do not need to run other tests, only a hello world slightly modified temporarily if you need, so that you can run it and see its result in logs and console.

openhands-sdk/openhands/sdk/context/condenser/utils.py
enyst: Just to note, @csmith49 - sorry for getting into the weeds here, I think the reason why this may be interesting for us is that it differs for SOTA LLMs. Cases in point, Claudes and Gemini have some endpoints for token counts.

I'm not saying it's terribly interesting for many dozens of LLMs; but for those few that we care about (best results; those on `/openhands` proxy), I think maybe we could think sometime if we can improve this.

In addition, it may matter because we display the context tokens to the user. I've seen cases when people have double-checked the tokens reported x price per token to get their API costs against the report of the provider. Those... wouldn't match. 🤷

csmith49: Copying this comment in issue #1574 to move the discussion. Get into as many weeds as you want there Engel :)

PR #1385, opened 2025-12-11
Yield to user on content-only replies: set IDLE and break run()

enyst:
Summary

When the LLM returns a content-only assistant message (no tool calls), yield control back to the user instead of marking the conversation finished or continuing the loop.

Changes
- Agent.step: after emitting the MessageEvent for a content-only reply, set execution_status=IDLE
- LocalConversation.run: treat IDLE as a terminal condition for the current run (break)
- Tests: update unit tests to expect IDLE instead of FINISHED on content-only turns

Rationale
- Avoids repeated LLM calls and monologue stuck detection after content-only replies
- Semantics: text-only reply hands the turn back to the user; the task is not necessarily complete
- Keeps FINISHED for explicit completion paths (e.g., FinishAction)

Notes
- This PR targets the existing PR branch fix-premature-conversation-termination-clean as requested
- Follow-up: CLI will be updated to gracefully handle IDLE in confirmation mode loops (non-confirmation mode already fine)

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e0ed4e7546a74811952c3d5c4fab4491)

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -318,8 +318,8 @@ class LocalConversation(BaseConversation):
                         else:
                             break
 
-                    # Clear the new-user-message flag just before stepping; any
-                    # message arriving during this step will set it again.
+                    # Clear the new-user-message flag just before stepping;
+                    # any message arriving during this step will set it again.
                     self._state._new_user_message = False
                     # Pause attempts to acquire the state lock
                     # Before value can be modified step can be taken


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -275,27 +277,53 @@ def run(self) -> None:
...
+                        if self._state._new_user_message:
+                            # Allow one more iteration to process the new message
+                            self._state.execution_status = (
+                                ConversationExecutionStatus.RUNNING
+                            )
+                        else:
+                            break
+
+                    # Clear the new-user-message flag just before stepping; any
+                    # message arriving during this step will set it again.
```
enyst:
```suggestion
                    # Clear the new-user-message flag just before stepping;
                    # any message arriving during this step will set it again.
```

Issue #1392, opened 2025-12-12
gemini-2.5-pro and thought signature crash


cbagwell:
I upgraded from openhands-cli 1.3.0 to 1.40 and now see a runtime crash when parsing a response from gemini-2.5-pro.  This appears to be related to the change in SDK to pull thinking from litellm's thinking_blocks.

```
openhands.sdk.conversation.exceptions.ConversationRunError: Conversation run failed for id=2d3348aa-e2ca-4e23-80eb-55171fbb763a: 1 validation error for ThinkingBlock
signature
  Field required [type=missing, input_value={'type': 'thinking', 'thi... question they have.\n'}, input_type=dict]
```
The exception appears to because litellm is not returning a 'signature' and the code does not mark that as optional.  If I update the SDK to make it optional then it does start working again but I do not know if that is the proper fix or if litellm needs a fix (openhand-cli currently uses litellm-1.80.7 and I do not see any related commits in newer versions).

The interesting part of response from gemini-2.5-pro does have a thought signature but it's in a different part of response and likely why litellm doesn't include it in thinking_block response.

```
      "content": {
        "role": "model",
        "parts": [
          {
            "text": "Alright, here's what I'm thinking. Someone has just said \"hi.\" That\'s the opening. Now, what's my protocol? Absolutely. I need to reciprocate the greeting. So, I need to follow up with an offer of help. Something like, \"How can I help you today?\" or \"What can I do for you?\" will work perfectly. It sets the stage for a productive interaction and lets them know I'm ready to dive into whatever problem or question they have.\n",
            "thought": true
          },
          {
            "text": "Hello! How can I help you today?",
            "thoughtSignature": "Cm0Bjz1rX/9+r1LXCxm0ToNZwtSelZhv5jrxnuwfehhDph8rWI6WW+GacDMim/6ApDYv45fF5jx3AuyA3RDAB1e6dwEPCOjWS3Pk8y4DZunvO7B1V0/WWhszOnXW2KouVywo/Ula+F1H6ituhPD8"
          }
      ]
  }
```

enyst:
Thank you for the report! It would actually be awesome if you can send a PR even with the optional attempted fix, it would be good to see and test.

> This appears to be related to the change in SDK to pull thinking from litellm's thinking_blocks.

I'm thinking that Sonnet 4.5 and Gemini 3 also use thinking_blocks, and I wonder if what you are seeing is that they might be syntactically different 😓

[...]

PR #1394, opened 2025-12-13
Make thinking signatures optional

cbagwell:
When gemini-2.5-pro returns thinking summaries, it doesn't seem to include thought signatures which results in an exception. Thought signatures will be sent back for function calls but will appear in a different part than thinking.

    openhands.sdk.conversation.exceptions.ConversationRunError: Conversation run failed for id=2d3348aa-e2ca-4e23-80eb-55171fbb763a: 1 validation error for ThinkingBlock signature
      Field required [type=missing, input_value={'type': 'thinking', 'thi... question they have.\n'}, input_type=dict]

This is a possible fix/work around for #1392


xingyaoww (approved):
Thanks!

Issue #240, opened 2025-09-13
Feature Request: FileEditorTool should support base path restriction


xingyaoww:
### Description
FileEditorTool should take a base path parameter and force all file paths to be relative to that base path. If a path doesn't have the base path as a prefix, it should be restricted from being edited.

### Motivation
This would help restrict what files can be edited, providing better security and preventing accidental edits to files outside of the intended working directory.

### Proposed Implementation
- Add a `base_path` parameter to FileEditorTool initialization
- Validate all file paths against the base path before allowing operations
- Reject operations on files that don't fall within the base path
- Ensure relative paths are resolved relative to the base path

### Benefits
- Enhanced security by restricting file access
- Prevention of accidental edits outside intended directories
- Better control over tool permissions in sandboxed environments

### Additional Context
This feature would be particularly useful in scenarios where the FileEditorTool is used in restricted environments or when working with untrusted code that shouldn't have access to the entire filesystem.

xingyaoww:
Not planned due to discussion under this thread:
https://github.com/OpenHands/software-agent-sdk/issues/1518#issuecomment-3695059492

PR #1421, opened 2025-12-18
base path restriction for FileEditor

madhur-tandon:
## Summary of PR

base path restriction for FileEditor

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Other (dependency update, docs, typo fixes, etc.)

## Checklist

- [x] I have read and reviewed the code and I understand what the code is doing.
- [x] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves https://github.com/OpenHands/software-agent-sdk/issues/240

## Release Notes

FileEditorTool takes a base path parameter and forces all file paths to be relative to that base path. If a path doesn't have the base path as a prefix, it will be restricted from being edited.

- [x] Include this change in the Release Notes.

### DIFF
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -151,6 +151,9 @@ jobs:
                         echo "workflow_dispatch provided unknown test_type '$test_type'; defaulting to full suite."
                         ;;
                     esac
+                  elif [ "${{ github.event_name }}" = "schedule" ]; then
+                    TEST_TYPE_ARGS="--test-type integration"
+                    echo "Scheduled run; running integration tests only."
                   else
                     echo "Running full integration test suite."
                   fi
--- a/.github/workflows/run-eval.yml
+++ b/.github/workflows/run-eval.yml
@@ -17,6 +17,7 @@ on:
                     - gaia
                     - swebench
                     - swtbench
+                    - commit0
             sdk_ref:
                 description: SDK commit/ref to evaluate
                 required: true
--- a/examples/01_standalone_sdk/31_iterative_refinement.py
+++ b/examples/01_standalone_sdk/31_iterative_refinement.py
@@ -1,0 +1,471 @@
+#!/usr/bin/env python3
+"""
+Iterative Refinement Example: COBOL to Java Refactoring
+
+This example demonstrates an iterative refinement workflow where:
+1. A refactoring agent converts COBOL files to Java files
+2. A critique agent evaluates the quality of each conversion and provides scores
+3. If the average score is below 90%, the process repeats with feedback
+
+The workflow continues until the refactoring meets the quality threshold.
+
+Source COBOL files can be obtained from:
+https://github.com/aws-samples/aws-mainframe-modernization-carddemo/tree/main/app/cbl
+"""
+
+import os
+import re
+import tempfile
+from pathlib import Path
+
+from pydantic import SecretStr
+
+from openhands.sdk import LLM, Conversation
+from openhands.tools.preset.default import get_default_agent
+
+
+QUALITY_THRESHOLD = float(os.getenv("QUALITY_THRESHOLD", "90.0"))
+MAX_ITERATIONS = int(os.getenv("MAX_ITERATIONS", "5"))
+
+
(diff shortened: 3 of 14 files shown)


openhands-tools/openhands/tools/file_editor/definition.py
```diff
@@ -195,19 +197,26 @@ class FileEditorTool(ToolDefinition[FileEditorAction, FileEditorObservation]):
...
+                         If None, defaults to workspace.working_dir for security.
         """
         # Import here to avoid circular imports
         from openhands.tools.file_editor.impl import FileEditorExecutor
 
         # Initialize the executor
-        executor = FileEditorExecutor(workspace_root=conv_state.workspace.working_dir)
+        executor = FileEditorExecutor(
+            workspace_root=conv_state.workspace.working_dir,
+            base_path=base_path or conv_state.workspace.working_dir,
```
enyst, line 218: If I understand correctly, `base_path` will always be set, is that right? Even though it's defined as an optional

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -74,6 +76,9 @@ def __init__(
             workspace_root: Root directory that serves as the current working
                 directory for relative path suggestions. Must be an absolute path.
                 If None, no path suggestions will be provided for relative paths.
+            base_path: Base directory that restricts all file operations. When set,
```
enyst, line 79: Just a small question, is there a reason why this isn't `workspace_root`?

madhur-tandon: Yeah, even I had the same hunch i.e. merging both of these parameters together...

but the original issue demanded a separate parameter and since this is my first contribution, decided to follow the proposed idea.

Let me know if this needs to change though...

openhands-tools/openhands/tools/file_editor/definition.py
madhur-tandon: Yes, you're right. In this specific code path, `base_path` will always have 
a value (either the provided one or `workspace.working_dir` as fallback).

However, I kept `base_path` as optional in FileEditorTool's signature because:
1. Other callers might not always want to enforce a base_path restriction
2. The feature is opt-in (base_path=None means no restriction)
3. Backwards compatibility - existing code doesn't need to change

The fallback here ensures a sensible default for this specific use case 
while keeping the tool flexible for other scenarios.

enyst: Just to clarify, you're right we need to do this, to implement the restriction, but I do think you're right it should be optional too, though. The use case I have in mind is with the OpenHands-CLI, and I think it should be possible for the CLI user to enable or disable this.

This bit of code creates the executor... which I think all callers need? It's not clear to me if we have enough flexibility for the CLI to just say e.g. _"`--always-approve` means yolo, we'll disable path enforcement"_? Please correct me if I'm wrong! It's just a tiny question, nothing else, I'd like to understand this.

openhands-tools/openhands/tools/file_editor/editor.py
enyst: Got it, I think the issue is written with the help of the agent too, so who knows, we can take it as starting point, but change it freely if it makes sense. 

I think maybe workspace root makes sense

enyst:
Good point about bash @xingyaoww !

@madhur-tandon Frankly, that’s why I was asking about optional: to disable it on my machine. I’ve seen too many times the agent saying “I can’t use file editor, lemme make a python script and run it to replace stuff in file”, which it… does. 😅

I think this, restricting the agent access, is important for many users, but maybe we could think of a more fundamental solution (like sandboxing, which applications based on the SDK can already use). WDYT?

neubig:
Hey @xingyaoww in that case do you think we should close the original issue that you opened as "not planned"?

xingyaoww:
@neubig fair enough! will close

neubig:
Hi @madhur-tandon really sorry that we had the issue open then decided not to implement it... But because of this I'll close the pull request, but we welcome future ones!

Issue #1101, opened 2025-11-07
Refactor: Improve HTTP client credential update pattern in Remote Workspace


xingyaoww:
## Summary

The current authentication pattern in `APIRemoteWorkspace` requires manually resetting and re-initializing the HTTP client when credentials change. This is error-prone and relies on side effects that are not immediately obvious.

## Current Implementation

In `openhands-workspace/openhands/workspace/remote_api/workspace.py` (lines 114-116):

```python
self.api_key = self._session_api_key
self._client = None  # Reset HTTP client with new host and API key
_ = self.client  # Initialize client by accessing the property
```

## Problem

This pattern has several issues:

1. **Non-obvious side effects**: Accessing the `.client` property triggers lazy initialization, but this is not apparent from the code
2. **Order dependency**: The operations must happen in exact order (set api_key → reset client → access client)
3. **Fragility**: Changes to the parent class `RemoteWorkspace` could break this pattern silently
4. **Code smell**: The underscore assignment `_ = self.client` indicates we're calling something purely for side effects
5. **Separation of concerns**: We have two different authentication mechanisms mixed together:
   - Runtime API auth: `runtime_api_key` with `X-API-Key` header (for managing containers)
   - Agent Server auth: `session_api_key` with `X-Session-API-Key` header (for workspace operations)

## Why It Works This Way

The parent class `RemoteWorkspace` has a lazy-initialized `client` property:

```python
@property
def client(self) -> httpx.Client:
    client = self._client
    if client is None:
        timeout = httpx.Timeout(connect=10.0, read=60.0, write=10.0, pool=10.0)
        client = httpx.Client(base_url=self.host, timeout=timeout, headers=self._headers)
        self._client = client
    return client
```

The `_headers` property (from `RemoteWorkspaceMixin`) uses `self.api_key` to construct authentication headers:

```python
@property
def _headers(self):
    headers = {}
    if self.api_key:
        headers["X-Session-API-Key"] = self.api_key
    return headers
```

So when the runtime starts and we receive a `session_api_key`, we need to:
1. Update `api_key` field to the new value
2. Recreate the HTTP client to use the new credentials
3. But the client is already created with old (None) credentials

The current workaround is to reset `_client = None` and then access `.client` to trigger recreation.

### Option 1: Add a Client Update Method

Add a method to `RemoteWorkspace` or `RemoteWorkspaceMixin`:

```python
def _update_client_credentials(self, api_key: str | None = None, host: str | None = None):
[...]

PR #1425, opened 2025-12-18
client credential update

madhur-tandon:
## Summary of PR

Improve HTTP client credential update pattern in Remote Workspace.
Implements option 2 of the recommendations mentioned in the original issue.
Re-creates the client dynamically behind the scenes when credentials (host, api key, etc.) are changed.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor
- [ ] Other (dependency update, docs, typo fixes, etc.)

## Checklist

- [x] I have read and reviewed the code and I understand what the code is doing.
- [x] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves https://github.com/OpenHands/software-agent-sdk/issues/1101

## Release Notes

- [ ] Include this change in the Release Notes.

### DIFF
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -151,6 +151,9 @@ jobs:
                         echo "workflow_dispatch provided unknown test_type '$test_type'; defaulting to full suite."
                         ;;
                     esac
+                  elif [ "${{ github.event_name }}" = "schedule" ]; then
+                    TEST_TYPE_ARGS="--test-type integration"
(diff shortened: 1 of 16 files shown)


neubig:
@xingyaoww I'm not very confident in my ability to review this one, it'd be great if you could take a look.

madhur-tandon:
@xingyaoww Fair enough, this PR basically does automatic detection so you don't even need to call `reset_client()` explicitly which was originally proposed in the issue...

So maybe that issue should be closed then 😅

PR #1432, opened 2025-12-18
fix end span in local conversation run

Rainhunter13:
There is a mistake in the way the `_end_observability_span` is called as span shouldn't be ended manually inside the `@observe` decorator. This behaviour results in the exception during context detaching. We can remove the observability span end form `run()` function as it happens already inside `close()`.


enyst (approved):
LGTM, thank you!

PR #1433, opened 2025-12-18
feat: add early stopping/pruner for behavior tests cost optimization

### LINKED ISSUE #1417: Early stopping/pruner for cost optimization in behavior tests
Currently, we already have some simple methods to make running behavior tests cheaper (for both frequent use and prompt optimization purposes):
- Truncate observation length to max 5000 chars in trajectory summary sent to LLM judge
- Use a cheap model (Haiku 4.5) for the LLM judge 
- Set max iter to 100 instead of 500 by default

However, the total cost for **5 tests x 6 models** is still pretty high, ranging from [$10 to $15 on average](https://github.com/OpenHands/software-agent-sdk/pull/1348#issuecomment-3660619676).

Another option is to implement some kind of pruner/early stopping mechanism so that we can detect failures early and stop the test case right there, instead of waiting for the long trajectory to finish.

ixchio:
## Summary
Closes #1417

Implements early stopping mechanism to detect failures early and terminate behavior tests before full trajectory completes, reducing LLM costs.

## Changes
- Add `early_stopper.py` with 6 pruner classes
- Integrate early stopping in `BaseIntegrationTest` callback  
- Add `get_early_stopper()` hook in `SoftwareAgentSDKBehaviorTest`
- Apply early stoppers to b01, b02, b05 behavior tests
- Add 17 unit tests for early stopper functionality

## Per discussion with @ryanhoangt
- Pattern-based pruning first (zero LLM cost)
- Stop on first failure signal
- Skip final LLM judge when early stopped
- Reusable pruner classes with base interface

## Test Results
All 17 unit tests passing

### DIFF
--- a/.github/workflows/README-RELEASE.md
+++ b/.github/workflows/README-RELEASE.md
@@ -55,9 +55,22 @@ Once the release is published, the **pypi-release.yml** workflow will automatica
 
 You can monitor the progress in the [Actions tab](https://github.com/OpenHands/software-agent-sdk/actions/workflows/pypi-release.yml).
 
-### Step 5: Post-Release Tasks
+### Step 5: Version Bump PRs (Automated)
+
+After successful PyPI publication, the workflow will automatically create PRs to update SDK versions in downstream repositories:
+
+- **[OpenHands](https://github.com/All-Hands-AI/OpenHands)** - Updates `openhands-sdk`, `openhands-tools`, and `openhands-agent-server` versions
+- **[OpenHands-CLI](https://github.com/All-Hands-AI/openhands-cli)** - Updates `openhands-sdk` and `openhands-tools` versions
+
+These PRs will:
+- Be created automatically with branch name `bump-sdk-X.Y.Z`
+- Include links back to the SDK release
+- Need to be reviewed and merged by the respective repository maintainers
+
+### Step 6: Post-Release Tasks
 
 - [ ] Merge the release PR to main
+- [ ] Review and merge the auto-created version bump PRs in OpenHands and OpenHands-CLI
 - [ ] Run evaluation on OpenHands Index (manual step)
 - [ ] Announce the release
 
--- a/.github/workflows/pypi-release.yml
+++ b/.github/workflows/pypi-release.yml
@@ -11,9 +11,25 @@ on:
 jobs:
     publish:
         runs-on: blacksmith-4vcpu-ubuntu-2404
+        outputs:
+            version: ${{ steps.extract_version.outputs.version }}
         steps:
             - name: Checkout
               uses: actions/checkout@v5
+
+            - name: Extract version from release tag
+              id: extract_version
+              run: |
+                  # Get version from release tag (e.g., v1.2.3 -> 1.2.3)
+                  if [[ "${{ github.event_name }}" == "release" ]]; then
+                    VERSION="${{ github.event.release.tag_name }}"
+                    VERSION="${VERSION#v}"  # Remove 'v' prefix if present
+                  else
+                    # For manual dispatch, extract from pyproject.toml
+                    VERSION=$(grep -m1 '^version = ' openhands-sdk/pyproject.toml | cut -d'"' -f2)
+                  fi
+                  echo "version=$VERSION" >> $GITHUB_OUTPUT
+                  echo "📦 Version: $VERSION"
 
             - name: Install uv
               uses: astral-sh/setup-uv@v7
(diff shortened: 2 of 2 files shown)


ryanhoangt:
@ixchio I think the current implementation has some issues as indicated from the comment above, could you take a look?

ixchio:
> @ixchio I think the current implementation has some issues as indicated from the comment above, could you take a look?

Hey @ryanhoangt, good catch! I just pushed a fix in 3fad22e   the issue was that 
early_stopper
was being accessed in the callback before it was initialized. Moved the initialization before LocalConversation creation. Should be good now, can you re-trigger the tests?

tests/integration/tests/b02_no_oververification.py
```diff
@@ -33,6 +34,82 @@ class NoOververificationTest(SoftwareAgentSDKBehaviorTest):
...
+                    )
+
+                from tests.integration.early_stopper import EarlyStopResult
+
+                return EarlyStopResult(should_stop=False)
+
+        return TerminalTestAwarePruner(
+            max_test_commands=5,  # More targeted test runs allowed
+            broad_test_patterns=[
+                "pytest tests/ ",  # Running entire tests dir (trailing space)
```
ryanhoangt: If there's a trailing space here, will it be able to catch commands such as `pytest tests`? I think we can just use `pytest tests/` pattern?

tests/integration/test_early_stopper.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+# Patch the isinstance checks to work with our mocks
+@pytest.fixture(autouse=True)
+def patch_imports(monkeypatch):
+    """Patch imports to use mocks."""
+    import tests.integration.early_stopper as early_stopper_module
+
+    # Use the real ActionEvent check but treat our mocks as valid
+    original_check = early_stopper_module.FileEditPruner.check
```
ryanhoangt: This seems weird to me since these tests don't examine the real implementation. Is it possible to construct real `FileEditorAction` /
`TerminalAction` objects and test the 
actual methods?

ryanhoangt (commented):
Thank you for the PR, I left a few comments below, and currently the linting workflow seems to fail. One thing I'm concerned about is we have too many concrete implementations of the pruners, which is a bit repetitive and hard to maintain. OTOH I still don't have a clear picture what the best design should look like. @xingyaoww do you have any thoughts about this?

---
This is a proposal from OH, which seems pretty good to me:

Here’s a concrete sketch for a simpler, composable early-stopping system you can drop 
into `tests/integration/early_stopper.py`. The key ideas:

* keep only two reusable pruners (`CommandPatternPruner` and `ActionLimitPruner`)
* provide a `compose` helper to combine them when needed
* let behavior tests configure early-stopping declaratively instead of subclassing

---

### Core primitives

```python
# tests/integration/early_stopper.py

from __future__ import annotations

from typing import Callable, Iterable

from openhands.sdk.event.base import Event
from openhands.sdk.event.llm_convertible.action import ActionEvent

from openhands.tools.file_editor.definition import FileEditorAction, FileEditorTool
from openhands.tools.terminal.definition import TerminalAction, TerminalTool


@dataclass
class EarlyStopResult:
    should_stop: bool
    reason: str | None = None


Checker = Callable[[Iterable[Event]], EarlyStopResult]


def compose(*checkers: Checker) -> Checker:
    """Return a checker that stops when any sub-checker stops."""
    def combined(events: Iterable[Event]) -> EarlyStopResult:
        for checker in checkers:
            result = checker(events)
            if result.should_stop:
                return result
        return EarlyStopResult(False)
    return combined
```

---

### Command-pattern pruner

```python
def CommandPatternPruner(
    tool: str,
    action_filter: Callable[[ActionEvent], bool],
    forbidden_patterns: list[str],
    *,
    reason_template: str,
) -> Checker:
    """Stop when an action's `command` contains a forbidden pattern."""

    def checker(events: Iterable[Event]) -> EarlyStopResult:
        for event in events:
            if isinstance(event, ActionEvent) and event.tool_name == tool:
                action = event.action
                if action is not None and action_filter(event):
                    command = getattr(action, "command", "")
                    for pattern in forbidden_patterns:
                        if pattern in command:
                            return EarlyStopResult(
                                True,
                                reason_template.format(pattern=pattern, 
command=command),
                            )
        return EarlyStopResult(False)

    return checker
```

Example usage (see below for b01/b02).

---

### Action-count limiter

```python
def ActionLimitPruner(
    tool: str,
    action_filter: Callable[[ActionEvent], bool],
    *,
    max_count: int,
    ignore: Callable[[ActionEvent], bool] | None = None,
    reason_template: str,
) -> Checker:
    """Stop when matching actions exceed `max_count`."""

    ignore = ignore or (lambda _: False)

    def checker(events: Iterable[Event]) -> EarlyStopResult:
        matching = 0
        for event in events:
            if isinstance(event, ActionEvent) and event.tool_name == tool:
                if event.action is None:
                    continue
                if ignore(event):
                    continue
                if action_filter(event):
                    matching += 1
                    if matching > max_count:
                        return EarlyStopResult(
                            True, reason_template.format(count=matching)
                        )
        return EarlyStopResult(False)

    return checker
``` 

#### b01 – “no premature implementation”

```python
def get_early_stopper(self):
    return CommandPatternPruner(
        tool=FileEditorTool.name,
        action_filter=lambda ev: isinstance(ev.action, FileEditorAction),
        forbidden_patterns=["create", "str_replace", "insert", "undo_edit"],
        reason_template=(
            "Agent attempted forbidden file edit: '{pattern}' in {command}"
        ),
    )
```

#### b02 – “no over-verification”

```python
def get_early_stopper(self):
    broad_tests = CommandPatternPruner(
        tool=TerminalTool.name,
        action_filter=lambda ev: isinstance(ev.action, TerminalAction)
        and ("pytest" in ev.action.command or "python -m unittest" in 
ev.action.command),
        forbidden_patterns=[
            "pytest tests/",
            "pytest .",
            "python -m pytest .",
            "pytest -x tests/",
        ],
        reason_template="Detected overly broad test command containing '{pattern}'",
    )

    test_spam = ActionLimitPruner(
        tool=TerminalTool.name,
        action_filter=lambda ev: isinstance(ev.action, TerminalAction)
        and ("pytest" in ev.action.command or "python -m unittest" in 
ev.action.command),
        ignore=lambda ev: "tests/tools/terminal" in ev.action.command,
        max_count=5,
        reason_template="Executed {count} test commands (limit 5)",
    )

    return compose(broad_tests, test_spam)
```

#### b05 – “no redundant files”

```python
def get_early_stopper(self):
    return ActionLimitPruner(
        tool=FileEditorTool.name,
        action_filter=lambda ev: isinstance(ev.action, FileEditorAction)
        and ev.action.command == "create",
        ignore=lambda ev: any(
            pattern in ev.action.path for pattern in (".openhands/", "__pycache__/")
        ),
        max_count=3,  # training script, optional README, optional test file
        reason_template="Created {count} files (limit 3)",
    )
```

enyst:
> Command-pattern pruner
> Action-count limiter

Interestingly, this seems to me maybe like a good use case for
- https://github.com/OpenHands/software-agent-sdk/pull/1467

ixchio:
hey @ryanhoangt - fixed both things you mentioned!

**trailing space pattern**: yea good catch, changed it to just `pytest tests/` without the trailing space. makes way more sense 👍

**mock tests**: rewrote them to use real ActionEvent/TerminalAction objects now. was kinda silly to test with mocks that bypass the actual implementation lol. all 17 tests pass with the real deal.

@enyst interesting point about #1467! that hooks system could def clean this up. lmk if you want me to wait for that or just go with this for now

enyst:
> @enyst interesting point about #1467! that hooks system could def clean this up. lmk if you want me to wait for that or just go with this for now

Oh, IMHO we could go with this just fine, thank you! 

I didn’t mean to imply waiting for hooks (though I think it’s ~mergeable), just that it will be great to put to good use. 😅 But it could be in a follow up refactoring.

I’ll defer to @ryanhoangt either way.

ixchio:
> > @enyst interesting point about #1467! that hooks system could def clean this up. lmk if you want me to wait for that or just go with this for now
> 
> Oh, IMHO we could go with this just fine, thank you!
> 
> I didn’t mean to imply waiting for hooks (though I think it’s ~mergeable), just that it will be great to put to good use. 😅 But it could be in a follow up refactoring.
> 
> I’ll defer to @ryanhoangt either way.

sounds good, will wait on @ryanhoangt for final approval 🙏

tests/integration/early_stopper.py
```diff
@@ -0,0 +1,357 @@
...
+        if test_commands > self.max_test_commands:
+            return EarlyStopResult(
+                should_stop=True,
+                reason=(
+                    f"Executed {test_commands} test commands, "
+                    f"exceeding limit of {self.max_test_commands}"
+                ),
+            )
+
+        return EarlyStopResult(should_stop=False)
```
xingyaoww: This early stopper seems a bit birttle, maybe we can consider remove this?

tests/integration/early_stopper.py
```diff
@@ -0,0 +1,357 @@
...
+                event_count,
+                "STOP" if self._cached_result.should_stop else "CONTINUE",
+            )
+
+        except Exception as e:
+            # On error, don't stop - let the test continue
+            logger.warning("LLM judge pruner error: %s", e)
+            self._cached_result = EarlyStopResult(should_stop=False)
+
+        return self._cached_result
```
xingyaoww: This is probably also an overkill too?

tests/integration/tests/b05_do_not_create_redundant_files.py
```diff
@@ -26,12 +28,77 @@
...
+        if create_count > self.max_creates:
+            return EarlyStopResult(
+                should_stop=True,
+                reason=(
+                    f"Created {create_count} files (max {self.max_creates}): "
+                    f"{', '.join(created_files[-3:])}"
+                ),
+            )
+
+        return EarlyStopResult(should_stop=False)
```
xingyaoww: I think this can also be a bit brittle

tests/integration/tests/b02_no_oververification.py
```diff
@@ -33,6 +34,82 @@ class NoOververificationTest(SoftwareAgentSDKBehaviorTest):
...
+
+        return TerminalTestAwarePruner(
+            max_test_commands=5,  # More targeted test runs allowed
+            broad_test_patterns=[
+                "pytest tests/",  # Running entire tests dir
+                "pytest .",  # Running all tests
+                "python -m pytest .",  # Running all tests
+                "pytest -x tests/",  # Broad with stop-on-first
+            ],
+        )
```
xingyaoww: I think it will be better if we leave it to the final llm judge to determine this, since rules like this might not always work 🤔

xingyaoww (commented):
I like the overall idea! But i do agree with @ryanhoangt that some of these early stopper could be a bit birttle.

@ixchio wdyt about removing the brittle ones, and we can get the base class and simpler earlystopper like `FileEditPruner` / `TerminalPruner` in?

tests/integration/early_stopper.py
ixchio: 👍 on it - stripping out the brittle stuff, keeping FileEditPruner/BashCommandPruner/CompositeEarlyStopper

tests/integration/early_stopper.py
ixchio: agreed, gone

tests/integration/tests/b05_do_not_create_redundant_files.py
ixchio: fair, dropping this one too

tests/integration/tests/b02_no_oververification.py
ixchio: > I think it will be better if we leave it to the final llm judge to determine this, since rules like this might not always work 🤔

makes sense, judge handles this better anyway

ixchio:
> I like the overall idea! But i do agree with @ryanhoangt that some of these early stopper could be a bit birttle.
> 
> @ixchio wdyt about removing the brittle ones, and we can get the base class and simpler earlystopper like `FileEditPruner` / `TerminalPruner` in?

👍 pushed - kept the simple ones, nuked the brittle stuff

tests/integration/tests/b02_no_oververification.py
```diff
@@ -33,6 +34,14 @@ class NoOververificationTest(SoftwareAgentSDKBehaviorTest):
...
     INSTRUCTION: str = INSTRUCTION
 
+    def get_early_stopper(self) -> EarlyStopperBase | None:
+        """No early stopper for this test.
+
+        Over-verification is a nuanced behavior that's better evaluated
+        by the final LLM judge rather than pattern-based pruning.
+        """
+        return None
+
```
xingyaoww: Let's remove this, since it will default to None?
```suggestion
```

ixchio: > Let's remove this, since it will default to None?

done

PR #1435, opened 2025-12-18
Workspace env vars for laminar evaluations

Rainhunter13:
Add option to pass environment variable to a `RemoteAPIWorkspace`, similar to how `DockerWorkspace` already behaves. This is necessary for Laminar evaluation to have a single continuous trace through different environments.

This is not yet tested.


Rainhunter13:
solved with https://github.com/OpenHands/software-agent-sdk/pull/1540

PR #1443, opened 2025-12-18
Save screenshots from browser tool

hamiltop:
When using the browser we often will want to capture screenshots and other browser data to use in deterministic scripts.

Screenshots are useful for recording evidence during QA and other flows. They can also be used with a captcha solver when needed.

Being able to get and set cookies and other data allows us to reproduce issues we hit in the browser in more localized scripts and inspect the data. We can also use it to take a browser session that was manually initiated and allow the agent to continue the session.


xingyaoww:
@OpenHands please fix the failing actions on PR #1443 at a forked branch `hamiltop:save_screenshots_from_browser_tool`

PR #1467, opened 2025-12-20
feat: Add hooks system for event interception

ak684:
Adds a hooks system that allows running shell scripts at key lifecycle events during agent execution, enabling deterministic control over agent behavior.

Features:
- 5 event types wired: PreToolUse, PostToolUse, UserPromptSubmit, SessionStart, SessionEnd
- 1 event type API-ready: Stop (requires core run loop changes for follow-up)
- Exit code semantics: 0=success, 2=block operation
- JSON I/O with decision/reason fields (additionalContext parsed, injection in follow-up)
- Matcher patterns: wildcard, exact match, regex auto-detection
- Per-hook timeout configuration
- Environment variables: OPENHANDS_PROJECT_DIR, OPENHANDS_SESSION_ID, OPENHANDS_TOOL_NAME, OPENHANDS_EVENT_TYPE

Blocking integration:
- PreToolUse: blocked_actions field on ConversationState, Agent emits UserRejectObservation instead of executing
- UserPromptSubmit: blocked_messages field, Agent skips LLM processing for blocked messages
- Persistence: blocked state survives conversation resume

Auto-wiring:
- hook_config parameter on Conversation() factory auto-wires SessionStart/SessionEnd
- RemoteConversation supports SessionStart/SessionEnd only (warns about unsupported hooks)

Config locations:
- .openhands/hooks.json (project-local)
- ~/.openhands/hooks.json (user-global)

Includes 38 unit/integration tests and an example (examples/01_standalone_sdk/33_hooks.py).

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:9cbd9b4-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-9cbd9b4-python \
  ghcr.io/openhands/agent-server:9cbd9b4-python
```

```
ghcr.io/openhands/agent-server:9cbd9b4-golang-amd64
ghcr.io/openhands/agent-server:9cbd9b4-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:9cbd9b4-golang-arm64
ghcr.io/openhands/agent-server:9cbd9b4-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:9cbd9b4-java-amd64
ghcr.io/openhands/agent-server:9cbd9b4-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:9cbd9b4-java-arm64
ghcr.io/openhands/agent-server:9cbd9b4-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:9cbd9b4-python-amd64
ghcr.io/openhands/agent-server:9cbd9b4-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:9cbd9b4-python-arm64
ghcr.io/openhands/agent-server:9cbd9b4-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:9cbd9b4-golang
ghcr.io/openhands/agent-server:9cbd9b4-java
ghcr.io/openhands/agent-server:9cbd9b4-python
```

- Each variant tag (e.g., `9cbd9b4-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `9cbd9b4-python-amd64`) are also available if needed

### DIFF
--- a/examples/01_standalone_sdk/33_hooks.py
+++ b/examples/01_standalone_sdk/33_hooks.py
@@ -6,7 +6,7 @@ Hooks are shell scripts that run at key lifecycle events, enabling:
 - Logging tool usage after execution (PostToolUse)
 - Processing user messages before they reach the agent (UserPromptSubmit)
 
-Hooks are configured in .openhands/hooks.json or passed programmatically via hook_config.
+Hooks are configured in .openhands/hooks.json or passed programmatically.
 """
 
 import os
@@ -150,9 +150,7 @@ def main():
         print("- Exit code 2 from a hook blocks the operation")
         print("- Hooks receive JSON on stdin with event details")
         print("- Environment variables like $OPENHANDS_TOOL_NAME simplify simple hooks")
-        print(
-            "- Hook config can be in .openhands/hooks.json or passed via hook_config"
-        )
+        print("- Hook config can be in .openhands/hooks.json or passed via hook_config")
 
         # Report cost
         cost = conversation.conversation_stats.get_combined_metrics().accumulated_cost


openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
```diff
@@ -0,0 +1,185 @@
...
+        # Store pending action for PostToolUse
+        self._pending_actions[event.id] = event
+
+    def _handle_post_tool_use(self, event: ObservationEvent) -> None:
+        """Handle PostToolUse hooks after an action completes."""
+        if not self.hook_manager.has_hooks(HookEventType.POST_TOOL_USE):
+            return
+
+        # Find the corresponding action
+        action_event = self._pending_actions.pop(event.action_id, None)
```
enyst: Isn't this ... possible to find in another way? We have its observation, right, and IIRC it has a reference to the action it came from? Or maybe the tool id can help us find the action.

Just thinking, maybe we don't need to store `_pending_actions` in the HookManager. We have another `pending_actions` too and it feels a bit confusing and might be error-prone

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
```diff
@@ -0,0 +1,185 @@
...
+        should_continue, additional_context, results = (
+            self.hook_manager.run_user_prompt_submit(message=message)
+        )
+
+        if not should_continue:
+            reason = self.hook_manager.get_blocking_reason(results)
+            logger.warning(f"Hook blocked user message: {reason}")
+
+        # TODO: Inject additional_context into the message
+        if additional_context:
```
enyst, line 179: Exactly! We currently implement this behavior directly, to inject keyword-based Skills/microagents, we could redesign it as a hook, provided and enabled by default in the SDK.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -0,0 +1,155 @@
...
+        env: dict[str, str] | None = None,
+    ) -> HookResult:
+        """Execute a single hook."""
+        # Prepare environment
+        hook_env = os.environ.copy()
+        hook_env["OPENHANDS_PROJECT_DIR"] = self.working_dir
+        hook_env["OPENHANDS_SESSION_ID"] = event.session_id or ""
+        hook_env["OPENHANDS_EVENT_TYPE"] = event.event_type
+        if event.tool_name:
+            hook_env["OPENHANDS_TOOL_NAME"] = event.tool_name
```
enyst, line 58: Out of curiosity, why would we need these env vars? Are they really useful or more like a guess of what could be useful? I don't see it...

openhands-sdk/openhands/sdk/hooks/types.py
```diff
@@ -0,0 +1,42 @@
...
+    model_config = {"use_enum_values": True}
+
+
+class HookDecision(str, Enum):
+    """Decisions a hook can make about an operation."""
+
+    ALLOW = "allow"
+    DENY = "deny"
+    BLOCK = "block"  # Alias for deny
+    APPROVE = "approve"  # Alias for allow
```
enyst: Why do we need aliases, can't we just choose the words we prefer?

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
ak684: will try another approach!

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
ak684: precisely!

openhands-sdk/openhands/sdk/hooks/executor.py
ak684: Yea so admittedly this is a direct mirror of CC's hook env var approach but I think it's useful. The thinking is that some hook authors might want to reference event metadata for their script use cases. Depending on the dev environment, manually parsing this information from JSON can get cumbersome ie.

**without env vars example**
```bash
#!/bin/bash

# always need to write this to access input
input=$(cat)

# Option A: Fragile grep/sed (breaks on nested JSON, special chars, etc.)
tool_name=$(echo "$input" | grep -o '"tool_name":"[^"]*"' | sed 's/"tool_name":"//;s/"//')

# Option B: Use Python (requires dependency, slower)
tool_name=$(echo "$input" | python3 -c "import sys,json; print(json.load(sys.stdin).get('tool_name',''))")

# Option C: Use node (requires dependency)
tool_name=$(echo "$input" | node -e "let d='';process.stdin.on('data',c=>d+=c);process.stdin.on('end',()=>console.log(JSON.parse(d).tool_name))")

# Option D: Use jq (requires dependency)
tool_name=$(echo "$input" | jq -r '.tool_name')

# the core point of the hook
if [[ "$tool_name" == "terminal" ]]; then
    echo '{"decision": "deny", "reason": "terminal blocked"}'
    exit 2
fi
exit 0
```

If we're open to supporting env vars, `input=$(cat)` plus all of the options shown above can be simplified to:

**env vars example**
```bash
#!/bin/bash
# No regex or dependencies required! Just plain bash.
if [[ "$OPENHANDS_TOOL_NAME" == "terminal" ]]; then
    echo '{"decision": "deny", "reason": "terminal blocked"}'
    exit 2
fi
exit 0
```
Plus, env vars are generally more convenient for use cases involving subprocesses (inherited automatically), shell functions (no need to pass args), and piping stdin to external tools.

openhands-sdk/openhands/sdk/hooks/types.py
ak684: yea this is a vestige from a former approach, will remove!

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -462,8 +463,26 @@ def _execute_action_event(
...
+        of executing the action.
         """
         state = conversation.state
+
+        # Check if this action was blocked by a PreToolUse hook
+        if action_event.id in state.blocked_actions:
+            reason = state.blocked_actions.pop(action_event.id)
+            logger.info(f"Action '{action_event.tool_name}' blocked by hook: {reason}")
+            rejection = UserRejectObservation(
+                action_id=action_event.id,
```
enyst, line 489: A thought here, in the current use, when we use `UserRejectObservation`, we seem to reject _all_ pending actions, or so it looks like from the code (sorry I didn't take the reference, but it's the other use of `UserRejectObservation`)

This rejects only the offending one. I actually think this should work fine, and since it's just executing it, it should be ok to have this associated obs. Idk, just noting the (apparent) different behavior, or I could be wrong it's different 🤔

openhands-sdk/openhands/sdk/hooks/config.py
```diff
@@ -0,0 +1,163 @@
...
+from openhands.sdk.hooks.types import HookEventType
+
+
+logger = logging.getLogger(__name__)
+
+
+class HookDefinition(BaseModel):
+    """A single hook definition (command to execute)."""
+
+    type: str = "command"
```
enyst: Totally a nit: this could maybe be an enum? Not necessary for this PR IMHO, maybe a follow up, or with some other updates when Claude finds a way, ~as it surely will~. Oops sorry did I say out loud

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
```diff
@@ -0,0 +1,187 @@
...
+    def _handle_post_tool_use(self, event: ObservationEvent) -> None:
+        """Handle PostToolUse hooks after an action completes."""
+        if not self.hook_manager.has_hooks(HookEventType.POST_TOOL_USE):
+            return
+
+        # Reverse look up corrosponding action from state events
+        action_event = None
+        if self._conversation_state is not None:
+            for e in reversed(self._conversation_state.events):
+                if isinstance(e, ActionEvent) and e.id == event.action_id:
```
enyst: Aha, OK. There won't be many events, until it gets to it

@openhands remind me please, does `reversed(self._conversation_state.events)` create a new list?

Note that events could be in the thousands, analyze performance of this code, do you have any optimization suggestions?

enyst:

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
enyst: Ha. OpenHands option ii is pending actions. 😅 

Option i, though, could maybe work better? I just suddenly remembered there must be a fast way to search events 
```
Yes. EventLog maintains an in-memory id→index map and exposes:
- get_index(event_id) -> int
- get_id(idx) -> EventID
- __getitem__(idx) -> Event

For your case, you can replace the reverse scan with:
events = self._conversation_state.events
try:
    e = events[events.get_index(event.action_id)]
    if isinstance(e, ActionEvent):
        action_event = e
except KeyError:
    action_event = None

This is O(1) for the lookup plus a single file read, versus O(n) reverse iteration. There isn’t a built-in predicate search; for anything other than by-id you’d still iterate.
```
WDYT?

openhands-sdk/openhands/sdk/hooks/executor.py
enyst: Ahh you're only setting them for the hook env, not reading them from the SDK env! I'm dense, sorry. Sounds good!

enyst:
Curious what you think about its "review-roasted", @ak684 . Personally, some quick thoughts
- maybe the config issue is worth looking at?
- JSON, user prompt,`jq` don't seem like problems to me
- pretty sure we truncate tool outputs already (not sure if soon enough), that's probably a non-issue
- "blocked_actions mutations won’t trigger autosave; document as ephemeral or provide a setter if persistence is intended." - seems like a good question to me, for conversation restore... but we have the reject obs which documents the reason, maybe that's ok 🤔

openhands-sdk/openhands/sdk/hooks/config.py
ak684: it's so simple, I'll make it an enum now!

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
ak684: agreed!

openhands-sdk/openhands/sdk/agent/agent.py
ak684: You're correct here in the differences!

ak684: _Seems_ like you're already onboard with the semantics (lmk if not though)! for posterity I'll add that this is another instance were we match CC's UI/UX behavior. I think it makes sense as well though because in general:

1. for automated hook rejection, I, as the hook author, am writing my hook with the expectation that my hook will block precisely what I match in the rule (and nothing more)
2. for manual, reject-all use cases in CLI/SaaS/Cloud (not powered by hooks), when i reject the llm's actions I expect everything to stop, often because i want to immediately send a message to give further instruction or redirection

ak684:
> Curious what you think about its "review-roasted", @ak684 . Personally, some quick thoughts
> 
> * maybe the config issue is worth looking at?
> * JSON, user prompt,`jq` don't seem like problems to me
> * pretty sure we truncate tool outputs already (not sure if soon enough), that's probably a non-issue
> * "blocked_actions mutations won’t trigger autosave; document as ephemeral or provide a setter if persistence is intended." - seems like a good question to me, for conversation restore... but we have the reject obs which documents the reason, maybe that's ok 🤔

outside of what's already covered via your pr review - imo the best thing from this review-roasted was the config call! that's something that didnt show up as an issue in my local SDK/CLI testing but very likely would break on Cloud/SaaS/OSS. included a fix for this.

The other interesting call out is the non-truncated tool outputs point, because actually I do not believe we truncate on `Observation.__str__()`, we only truncate on `ObservationEvent.__str__()`.. so I'm curious what you think is the best way to address that (assuming we want to address it)? if so, one idea is to truncate in the hooks code to `DEFAULT_TEXT_CONTENT_LIMIT`? haven't added this fix just yet bc want to know you're thoughts

added a few simple nits that I genuinely thought added value, but also left out other nits in the roast i thought were fine

oh! also i wired up SessionStart, SessionEnd now after being called out, instead of waiting until follow up, since it was only a few lines of code.. and re: blocked_actions.. so we're using blocked_actions as a temporary, mechanical object, to signal from the hook callback to the agent loop "don't execute this, reject it instead." If we'd like to make that ephemeral nature more clear via comments or something i can do that.. For resumes, using the permanent record of UserRejectObservation should be more than enough.

openhands-sdk/openhands/sdk/agent/agent.py
enyst: Ahh, that’s a good point! Yes, I was already onboard with the semantics, but I didn’t realize that Point 2 is the correct behavior for user rejection. Yes, yes it is, this makes sense to me. (maybe we could even comment it somewhere in the code, both places, so that agents don’t mess it up next time). Thank you!

examples/01_standalone_sdk/33_hooks.py
enyst, line 1: To note, I think maybe there’s some tiny fixes needed here, maybe file name (not sure which is the last in `main`, they should be in order), and it should end with a cost calculation, because otherwise `test-examples` workflow will fail, and that would give Xingyao headaches for releases

@openhands verify this example file and fix as necessary to be like the others and work right with the workflow.

PR #1470, opened 2025-12-22
Add compatibility fixes for Nemotron and similar models

pankajarm:
Key improvements:
1. FN_REGEX_PATTERN: Make newline optional after function name
2. Handle missing/None content in messages gracefully
3. Strip whitespace from parameter values (models add extra newlines)
4. Add _preprocess_model_output() to clean wrapper tags SAFELY
   - Only strips </think>, <tool_call>, </tool_call> at boundaries
   - Preserves these strings if they appear inside parameter values
5. Apply tool name aliases only as fallback (try exact match first)
6. Skip security_risk from required param check in fn_call_converter (validation happens later in _extract_security_risk with proper context)
7. Skip ICL examples for nemotron (limited context window)

Best practices applied:
- All changes are backwards-compatible
- Tool aliases only applied if exact name not found (no regression)
- Preprocessing only at boundaries (preserves param content)
- security_risk schema unchanged (still required for GPT-4/Claude)

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -24,7 +24,7 @@ jobs:
             - name: Checkout repository
               uses: actions/checkout@v5
               with:
-                  token: ${{ secrets.GITHUB_TOKEN }}
+                  token: ${{ secrets.ALLHANDS_BOT_GITHUB_PAT }}
 
             - name: Install uv
               uses: astral-sh/setup-uv@v7
@@ -65,7 +65,7 @@ jobs:
 
             - name: Create Pull Request
               env:
-                  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  GH_TOKEN: ${{ secrets.ALLHANDS_BOT_GITHUB_PAT }}
               run: |
                   cat > pr_body.txt << 'EOF'
                   ## Release v${{ inputs.version }}
@@ -75,8 +75,9 @@ jobs:
                   ### Release Checklist
                   - [x] Version set to ${{ inputs.version }}
                   - [ ] Fix any deprecation deadlines if they exist
-                  - [ ] Integration tests pass (tagged with `integration-tests`)
-                  - [ ] Example checks pass (tagged with `test-examples`)
+                  - [ ] Integration tests pass (tagged with `integration-test`)
+                  - [ ] Behavior tests pass (tagged with `behavior-test`)
+                  - [ ] Example tests pass (tagged with `test-examples`)
                   - [ ] Draft release created at https://github.com/OpenHands/software-agent-sdk/releases/new
                     - [ ] Select tag: `v${{ inputs.version }}`
                     - [ ] Select branch: `${{ env.BRANCH_NAME }}`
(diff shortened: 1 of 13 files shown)


pankajarm:
Tested all 7 checks locally including tools-tests which was changed before pushing this.

openhands-sdk/openhands/sdk/llm/mixins/fn_call_converter.py
```diff
@@ -1124,15 +1151,32 @@ def convert_non_fncall_messages_to_fncall_messages(
...
+                    )
+
+                matching_tool = _find_tool(fn_name)
+                # Try aliases if tool not found (some models use legacy names)
+                if not matching_tool:
+                    TOOL_NAME_ALIASES = {
+                        "str_replace_editor": "file_editor",
+                        "bash": "terminal",
+                        "execute_bash": "terminal",
+                        "str_replace": "file_editor",
```
enyst, line 1175: This... is this really Nemotron doing it? Or could it have been due to an outdated prompt or anything in the context? No, it probably isn't prompting... 🤔

xingyaoww: Yeah i think so, nemotron is really overfitted to `execute_bash` in my experience 😓

pankajarm: > Yep, it's definitely Nemotron being I guess Nemotron 😅
>
> We ran into this while testing - even when the prompt clearly says the tools are terminal and file_editor, it keeps outputting execute_bash and str_replace_editor. can't help but think it may have been trained on older OpenHands data with those names baked in pretty hard.
>
> The alias fix is just a fallback though - we try exact match first, so it won't mess with models that already get the names right.
>
> Also hit a few other quirks: it skips the newline after <function=...>, wraps stuff in <tool_call> tags, and sometimes throws in extra whitespace. Small stuff but enough to break the parser.
>
> Oh and we skip ICL examples for it too - similar to what's already done for devstral. Context window issues.

LMK if all looks ok.

enyst: > can't help but think it may have been trained on older OpenHands data with those names baked in pretty hard

🫠 

My thought too 🤷‍♂️

pankajarm: Yeah I don't think Nemotron-3-Nano-30B-A3B has enough chops, Here are all the trajectories and metrics:
https://huggingface.co/datasets/pankajmathur/nemotron-nano-swebench-verified-traj

Metric 	Value
Total Instances 	500
Submitted 	500 (100%)
Evaluated 	355 (71%)
Resolved 	82 (23.1%)
Empty Patches 	145 (29%)

pankajarm:
thanks, just want to add that I don't think Nemotron-3-Nano-30B-A3B has enough chops, Here are all the trajectories and metrics:
https://huggingface.co/datasets/pankajmathur/nemotron-nano-swebench-verified-traj

Metric 	Value
Total Instances 	500
Submitted 	500 (100%)
Evaluated 	355 (71%)
Resolved 	82 (23.1%)
Empty Patches 	145 (29%)

PR #1506, opened 2025-12-24
fix: make StaticSecret.value optional to fix conversation resume

### LINKED ISSUE #1505: Conversations fail to resume after pod restart when StaticSecret values are redacted
## Problem

Users with long conversations on app-all-hands.dev get "An error occurred. Please try again" when reconnecting because their conversation state cannot be loaded.

## Root Cause

The bug is a serialization/deserialization mismatch:

1. In `state.py` line 150, conversation state is saved with:
   ```python
   payload = self.model_dump_json(exclude_none=True)
   ```

2. When secrets are redacted (for security), the `StaticSecret.value` field becomes `None`

3. The `exclude_none=True` flag then **removes the `value` field entirely** from the JSON

4. When the conversation is resumed and state is loaded, Pydantic validation fails because `StaticSecret.value` is defined as a required field (no default value):
   ```python
   class StaticSecret(SecretSource):
       value: SecretStr  # Required! No default
   ```

## Error from Datadog Logs

```
input_value={'kind': 'StaticSecret'}
pydantic_core._pydantic_core.ValidationError: 
  secret_registry.secret_sources.LLM_API_KEY.StaticSecret.value
    Field required [type=missing, ...]
```

## Why Some Conversations Work

Conversations that don't use `StaticSecret` (only `LookupSecret`) or haven't had their pod restarted after secrets were redacted will continue to work. The error only manifests on the second resume after secrets have been serialized with `None` values.

## Affected Files

- `openhands-sdk/openhands/sdk/secret/secrets.py` - `StaticSecret.value` needs a default
- `openhands-sdk/openhands/sdk/conversation/state.py` - uses `exclude_none=True`

## Proposed Fix

Understand why secrets are being redacted before saving conversations in prod.
This is probably due to a missing Pydantic cipher context, which should be made available.

neubig:
## Summary

Fixes #1505 

Users with long conversations on app-all-hands.dev were getting "An error occurred. Please try again" when reconnecting because their conversation state could not be loaded.

## Root Cause

When conversation state is serialized with `exclude_none=True` (in `state.py` line 150), and `StaticSecret.value` is `None` (redacted for security), the value field is omitted entirely from the JSON. This caused a Pydantic validation error when resuming conversations after pod restart since `StaticSecret.value` was a required field.

## Fix

Make `StaticSecret.value` optional with a default value of `None`:
```python
value: SecretStr | None = None
```

Also update `get_value()` to handle the `None` case.

## Testing

Added regression test `test_conversation_state_resume_with_redacted_secrets` that:
1. Creates a conversation with secrets in the secret_registry
2. Simulates the production scenario by writing state with missing `value` fields
3. Confirms the conversation can be resumed without validation errors

**Test confirmed to fail before fix and pass after fix.**

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:fcd85e9-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-fcd85e9-python \
  ghcr.io/openhands/agent-server:fcd85e9-python
```

```
ghcr.io/openhands/agent-server:fcd85e9-golang-amd64
ghcr.io/openhands/agent-server:fcd85e9-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:fcd85e9-golang-arm64
ghcr.io/openhands/agent-server:fcd85e9-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:fcd85e9-java-amd64
ghcr.io/openhands/agent-server:fcd85e9-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:fcd85e9-java-arm64
ghcr.io/openhands/agent-server:fcd85e9-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:fcd85e9-python-amd64
ghcr.io/openhands/agent-server:fcd85e9-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:fcd85e9-python-arm64
ghcr.io/openhands/agent-server:fcd85e9-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:fcd85e9-golang
ghcr.io/openhands/agent-server:fcd85e9-java
ghcr.io/openhands/agent-server:fcd85e9-python
```

- Each variant tag (e.g., `fcd85e9-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `fcd85e9-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/secret/secrets.py
+++ b/openhands-sdk/openhands/sdk/secret/secrets.py
@@ -1,12 +1,23 @@
 """Secret sources and types for handling sensitive data."""
 
+import logging
 from abc import ABC, abstractmethod
 
 import httpx
-from pydantic import Field, SecretStr, field_serializer, field_validator
+from pydantic import (
+    Field,
+    FieldSerializationInfo,
+    SecretStr,
+    field_serializer,
+    field_validator,
+)
+from pydantic.types import _secret_display
 
 from openhands.sdk.utils.models import DiscriminatedUnionMixin
 from openhands.sdk.utils.pydantic_secrets import serialize_secret, validate_secret
+
+
+logger = logging.getLogger(__name__)
 
 
 class SecretSource(DiscriminatedUnionMixin, ABC):
@@ -38,8 +49,27 @@ class StaticSecret(SecretSource):
         return validate_secret(v, info)
 
     @field_serializer("value", when_used="always")
-    def _serialize_secrets(self, v: SecretStr | None, info):
-        return serialize_secret(v, info)
+    def _serialize_secrets(
+        self, v: SecretStr | None, info: FieldSerializationInfo
+    ) -> str | None:
+        if v is None:
+            return None
+
+        result = serialize_secret(v, info)
+
+        # Check if the secret was masked by Pydantic
+        # _secret_display returns "**********" for non-empty secrets
+        if isinstance(result, SecretStr) and str(result) == _secret_display(
+            v.get_secret_value()
+        ):
+            logger.warning(
+                "No cipher context available, secret will be lost during serialization"
+            )
+            return None
+
+        # At this point result should be a string (encrypted or exposed)
+        assert isinstance(result, str)
+        return result
 
 
 class LookupSecret(SecretSource):
--- a/tests/sdk/conversation/test_secret_source.py
+++ b/tests/sdk/conversation/test_secret_source.py
@@ -100,6 +100,25 @@ def test_static_secret_with_value():
     assert validated.get_value() == "my-secret"
 
 
+def test_static_secret_serializes_to_none_without_cipher():
+    """Test StaticSecret serializes value as None when no cipher/expose context.
+
+    This ensures secrets are clearly marked as lost (not persisted) rather than
+    using ambiguous "**********" masking.
+    """
+    secret = StaticSecret(value=SecretStr("my-secret"))
+
+    # Without context, value should serialize to None (not "**********")
+    dumped = secret.model_dump(mode="json")
+    assert dumped == {"kind": "StaticSecret", "description": None, "value": None}
+
+    # When value is actually None internally, exclude_none omits it
+    secret_none = StaticSecret(value=None)
+    dumped_none = secret_none.model_dump(mode="json", exclude_none=True)
+    assert dumped_none == {"kind": "StaticSecret"}
+    assert "value" not in dumped_none
+
+
 def test_conversation_state_resume_with_redacted_secrets():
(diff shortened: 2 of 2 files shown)


enyst (dismissed):
Sorry, could you please explain why is it None? I think we have an env var to allow for encrypted secrets to be persisted, is it not in use? 

If it's not in use, then even if the conversation resumes, it wouldn't have the secrets anymore... which doesn't seem like the correct behavior, unless I'm missing something.

_Edited to add:_ the env var in this [slack reference](https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1766175998707049?thread_ts=1766161785.015039&cid=C06U8UTKSAD) maybe?

enyst:
To note, I'm most happy to dismiss this request for changes if you're confident this is the correct behavior now. I spoke from my current understanding which could be outdated. I'm not aware the behavior has changed, but it _could_ have changed. (a possible alternative could be, it seems to me, that we re-send specifically secrets on restore)

I'd love to know what @tofarr thinks here, he implemented this functionality. I'd also love if we could somehow test this case... This is important behavior of the application, losing secrets has been a long running pain in multiple forms and reasons...

neubig:
OK, I think this is more solid now:
1. More explicit checks about secrets
2. Puts out a warning when a secret is lost in serialization

There's a separate issue of trying to fix the cipher context in the prod openhands app, which should be fixed too.

openhands-sdk/openhands/sdk/secret/secrets.py
```diff
@@ -36,8 +49,27 @@ def _validate_secrets(cls, v: SecretStr | None, info):
         return validate_secret(v, info)
 
     @field_serializer("value", when_used="always")
-    def _serialize_secrets(self, v: SecretStr | None, info):
-        return serialize_secret(v, info)
+    def _serialize_secrets(
+        self, v: SecretStr | None, info: FieldSerializationInfo
+    ) -> str | None:
+        if v is None:
+            return None
```
enyst, line 56: If it’s None, this secret was lost already, if I understand this correctly? If we really want this `| None`, maybe we could add a warning here too.

neubig:
To clarify, LLM_API_KEY is a secret that the user set, so the agent will still work but it won't have access to the env var LLM_API_KEY.

That being said, I do agree with you that it might be better to solve the underlying issue. I'll close this.

PR #1537, opened 2025-12-29
feat: relax tool matching on resume

### LINKED ISSUE #1533: Allow adding new tools when resuming conversations without failing
## Problem

Currently, when resuming a conversation, tools have to match exactly or the whole resume fails. This is strict and prevents users from adding new tools to an existing conversation.

## Proposed Solution

Relax the tool matching validation during conversation resume:
- **Only fail** if a tool that was **used in history** is missing
- **Allow** adding new tools that were not present in the original conversation

This would provide a better user experience while still maintaining consistency for tools that were actually used.

## Context

This suggestion came from the discussion in #1451 (Proposal: agent as composed of immutable instances).

As noted by @ixchio:
> One thing I noticed tho - tools have to match exactly or the whole resume fails. Makes sense for consistency but feels kinda strict? Like if I just want to add a new tool to an existing convo, why should that break things. Maybe only fail if a tool used in history is missing, not if a new one is added.

## Related

- #1451

ixchio:
## Summary
Only fail if a tool that was *used in history* is missing. Allow adding new tools freely when resuming conversations.

## Changes
- Add optional `used_tools: set[str] | None` param to `resolve_diff_from_deserialized()`
- Extract used tool names from event history in `state.py` before reconciliation
- When `used_tools` is provided, only validate that those tools exist in runtime
- New tools can be added to resumed conversations without breaking anything

## Tests
- `test_conversation_allows_removing_unused_tools` - removing tools that were never used is allowed
- `test_conversation_allows_adding_new_tools` - adding new tools when resuming is allowed

All existing reconciliation tests continue to pass.

Fixes #1533

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+## Summary
+
+[fill in a summary of this PR]
+
+## Checklist
+
+- [ ] If the PR is changing/adding functionality, are there tests to reflect this?
+- [ ] If there is an example, have you run the example to make sure that it works?
+- [ ] If there are instructions on how to run the code, have you followed the instructions and made sure that it works?
+- [ ] If the feature is significant enough to require documentation, is there a PR open on the OpenHands/docs repository with the same branch name?
+- [ ] Is the github CI passing?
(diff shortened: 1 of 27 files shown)


neubig:
@openhands /codereview-roasted including checking if you can (1) reproduce the issue with the previous code, and (2) run the new code and demonstrate that the problem that existed with the previous code is now fixed.

ixchio:
Closing - recreating with clean single commit

neubig:
BTW @ixchio it's not necessary to create a single commit -- we squash merge all commits, so they get merged into a single commit anyway.

ixchio:
> BTW @ixchio it's not necessary to create a single commit -- we squash merge all commits, so they get merged into a single commit anyway.

Ah good to know! Thanks for the tip - will skip the cleanup next time since you squash merge anyway. Recreated as #1538 with clean branch, but same code.

PR #1542, opened 2025-12-29
refactor: remove reconciliation methods, use runtime agent directly

### LINKED ISSUE #1451: Proposal: agent as composed of immutable instances
### **Problem:**

We see quite a number of bug reports on restoring conversations with different settings. Many of those I’ve seen, I think, are LLM settings like model name, set encrypted reasoning, etc, but also MCP add/remove, Skills modification between sessions, and others.

The issues come from agent immutability, which seemed to imply LLM immutability, agent context immutability (like skills), condenser llm etc. The interesting part is, this “freezing” of the agent was one of the design principles of V1.

It’s useful for reproducibility, but it’s maybe not so good for user experience. As of now, we seem to patch things one by one in places where they’re reported, e.g. in metadata, in skills, in security, in condenser, in reasoning attributes.

### **Proposed solution:**

Regard the Agent as:
- composable, that is composed of
- immutable parts (llm, condenser, initial context, tools, MCP)

This is the minimal change I can think of, to keep a lot of the guarantees of immutability, and also unblock development of flexible user interfaces and make these bug reports a thing of the past.

### **References:**

1. We actually touched this from the PR we did it as [an experiment in how to serialize the agent]().
2. Xingyao actually [foresaw us ending this practice](https://github.com/OpenHands/software-agent-sdk/pull/41/files/e8ad8621f6cc45d8564e649dff20dadc47e07449#r2337733217) here.

enyst:
## Summary

This PR removes all reconciliation methods (`resolve_diff_from_deserialized`) and uses the provided Agent directly when restoring conversations. This is an alternative approach to issue #1451.

### What Was Happening on Main

On `main`, when restoring a conversation, `resolve_diff_from_deserialized` would:

**Override from runtime:**
- `agent_context` (skills, system_message_suffix, user_message_suffix, secrets)
- `llm` secrets (api_key, aws credentials, litellm_extra_body)
- `condenser.llm` secrets

**Restore from persistence (and require exact match with runtime):**
- `tools`
- `mcp_config`
- `filter_tools_regex`
- `system_prompt_filename`
- `security_policy_filename`
- `system_prompt_kwargs`
- `condenser` (except its llm secrets)
- `llm` config (model, temperature, etc.)

The final equality check meant users effectively couldn't change most Agent configuration between sessions.

### What This PR Does

**Removes reconciliation.** The provided Agent is used directly - subject to limitations that would otherwise not work at all, such as, it has to be the same Agent class, or the same tools.

Users are now free to change Agent configuration between sessions:
- `llm` (model, api_key, all settings)
- `mcp_config`
- `filter_tools_regex`
- `agent_context`
- `system_prompt_filename`
- `security_policy_filename`
- `system_prompt_kwargs`
- `condenser`

#### Limitations:
- `tools`
- Agent's class/type
- non-Agent `state` attributes like Confirmation Policy

#### New Conversation:
1. Create ConversationState with the provided Agent (Pydantic validation happens here)
2. Initialize EventLog for event storage
3. Save initial base state to persistence
4. Return the new state

#### Restored Conversation:
1. Load persisted base_state.json (to get conversation metadata)
2. Verify conversation ID matches
3. Create ConversationState with the provided Agent (Pydantic validation happens here)
4. Restore persisted conversation metadata (execution_status, confirmation_policy, etc.)
5. Attach EventLog to load persisted events
6. Save updated base state (with the provided Agent)
7. Return the resumed state

### Validation

Pydantic validation happens when creating instances (LLM, Agent, ConversationState) via the constructor.

### Note on Tools

From issue #1533 , tools already used in the conversation history are still available.

### Scope: LocalConversation Only

This PR only affects **LocalConversation**. 

For **RemoteConversation**, the server always creates the Agent from the persisted `meta.json` - the client's Agent is ignored when restoring. Making RemoteConversation support Agent changes would require:
1. Client sends new Agent config when attaching to existing conversation
2. Server accepts and uses the new Agent config instead of persisted one

This is out of scope for this PR but could be a follow-up.

Closes #1451

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:521d39f-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-521d39f-python \
  ghcr.io/openhands/agent-server:521d39f-python
```

```
ghcr.io/openhands/agent-server:521d39f-golang-amd64
ghcr.io/openhands/agent-server:521d39f-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:521d39f-golang-arm64
ghcr.io/openhands/agent-server:521d39f-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:521d39f-java-amd64
ghcr.io/openhands/agent-server:521d39f-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:521d39f-java-arm64
ghcr.io/openhands/agent-server:521d39f-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:521d39f-python-amd64
ghcr.io/openhands/agent-server:521d39f-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:521d39f-python-arm64
ghcr.io/openhands/agent-server:521d39f-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:521d39f-golang
ghcr.io/openhands/agent-server:521d39f-java
ghcr.io/openhands/agent-server:521d39f-python
```

- Each variant tag (e.g., `521d39f-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `521d39f-python-amd64`) are also available if needed

### DIFF
--- a/.github/workflows/run-eval.yml
+++ b/.github/workflows/run-eval.yml
@@ -1,5 +1,6 @@
 ---
 name: Run Eval
+run-name: Run Eval (${{ inputs.benchmark || 'swebench' }}) ${{ inputs.reason || github.event.label.name || 'release' }}
 
 on:
     pull_request_target:


openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -300,72 +299,6 @@ def step(
...
-        reconciled = persisted.model_copy(update=updates)
-        if self.model_dump(exclude_none=True) != reconciled.model_dump(
-            exclude_none=True
-        ):
-            raise ValueError(
-                "The Agent provided is different from the one in persisted state.\n"
-                f"Diff: {pretty_pydantic_diff(self, reconciled)}"
-            )
-        return reconciled
-
```
xingyaoww, line 406: To me, it seems the whole goal of removing this "resolve_diff" is to increase the flexibility of dynamic configuration. One reason that i want to keep resolve diff function is that it give us control over what gets merged and what's not. IRRC, The real root cause that's causing us trouble was that we had a too strict default -- In this case, how about we simply relax the logic in `resolve_diff_from_deserialized` and allow overwritten most of the config, and log accordingly? 

In practice, it will behave the same as this PR, just with fewer code changes.

It will also make it easier to implement https://github.com/OpenHands/software-agent-sdk/issues/1533, since one only need to add a few lines in this consolidation function to tighten up the loading of particular attributes?

WDYT?

enyst: That makes sense, to be honest I’d prefer to clean up the code from the special cases excepted from the rules, and rebuild anew. But yes, for tools it makes sense and @ixchio has the PR for a middle ground there.

Maybe we can restore it but name it `load` or `create_from (persistent_agent)`?

@openhands lets try to reintroduce the code restricting the restore path for agent’s tools to the same tools as those persistent, keep the same restriction as it was in the reconcile method, name the method `load` or `create_from` according to what it does; and ultrathink the `conversationstate` code to work nicely on new and restored paths.

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -191,26 +216,42 @@ def create(
...
+            if persisted_state.id != id:
                 raise ValueError(
                     f"Conversation ID mismatch: provided {id}, "
-                    f"but persisted state has {state.id}"
+                    f"but persisted state has {persisted_state.id}"
                 )
 
-            # Reconcile agent config with deserialized one
-            resolved = agent.resolve_diff_from_deserialized(state.agent)
+            # Validate tools match - they may have been used in conversation history.
```
enyst:
```suggestion
            # Limit to the tools that match - they may have been used in conversation history.
```

openhands-sdk/openhands/sdk/agent/base.py
xingyaoww: > Maybe we can restore it but name it load or create_from (persistent_agent)?

Sounds great!

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -191,26 +216,42 @@ def create(
...
                     f"Conversation ID mismatch: provided {id}, "
-                    f"but persisted state has {state.id}"
+                    f"but persisted state has {persisted_state.id}"
                 )
 
-            # Reconcile agent config with deserialized one
-            resolved = agent.resolve_diff_from_deserialized(state.agent)
+            # Limit to the tools that match - they may have been used in conversation history.
+            # All other config (LLM, agent_context, condenser, etc.) can change freely.
+            agent.load(persisted_state.agent)
```
xingyaoww: Shall we get the return value of this function?

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -191,26 +216,42 @@ def create(
...
+                stuck_detection=stuck_detection,
+                # Restore persisted conversation metadata
+                execution_status=persisted_state.execution_status,
+                confirmation_policy=persisted_state.confirmation_policy,
+                security_analyzer=persisted_state.security_analyzer,
+                activated_knowledge_skills=persisted_state.activated_knowledge_skills,
+                blocked_actions=persisted_state.blocked_actions,
+                blocked_messages=persisted_state.blocked_messages,
+                secret_registry=persisted_state.secret_registry,
+            )
```
xingyaoww: Instead of creating this from scratch, can we just explicitly overwrite the few keys for `persisted_states`, in case we add more fields in the future but forgot to add a new key here

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -300,49 +299,36 @@ def step(
         NOTE: state will be mutated in-place.
         """
 
-    def resolve_diff_from_deserialized(self, persisted: "AgentBase") -> "AgentBase":
-        """
-        Return a new AgentBase instance equivalent to `persisted` but with
-        explicitly whitelisted fields (e.g. api_key) taken from `self`.
+    def load(self, persisted: "AgentBase") -> "AgentBase":
+        """Validate that this agent can resume a conversation from persisted state.
```
enyst:
```suggestion
        """Verify that we can resume this agent from persisted state.
```

openhands-sdk/openhands/sdk/conversation/state.py
enyst:
```suggestion
            verified_agent = agent.load(persisted_state.agent)
```
TBH the method currently is a "verify" rather than a "load"

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -193,22 +218,27 @@ def create(
...
+            # Attach runtime handles
             state._fs = file_store
             state._events = EventLog(file_store, dir_path=EVENTS_DIR)
             state._autosave_enabled = True
-            state.agent = resolved
 
+            # Override with runtime-provided values
+            state.agent = verified_agent
+            state.workspace = workspace
+            state.max_iterations = max_iterations
```
enyst, line 241: @xingyaoww how does this look like? So we took all from persistent state: except for agent, where we just enforce some limitations; for max iterations, the idea is that if the user set "ok 700 from now on", it will be applied; as for workspace, idk, WDYT?
I'm not sure I understand why it was _from persisted state_, on main. Wouldn't that prevent "teleporting" a conversation to remote and back to local?

enyst: Ah, maybe I'm confused. @openhands what exactly does the "workspace" variable here in `ConversationState` refer to, and what possible values it has? Trace all execution paths it's on.

openhands-sdk/openhands/sdk/conversation/state.py
enyst: Done!

openhands-sdk/openhands/sdk/conversation/state.py
enyst: Done! Currently we don't return something else than `self` or `ValueError`, but just in case.

enyst:
@openhands I look at the diff of this PR and I don't see all I expected to see. For example, in llm.py we have a class var named OVERRIDE... something, which is, I think, only used in the reconciliation method we removed from LLM. Please track it down and confirm what I said, and if it's unused let's clean it up.

Verify if in the other files in this PR, e.g. state or agentbase, we have something similar and do the same.

Review the code and clean it up from such redundancies or similar.

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -193,22 +221,27 @@ def create(
         if base_text:
             state = cls.model_validate(json.loads(base_text))
 
-            # Enforce conversation id match
             if state.id != id:
```
enyst, line 224:
```suggestion
            # Restore the conversation with the same id
            if state.id != id:
```

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -193,22 +221,28 @@ def create(
...
                 raise ValueError(
                     f"Conversation ID mismatch: provided {id}, "
                     f"but persisted state has {state.id}"
                 )
 
-            # Reconcile agent config with deserialized one
-            resolved = agent.resolve_diff_from_deserialized(state.agent)
+            # Verify tools match - they may have been used in conversation history.
+            # All other config (LLM, agent_context, condenser, etc.) can change freely.
```
enyst:
```suggestion
```

enyst:
@openhands Understand the goal that this PR aims to achieve. We have two tests files, modified in this PR, which are no longer named appropriately since we removed reconciliation and replace it with "verification". Rename test_agent_reconciliation.py to test_agent_loading.py (or verification if you prefer, but maybe loading will do, wdyt?)

Review the tests in tests/sdk/llm/test_llm_reconciliation.py and if they're not duplicated in other tests/sdk llm-related files and they're still useful (are they?), then move them in a test llm file.

openhands-sdk/openhands/sdk/conversation/state.py
xingyaoww: Looks good!

enyst:
@openhands Understand the goal of this PR and its diff. We need to merge main into it, and fix the conflicts. Find the recent commit that introduced the conflicts (it's very recent, last five or so) and understand its goal too, so we can keep that behavior as well as this one.

Then do it, merge and fix.

tests/integration/tests/t10_restore_conversation.py
```diff
@@ -0,0 +1,178 @@
...
+                    f"got {conv2.state.execution_status!r}"
+                ),
+            )
+        if not conv2.state.confirmation_policy.should_confirm():
+            return TestResult(
+                success=False,
+                reason="confirmation_policy was not restored from persistence",
+            )
+
+        # 2) Runtime agent/LLM should be used (no reconciliation).
```
enyst:
```suggestion
        # 2) Runtime agent/LLM should be used.
```

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -366,43 +353,23 @@ def resolve_diff_from_deserialized(
                 if isinstance(event, ActionEvent) and event.tool_name
             }
 
-            # Only require tools that were actually used in history
             missing_used_tools = used_tools - runtime_names
```
enyst, line 357:
```suggestion
            # Only require tools that were actually used in history
            missing_used_tools = used_tools - runtime_names
```

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -236,13 +264,12 @@ def create(
...
             stuck_detection=stuck_detection,
         )
-        # Record existing analyzer configuration in state
         state.security_analyzer = state.security_analyzer
         state._fs = file_store
         state._events = EventLog(file_store, dir_path=EVENTS_DIR)
         state.stats = ConversationStats()
 
-        state._save_base_state(file_store)  # initial snapshot
+        state._save_base_state(file_store)
```
enyst:
```suggestion
        state._save_base_state(file_store)  # initial snapshot
```

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -236,7 +264,6 @@ def create(
             max_iterations=max_iterations,
             stuck_detection=stuck_detection,
         )
-        # Record existing analyzer configuration in state
         state.security_analyzer = state.security_analyzer
```
enyst:
```suggestion
```
This is weird. I could just delete it 🤔

tests/cross/test_agent_loading.py
```diff
@@ -106,6 +106,11 @@ def test_conversation_restarted_with_changed_working_directory(tmp_path_factory)
     )
 
 
+# # Tests for agent tools restriction and LLM flexibility
+def test_conversation_with_different_agent_tools_fails():
+    """Test that using an agent with different tools fails (tools must match)."""
+
+
```
enyst:
```suggestion
```

tests/cross/test_agent_loading.py
```diff
@@ -106,6 +106,7 @@ def test_conversation_restarted_with_changed_working_directory(tmp_path_factory)
     )
 
 
+# # Tests for agent tools restriction and LLM flexibility
 # Tests from test_local_conversation_tools_integration.py
```
enyst:
```suggestion
# Tests for agent tools restriction and LLM flexibility
```

tests/cross/test_agent_loading.py
```diff
@@ -458,6 +452,75 @@ def test_conversation_persistence_lifecycle(mock_completion):
...
 
 
+def test_conversation_resume_overrides_agent_llm_but_preserves_state_settings():
+    """Test resume behavior when changing runtime Agent/LLM settings.
+
+    Expectations:
+    - Some conversation *state* settings are persisted and should not be overridden
+      on resume (e.g., confirmation_policy, execution_status).
+    - Agent/LLM settings should come from the runtime-provided Agent on resume
+      (no reconciliation).
```
enyst:
```suggestion
```

tests/integration/tests/t10_restore_conversation.py
```diff
@@ -0,0 +1,182 @@
+"""Test conversation restore (resume) behavior.
+
+This integration test exercises the key behavior of PR #1542:
+- On resume, we use the runtime-provided Agent directly (no reconciliation).
```
enyst:
```suggestion
- On resume, we use the runtime-provided Agent.
```

tests/integration/tests/t10_restore_conversation.py
```diff
@@ -0,0 +1,196 @@
...
+    ConversationExecutionStatus,
+)
+from openhands.sdk.llm import LLM
+from openhands.sdk.security.confirmation_policy import AlwaysConfirm
+from openhands.sdk.tool import Tool, register_tool
+from openhands.tools.terminal import TerminalTool
+from tests.integration.base import BaseIntegrationTest, TestResult
+
+
+INSTRUCTION = "Create a new conversation."  # Not used; we validate restore behavior.
```
enyst:
```suggestion
INSTRUCTION = "Create a new conversation."
```

enyst (commented):
@xingyaoww We made here both `cross` test and `integration test`. Both ran and worked repeatedly in agent's workspace.

I'm not sure about the integration test TBH. 🤔 Verifying the correct behavior on conversation restore doesn't really depend on LLM; so running it repeatedly for 6 LLMs seems a bit ... not so useful perhaps? What it needs is at most 2 in use during the test itself.

**Options:**

**1. Two real LLMs**
if we want real LLM calls, and with 2 LLMs, the agent suggested something like:

  - Add a new job in .github/workflows/integration-runner.yml, something like `run-restore-test-once`:
    - runs only when label is integration-test (and maybe schedule / dispatch)
    - sets env vars:
        - LLM_CONFIG_1 = first model config from matrix (e.g. Sonnet)
        - LLM_CONFIG_2 = second model config from matrix (e.g. GPT-5.1-codex-max)
    - runs:
        - `uv run python tests/integration/run_infer.py --eval-ids t10_restore_conversation --llm-config "$LLM_CONFIG_1" ...`
        - and passes the “second model” to the test via an env var (e.g. `RESTORE_LLM_CONFIG_2`), or via a dedicated CLI arg if we extend `run_infer.py`.
  - In t10, stop hardcoding model names; instead:
    - use self.llm for llm1 (this is provided by --llm-config)
    - use the second config from env/arg for llm2 (the “other model”)
    - keep using the same LLM_API_KEY / LLM_BASE_URL from CI secrets.

**2. One real LLM**
 - alternatively, if we want with a single LLM, but still actually run `send_message` for real, we can do only the first part above, and avoid the roundabout with the `RESTORE_LLM_CONFIG_2` part

**3. Not run real LLMs**
- run a `cross` test verifying the restore conversation behavior. (the right attributes, some from persisted, some from runtime)

WDYT?

openhands-sdk/openhands/sdk/conversation/state.py
xingyaoww: yep!

xingyaoww:
@enyst hmm, i would hope to have two real llms :), but i do agree with you that we might not need to run it for all the model matrix we have.

Maybe for the context of getting this PR merged, we can do 3 for now? I'm hoping i can include this PR in the next minor release :)

enyst:
@openhands Please understand the goal of this PR, and merge main into it, fix the conflicts.

Note: look at the very recent commit on main that lead to the conflicts, understand it, so you can make sure to keep it and keep the tests accurate and complete - we want good tests!

tests/sdk/conversation/local/test_state_serialization.py
```diff
@@ -594,46 +748,59 @@ def test_conversation_state_stats_preserved_on_resume():
             response_id="test-response-1",
         )
 
-        # Verify stats are set correctly before saving
         combined_metrics = state.stats.get_combined_metrics()
```
enyst, line 753:
```suggestion
        # Verify stats are set correctly before saving
        combined_metrics = state.stats.get_combined_metrics()
```

tests/sdk/conversation/local/test_state_serialization.py
```diff
@@ -579,10 +735,8 @@ def test_conversation_state_stats_preserved_on_resume():
             id=conv_id,
         )
 
-        # Register the LLM and add some metrics
         state.stats.register_llm(RegistryEvent(llm=llm))
 
-        # Add token usage with context_window
         assert llm.metrics is not None
```
enyst, line 741:
```suggestion
        # Add token usage with context_window
        assert llm.metrics is not None
```

enyst:
@openhands Understand this PR. Then split the integration test changes from it, in a new branch based on this branch, and make a PR to this PR.

We want to get those right, but without running it with the whole LLM matrix, and in the same time, we want this branch ready.

enyst:
> @enyst hmm, i would hope to have two real llms :), but i do agree with you that we might not need to run it for all the model matrix we have.
> 
> Maybe for the context of getting this PR merged, we can do 3 for now? I'm hoping i can include this PR in the next minor release :)

@xingyaoww 
- I adjusted `cross` tests to do more, test also another detail. I do think the code should be ready for review.
- I also tried running separately "2 LLMs" option in integration tests, but that didn't work, it requires more testing. I split it into its own branch, in case we want to merge this one.
- To note, the original integration test with 2 LLMs worked well. It was restoring for real and continued conversation with another event. Need to figure out how to run it in github actions correctly without the matrix.
- I'm meanwhile trying to test manually with a CLI based on this branch. 🤔

xingyaoww:
@OpenHands please create an issue in OpenHands/OpenHands-CLI - based on the change of this PR, we should update how we handle `agent_settings` systematically in CLI.

We should only keep the non-default fields in `agent_settings` for CLI (instead of saving every field like `.model_dump()` and always override them - and keep everything else as default.

enyst:
This is the CLI on this branch, restoring with another LLM, remembering the events, and it did also delete that file:
image

PR #1547, opened 2025-12-30
feat(sdk): complete hooks implementation with additional context and stop hook

### LINKED ISSUE #1527: Feature Request: Add Stop Hook to prevent early agent termination
## Summary

Add a **Stop Hook** that triggers when the agent decides to stop/terminate, allowing users to evaluate whether the agent should actually stop or continue working.

## Motivation

This feature addresses the "early exit" or "lazy agent" problem where agents may decide to stop before all requested work is complete. A Stop Hook would allow:

1. **Intelligent evaluation** of whether all tasks are truly complete
2. **Automatic continuation** if outstanding work remains
3. **Better agent engagement** for longer, more complex tasks

## Reference Implementation

Claude Code has implemented this feature with their [Stop Hook](https://code.claude.com/docs/en/hooks#example%3A-intelligent-stop-hook):

```json
{
  "hooks": {
    "Stop": [
      {
        "hooks": [
          {
            "type": "prompt",
            "prompt": "Evaluate if Claude should stop working. Context: $ARGUMENTS\n\nAnalyze the conversation and determine if:\n1. All user-requested tasks are complete\n2. Any errors need to be addressed\n3. Follow-up work is needed\n\nRespond with JSON: {\"decision\": \"approve\" or \"block\", \"reason\": \"your explanation\"}"
          }
        ]
      }
    ]
  }
}
```

## Current State

OpenHands has a hooks system ([docs](https://docs.openhands.dev/sdk/guides/hooks#hooks)), but there is no explicit "stop" event type in the [event types](https://docs.openhands.dev/sdk/arch/events#event-types).

Currently, `MessageEvent (Agent)` may indicate conversation termination, but this is not explicit and doesn't provide a clear mechanism to:
- Intercept the stop decision
- Evaluate if work is complete
- Block the stop and continue working

## Proposed Solution

Add a new hook event type (e.g., `AgentStop` or `Stop`) that:

1. **Triggers** when the agent decides to stop/terminate
2. **Provides context** about the current state (conversation history, completed actions, etc.)
3. **Allows blocking** the stop action to keep the agent working
4. **Supports both command and prompt-based hooks** for flexibility

## Use Cases

- Ensuring all items in a task list are completed before stopping
- Verifying test suites pass before considering work done
- Checking that all requested files have been modified
- Preventing premature termination during complex refactoring tasks

## Related Discussion

This was mentioned in the OpenHands webinar on "automating massive refactors" as a way to address agents "being lazy" and exiting early.

ak684:
follow up to: https://github.com/OpenHands/software-agent-sdk/pull/1467

fixes #1527 

- Implement additional_context injection in UserPromptSubmit hooks
  - Hook context is appended to MessageEvent.extended_content
  - Context is appended to MessageEvent.extended_content and included in LLM messages

- Implement stop hook integration in conversation run loop
  - Stop hooks can deny premature agent completion
  - Feedback from hooks is injected as a user message
  - Agent continues running after stop hook denial

- Add comprehensive tests for both features
  - Tests for context appearing in extended_content
  - Tests for context appearing in to_llm_message() output
  - Tests for stop hook denial with feedback injection
  - Integration tests for full conversation loop with stop hooks

- Add advanced hooks example (34_hooks_advanced.py)
  - Demonstrates UserPromptSubmit with additional_context
  - Demonstrates Stop hook with conditional denial

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:ab9d4ec-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-ab9d4ec-python \
  ghcr.io/openhands/agent-server:ab9d4ec-python
```

```
ghcr.io/openhands/agent-server:ab9d4ec-golang-amd64
ghcr.io/openhands/agent-server:ab9d4ec-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:ab9d4ec-golang-arm64
ghcr.io/openhands/agent-server:ab9d4ec-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:ab9d4ec-java-amd64
ghcr.io/openhands/agent-server:ab9d4ec-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:ab9d4ec-java-arm64
ghcr.io/openhands/agent-server:ab9d4ec-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:ab9d4ec-python-amd64
ghcr.io/openhands/agent-server:ab9d4ec-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:ab9d4ec-python-arm64
ghcr.io/openhands/agent-server:ab9d4ec-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:ab9d4ec-golang
ghcr.io/openhands/agent-server:ab9d4ec-java
ghcr.io/openhands/agent-server:ab9d4ec-python
```

- Each variant tag (e.g., `ab9d4ec-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `ab9d4ec-python-amd64`) are also available if needed


rbren:
This likely solves https://github.com/OpenHands/software-agent-sdk/issues/1527!

examples/01_standalone_sdk/34_hooks_advanced.py
xingyaoww, line 1: How do we feel about merging this with our existing hook example? We can turn the hook example into a folder, and put these scripts directly into that folder

examples/01_standalone_sdk/34_hooks_advanced.py
```diff
@@ -0,0 +1,155 @@
...
+        print("\n" + "=" * 60)
+        print("Example Complete!")
+        print("=" * 60)
+
+        cost = conversation.conversation_stats.get_combined_metrics().accumulated_cost
+        print(f"\nEXAMPLE_COST: {cost}")
+
+
+if __name__ == "__main__":
+    main()
```
xingyaoww: I'd throw away the "main" function and directly write the script out - fewer indentation & easier to understand

examples/01_standalone_sdk/34_hooks_advanced.py
```diff
@@ -0,0 +1,155 @@
...
+                    }
+                ],
+                "Stop": [
+                    {
+                        "hooks": [{"type": "command", "command": str(stop_script)}],
+                    }
+                ],
+            }
+        }
+    )
```
xingyaoww: NIT: could we make this a HookConfig a pydantic class that contains `UserPromptSubmit` `Stop`, etc, which are other pydantic class, so we can have stricter type check instead of putting together a dict like this?

examples/01_standalone_sdk/34_hooks_advanced.py
ak684: done ✅

examples/01_standalone_sdk/34_hooks_advanced.py
ak684: I think this is a great idea. I’ll plan on implementing it in a follow-up PR so that we can keep this PR’s scope focused.

examples/01_standalone_sdk/33_hooks/hook_scripts/block_dangerous.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# PreToolUse hook: Block dangerous rm -rf commands
+# Uses jq for JSON parsing (needed for nested fields like tool_input.command)
+
+input=$(cat)
+command=$(echo "$input" | jq -r '.tool_input.command // ""')
```
xingyaoww: If `jq` is not installed on user's system, will we surface the error message to the user, or will we just fail silently?

examples/01_standalone_sdk/33_hooks/33_hooks.py
xingyaoww, line 1: @OpenHands please send a PR to OpenHands/docs repo to update the old 33_hooks.py to this new one

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
```diff
@@ -51,11 +55,11 @@ def on_event(self, event: Event) -> None:
 
         # Run UserPromptSubmit hooks for user messages
         if isinstance(event, MessageEvent) and event.source == "user":
-            self._handle_user_prompt_submit(event)
+            callback_event = self._handle_user_prompt_submit(event)
```
xingyaoww, line 58: @OpenHands, does this mean we will modify the event passed to the callback? Does it break the pattern we currently have for `on_event`? Ideally we should never mutate event passed to `on_event`, please investigate. Do not modify any code

xingyaoww: > However, there's an integration issue: in LocalConversation.__init__, create_hook_callback is called without passing original_callback, so it defaults to None. This means the modified event is created but never actually passed to downstream callbacks - it's lost.

@ak684 is this intentional?

examples/01_standalone_sdk/33_hooks/hook_scripts/block_dangerous.sh
ak684: we’ll surface the error but i decided to replace jq with grep which should work better out of the box

openhands-sdk/openhands/sdk/hooks/conversation_hooks.py
ak684: updated to pass to callback, also added test

enyst: @xingyaoww I wonder WDYT about this solution?

@ak684 
Please let me take a step back here and look at our options. We have the goal to inject some extra content in the user message, just like a keyword-skill.

1. modify the event
2. **add a new event (this PR’s solution)**
3. intervene before the user message event is created

1 and 2 are after the event was created, and saved in EventLog (I think it was saved too?)

For 2, we then have a question:
  a. do we notify listeners about the new event
  b. or about both?

Also for 2, I think we have another question that I’m most interested in:
  - do we send to the LLM only the second, or do we send both?
  - (IMHO we should send only the second, since it seems to me it contains all `content` of the first, too; or make it have only the `extended content`)

ak684: So, my understanding of the code in this PR is that the hook processor intercepts events before they reach EventLog.append(), so only one event is ever persisted and sent to the LLM: either the original "no modification needed" event, or the replacement event with extended context.

The test `test_modified_events_with_additional_context_persisted` for example asserts there's exactly one MessageEvent in EventLog, and that event contains the hook-injected context. In this way @enyst, the current approach may actually trend closer to your option 3 (intervene before save) than option 2 (add a new event after save).

Here's the flow I'm tracing:

1. send_message() creates the MessageEvent
2. It goes to _on_event (the hook processor, NOT EventLog)
3. Hook processor runs, potentially creates a replacement with extended_content
4. Only then does the (possibly replaced) event flow to _default_callback which calls EventLog.append()

I'm open to other approaches though! If you or @xingyaoww have a specific alternative pattern you'd prefer to see, please let me know so I can implement it and we can move forward.

enyst: Ahh, thank you, I like this! I'd love to know what Xingyao thinks about it.

xingyaoww: > It goes to _on_event (the hook processor, NOT EventLog)
> Hook processor runs, potentially creates a replacement with extended_content

The high level idea looks good to me -- The only nit here is that i'd hope we name it differently: not `_on_event` because it confuses me w/ the `on_event` callback that should not modify the event. Maybe we can call it something like `event_hook_fn`?

xingyaoww: Actually, `self._hook_processor, self._on_event = create_hook_callback(`, also works for me -- feel free to disregard my previous comment

xingyaoww (approved):
LGTM! Thank you!

xingyaoww:
@ak684 
PR looks good to me - just need to get these doc merged as well https://github.com/OpenHands/docs/pull/230

Issue #1518, opened 2025-12-28
[Bug] Condensation for 0 events


xingyaoww:
I got these errors:

```
{"kind": "Condensation", "id": "1dd50966-6311-4337-8cc8-f20f8ce3f41d", "timestamp": "2025-12-28T18:28:51.883055", "source": "environment", "forgotten_event_ids": [], "summary": "I don't see any events provided to summarize. Please share the list of events that need to be summarized, and I'll create a context-aware state summary following the format guidelines.\n\nIf you're starting a new session, please provide:\n1. The events/actions taken by the agent\n2. Any relevant context about the task being performed\n\nI'll then generate an appropriate summary tracking user context, task status, and any relevant state information.", "summary_offset": 122, "llm_response_id": "chatcmpl-8fb35e47-ca68-4f72-bd13-bc38a07d7a94"}
```

DM me for the covnersation event dumps

xingyaoww:
## Root Cause Analysis

After analyzing the event dump, I've identified the root cause of this bug.

### The Problem

When a **tool loop** (a sequence of ActionEvent/ObservationEvent pairs starting with thinking blocks) spans almost the entire view, the `manipulation_indices` computed by `View` leave no valid range for forgetting events.

From the event dump:
- **Total events before first condensation**: 122
- **Event structure**:
  - Event 0: `SystemPromptEvent`
  - Event 1: `MessageEvent` (user message)
  - Events 2-121: A continuous tool loop (60 action-observation pairs)

The tool loop starts at index 2 (first `ActionEvent` with thinking blocks) and continues uninterrupted to index 121.

### Manipulation Indices

The `View.manipulation_indices` property computes safe boundaries for forgetting events while preserving atomic units. For this event sequence:

```
manipulation_indices = [0, 1, 2, 122]
```

This means:
- Index 0: Before SystemPromptEvent
- Index 1: Before MessageEvent  
- Index 2: Start of tool loop
- Index 122: End of tool loop (and end of view)

### Condensation Logic Failure

When `LLMSummarizingCondenser._get_forgotten_events()` runs with `keep_first=4`:

1. `forgetting_start = find_next_manipulation_index(keep_first=4, strict=True)` → Returns `122` (smallest index > 4)
2. `forgetting_end = find_next_manipulation_index(naive_end=67, strict=False)` → Returns `122` (smallest index >= 67)
3. `forgotten_events = view[122:122]` → **Empty list!**

[...]

PR #1577, opened 2026-01-02
fix(condenser): Retry on empty condensation

csmith49:
## Summary

Adds a graceful failure mechanism when a condenser doesn't forget any events. Instead of throwing an uncaught exception, we now throw a `NoCondensationAvailableException` which is caught by the base condenser class and returns the view provided.

Notably, this means if we cannot find a condensation (because, e.g., the entirety of the message history is a tool loop) then the condensation will be effectively delayed until a valid chunk of events can be forgotten.

Also updates the condenser integration test to report the number of events condensed in the first condensation, adds unit tests for the exception handling, and updates a handful of other related tests.

Intended to address #1518.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:22eb6f2-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-22eb6f2-python \
  ghcr.io/openhands/agent-server:22eb6f2-python
```

```
ghcr.io/openhands/agent-server:22eb6f2-golang-amd64
ghcr.io/openhands/agent-server:22eb6f2-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:22eb6f2-golang-arm64
ghcr.io/openhands/agent-server:22eb6f2-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:22eb6f2-java-amd64
ghcr.io/openhands/agent-server:22eb6f2-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:22eb6f2-java-arm64
ghcr.io/openhands/agent-server:22eb6f2-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:22eb6f2-python-amd64
ghcr.io/openhands/agent-server:22eb6f2-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:22eb6f2-python-arm64
ghcr.io/openhands/agent-server:22eb6f2-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:22eb6f2-golang
ghcr.io/openhands/agent-server:22eb6f2-java
ghcr.io/openhands/agent-server:22eb6f2-python
```

- Each variant tag (e.g., `22eb6f2-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `22eb6f2-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -489,7 +489,23 @@ class RemoteConversation(BaseConversation):
         self._hook_processor = None
         self._cleanup_initiated = False
 
-        if conversation_id is None:
+        should_create = conversation_id is None
+        if conversation_id is not None:
+            # Try to attach to existing conversation
+            resp = _send_request(
+                self._client,
+                "GET",
+                f"/api/conversations/{conversation_id}",
+                acceptable_status_codes={404},
+            )
+            if resp.status_code == 404:
+                # Conversation doesn't exist, we'll create it
+                should_create = True
+            else:
+                # Conversation exists, use the provided ID
+                self._id = conversation_id
+
+        if should_create:
             # Import here to avoid circular imports
             from openhands.sdk.tool.registry import get_tool_module_qualnames
 
@@ -518,6 +534,9 @@ class RemoteConversation(BaseConversation):
                 else:
                     threshold_config = stuck_detection_thresholds
                 payload["stuck_detection_thresholds"] = threshold_config.model_dump()
+            # Include conversation_id if provided (for creating with specific ID)
+            if conversation_id is not None:
+                payload["conversation_id"] = str(conversation_id)
             resp = _send_request(
                 self._client, "POST", "/api/conversations", json=payload
             )
@@ -529,11 +548,6 @@ class RemoteConversation(BaseConversation):
                     "Invalid response from server: missing conversation id"
                 )
             self._id = uuid.UUID(cid)
-        else:
-            # Attach to existing
-            self._id = conversation_id
-            # Validate it exists
-            _send_request(self._client, "GET", f"/api/conversations/{self._id}")
 
         # Initialize the remote state
         self._state = RemoteState(self._client, str(self._id))
--- a/tests/sdk/conversation/remote/test_remote_conversation.py
+++ b/tests/sdk/conversation/remote/test_remote_conversation.py
@@ -53,6 +53,7 @@ class TestRemoteConversation:
                 # Return conversation info response with finished status
                 # (needed for run() polling to complete)
                 response = Mock()
+                response.status_code = 200
                 response.raise_for_status.return_value = None
                 conv_info = mock_conv_response.json.return_value.copy()
                 conv_info["execution_status"] = "finished"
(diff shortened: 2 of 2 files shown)


csmith49:
There are two integration test runs here -- the first one has the token condenser test failing for Devstral, but it's the same failure mode as with GPT 5.1. I'm hesitant to mark it skipped because the failure seems to be intermittent.

csmith49:
IMPORTANT NOTE: This PR is not a complete solution. There are still some strange cases where we can get caught in loops. For example, if a context window exception is thrown we'll end up retrying into infinity.

xingyaoww:
> For example, if a context window exception is thrown we'll end up retrying into infinity.

Could we fail hard in this case? IMO fail hard is better here than going into infinite loop though

csmith49:
> Could we fail hard in this case? IMO fail hard is better here than going into infinite loop though

That's doable. The challenge right now is just figuring out when we're in that situation. I've got some draft changes that make that easy though.

openhands-sdk/openhands/sdk/context/condenser/base.py
```diff
@@ -89,11 +100,41 @@ def get_condensation(
...
+
+                # Check if all condensation requests are continuable. If so, or if there
+                # are no requests, we can safely return the uncondensed view.
+                all_continuable = all(
+                    self.is_continuable_reason(reason)
+                    for reason in view.unhandled_condensation_request_reasons
+                )
+                if not view.unhandled_condensation_request or all_continuable:
+                    return view
+
```
xingyaoww: @OpenHands can you help me explain this code better?

What will happen if there's an unhandled "manual" condensation request AND we run into NoCondensationAvailableException? What about other types?

xingyaoww: @csmith49 
I wonder if we should handle "REQUEST" the same way? When user requests for "/condense", maybe we should let them know when it fails, rather than proceed w/o condensation silently

csmith49: That's a good point. Also catches a potential bug -- the way the `conversation.condense()` function works is by adding a `MANUAL` request and then taking a step. If the conversation starts up again there that's pretty bad.

But it also means these last few commits aren't needed, because we'll treat all `CondensationRequest` events the same. Let me rework the flow so we only continue silently from a failed condensation if it's based on the resource usage limits.

csmith49:
Hey @xingyaoww can you take another look now? Here are the changes from what you saw:

* The `CondensationRequest` event has no more enum -- they're all the same, regardless of whether they come from the agent or from the user (via `LocalConversation.condense`, for example).
* The `RollingCondenser` base class has had the API slightly changed. The function `should_condense` is replaced by `condensation_requirement`, which has return type `CondensationRequirement | None`. The requirements can be either hard or soft and the condenser gets to pick when overriding this function.
    * If the requirement is hard and there's no possible condensation, throw an exception. This is the behavior before this change.
    * If the requirement is soft and there's no possible condensation, just skip it for now.

This flow lets us wait until the manipulation indices are good when we cross the event count threshold but throw an error in pretty much all other cases.

Next step (in a separate PR) is instead condensing _everything_ if we get a hard requirement and no condensation.

xingyaoww (approved):
LGTM!

PR #1584, opened 2026-01-03
Integration test for Opus thinking block constraints

csmith49:
## Summary

Opus 4.5 seems to occasionally respond with a malformed signature error. This PR adds an integration test that reproduces a possible scenario as simply as possible.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:143a01d-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-143a01d-python \
  ghcr.io/openhands/agent-server:143a01d-python
```

```
ghcr.io/openhands/agent-server:143a01d-golang-amd64
ghcr.io/openhands/agent-server:143a01d-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:143a01d-golang-arm64
ghcr.io/openhands/agent-server:143a01d-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:143a01d-java-amd64
ghcr.io/openhands/agent-server:143a01d-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:143a01d-java-arm64
ghcr.io/openhands/agent-server:143a01d-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:143a01d-python-amd64
ghcr.io/openhands/agent-server:143a01d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:143a01d-python-arm64
ghcr.io/openhands/agent-server:143a01d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:143a01d-golang
ghcr.io/openhands/agent-server:143a01d-java
ghcr.io/openhands/agent-server:143a01d-python
```

- Each variant tag (e.g., `143a01d-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `143a01d-python-amd64`) are also available if needed

### DIFF
--- a/tests/integration/tests/t10_condensation_thinking.py
+++ b/tests/integration/tests/t10_condensation_thinking.py
@@ -8,9 +8,11 @@ when earlier thinking blocks are condensed away while later ones remain.
 from itertools import chain
 
 from openhands.sdk import get_logger
-from openhands.sdk.context.condenser import LLMSummarizingCondenser
+from openhands.sdk.context.condenser import CondenserBase
+from openhands.sdk.context.view import View
 from openhands.sdk.event import ActionEvent
 from openhands.sdk.event.condenser import Condensation
+from openhands.sdk.llm import LLM
 from openhands.sdk.llm.utils.model_features import get_features
 from openhands.sdk.tool import Tool, register_tool
 from openhands.tools.terminal import TerminalTool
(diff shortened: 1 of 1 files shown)


tests/integration/tests/t10_condensation_thinking.py
```diff
@@ -0,0 +1,296 @@
...
+
+        # Find the first atomic unit that contains thinking blocks
+        first_thinking_block_unit_idx = None
+        for i in range(len(indices) - 1):
+            start_idx = indices[i]
+            end_idx = indices[i + 1]
+
+            # Check if this atomic unit has any events with thinking blocks
+            for event in view.events[start_idx:end_idx]:
+                if isinstance(event, ActionEvent) and event.thinking_blocks:
```
enyst, line 89: Thank you! I appreciate this, I do wonder though if it belongs as integration test, or maybe something else.

`thinking_blocks` are an Anthropic Claudes' parameter, Gemini 3 has this but they go by different rules (and different from Gemini 2.5 too 😅 ...); Minimax I think has them too. And I can't think of another.

Since it feels very LLM specific, I wonder if maybe a script in scripts/ would work? Alternatively, we have now an examples/ directory named llm_specific/ ... None seems great, idk, WDYT?

csmith49: It probably is too LLM specific to be an integration test. The scaffold is just convenient for condenser tests -- repeatable, no mocked objects, arbitrary LLMs configured from outside the test, etc.

What if we made it like the behavior tests? Using the same infra as the integration tests, but triggered separately and non-blocking. This would basically become `c01_test...` instead.

Would be good to have some batch of tests that stress the condenser across multiple LLMs we can run on occasion to make sure behavior isn't drifting.

enyst: I actually just have a somewhat similar problem: [we want an integration test](https://github.com/OpenHands/software-agent-sdk/pull/1542#issuecomment-3712960525) for conversation restore, where the user does a few actions with an LLM, exit and change to another LLM, then restore the conversation.

The structure of integration tests is almost okay, except that it is actually unnecessary to run all 6 LLMs or so, the test doesn't really depend on them. It just seems to me a bit of a waste, it matters that it "really" works, with real LLMs picking up and continuing the conversation, but it doesn't matter for all matrix. Maybe it matters for more than a pair (we might want a reasoning LLM paired with a non-reasoning one, for fun), but those are... different rules to choose LLM than the current matrix.

I'd call it similar with this, in the sense where `thinking_blocks` are under test, so we know exactly which LLM(s) to run it for, and not run it for others? If that's possible within the integration-test framework... for `C01, C02, ...`

PR #1596, opened 2026-01-05
feat(condenser): Hard context reset on unrecoverable error

csmith49:
## Summary

Extends the rolling condenser API to support hard context resets, and adds one such strategy for the LLM summarizing condenser. This strategy triggers when a hard condensation request comes through but one can't be found.

The quality of the summary and the agent's behavior afterwards is unknown, but we'll be fine-tuning the behavior in future PRs. 

## Changes

* Extends `RollingCondenser` base class to expose `hard_context_reset` -- an optional strategy for last-ditch context management
* Implementation of a hard context reset strategy for `LLMSummarizingCondenser`
* Updating the `c02` condensation integration test to check for the hard context reset instead of the expected error

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:a17b091-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-a17b091-python \
  ghcr.io/openhands/agent-server:a17b091-python
```

```
ghcr.io/openhands/agent-server:a17b091-golang-amd64
ghcr.io/openhands/agent-server:a17b091-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:a17b091-golang-arm64
ghcr.io/openhands/agent-server:a17b091-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:a17b091-java-amd64
ghcr.io/openhands/agent-server:a17b091-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:a17b091-java-arm64
ghcr.io/openhands/agent-server:a17b091-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:a17b091-python-amd64
ghcr.io/openhands/agent-server:a17b091-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:a17b091-python-arm64
ghcr.io/openhands/agent-server:a17b091-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:a17b091-golang
ghcr.io/openhands/agent-server:a17b091-java
ghcr.io/openhands/agent-server:a17b091-python
```

- Each variant tag (e.g., `a17b091-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `a17b091-python-amd64`) are also available if needed

### DIFF
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -114,6 +114,11 @@ MODELS = {
             "temperature": 0.0,
         },
     },
+    "glm-4.7": {
+        "id": "glm-4.7",
+        "display_name": "GLM-4.7",
+        "llm_config": {"model": "litellm_proxy/openrouter/z-ai/glm-4.7"},
+    },
 }
 
 
--- a/openhands-sdk/openhands/sdk/agent/agent.py
+++ b/openhands-sdk/openhands/sdk/agent/agent.py
@@ -67,6 +67,10 @@ from openhands.sdk.tool.builtins import (
 logger = get_logger(__name__)
 maybe_init_laminar()
 
+# Maximum number of events to scan during init_state defensive checks.
+# SystemPromptEvent must appear within this prefix (at index 0 or 1).
+INIT_STATE_PREFIX_SCAN_WINDOW = 3
+
 
 class Agent(AgentBase):
     """Main agent implementation for OpenHands.
@@ -102,53 +106,82 @@ class Agent(AgentBase):
         state: ConversationState,
         on_event: ConversationCallbackType,
     ) -> None:
+        """Initialize conversation state.
+
+        Invariants enforced by this method:
+        - If a SystemPromptEvent is already present, it must be within the first 3
+          events (index 0 or 1 in practice; index 2 is included in the scan window
+          to detect a user message appearing before the system prompt).
+        - A user MessageEvent should not appear before the SystemPromptEvent.
+
+        These invariants keep event ordering predictable for downstream components
+        (condenser, UI, etc.) and also prevent accidentally materializing the full
+        event history during initialization.
+        """
         super().init_state(state, on_event=on_event)
-        # TODO(openhands): we should add test to test this init_state will actually
-        # modify state in-place
 
         # Defensive check: Analyze state to detect unexpected initialization scenarios
         # These checks help diagnose issues related to lazy loading and event ordering
         # See: https://github.com/OpenHands/software-agent-sdk/issues/1785
-        events = list(state.events)
-        has_system_prompt = any(isinstance(e, SystemPromptEvent) for e in events)
-        has_user_message = any(
-            isinstance(e, MessageEvent) and e.source == "user" for e in events
-        )
-        has_any_llm_event = any(isinstance(e, LLMConvertibleEvent) for e in events)
+        #
+        # NOTE: len() is O(1) for EventLog (file-backed implementation).
+        event_count = len(state.events)
 
+        # NOTE: state.events is intentionally an EventsListBase (Sequence-like), not
+        # a plain list. Avoid materializing the full history via list(state.events)
+        # here (conversations can reach 30k+ events).
+        #
+        # Invariant: when init_state is called, SystemPromptEvent (if present) must be
+        # at index 0 or 1.
+        #
+        # Rationale:
+        # - Local conversations start empty and init_state is responsible for adding
+        #   the SystemPromptEvent as the first event.
+        # - Remote conversations may receive an initial ConversationStateUpdateEvent
+        #   from the agent-server immediately after subscription. In a typical remote
+        #   session prefix you may see:
+        #     [ConversationStateUpdateEvent, SystemPromptEvent, MessageEvent, ...]
+        #
+        # We intentionally only inspect the first few events (cheap for both local and
+        # remote) to enforce this invariant.
+        prefix_events = state.events[:INIT_STATE_PREFIX_SCAN_WINDOW]
+
+        has_system_prompt = any(isinstance(e, SystemPromptEvent) for e in prefix_events)
+        has_user_message = any(
+            isinstance(e, MessageEvent) and e.source == "user" for e in prefix_events
+        )
         # Log state for debugging initialization order issues
         logger.debug(
             f"init_state called: conversation_id={state.id}, "
-            f"event_count={len(events)}, "
+            f"event_count={event_count}, "
             f"has_system_prompt={has_system_prompt}, "
-            f"has_user_message={has_user_message}, "
-            f"has_any_llm_event={has_any_llm_event}"
+            f"has_user_message={has_user_message}"
         )
 
         if has_system_prompt:
-            # SystemPromptEvent already exists - this is unexpected during normal flow
-            # but could happen in persistence/resume scenarios
-            logger.warning(
-                f"init_state called but SystemPromptEvent already exists. "
-                f"conversation_id={state.id}, event_count={len(events)}. "
-                f"This may indicate double initialization or a resume scenario."
+            # Restoring/resuming conversations is normal: a system prompt already
+            # present means this conversation was initialized previously.
+            logger.debug(
+                "init_state: SystemPromptEvent already present; skipping init. "
+                f"conversation_id={state.id}, event_count={event_count}."
             )
             return
 
-        # Assert: If there are user messages but no system prompt, something is wrong
-        # The system prompt should always be added before any user messages
+        # Assert: A user message should never appear before the system prompt.
+        #
+        # NOTE: This is a best-effort check based on the first few events only.
+        # Remote conversations can include a ConversationStateUpdateEvent near the
+        # start, so we scan a small prefix window.
         if has_user_message:
-            event_types = [type(e).__name__ for e in events]
+            event_types = [type(e).__name__ for e in prefix_events]
             logger.error(
-                f"init_state: User message exists without SystemPromptEvent! "
-                f"conversation_id={state.id}, events={event_types}"
+                f"init_state: User message found in prefix before SystemPromptEvent! "
+                f"conversation_id={state.id}, prefix_events={event_types}"
             )
-            assert not has_user_message, (
-                f"Unexpected state: User message exists before SystemPromptEvent. "
-                f"conversation_id={state.id}, event_count={len(events)}, "
-                f"event_types={event_types}. "
-                f"This indicates an initialization order bug - init_state should be "
-                f"called before any user messages are added to the conversation."
+            raise AssertionError(
+                "Unexpected state: user message exists before SystemPromptEvent. "
+                f"conversation_id={state.id}, event_count={event_count}, "
+                f"prefix_event_types={event_types}."
             )
 
         # Prepare system message
--- a/openhands-sdk/openhands/sdk/secret/secrets.py
+++ b/openhands-sdk/openhands/sdk/secret/secrets.py
@@ -92,7 +92,19 @@ class LookupSecret(SecretSource):
         return result
 
 
-_SECRET_HEADERS = ["AUTHORIZATION", "COOKIE", "KEY", "SECRET", "SESSION", "TOKEN"]
+# Patterns used for substring matching against header names (case-insensitive).
+# Headers containing any of these patterns will be redacted during serialization.
+# Examples: X-Access-Token, Cookie, Authorization, X-API-Key, X-API-Secret
+_SECRET_HEADERS = [
+    "AUTHORIZATION",
+    "COOKIE",
+    "CREDENTIAL",
+    "KEY",
+    "PASSWORD",
+    "SECRET",
+    "SESSION",
+    "TOKEN",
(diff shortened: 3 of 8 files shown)


csmith49:
This PR looks like it works -- until you try to continue a conversation after a hard reset. The `View` objects can only support a single summary, so things start to break after the first non-hard-reset condensation.

I'm working on an extension to address this issue in a separate PR, and will proceed with this one when that feature is merged.

enyst:
@openhands Do a /codereview-roasted on this PR. Reply with your feedback as a PR comment.

enyst:
## 🔴 Taste Rating: **Needs Improvement**

*"Talk is cheap. Show me the code."* — Well, I've seen the code, and we need to talk.

---

### **[CRITICAL ISSUES]** (Must fix - these break fundamental principles)

**[PR Description] 💀 Shipping Known-Broken Code**
> *"The `View` objects can only support a single summary, so things start to break after the first non-hard-reset condensation."*

You're telling me you're shipping code that **you know breaks** after the first use? This is the software equivalent of selling a car that only works for the first trip. The fact that there's a "separate PR" to fix this doesn't make it acceptable to merge broken code into main. **We don't break userspace.**

**[base.py, Lines 159-177] 🏗️ Complexity Explosion**
```python
elif request == CondensationRequirement.HARD:
    try:
        hard_reset_condensation = self.hard_context_reset(...)
        if hard_reset_condensation is not None:
            return hard_reset_condensation
    except Exception as hard_reset_exception:
        raise hard_reset_exception from e
raise e
```

This is now 3+ levels of nesting with multiple special cases. The original code was clean: try condensation, handle soft/hard requirements. Now we have:
1. Try condensation
2. If exception AND soft → return
3. If exception AND hard → try hard reset
4. If hard reset fails → chain exceptions
5. If hard reset returns None → re-raise original
6. Otherwise re-raise

This is exactly the kind of "add another if statement" approach that creates unmaintainable code. **If you need more than 3 levels of indentation, you're screwed.**

**[PR Description] 🎯 Solving an Imaginary Problem Badly**
> *"The quality of the summary and the agent's behavior afterwards is not likely to be good. This is intended solely as a last-ditch fallback option."*

So we're adding ~700 lines of code (including tests) for a feature that:
1. Doesn't work well
2. Breaks after first use
3. Is a "last-ditch fallback"

If the fallback doesn't actually help users recover, why ship it? **Theory and practice sometimes clash. Theory loses. Every single time.**

---

### **[IMPROVEMENT OPPORTUNITIES]** (Should fix - violates good taste)

**[llm_summarizing_condenser.py, Lines 254-268] 🔇 Unused Parameter**
```python
def hard_context_reset(
    self,
    view: View,
    agent_llm: LLM | None = None,  # noqa: ARG002
) -> Condensation | None:
```

You're accepting `agent_llm` but ignoring it with a `noqa` comment. If you don't need it, don't accept it. If the base class requires it, that's a design smell in the base class. Don't paper over bad design with `noqa`.

**[c02_hard_context_reset.py] 📏 497 Lines for a "Simple Fallback"**

A 497-line test file for a "last-ditch fallback" feature? This test is doing:
- Hard reset verification
- Normal condensation verification  
- Task completion verification
- View structure verification
- Event counting
- Summary content validation

If your test needs to verify this many things, your feature is too complex. The test complexity is a code smell for the feature complexity.

**[base.py, Line 103-119] 📝 Docstring Promises What Code Can't Deliver**
```python
"""Perform a hard context reset, if supported by the condenser.
...
This method is invoked when:
- A HARD condensation requirement is triggered
- But the condenser raises a NoCondensationAvailableException error
"""
```

The docstring makes this sound like a clean recovery mechanism. The PR description admits it's a broken hack. Pick one.

---

### **[STYLE NOTES]** (Minor - only mention if genuinely important)

**[test_rolling_condenser.py, Lines 158-238] ✅ Good Test Structure**

Credit where due: the unit tests for the base class behavior are well-structured and test the right things. The mock implementations are clean.

**[base.py, Line 103] ✅ Good Default Behavior**

Returning `None` by default and letting subclasses opt-in is the right pattern. At least this part shows good taste.

---

## **VERDICT:**
❌ **Needs rework**: Fundamental design issues must be addressed first

The author's own comments admit this feature:
1. Breaks after first use
2. Produces poor quality results
3. Is a "last-ditch fallback"

Don't merge code that you know is broken with a promise to fix it later. Fix it first, or don't ship it.

---

## **KEY INSIGHT:**
This PR adds complexity to handle a failure case, but the handling itself is acknowledged to fail. That's not a solution—that's moving the problem.

*"Bad programmers worry about the code. Good programmers worry about data structures and their relationships."* — The real fix is making `View` support multiple summaries, not adding a broken fallback for when it can't.

enyst:
Sorry @csmith49 , I hadn't seen your last comment! 

I was wondering if this PR can indirectly help with this:
- https://github.com/OpenHands/software-agent-sdk/issues/1782

That is, I was wondering if we could go in the opposite direction of the discussion there: use a condensation intently when an error like that, which makes the conversation unusable, happens. We would just continue with a summary, instead of not continuing at all.

csmith49:
> That is, I was wondering if we could go in the opposite direction of the discussion there: use a condensation intently when an error like that, which makes the conversation unusable, happens. We would just continue with a summary, instead of not continuing at all.

That's the intent of this PR, although we still need some specialized logic to surface the exception and get it to the condenser (like we do context window exceptions). The real blocker for this PR is #1721.

csmith49:
That last condenser integration test is a bummer. Checking the test results manually confirms the test in question (`c02`) passed for both models.

csmith49:
@OpenHands please fix the failing actions on PR #1596 at branch `feat/hard-context-reset`

csmith49:
The failing test doesn't appear to be related to the changes here -- it's some cross test that seems a little flaky.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -116,17 +120,49 @@ def condensation_requirement(
...
+
+        output = str(event)
+
+        if len(output) <= max_event_str_length:
+            return output
+
+        # Cut out the middle part of the string representation to ensure the size is
+        # within limits, while preserving the start and end of the event.
+        half_length = (max_event_str_length - len(truncation_marker)) // 2
+        return output[:half_length] + truncation_marker + output[-half_length:]
```
enyst: 🤔 I wonder what does this mean for images? I think we have them, right?

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -232,6 +273,49 @@ def _get_forgotten_events(
...
+                )
+            except Exception as e:
+                # If we haven't set a max_event_str_length yet, set it as the largest
+                # event string length.
+                if max_event_str_length is None:
+                    max_event_str_length = max(len(str(event)) for event in view.events)
+
+                # Since the summarization failed, reduce the max_event_str_length by 20%
+                assert max_event_str_length is not None
+                max_event_str_length = int(max_event_str_length * 0.8)
```
enyst: 🤔 We don't seem to discriminate between event types, e.g. a long user prompt and a long bash output.

The user prompt ... unlikely to be _too_ large perhaps; the tmux output bothers me a bit because it's not clear to me that it has a hard limit. I propose to give it one here:
- https://github.com/OpenHands/software-agent-sdk/pull/1823

I feel there could be (easily?) ways in which the distribution looks like, with one or a few large events and many small, that might not pass 5x?

What if the main LLM is Gemini with 1 million tokens context window, and the summarizer Sonnet with 200k?

tests/cross/test_remote_conversation_live_server.py
```diff
@@ -133,9 +133,11 @@ def server_env(
         thread.join(timeout=2)
 
         # Clean up any leftover directories created during the test
+        # Use ignore_errors=True to handle race conditions where the server may still
+        # be writing files during cleanup
         cwd_conversations = Path("workspace/conversations")
         if cwd_conversations.exists():
-            shutil.rmtree(cwd_conversations)
+            shutil.rmtree(cwd_conversations, ignore_errors=True)
```
enyst: Is this intended?

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
enyst: Actually, what I'd like to note is, we don't seem to truncate the `terminal` outputs at the moment, and others are unclear (MCP? not sure), but on the other hand, they are not sent to the LLM with full size:
https://github.com/OpenHands/software-agent-sdk/blob/ae0dc8a992473caeba315aaa2502bc5fdd54fc90/openhands-sdk/openhands/sdk/llm/message.py#L178

I'm reading this as: because it uses events stringified directly, the condenser doesn't see the same events nor the same sizes, as the agent LLM, it sees events potentially much larger. Does that seem accurate?

enyst:
> The failing test doesn't appear to be related to the changes here -- it's some cross test that seems a little flaky.

You're right, it's slated for fix here:
- https://github.com/OpenHands/software-agent-sdk/pull/1832

tests/cross/test_remote_conversation_live_server.py
csmith49: It's unrelated, crept in when I blindly got OH to resolve some flaky tests. Will remove that commit to keep this PR clean.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
csmith49: It looks like they get subbed in as `[Image: X URLs]` during stringification.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
csmith49: > I'm reading this as: because it uses events stringified directly, the condenser doesn't see the same events nor the same sizes, as the agent LLM, it sees events potentially much larger. Does that seem accurate?

Partially right. It uses the `__str__` function on `LLMConvertibleEvent` to render them, which has some in-built truncation on text components. But it's true that we don't use the normal `Event` -> `Message` pipeline, so the condenser's view is slightly different than what the agent sees.

> I feel there could be (easily?) ways in which the distribution looks like, with one or a few large events and many small, that might not pass 5x?

Probably! I'll make the parameters here configurable so folks can make it shrink more aggressively or try for longer.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
enyst: Sorry I'm dense, do you mean base64 data is replaced with a short indicator that there was an image there?

csmith49: Yes. No image data is ever shown to the condenser, they're just replaced by a string that indicates an image _was_ there.

I'm not 100% on the v0 behavior but this stringification of events has always been the implementation, so it's reasonable to expect something similar was happening.

For reference, this is the relevant function: https://github.com/OpenHands/software-agent-sdk/blob/ae0dc8a992473caeba315aaa2502bc5fdd54fc90/openhands-sdk/openhands/sdk/event/base.py#L65-L88

And tools that take screenshots _should_ end up with those screenshots getting converted through this function: https://github.com/OpenHands/software-agent-sdk/blob/ae0dc8a992473caeba315aaa2502bc5fdd54fc90/openhands-sdk/openhands/sdk/llm/message.py#L683-L694

enyst: Ahh I misunderstood, I think this is great actually, thank you!

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -21,6 +22,9 @@
 from openhands.sdk.observability.laminar import observe
 
 
+logger = getLogger(__name__)
```
xingyaoww: @OpenHands we should use openhands.sdk's logger here.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -116,17 +128,49 @@ def condensation_requirement(
...
+
+        output = str(event)
+
+        if len(output) <= max_event_str_length:
+            return output
+
+        # Cut out the middle part of the string representation to ensure the size is
+        # within limits, while preserving the start and end of the event.
+        half_length = (max_event_str_length - len(truncation_marker)) // 2
+        return output[:half_length] + truncation_marker + output[-half_length:]
```
xingyaoww: We have a `maybe_truncate` function in the SDK utils, shall we re-use them here?

PR #1614, opened 2026-01-06
fix: add file-based locking to EventLog.append()

### LINKED ISSUE #1545: Race condition in EventLog.append() causes duplicate indices when multiple instances write concurrently
## Summary

EventLog.append() is not thread-safe. When multiple EventLog instances write to the same directory concurrently, they can produce duplicate file indices, leading to data corruption and downstream errors.

## Root Cause

The `append()` method reads `self._length`, writes a file, then increments `self._length`. If two EventLog instances (or threads) call `append()` concurrently:
1. Both read the same `_length` value
2. Both write to the same file path (overwriting each other)
3. Index corruption occurs

## Reproduction

This was discovered when investigating an Anthropic API error:
```
messages.93: tool_use ids were found without tool_result blocks immediately after
```

Analysis of conversation events showed:
- Two parallel event streams writing to overlapping indices (112-155)
- Stream 1: Agent writing events during processing
- Stream 2: User message handler creating a new EventLog instance that didn't see recent writes

## Impact

- Duplicate event files with same index- Duplicate event tent
- Event ordering corruption
- LLM message conversion fails due to tool_use/tool_result mismatch
- Conversation becomes unusable

## Proposed Fix

Add file-based locking to `EventLog.append()` to ensure Add fine procesAdd file-based locking to `EventLog.append()` to ensure Add fine procesAdd file-based locking to `EventLog.append()` to ensure Add fine procesAdd file-based lockiReAdd file-based locking to `EventLogands-CLI when users send messages while the agent is still procesAdd file-based locking toll Add file-basedOpenHands-CLI to investigate the root cause there.

ixchio:
Fixes #1545

Hey folks! 👋

So I was digging into some weird duplicate index issues and found the culprit - classic race condition in `EventLog.append()`. When multiple instances hit it at the same time, they both grab the same `_length`, write to the same file path, and things go sideways fast.

## What's changed:

- Pulled in `filelock` for proper cross-process locking
- Wrapped the critical section in `EventLog.append()` with a `FileLock`  
- After grabbing the lock, we re-check disk state in case someone else wrote while we waited
- Added `exists()` and `get_absolute_path()` to `FileStore` interface (needed for lock file handling)

## Tests:

All existing tests still pass, plus added:
- `test_event_log_concurrent_append_thread_safety` - hammers it with 5 threads doing 10 writes each
- `test_event_log_concurrent_writes_serialized` - two EventLog instances sharing the same dir

Lmk if anything looks off or if you'd prefer a different approach!

### DIFF
--- a/openhands-sdk/openhands/sdk/critic/impl/agent_finished.py
+++ b/openhands-sdk/openhands/sdk/critic/impl/agent_finished.py
@@ -2,14 +2,14 @@
 AgentFinishedCritic implementation.
 
 This critic evaluates whether an agent properly finished a task by checking:
-1. The agent's last action was a FinishAction (proper completion)
+1. The agent finished properly (via FinishAction or final MessageEvent)
 2. The generated git patch is non-empty (actual changes were made)
 """
 
 from collections.abc import Sequence
 
 from openhands.sdk.critic.base import CriticBase, CriticResult
-from openhands.sdk.event import ActionEvent, LLMConvertibleEvent
+from openhands.sdk.event import ActionEvent, LLMConvertibleEvent, MessageEvent
 from openhands.sdk.logger import get_logger
 from openhands.sdk.tool.builtins.finish import FinishAction
 
@@ -22,7 +22,10 @@ class AgentFinishedCritic(CriticBase):
     Critic that evaluates whether an agent properly finished a task.
 
     This critic checks two main criteria:
-    1. The agent's last action was a FinishAction (proper completion)
+    1. The agent finished properly - either via:
+       - A FinishAction (explicit completion), or
+       - A final MessageEvent from the agent (implicit completion when LLM
+         produces content and awaits user input)
     2. The generated git patch is non-empty (actual changes were made)
     """
 
@@ -51,10 +54,10 @@ class AgentFinishedCritic(CriticBase):
                 + "; ".join(reasons),
             )
 
-        # Check if agent properly finished with FinishAction
-        if not self._has_finish_action(events):
-            reasons.append("No FinishAction found")
-            logger.debug("AgentFinishedCritic: No FinishAction")
+        # Check if agent properly finished
+        if not self._has_proper_finish(events):
+            reasons.append("No proper finish signal found")
+            logger.debug("AgentFinishedCritic: No proper finish signal")
             return CriticResult(
                 score=0.0,
                 message="Agent did not finish properly. " + "; ".join(reasons),
@@ -63,21 +66,43 @@ class AgentFinishedCritic(CriticBase):
         logger.debug("AgentFinishedCritic: Successfully completed")
         return CriticResult(
             score=1.0,
-            message="Agent completed with FinishAction and non-empty patch",
+            message="Agent completed properly with non-empty patch",
         )
 
-    def _has_finish_action(self, events: Sequence[LLMConvertibleEvent]) -> bool:
-        """Check if the last action was a FinishAction."""
+    def _has_proper_finish(self, events: Sequence[LLMConvertibleEvent]) -> bool:
+        """Check if the agent finished properly.
+
+        The agent can finish in two ways:
+        1. By calling FinishAction explicitly
+        2. By sending a final MessageEvent (when LLM produces content and
+           the conversation awaits user input - this sets execution_status
+           to FINISHED in the SDK)
+
+        We check for either signal by looking at the last meaningful event
+        (skipping ConversationStateUpdateEvent which is just metadata).
+        """
         if not events:
             return False
 
-        # Look for the last ActionEvent in the history
+        # Look for the last meaningful event in the history
         for event in reversed(events):
+            # Check for FinishAction
             if isinstance(event, ActionEvent):
-                # Check if this is a FinishAction
                 if event.action and isinstance(event.action, FinishAction):
+                    logger.debug("AgentFinishedCritic: Found FinishAction")
                     return True
-                # If we find any other action type, the agent didn't finish
+                # If we find any other action type as the last event,
+                # the agent didn't finish properly
+                action_name = type(event.action).__name__
+                logger.debug(
+                    f"AgentFinishedCritic: Last action was {action_name}, "
+                    "not FinishAction"
+                )
                 return False
+
+            # Check for final MessageEvent from agent (implicit finish)
+            if isinstance(event, MessageEvent) and event.source == "agent":
+                logger.debug("AgentFinishedCritic: Found final MessageEvent from agent")
+                return True
 
         return False
--- a/tests/sdk/critic/test_critic.py
+++ b/tests/sdk/critic/test_critic.py
@@ -11,8 +11,8 @@ from openhands.sdk.critic import (
     EmptyPatchCritic,
     PassCritic,
 )
-from openhands.sdk.event import ActionEvent
-from openhands.sdk.llm import MessageToolCall, TextContent
+from openhands.sdk.event import ActionEvent, MessageEvent
+from openhands.sdk.llm import Message, MessageToolCall, TextContent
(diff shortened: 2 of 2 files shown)


openhands-sdk/pyproject.toml
```diff
@@ -7,14 +7,15 @@ requires-python = ">=3.12"
 dependencies = [
     "deprecation>=2.1.0",
     "fastmcp>=2.11.3",
+    "filelock>=3.15.0",
```
jpshackelford: Vulnerable to CVE-2025-68146: TOCTOU symlink attack vulnerability. Version 3.20.1 is the first patched version. Anyone running on 3.15.x-3.20.0 can have arbitrary files truncated by a local attacker. Pin to >=3.20.1

jpshackelford:
Currently, there's no inter-process locking in the production SDK code. This PR introduces the first such mechanism. 

The codebase has a custom FIFOLock implementation that provides:

* FIFO ordering (fairness)
* Reentrancy
* Timeout support

This is used in ConversationState for thread coordination but it does not provide support for inter-process coordination which is required here.

In addition to not working reliably on NFS, flock() has some limitations we need to think about:

| Limitation | Relevant? | Analysis |
|------------|-----------|----------|
| **No Asyncio Support** | ⚠️ **Potentially** | `LocalConversation` calls `self._state.events.append(e)` synchronously from callbacks. However, `RemoteConversation` uses asyncio. If the codebase ever needs async event persistence, this becomes a blocker. Currently OK since `EventLog` is only used in `LocalConversation`. |
| **Thread-Local Behavior** | ✅ **Yes, but OK** | Each `EventLog` instance creates its own `FileLock`. The PR's design means the lock is per-EventLog instance, not shared across threads. This is actually the correct behavior - multiple threads sharing an `EventLog` will correctly serialize through the single lock instance. |
| **Lock Files Not Auto-Deleted** | ⚠️ **Minor concern** | `.eventlog.lock` files will persist in every conversation's `events/` directory. For long-running systems with many conversations, these accumulate. Not a functional problem, but adds clutter. Could add cleanup in a future PR. |
| **Android/Termux Unsupported** | ✅ **Not relevant** | Probably not an issue though I know of one community member doing something with OpenHands on Android... |
| **No Deadlock Detection** | ⚠️ **Potentially** | The code only uses ONE lock (`_lock` in EventLog), so circular deadlocks are impossible within EventLog itself. However, if `EventLog.append()` is called while holding `ConversationState._lock` (the `FIFOLock`), and something inside `append()` tried to acquire `ConversationState._lock`, you'd deadlock. Looking at the code, this doesn't happen currently - `append()` only touches the file store. **Low risk but worth noting.** |

#### The Key Question: Is `EventLog.append()` Called from Async Code?

Looking at the call chain:
```
LocalConversation._default_callback(e)
  → self._state.events.append(e)    # This is EventLog.append()
```

This callback is invoked synchronously during agent execution. The `RemoteConversation` has a different path that doesn't use local `EventLog` at all (it streams to a server).

**Verdict: Asyncio limitation is NOT currently blocking**, but the architecture should be aware that `EventLog` is sync-only.

jpshackelford:
If we go with this locking library / mechanism a few things we'd probably want to add:

- [ ]  Consider a timeout on lock acquisition and raising an error rather than silent hang for cases where lock holder itself is stuck. 

- [ ]  Fix `InMemoryFileStore.get_absolute_path()` to include a unique instance identifier. Currently all `InMemoryFileStore` instances return the same lock file path (`/tmp/openhands_inmemory/...`), meaning separate in-memory stores across different tests or processes will contend on the same lock. Adding a UUID or similar per-instance makes each store's lock independent.

- [ ]  Log exceptions in `_count_events_on_disk()` and`_sync_from_disk()` instead of silently swallowing them. Currently these methods catch all exceptions and return quietly, making debugging difficult. At minimum, `FileNotFoundError` (expected when directory doesn't exist yet) should be handled separately from unexpected errors, which should be logged with context.

- [ ]  Document the NFS/network filesystem limitation in the class docstring. File locking via flock() does not work reliably on NFS mounts or network filesystems. Users deploying with shared storage should be aware

ixchio:
@jpshackelford  Hey! Just pushed updates addressing all your feedback. Added lock timeout w/ proper error handling, split out FileNotFoundError from the catch-all, added logging, and pinned filelock to 3.20.1 for the CVE. 
tests still passing 👍

openhands-sdk/openhands/sdk/io/memory.py
```diff
@@ -49,6 +52,20 @@ def delete(self, path: str) -> None:
             keys_to_delete = [key for key in self.files.keys() if key.startswith(path)]
             for key in keys_to_delete:
                 del self.files[key]
-            logger.debug(f"Cleared in-memory file store: {path}")
+            logger.debug("Cleared in-memory file store: %s", path)
```
tofarr: Nit: Even though logging is more efficient with formatted strings, we settled on interpolated strings due to the simpler syntax.

openhands-sdk/openhands/sdk/io/memory.py
```diff
@@ -9,9 +10,11 @@
 
 class InMemoryFileStore(FileStore):
     files: dict[str, str]
+    _instance_id: str
 
     def __init__(self, files: dict[str, str] | None = None) -> None:
         self.files = {}
+        self._instance_id = str(uuid.uuid4())
```
tofarr: Nit: uuid.uuid4().hex would yield a string without dashes - we use this in most places as they are shorter and can be double clicked on to select in many environments.

openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -75,26 +94,83 @@ def __iter__(self) -> Iterator[Event]:
...
+        except Exception as e:
+            logger.warning("Error listing event directory %s: %s", self._dir, e)
+            return 0
+        return sum(
+            1
+            for p in paths
+            if p.rsplit("/", 1)[-1].startswith("event-") and p.endswith(".json")
+        )
+
+    def _sync_from_disk(self, disk_length: int) -> None:
```
tofarr, line 150: This feels substantially similar to the _scan_and_build_index method which already exists. Could we reuse / adapt that?

ixchio:
Done 👍

ixchio:
Refactored to reuse _scan_and_build_index - see latest commit 👍

openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -16,17 +18,35 @@
...
+    Note:
+        File locking via flock() does NOT work reliably on NFS mounts or
+        network filesystems. Users deploying with shared storage should
+        use alternative coordination mechanisms.
+    """
+
     _fs: FileStore
     _dir: str
     _length: int
+    _lock: BaseFileLock
```
tofarr: I wonder would it make sense to add the locking mechanism to the FileStore interface? The event log as it stands tries to avoid being bound to the file system - and I wonder if having the BaseFileLock here would clash with implementations like the InMemoryFileStore which don't actually have an attached file system?

ixchio: Solid point - yeah shoving FileLock into EventLog feels kinda janky when InMemoryFileStore doesn't need flock at all. I'll move it behind __enter__/__exit__ on the FileStore interface....cheers!

ixchio:
Refactored locking into FileStore interface with `__enter__`/`__exit__`. EventLog now uses `with self._fs.lock(...)`. Added threading lock for InMemoryFileStore to keep tests happy. Ready for another look! 🚀

ixchio:
Glad we got this one across the finish line! 🍰 I think I’ve officially caught the OpenHands bug—that’s 5 PRs down and I’m just getting warmed up. Great working with you all on the technical deep-dives; let’s keep this momentum going! 💪

PR #1619, opened 2026-01-07
fix: several issues related to scalability

Co1lin:
## Summary

This PR tries to fix several issues relevant to scalability. It is motivated from my large-scale agentic data synthesis jobs.

- In RemoteWorkspace, the httpx client is responsible for making all network requests, including querying LLMs. However, the current 60 second read timeout is not enough for the following settings:
  - Queries with long context and reasoning: I used `vertex_ai/gemini-3-pro-preview` and `vertex_ai/gemini-2.5-pro` in my jobs, and it is quite common for them to take more than 60 seconds to respond, when the context is long and the reasoning effort is high.
  - Querying self-host LLMs: self-hosted models with vLLM and SGLang can often have a latency longer than 60 seconds, due to the speed of the used GPUs and the batch size of the queries, such as large batch sizes in data synthesis settings.
- Consider the example code below showing a toy example of large-scale agentic data synthesis. In this example, I want the agent to do read-only explorations in the repo. To reduce overhead, it is straightforward to use a single workspace for all tasks, instead of launching an independent workspace for each sample synthesis. With such a setting, I encountered several issues and therefore propose the following fixes:
  - I propose to remove [the default limit of 100](https://github.com/encode/httpx/blob/ae1b9f66238f75ced3ced5e4485408435de10768/httpx/_config.py#L247) on the max connections. In my tasks, it is quite common to use > 100 workers in parallel to query LLMs to speed up large-scale data synthesis.
  - When using one DockerWorkspace for many conversations, there will be many tmux sessions, each for one conversation. This can trigger the "too many open files" errors. Though there could be other ways to trigger this error. So I propose to increase the limit when starting the container.
  - If too many conversations reuse one container, currently the tmux sessions are not closed when their conversations close, which will accumulate many tmux sessions, resulting in the whole system getting pretty slow and stuck. (I tried to enter into the container and I cannot even successfully run `tmux ls` or attach to any session inside.) To solve this issue, we can send the DELETE request from the client to trigger server-side [`delete_conversation`](https://github.com/OpenHands/software-agent-sdk/blob/2c5a439635dbdcecfc0b58c283995b1dc4552fae/openhands-agent-server/openhands/agent_server/conversation_router.py#L159), which I think will eventually trigger [`TmuxTerminal.close`](https://github.com/OpenHands/software-agent-sdk/blob/2c5a439635dbdcecfc0b58c283995b1dc4552fae/openhands-tools/openhands/tools/terminal/terminal/tmux_terminal.py#L93) after a long chain. I verified on my ~10k scale job that this can effectively prevent the tmux session accumulation over time (I monitored the number of sessions in the container) and keep the system smooth.

<details>
<summary>too many open files error I got</summary>
```python
[DOCKER] {"asctime": "2025-12-11 12:11:49,946", "levelname": "ERROR", "name": "openhands.agent_server.api", "filename": "api.py", "lineno": 223, "message": "Unhandled exception on POST /api/conversations", "exc_info": "Traceback (most recent call last):\n  File \"starlette/middleware/errors.py\", line 164, in __call__\n  File \"starlette/middleware/cors.py\", line 85, in __call__\n  File \"starlette/middleware/exceptions.py\", line 63, in __call__\n  File \"starlette/_exception_handler.py\", line 53, in wrapped_app\n  File \"starlette/_exception_handler.py\", line 42, in wrapped_app\n  File \"fastapi/middleware/asyncexitstack.py\", line 18, in __call__\n  File \"starlette/routing.py\", line 716, in __call__\n  File \"starlette/routing.py\", line 736, in app\n  File \"starlette/routing.py\", line 290, in handle\n  File \"fastapi/routing.py\", line 123, in app\n  File \"starlette/_exception_handler.py\", line 53, in wrapped_app\n  File \"starlette/_exception_handler.py\", line 42, in wrapped_app\n  File \"fastapi/routing.py\", line 109, in app\n  File \"fastapi/routing.py\", line 389, in app\n  File \"fastapi/routing.py\", line 288, in run_endpoint_function\n  File \"openhands/agent_server/conversation_router.py\", line 137, in start_conversation\n  File \"openhands/agent_server/conversation_service.py\", line 196, in start_conversation\n  File \"openhands/agent_server/conversation_service.py\", line 418, in _start_event_service\n  File \"openhands/agent_server/event_service.py\", line 355, in start\n  File \"openhands/sdk/conversation/impl/local_conversation.py\", line 164, in __init__\n  File \"openhands/sdk/agent/agent.py\", line 100, in init_state\n  File \"openhands/sdk/agent/base.py\", line 189, in init_state\n  File \"openhands/sdk/agent/base.py\", line 200, in _initialize\n  File \"openhands/sdk/tool/registry.py\", line 156, in resolve_tool\n  File \"openhands/sdk/tool/registry.py\", line 103, in _resolve\n  File \"openhands/tools/terminal/definition.py\", line 272, in create\n  File \"openhands/tools/terminal/impl.py\", line 57, in __init__\n  File \"openhands/tools/terminal/terminal/terminal_session.py\", line 83, in initialize\n  File \"openhands/tools/terminal/terminal/tmux_terminal.py\", line 57, in initialize\n  File \"libtmux/server.py\", line 574, in new_session\nlibtmux.exc.LibTmuxException: ['create window failed: fork failed: Too many open files']"}
```
</details>

The toy example showing the concept of a large-scale data synthesis job:

```python
import asyncio
from pathlib import Path

from tqdm.asyncio import tqdm as async_tqdm

from openhands.sdk import (
    LLM,
    Conversation,
    RemoteConversation,
)
from openhands.tools.preset.default import get_default_agent
from openhands.workspace import DockerWorkspace


llm = LLM(
    model='openai/gpt-5.1-codex-mini',
)

agent = get_default_agent(llm=llm, cli_mode=True)


async def work(workspace: DockerWorkspace, code_path: str, semaphone: asyncio.Semaphore) -> RemoteConversation:
    async with semaphone:
        loop = asyncio.get_event_loop()

        conversation = Conversation(agent=agent, workspace=workspace)

        def run():
            conversation.send_message(
                f"Read the code in {code_path} and summarize it. You should respond with a report in markdown format in your last message. Your report should only focus on the code in {code_path} without information about other files. Your report should be informative, detailed, and easy to understand. Do NOT change or write any files into the environment."
            )
            conversation.run()

        await loop.run_in_executor(None, run)  # run()
        return conversation


if __name__ == "__main__":
    host_dir = Path(__file__).parent / "z_code_test"
    src_dir = host_dir / "src"
    src_dir.mkdir(parents=True, exist_ok=True)
    (src_dir / "code_0.py").write_text("print('Hello, world!')\n")
    (src_dir / "code_1.py").write_text("print('Goodbye, world!')\n")
    (src_dir / "code_2.py").write_text("print('Another code file')\n")
    # there could be more code files for analysis...

    num_samples = 100_000
    num_workers = 128
    semaphone = asyncio.Semaphore(num_workers)

    with DockerWorkspace(
        server_image="ghcr.io/openhands/agent-server:latest-python",
        host_port=8011,
        mount_dir=str(host_dir.resolve()),
        working_dir='/workspace/src',
    ) as workspace:
        import httpx
        workspace._client = httpx.Client(
            base_url=workspace.host,
            headers=workspace._headers,
            timeout=httpx.Timeout(connect=10.0, read=600.0, write=10.0, pool=10.0),
            limits=httpx.Limits(max_connections=None),
        )

        code_paths = [f"code_{i % 3}.py" for i in range(num_samples)]

        rets = asyncio.run(
            async_tqdm.gather(
                *[work(workspace, code_path, semaphone) for code_path in code_paths],
                desc="working",
            )
        )
```

## Checklist


openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -984,6 +984,14 @@ def close(self) -> None:
         if self._hook_processor is not None:
             self._hook_processor.run_session_end()
         try:
+            # trigger server-side delete_conversation to release resources
+            # like tmux sessions
+            _send_request(
+                self._client,
+                "DELETE",
+                f"/api/conversations/{self.id}",
```
enyst: I can see the need for this in the use case you described, but I'm not sure we risk here unintended consequences on other use cases, such as: stop a conversation, and restart it later 🤔 

I wonder, since you have a reference to the conversation in your script, is it possible to delete it?

Co1lin: Thanks for your comment!
Do you mean deleting the reference to a conversation, i.e., `del conversation`, can release the tmux session of it?
I think in my script, the conversation object is only available in each `work` coroutine, so it will be released by Python GC after each coroutine finishes. However, my observation is that this does not kill the tmux session of it, leading to accumulation.
I think the `close` method is designed to "close the conversation and clean up resources", and I can only see `__del__` calls it. I think the current codebase also does not support closing a conversation and restarting it later because it already releases other resources? As users, if a conversation is needed later, then I guess we just do not call `close` and leave it.
Welcome your thoughts!

openhands-sdk/openhands/sdk/workspace/remote/base.py
```diff
@@ -49,12 +49,15 @@ def client(self) -> httpx.Client:
         if client is None:
             # Configure reasonable timeouts for HTTP requests
             # - connect: 10 seconds to establish connection
-            # - read: 60 seconds to read response (for LLM operations)
+            # - read: 600 seconds to read response (for LLM operations)
             # - write: 10 seconds to send request
             # - pool: 10 seconds to get connection from pool
-            timeout = httpx.Timeout(connect=10.0, read=60.0, write=10.0, pool=10.0)
+            timeout = httpx.Timeout(connect=10.0, read=600.0, write=10.0, pool=10.0)
```
xingyaoww: What do you feel if we make the read timeout configurable? e.g., making it a properly of the class and increase the default to 600?

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
xingyaoww: Sometimes when we close a conversation, we might still want the agent to keep running.

What do you think about adding a `delete_on_close` field and default to "False" so it keeps the current bahavior.

When you want to clean up the convo on exit, you can just set it to true?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -535,20 +538,23 @@ def close(self) -> None:
...
-        for tool in tools_map.values():
+        if self.stop_agent_on_close:
             try:
-                executable_tool = tool.as_executable()
-                executable_tool.executor.close()
-            except NotImplementedError:
-                # Tool has no executor, skip it without erroring
-                continue
-            except Exception as e:
-                logger.warning(f"Error closing executor for tool '{tool.name}': {e}")
```
Co1lin, line 724: Hi @xingyaoww , I checked the current implementation in `LocalConversation`, and I think it already stops the agent running when the conversation is closed.
Therefore, in the last commit of this PR, I added a new attribute `stop_agent_on_close` in both `LocalConversation` and `RemoteConversation` to decide whether to keep agent running after closing the conversation.

Co1lin:
Thanks for both of you following up on this! :) Let me know what do you think on the current version.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
xingyaoww: I don't actually quite understand the need to have `stop_agent_on_close`, since it only close the tools 🤔 

And putting it inside `close()` function might not be the best idea, when you call `agent.close()` you kinda expect the agent to be closed

openhands-sdk/openhands/sdk/workspace/remote/base.py
```diff
@@ -50,12 +50,17 @@ def client(self) -> httpx.Client:
...
-            timeout = httpx.Timeout(connect=10.0, read=60.0, write=10.0, pool=10.0)
+            timeout = httpx.Timeout(
+                connect=10.0, read=self.read_timeout, write=10.0, pool=10.0
+            )
             client = httpx.Client(
-                base_url=self.host, timeout=timeout, headers=self._headers
+                base_url=self.host,
+                timeout=timeout,
+                headers=self._headers,
+                limits=httpx.Limits(max_connections=None),
```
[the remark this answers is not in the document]
xingyaoww: How about we keep the limits to be default?

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -439,6 +439,7 @@ class RemoteConversation(BaseConversation):
     _client: httpx.Client
     _hook_processor: HookEventProcessor | None
     _cleanup_initiated: bool
+    stop_agent_on_close: bool = False
```
xingyaoww: Shall we consider rename it to smth like: `delete_on_close`?

Co1lin: > How about we keep the limits to be default?
This is for supporting running > 100 conversations in parallel. If we keep it to be default, one needs to hack the code to make it work. I can also turn it into an argument. But you can decide if we want this or not.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
Co1lin: I think here we may have some misunderstanding to each other regarding closing conversation, closing agent, and cleaning up tools. Sorry for the delay but I will come back after some rethinking.

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
xingyaoww: > This is for supporting running > 100 conversations in parallel. If we keep it to be default, one needs to hack the code to make it work. I can also turn it into an argument. But you can decide if we want this or not.

Yeah, let's make `max_connections` an argument and default to None!

@OpenHands please address all of these discussed item and push it up

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -58,6 +58,7 @@ class LocalConversation(BaseConversation):
     llm_registry: LLMRegistry
     _cleanup_initiated: bool
     _hook_processor: HookEventProcessor | None
+    delete_on_close: bool = False
```
xingyaoww: @OpenHands let's modify the default value for this to keep the old behavior here: which should be we by default delete on close

xingyaoww (approved):
LGTM, thank you! Sorry for taking so long to approve this

xingyaoww:
@OpenHands pls merge from main and resolve all the conflictss

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -439,6 +439,7 @@ class RemoteConversation(BaseConversation):
     _client: httpx.Client
     _hook_processor: HookEventProcessor | None
     _cleanup_initiated: bool
+    delete_on_close: bool = False
```
xingyaoww, line 558: @OpenHands please send a PR to OpenHands/benchmarks repo and set `delete_on_close` for ALL usage of `RemoteConversation` there

Issue #1473, opened 2025-12-22
Support AgentSkills standard


neubig:
## Overview

Add support for the [AgentSkills standard](https://agentskills.io/) - an open format for giving agents new capabilities. This will enable skills created for the OpenHands SDK to be compatible with other platforms that adopt the standard (e.g., Claude Code), and vice versa.

## Background

The AgentSkills standard defines:
- **SKILL.md format**: YAML frontmatter with required `name` and `description` fields, plus optional `license`, `compatibility`, `metadata`, and `allowed-tools` fields
- **Directory structure**: Skills are directories containing `SKILL.md`, with optional `scripts/`, `references/`, and `assets/` subdirectories
- **Progressive disclosure**: Metadata loaded at startup, full instructions on activation, resources on demand
- **Validation**: Name rules (64 chars, lowercase, hyphens), description limits (1024 chars), etc.

The reference implementation is available at [agentskills/agentskills](https://github.com/agentskills/agentskills) with a `skills-ref` Python library.

## Current State

The OpenHands SDK already has a skills system (`openhands.sdk.context.skills`) with:
- `Skill` model with `name`, `content`, `trigger`, `source`, `mcp_tools`, `inputs`
- Loading from `.md` files with frontmatter
- Support for keyword and task triggers
- MCP tools integration

## Sub-Issues

This work is broken into 5 implementation tasks:

- [ ] #1474 - Add skills-ref dependency and extend Skill model with AgentSkills fields
- [ ] #1475 - Support SKILL.md file convention and name validation
- [ ] #1476 - Support .mcp.json for MCP server configuration
- [ ] #1477 - Support resource directories (scripts/, references/, assets/)
- [ ] #1478 - Add validation and prompt generation utilities

## Compatibility Notes

OpenHands-specific extensions (`trigger`, `mcp_tools`, `inputs`) will be preserved:
- These can be stored in the `metadata` field for AgentSkills compatibility
- Skills using OpenHands extensions will still work but may not be portable to other platforms
- `.mcp.json` support provides Claude Code compatibility for MCP configuration

## References

- [AgentSkills Specification](https://agentskills.io/specification)
- [skills-ref library](https://github.com/agentskills/agentskills/tree/main/skills-ref)
- [Anthropic Skills Repository](https://github.com/anthropics/skills)
- [Claude Code Skills Documentation](https://code.claude.com/docs/en/skills)
- https://github.com/OpenHands/software-agent-sdk/issues/1452
[...]

PR #1647, opened 2026-01-08
feat(plugin): Add Plugin.fetch() for remote plugin fetching and caching

### LINKED ISSUE #1645: feat(plugin): Add Plugin.fetch() for remote plugin fetching and caching
## Summary

Add the ability to fetch plugins from remote sources (GitHub repositories, git URLs) and cache them locally. This extends the existing `Plugin.load()` which only works with local paths.

This is a follow-up to https://github.com/OpenHands/software-agent-sdk/issues/1473. 

## Context

This is a building block for the Plugin Directory feature ([OpenHands/OpenHands#12088](https://github.com/OpenHands/OpenHands/issues/12088)). When a user launches a plugin from the Plugin Directory, the backend needs to fetch the plugin from its source before loading it.

The SDK already has a similar pattern for public skills in `update_skills_repository()` which clones/updates a git repo to a local cache. This issue extends that pattern to plugins.

See https://github.com/OpenHands/OpenHands/issues/12085#issuecomment-3725059694 for how this fits together with other pieces.

### New Function: `Plugin.fetch()`

```python
@classmethod
def fetch(
    cls,
    source: str,
    cache_dir: Path | None = None,
    ref: str | None = None,
) -> Path:
    """Fetch a plugin from a remote source and return the local cached path.
    
    Args:
        source: Plugin source - can be:
            - "github:owner/repo" - GitHub repository shorthand
            - "https://github.com/owner/repo.git" - Full git URL
            - "/local/path" - Local path (returned as-is)
        cache_dir: Directory for caching. Defaults to ~/.openhands/cache/plugins/
        ref: Optional branch, tag, or commit to checkout
        
    Returns:
        Path to the local plugin directory (ready for Plugin.load())
        
    Raises:
        PluginFetchError: If fetching fails
    """
```

### Caching Strategy

- Cache location: `~/.openhands/cache/plugins/{source_hash}/`
- If cached and exists: optionally update (git fetch) or use as-is
- If not cached: clone the repository
- Support shallow clones for efficiency (`--depth 1`)

### Source Parsing

```python
def parse_plugin_source(source: str) -> tuple[str, str]:
    """Parse plugin source into (type, url).
    
    Examples:
        "github:owner/repo" → ("github", "https://github.com/owner/repo.git")
        "https://gitlab.com/org/repo.git" → ("git", "https://gitlab.com/org/repo.git")
        "/local/path" → ("local", "/local/path")
    """
```

## Files to Modify

- `openhands-sdk/openhands/sdk/plugin/plugin.py` - Add `fetch()` classmethod
- `openhands-sdk/openhands/sdk/plugin/__init__.py` - Export new function
- New: `openhands-sdk/openhands/sdk/plugin/fetch.py` - Fetching logic (optional, could be in plugin.py)

## Acceptance Criteria

- [ ] `Plugin.fetch("github:owner/repo")` clones and returns local path
- [ ] `Plugin.fetch("https://...")` works with any git URL
- [ ] `Plugin.fetch("/local/path")` returns path unchanged
- [ ] Caching works - second fetch uses cached version
- [ ] `ref` parameter checks out specific branch/tag
- [ ] Errors are handled gracefully with clear messages
- [ ] Unit tests cover all source types and caching behavior

## Related Issues

- OpenHands/OpenHands#12085 - [PRD] OpenHands Plugins
- OpenHands/OpenHands#12088 - Plugin Marketplace GUI
- #1453 - Support for bare git repository plugin marketplaces

jpshackelford:
## Summary

Implements #1645 - adds the ability to fetch plugins from remote sources (GitHub repositories, git URLs) and cache them locally. This is a building block for the Plugin Directory feature ([OpenHands/OpenHands#12088](https://github.com/OpenHands/OpenHands/issues/12088)).

### New Function: `Plugin.fetch()`

```python
@classmethod
def fetch(
    cls,
    source: str,
    cache_dir: Path | None = None,
    ref: str | None = None,
    update: bool = True,
    subpath: str | None = None,
) -> Path:
    """Fetch a plugin from a remote source and return the local cached path."""
```

### Source Parsing

The `parse_plugin_source()` function handles various source formats:
- `"github:owner/repo"` - GitHub repository shorthand
- `"https://github.com/owner/repo.git"` - Full git URL  
- `"git@github.com:owner/repo.git"` - SSH git URL
- `"/local/path"` - Local path (returned as-is)

### Caching Strategy

- Cache location: `~/.openhands/cache/plugins/{name}-{hash}/`
- If cached and exists: optionally update (git fetch) or use as-is
- If not cached: clone the repository
- Supports shallow clones for efficiency (`--depth 1`)

### Usage Example

```python
# Fetch and load a plugin
path = Plugin.fetch("github:owner/my-plugin")
plugin = Plugin.load(path)

# With specific version
path = Plugin.fetch("github:owner/my-plugin", ref="v1.0.0")
plugin = Plugin.load(path)

# Fetch a plugin from a subdirectory (monorepo support)
path = Plugin.fetch("github:owner/monorepo", subpath="plugins/my-plugin")
plugin = Plugin.load(path)

# Fetch and load in one step
plugin = Plugin.load(Plugin.fetch("github:owner/my-plugin"))
```

## Files Changed

- `openhands-sdk/openhands/sdk/plugin/fetch.py` - New module with fetching logic
- `openhands-sdk/openhands/sdk/plugin/plugin.py` - Added `fetch()` classmethod
- `openhands-sdk/openhands/sdk/plugin/__init__.py` - Export new types
- `openhands-sdk/openhands/sdk/git/cached_repo.py` - Shared git caching infrastructure (`GitHelper`, `cached_clone_or_update`)
- `openhands-sdk/openhands/sdk/git/utils.py` - Added timeout parameter to `run_git_command`
- `openhands-sdk/openhands/sdk/context/skills/utils.py` - Refactored to use shared git caching
- `tests/sdk/plugin/test_plugin_fetch.py` - Unit tests for fetch functionality
- `tests/sdk/plugin/test_plugin_fetch_integration.py` - Integration tests

## Checklist

- [x] `Plugin.fetch("github:owner/repo")` clones and returns local path
- [x] `Plugin.fetch("https://...")` works with any git URL
- [x] `Plugin.fetch("/local/path")` returns path unchanged
- [x] Caching works - second fetch uses cached version
- [x] `ref` parameter checks out specific branch/tag
- [x] `subpath` parameter supports fetching from subdirectories (monorepo support)
- [x] Errors are handled gracefully with `PluginFetchError`
- [x] Shared git caching infrastructure extracted for reuse by skills system
- [x] Unit tests cover all source types and caching behavior

## Related Issues

- Closes #1645
- Related to OpenHands/OpenHands#12085 - [PRD] OpenHands Plugins
- Related to OpenHands/OpenHands#12088 - Plugin Marketplace GUI
- Follow-up to #1473 (AgentSkills standard support - now closed)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:29b4dcc-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-29b4dcc-python \
  ghcr.io/openhands/agent-server:29b4dcc-python
```

```
ghcr.io/openhands/agent-server:29b4dcc-golang-amd64
ghcr.io/openhands/agent-server:29b4dcc-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:29b4dcc-golang-arm64
ghcr.io/openhands/agent-server:29b4dcc-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:29b4dcc-java-amd64
ghcr.io/openhands/agent-server:29b4dcc-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:29b4dcc-java-arm64
ghcr.io/openhands/agent-server:29b4dcc-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:29b4dcc-python-amd64
ghcr.io/openhands/agent-server:29b4dcc-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:29b4dcc-python-arm64
ghcr.io/openhands/agent-server:29b4dcc-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:29b4dcc-golang
ghcr.io/openhands/agent-server:29b4dcc-java
ghcr.io/openhands/agent-server:29b4dcc-python
```

- Each variant tag (e.g., `29b4dcc-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `29b4dcc-python-amd64`) are also available if needed

### DIFF
--- a/tests/sdk/context/skill/test_load_public_skills.py
+++ b/tests/sdk/context/skill/test_load_public_skills.py
@@ -209,6 +209,8 @@ def test_update_skills_repository_update_existing(tmp_path):
 
     mock_result = MagicMock()
     mock_result.returncode = 0
+    # Simulate being on a branch (not detached HEAD) so reset is called
+    mock_result.stdout = "main"
 
     with patch(
         "openhands.sdk.git.utils.subprocess.run", return_value=mock_result
@@ -220,14 +222,13 @@ def test_update_skills_repository_update_existing(tmp_path):
         )
 
         assert result_path == repo_path
-        # The shared git helper does: fetch, then _checkout_ref (which does
-        # fetch ref, checkout, reset)
+        # The git operations are: fetch, checkout, get_current_branch, reset
+        # (get_current_branch returns branch name so reset is called)
         assert mock_run.call_count == 4
         all_commands = [call[0][0] for call in mock_run.call_args_list]
-        # Should have fetch, fetch ref, checkout, reset
         assert all_commands[0][:3] == ["git", "fetch", "origin"]
-        assert all_commands[1][:3] == ["git", "fetch", "origin"]
-        assert all_commands[2][:2] == ["git", "checkout"]
+        assert all_commands[1][:2] == ["git", "checkout"]
+        assert all_commands[2] == ["git", "rev-parse", "--abbrev-ref", "HEAD"]
         assert all_commands[3][:3] == ["git", "reset", "--hard"]
 
 


jpshackelford:
## Requested Changes: Add `subpath` parameter

### Context
During testing, we identified a bug where plugin paths (e.g., `github:owner/repo/plugins/sub-plugin`) were being rejected by `parse_plugin_source()` because it validates that GitHub shorthand has only one `/` in the repo path. Rather than embedding the path in the source string, we've decided to keep `path` as a separate parameter throughout the entire stack for consistency and clarity.

### Changes Needed

**`openhands-sdk/openhands/sdk/plugin/fetch.py`** - Add `subpath` parameter to `fetch_plugin()`:

```python
def fetch_plugin(
    source: str,
    cache_dir: Path | None = None,
    ref: str | None = None,
    update: bool = True,
    subpath: str | None = None,  # NEW PARAMETER
    git_helper: GitHelper | None = None,
) -> Path:
    """Fetch a plugin from a remote source and return the local cached path.

    Args:
        source: Plugin source - "github:owner/repo", git URL, or local path
        cache_dir: Directory for caching
        ref: Optional branch, tag, or commit
        update: If True, update cached repo
        subpath: Optional subdirectory path within the repo
        git_helper: GitHelper instance

    Returns:
        Path to the plugin directory (with subpath applied if specified)
    """
    source_type, url = parse_plugin_source(source)
    
    # ... existing clone/update logic ...
    
    # NEW: Apply subpath if specified (add before final return)
    if subpath:
        final_path = plugin_path / subpath.strip('/')
        if not final_path.exists():
            raise PluginFetchError(
                f"Subdirectory '{subpath}' not found in plugin repository"
            )
        return final_path
    
    return plugin_path
```

**`openhands-sdk/openhands/sdk/plugin/plugin.py`** - Add `subpath` parameter to `Plugin.fetch()`:

```python
@classmethod
def fetch(
    cls,
    source: str,
    cache_dir: Path | None = None,
    ref: str | None = None,
    update: bool = True,
    subpath: str | None = None,  # NEW PARAMETER
) -> Path:
    """Fetch a plugin from a remote source."""
    return fetch_plugin(
        source, 
        cache_dir=cache_dir, 
        ref=ref, 
        update=update,
        subpath=subpath,  # NEW
    )
```

This maintains API shape consistency across all layers (see https://github.com/OpenHands/OpenHands/issues/12087#issuecomment-3733464400 for the full API design).

openhands-sdk/openhands/sdk/plugin/git_helper.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class GitError(Exception):
+    """Raised when a git operation fails."""
+
+    pass
+
+
+class GitHelper:
+    """Abstraction for git operations, enabling easy mocking in tests."""
```
enyst: I think we already have git clone in the codebase?

openhands-sdk/openhands/sdk/git/cached_repo.py
```diff
@@ -0,0 +1,196 @@
...
+                timeout=timeout,
+            )
+        except subprocess.CalledProcessError as e:
+            stderr = e.stderr.decode() if e.stderr else str(e)
+            raise GitError(f"Fetch failed: {stderr}") from e
+        except subprocess.TimeoutExpired as e:
+            raise GitError(f"Fetch timed out after {timeout}s") from e
+
+    def checkout(self, repo_path: Path, ref: str, timeout: int = 30) -> None:
+        """Checkout a ref (branch, tag, or commit).
```
enyst, line 94: I think we already needed to do `git checkout`, so I wonder if this code duplicates it. Maybe in the `workspace` package?

I believe we checkout:
- user's repo (I think this is a `workspace` thing)
- organization skills repo, e.g. github/acme/.openhands/skills or so

This kind of feature feels like the second, like checking out a repo for this user/org from some remote?

openhands-sdk/openhands/sdk/git/cached_repo.py
```diff
@@ -0,0 +1,196 @@
...
+                capture_output=True,
+                timeout=timeout,
+            )
+        except subprocess.CalledProcessError as e:
+            stderr = e.stderr.decode() if e.stderr else str(e)
+            raise GitError(f"Reset failed: {stderr}") from e
+        except subprocess.TimeoutExpired as e:
+            raise GitError(f"Reset timed out after {timeout}s") from e
+
+    def get_current_branch(self, repo_path: Path, timeout: int = 10) -> str | None:
```
enyst, line 119: Ditto, we must have this code already, is this a duplicate?

pyproject.toml
```diff
@@ -79,6 +79,24 @@ python_classes = ["Test*"]
 python_functions = ["test_*"]
 addopts = "-v --tb=short"
 asyncio_mode = "auto"
+markers = [
+    "forked: marks tests as requiring process isolation (deselect with '-m \"not forked\"')",
+]
+
+# Coverage configuration - needed for --forked tests to collect coverage data
+[tool.coverage.run]
```
enyst: This doesn't seem part of this PR, maybe it could be in a separate PR if you'd like it in?

enyst:
@openhands Install yourself `gh` if you don't already have it, and then get all this PR diff and the inline comments to it.

GOAL: answer a code design question: does this feature implementation belong in the sdk package of the SDK (like this PR), in the workspace package of the SDK (`Workspace` classes), or in the client apps?

I want you to make a throughout research in the SDK packages and try to understand the git-related operations and responsibilities. Answer directly in a github comment.

enyst:
Thanks for flagging this — I dug through the SDK packages to see where “git-ish” responsibilities currently live.

### What exists today

**SDK (`openhands-sdk`) already does small, self-contained git + caching for *SDK-level assets*.**

* `openhands.sdk.context.skills.utils.update_skills_repository()` maintains a local clone under `~/.openhands/cache/skills/...` and does `git clone/fetch/reset` via `subprocess`.
  * This is *not* workspace-scoped: it’s a pure “fetch a read-only dependency into a user cache” helper.

**Workspace (`openhands-workspace`) is about running commands / managing environments**, not about bringing in additional repos into ~/.openhands caches.

* The workspace layer exposes git *introspection* APIs (`git_changes`, `git_diff`) for the repo inside the workspace, but it does **not** implement `git clone` / remote fetching logic. It’s basically “operate on whatever repo is already in the workspace”.

### So where should `Plugin.fetch()` live?

**Assessment: `Plugin.fetch()` belongs in `openhands-sdk` (the sdk package), but ideally as a shared “remote repo fetch + cache” utility used by both skills + plugins, not in `Workspace` or client apps.**

**Why SDK (not Workspace):**
- The feature is “resolve a plugin source string → ensure a local cached copy exists → return a filesystem path”. That mirrors the existing public-skills behavior and is independent of any particular workspace implementation (Docker, API remote, cloud, etc).
- The cache location is under the user’s home (`~/.openhands/cache/...`), which is outside the workspace’s conceptual boundary. Workspaces should generally avoid mutating host-level caches (especially for remote workspaces).
- Workspace APIs today are about *a repo that already exists in the workspace* (diff/changes), not “bring external dependencies into existence”.

**Why not client apps:**
- If every client re-implements “git clone with caching + ref checkout + update semantics”, we’ll end up with duplicated, inconsistent behavior. This is a reusable SDK primitive (plugins, skills registry, future org registries, etc).
- Keeping it in the SDK makes it testable and available to multiple clients.

### Design suggestion (to reduce duplication)
Right now the PR introduces `openhands.sdk.plugin.git_helper.GitHelper` while `openhands.sdk.git.utils` and `skills/utils.py` already contain git wrappers.

I’d recommend unifying these as:
1) a single `openhands.sdk.git` abstraction for *both* "git introspection" and "git fetch/clone" operations, and
2) small domain wrappers:
   - `skills` calls something like `sdk.git.cached_clone_or_update(...)`
   - `plugin` calls the same and adds “parse plugin source → choose cache key/name”.

That keeps git operations centralized and avoids the “we have 2-3 ways to call git” situation.

jpshackelford:
@enyst: 

## Git Operations Consolidation - Complete

Per your feedback about duplicate git operations, we've completed a full consolidation. Here's what was done:

### Summary

We've refactored `GitHelper` in `cached_repo.py` to use the existing `run_git_command()` utility from `git/utils.py` instead of calling `subprocess` directly. This eliminates the code duplication you identified.

### Changes Made

1. **Extended `run_git_command()`** with an optional `timeout` parameter (default: 30s) to support the varying timeout needs of different git operations (clone: 120s, fetch: 60s, etc.)

2. **Refactored `GitHelper` class** - All methods now delegate to `run_git_command()`:
   - `clone()`, `fetch()`, `checkout()`, `reset_hard()`, `get_current_branch()`
   - Removed ~70 lines of duplicate subprocess handling code

3. **Unified error handling** - All git operations now use `GitCommandError` (the structured exception with `command`, `exit_code`, and `stderr` fields) instead of the simpler `GitError` with just a message string

4. **Updated exports** - Added `GitCommandError` to the plugin module's public API for backwards compatibility

### Result: Single Code Path

All git subprocess calls in the SDK now flow through `run_git_command()`:

```
git/utils.py:run_git_command()     ← Single entry point for all git subprocess calls
    ↑
git/cached_repo.py:GitHelper       ← Used by skills & plugins
    ↑
├── context/skills/utils.py        ← Skills repository caching  
└── plugin/fetch.py                ← Plugin repository caching
```

openhands-sdk/openhands/sdk/git/cached_repo.py
jpshackelford: See https://github.com/OpenHands/software-agent-sdk/pull/1647#issuecomment-3752531365

openhands-sdk/openhands/sdk/plugin/git_helper.py
jpshackelford: https://github.com/OpenHands/software-agent-sdk/pull/1647#issuecomment-3752531365

openhands-sdk/openhands/sdk/git/cached_repo.py
jpshackelford: https://github.com/OpenHands/software-agent-sdk/pull/1647#issuecomment-3752531365

jpshackelford:
## Code Review Follow-up

Addressing the following cleanup items from /codereview-roasted

1. **Remove global mutable state for GitHelper** - Replace `get_git_helper()` / `set_git_helper()` pattern with explicit instantiation at call sites
2. **Clean up exports from plugin/__init__.py** - Remove internal `GitHelper`, `GitError`, `GitCommandError` from plugin module exports
3. **Fix silent exception swallowing** - Make error handling consistent in `_update_repository` and `_checkout_ref`
4. **Consolidate duplication** - Clean up overlap between `fetch_plugin()` and `cached_clone_or_update()`

Each will be addressed in a separate commit.

openhands-sdk/openhands/sdk/plugin/fetch.py
```diff
@@ -0,0 +1,252 @@
...
+
+    return name[:32] if name else "plugin"
+
+
+def fetch_plugin(
+    source: str,
+    cache_dir: Path | None = None,
+    ref: str | None = None,
+    update: bool = True,
+    subpath: str | None = None,
```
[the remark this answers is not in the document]
jpshackelford: Updated description.

openhands-sdk/openhands/sdk/git/cached_repo.py
```diff
@@ -0,0 +1,282 @@
...
+        Path to the local repository if successful, None on failure.
+        Returns None (not raises) on git errors to allow graceful degradation.
+    """
+    git = git_helper if git_helper is not None else GitHelper()
+
+    try:
+        if repo_path.exists() and (repo_path / ".git").exists():
+            if update:
+                logger.debug(f"Updating repository at {repo_path}")
+                _update_repository(repo_path, ref, git)
```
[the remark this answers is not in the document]
jpshackelford: Fixed in 30fc4d8e. Removed the return value from `_update_repository()` since the caller doesn't need to distinguish between "updated successfully" vs "using stale cache" — both cases return a usable path. The function now returns `None` and logs warnings internally on failure.

openhands-sdk/openhands/sdk/git/cached_repo.py
```diff
@@ -0,0 +1,282 @@
...
+        else:
+            current_branch = git.get_current_branch(repo_path)
+            if current_branch:
+                git.reset_hard(repo_path, f"origin/{current_branch}")
+    except GitCommandError as e:
+        logger.warning(f"Failed to checkout/reset: {e}. Using cached version.")
+        return False
+
+    logger.debug("Repository updated successfully")
+    return True
```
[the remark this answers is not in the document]
jpshackelford: Fixed in 30fc4d8e. Rather than just adding a log message, we now properly handle detached HEAD state by recovering to the default branch. When `update=True` is called without a ref and the repo is in detached HEAD (e.g., from a previous tag checkout), we:

1. Query `origin/HEAD` to determine the remote's default branch
2. Checkout and reset to that branch

This ensures `Plugin.fetch(source, update=True)` always means "get the latest" rather than staying stuck on a previous ref. Added `get_default_branch()` method to `GitHelper` and comprehensive tests for this behavior.

openhands-sdk/openhands/sdk/git/cached_repo.py
```diff
@@ -0,0 +1,282 @@
...
+            timeout: Timeout in seconds.
+
+        Raises:
+            GitCommandError: If clone fails.
+        """
+        cmd = ["git", "clone"]
+
+        if depth is not None:
+            cmd.extend(["--depth", str(depth)])
+
```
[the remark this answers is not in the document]
jpshackelford: Fixed in 30fc4d8e. Added documentation to `GitHelper.clone()` docstring:

```
depth: Clone depth (None for full clone, 1 for shallow). Note that
    shallow clones only fetch the tip of the specified branch. If you
    later need to checkout a specific commit that isn't the branch tip,
    the checkout may fail. Use depth=None for full clones if you need
    to checkout arbitrary commits.
```

enyst:
@openhands Do a /codereview-roasted on this PR. Post your feedback as comment on github, and use real newlines; take into account that your comment will be rendered in markdown so make it with sections, headings, bulleted lists, etc.

enyst:
## /codereview-roasted: `Plugin.fetch()` (PR OpenHands/OpenHands#1647)

**Taste rating:** 🟡 *Acceptable* — the core idea is fine and the direction (SDK-level, cacheable, testable) makes sense, but the implementation still trips over a few classic “we made it work, now we have to live with it” problems.

This PR is basically two features stapled together:

1. `Plugin.fetch()` UX/API
2. A generic “cached git repo” mechanism (now shared with skills)

That’s the right split *in principle*, but there are a couple of places where the split is leaky or the semantics are surprising.

---

## [CRITICAL ISSUES] (must fix / likely to bite users)

### 1) `cached_clone_or_update()` returning `None` on failure is a footgun
[openhands-sdk/openhands/sdk/git/cached_repo.py, ~168–223]

`cached_clone_or_update()` is a low-level infra primitive, yet it returns `Path | None` and silently degrades on basically any exception.

Problems:
- Callers have to remember to check for `None` (some will forget).
- The actual failure details are now “buried in logs” — great if you love debugging CI logs, not great if you’re a user.
- “Graceful degradation” is not always appropriate (plugins and skills are dependencies; stale/corrupt state can be worse than a hard failure).

Suggestion:
- Make the primitive **raise** a typed error (`GitCommandError` is already there) and let *callers* decide whether to degrade.
- If you really want a soft-fail helper, provide **two** entry points:
  - `cached_clone_or_update(...) -> Path` (strict)
  - `try_cached_clone_or_update(...) -> Path | None` (explicitly best-effort)

Right now the API advertises “maybe you get a repo, maybe you don’t, shrug”. That’s not infrastructure — that’s a liability.

### 2) `ref` semantics are mushy (branch/tag/commit are not the same thing)
[openhands-sdk/openhands/sdk/git/cached_repo.py, ~366–398]

The code tries to treat `ref` as “branch or tag or commit”, and then does a best-effort `reset --hard origin/{ref}`.

Issues:
- For tags: `origin/{tag}` typically doesn’t exist.
- For commits: `git fetch origin <sha>` is not a meaningful refspec on most remotes.
- For branches: `checkout <branch>` may create a local branch that does *not* track origin unless configured.

You partially paper over this with a try/except + debug logs, but the behavior is still inconsistent.

Suggestion:
- Decide what the contract is:
  - If `ref` is a branch name → update it to match remote branch.
  - If `ref` is a tag → checkout tag (detached head is expected).
  - If `ref` is a commit sha → either disallow for shallow clones or do an explicit fetch with `--depth`/`--filter` strategy.
- Implement ref-type detection via `git rev-parse --verify` / `show-ref` / `ls-remote` instead of “try it and see”.

### 3) Concurrency: cache directory is not safe under parallel usage
[openhands-sdk/openhands/sdk/git/cached_repo.py, ~168–246]

Multiple processes/threads can race on the same cache path:
- one clones while another fetches
- one deletes the directory while another is reading
- partial clone leaves a broken repo

Right now it’s “best effort + rmtree”. That’s a recipe for occasional, un-reproducible user pain.

Suggestion:
- Add a lock (even a simple `filelock` style lockfile) per `repo_path`.
- Clone into a temp dir and `rename()` atomically into place.

---

## [IMPROVEMENT OPPORTUNITIES] (good taste / maintainability)

### 4) `parse_plugin_source()` is doing too much guessing
[openhands-sdk/openhands/sdk/plugin/fetch.py, ~24–79]

The heuristics:
- treat anything with `/` and without `://` as “local”

…means things like `owner/repo` (common shorthand) will be interpreted as a local path. That’s not “helpful”; it’s silently wrong.

Suggestion:
- Keep parsing strict and explicit. If you want `owner/repo` shorthand, add it intentionally (and document it).
- Otherwise, require either `github:` or a URL scheme for remote.

### 5) `fetch_plugin()` mixes responsibilities
[openhands-sdk/openhands/sdk/plugin/fetch.py, ~176–252]

`fetch_plugin()`:
- parses and normalizes sources
- selects cache path
- performs git operations
- applies `subpath`

It’s not *huge*, but it’s already becoming a kitchen sink.

Suggestion:
- Split into:
  - `resolve_source(source) -> ResolvedSource(type, url_or_path)`
  - `ensure_repo_cached(url, cache_dir, ref, update) -> Path`
  - `apply_subpath(root, subpath) -> Path`

This also makes it easier to evolve behavior later (e.g., auth, GitHub API tarballs, signature verification).

### 6) Export surface area: don’t turn `plugin` into a grab bag
[openhands-sdk/openhands/sdk/plugin/__init__.py]

Exporting `parse_plugin_source` is… questionable. It’s an internal helper, not a stable API.

Suggestion:
- Keep the public API to `Plugin.fetch()` + `PluginFetchError`.
- If you want to expose parsing, make it clearly public and stable (`PluginSource.parse(...)` style), otherwise keep it private.

### 7) Tests: good coverage, but integration tests still lean on `subprocess.run`
[tests/sdk/plugin/test_plugin_fetch_integration.py]

You introduced `run_git_command()` to centralize safe git execution, but the integration tests still call raw `subprocess.run` everywhere.

Suggestion:
- For test repo setup, `subprocess.run` is fine.
- For “the thing you’re testing” (git behaviors), prefer the same helper used by prod code, or at least be consistent about timeouts/error formatting.

---

## [STYLE NOTES] (minor, but annoying)

### 8) “`pass` Exception” class is noise
[openhands-sdk/openhands/sdk/plugin/fetch.py, 18–21]

`class PluginFetchError(Exception): pass` adds nothing. Either add meaningful doc + fields, or just use a plain exception type.

Suggestion:
- If you keep it, consider storing structured context (`source`, `ref`, `cache_dir`, `cause`) so callers can show actionable error messages.

### 9) Logging level mismatch
[cached_repo.py]

Clone logs at `info`, update logs at `debug`, failures at `warning`. That’s inconsistent for a cache primitive.

Suggestion:
- Pick a consistent scheme: `info` for one-liners that users care about, `debug` for details, `error` only when you’re going to raise.

---

## Verdict

✅ **Concept is worth merging** (remote plugin fetching + cache is a real need).

❌ **Implementation needs tightening** around:
- failure semantics (stop returning `None` from infra primitives)
- ref handling (branch/tag/commit behavior should be explicit)
- cache concurrency (locking/atomic clone)

**Key insight:** you’re building a dependency manager. Dependency managers don’t get to be “best effort” unless you also enjoy support tickets.

openhands-sdk/openhands/sdk/plugin/fetch.py
```diff
@@ -0,0 +1,252 @@
...
+        return ("local", source)
+
+    raise PluginFetchError(
+        f"Unable to parse plugin source: {source}. "
+        f"Expected formats: 'github:owner/repo', git URL, or local path"
+    )
+
+
+def _is_git_url(source: str) -> bool:
+    """Check if a source string looks like a git URL.
```
enyst: Totally a nit, we know these functions happen to enter into the codebase together with the plugin implementation, but this is `git` stuff and it's potentially useful for more `git` uses, so it probably belongs in the `git` module

openhands-sdk/openhands/sdk/plugin/fetch.py
```diff
@@ -0,0 +1,252 @@
...
+        return True
+
+    # File protocol (for testing)
+    if source.startswith("file://"):
+        return True
+
+    return False
+
+
+def _normalize_git_url(url: str) -> str:
```
enyst: Ditto

openhands-sdk/openhands/sdk/plugin/fetch.py
```diff
@@ -0,0 +1,252 @@
...
+    readable_name = _extract_readable_name(source)
+
+    cache_name = f"{readable_name}-{source_hash}"
+    return cache_dir / cache_name
+
+
+def _extract_readable_name(source: str) -> str:
+    """Extract a human-readable name from a source URL/path.
+
+    Extracts the last path component (repo name) and sanitizes it.
```
enyst: Ditto

FWIW this one feels duplicate too? I mean I feel we've done before some extraction of the repo name? Though I don't recall if it's currently in the SDK, it just would be nice to have it as part of a small `git utilities that a software SDK likely needs` module, which is apparently our `git` module

pyproject.toml
enyst: Sorry, this doesn't seem solved so I marked it unsolved. Could we move it in a separate PR or is there a reason you need it here?

jpshackelford: Removed in 28e6aa4a. The pytest markers and coverage config were not needed for the plugin fetch feature itself.

jpshackelford:
Thanks for the thorough review! I've addressed several of the points raised:

## Fixed

### OpenHands/OpenHands#6 - Remove `parse_plugin_source` from public exports ✅
Commit 8d2e290: Removed `parse_plugin_source` from `__all__`. The public API is now just `Plugin.fetch()` + `PluginFetchError`.

### OpenHands/OpenHands#5 - Split `fetch_plugin()` into smaller functions ✅
Commit 64098bb: Extracted three helper functions:
- `_resolve_local_source()` - handles local path resolution
- `_fetch_remote_source()` - handles remote git fetching  
- `_apply_subpath()` - validates and applies subpath

The main `fetch_plugin()` is now ~15 lines.

### OpenHands/OpenHands#3 - Add cache concurrency safety ✅
Commit d472139: Added file-based locking using `filelock` (already a dependency). The lock file is created adjacent to the repo directory (`repo_path.lock`). Default timeout is 30 seconds (configurable). Lock is released automatically on success or error.

openhands-sdk/openhands/sdk/plugin/fetch.py
jpshackelford: done

openhands-sdk/openhands/sdk/plugin/fetch.py
jpshackelford: done

openhands-sdk/openhands/sdk/plugin/fetch.py
jpshackelford: done

enyst (approved):
Thank you for the work on this! I gave quite a bit of thought to this functionality and I believe we're on the right track thanks to your work on it. I had some tiny comments inline, but they were just nits anyway.

jpshackelford:
**Re: ref semantics are mushy (branch/tag/commit are not the same thing)**

Thanks for the thorough review. You're right that the original code had issues:
- `origin/{tag}` doesn't exist (tags aren't namespaced under origin/)
- `git fetch origin <sha>` isn't a valid refspec
- The try/except approach was wasteful and the contract unclear

**What we considered:**

Explicit ref-type pre-detection as you suggested, using `git rev-parse --verify`, `show-ref`, or `ls-remote` to determine if a ref is a branch, tag, or commit before acting on it.

**Why we decided against it:**

1. **Network latency**: `git ls-remote` requires a network round-trip on every checkout to check remote refs
2. **Ambiguous edge cases**: Local ref parsing can't reliably distinguish short SHAs from branch names. A branch can legally be named `abc1234` or `deadbeef`, making regex-based detection unreliable
3. **Complexity**: Would need to handle cases where a name exists as both a branch and tag, or where local and remote refs differ

**What we did instead:**

Detect ref type *after* checkout by checking HEAD state:
- `git checkout <ref>` resolves the ref type (git already handles the ambiguity)
- `git rev-parse --abbrev-ref HEAD` returns `"HEAD"` for detached state (tag/commit) or the branch name
- Only reset to `origin/{branch}` when we're actually on a branch

This is a single local command with no ambiguity, since git has already resolved what the ref was during checkout.

PR #1649, opened 2026-01-08
fix(condenser): View properties for capturing API assumptions

csmith49:
## Summary

This PR is a WIP.

The goal is to refactor the `View` object so that all of the consistency checks -- batch structure, tool loops + thinking blocks, etc -- can be easily represented in parallel.

### Why

The `View` inherits some behavior from v0's `ConversationMemory` class, which was responsible for making sure that certain properties held in the event stream before it was sent up to the LLM. Example properties:

* Action/Observation pairs not split
* Action batches not split
* Tool use loops not split

We expect the LLM API to preserve these features. They'll only be violated when we start to mess with the event stream...like the condenser does.

In recent PRs the `View`/`Condenser` interaction has been changed to rely on manipulation indices, where the condenser now tries to maintain the properties as it drops events. We've kept the "enforcement" side of things as a safety plan, but ideally it never matters.

However, both of those processes are tightly-coupled (wrt the properties enforced) loops. To make it easier to manage these properties this PR works to separate the properties into classes that can be managed in parallel.

### Design Decisions and Assumptions

View properties are simple classes that implement a two-function API:

1. `enforce`, which checks that the property holds, and
2. `manipulation_indices`, which produces a set of indices that the condenser can use while preserving the property

In theory only the manipulation indices are necessary, but experience has taught us that sometimes weird things happen. By keeping the enforcement function around and logging the output with big scary warnings we can recover from these weird things, with the option to disable enforcement or simplify the codebase in the future.

#### Manipulation indices

One reason this works is because the manipulation index calculations can be done independently and combined simply (just taking the intersection of the sets).

It's possible that future properties won't be representable within this framework. For example, maybe we're allowed to drop at most one thinking block from the previous message. There's no way to capture that with the intersection of manipulation indices.

That may never end up being a problem. But if so all we need to do is update the manipulation index representation and how the condensers use them. That's a big implementation lift but it still fits in this conceptual framework.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:7ae4e94-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-7ae4e94-python \
  ghcr.io/openhands/agent-server:7ae4e94-python
```

```
ghcr.io/openhands/agent-server:7ae4e94-golang-amd64
ghcr.io/openhands/agent-server:7ae4e94-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:7ae4e94-golang-arm64
ghcr.io/openhands/agent-server:7ae4e94-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:7ae4e94-java-amd64
ghcr.io/openhands/agent-server:7ae4e94-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:7ae4e94-java-arm64
ghcr.io/openhands/agent-server:7ae4e94-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:7ae4e94-python-amd64
ghcr.io/openhands/agent-server:7ae4e94-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:7ae4e94-python-arm64
ghcr.io/openhands/agent-server:7ae4e94-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:7ae4e94-golang
ghcr.io/openhands/agent-server:7ae4e94-java
ghcr.io/openhands/agent-server:7ae4e94-python
```

- Each variant tag (e.g., `7ae4e94-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `7ae4e94-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/context/view/properties/base.py
+++ b/openhands-sdk/openhands/sdk/context/view/properties/base.py
@@ -24,7 +24,9 @@ class ViewPropertyBase(ABC):
 
     @abstractmethod
     def enforce(
-        self, current_view_events: list[LLMConvertibleEvent], all_events: list[Event]
+        self,
+        current_view_events: Sequence[LLMConvertibleEvent],
+        all_events: Sequence[Event],
     ) -> set[EventID]:
         """Enforce the property on a list of events.
 
@@ -41,7 +43,9 @@ class ViewPropertyBase(ABC):
 
     @abstractmethod
     def manipulation_indices(
-        self, current_view_events: list[LLMConvertibleEvent], all_events: list[Event]
+        self,
+        current_view_events: Sequence[LLMConvertibleEvent],
+        all_events: Sequence[Event],
     ) -> ManipulationIndices:
         """Get manipulation indices for the property on a list of events.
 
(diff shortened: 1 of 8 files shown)


csmith49:
Closing in favor of #2116

Issue #1671, opened 2026-01-09
[Bug] ConversationState fails to restore when secrets are serialized without cipher/expose_secrets


xingyaoww:
## Bug Description

When a `ConversationState` is persisted and later restored, the deserialization fails with a Pydantic validation error if the secrets in `secret_registry.secret_sources` were serialized without a cipher or `expose_secrets=True` context.

## Root Cause

The issue is in the secret serialization/deserialization flow:

1. When `ConversationState._save_base_state()` is called, it uses `model_dump_json(exclude_none=True)` without providing a cipher or `expose_secrets` context
2. The `serialize_secret()` function in `utils/pydantic_secrets.py` returns the raw `SecretStr` object when no cipher/expose_secrets is provided (line 28)
3. Pydantic then serializes this as `"**********"` (default masking)
4. When deserializing, `validate_secret()` sees `"**********"` and returns `None` (line 52: `if secret_value == "**********": return None`)
5. However, `StaticSecret.value` is defined as `value: SecretStr` (required, not optional), so validation fails

## Error Message

```
pydantic_core._pydantic_core.ValidationError: 6 validation errors for ConversationState
secret_registry.secret_sources.LLM_API_KEY.value
  Field required [type=missing, input_value={"description": "This is ..."}, input_type=dict]
secret_registry.secret_sources.DD_API_KEY.value
  Field required [type=missing, input_value={"description": "Datadog API Key"}, input_type=dict]
...
```

## Steps to Reproduce

1. Create a `ConversationState` with secrets in `secret_registry`
2. Let the state auto-persist (or manually call `_save_base_state()`)
3. Restart the server/process
4. Try to restore the conversation using `ConversationState.create()` with the same persistence_dir
5. Observe the Pydantic validation error

## Expected Behavior

The conversation should be restored successfully, either by:
- Making `StaticSecret.value` optional (`value: SecretStr | None = None`)
- Or gracefully handling the case where secrets cannot be restored (with appropriate logging/warning)
- Or always using a cipher when persisting secrets

## Actual Behavior

The conversation fails to restore with a validation error, making the conversation permanently inaccessible.

## Environment

- openhands-sdk version: 1.7.4 (also verified in 1.8.1)
- Discovered in production via Datadog logs

## Affected Files

- `openhands/sdk/secret/secrets.py` - `StaticSecret` class
- `openhands/sdk/utils/pydantic_secrets.py` - `serialize_secret()` and `validate_secret()` functions
[...]

PR #1672, opened 2026-01-09
Fix conversation restore failing due to secret serialization/deserialization mismatch

tofarr:
## Summary

This PR fixes an issue where V1 conversations fail to restore from persistence when the conversation state contains secrets in the `secret_registry`.

### Problem
Secrets were serialized as redacted values (`**********`) but `StaticSecret.value` was a required field that couldn't accept `None` after validation converted `**********` to `None`.

#### Part 1: Make secrets optional (prevents crash)
- Made `StaticSecret.value` optional with `None` default
- Updated `StaticSecret.get_value()` to handle `None` values gracefully
- Fixed `LookupSecret` header validators to skip redacted headers instead of crashing

#### Part 2: Add cipher support (preserves secrets)
- Added `cipher` parameter to `ConversationState.create()` and `LocalConversation.__init__()`
- When a cipher is provided, secrets are encrypted on save and decrypted on load
- `EventService` now passes its cipher to `LocalConversation`

### How it works now

1. **With cipher (recommended for production):**
   - Secrets are encrypted when saving state
   - Secrets are decrypted when loading state
   - Secret values are preserved across restarts

2. **Without cipher (fallback):**
   - Secrets are redacted as `**********` during serialization
   - Redacted secrets deserialize to `StaticSecret(value=None)`
   - Conversation loads successfully, but secrets need to be re-provided

### Files Changed
- `openhands-sdk/openhands/sdk/secret/secrets.py` - Make StaticSecret.value optional, fix LookupSecret headers
- `openhands-sdk/openhands/sdk/conversation/state.py` - Add cipher support for encryption/decryption
- `openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py` - Add cipher parameter
- `openhands-agent-server/openhands/agent_server/event_service.py` - Pass cipher to LocalConversation
- `tests/sdk/conversation/local/test_state_serialization.py` - Add regression tests
- `tests/sdk/conversation/test_secret_source.py` - Add regression tests

## Checklist

@tofarr can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

## Fixes

* https://github.com/OpenHands/software-agent-sdk/issues/1505
* https://github.com/OpenHands/software-agent-sdk/issues/1671
* https://github.com/OpenHands/OpenHands/issues/12344

## Testing

image
image
image
image
image

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:7cc139d-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-7cc139d-python \
  ghcr.io/openhands/agent-server:7cc139d-python
```

```
ghcr.io/openhands/agent-server:7cc139d-golang-amd64
ghcr.io/openhands/agent-server:7cc139d-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:7cc139d-golang-arm64
ghcr.io/openhands/agent-server:7cc139d-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:7cc139d-java-amd64
ghcr.io/openhands/agent-server:7cc139d-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:7cc139d-java-arm64
ghcr.io/openhands/agent-server:7cc139d-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:7cc139d-python-amd64
ghcr.io/openhands/agent-server:7cc139d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:7cc139d-python-arm64
ghcr.io/openhands/agent-server:7cc139d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:7cc139d-golang
ghcr.io/openhands/agent-server:7cc139d-java
ghcr.io/openhands/agent-server:7cc139d-python
```

- Each variant tag (e.g., `7cc139d-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `7cc139d-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/sockets.py
+++ b/openhands-agent-server/openhands/agent_server/sockets.py
@@ -142,7 +142,7 @@ async def _send_event(event: Event, websocket: WebSocket):
         dumped = event.model_dump(mode="json")
         await websocket.send_json(dumped)
     except Exception:
-        logger.exception("error_sending_event:{event}", stack_info=True)
+        logger.exception("error_sending_event: %r", event, stack_info=True)
 
 
 @dataclass
@@ -160,7 +160,7 @@ async def _send_bash_event(event: BashEventBase, websocket: WebSocket):
         dumped = event.model_dump(mode="json")
         await websocket.send_json(dumped)
     except Exception:
-        logger.exception("error_sending_event:{event}", stack_info=True)
+        logger.exception("error_sending_bash_event: %r", event, stack_info=True)
 
 
 @dataclass


tofarr:
@OpenHands Please implement the suggestions left by all-hands-bot

openhands-agent-server/openhands/agent_server/config.py
```diff
@@ -26,6 +26,9 @@ def _default_session_api_keys():
 
 def _default_secret_key() -> SecretStr | None:
     session_api_key = os.getenv(SESSION_API_KEY_ENV)
+    if session_api_key:
+        return SecretStr(session_api_key)
+    session_api_key = os.getenv("OH_SESSION_API_KEYS_0")
```
neubig: I didn't understand the necessity for this, could you at least leave a comment here to explain why this two part solution is necessary?

Also,
```
session_api_key = os.getenv(SESSION_API_KEY_ENV) or os.getenv("OH_SESSION_API_KEYS_0")
```
may be more concise.

enyst:
@openhands Read the full diff of this PR and replace ALL occurrences of “ALL-4846” with “issue 1505”. Double check the diff when you’re done. Commit and push.

openhands-agent-server/openhands/agent_server/config.py
enyst: I’d also love to know why are more than 1 needed? `…KEY_0` suggests than more the one is used

(While I can’t find it right now, I’ve seen that people have documented a variable OH_SESSION_KEY I believe, for LLM key encryption. I wonder if such key still works?)

tofarr: Having more than one seasion_api_key allowed here allows key rotation. The first one in the list is the default, but others are permitted keys. This is not a workflow we use at present, but will likely become necessary as we start to move into an architecture where sandboxes contain multiple conversations, persist longer and are tied to a user/org.

The actual change here was to support the key format we already use in the docker container in the OpenHands repo

tofarr: The single SESSION_API_KEY is still supported for backwards compatibility - this has not changed.

I think I need to add a readme in a separate PR describing all this

tofarr: And it goes without saying - the current runtime API only supports a single session api key.

tofarr: @enyst - you may also be thinking of the JWTService in the app server which also has the capacity for key rotation, but is more complex than this. That service:
* Is based on JWT / JWE
* Allows for multiple keys.
* Each key has a unique id used for decryption.
* The JWT tokens include the `kid` header (key id), specifying which key was used to encrypt / sign them.

tofarr: Updated docstrings

openhands-agent-server/openhands/agent_server/config.py
neubig:
```suggestion
SESSION_API_KEY_ENVS = ["SESSION_API_KEY", "OH_SESSION_API_KEYS_0"]
```

neubig, line 30:
```suggestion
result = []
for env_var in SESSION_API_KEY_ENVS:
    session_api_key = os.getenv(env_var)
    if session_api_key:
        result.append(session_api_key)
return result
```

openhands-agent-server/openhands/agent_server/config.py
```diff
@@ -25,7 +25,16 @@ def _default_session_api_keys():
...
+    check OH_SESSION_API_KEYS_0 - The new variable, since we support multiple values
+    for key rotation.
+    """
     session_api_key = os.getenv(SESSION_API_KEY_ENV)
+    if session_api_key:
+        return SecretStr(session_api_key)
+    session_api_key = os.getenv("OH_SESSION_API_KEYS_0")
     if session_api_key:
         return SecretStr(session_api_key)
     return None
```
neubig:
```suggestion
session_api_keys = _default_session_api_keys()
if len(session_api_keys) > 0:
    return SecretStr(session_api_keys[0])
return None
```

openhands-agent-server/openhands/agent_server/config.py
tofarr: I've updated the docs again - hopefully to make things clearer.

I think the confusion comes from understanding of the `EnvParser` which will automatically read pydantic models from the environment. These default methods are fallbacks for when we are doing something unusual - in this case grabbing the session_api_key from the legacy variable if it is not defined.

We use this [EnvParser](https://github.com/OpenHands/software-agent-sdk/blob/f011127916917cb3a3a2c29d2c5b151ef2464f44/openhands-agent-server/openhands/agent_server/env_parser.py) mechanism to configure complex structures like webhooks:
https://github.com/OpenHands/OpenHands/blob/7862e10f038d304ec79cdee46f7fb1e1842325bd/openhands/app_server/sandbox/docker_sandbox_service.py#L320

And we use it heavily in the app server for configuration - All the injected services can literally be replaced using the environment if needed.

enyst: > These default methods are fallbacks for when we are doing something unusual

We could maybe name those methods something else than `default_` since in some sense, they're not actually the default values that just come into place, they're for when unusual handling is needed.

PR #1721, opened 2026-01-13
feat(condenser): Multi-summary views

csmith49:
## Summary

While working on #1596 we came across a problem: `View` objects are limited to a single summary event. This limits our condensation abilities and adds some brittle corner-case logic in the condensers.

The fix is to make it so views can have multiple summaries simultaneously.

This immediately introduces some problems:
* A lot of specialized functions assume there's at most one summary.
* With multiple summaries, we might not keep/forget all at the same time.
* Summaries are "synthetic" events introduced between the conversation and the agent, and so can't be referenced by `Condensation` events because they don't have consistent IDs.

So this PR deletes the specialized functions and adds a standard pattern for summary IDs. The logic for constructing the view is also simplified to process condensations one at a time instead of all at once, and some functions that were enforcing constraints on the view have their signatures reworked.

Additional changes include:
* Updated tests to remove references to unused functions
* A new test suite for multi-summary views
* Updates to the `LLMSummarizingCondenser` to handle the new view structure.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:2003742-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-2003742-python \
  ghcr.io/openhands/agent-server:2003742-python
```

```
ghcr.io/openhands/agent-server:2003742-golang-amd64
ghcr.io/openhands/agent-server:2003742-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:2003742-golang-arm64
ghcr.io/openhands/agent-server:2003742-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:2003742-java-amd64
ghcr.io/openhands/agent-server:2003742-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:2003742-java-arm64
ghcr.io/openhands/agent-server:2003742-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:2003742-python-amd64
ghcr.io/openhands/agent-server:2003742-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:2003742-python-arm64
ghcr.io/openhands/agent-server:2003742-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:2003742-golang
ghcr.io/openhands/agent-server:2003742-java
ghcr.io/openhands/agent-server:2003742-python
```

- Each variant tag (e.g., `2003742-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `2003742-python-amd64`) are also available if needed

### DIFF
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -65,6 +65,11 @@ MODELS = {
         "display_name": "MiniMax M2",
         "llm_config": {"model": "litellm_proxy/minimax/minimax-m2"},
     },
+    "minimax-m2.1": {
+        "id": "minimax-m2.1",
+        "display_name": "MiniMax M2.1",
+        "llm_config": {"model": "litellm_proxy/minimax/MiniMax-M2.1"},
+    },
     "deepseek-v3.2-reasoner": {
         "id": "deepseek-v3.2-reasoner",
         "display_name": "DeepSeek V3.2 Reasoner",
--- a/.github/workflows/pr-artifacts.yml
+++ b/.github/workflows/pr-artifacts.yml
@@ -1,0 +1,139 @@
+---
+name: PR Artifacts
+
+on:
+    workflow_dispatch: # Manual trigger for testing
+    pull_request:
+        types: [opened, synchronize, reopened]
+        branches: [main]
+    pull_request_review:
+        types: [submitted]
+
+jobs:
+  # Auto-remove .pr/ directory when a reviewer approves
+    cleanup-on-approval:
+        concurrency:
+            group: cleanup-pr-artifacts-${{ github.event.pull_request.number }}
+            cancel-in-progress: false
+        if: github.event_name == 'pull_request_review' && github.event.review.state == 'approved'
+        runs-on: ubuntu-latest
+        permissions:
+            contents: write
+            pull-requests: write
+        steps:
+            - name: Check if fork PR
+              id: check-fork
+              run: |
+                  if [ "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.event.pull_request.base.repo.full_name }}" ]; then
+                    echo "is_fork=true" >> $GITHUB_OUTPUT
+                    echo "::notice::Fork PR detected - skipping auto-cleanup (manual removal required)"
+                  else
+                    echo "is_fork=false" >> $GITHUB_OUTPUT
+                  fi
+
+            # Use PAT so the push triggers CI workflows that will complete and
+            # satisfy branch protection. We can't use [skip ci] because the Vercel
+            # GitHub App creates stuck checks that block merging.
+            - uses: actions/checkout@v5
+              if: steps.check-fork.outputs.is_fork == 'false'
+              with:
+                  ref: ${{ github.event.pull_request.head.ref }}
+                  token: ${{ secrets.ALLHANDS_BOT_GITHUB_PAT }}
+
+            - name: Remove .pr/ directory
+              id: remove
+              if: steps.check-fork.outputs.is_fork == 'false'
+              run: |
+                  if [ -d ".pr" ]; then
+                    git config user.name "allhands-bot"
+                    git config user.email "allhands-bot@users.noreply.github.com"
+                    git rm -rf .pr/
+                    git commit -m "chore: Remove PR-only artifacts [automated]"
+                    git push || {
+                      echo "::error::Failed to push cleanup commit. Check branch protection rules."
+                      exit 1
+                    }
+                    echo "removed=true" >> $GITHUB_OUTPUT
+                    echo "::notice::Removed .pr/ directory"
+                  else
+                    echo "removed=false" >> $GITHUB_OUTPUT
+                    echo "::notice::No .pr/ directory to remove"
+                  fi
+
+            - name: Update PR comment after cleanup
+              if: steps.check-fork.outputs.is_fork == 'false' && steps.remove.outputs.removed == 'true'
+              uses: actions/github-script@v7
+              with:
+                  script: |
+                      const marker = '<!-- pr-artifacts-notice -->';
+                      const body = `${marker}
+                      ✅ **PR Artifacts Cleaned Up**
+
+                      The \`.pr/\` directory has been automatically removed.
+                      `;
+
+                      const { data: comments } = await github.rest.issues.listComments({
+                        owner: context.repo.owner,
+                        repo: context.repo.repo,
+                        issue_number: context.issue.number,
+                      });
+
+                      const existing = comments.find(c => c.body.includes(marker));
+                      if (existing) {
+                        await github.rest.issues.updateComment({
+                          owner: context.repo.owner,
+                          repo: context.repo.repo,
+                          comment_id: existing.id,
+                          body: body,
+                        });
+                      }
+
+  # Warn if .pr/ directory exists (will be auto-removed on approval)
+    check-pr-artifacts:
+        if: github.event_name == 'pull_request'
+        runs-on: ubuntu-latest
+        permissions:
+            contents: read
+            pull-requests: write
+        steps:
+            - uses: actions/checkout@v5
+
+            - name: Check for .pr/ directory
+              id: check
+              run: |
+                  if [ -d ".pr" ]; then
+                    echo "exists=true" >> $GITHUB_OUTPUT
+                    echo "::warning::.pr/ directory exists and will be automatically removed when the PR is approved. For fork PRs, manual removal is required before merging."
+                  else
+                    echo "exists=false" >> $GITHUB_OUTPUT
+                  fi
+
+            - name: Post or update PR comment
+              if: steps.check.outputs.exists == 'true'
+              uses: actions/github-script@v7
+              with:
+                  script: |
+                      const marker = '<!-- pr-artifacts-notice -->';
+                      const body = `${marker}
+                      📁 **PR Artifacts Notice**
+
+                      This PR contains a \`.pr/\` directory with PR-specific documents. This directory will be **automatically removed** when the PR is approved.
+
+                      > For fork PRs: Manual removal is required before merging.
+                      `;
+
+                      const { data: comments } = await github.rest.issues.listComments({
+                        owner: context.repo.owner,
+                        repo: context.repo.repo,
+                        issue_number: context.issue.number,
+                      });
+
+                      const existing = comments.find(c => c.body.includes(marker));
+                      if (!existing) {
+                        await github.rest.issues.createComment({
+                          owner: context.repo.owner,
+                          repo: context.repo.repo,
+                          issue_number: context.issue.number,
+                          body: body,
+                        });
+                      }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,6 +118,45 @@ When reviewing code, provide constructive feedback:
 
 </DEV_SETUP>
 
+<PR_ARTIFACTS>
+# PR-Specific Documents
+
+When working on a PR that requires design documents, scripts meant for development-only, or other temporary artifacts that should NOT be merged to main, store them in a `.pr/` directory at the repository root.
+
+## Usage
+
+```bash
+# Create the directory if it doesn't exist
+mkdir -p .pr
+
+# Add your PR-specific documents
+.pr/
+├── design.md       # Design decisions and architecture notes
+├── analysis.md     # Investigation or debugging notes
+└── notes.md        # Any other PR-specific content
+```
+
+## How It Works
+
+1. **Notification**: When `.pr/` exists, a single comment is posted to the PR conversation alerting reviewers
+2. **Auto-cleanup**: When the PR is approved, the `.pr/` directory is automatically removed via commit
+3. **Fork PRs**: Auto-cleanup cannot push to forks, so manual removal is required before merging
+
+## Important Notes
+
+- Do NOT put anything in `.pr/` that needs to be preserved
+- The `.pr/` check passes (green ✅) during development - it only posts a notification, not a blocking error
+- For fork PRs: You must manually remove `.pr/` before the PR can be merged
+
+## When to Use
+
+- Complex refactoring that benefits from written design rationale
+- Debugging sessions where you want to document your investigation
+- Feature implementations that need temporary planning docs
+- Temporary script that are intended to show reviewers that the feature works
+- Any analysis that helps reviewers understand the PR but isn't needed long-term
+</PR_ARTIFACTS>
+
 <CODE>
 - Avoid hacky trick like `sys.path.insert` when resolving package dependency
 - Use existing packages/libraries instead of implementing yourselves whenever possible.
--- a/examples/01_standalone_sdk/34_critic_example.py
+++ b/examples/01_standalone_sdk/34_critic_example.py
@@ -1,0 +1,99 @@
+"""Example demonstrating critic-based evaluation of agent actions.
+
+This is EXPERIMENTAL.
+
+This shows how to configure an agent with a critic to evaluate action quality
+in real-time. The critic scores are displayed in the conversation visualizer.
+
+For All-Hands LLM proxy (llm-proxy.*.all-hands.dev), the critic is auto-configured
+using the same base_url with /vllm suffix and "critic" as the model name.
+"""
+
+import os
+import re
+
+from openhands.sdk import LLM, Agent, Conversation, Tool
+from openhands.sdk.critic import APIBasedCritic
+from openhands.sdk.critic.base import CriticBase
+from openhands.tools.file_editor import FileEditorTool
+from openhands.tools.task_tracker import TaskTrackerTool
+from openhands.tools.terminal import TerminalTool
+
+
+def get_required_env(name: str) -> str:
+    value = os.getenv(name)
+    if value:
+        return value
+    raise ValueError(
+        f"Missing required environment variable: {name}. "
+        f"Set {name} before running this example."
+    )
+
+
+def get_default_critic(llm: LLM) -> CriticBase | None:
+    """Auto-configure critic for All-Hands LLM proxy.
+
+    When the LLM base_url matches `llm-proxy.*.all-hands.dev`, returns an
+    APIBasedCritic configured with:
+    - server_url: {base_url}/vllm
+    - api_key: same as LLM
+    - model_name: "critic"
+
+    Returns None if base_url doesn't match or api_key is not set.
+    """
+    base_url = llm.base_url
+    api_key = llm.api_key
+    if base_url is None or api_key is None:
+        return None
+
+    # Match: llm-proxy.{env}.all-hands.dev (e.g., staging, prod, eval)
+    pattern = r"^https?://llm-proxy\.[^./]+\.all-hands\.dev"
+    if not re.match(pattern, base_url):
+        return None
+
+    return APIBasedCritic(
+        server_url=f"{base_url.rstrip('/')}/vllm",
+        api_key=api_key,
+        model_name="critic",
+    )
+
+
+llm_api_key = get_required_env("LLM_API_KEY")
+
+llm = LLM(
+    model=os.getenv("LLM_MODEL", "anthropic/claude-sonnet-4-5-20250929"),
+    api_key=llm_api_key,
+    base_url=os.getenv("LLM_BASE_URL", None),
+)
+
+# Try auto-configuration for All-Hands proxy, fall back to explicit env vars
+critic = get_default_critic(llm)
+if critic is None:
+    critic = APIBasedCritic(
+        server_url=get_required_env("CRITIC_SERVER_URL"),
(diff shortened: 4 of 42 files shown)


openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -117,33 +116,15 @@ def condensation_requirement(
         if Reason.REQUEST in reasons:
             return CondensationRequirement.HARD
 
-    def _get_summary_event_content(self, view: View) -> str:
```
csmith49, line 120: No longer needed -- the previous summaries are now captured as regular events in the summary generation.

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -291,96 +265,112 @@ def __getitem__(
...
-        return updated_removed_ids
+        # Filter out events that need to be removed
+        return [event for event in view_events if event.id not in ids_to_remove]
 
     @staticmethod
-    def filter_unmatched_tool_calls(
-        events: list[LLMConvertibleEvent],
+    def _filter_unmatched_tool_calls(
+        view_events: Sequence[LLMConvertibleEvent],
+        all_events: Sequence[Event],  # noqa: ARG004
```
csmith49: This argument isn't used just yet. But it aligns this function with `_enforce_batch_atomicity` and will make future PRs like #1649 easier to adopt.

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -291,96 +265,112 @@ def __getitem__(
...
 
         Removes ActionEvents and ObservationEvents that have tool_call_ids
         but don't have matching pairs. Also enforces batch atomicity - if any
         ActionEvent in a batch is filtered out, all ActionEvents in that batch
         are also filtered out.
+
+        Args:
+            view_events: The list of events to filter
+            all_events: The complete original list of all events (unused for now,
+                        but included to match signature pattern)
```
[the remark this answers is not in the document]
csmith49: See comments elsewhere, the unused arguments are deliberate.

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -291,96 +265,112 @@ def __getitem__(
...
+        kept_event_ids = {event.id for event in kept_events}
         tool_call_ids_to_remove: set[ToolCallID] = set()
-        for action_id in removed_event_ids:
-            if action_id in action_batch.action_id_to_tool_call_id:
-                tool_call_ids_to_remove.add(
-                    action_batch.action_id_to_tool_call_id[action_id]
-                )
+        for action_id, tool_call_id in action_batch.action_id_to_tool_call_id.items():
+            if action_id not in kept_event_ids:
+                tool_call_ids_to_remove.add(tool_call_id)
```
VascoSch92:
```python
tool_call_ids_to_remove: set[ToolCallID] = {
    tool_call_id 
    for action_id, tool_call_id in action_batch.action_id_to_tool_call_id.items()
    if action_id not in kept_event_ids
}
```

It should be slightly faster ;-)

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -435,66 +425,89 @@ def find_next_manipulation_index(self, threshold: int) -> int:
...
 
-        # Enforce batch atomicity: if any event in a multi-action batch is forgotten,
-        # forget all events in that batch to prevent partial batches with thinking
-        # blocks separated from their tool calls
-        forgotten_event_ids = View._enforce_batch_atomicity(events, forgotten_event_ids)
+        # Copy the events into output, skipping forgotten events.
+        for event in events:
+            if event.id in condensation.forgotten_event_ids:
+                continue
+            output.append(event)
```
VascoSch92:
```python
output: list[LLMConvertibleEvent] = [
    event
    for event in events
    if event.id not in condensation.forgotten_event_ids
]
```

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -435,66 +425,89 @@ def find_next_manipulation_index(self, threshold: int) -> int:
...
+                output.append(event)
 
-        for event in reversed(events):
-            if isinstance(event, Condensation):
-                break
+            else:
+                logger.debug(
+                    f"Skipping non-LLMConvertibleEvent of type {type(event)} "
+                    f"in View.from_events"
+                )
```
VascoSch92, line 472: This is the perfect place to use a `match`.

I like it because it reads like a table of contents. You immediately see the type and the action associated with it.

Of course, it’s mostly a matter of personal preference, so feel free to ignore the suggestion if you prefer the current way!

```python
for event in events:
    match event:
        case CondensationRequest():
            continue

        case Condensation():
            condensations.append(event)
            output = View.apply_condensation(output, event)

        case LLMConvertibleEvent():
            output.append(event)

        case _:
            logger.debug(
                f"Skipping non-LLMConvertibleEvent of type {type(event)} "
                f"in View.from_events"
            )
```

csmith49: Love match cases. Interestingly, there aren't many in our code base. LLMs seem to bias against them when generating code. Will probably keep this as-is, but if we could take advantage of the other features of match cases (unpacking, more complex guards) I'd make the switch.

openhands-sdk/openhands/sdk/context/view.py
csmith49: List comprehension will make this faster for sure. When you're leaving comments like this feel free to use GitHub's suggestion feature so we can easily make a commit :)

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -435,66 +425,89 @@ def find_next_manipulation_index(self, threshold: int) -> int:
...
+            output.append(event)
+
+        # Insert the condensation summary event at the specified offset, if present.
+        if condensation.summary is not None and condensation.summary_offset is not None:
+            logger.debug(f"Inserting summary at offset {condensation.summary_offset}")
+
+            # Use deterministic ID based on condensation's llm_response_id
+            # This enables subsequent condensations to reference and forget summaries
+            # Replace underscores with hyphens to comply with file persistence regex
+            summary_id = f"{condensation.llm_response_id.replace('_', '-')}-summary"
```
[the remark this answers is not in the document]
csmith49: Replaced with event ID, which means we can assume it's reasonably well-formed.

csmith49:
@enyst While working on this PR I remembered why the manipulation index calculations were strict wrt the `keep_first` parameter: because we'd keep `keep_first` events at the start, and then the summary, and _then_ all the events we cared about.

Never really mattered with the way we filtered summary events, but definitely matters after this PR.

enyst:
Ah I see.

@openhands Do a /codereview-roasted on this PR. Investigate deeply! Respond as a comment on github on this PR, so consider that your output will be rendered as markdown.

enyst:
🟡 **Taste rating: Acceptable**  
You’re solving a real problem (single-summary Views were a straightjacket), but the implementation trades “brittle corner-cases” for “now *everything* is a corner-case unless you’re very disciplined”. The core idea is fine; the execution has some sharp edges that will absolutely cut the next person who touches this.

---

## [CRITICAL ISSUES] (must fix / at least must justify loudly)

### 1) Deterministic summary IDs are a hack glued to an unrelated constraint
[openhands-sdk/openhands/sdk/context/view.py, ~451]  
```py
summary_id = f"{condensation.id.replace('_', '-')}-summary"
```

You’re encoding a *file persistence regex constraint* into the semantic identity of an event. That’s the kind of “works today, explodes later” coupling that makes systems rot.

- **Why this is bad:** IDs are now shaped by persistence quirks, not domain meaning. If another component later expects event IDs to be stable across transports, or expects underscore preservation, you’ve just created a silent incompatibility.
- **Concrete improvement:** introduce a dedicated `summary_event_id_for(condensation_id: EventID) -> EventID` helper (or better: a structured ID like `CondensationSummaryEventID(condensation_id=...)`) and keep normalization/persistence rules in one place. If the regex requires hyphens, *encode that rule centrally*.

### 2) `summary_offset` semantics are still underspecified and will bite you with multiple condensations
[openhands-sdk/openhands/sdk/context/view.py, ~429-457] `apply_condensation()`

You apply condensations sequentially, which is good, but **what does `summary_offset` mean**?
- offset into the *original conversation*?
- offset into the *post-filtered view events* at time of condensation?
- offset into the list *after forgetting but before inserting*?

Right now it’s “whatever list you happen to have at that moment”. That’s fine internally, but then you also want condensations to be referential and stable over time. Those goals fight each other.

- **Failure mode:** a later condensation generated with one interpretation of offsets will be replayed under another interpretation (because the view-building algorithm changed again), and your summaries drift.
- **Concrete improvement:** document it as an invariant in the `Condensation` event model and add assertions/tests that enforce it. If `summary_offset` is “offset in the view after removing forgotten events from the then-current view”, say that explicitly.

### 3) Batch atomicity now depends on having *all events* available, but the API doesn’t enforce that
[openhands-sdk/openhands/sdk/context/view.py, ~265+] `_enforce_batch_atomicity(view_events, all_events)`

You changed the signature to require `all_events` (good), but nothing stops a caller from passing a subset and getting garbage behavior. The function name doesn’t communicate that it’s doing a join against global batch metadata.

- **Concrete improvement:** either:
  - move batch construction higher so `View.from_events()` owns it (and pass a precomputed batch map), or
  - rename params to something more honest like `source_events` / `view_events`, and assert that `view_events` is a subsequence of `source_events` IDs.

---

## [IMPROVEMENT OPPORTUNITIES] (good taste / maintainability)

### 4) You deleted “specialized functions” and replaced them with “specialized inline conventions”
This PR claims to remove brittle helper logic, but the brittleness didn’t disappear — it moved:
- “summary IDs are synthetic but deterministic”
- “condensation requests are unhandled if they appear after the last condensation”
- “summary insertion uses offsets into a mutable stream”
- “some events aren’t LLMConvertible, silently skipped”

This is still specialized behavior; it’s just *less discoverable* now.

- **Concrete improvement:** formalize the view-building pipeline as named steps with explicit contracts:
  1) strip requests  
  2) stream condensations  
  3) apply forgetting  
  4) insert summaries  
  5) enforce atomicity  
  6) enforce tool-call pairing  
  Then unit-test those steps independently.

### 5) `unhandled_condensation_request_exists()` is linear scan every time — and computed after already scanning
[openhands-sdk/openhands/sdk/context/view.py, ~460+]  

You already iterate through `events` once in `from_events()`, then you reverse-scan again. Not a performance crisis, but it’s sloppy and reads like “I couldn’t be bothered to track a boolean”.

- **Concrete improvement:** compute it during the first pass (or track “seen_condensation” and “seen_request_after_last_condensation”).

### 6) `_filter_unmatched_tool_calls()` docstring lies (and the signature is suspicious)
[openhands-sdk/openhands/sdk/context/view.py, ~319+]  
It says `all_events` is “unused for now” but then you *do* use it for batch atomicity via `ActionBatch.from_events(all_events)`.

That mismatch makes it harder to reason about correctness.

- **Concrete improvement:** fix the docstring and rename `batch` vs `action_batch` consistently (you currently use both concepts across edits, which reads like half-refactor).

### 7) Condenser prompt change removes an explicit “previous summary” channel — hope you like regression roulette
[openhands-sdk/openhands/sdk/context/condenser/prompts/summarizing_prompt.j2]  
You dropped `<PREVIOUS SUMMARY>` and now rely on summaries being present in the events list. That might be correct structurally, but it changes the prompt semantics:
- models previously got “previous summary” as a privileged signal
- now it’s just another event in the stream, competing with noise

Your non-blocking condenser tests passed (good), but this is exactly the kind of change that passes a tiny suite and then degrades quality in real conversations.

- **Concrete improvement:** keep the explicit previous summary section but derive it from “most recent summary event” (or “all summaries in view”, but clearly separated). Multi-summary doesn’t mean “throw away the one thing that guided the model”.

---

## [STYLE NOTES] (minor, but still worth fixing)

### 8) Variable name `_new_summary_event` screams “I fought the linter”
[openhands-sdk/openhands/sdk/context/view.py, ~452-456]  
Leading underscore in a local is weird and communicates nothing.

### 9) Silent skipping of non-LLMConvertible events is risky
[openhands-sdk/openhands/sdk/context/view.py, ~495+]  
Logging at debug means most users won’t see “we dropped events on the floor”. If that’s intended, fine — but then codify it as policy, not as a debug log.

---

## VERDICT
✅ **Worth merging** *if* you tighten the semantics around `summary_offset` and stop baking persistence quirks into event identity.  
Right now it works, but it’s the kind of “works” that future you will pay interest on.

**Key insight:** you fixed “Views can only represent one summary” by turning summaries into *stateful stream mutations* — that’s fine, but only if the mutation semantics (IDs, offsets, replay guarantees) are defined like an API, not implied by whatever `from_events()` happens to do this month.

csmith49:
> ## [CRITICAL ISSUES] (must fix / at least must justify loudly)
> 1) Deterministic summary IDs are a hack glued to an unrelated constraint
> ...
> 2) `summary_offset` semantics are still underspecified and will bite you with multiple condensations
> ...

The summary offset semantics haven't changed, but documentation could be more verbose. Addressed these two points by moving the condensation semantics to the `Condensation` event instead of the `View`.

> 3) Batch atomicity now depends on having _all events_ available, but the API doesn’t enforce that
> ...

Semantics of batch atomicity hasn't changed, just the way the code is structured. The API is a bit messy but that's what [this PR](https://github.com/OpenHands/software-agent-sdk/pull/1649) is for.

> ## [IMPROVEMENT OPPORTUNITIES] (good taste / maintainability)
> 4) You deleted “specialized functions” and replaced them with “specialized inline conventions”
> ...
> 5) `unhandled_condensation_request_exists()` is linear scan every time — and computed after already scanning
> [openhands-sdk/openhands/sdk/context/view.py, ~460+]
> ...

The suggested fixes here are incompatible. The pipeline requires multiple linear passes if you actually want to separate the steps in code. We'd like a cleaner pipeline (and this PR improves the structure dramatically) but larger changes are beyond the scope of this PR.

> 6) `_filter_unmatched_tool_calls()` docstring lies (and the signature is suspicious)
> [openhands-sdk/openhands/sdk/context/view.py, ~319+] It says `all_events` is “unused for now” but then you _do_ use it for batch atomicity via `ActionBatch.from_events(all_events)`.
> ...

Addressed.

> 7) Condenser prompt change removes an explicit “previous summary” channel — hope you like regression roulette
> ...

Regressions are a concern, for sure, but the changes suggested here aren't a fix because they ignore the _positional_ nature of summaries. Anecdotally I haven't noticed any performance degradation.

> ## [STYLE NOTES] (minor, but still worth fixing)
> 8) Variable name `_new_summary_event` screams “I fought the linter”
> ...

Says the model that wrote that line of code! (Also removed when addressing other issues).

> 9) Silent skipping of non-LLMConvertible events is risky
> ...

This is standard and expected behavior (which is why the log level is `DEBUG`).

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -291,96 +264,112 @@ def __getitem__(
...
+        all ActionEvents in that batch are removed.
 
         This prevents partial batches from being sent to the LLM, which can cause
         API errors when thinking blocks are separated from their tool calls.
 
         Args:
-            events: The original list of events
-            removed_event_ids: Set of event IDs that are being removed
+            view_events: The list of events that are being kept in the view
+            all_events: The complete original list of all events
```
enyst, line 278: Sorry, is this complete original list ~240 events when it's 240 max size, or is it full history? Full history could be  ~ thousands 😅

openhands-sdk/openhands/sdk/context/view.py
```diff
@@ -291,96 +264,112 @@ def __getitem__(
...
+                ids_to_remove.update(batch_event_ids)
                 logger.debug(
                     f"Enforcing batch atomicity: removing entire batch "
                     f"with llm_response_id={llm_response_id} "
                     f"({len(batch_event_ids)} events)"
                 )
 
-        return updated_removed_ids
+        # Filter out events that need to be removed
+        return [event for event in view_events if event.id not in ids_to_remove]
```
enyst, line 308: (Not a review comment, sorry just pondering) I wonder if we remove more than we really _need_ to remove 🤔 

Maybe because of Anthropic quirks with extended thinking... I think, special blocks / signatures aside, we could have paired action / obs (not the entire batch), but I'm just not sure about all cases. It's probably safer as we do it here. 

Please correct me if wrong though: the batches are batches with the same `llm_response_id`, so buckets of parallel tool calls, not other criteria?

enyst:
@openhands Give me a second opinion here. Investigate deeply. We need this PR, so /codereview in depth. You can post your feedback directly on github.

enyst:
Gemini [went off the deep end](https://openhands-ai.slack.com/archives/C06PB3T5ZK6/p1769058830161239)!

FWIW, its actual review was:

**Style and Formatting**
[[openhands-sdk/openhands/sdk/context/view.py, Line 30]](https://github.com/OpenHands/software-agent-sdk/blob/feat/multi-summary-views/openhands-sdk/openhands/sdk/context/view.py#L30) :memo: Documentation: `ActionBatch` is a helper class but is defined at the top level. If it's only used within `View`, consider making it private (`_ActionBatch`) or moving it to a utils module to reduce the public API surface of `view.py`.

**Clarity and Readability**
[[openhands-sdk/openhands/sdk/context/view.py, Lines 477-478]](https://github.com/OpenHands/software-agent-sdk/blob/feat/multi-summary-views/openhands-sdk/openhands/sdk/context/view.py#L477-L478) :recycle: Redundancy: `_enforce_batch_atomicity` is called here, but it is also called inside `_filter_unmatched_tool_calls` (Line 347). Since `_filter_unmatched_tool_calls` performs the atomicity check against `all_events`, the explicit call in `from_events` is redundant. Removing it would avoid iterating over the events an extra time.

[[openhands-sdk/openhands/sdk/context/view.py, Lines 65-186]](https://github.com/OpenHands/software-agent-sdk/blob/feat/multi-summary-views/openhands-sdk/openhands/sdk/context/view.py#L65-L186) :mag: Readability: `manipulation_indices` is a very long and complex method (over 120 lines). It handles batch grouping, tool loop detection, and index calculation. Consider refactoring the tool loop detection logic into a separate private method (e.g., `_identify_tool_loops`) to improve readability and testability.

[[openhands-sdk/openhands/sdk/context/view.py, Line 423]](https://github.com/OpenHands/software-agent-sdk/blob/feat/multi-summary-views/openhands-sdk/openhands/sdk/context/view.py#L423) :memo: Documentation: The `unhandled_condensation_request_exists` method iterates `reversed(events)`. It might be worth noting in the docstring that this optimization assumes condensation requests are likely to be recent, or that we only care about the *latest* state.

**Security and Common Bug Patterns**
[[openhands-sdk/openhands/sdk/event/condenser.py, Line 70]](https://github.com/OpenHands/software-agent-sdk/blob/feat/multi-summary-views/openhands-sdk/openhands/sdk/event/condenser.py#L70) :link: Coupling: The deterministic ID generation `f"{self.id}-summary"` couples the summary event ID to the condensation event ID string representation. While this ensures uniqueness per condensation, it relies on string formatting. Ensure that `self.id` is always safe for this concatenation (e.g. doesn't contain characters that might confuse downstream parsers if any).

openhands-sdk/openhands/sdk/context/view.py
csmith49: It's the full list. Which I think we need? We only focus on enforcing the property on the current view, but the batch detection algorithm I think needs to look through all the _other_ events to make sure the batches aren't split by condensations.

openhands-sdk/openhands/sdk/context/view.py
csmith49: I believe that's what batches mean here. And we probably need it because of the way we split parallel tool calls into multiple events -- if we drop any one of the batch action/obs pairs we can't recreate the parallel tool call event and Anthropic yells at us.

Or at least it was that way once upon a time, probably worth investigating again.

csmith49:
> I set off Gemini out of curiosity, to see what it will tell us. Up to you though.

It's not bad feedback. It does feel like it's ignoring the historical context of this code and the PRs stacked behind it, but 🤷 

I believe all of these concerns will be addressed in upcoming PRs, so I plan to leave it alone for now.

openhands-sdk/openhands/sdk/context/view.py
enyst: Maybe worth noting, I think we had problems before with the full history in memory. I must have seen multiple times questions, coming from working on the servers, like, do we really have to keep thousands of them in memory / load from disk when the agent is only using ~100? Also I thought EventLog avoids having them all in memory? 🤔

PR #1732, opened 2026-01-14
fix: move async operations outside sync lock in event subscription

neubig:
## Summary

This PR moves `await` operations outside the `FIFOLock` context in `subscribe_to_events()` and `_publish_state_update()` to prevent potential issues when mixing synchronous locks with async I/O operations.

## Problem

During investigation of conversation `ea1af5d3-0bbf-475a-a70d-9c93f9d113e1`, we observed a pod hang after logging "WebSocket connected" with no subsequent output. The pod was running v1.8.1 and went completely silent during WebSocket event subscription.

The current code acquires a synchronous `FIFOLock` and then performs async operations while holding that lock:

```python
with state:  # Acquires FIFOLock
    state_update_event = ConversationStateUpdateEvent.from_conversation_state(state)
    await subscriber(state_update_event)  # Async I/O while holding sync lock
```

This pattern can lead to issues because:
1. The `await` yields control to the event loop while holding a sync lock
2. If other coroutines or threads need this lock, contention occurs
3. The state serialization (`model_dump()`) for large states can be CPU-intensive

### Frontend Reconnection Behavior

The frontend WebSocket implementation ([`use-websocket.ts`](https://github.com/OpenHands/OpenHands/blob/main/frontend/src/hooks/use-websocket.ts#L107-L109)) has automatic reconnection enabled with a **3-second delay**. If the subscription hangs for more than 3 seconds due to lock contention, the frontend will attempt to reconnect, creating a new subscription request while the previous one is still blocked. This can compound the problem:

1. First subscription hangs while holding the lock
2. After 3 seconds, frontend reconnects and sends another subscription request
3. The new request also needs the lock, creating additional contention
4. This cycle can continue, potentially leading to resource exhaustion

## Solution

The state snapshot (`ConversationStateUpdateEvent`) is an **immutable copy** of the state data created by `model_dump()`. The `Event` base class has `frozen=True` in its `model_config`, making all events immutable after creation. Once created, the snapshot doesn't need continued lock protection. We can safely:

1. Acquire the lock
2. Create the immutable snapshot
3. Release the lock
4. Send the snapshot asynchronously

```python
with state:  # Acquire lock
    state_update_event = ConversationStateUpdateEvent.from_conversation_state(state)
# Lock released - snapshot is frozen (immutable), doesn't need protection

await subscriber(state_update_event)  # Send outside lock
```

## Changes

- `subscribe_to_events()`: Move `await subscriber(state_update_event)` outside the `with state:` block
- `_publish_state_update()`: Move `await self._pub_sub(state_update_event)` outside the `with state:` block
- Added explicit comments explaining the immutability guarantee from `Event.frozen=True`
- Added note about sequential subscriber notification in `_pub_sub` (potential future optimization with `asyncio.gather()`)

## Testing

- All existing tests pass
- Pre-commit checks pass
- Added `TestEventServiceConcurrentSubscriptions` test class with 4 new tests:
  - `test_concurrent_subscriptions_no_deadlock`: Verifies multiple concurrent subscriptions complete without hanging
  - `test_slow_subscriber_does_not_block_lock`: **Verifies lock is released before async subscriber call** (this test fails with the original code and passes with the fix)
  - `test_subscription_during_state_update`: Verifies subscriptions and state updates can interleave without deadlock
  - `test_multiple_state_updates_with_slow_subscribers`: Verifies rapid state updates work with slow subscribers

### DIFF
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -319,13 +319,15 @@ class EventService:
         # Send current state to the new subscriber immediately
         if self._conversation:
             state = self._conversation._state
-            # Create state snapshot while holding the lock to ensure consistency
+            # Create state snapshot while holding the lock to ensure consistency.
+            # ConversationStateUpdateEvent inherits from Event which has frozen=True
+            # in its model_config, making the snapshot immutable after creation.
             with state:
                 state_update_event = (
                     ConversationStateUpdateEvent.from_conversation_state(state)
                 )
 
-            # Send state update outside the lock - the event is an immutable snapshot,
+            # Send state update outside the lock - the event is frozen (immutable),
             # so we don't need to hold the lock during the async send operation.
             # This prevents potential deadlocks between the sync FIFOLock and async I/O.
             try:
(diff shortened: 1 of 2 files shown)


neubig:
@openhands all of the above are good suggestions, please fix

tofarr (approved):
General Principal Makes sense - do not hold the lock for longer than required.

PR #1800, opened 2026-01-23
refactor(message): move serialization control to to_chat_dict() parameters

neubig:
## Summary

Following up on PR #1798 (which changed the 5 serialization fields to default to `None`), this PR moves those fields out of the `Message` class entirely and makes them required parameters of `to_chat_dict()`.

### Message Class

**Removed 5 fields:**
- `cache_enabled`
- `vision_enabled`
- `function_calling_enabled`
- `force_string_serializer`
- `send_reasoning_content`

**Updated methods:**
- `to_chat_dict()` - now accepts all 5 as required keyword-only parameters
- `_list_serializer()` - now accepts `vision_enabled` parameter

### LLM Class

**Updated `format_messages_for_llm()`:**
```python
def format_messages_for_llm(self, messages: list[Message]) -> list[dict]:
    # ... caching logic ...
    
    model_features = get_features(self._model_name_for_capabilities())
    cache_enabled = self.is_caching_prompt_active()
    vision_enabled = self.vision_is_active()
    function_calling_enabled = self.native_tool_calling
    force_string_serializer = (
        self.force_string_serializer
        if self.force_string_serializer is not None
        else model_features.force_string_serializer
    )
    send_reasoning_content = model_features.send_reasoning_content

    formatted_messages = [
        message.to_chat_dict(
            cache_enabled=cache_enabled,
            vision_enabled=vision_enabled,
            function_calling_enabled=function_calling_enabled,
            force_string_serializer=force_string_serializer,
            send_reasoning_content=send_reasoning_content,
        )
        for message in messages
    ]
    return formatted_messages
```

## Benefits

1. **Cleaner separation of concerns**: Message defines its structure, LLM provides serialization context at call time
2. **No more None validation**: Parameters are required, so no runtime checks needed
3. **Simpler Message data model**: 5 fewer fields for storage/serialization
4. **More explicit tests**: Tests clearly show which serialization settings are being tested

## Testing

- All 2004 SDK tests pass
- Pre-commit hooks pass (pyright, ruff, etc.)

## Breaking Changes

This is a breaking change for any code that:
1. Calls `message.to_chat_dict()` directly (must now pass parameters)
2. Calls `message._list_serializer()` directly (must now pass `vision_enabled`)
3. Sets the removed fields on Message objects (will get validation errors)

Most users won't be affected since they use `LLM.format_messages_for_llm()` which handles this internally.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:6400a86-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-6400a86-python \
  ghcr.io/openhands/agent-server:6400a86-python
```

```
ghcr.io/openhands/agent-server:6400a86-golang-amd64
ghcr.io/openhands/agent-server:6400a86-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:6400a86-golang-arm64
ghcr.io/openhands/agent-server:6400a86-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:6400a86-java-amd64
ghcr.io/openhands/agent-server:6400a86-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:6400a86-java-arm64
ghcr.io/openhands/agent-server:6400a86-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:6400a86-python-amd64
ghcr.io/openhands/agent-server:6400a86-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:6400a86-python-arm64
ghcr.io/openhands/agent-server:6400a86-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:6400a86-golang
ghcr.io/openhands/agent-server:6400a86-java
ghcr.io/openhands/agent-server:6400a86-python
```

- Each variant tag (e.g., `6400a86-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `6400a86-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/message.py
+++ b/openhands-sdk/openhands/sdk/llm/message.py
@@ -11,10 +11,11 @@ from litellm.types.responses.main import (
 from litellm.types.utils import Message as LiteLLMMessage
 from openai.types.responses.response_output_message import ResponseOutputMessage
 from openai.types.responses.response_reasoning_item import ResponseReasoningItem
-from pydantic import BaseModel, ConfigDict, Field, field_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 from openhands.sdk.logger import get_logger
 from openhands.sdk.utils import DEFAULT_TEXT_CONTENT_LIMIT, maybe_truncate
+from openhands.sdk.utils.deprecation import warn_deprecated
 
 
 logger = get_logger(__name__)
@@ -229,6 +230,47 @@ class Message(BaseModel):
         default=None,
         description="OpenAI Responses reasoning item from model output",
     )
+
+    # Deprecated fields that were moved to to_chat_dict() parameters.
+    # These fields are ignored but accepted for backward compatibility.
+    # REMOVE_AT: 1.12.0 - Remove this list and the _handle_deprecated_fields validator
+    _DEPRECATED_FIELDS: ClassVar[tuple[str, ...]] = (
+        "cache_enabled",
+        "vision_enabled",
+        "function_calling_enabled",
+        "force_string_serializer",
+        "send_reasoning_content",
+    )
+
+    model_config = ConfigDict(extra="ignore")
+
+    @model_validator(mode="before")
+    @classmethod
+    def _handle_deprecated_fields(cls, data: Any) -> Any:
+        """Handle deprecated fields by emitting warnings and removing them.
+
+        REMOVE_AT: 1.12.0 - Remove this validator along with _DEPRECATED_FIELDS
+        """
+        if not isinstance(data, dict):
+            return data
+
+        deprecated_found = [f for f in cls._DEPRECATED_FIELDS if f in data]
+        for field in deprecated_found:
+            warn_deprecated(
+                f"Message.{field}",
+                deprecated_in="1.9.1",
+                removed_in="1.12.0",
+                details=(
+                    f"The '{field}' field has been removed from Message. "
+                    "Pass it as a parameter to to_chat_dict() instead, or use "
+                    "LLM.format_messages_for_llm() which handles this automatically."
+                ),
+                stacklevel=4,  # Adjust for validator call depth
+            )
+            # Remove the deprecated field so Pydantic doesn't complain
+            del data[field]
+
+        return data
 
     @property
     def contains_image(self) -> bool:
--- a/tests/sdk/llm/test_message.py
+++ b/tests/sdk/llm/test_message.py
@@ -392,3 +392,67 @@ def test_message_with_reasoning_content_list_serializer():
     assert isinstance(result["content"], list)
     assert result["content"][0]["text"] == "Final answer"
     assert result["reasoning_content"] == "Step by step reasoning"
+
+
+def test_message_deprecated_fields_emit_warnings():
+    """Test that deprecated fields emit deprecation warnings but don't fail."""
+    import warnings
+
+    from deprecation import DeprecatedWarning
+
+    from openhands.sdk.llm.message import Message
+
+    deprecated_fields = [
+        "cache_enabled",
+        "vision_enabled",
+        "function_calling_enabled",
+        "force_string_serializer",
+        "send_reasoning_content",
+    ]
+
+    # Test each deprecated field individually using model_validate with dict
+    for field in deprecated_fields:
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            Message.model_validate({"role": "user", "content": "test", field: True})
+            # Should have received a deprecation warning
+            deprecation_warnings = [
+                x for x in w if issubclass(x.category, DeprecatedWarning)
+            ]
+            assert len(deprecation_warnings) == 1, (
+                f"Expected 1 warning for {field}, got {len(deprecation_warnings)}"
+            )
+            assert field in str(deprecation_warnings[0].message)
+
+
+def test_message_deprecated_fields_are_ignored():
+    """Test that deprecated fields are ignored and don't affect the Message."""
+    import warnings
+
+    from openhands.sdk.llm.message import Message
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")  # Suppress warnings for this test
+        # Use model_validate to pass extra fields that pyright doesn't know about
+        message = Message.model_validate(
+            {
+                "role": "user",
+                "content": "test",
+                "cache_enabled": True,
+                "vision_enabled": True,
+                "function_calling_enabled": True,
+                "force_string_serializer": True,
+                "send_reasoning_content": True,
+            }
+        )
+
+    # The message should be created successfully
+    assert message.role == "user"
+    assert len(message.content) == 1
+
+    # The deprecated fields should not exist on the model
+    assert not hasattr(message, "cache_enabled")
+    assert not hasattr(message, "vision_enabled")
+    assert not hasattr(message, "function_calling_enabled")
+    assert not hasattr(message, "force_string_serializer")
+    assert not hasattr(message, "send_reasoning_content")


enyst:

Post your feedback directly on github as a comment to the PR. Note that it will be rendered as markdown.

Once you are done, answer me: WDYT about the change in this PR, it seems minor but do these fields "belong" in Message or are they better as on-the-fly parameters? Explain your thinking. Do it in the same github comment.

enyst:
## /codereview-roasted 🔥

**Taste Rating:** 🟡 **Acceptable** (directionally correct), with some unnecessary API footguns and a bit of “we moved complexity, didn’t delete it”.

This PR *looks* like a minor refactor, but it’s really an API surface redesign:
- you removed 5 fields from `Message`
- you made them **required** kwargs to `Message.to_chat_dict()`
- you pushed the responsibility for choosing them into `LLM.format_messages_for_llm()`

That’s sane *if* `Message` is a pure data container and serialization policy lives at the edge. But right now it’s halfway there.

---

### [CRITICAL ISSUES]

- **Breaking-change blast radius is larger than the PR admits.**
  - The PR description says “most users won’t be affected”. That’s optimistic.
  - Any downstream that ever called `to_chat_dict()` (and lots of people do, for logging / tracing / adapters / custom providers) is now forced to thread 5 flags everywhere.
  - Yes, you added a compatibility shim for *constructor fields* via `extra='ignore'` + `model_validator(mode="before")`, but you did **not** provide any compatibility for `to_chat_dict()` callers.
  - If you want “never break userspace”, you need at least **a deprecation period** where `to_chat_dict()` keeps defaults (or a `to_chat_dict_for_llm(llm)` helper), *or* you provide a single `serialization_context` object.

- [`openhands-sdk/openhands/sdk/llm/message.py`] **The deprecation shim is a little too clever and a bit sloppy.**
  - `model_config = ConfigDict(extra="ignore")` is a broad hammer: it silently ignores *any* extra keys, not just the 5 fields. That’s a debugging nightmare when someone misspells a real field.
  - You then compensate by manually warning only for the deprecated ones. Fine, but it still means typos won’t be caught.
  - If you must ignore extras, consider `extra="forbid"` + a targeted `model_validator` that strips only known deprecated keys (and errors on unknown ones). Right now you’re trading away validation for convenience.

---

- **The API is now “5 booleans of doom”.**
  - `to_chat_dict(cache_enabled=..., vision_enabled=..., function_calling_enabled=..., force_string_serializer=..., send_reasoning_content=...)`
  - This is tedious, easy to get wrong, and encourages copy/paste cargo culting.
  - A better shape is a single context object, e.g.:
    - `SerializationContext(cache_enabled=..., vision_enabled=..., ...)`
    - or even a `model_features`-like object passed through.
  - It also makes future flags cheaper to add without turning the method signature into a landfill.

- [`LLM.format_messages_for_llm()`] **Good move to compute feature flags once.**
  - This is a real improvement: you stopped mutating message objects as a side-effect.
  - But you still have policy (features + overrides + caching activation) scattered across `LLM` + `Message`.

- **The naming is lying a bit.**
  - `cache_enabled`, `vision_enabled`, etc. are not properties of a `Message`.
  - They’re properties of the *serialization target* (provider/model) and *the invocation*.
  - So the new approach is more honest.

- **Tests now duplicate the same opts dict in multiple files.**
  - You added `DEFAULT_SERIALIZATION_OPTS` in two test modules and another dict in fixture generation.
  - This is the classic “refactor created boilerplate”.
  - If this API stays, at least centralize a helper in tests (or provide a `Message.to_chat_dict_default()` for internal use).

---

### [STYLE NOTES]

- `stacklevel=4` in the validator is a magic number that will rot. It’s not terrible, but it’s also not stable as code moves.

---

## WDYT: do these fields “belong” in `Message`, or are they better as on-the-fly parameters?

They **do not belong** in `Message` *as persisted model fields*.

**Why:**
- They are not intrinsic properties of the message payload (“what was said”).
- They’re **policy knobs** for how you serialize **for a particular model/provider call**.
- Persisting them on the message couples stored conversation history to one specific transport format, which is exactly the kind of “data structure contamination” that makes systems brittle.

So: **moving them out of `Message` is correct.**

But making them **5 required kwargs on every `to_chat_dict()` call** is a UX trap:
- it forces policy plumbing into every call site
- it will lead to inconsistent combos (people will pass mismatched flags)
- it makes the “easy thing” (serialize a message) artificially hard

**My take:** keep the policy out of the `Message` data model, *but* don’t spray booleans everywhere.

Best compromise:
- Introduce a single `SerializationContext`/`LLMSerializationOptions` struct
- `LLM.format_messages_for_llm()` builds it once
- `Message.to_chat_dict(ctx=...)` consumes it
- Optionally keep a deprecated `to_chat_dict()` signature for one release cycle that defaults to “safe string serializer” to avoid breaking users.

**Verdict:** ✅ worth merging *if* you accept the API break and/or add a more stable abstraction; otherwise this is a “minor refactor” that quietly makes downstream code miserable.

enyst:
### Suggestion: introduce an `LLMSerializationContext` (aka “stop passing 5 booleans around”)

Right now `Message.to_chat_dict()` wants 5 keyword-only bools:

```py
message.to_chat_dict(
  cache_enabled=...,
  vision_enabled=...,
  function_calling_enabled=...,
  force_string_serializer=...,
  send_reasoning_content=...,
)
```

That’s workable internally, but it’s a leaky, high-friction public API. The natural abstraction is a *context object* that represents “how are we serializing for this model/provider/invocation?”.

#### What it should look like

```py

@dataclass(frozen=True, slots=True)
class LLMSerializationContext:
    # Transport/provider/model feature toggles (per call)
    cache_enabled: bool
    vision_enabled: bool
    function_calling_enabled: bool

    # Serialization quirks / compatibility knobs
    force_string_serializer: bool

    # Output shaping
    send_reasoning_content: bool

    @classmethod
    def from_llm(cls, llm: "LLM") -> "LLMSerializationContext":
        model_features = get_features(llm._model_name_for_capabilities())
        return cls(
            cache_enabled=llm.is_caching_prompt_active(),
            vision_enabled=llm.vision_is_active(),
            function_calling_enabled=llm.native_tool_calling,
            force_string_serializer=(
                llm.force_string_serializer
                if llm.force_string_serializer is not None
                else model_features.force_string_serializer
            ),
            send_reasoning_content=model_features.send_reasoning_content,
        )

    def should_use_list_serializer(self) -> bool:
        # Centralize the “policy” that currently lives in Message.to_chat_dict
        return (not self.force_string_serializer) and (
            self.cache_enabled or self.vision_enabled or self.function_calling_enabled
        )
```

Then `Message` becomes:

```py
def to_chat_dict(self, *, ctx: LLMSerializationContext) -> dict[str, Any]:
    if ctx.should_use_list_serializer():
        message_dict = self._list_serializer(vision_enabled=ctx.vision_enabled)
    else:
        message_dict = self._string_serializer()

    ... tool_calls/tool_call_id/name injection ...

    if ctx.send_reasoning_content and self.reasoning_content:
        message_dict["reasoning_content"] = self.reasoning_content

    return message_dict
```

And `LLM.format_messages_for_llm()` becomes:

```py
ctx = LLMSerializationContext.from_llm(self)
formatted = [m.to_chat_dict(ctx=ctx) for m in messages]
```

#### Why this is better

- **Single parameter instead of 5** → less boilerplate, less copy/paste config drift.
- **One place to encode policy** (`should_use_list_serializer`) instead of scattering it across LLM + Message.
- **Extensible without signature churn**: adding a new serialization knob doesn’t break every caller with a new required kwarg.
- **Better “userspace compatibility” story**: you can keep the old signature temporarily as:
  ```py
  def to_chat_dict(self, *, ctx: LLMSerializationContext | None = None, **legacy_kwargs):
      ...
  ```
  and emit a targeted deprecation warning when legacy kwargs are used.

#### Optional refinement: split “capabilities” vs “invocation toggles”

If you want to be more explicit:
- `ModelCapabilities` / `ProviderCapabilities` (static per model)
- `InvocationSerializationOptions` (dynamic per call)

…but for now a single `LLMSerializationContext` is the pragmatic “good taste” improvement.

#### One caution

If you keep `Message.model_config = extra="ignore"` for deprecated constructor keys, I’d strongly recommend **not** making it permanent; ideally strip only known deprecated keys and still error on unknown extras, otherwise typos quietly disappear.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -989,19 +989,27 @@ def format_messages_for_llm(self, messages: list[Message]) -> list[dict]:
...
+        )
+        send_reasoning_content = model_features.send_reasoning_content
+
+        formatted_messages = [
+            message.to_chat_dict(
+                cache_enabled=cache_enabled,
+                vision_enabled=vision_enabled,
+                function_calling_enabled=function_calling_enabled,
+                force_string_serializer=force_string_serializer,
+                send_reasoning_content=send_reasoning_content,
```
enyst, line 1009: (Not a review comment)
This one seems to confuse people, because it reads as "if not set, then it means we're not sending reasoning content" - which is not accurate for a number of well known LLMs (claude, gemini, openai)

Maybe sometime we could rename it, though I'm not sure to what... "special_provider_reasoning_needed_knob" 😅

neubig:
Thanks @enyst ! I agree with your comment, but it seems a slight bit orthogonal to this PR. With respect to the AI code review, I actually don't agree with it. I think the current design is preferable to introducing a complex struct, especially because the to_chat_dict function is only called in one place. Happy to discuss more if you agree though.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -989,19 +989,27 @@ def format_messages_for_llm(self, messages: list[Message]) -> list[dict]:
...
+        function_calling_enabled = self.native_tool_calling
+        force_string_serializer = (
+            self.force_string_serializer
+            if self.force_string_serializer is not None
+            else model_features.force_string_serializer
+        )
+        send_reasoning_content = model_features.send_reasoning_content
+
+        formatted_messages = [
+            message.to_chat_dict(
```
enyst, line 1004: It makes sense to me, and it reads well. I think maybe I would consider also the `LLMSerializationContext` bucket of things, which the agent suggested in a comment, even if that isn't much of a difference either. I guess as long as it takes them a bit out of the way (and encapsulating them might be a way to do that a little bit), the code might become more readable - they are serialization stuff. 

`function_calling_enabled` doesn't even do anything, anything at all over there: except is part of the heuristic to choose list/string. 😢 

But it's ~ the same thing at this point, not strong opinions.

enyst:
Sorry, I was still thinking, and trying to pass a few options through the agent.

Maybe you're right! Though I think it's also the fact that these options are recomputed every step, that is a bit strange. I'm not sure that's necessary. But I don't know what the best is, I'll try something and we'll see if we like it.

openhands-sdk/openhands/sdk/llm/message.py
```diff
@@ -249,6 +231,47 @@ class Message(BaseModel):
         description="OpenAI Responses reasoning item from model output",
     )
 
+    # Deprecated fields that were moved to to_chat_dict() parameters.
+    # These fields are ignored but accepted for backward compatibility.
```
enyst, line 235: This is a bit strange? I was just looking at a deprecation in another PR and I told the agent to take this PR as example. It seems a bit strange because if they are in use, then the application using them will have different behavior and potentially fail (with LLM providers).

I think the concern is low here, but it feels like deprecation might not actually fill its purpose? I think it's supposed to work as expected (while warning), until we remove them.

enyst: @neubig What do you think about something like this?

https://github.com/OpenHands/software-agent-sdk/pull/1803/files#diff-107251f5114a3fccff331a5e3d9be4aec09473e1a859dd2231da6f988cc76330R287

The field remains available, the API is the same, the code continues to work, for 5 more versions.

Otherwise I'm not sure deprecation is useful 🤔 , it can lead to strange unexpected behaviors if we keep formal compatibility, but not actual.

neubig: @openhands check if the previous behavior and current behavior actually change as a result of PR 1800. I think that maybe the answer is "no" because the serialization-related parameters are derived from the LLM regardless of what value is stored in Message, but I might be wrong.

neubig: So I think that since this doesn't change behavior it'd be OK to merge as-is, but thanks for the discussion!

enyst: If I understand correctly, we could:
- think of the python API as a public API, which client applications can use however they wish
- think of our uses cases as the public API, in which case maybe we look at `Message` class as an "internal" construction.

I am assuming the first is the case. I think maybe you're looking at the second, @neubig ?

PR #1807, opened 2026-01-23
Fix issue with remote docker workspaces.

hamiltop:
When using a remote docker environment, we need to connect to the docker host rather than localhost.

## Summary

I'm not sure why this was hardcoded to localhost when the original implementation used the app server. I'm also not sure if we need the utility function to replace localhost with the app server url. For reference, here's the original implementation:

https://github.com/OpenHands/OpenHands/blob/6c5ef256fd24f1108f919613f52d5f2426f8df5c/openhands/app_server/sandbox/docker_sandbox_service.py#L197C24-L203

## Checklist


hamiltop:
I would like to get this merged in. It's a pretty regression from the pre v1 openhands.

enyst:
Thank you, and sorry for the delay on this!

enyst:
@hamiltop I think the reviewer agent is a bit confused, but I sympathize that the code is a bit unclear, could we maybe document enough to help it understand the intention behind?

neubig:
[automated message] @xingyaoww assigned for review according to git blame

PR #1817, opened 2026-01-24
fix: handle PS1 metadata corruption in command output

neubig:
## Summary

This PR fixes a bug where PS1 metadata blocks get corrupted when programs output content concurrently with the shell's PS1 prompt rendering, causing evaluation failures.

## The Problem

When a program outputs to stdout while the shell is rendering the PS1 prompt, the outputs get **interleaved** due to terminal buffering. This creates nested `###PS1JSON###` markers:

```
###PS1JSON###
{corrupted JSON}
concurrent output here
###PS1JSON###
{valid JSON}
###PS1END###
```

The original non-greedy regex `###PS1JSON###(.*?)###PS1END###` matches from the FIRST marker to the ONLY end marker, capturing corrupted content that fails JSON parsing.

## The Fix

Changed the regex to use **negative lookahead**:

```python
# Before
r"###PS1JSON###(.*?)###PS1END###"

# After  
r"###PS1JSON###((?:(?!###PS1JSON###).)*?)###PS1END###"
```

This ensures the regex matches only content that does NOT contain another `###PS1JSON###` marker, effectively matching only the LAST valid block before each `###PS1END###`.

## Test Coverage

Added comprehensive tests in `test_ps1_corruption.py` covering:
- Nested marker handling
- Corrupted output recovery
- Pager output scenarios
- Multiple valid blocks

## Verification

```bash
uv run pytest tests/tools/terminal/test_ps1_corruption.py tests/tools/terminal/test_terminal_ps1_metadata.py -v
# 23 passed
```

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:8a1a337-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-8a1a337-python \
  ghcr.io/openhands/agent-server:8a1a337-python
```

```
ghcr.io/openhands/agent-server:8a1a337-golang-amd64
ghcr.io/openhands/agent-server:8a1a337-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:8a1a337-golang-arm64
ghcr.io/openhands/agent-server:8a1a337-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:8a1a337-java-amd64
ghcr.io/openhands/agent-server:8a1a337-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:8a1a337-java-arm64
ghcr.io/openhands/agent-server:8a1a337-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:8a1a337-python-amd64
ghcr.io/openhands/agent-server:8a1a337-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:8a1a337-python-arm64
ghcr.io/openhands/agent-server:8a1a337-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:8a1a337-golang
ghcr.io/openhands/agent-server:8a1a337-java
ghcr.io/openhands/agent-server:8a1a337-python
```

- Each variant tag (e.g., `8a1a337-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `8a1a337-python-amd64`) are also available if needed


neubig:
@openhands it is not very clear from the PR description what it means to have "PS1 blocks corrupted by ascii art". What is the full scope of things that can be corrupted and what is the underlying mechanism that caused this to happen? For evidence that this happened in the past, check previous PRs to OpenHands/openhands-index-results and download the results from SWE-Bench-Multimodal for claude-sonnet-4.5. The evidence is in one of those sets of results from last week (probably Thursday or Friday). Find more examples and also understand the causal mechanism, and update the PR description to be more precise.

xingyaoww:
@OpenHands /codereview-roasted can you checkout a new temporary branch from this PR (but do NOT push!), then try to remove all the implementations from this PR. Then tell me, can you reproduce the test failure mentioned in the PR description?

neubig:
@openhands read the critical comment from all-hands-bot and write a test to reproduce the "critical bug"

neubig:
@openhands check all the review threads, and if they have already been resolved in the code, mark them as resolved using the github API. You will need to continue doing this until CI passes. Further, there are mentions in the comments about "ascii art". it is not necessarily only ascii art that causes this issue, so make the description more general. also try to make sure that the changes are as concise as they can possibly be whil still fixing the core issue.

openhands-tools/openhands/tools/terminal/constants.py
xingyaoww, line 1: @OpenHands /codereview-roasted tell me, in this updated version of the code, what will the terminal output to the agent IF the output somehow interleaved with the PS1 metadata? Is there a better solution / more fundamental solution?

xingyaoww: the reason why we need PS1 metadata with terminal output is that we want a way to know that the current program finishes the execution. However, if we put it into a separate fd, it probably won't work like that. Tmux just give us these interleaved output - in this case, i felt this workaround at least give agent all the info it needs and is a net improvement

xingyaoww:
## Code Review Analysis

### What happens when output interleaves with PS1 metadata?

With this fix, when interleaving occurs:

1. **Metadata extraction works correctly** - The negative lookahead regex matches only the LAST valid `###PS1JSON###` block before `###PS1END###`, so we get correct metadata (exit code, cwd, etc.)

2. **However, the agent sees corrupted output** - The corrupted first PS1 block (including partial JSON and `###PS1JSON###` markers) appears in the command output shown to the agent:

```
###PS1JSON###
{
  "pid": "",
  "exit_code": "0",
  ...
 8   -_-_-_-_-_,------,
 0#PS-_-_-_-_-_|   /\_/\
...
  8 passing (6ms)

Done.
```

This is confusing for the agent - it sees PS1 markers and partial JSON as part of the command output.

### Is there a better/more fundamental solution?

**Yes.** I have created issue #1972 documenting a more fundamental approach:

**PROMPT_COMMAND + File-based Metadata:**
```bash
PROMPT_COMMAND='echo "{...}" > /tmp/.ps1_meta_$$'
PS1='###PS1READY###\n'
```

- `PROMPT_COMMAND` runs BEFORE PS1, writing metadata to a file (no interleaving possible)
- PS1 contains only a simple completion marker
- Metadata is NEVER corrupted - only the marker can be corrupted
- If marker is corrupted, we miss completion detection (safer failure mode vs wrong metadata)

### Verdict on this PR

This PR is a valid bandage fix - it correctly recovers metadata from corrupted output. However:

1. It does not prevent the corrupted content from appearing in agent output
2. A more fundamental solution (PROMPT_COMMAND + file) would be more robust

See #1972 for the detailed proposal.

PR #1823, opened 2026-01-26
Truncate terminal outputs before persisting events

enyst:
HUMAN: In V1, currently
(1)- we cap Terminal obs thanks to 10k lines limit
(2)- no max chars limit; I don't know if lines are limited
(3)- `to_llm_message` will cap with a max chars limit, the content actually used (sent to the LLM)

This PR proposes to address "layer 2" as the agent calls it below, capping per chars too. Reasons include: potentially high observations loaded in memory, transferred over the wire, loaded in the UIs even if not displayed in full.

Note: it seems the browser tools already cap to 30k chars, before creating the observation. Basically this PR proposes to make sure Terminal does too.

Related to https://github.com/OpenHands/software-agent-sdk/issues/1824

---

Port layer-2 truncation (truncate before store) for terminal tool results.

Background: OpenHands/OpenHands#7404 (V0) truncated CmdOutputObservation before saving into the event stream to avoid persisting multi-megabyte outputs the LLM never sees.

Proposed fix:
- TerminalSession now truncates command output to MAX_CMD_OUTPUT_SIZE (30k) before creating TerminalObservation (completed/nochange-timeout/hard-timeout + previous-command-running).
- Added a session-level test (runs on tmux + subprocess) asserting large command output is truncated in obs.text.

Local: pre-commit clean; terminal session + truncation tests pass.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:9b0ee17-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-9b0ee17-python \
  ghcr.io/openhands/agent-server:9b0ee17-python
```

```
ghcr.io/openhands/agent-server:9b0ee17-golang-amd64
ghcr.io/openhands/agent-server:9b0ee17-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:9b0ee17-golang-arm64
ghcr.io/openhands/agent-server:9b0ee17-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:9b0ee17-java-amd64
ghcr.io/openhands/agent-server:9b0ee17-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:9b0ee17-java-arm64
ghcr.io/openhands/agent-server:9b0ee17-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:9b0ee17-python-amd64
ghcr.io/openhands/agent-server:9b0ee17-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:9b0ee17-python-arm64
ghcr.io/openhands/agent-server:9b0ee17-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:9b0ee17-golang
ghcr.io/openhands/agent-server:9b0ee17-java
ghcr.io/openhands/agent-server:9b0ee17-python
```

- Each variant tag (e.g., `9b0ee17-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `9b0ee17-python-amd64`) are also available if needed


openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -100,6 +100,50 @@ class TerminalObservation(Observation):
         description="Directory where full output files are saved",
     )
 
+
```
enyst:
```suggestion
```

enyst:
(AGENT)

Refactor per discussion: moved “truncate before store” to the TerminalSession layer (where observations are created), instead of a TerminalObservation validator.

- TerminalSession now truncates command output to MAX_CMD_OUTPUT_SIZE (30k) before creating TerminalObservation (completed/nochange-timeout/hard-timeout + previous-command-running).
- Removed TerminalObservation model validator and the full_output_save_dir persistence logic.
- Added a session-level test (runs on tmux + subprocess) asserting large command output is truncated in obs.text.

Local: pre-commit clean; terminal session + truncation tests pass.

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -124,11 +124,13 @@ def to_llm_content(self) -> Sequence[TextContent | ImageContent]:
         if self.metadata.exit_code != -1:
             ret += f"\n[Command finished with exit code {self.metadata.exit_code}]"
 
-        # Use enhanced truncation with file saving if working directory is available
+        # Ensure the LLM payload remains bounded.
+        # The full output (if any) is already optionally persisted when we truncate
+        # for storage. Avoid saving again here.
```
enyst:
```suggestion
```

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -124,11 +124,11 @@ def to_llm_content(self) -> Sequence[TextContent | ImageContent]:
         if self.metadata.exit_code != -1:
             ret += f"\n[Command finished with exit code {self.metadata.exit_code}]"
 
-        # Use enhanced truncation with file saving if working directory is available
+        # Ensure the LLM payload remains bounded.
```
enyst:
```suggestion
        # Use enhanced truncation with file saving if working directory is available
```

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -128,7 +128,7 @@ def to_llm_content(self) -> Sequence[TextContent | ImageContent]:
         truncated_text = maybe_truncate(
             content=ret,
             truncate_after=MAX_CMD_OUTPUT_SIZE,
-            save_dir=self.full_output_save_dir,
+            save_dir=None,
```
enyst:
```suggestion
            save_dir=self.full_output_save_dir,
```

smolpaws:
Re: latest review note about metadata being lost. In this PR, TerminalSession truncates only the *command output* string (obs.text) *before* creating the TerminalObservation. The prefix/suffix (exit code, timeout info, etc.) live on CmdOutputMetadata (obs.metadata.prefix/suffix) and are not appended into command_output at this layer, so this truncation cannot drop those metadata messages.

The only place we concatenate prefix+text+suffix and then truncate is TerminalObservation.to_llm_content, which keeps head+tail; if we want to be extra safe, we can add a test (with a much smaller MAX via monkeypatch) asserting the suffix lines are preserved under truncation. (update: test added)

Also: since we’re not passing save_dir to maybe_truncate, we’re intentionally not persisting full outputs / doing dedup here.

enyst:
@openhands understand this PR and read all comments. Tell me : what was NOT fixed from the AI reviewer comments?

List them. Then take them one by one and address them: do you think it should it be fixed or not, and why. Respond as a comment on github on the PR, using smolpaws. Note that it will be rendered as markdown.

smolpaws:
smolpaws here.

I re-read the AI review (all-hands-bot) and compared it to the current PR state (after the refactor that moved truncation into `TerminalSession` and removed the `TerminalObservation` validator + `full_output_save_dir` logic).

## What AI review comments were **not** fixed?

1. **(Suggestion) Deduplicate the repeated truncation snippet** (4 call sites).
2. **(Suggestion) Broaden test coverage / add edge-case tests** beyond the basic “output is truncated”.

Everything else from that review is effectively addressed by the refactor:
- **Double file saving**: no longer applicable because the validator + `full_output_save_dir` saving was removed, and we’re not passing `save_dir` anywhere.
- **“Missing `save_dir` parameter”**: intentionally not implemented (and now structurally impossible in this PR since `full_output_save_dir` is gone).
- **“Metadata could be lost”**: addressed by adding a test around `to_llm_content` tail preservation.

---

## Should each remaining item be fixed?

### 1) Code duplication (4 identical `maybe_truncate(..., truncate_after=MAX_CMD_OUTPUT_SIZE)` blocks)
**Do we need to fix it?** *Probably not for this PR.*

**Why:**
- The duplicated code is very small and local.
- Introducing a helper (e.g. `_truncate_command_output`) slightly improves DRYness, but also creates another abstraction to thread through (and reviewers often disagree on where that helper should live: `TerminalSession`, a utility module, etc.).
- Since this PR’s main goal is correctness + preventing large persisted events, the duplication doesn’t materially affect risk/behavior.

**When it *would* be worth doing:** if we expect to tune truncation behavior (e.g., different limits per status, or adding optional save-to-disk), a helper would become more valuable.

### 2) Additional tests / edge cases
**Do we need to fix it?** *Not strictly required*, but *nice-to-have*.

**Why:**
- The PR already covers the critical behavior: session-level truncation before persisting + ensuring LLM-facing content keeps the suffix in the tail.
- Adding more tests (exactly-at-limit, just-over-limit, multi-line behavior, head+tail invariants) increases confidence, but also increases test runtime/maintenance (especially with multi-backend terminal types).

**What I’d add if we decide to extend tests:**
- `len(obs.text) == small_max` behavior when input is much larger (already roughly covered)
- `obs.text.startswith(...)` and `.endswith(...)` to explicitly assert head+tail preservation
- A multi-line generator (e.g. `printf 'A\n%.0s' {1..N}`) to ensure truncation isn’t doing anything surprising with newline boundaries

---

If you want, I can follow up with a small refactor adding `_truncate_command_output()` + 1-2 lightweight edge case tests, but I’m leaning toward keeping this PR minimal.

xingyaoww (approved):
Thanks Engel! LGTM

PR #1820, opened 2026-01-24
refactor: remove redundant shutdown reconcile call

xingyaoww:
## Summary

This PR removes the `reconcile()` call that was added in PR #1791 after run completion. Testing shows this call is redundant because the WebSocket subscription handshake and the initial `reconcile()` call in `__init__` are sufficient to ensure all events are captured.

Maybe this is unnecessary 🤔 We could remove it to be a bit conservative / preserve better performance on WS

## Changes

Removed the following code from `_wait_for_run_completion()`:

```python
# Reconcile events to ensure we have all events that may have
# been emitted during the final moments of the run. This handles
# the race condition where events are published after the client
# detects "finished" status but before WebSocket delivers them.
self._state.events.reconcile()
```

## Testing

All tests pass without this call:

- `test_events_not_lost_during_client_disconnection` - 5 consecutive runs ✅
- `test_remote_conversation_over_real_server` - 5 consecutive runs ✅
- All 66 remote conversation tests ✅
- All 1271 SDK tests ✅

## Rationale

The original PR #1791 added two `reconcile()` calls:
1. In `__init__` after `wait_until_ready()` - to catch events between initial REST sync and WebSocket subscription
2. In `_wait_for_run_completion` after detecting "finished" status - to catch events during shutdown

The second call appears to be unnecessary because:
- The WebSocket subscription is already established and receiving events
- The initial reconcile in `__init__` ensures the client starts with a complete view
- Events emitted during the run are delivered via WebSocket in real-time

Removing this call simplifies the code and reduces an unnecessary REST API call at the end of each run.

Related to #1791

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e0e29647fc884279a1e6363196bfaa9d)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
| python | amd64, arm64 | `nikolaik/python-nodejs:python3.12-nodejs22` | [Link](https://hub.docker.com/_/nikolaik/python-nodejs:python3.12-nodejs22) |
| golang | amd64, arm64 | `golang:1.21-bookworm` | [Link](https://hub.docker.com/_/golang:1.21-bookworm) |

[...]

PR #1832, opened 2026-01-26
fix: wait for WebSocket terminal status to prevent event loss

xingyaoww:
## Summary

This PR fixes the race condition where events emitted during the final moments of a run could be lost if the WebSocket did not deliver them before `run()` returned.

## Problem

The test `test_events_not_lost_during_client_disconnection` was failing intermittently in CI (observed in [PR #1829](https://github.com/OpenHands/software-agent-sdk/actions/runs/21364607784/job/61492749827?pr=1829#step:7:5709)) with the following error:

```
AssertionError: ActionEvent with finish tool not found in client events. 
REST API has 1 ActionEvent(s) but client has 0.
```

## Root Cause

The race condition occurs when:
1. Server emits events (ActionEvent, ObservationEvent)
2. Client polls and sees "finished" status
3. `run()` returns before WebSocket delivers those events

PR #1820 removed the `reconcile()` call that was supposed to catch these events, causing the flaky test failure.

## Solution

This PR implements a more fundamental fix by waiting for the WebSocket to deliver the terminal status event before returning from `run()`:

1. **Added `_run_complete_event`**: A `threading.Event` that gets signaled when WebSocket delivers a `ConversationStateUpdateEvent` with terminal `execution_status` (idle, finished, error, stuck)

2. **Added `is_terminal()` method**: Added to `ConversationExecutionStatus` enum to check if a status represents a terminal state

3. **Modified `_wait_for_run_completion()`**: Now waits for the WebSocket to deliver the terminal status event instead of returning immediately when polling detects completion. The polling is kept as a health check for ERROR/STUCK states that need immediate attention.

### Why this approach works:

- **Fixes the root cause**: By waiting for WebSocket to deliver the terminal status event, we ensure all events are delivered before `run()` returns (since events come through the same WebSocket channel)
- **No additional API calls**: When WebSocket delivers events on time (the common case), no additional API calls are needed
- **Simple and reliable**: The WebSocket terminal status event is guaranteed to arrive after all other events from the run

## Reliable Reproduction

Added `test_event_loss_repro.py` which reliably reproduces the race condition by injecting a 3s delay in the WebSocket callback for finish events.

## Testing

- All cross tests pass locally
- The reproduction test passes with the fix

## Related

- Fixes the flaky test failure in PR #1829
- Addresses the issue introduced by PR #1820

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:71f40eb-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-71f40eb-python \
  ghcr.io/openhands/agent-server:71f40eb-python
```

```
ghcr.io/openhands/agent-server:71f40eb-golang-amd64
ghcr.io/openhands/agent-server:71f40eb-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:71f40eb-golang-arm64
ghcr.io/openhands/agent-server:71f40eb-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:71f40eb-java-amd64
ghcr.io/openhands/agent-server:71f40eb-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:71f40eb-java-arm64
ghcr.io/openhands/agent-server:71f40eb-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:71f40eb-python-amd64
ghcr.io/openhands/agent-server:71f40eb-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:71f40eb-python-arm64
ghcr.io/openhands/agent-server:71f40eb-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:71f40eb-golang
ghcr.io/openhands/agent-server:71f40eb-java
ghcr.io/openhands/agent-server:71f40eb-python
```

- Each variant tag (e.g., `71f40eb-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `71f40eb-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -7,7 +7,7 @@ import time
 import uuid
 from collections.abc import Mapping
 from queue import Empty, Queue
-from typing import SupportsIndex, overload
+from typing import TYPE_CHECKING, SupportsIndex, overload
 from urllib.parse import urlparse
 
 import httpx
@@ -15,6 +15,10 @@ import websockets
 
 from openhands.sdk.agent.base import AgentBase
 from openhands.sdk.conversation.base import BaseConversation, ConversationStateProtocol
+
+
+if TYPE_CHECKING:
+    from openhands.sdk.tool.schema import Action, Observation
 from openhands.sdk.conversation.conversation_stats import ConversationStats
 from openhands.sdk.conversation.events_list_base import EventsListBase
 from openhands.sdk.conversation.exceptions import (
@@ -1194,6 +1198,28 @@ class RemoteConversation(BaseConversation):
             HTTPError: If the server returns an error (e.g., no condenser configured).
         """
         _send_request(self._client, "POST", f"/api/conversations/{self._id}/condense")
+
+    def execute_tool(self, tool_name: str, action: "Action") -> "Observation":
+        """Execute a tool directly without going through the agent loop.
+
+        Note: This method is not yet supported for RemoteConversation.
+        Tool execution for remote conversations happens on the server side
+        during the normal agent loop.
+
+        Args:
+            tool_name: The name of the tool to execute
+            action: The action to pass to the tool executor
+
+        Raises:
+            NotImplementedError: Always, as this feature is not yet supported
+                for remote conversations.
+        """
+        raise NotImplementedError(
+            "execute_tool is not yet supported for RemoteConversation. "
+            "Tool execution for remote conversations happens on the server side "
+            "during the normal agent loop. Use LocalConversation for direct "
+            "tool execution."
+        )
 
     def close(self) -> None:
         """Close the conversation and clean up resources.
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -124,6 +124,11 @@ MODELS = {
         "display_name": "Qwen3 Coder Next",
         "llm_config": {"model": "litellm_proxy/openrouter/qwen/qwen3-coder-next"},
     },
+    "qwen3-coder-30b-a3b-instruct": {
+        "id": "qwen3-coder-30b-a3b-instruct",
+        "display_name": "Qwen3 Coder 30B A3B Instruct",
+        "llm_config": {"model": "litellm_proxy/Qwen3-Coder-30B-A3B-Instruct"},
+    },
 }
 
 
--- a/.github/workflows/assign-reviews.yml
+++ b/.github/workflows/assign-reviews.yml
@@ -24,7 +24,7 @@ jobs:
     run-task:
         # Only run scheduled jobs in the main repository, not in forks
         if: github.repository == 'OpenHands/software-agent-sdk' || github.event_name == 'workflow_dispatch'
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         env:
             # Configuration (modify these values as needed)
             AGENT_SCRIPT_URL: https://raw.githubusercontent.com/OpenHands/agent-sdk/main/examples/03_github_workflows/01_basic_action/agent_script.py
--- a/.github/workflows/condenser-runner.yml
+++ b/.github/workflows/condenser-runner.yml
@@ -46,7 +46,7 @@ jobs:
                 ) ||
                 github.event_name == 'workflow_dispatch'
             )
-        runs-on: blacksmith-4vcpu-ubuntu-2204
+        runs-on: ubuntu-22.04
         permissions:
             contents: read
             id-token: write
@@ -163,7 +163,7 @@ jobs:
                 ) ||
                 github.event_name == 'workflow_dispatch'
             )
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         permissions:
             contents: read
             pull-requests: write
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -10,7 +10,7 @@ on:
     workflow_dispatch:
 jobs:
     dispatch:
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         permissions:
             contents: write
         steps:
--- a/.github/workflows/deprecation-check.yml
+++ b/.github/workflows/deprecation-check.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
     check:
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         steps:
             - name: Checkout
               uses: actions/checkout@v5
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -66,7 +66,7 @@ jobs:
                 (github.event_name == 'schedule' && github.repository == 'OpenHands/software-agent-sdk')
             )
         needs: [post-initial-comment]
-        runs-on: blacksmith-4vcpu-ubuntu-2204
+        runs-on: ubuntu-22.04
         permissions:
             contents: read
             id-token: write
@@ -241,7 +241,7 @@ jobs:
                 github.event_name == 'workflow_dispatch' ||
                 (github.event_name == 'schedule' && github.repository == 'OpenHands/software-agent-sdk')
             )
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         permissions:
             contents: read
             pull-requests: write
--- a/.github/workflows/pr-review-by-openhands.yml
+++ b/.github/workflows/pr-review-by-openhands.yml
@@ -35,7 +35,7 @@ jobs:
         concurrency:
             group: pr-review-${{ github.event.pull_request.number }}
             cancel-in-progress: true
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         env:
             LLM_MODEL: litellm_proxy/claude-sonnet-4-5-20250929
             LLM_BASE_URL: https://llm-proxy.app.all-hands.dev
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
     pre-commit:
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
 
         steps:
             - name: Checkout code
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
     prepare-release:
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         steps:
             - name: Validate version format
               run: |
--- a/.github/workflows/pypi-release.yml
+++ b/.github/workflows/pypi-release.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
     publish:
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         outputs:
             version: ${{ steps.extract_version.outputs.version }}
         steps:
@@ -68,7 +68,7 @@ jobs:
 
     create-version-bump-prs:
         needs: publish
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         if: github.event_name == 'release'
         env:
             VERSION: ${{ needs.publish.outputs.version }}
--- a/.github/workflows/run-examples.yml
+++ b/.github/workflows/run-examples.yml
@@ -23,7 +23,7 @@ jobs:
         # Schedule trigger only runs in the main repository, not in forks
         if: github.event.label.name == 'test-examples' || github.event_name == 'workflow_dispatch' || (github.event_name == 'schedule' && 
             github.repository == 'OpenHands/software-agent-sdk')
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         timeout-minutes: 60
         steps:
             - name: Wait for agent server to finish build
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -116,7 +116,7 @@ jobs:
 
     check-openapi-schema:
         name: Check OpenAPI Schema
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
 
         steps:
             - name: Checkout PR branch
@@ -162,42 +162,44 @@ jobs:
                 # Each job specifies exactly what it builds and where it runs
                 include:
                     # Python variant
+                    # NOTE: Using Python 3.12 due to PyInstaller+libtmux compatibility
+                    # issue with Python 3.13. See issue #1886 for details.
                     - variant: python
                       arch: amd64
-                      base_image: nikolaik/python-nodejs:python3.13-nodejs22
-                      runner: blacksmith-8vcpu-ubuntu-2404
+                      base_image: nikolaik/python-nodejs:python3.12-nodejs22
+                      runner: ubuntu-24.04
                       platform: linux/amd64
 
                     - variant: python
                       arch: arm64
-                      base_image: nikolaik/python-nodejs:python3.13-nodejs22
-                      runner: blacksmith-8vcpu-ubuntu-2404-arm
+                      base_image: nikolaik/python-nodejs:python3.12-nodejs22
+                      runner: ubuntu-24.04-arm
                       platform: linux/arm64
 
                     # Java variant
                     - variant: java
                       arch: amd64
                       base_image: eclipse-temurin:17-jdk
-                      runner: blacksmith-8vcpu-ubuntu-2404
+                      runner: ubuntu-24.04
                       platform: linux/amd64
 
                     - variant: java
                       arch: arm64
                       base_image: eclipse-temurin:17-jdk
-                      runner: blacksmith-8vcpu-ubuntu-2404-arm
+                      runner: ubuntu-24.04-arm
                       platform: linux/arm64
 
                     # Golang variant
                     - variant: golang
                       arch: amd64
                       base_image: golang:1.21-bookworm
-                      runner: blacksmith-8vcpu-ubuntu-2404
+                      runner: ubuntu-24.04
                       platform: linux/amd64
 
                     - variant: golang
                       arch: arm64
                       base_image: golang:1.21-bookworm
-                      runner: blacksmith-8vcpu-ubuntu-2404-arm
+                      runner: ubuntu-24.04-arm
                       platform: linux/arm64
 
         runs-on: ${{ matrix.runner }}
@@ -210,7 +212,8 @@ jobs:
             ARCH: ${{ matrix.arch }}
             TARGET: binary
             PLATFORM: ${{ matrix.platform }}
-            GITHUB_SHA: ${{ github.sha }}
+            # Use PR head SHA for pull requests to match the image tag expected by run-examples.yml
+            GITHUB_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
             GITHUB_REF: ${{ github.ref }}
             CI: 'true'
 
@@ -224,8 +227,8 @@ jobs:
                   version: latest
                   python-version: '3.13'
 
-            - name: Set up Docker Buildx with Blacksmith
-              uses: useblacksmith/setup-docker-builder@v1
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v3
 
             - name: Log in to GHCR
               uses: docker/login-action@v3
@@ -268,9 +271,9 @@ jobs:
                   echo "Versioned tags: $VERSIONED_TAGS_CSV"
                   echo "===================="
 
-            - name: Build & Push with Blacksmith (${{ matrix.variant }}-${{ matrix.arch }})
+            - name: Build & Push (${{ matrix.variant }}-${{ matrix.arch }})
               id: build
-              uses: useblacksmith/build-push-action@v2
+              uses: docker/build-push-action@v6
               with:
                   context: ${{ steps.prep.outputs.build_context }}
                   file: ${{ steps.prep.outputs.dockerfile }}
@@ -278,6 +281,8 @@ jobs:
                   platforms: ${{ env.PLATFORM }}
                   push: true
                   tags: ${{ steps.prep.outputs.tags }}
+                  cache-from: type=gha
+                  cache-to: type=gha,mode=max
                   build-args: |
                       BASE_IMAGE=${{ env.BASE_IMAGE }}
 
@@ -329,7 +334,7 @@ jobs:
             github.event_name == 'push' ||
             (github.event_name == 'pull_request' &&
              !github.event.pull_request.head.repo.fork)
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         strategy:
             matrix:
                 variant: [python, java, golang]
@@ -352,8 +357,8 @@ jobs:
                   echo "short_sha=$SHORT_SHA" >> $GITHUB_OUTPUT
                   echo "Using SHORT_SHA: $SHORT_SHA"
 
-            - name: Set up Docker Buildx with Blacksmith
-              uses: useblacksmith/setup-docker-builder@v1
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v3
 
             - name: Log in to GHCR
               uses: docker/login-action@v3
@@ -370,7 +375,7 @@ jobs:
                   MANIFEST_TAG="${SHORT_SHA}-${VARIANT}"
 
                   # Create multi-arch manifest combining amd64 and arm64 using buildx imagetools
-                  # This properly handles manifest lists from Blacksmith builds
+                  # This properly handles manifest lists from Docker builds
                   echo "Creating multi-arch manifest: ${IMAGE}:${MANIFEST_TAG}"
                   docker buildx imagetools create -t ${IMAGE}:${MANIFEST_TAG} \
                     ${IMAGE}:${SHORT_SHA}-${VARIANT}-amd64 \
@@ -444,7 +449,7 @@ jobs:
         # Run if it's a PR and the matrix job completed (even if some variants failed)
         if: github.event_name == 'pull_request' && always() && (needs.build-and-push-image.result == 'success' || needs.build-and-push-image.result ==
             'failure')
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         outputs:
             build_summary: ${{ steps.consolidate.outputs.build_summary }}
         steps:
@@ -606,7 +611,7 @@ jobs:
         needs: consolidate-build-info
         # Only on PRs, and only if the consolidation succeeded
         if: github.event_name == 'pull_request' && needs.consolidate-build-info.result == 'success'
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         permissions:
             contents: read
             pull-requests: write
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -11,7 +11,7 @@ jobs:
     stale:
         # Only run scheduled jobs in the main repository, not in forks
         if: github.repository == 'OpenHands/software-agent-sdk'
-        runs-on: blacksmith-4vcpu-ubuntu-2204
+        runs-on: ubuntu-22.04
         steps:
             - uses: actions/stale@v10
               with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -140,7 +140,7 @@ jobs:
                   if-no-files-found: warn
 
     agent-server-tests:
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         steps:
             - name: Checkout
               uses: actions/checkout@v5
@@ -197,7 +197,7 @@ jobs:
                   if-no-files-found: warn
 
     cross-tests:
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         steps:
             - name: Checkout
               uses: actions/checkout@v5
@@ -254,7 +254,7 @@ jobs:
                   if-no-files-found: warn
 
     coverage-report:
-        runs-on: blacksmith-2vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         needs: [sdk-tests, tools-tests, agent-server-tests, cross-tests]
         if: always() && github.event_name == 'pull_request'
         steps:
--- a/.github/workflows/todo-management.yml
+++ b/.github/workflows/todo-management.yml
@@ -44,7 +44,7 @@ permissions:
 
 jobs:
     scan-todos:
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
     # Only run if triggered manually or if 'automatic-todo' label was added
         if: >
             github.event_name == 'workflow_dispatch' ||
@@ -120,7 +120,7 @@ jobs:
     process-todos:
         needs: scan-todos
         if: needs.scan-todos.outputs.todo-count > 0
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         strategy:
             matrix:
                 todo: ${{ fromJson(needs.scan-todos.outputs.todos) }}
@@ -297,7 +297,7 @@ jobs:
     summary:
         needs: [scan-todos, process-todos]
         if: always()
-        runs-on: blacksmith-4vcpu-ubuntu-2404
+        runs-on: ubuntu-24.04
         steps:
             - name: Generate Summary
               run: |
--- a/.openhands/skills/code-review.md
+++ b/.openhands/skills/code-review.md
@@ -5,7 +5,39 @@ triggers:
 - /codereview
 ---
 
-You are an expert code reviewer with a pragmatic, simplicity-focused approach. Provide clear, actionable feedback on code changes. Be direct but constructive.
+# OpenHands/software-agent-sdk Code Review Guidelines
+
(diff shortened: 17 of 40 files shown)


enyst:
@openhands Look at this github action failure: https://github.com/OpenHands/software-agent-sdk/actions/runs/21394941864/job/61590830061?pr=1840#step:7:5796

Analyze deeply why did test_events_not_lost_during_client_disconnection() fail there, not that the branch is not the same with this branch (that branch with the failure was openhands/bounded-init-state-scan).

Then check this PR's diff and tell me WDYT: does this PR fix the root cause of test_events_not_lost_during_client_disconnection() failures, as it intends to?

enyst:

We want a solution for the recent events that doesn’t require rethinking the ws/REST sync yet.

enyst:
### Comparison with PR #1858 (test-only fix)
I just cherry-picked the latest commit from PR #1840 into a new PR (#1858). That patch **only hardens the test** by:
- waiting longer for server shutdown + explicitly waiting for port release
- using `shutil.rmtree(..., ignore_errors=True)` to avoid teardown races
- retrying the REST events query (10x with short delay) to handle eventual consistency

No SDK behavior changes there — it’s strictly a *test stability* change.

### PR #1832 approach
This PR changes **RemoteConversation** behavior to avoid the event-loss race:
- waits for a terminal status *delivered via WebSocket* before `run()` returns
- falls back to polling if the WS signal is missed
- performs `reconcile_recent()` when polling wins to fetch late events
- adds a repro test with delayed WS callback

### Take / preference
If the underlying issue is **real event loss** (run() returns before WS delivers final events), **#1832 is the better fix** because it addresses the root cause and aligns client behavior with the REST API’s authoritative state.

The #1858-style fix mainly treats **CI flakiness / eventual consistency** — it reduces false negatives but could mask a genuine race in production. If the delayed-WS repro is valid, I’d favor the behavioral fix in #1832 and keep the test robust as a secondary guard.

**Minor nit:** `reconcile_recent()` uses `timestamp__gte`, so duplicates are expected; you already dedupe by `event.id`, which is good. Just worth noting that timestamp resolution might be coarse, so the dedupe path is essential.

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -910,21 +1011,43 @@ def _wait_for_run_completion(
...
                 status = self._poll_status_once()
             except Exception as exc:
                 self._handle_poll_exception(exc)
             else:
                 if self._handle_conversation_status(status):
+                    # Polling detected terminal status before WebSocket delivered it.
+                    # This is the race condition we're trying to avoid.
+                    # Use reconcile_recent() to fetch any events that may have
+                    # been emitted but not yet delivered via WebSocket.
+                    self._state.events.reconcile_recent()
```
xingyaoww: @OpenHands review diff against main, can we ONLY Keep the `_run_complete_event` and remove this `reconcile_recent` as it doesn't seems to be necessary?

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -706,8 +708,21 @@ def __init__(
...
 
+        # Add a callback that signals when run completes via WebSocket
+        # This ensures we wait for all events to be delivered before run() returns
+        def run_complete_callback(event: Event) -> None:
+            if isinstance(event, ConversationStateUpdateEvent):
+                if event.key == "execution_status" and event.value in (
+                    ConversationExecutionStatus.IDLE.value,
+                    ConversationExecutionStatus.FINISHED.value,
+                    ConversationExecutionStatus.ERROR.value,
+                    ConversationExecutionStatus.STUCK.value,
```
xingyaoww: @OpenHands do we have a list that defines these as a TERMINAL_STATUS=[] ?

Can you check for all usage of `ConversationExecutionStatus` and see if they are all very similar? If so, we should consider add a class method for `ConversationExecutionStatus` to check if a Status is a terminal state. This should help us avoid code like this in the future

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -910,20 +936,33 @@ def _wait_for_run_completion(
...
-                        elapsed,
-                    )
-                    return
-
-            time.sleep(poll_interval)
+                # _handle_conversation_status raises exceptions for ERROR/STUCK
+                # and returns True for terminal states (IDLE/FINISHED)
+                # We don't return here - we wait for WebSocket to deliver
+                # the terminal status event to ensure all events are received
+                self._handle_conversation_status(status)
```
xingyaoww, line 990: @OpenHands does _handle_conversation_status actually raise exception?

Can we increase the timeout for `self._run_complete_event.wait(timeout=poll_interval):` to be 5*poll_interval?

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -910,20 +936,38 @@ def _wait_for_run_completion(
                     ),
                 )
 
+            # Wait for either:
+            # 1. WebSocket delivers terminal status event (preferred)
+            # 2. Poll interval expires (fallback - check status via REST)
+            if self._run_complete_event.wait(timeout=5 * poll_interval):
```
xingyaoww:
```suggestion
            if self._run_complete_event.wait(timeout=poll_interval):
```

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -910,20 +936,38 @@ def _wait_for_run_completion(
...
+            # detects a terminal status (IDLE/FINISHED). Returning early would
+            # cause event loss because the WebSocket may not have delivered all
+            # events yet. Instead, we continue looping until the WebSocket
+            # delivers the terminal status event, which guarantees all events
+            # have been received through the same channel.
             try:
                 status = self._poll_status_once()
             except Exception as exc:
                 self._handle_poll_exception(exc)
             else:
```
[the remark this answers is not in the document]
xingyaoww: @OpenHands please review the diff against main and update the pr title and description so it macthes the implementation. We do not want to implement this reconcile_recent

tests/cross/test_event_loss_repro.py
```diff
@@ -0,0 +1,230 @@
...
+    ws_delay_s = 3.0
+    assert conv._ws_client is not None
+    orig_cb = conv._ws_client.callback
+
+    def delayed_cb(event: Event) -> None:
+        if (
+            isinstance(event, (ActionEvent, ObservationEvent))
+            and getattr(event, "tool_name", None) == "finish"
+        ):
+            time.sleep(ws_delay_s)
```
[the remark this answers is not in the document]
xingyaoww: already revert the wait timeout to 1 sec, so this shouldn't be an issue any more

enyst:

enyst:
@openhands look at CI, it’s failing. You can use remote gh. You should understand the problem the PR is addressing very well, investigate all you need for that. Then fix it please.

smolpaws:
## 🔴 /codereview-roasted: Needs improvement

**Taste Rating: 🔴 Needs improvement** — this fixes the symptom, but now correctness depends on one WS event arriving *at the right time*. That’s a new race, not a cure.

---

### **[CRITICAL ISSUES]** (Must fix)

**[openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py, Lines 925–970]**  
**Liveness regression**: `_run_complete_event.clear()` + waiting only on WS means you can miss a terminal event that already arrived between `run()` and `_wait_for_run_completion()`. Result: `run()` can hang until timeout even though the run finished.  
**Fix**: guard with a run generation, or check current status before clearing, or set the event if status is already terminal.

**[remote_conversation.py, Lines 939–970]**  
**WS-only dependency**: polling detects terminal states but you explicitly ignore them. If WS is delayed/disconnected, `run()` hangs even though REST says “finished”. That’s a behavioral regression.  
**Fix**: treat polling as a fallback exit (after N retries or when WS is unhealthy), not just as an error detector.

---

### **[IMPROVEMENT OPPORTUNITIES]** (Should fix)

**[tests/cross/test_event_loss_repro.py, Lines 36–86]**  
**Teardown race**: 2s join without confirmation the server actually stopped. That’s the same kind of flakiness you’re trying to avoid.  
**Fix**: wait for port release or an explicit shutdown acknowledgement.

**[tests/cross/test_event_loss_repro.py, Lines 166–181]**  
**Slow test**: hard-coded 3s delay per run. You can make it deterministic without burning 3 seconds.  
**Fix**: use shorter delay + retry/loop, or a barrier-based sync.

---

### **VERDICT: ❌ Needs rework**
You *did* fix the original race, but you replaced it with a WS liveness dependency. The minimal fix is a bounded WS wait with a REST fallback, not WS-or-timeout.

**Key insight:** the channel is unreliable; correctness can’t depend on it. Add a fallback exit when REST says terminal and WS hasn’t fired in a reasonable window.

smolpaws:
HUMAN NOTE: this is only about [this recent commit](https://github.com/OpenHands/software-agent-sdk/pull/1832/commits/5acf512d389f2e667bafb3896a625bfe1f897a1b) which fixed CI failures.

#### 🔴 /codereview-roasted (follow-up): this commit is **test-bandage**, not a fix

I pulled the latest commit (`5acf512d`), and it only changes **tests**. It does *not* address the hang/race in production code.

### What the commit actually does
- Adds `wait_until_ready.return_value = True` everywhere (fine, just to satisfy new WS handshake).
- In the run tests, it **mocks `_run_complete_event.wait()` to return True**, or swaps in a pre-set event.

### Why this is not a real fix
- It **masks the liveness problem** instead of solving it. Tests now skip the exact behavior that’s hanging CI.
- If the WS terminal event arrives early or never arrives, the real code can still hang. The tests no longer exercise that path.

### Verdict
**This commit is a test adaptation, not a fix.** It keeps CI green by bypassing the hang, but it doesn’t solve the underlying WS-only dependency. At best it hides the issue; at worst it makes us think the bug is gone.

**Bottom line:** good for unblocking tests, bad if treated as a solution. You still need a real fallback (REST terminal status or bounded wait) in the runtime code.

xingyaoww:
@OpenHands please review this comment https://github.com/OpenHands/software-agent-sdk/pull/1832#issuecomment-3826120464

Can you address the critical issue for me with the most appropriate/simplist solution you can come up? I want to avoid using the REST API to match the full data as much as possible as it is an expensive operation

xingyaoww:
@OpenHands please fix the failing actions on PR #1832 at branch `fix/restore-post-run-reconcile` and merge from main and resolve conflicts

xingyaoww:

xingyaoww:
@OpenHands please address these issues: https://github.com/OpenHands/software-agent-sdk/pull/1832#issuecomment-3836205812

enyst:
@openhands Do a /codereview-roasted on this PR. Investigate deeply the WS implementation and understand the synchronization problems it has, and this PR is trying to solve with some approach. WDYT?

Answer directly as a comment to the PR. Note that it will be rendered as markdown.

smolpaws:
## 🔴 /codereview-roasted: Needs rework

**Taste Rating:** 🔴 **Needs improvement** – fixes a real race, but adds new correctness gaps and still relies on ambiguous WS semantics.

---

### **[CRITICAL ISSUES]** (must fix)

**[openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py ~L706–L949] :boom: Silent ERROR/STUCK regression**
- `run_complete_callback` sets `_run_complete_event` for *any* terminal status, and `_wait_for_run_completion()` returns immediately on that flag without calling `_handle_conversation_status()`.
- Result: if the server ends in **ERROR** or **STUCK**, `run()` now **returns success** instead of raising `ConversationRunError`.
- Fix: capture the terminal status value from the WS event (store it), and always route it through `_handle_conversation_status()` before returning.

**[remote_conversation.py ~L706–L915] :bomb: No run correlation → stale WS can short‑circuit a new run**
- You clear `_run_complete_event` before POST `/run`, but if a **stale terminal `execution_status`** event (from a previous run or delayed WS delivery) arrives after the clear, `run()` will return immediately **even though the new run hasn’t finished**.
- There is **no run id / timestamp gating**. You need *some* monotonic marker: record run start time and only accept status events with `event.timestamp >= run_start`, or introduce a run counter in server state.

**[remote_conversation.py ~L920–L993] :warning: Hang risk when status flips to IDLE**
- The loop only exits on WS terminal status or **3 terminal REST polls**. If a new user message arrives concurrently, status becomes **IDLE** (non‑terminal) and your loop **never exits**, even though the run already ended.
- This is a correctness regression for multi‑threaded clients. If you insist on waiting for WS, you still need a **secondary completion condition** that can detect “run finished and no run in progress” rather than purely terminal state.

---

### **[IMPROVEMENT OPPORTUNITIES]** (should fix)

**[remote_conversation.py ~L706–L740] :hammer_and_wrench: `run()` now waits on user callbacks**
- `_run_complete_event` is set only after the composed callback chain finishes. A slow user callback now makes `run()` block, which is a surprising coupling.
- If the intent is “wait for WS delivery”, you should separate “WS receipt” from “user callback completion”. Otherwise, document this behavior loudly.

**[tests/cross/test_event_loss_repro.py] :sweat_smile: heavy integration test**
- Spawns a real server + sleeps 3s per run. Good repro, but this will make `cross` slow. Consider marking it slow or gating with a marker so CI doesn’t pay this cost everywhere.

---

### **VERDICT**
❌ **Needs rework** – directionally correct (sync on WS), but the current implementation **silently masks ERROR/STUCK**, **can return early on stale events**, and **can hang when status flips to IDLE**. These are user‑visible correctness regressions.

**Key insight:** Waiting on “terminal status via WS” is fine **only if you tie it to the *current run*** and still propagate error semantics; right now it’s just a global flag.

— smolpaws

xingyaoww:
@OpenHands please fix https://github.com/OpenHands/software-agent-sdk/pull/1832#issuecomment-3837769755

xingyaoww:
@OpenHands Do a /codereview-roasted on this PR. Investigate deeply the WS implementation and understand the synchronization problems it has, and this PR is trying to solve with some approach. WDYT?

Answer directly as a comment to the PR. Note that it will be rendered as markdown.

xingyaoww:
## 🔴 /codereview-roasted: Needs Improvement (but close to acceptable)

**Taste Rating: 🟡 Acceptable** - A pragmatic fix that addresses the root cause, but relies on implicit ordering guarantees and has some thread safety concerns.

---

### Deep Dive: The Synchronization Problem

The WebSocket implementation has a fundamental race condition between two parallel channels:

```
Server Timeline:
  ActionEvent(finish) → ObservationEvent(finish) → status=FINISHED

Client Timeline (before fix):
  REST poll sees FINISHED → run() returns → WebSocket still delivering events
```

The race window exists because REST polling can detect the terminal status before WebSocket delivers all events.

### Why This Fix Works

The key insight is that the **server guarantees event ordering**:

```python
# event_service.py, lines 531-543
async def _run_and_publish():
    try:
        await loop.run_in_executor(None, conversation.run)
    finally:
        # Wait for all pending events to be published
        if self._callback_wrapper:
            await loop.run_in_executor(
                None, self._callback_wrapper.wait_for_pending, 30.0
            )
        # THEN publish terminal status
        await self._publish_state_update()
```

Combined with:
1. **TCP ordering**: WebSocket messages arrive in order
2. **Sequential callback processing**: The client processes messages one at a time

So if we see the terminal status via WebSocket, all prior events have been processed. This is the correct fix.

---

### **[CRITICAL ISSUES]** (Should fix)

**[remote_conversation.py, Lines 555-558, 731-733] Thread Safety**: Shared state accessed from multiple threads without explicit synchronization.

```python
# Written from WebSocket thread:
self._terminal_status = event.value
self._terminal_status_run_counter = self._run_counter

# Read from main thread:
if self._terminal_status_run_counter == expected_run_counter:
    ws_status = self._terminal_status
```

While Python's GIL provides some protection, this is fragile. Consider using a `threading.Lock` or making these operations atomic via a single shared object.

**[remote_conversation.py, Line 951] Magic Number**: `TERMINAL_POLL_THRESHOLD = 3` is defined inline without clear justification.

```python
TERMINAL_POLL_THRESHOLD = 3  # Why 3? Document the reasoning
```

This adds 3 * poll_interval (default 3 seconds) of latency in the fallback path. Consider making this configurable or documenting why 3 is the right number.

---

### **[IMPROVEMENT OPPORTUNITIES]** (Should consider)

**[Implicit Contract] Server Ordering Guarantee**: The fix relies on the server sending events in order (all events before terminal status). This is currently true but not documented as a contract.

If someone changes `event_service.py` to publish the state update before `wait_for_pending()`, this fix breaks silently. Consider:
1. Adding a comment in `event_service.py` documenting this ordering guarantee
2. Or adding a server-side "run_complete" event that explicitly signals all events are delivered

**[test_event_loss_repro.py] Test Relies on Implementation Detail**: The test injects a delay in the callback, which works because WebSocket message processing is sequential. This is correct but subtle - the test should document why this works:

```python
# The delay works because WebSocket processes messages sequentially:
# 1. ActionEvent arrives → callback sleeps 3s
# 2. ConversationStateUpdateEvent waits in queue
# 3. Terminal status can't be processed until ActionEvent callback completes
```

**[state.py, is_terminal()] Good Design**: Excluding `IDLE` from terminal states is correct and well-documented. This prevents false positives when WebSocket delivers the initial state update during connection. 👍

---

### **[STYLE NOTES]** (Minor)

**[remote_conversation.py, Lines 719-735] Nested Callback Definition**: The `run_complete_callback` is defined inside `__init__`, which is fine but makes the class harder to test in isolation. Consider extracting to a method.

---

### **VERDICT:**
✅ **Worth merging with minor fixes**: The core logic is sound and addresses the root cause correctly. The thread safety concern is the main issue to address.

### **KEY INSIGHT:**
The fix correctly leverages the fact that WebSocket message processing is sequential - if we see the terminal status via WebSocket, all prior events have been processed. This is a fundamental property of the system that makes the fix reliable.

### **Comparison with PR #1858 (test-only fix)**:
PR #1858 hardens the test but doesn't fix the SDK. This PR (#1832) fixes the actual race condition in the SDK, which is the right approach. The test-only fix would leave users vulnerable to the same race condition in production.

xingyaoww:
@OpenHands please fix https://github.com/OpenHands/software-agent-sdk/pull/1832#issuecomment-3842280906

xingyaoww:
@OpenHands Do a /codereview-roasted on this PR. Investigate deeply the WS implementation and understand the synchronization problems it has, and this PR is trying to solve with some approach. WDYT?

Answer directly as a comment to the PR. Note that it will be rendered as markdown.

xingyaoww:
## 🔥 /codereview-roasted: Deep Dive into the WS Synchronization Problem

**Taste Rating: 🟡 Acceptable** - A pragmatic fix that addresses the immediate problem, though it adds complexity to paper over a fundamental dual-path architecture issue.

---

### The Core Problem (Well Understood)

The SDK has two parallel data paths that need synchronization:
1. **REST polling** → checks `execution_status` 
2. **WebSocket stream** → delivers events (ActionEvent, ObservationEvent, ConversationStateUpdateEvent)

The race condition:
```
Server: emit finish ActionEvent → emit ObservationEvent → set status=FINISHED → emit StateUpdateEvent
Client: [WS thread processing events...]  |  [REST poll sees FINISHED] → run() returns
                                          ↑
                                    Events still in flight!
```

### Why This PR's Approach Works

After digging into the server-side code (`event_service.py:525-543`), I found this critical guarantee:

```python
# Wait for all pending events to be published via
# AsyncCallbackWrapper before publishing the final state update.
if self._callback_wrapper:
    await loop.run_in_executor(None, self._callback_wrapper.wait_for_pending, 30.0)

# Clear task reference and publish state update
self._run_task = None
await self._publish_state_update()
```

**The server guarantees the terminal status event is sent AFTER all other events.** This makes the PR's approach of waiting for the WS terminal status fundamentally sound - if you receive the terminal status via WS, all events have already been sent through the same channel.

---

### **[CRITICAL ISSUES]** - None blocking merge

The previous review's concern about IDLE being a terminal status has been addressed - `is_terminal()` correctly excludes IDLE.

---

### **[IMPROVEMENT OPPORTUNITIES]**

**[remote_conversation.py, Lines 986-1012] Potential Tight Loop on Stale Events**
```python
if ws_status is not None:
    # ... handle and return
else:
    continue  # ← No sleep here
```
If there are multiple stale events from previous runs, this could spin without sleeping. In practice unlikely, but consider adding a small sleep or rate limit.

**[remote_conversation.py, Line 958] Magic Number**
```python
TERMINAL_POLL_THRESHOLD = 3
```
The comment explains the rationale well, but this could be a class constant or configurable parameter for different deployment scenarios (e.g., high-latency networks might want higher threshold).

**[test_event_loss_repro.py] Test Artificiality**
The test injects delay in the *callback processing*, not in network delivery. While it demonstrates the race condition, a more realistic test would simulate network latency. However, this is acceptable for a regression test.

---

### **[ARCHITECTURAL OBSERVATION]**

The complexity added here (5 new instance variables, run counter correlation, dual-path waiting logic) is a symptom of the fundamental tension between:
- **REST for reliability** (guaranteed delivery, easy retry)
- **WebSocket for real-time** (low latency, push-based)

The PR correctly identifies that the WS terminal status is the synchronization point, but the fallback mechanism (3 consecutive terminal polls + reconcile) adds significant complexity.

**Long-term consideration**: Could the SDK be simplified by making WS the primary path and REST purely for initial sync/fallback? The server already guarantees event ordering.

---

### **VERDICT: ✅ Worth Merging**

The core logic is sound:
1. ✅ Server guarantees terminal status comes after all events
2. ✅ Waiting for WS terminal status ensures event delivery
3. ✅ Fallback mechanism handles WS failures gracefully
4. ✅ Run counter prevents stale event issues
5. ✅ Thread safety with proper locking

The complexity is justified by the need to handle both happy path (WS works) and degraded path (WS fails).

---

### **KEY INSIGHT**

This PR correctly leverages the server's existing guarantee that terminal status is published *after* all events are flushed. The synchronization point was always there - this PR just makes the client wait for it properly.

xingyaoww:
@enyst 🤣 looks like after i repeat your prompt three times in a row, now openhands thinks the solution is ok

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -708,8 +718,28 @@ def __init__(
...
+                        status = ConversationExecutionStatus(event.value)
+                        if status.is_terminal():
+                            # Store the terminal status and run counter atomically.
+                            # This allows _wait_for_run_completion to:
+                            # 1. Check if status is for current run (not stale)
+                            # 2. Call _handle_conversation_status for ERROR/STUCK
+                            with self._terminal_status_lock:
+                                self._terminal_status = event.value
+                                self._terminal_status_run_counter = self._run_counter
+                            self._run_complete_event.set()
```
xingyaoww: @OpenHands Do a /codereview-roasted on this PR. Investigate deeply the WS implementation and understand the synchronization problems it has, and this PR is trying to solve with some approach. WDYT?

- Tell me, can we consolidate the `_terminal_status` and `_run_complete_event` into just ONE event?
- Also, can we simplify the current implementation?

Answer directly as a comment to the PR. Note that it will be rendered as markdown.

xingyaoww:
## 🔥 Roasted Code Review

### Taste Rating: 🟡 **Acceptable** - Works but could be cleaner

---

### The Good

This PR solves a **real problem** - the event loss race condition is a legitimate bug that causes flaky tests. The diagnosis is correct: when polling sees "finished" before WebSocket delivers events, `run()` returns too early.

The test reproduction (`test_event_loss_repro.py`) is excellent - injecting a 3s delay to reliably reproduce the race condition is the right way to validate the fix.

---

### The Not-So-Good: Synchronization Complexity

**5 pieces of state for one synchronization problem?** Let me count:

```python
_run_complete_event: threading.Event      # Signal
_run_counter: int                         # Counter
_terminal_status: str | None              # Payload
_terminal_status_run_counter: int         # Counter correlation
_terminal_status_lock: threading.Lock     # Protection
```

This is over-engineered. The classic "I need to pass data between threads" problem has simpler solutions.

---

### Answer to Your Questions

#### 1. Can we consolidate `_terminal_status` and `_run_complete_event` into ONE event?

**Yes, absolutely.** Use a `queue.Queue`:

```python
from queue import Queue, Empty

class RemoteConversation:
    _terminal_status_queue: Queue[str]  # Thread-safe, blocks on get()
    
    def __init__(self, ...):
        self._terminal_status_queue = Queue()
    
    # In callback:
    def run_complete_callback(event: Event) -> None:
        if isinstance(event, ConversationStateUpdateEvent):
            if event.key == "execution_status":
                status = ConversationExecutionStatus(event.value)
                if status.is_terminal():
                    self._terminal_status_queue.put(event.value)
    
    # In _wait_for_run_completion:
    def _wait_for_run_completion(self, poll_interval, timeout):
        while True:
            try:
                ws_status = self._terminal_status_queue.get(timeout=poll_interval)
                self._handle_conversation_status(ws_status)
                return
            except Empty:
                # Fallback: poll REST API
                ...
```

This eliminates:
- `_run_complete_event` (Queue.get() blocks)
- `_terminal_status_lock` (Queue is thread-safe)
- `_terminal_status` (it's in the queue)

#### 2. Can we simplify the run_counter mechanism?

**Yes.** Instead of tracking run counters, just drain the queue at the start of `run()`:

```python
def run(self, ...):
    # Drain any stale terminal status from previous runs
    while not self._terminal_status_queue.empty():
        try:
            self._terminal_status_queue.get_nowait()
        except Empty:
            break
    
    # ... trigger run ...
    self._wait_for_run_completion(...)
```

This eliminates:
- `_run_counter`
- `_terminal_status_run_counter`

---

### Simplified Implementation (5 variables → 1)

```python
from queue import Queue, Empty

class RemoteConversation:
    _terminal_status_queue: Queue[str]
    
    def __init__(self, ...):
        self._terminal_status_queue = Queue()
    
    def _setup_ws_callback(self):
        def run_complete_callback(event: Event) -> None:
            if isinstance(event, ConversationStateUpdateEvent):
                if event.key == "execution_status":
                    try:
                        status = ConversationExecutionStatus(event.value)
                        if status.is_terminal():
                            self._terminal_status_queue.put(event.value)
                    except ValueError:
                        pass
        return run_complete_callback
    
    def run(self, ...):
        # Drain stale events
        while True:
            try:
                self._terminal_status_queue.get_nowait()
            except Empty:
                break
        
        # Trigger run
        self._trigger_run()
        self._wait_for_run_completion(poll_interval, timeout)
    
    def _wait_for_run_completion(self, poll_interval, timeout):
        start_time = time.monotonic()
        consecutive_terminal_polls = 0
        TERMINAL_POLL_THRESHOLD = 3
        
        while True:
            elapsed = time.monotonic() - start_time
            if elapsed > timeout:
                raise ConversationRunError(...)
            
            try:
                # Wait for WS terminal status (preferred path)
                ws_status = self._terminal_status_queue.get(timeout=poll_interval)
                self._handle_conversation_status(ws_status)
                logger.info("Run completed via WebSocket (status: %s)", ws_status)
                return
            except Empty:
                # Fallback: poll REST API
                try:
                    status = self._poll_status_once()
                except Exception as exc:
                    self._handle_poll_exception(exc)
                    consecutive_terminal_polls = 0
                    continue
                
                self._handle_conversation_status(status)
                
                if status and ConversationExecutionStatus(status).is_terminal():
                    consecutive_terminal_polls += 1
                    if consecutive_terminal_polls >= TERMINAL_POLL_THRESHOLD:
                        logger.info("Run completed via REST fallback")
                        self._state.events.reconcile()
                        return
                else:
                    consecutive_terminal_polls = 0
```

---

### Other Observations

1. **`_handle_conversation_status` return value is now ignored** in the WS path (line 1001-1002). The method returns `bool` but you're not using it. Either use it or change the method signature.

2. **The fallback path is reasonable** - 3 consecutive terminal polls before returning is a good balance. But document why 3 (you did, good).

3. **The `is_terminal()` method is clean** - good addition to the enum.

---

### Verdict: ✅ **Worth merging with simplification**

The core logic is sound and solves a real problem. But the synchronization mechanism is more complex than necessary. Consider simplifying with a Queue before merging, or merge as-is and simplify in a follow-up.

**Key Insight:** When you need to pass data between threads with blocking semantics, `queue.Queue` is almost always simpler than `Event + shared variable + Lock + counter`.

xingyaoww:
@OpenHands please update this PR following the suggestions here: https://github.com/OpenHands/software-agent-sdk/pull/1832#issuecomment-3842474228

xingyaoww:
@OpenHands Do a /codereview-roasted on this PR. Investigate deeply the WS implementation and understand the synchronization problems it has, and this PR is trying to solve with some approach. WDYT?

Answer directly as a comment to the PR. Note that it will be rendered as markdown.

xingyaoww:
## 🔥 /codereview-roasted: Deep Dive into the WS Synchronization Problem

**Taste Rating: 🟡 Acceptable** - A pragmatic tactical fix that papers over a fundamental architectural issue. Works, but adds complexity.

---

### The Core Problem (Well Identified)

The race condition is a classic producer-consumer synchronization issue:

```
Timeline:
Server:  [emit ActionEvent] → [emit ObservationEvent] → [set status=finished]
                                                              ↓
REST:    ─────────────────────────────────────────────→ [poll sees "finished"]
                                                              ↓
WS:      ────────────[ActionEvent]────────[ObservationEvent]──┼──[status event]
                                                              ↓
Client:  ─────────────────────────────────────────────→ run() returns! 😱
                                                        (events still in flight)
```

The PR correctly identifies that REST polling can see terminal status before WS delivers all events.

---

### **[CRITICAL ISSUES]** (Must fix or acknowledge)

**[remote_conversation.py, Line 209] Data Flow / Silent Failure**: `ConnectionClosed` breaks the WS loop entirely instead of retrying.

```python
except websockets.exceptions.ConnectionClosed:
    break  # <-- WS client stops forever, no retry
```

If the server closes the connection gracefully (e.g., during shutdown or load balancing), the WS client dies silently. The fallback mechanism saves us, but there's no logging or visibility into this failure mode. Consider:
- Logging a warning when WS disconnects
- Attempting reconnection for graceful closes (not just errors)

**[remote_conversation.py, Lines 881-887] Race Condition / Queue Draining**: The queue drain at `run()` start has a subtle race.

```python
# Drain any stale terminal status events from previous runs.
while True:
    try:
        self._terminal_status_queue.get_nowait()
    except Empty:
        break
```

What if a terminal status arrives between draining and the server actually starting the run? The 409 handling helps, but consider this sequence:
1. Previous run finishes, terminal status queued
2. `run()` starts, drains queue
3. Server hasn't started new run yet
4. WS delivers another terminal status from previous run (delayed)
5. New run starts
6. `_wait_for_run_completion()` sees the stale terminal status

This is unlikely but possible. A more robust approach would be to include a run ID or sequence number in the terminal status.

---

### **[IMPROVEMENT OPPORTUNITIES]** (Should fix - violates good taste)

**[remote_conversation.py, Line 945] Magic Number**: `TERMINAL_POLL_THRESHOLD = 3` is a magic number with a comment, but it's not configurable.

```python
TERMINAL_POLL_THRESHOLD = 3
```

The comment explains the reasoning well, but this adds 3 seconds of latency in the fallback path. Consider making this configurable via a parameter or class attribute for users who need lower latency and are willing to accept the risk.

**[remote_conversation.py, Lines 958-974] Complexity**: The dual-path waiting logic is getting complex.

```python
try:
    ws_status = self._terminal_status_queue.get(timeout=poll_interval)
    # Handle ERROR/STUCK states - raises ConversationRunError
    self._handle_conversation_status(ws_status)
    logger.info(...)
    return
except Empty:
    pass  # Queue.get() timed out, fall through to REST polling
```

This is acceptable, but the method is doing two things: waiting for WS and polling REST. Consider extracting the WS wait into a separate method for clarity.

**[test_event_loss_repro.py] Test Validity**: The test validates callback completion, not WS-before-REST ordering.

The test injects a delay in the callback:
```python
def delayed_cb(event: Event) -> None:
    if isinstance(event, (ActionEvent, ObservationEvent)) and ...:
        time.sleep(ws_delay_s)  # Delay BEFORE calling orig_cb
    orig_cb(event)
```

Since `run_complete_callback` is part of the composed callbacks (added after user callbacks), the terminal status signal is also delayed. This means the test validates that we wait for callbacks to complete, but doesn't directly validate the WS-before-REST ordering guarantee.

A more direct test would mock the WS to deliver events with a delay AFTER the REST poll sees terminal status.

---

### **[ARCHITECTURAL OBSERVATIONS]** (For future consideration)

**The Fundamental Issue**: The dual REST/WS synchronization model is inherently racy.

The PR description acknowledges this: *"We want a solution for the recent events that doesn't require rethinking the ws/REST sync yet."*

For a more robust long-term solution, consider:
1. **Server-side ordering guarantee**: Ensure the server sends the terminal status event AFTER all other events
2. **Sequence numbers**: Include event sequence numbers so the client can detect gaps
3. **Event ID tracking**: Track the last event ID and fetch missing events by ID range
4. **Abandon WS for events**: Use REST with long-polling for event delivery (simpler, more reliable)

**WS Reconnection Gap**: If WS disconnects and reconnects during a run, events during the gap are lost. The fallback reconcile helps, but only if we hit the 3-poll threshold. This is a known limitation.

---

### **VERDICT:**

✅ **Worth merging** - The core logic is sound and significantly improves reliability. The fix correctly waits for WS to deliver terminal status before returning, ensuring events are delivered through the same channel.

**Key Insight**: The PR's approach of "wait for WS terminal status, fall back to REST + reconcile" is a pragmatic solution that works for the common case (WS delivers events on time) while providing a safety net for edge cases (WS delayed/disconnected).

**Remaining Risks**:
1. WS reconnection during run can still lose events (mitigated by fallback reconcile)
2. 3-poll fallback adds latency when WS fails
3. Silent WS disconnection has no visibility

**Recommendation**: Merge with the understanding that this is a tactical fix. Consider adding WS health logging and making `TERMINAL_POLL_THRESHOLD` configurable in a follow-up.

xingyaoww:
@OpenHands please merge from main, resolve the merge conflict. 

And then try run tests/sdk/conversation/remote/test_remote_conversation.py 1000 times and see if we run into concurrency issue / event loss? Then try switch back to main and run it 1000 times. Record the avg time taken to finish the test and the avg success rate and report back

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -913,20 +955,57 @@ def _wait_for_run_completion(
...
+                # Handle ERROR/STUCK states - raises ConversationRunError
+                self._handle_conversation_status(ws_status)
+
+                logger.info(
+                    "Run completed via WebSocket notification "
+                    "(status: %s, elapsed: %.1fs)",
+                    ws_status,
+                    elapsed,
+                )
+                return
```
xingyaoww, line 976: @OpenHands tell me, is it possible that the terminal status is streamed back FIRST, before the other events are streaming back via websocket?

PR #1850, opened 2026-01-29
[FileStore]: fix Use caching on demand to reduce memory usage

CLFutureX:
### **Background:**
Some FileStore usage scenarios actually don’t need caching.For example, _save_base_state() and _save_processing_history() cause extra memory usage.
### **Change:**
Add direct write and direct read methods to enable on-demand caching.


openhands-tools/openhands/tools/tom_consult/executor.py
```diff
@@ -417,7 +417,7 @@ def _save_processing_history(self, session_ids: list[str]) -> None:
 
             history_file = f"{get_usermodeling_dir(self.user_id)}/processed_sessions_timestamps.json"  # noqa: E501
 
-            self.file_store.write(history_file, json.dumps(history, indent=2))
+            self.file_store.directWrite(history_file, json.dumps(history, indent=2))
```
enyst: Sorry I'm dense, if `read/write` get `cache` argument, do we need `directRead/directWrite`, isn't it with cache false?

CLFutureX: > Sorry I'm dense, if `read/write` get `cache` argument, do we need `directRead/directWrite`, isn't it with cache false?

Yes, i initially planned to add the cache parameter. However, i found it would result in redundant parameters in InMemoryFileStore and cause validation failures. Also, to make the interface clearer and easier to use, i decided to add new methods instead.

neubig:
[automated message] @neubig assigned for review according to git blame

neubig (changes requested):
Hi @CLFutureX , thank you for opening the PR. Before I fully assess the PR, could you please open an issue with a minimal reproducible example of the problem that you're trying to solve? For instance, a command line argument or script that runs the agent where you monitor the agent's memory and find that it is very large. You could then run the same command with this fix implemented and demonstrate that it significantly reduces memory.

That would help me (1) understand the scope of the problem, and (2) be able to confirm the error on my end as well.

Thank you! And once you are done with that, please click the "re-request review" button next to my name on the reviewers panel.

CLFutureX:
> Hi @CLFutureX , thank you for opening the PR. Before I fully assess the PR, could you please open an issue with a minimal reproducible example of the problem that you're trying to solve? For instance, a command line argument or script that runs the agent where you monitor the agent's memory and find that it is very large. You could then run the same command with this fix implemented and demonstrate that it significantly reduces memory.
> 
> That would help me (1) understand the scope of the problem, and (2) be able to confirm the error on my end as well.
> 
> Thank you! And once you are done with that, please click the "re-request review" button next to my name on the reviewers panel.

Thanks for the review. The reason I opened this PR is exactly as described.
Right now, all read and write operations in LocalFileStore use caching by default. I originally added this cache layer to avoid reading from the local file on every event iteration.
However, LocalFileStore is not only used for event storage — it’s also used for other things like reading and writing base_state. In these scenarios, we only need to write to the file and read it once, so caching is unnecessary.
That’s why this PR adds a flag to control whether caching is enabled for read/write operations.

Also, @enyst please take a look at the latest changes: I think adding a parameter is indeed better than adding a new method.

neubig (commented):
Hi @CLFutureX , we still need a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example) for this bug so we can confirm the problem and that it is fixed after the PR. Thank you! If you are able to provide this we can review it, otherwise we will have to close it, because we won't be able to confirm the issue.

Relatedly, please add a test that fails before the PR and passes after the PR demonstrating the problem.

CLFutureX:
> Hi @CLFutureX , we still need a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example) for this bug so we can confirm the problem and that it is fixed after the PR. Thank you! If you are able to provide this we can review it, otherwise we will have to close it, because we won't be able to confirm the issue.
> 
> Relatedly, please add a test that fails before the PR and passes after the PR demonstrating the problem.

Thanks for your review. I'll find some time to add the relevant examples.

CLFutureX:
@neubig 
## Description
This change is an enhancement to the existing functionality, so i have added functional correctness tests:
- `tests/sdk/io/test_filestore_cache.py#test_cache_enable_and_disable`

I also added tests for the usage scenario of this feature:
- `tests/sdk/conversation/local/test_state_serialization.py#test_conversation_state_fs_cache`

### Key Test Detail (test_conversation_state_fs_cache)
Before this change, the cache count was 1 for this test case. After adding the "cache on demand" feature (allowing cache to be enabled/disabled as needed), the conversation state is no longer cached in memory. This is verified by the assertion:
```python
assert len(loaded_state._fs.cache) == 0

neubig (changes requested):
Thanks for adding the tests. However, this still is not the MRE I requested.

The new tests check the internal caching behavior after your change, but they do not minimally reproduce the original problem on `main` (i.e., excessive memory usage), nor do they show that the memory issue is significant in a real runnable scenario.

Please provide a true minimal reproducible example that includes:

1. an exact script or command I can run,
2. the minimal setup/input needed,
3. the expected vs actual behavior,
4. a concrete way to measure memory usage, and
5. ideally, numbers showing the behavior on `main` and on this branch.

For example, this could be a small script that runs the agent or conversation state persistence in a loop, measures RSS / memory usage, and demonstrates that memory grows much larger than expected on `main`.

The assertion `assert len(loaded_state._fs.cache) == 0` is not sufficient for this, because it only checks an internal implementation detail of your fix, not the original bug itself.

Once there is a runnable MRE, I’ll be happy to review again.

CLFutureX:
> Thanks for adding the tests. However, this still is not the MRE I requested.
> 
> The new tests check the internal caching behavior after your change, but they do not minimally reproduce the original problem on `main` (i.e., excessive memory usage), nor do they show that the memory issue is significant in a real runnable scenario.
> 
> Please provide a true minimal reproducible example that includes:
> 
> 1. an exact script or command I can run,
> 2. the minimal setup/input needed,
> 3. the expected vs actual behavior,
> 4. a concrete way to measure memory usage, and
> 5. ideally, numbers showing the behavior on `main` and on this branch.
> 
> For example, this could be a small script that runs the agent or conversation state persistence in a loop, measures RSS / memory usage, and demonstrates that memory grows much larger than expected on `main`.
> 
> The assertion `assert len(loaded_state._fs.cache) == 0` is not sufficient for this, because it only checks an internal implementation detail of your fix, not the original bug itself.
> 
> Once there is a runnable MRE, I’ll be happy to review again.

thanks for  the review,  I think we might have a slight misunderstanding about this PR 

This is actually a feature enhancement. Its main goal is to implement on-demand caching to reduce unnecessary memory usage, rather than fixing the current high memory consumption issue.

For example, this check:
assert len(loaded_state._fs.cache) == 0

What it really shows is:
Previously, we cached session states unconditionally for all read and write operations in LocalFileStore. But in fact, caching should only be used in high-frequency reading scenarios.
With on-demand caching now supported, session states will no longer be cached by default.

neubig:
I see your argument, but the thing is we don't add features unless they are solving a problem.

At the moment, you have not yet demonstrated that there is a problem. What I am asking for is that you demonstrate that there is a problem, for instance demonstrating that OpenHands is using more memory than it should. Once you demonstrate that there is a problem by providing the commands that you used to run the program and monitor memory, you could demonstrate that you ran the same commands after the fix and the memory usage went down.

neubig:
Closing due to lack of activity!

PR #1871, opened 2026-01-31
fix: Add AttributeError to escape_bash_special_chars exception handling

### LINKED ISSUE #1870: [Bug]: AttributeError crash in escape_bash_special_chars when parsing comment-only bash commands
## Bug Description

When executing a bash command that contains only a comment (e.g., `# No command needed here`), the `escape_bash_special_chars()` function crashes with:

```
AttributeError: 'str' object has no attribute 'kind'
```

This crashes the entire conversation instead of gracefully handling the input.

## Root Cause

In `openhands-tools/openhands/tools/terminal/utils/command.py`, the `visit_node()` function inside `escape_bash_special_chars()` recursively visits nodes from bashlex. However, `node.parts` can contain **strings** (not just node objects) for certain tokens like heredoc end markers.

When `visit_node(part)` is called with a string, accessing `node.kind` fails because strings don't have a `.kind` attribute.

## Current Code (line 144)

```python
except (ParsingError, NotImplementedError, TypeError):
```

## Proposed Fix

Add `AttributeError` to the exception handling:

```python
except (ParsingError, NotImplementedError, TypeError, AttributeError):
```

**Note:** This is consistent with `split_bash_commands()` in the same file which already catches `AttributeError` (lines 19-24).

## Steps to Reproduce

1. Start a conversation
2. Execute a terminal action with only a comment: `# This is just a comment`
3. Observe crash: `AttributeError: 'str' object has no attribute 'kind'`

## Expected Behavior

The comment-only command should be handled gracefully (either executed as a no-op or returned as-is).

## Environment

- SDK version: latest
- Python: 3.12

## Additional Context

The same fix pattern was already applied to `split_bash_commands()` in the same file for issue #8369 (see comment on line 25-26), but was missed for `escape_bash_special_chars()`.

daviskay75:
## Summary

Fixes #1870

When parsing comment-only bash commands (e.g., `# comment`), bashlex can return strings in `node.parts` instead of node objects. Accessing `.kind` on these strings causes an `AttributeError` crash that kills the entire conversation.

## Changes

- Added `AttributeError` to the except clause in `escape_bash_special_chars()` (line 144)

## Why this fix?

This is consistent with `split_bash_commands()` in the same file which already handles this case:

```python
# Lines 19-24
except (
    ParsingError,
    NotImplementedError,
    TypeError,
    AttributeError,  # Already handled here!
):
    # Added AttributeError to catch 'str' object has no attribute 'kind' error
    # (issue #8369)
```

The same pattern was missed for `escape_bash_special_chars()`.

## Test plan

- [x] Verify comment-only bash commands no longer crash
- [x] Verify normal bash commands still work correctly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/openhands-tools/openhands/tools/terminal/utils/command.py
+++ b/openhands-tools/openhands/tools/terminal/utils/command.py
@@ -141,7 +141,7 @@ def escape_bash_special_chars(command: str) -> str:
         remaining = command[last_pos:]
         parts.append(remaining)
         return "".join(parts)
-    except (ParsingError, NotImplementedError, TypeError):
+    except (ParsingError, NotImplementedError, TypeError, AttributeError):
         logger.debug(
             f"Failed to parse bash commands for special characters escape\n[input]: "
             f"{command}\n[warning]: {traceback.format_exc()}\nThe original command "


xingyaoww (approved):
Thanks

PR #1880, opened 2026-02-02
feat(websocket): add after_timestamp filter for bi-directional event loading

jpshackelford:
## Summary

Add an `after_timestamp` query parameter to the WebSocket events endpoint that filters events during `resend_all` to only return events with timestamps >= the specified value.

## Background

This is a prerequisite for the bi-directional event loading optimization described in [OpenHands/OpenHands#12705](https://github.com/OpenHands/OpenHands/issues/12705).

## Changes

- Added `after_timestamp: datetime | None` query parameter to `/sockets/events/{conversation_id}` WebSocket endpoint
- When `resend_all=True` and `after_timestamp` is provided, only events with `timestamp >= after_timestamp` are resent
- Added comprehensive tests for the new parameter

## Use Case

This enables efficient bi-directional loading where:
1. **REST API** fetches historical events (paginated backwards from newest)
2. **WebSocket** handles real-time events after a specific timestamp

```
Timeline:
[oldest] ←←←← REST (backwards) ←←←← [newest] →→→→ WebSocket (forward) →→→→ [new events]
```

The underlying `search_events` already supports `timestamp__gte` filtering, so this change simply exposes that capability through the WebSocket API.

## Testing

- Added 3 new tests in `TestAfterTimestampFiltering` class
- Updated existing `test_resend_all_true_resends_events` to verify the new parameter passing

## Related Issue

Resolves: https://github.com/OpenHands/OpenHands/issues/12705 (agent-server prerequisite)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:bce2fa7-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-bce2fa7-python \
  ghcr.io/openhands/agent-server:bce2fa7-python
```

```
ghcr.io/openhands/agent-server:bce2fa7-golang-amd64
ghcr.io/openhands/agent-server:bce2fa7-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:bce2fa7-golang-arm64
ghcr.io/openhands/agent-server:bce2fa7-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:bce2fa7-java-amd64
ghcr.io/openhands/agent-server:bce2fa7-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:bce2fa7-java-arm64
ghcr.io/openhands/agent-server:bce2fa7-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:bce2fa7-python-amd64
ghcr.io/openhands/agent-server:bce2fa7-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:bce2fa7-python-arm64
ghcr.io/openhands/agent-server:bce2fa7-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:bce2fa7-golang
ghcr.io/openhands/agent-server:bce2fa7-java
ghcr.io/openhands/agent-server:bce2fa7-python
```

- Each variant tag (e.g., `bce2fa7-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `bce2fa7-python-amd64`) are also available if needed


enyst:

When you’re done, look at the linked issue and tell WDYT.

Post your feedback directly as a comment on github here. Use smolpaws. Note that it will be rendered as markdown.

smolpaws:
**Taste Rating:** 🟡 **Acceptable** — small change, but there’s a correctness landmine hiding in the timestamp handling.

---

## Linus-Style Analysis

**[CRITICAL ISSUES]**
- [openhands-agent-server/openhands/agent_server/event_service.py, Lines 110–141; openhands-agent-server/openhands/agent_server/sockets.py, Lines 81–89] **Correctness / Data semantics**: `after_timestamp` is a `datetime`, but events store `timestamp` as **string** and comparisons are **lexicographic**. That “works” only if *all* timestamps are normalized to the same ISO format **and** timezone handling is consistent. Right now events default to `datetime.now().isoformat()` (naive), while `after_timestamp` can be tz-aware, which means `"2024-01-15T10:30:00+00:00"` vs `"2024-01-15T10:30:00"` lexicographic order is not reliable. That can silently drop or duplicate events at the boundary. **Fix**: normalize to UTC (or strip tzinfo) before comparison, or parse `event.timestamp` to `datetime` and compare actual datetimes. At minimum, enforce/validate timezone format in the WebSocket param and document it.

- [openhands-agent-server/openhands/agent_server/sockets.py, Lines 47–57] **API contract clarity**: the new param is only honored when `resend_all=True`, but the API silently ignores it otherwise. Consider returning a 400 when `after_timestamp` is set without `resend_all`, or explicitly document in the OpenAPI/README for the socket endpoint. Silent no-ops are a debugging tax.
- [tests/agent_server/test_event_router_websocket.py, Lines 452–588] **Coverage gap**: tests only assert that the param is passed through. You’re not asserting the boundary behavior (`event.timestamp == after_timestamp`). Add one test case to prove inclusive filtering doesn’t cause duplicates or misses — especially given the string-compare behavior.

**[STYLE NOTES]**
- [tests/agent_server/test_event_router_websocket.py, Lines 459–546] Minor: repeated local imports (`datetime`, `cast`) inside tests make the file noisy; pull them to the top-level unless there’s a cycle.

---

**VERDICT:** ✅ **Worth merging** once timestamp normalization is addressed. Otherwise you’re adding a correctness bug under load.

**KEY INSIGHT:** If timestamps aren’t normalized, the filter is gambling on string ordering — that’s not a protocol, it’s a coincidence.

---

## WDYT on OpenHands/OpenHands#12705
The bi-directional loading plan is solid and pragmatic — REST for bulk history + WebSocket for tail is the right shape. Two gotchas to keep in mind:
1. **Boundary duplication**: using `>=` on a timestamp boundary will duplicate if multiple events share the same timestamp or the REST page ends on that exact timestamp. Either make the websocket filter **strictly greater**, or include `(timestamp, id)` ordering and dedupe client-side.
2. **Timestamp consistency**: the backend currently treats timestamps as strings; if the frontend starts mixing timezones or formats, you’ll get ordering bugs. Standardize on UTC ISO strings (or switch to numeric epoch) end-to-end.

Overall: good direction, just nail the timestamp semantics so you don’t ship a subtle “missing events” bug.

jpshackelford:
Thanks for the detailed review! I've pushed fixes that address the timezone handling concerns. Let me explain the architecture and our decisions.

## Background: How Event Timestamps Work

Event timestamps are defined in `openhands-sdk` (not `openhands-agent-server`):

```python
# openhands-sdk/openhands/sdk/event/base.py
timestamp: str = Field(
    default_factory=lambda: datetime.now().isoformat(),
)  # consistent with V1
```

Key points:
- Stored as **string** (ISO format), not datetime object
- Uses **naive local time** (`datetime.now()` without tzinfo)
- Marked "consistent with V1" - backward compatibility constraint

The agent-server's filtering code must work with this existing format.

## The Problem

The original implementation used string comparison for timestamp filtering:
```python
if event.timestamp < timestamp_gte_str:  # string comparison
```

This breaks when ISO strings have different timezone suffixes:
- `"2024-01-15T10:30:00"` (naive)
- `"2024-01-15T10:30:00+00:00"` (tz-aware)

These represent potentially different moments but don't compare correctly as strings.

## Our Solution

We had two options to enable proper datetime comparison:

| Option | Approach | Tradeoff |
|--------|----------|----------|
| A | Normalize filter to naive | One-line change in `normalize_datetime_to_server_timezone` |
| B | Add tzinfo when parsing event timestamps | Changes in `event_service.py` for every comparison |

We chose **Option A** because:
1. Simpler - change is localized to the normalization function
2. Matches the source of truth - events are stored naive, so filters should be naive
3. Both REST and WebSocket already call `normalize_datetime_to_server_timezone`

## Changes Made

1. **WebSocket endpoint**: Added `normalize_datetime_to_server_timezone()` (matching REST)

2. **`normalize_datetime_to_server_timezone()`**: Now returns naive datetime
   ```python
   # Converts to server local time AND strips tzinfo
   return dt.astimezone(None).replace(tzinfo=None)
   ```

3. **`event_service.py`**: Replaced string comparison with datetime comparison
   ```python
   event_dt = datetime.fromisoformat(event.timestamp)
   if event_dt < timestamp__gte:
   ```

## API Contract

| Client provides | Behavior |
|-----------------|----------|
| Naive datetime | Assumed to be server local time, used as-is |
| Tz-aware datetime | Converted to server local time, made naive |

This handles timezones correctly - a client can pass UTC and it will be properly converted to server time before comparison.

## No Breaking Changes

Verified that OpenHands/OpenHands frontend doesn't use `after_timestamp` yet - it only uses `resend_all: true`. This is a new feature, so clients will adopt the documented behavior from the start.

neubig:
[automated message] @neubig assigned for review according to git blame

tests/agent_server/test_event_router.py
```diff
@@ -10,14 +11,76 @@
...
+    normalize_datetime_to_server_timezone,
+)
 from openhands.agent_server.event_service import EventService
 from openhands.agent_server.models import SendMessageRequest
 from openhands.sdk import Message
 from openhands.sdk.event.llm_convertible.message import MessageEvent
 from openhands.sdk.llm.message import ImageContent, TextContent
 
 
+class TestNormalizeDatetimeToServerTimezone:
```
neubig: Typically we avoid test classes and just use test functions, unless a class is actually necessary.

tests/agent_server/test_event_router_websocket.py
```diff
@@ -444,3 +447,362 @@ async def test_resend_all_handles_send_json_exception(
         # WebSocket should still be subscribed and unsubscribed normally
         mock_event_service.subscribe_to_events.assert_called_once()
         mock_event_service.unsubscribe_from_events.assert_called_once()
+
+
+class TestAfterTimestampFiltering:
```
neubig: ditto

tests/agent_server/test_event_router_websocket.py
```diff
@@ -444,3 +447,362 @@ async def test_resend_all_handles_send_json_exception(
...
+    @pytest.mark.asyncio
+    async def test_after_timestamp_timezone_aware_is_normalized(
+        self, mock_websocket, mock_event_service, sample_conversation_id
+    ):
+        """Test that timezone-aware timestamps are normalized to naive server time."""
+        from datetime import datetime
+        from typing import cast
+
+        from openhands.agent_server.models import EventPage
+        from openhands.sdk.event import Event
```
neubig: I think these can be module level imports.

tests/agent_server/test_event_router_websocket.py
```diff
@@ -444,3 +447,142 @@ async def test_resend_all_handles_send_json_exception(
...
+
+        # search_events should be called with timestamp__gte parameter
+        mock_event_service.search_events.assert_called_once_with(
+            page_id=None, timestamp__gte=test_timestamp
+        )
+
+    @pytest.mark.asyncio
+    async def test_after_timestamp_without_resend_all_no_effect(
+        self, mock_websocket, mock_event_service, sample_conversation_id
+    ):
```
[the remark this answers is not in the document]
neubig: agreed

tests/agent_server/test_event_router_websocket.py
neubig, line 1: The added tests in this file seem a bit excessive, maybe we only need 1-2 to test the new functionality?

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -41,8 +42,20 @@ async def events_socket(
     websocket: WebSocket,
     session_api_key: Annotated[str | None, Query(alias="session_api_key")] = None,
     resend_all: Annotated[bool, Query()] = False,
+    after_timestamp: Annotated[datetime | None, Query()] = None,
```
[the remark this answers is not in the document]
jpshackelford: The Query description was already added in a previous commit by the PR author. See sockets.py lines 46-55.

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -41,8 +42,20 @@ async def events_socket(
...
+
+    Args:
+        conversation_id: The conversation ID to subscribe to.
+        websocket: The WebSocket connection.
+        session_api_key: Optional API key for authentication.
+        resend_all: If True, resend all existing events when connecting.
+        after_timestamp: If provided with resend_all=True, only resend events
+            with timestamps >= this value. Enables efficient bi-directional
+            loading where REST fetches historical events and WebSocket handles
+            events after a specific point in time.
```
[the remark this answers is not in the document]
jpshackelford: The docstring with timezone documentation was already added in a previous commit by the PR author. See sockets.py lines 57-69.

tests/agent_server/test_event_router_websocket.py
```diff
@@ -444,3 +447,142 @@ async def test_resend_all_handles_send_json_exception(
...
         mock_event_service.subscribe_to_events.assert_called_once()
         mock_event_service.unsubscribe_from_events.assert_called_once()
+
+
+class TestAfterTimestampFiltering:
+    """Test cases for after_timestamp parameter functionality."""
+
+    @pytest.mark.asyncio
+    async def test_after_timestamp_passed_to_search_events(
+        self, mock_websocket, mock_event_service, sample_conversation_id
```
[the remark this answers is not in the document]
jpshackelford: Fixed in 00985c47 - moved datetime and cast imports to module level.

tests/agent_server/test_event_router_websocket.py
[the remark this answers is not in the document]
jpshackelford: Fixed in 00985c47 - all imports are now at module level.

tests/agent_server/test_event_router.py
jpshackelford: Fixed in 00985c47 - converted TestNormalizeDatetimeToServerTimezone class to standalone test functions.

tests/agent_server/test_event_router_websocket.py
jpshackelford: Fixed in 00985c47 - converted all test classes in test_event_router_websocket.py to standalone test functions.

tests/agent_server/test_event_router_websocket.py
jpshackelford: Fixed in 00985c47 - moved datetime, cast, EventPage, and Event imports to module level.

tests/agent_server/test_event_router_websocket.py
jpshackelford: Fixed in 00985c47 - reduced test count significantly. Removed TestAfterTimestampFilteringBehavioral class entirely (duplicative with event_service tests), consolidated TestAfterTimestampFiltering to 2 key tests, and removed redundant test classes. Total websocket tests reduced from 35+ to 11.

jpshackelford:
## Summary: API Breakage Checks

### ✅ **API breakage checks (Griffe)** - **PASSED**
- Checks Python SDK's programmatic API for breaking changes
- No breaking changes detected in `openhands-sdk`, `openhands-workspace`, or `openhands-tools`
- This PR doesn't introduce any Python API breaking changes

### ⚠️ **Agent server REST API breakage checks (OpenAPI)** - **Failed (but non-blocking)**

**What happened:**
```
Breaking REST API change detected without MINOR version bump (1.12.0 -> 1.12.0).
- the 'file' request property type/format changed from 'string'/'' to 'string'/'binary'
```

**Key findings:**
1. **This failure is NOT caused by this PR** - I verified by running the same check on `origin/main` and it produces the identical error
2. The `file` property change (from `string/''` to `string/'binary'`) is a pre-existing breaking change in main that hasn't been addressed
3. **The check has `continue-on-error: true`** for non-release PRs (`!startsWith(github.head_ref, 'rel-')`) so it won't block this PR from merging

**Why it doesn't block:**
- The workflow is configured to report failures via PR comment but not gate merges for feature branches
- Only release branches (`rel-*`) are blocked by this check

**Action needed:**
- None for this PR - it's a pre-existing issue in main
- The repo maintainers may want to address this separately by either:
  1. Bumping the MINOR version to acknowledge the breaking change, or
  2. Investigating why the `file` property type/format changed

jpshackelford:
@neubig Ready for final review

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -92,8 +94,30 @@ async def events_socket(
...
-    """WebSocket endpoint for conversation events."""
+    """WebSocket endpoint for conversation events.
+
+    Args:
+        conversation_id: The conversation ID to subscribe to.
+        websocket: The WebSocket connection.
+        session_api_key: Optional API key for authentication.
+        resend_all: If True, resend all existing events when connecting.
+        after_timestamp: If provided with resend_all=True, only resend events
+            with timestamps >= this value. Timestamps are interpreted in server
```
enyst: If I use an API and see it has `resend_all` to resend all events after connecting, and `after_timestamp` to send only those after timestamp, what seems to make sense is that:
- when `resend_all=False` or None 
- and `after_timestamp` is set, then send only the newer events.

Shouldn't we switch this boolean's meaning and behavior?

jpshackelford: even more clear would be to deprecate resend_all and have resend_mode: None | all | after_timestamp

enyst: That sounds good!

neubig:
@jpshackelford , thanks! When you want to re-request my review, could you press the "cycle" button next to my name under reviewers? That'll put it back on my review queue, and I'm better at checking that than my github messages (which are often a bit noisy). I'll click it for myself for now.

jpshackelford:
## Implemented: `resend_mode` API with `resend_all` deprecation

Following the discussion with @enyst, I've implemented the cleaner API design:

### New API

```
resend_mode: Literal['all', 'since'] | None
after_timestamp: datetime | None
```

**Usage:**
- `?resend_mode=all` → sends all events (replaces `resend_all=true`)
- `?resend_mode=since&after_timestamp=2024-01-15T10:30:00Z` → sends events since timestamp
- No params (default) → just subscribe, no historical events

### Backward Compatibility

`resend_all` is **deprecated but still functional**:
- Hidden from OpenAPI schema (`include_in_schema=False`)
- Logs a deprecation warning when used
- `?resend_all=true` continues to work (treated as `resend_mode='all'`)

### Migration Path

| Phase | API Surface | Breaking? |
|-------|-------------|-----------|
| Now | `resend_mode`, `after_timestamp`, `resend_all` (hidden) | No |
| Deprecation period | Warn in logs when `resend_all` used | No |
| Future major release | Remove `resend_all` | Yes |
| Final | `resend_mode`, `after_timestamp` | - |

### Frontend Migration (when ready)

```diff
- { resend_all: true }
+ { resend_mode: 'all' }
```

Commit: 2d7f74b

openhands-agent-server/openhands/agent_server/sockets.py
jpshackelford: Implemented in 2d7f74b - Added `resend_mode` parameter with values `all`, `since`, or `None`. Deprecated `resend_all` (hidden from schema, logs warning when used). See PR comment for details.

Issue #781, opened 2025-10-17
Proposal: Multiple LLM profiles


enyst:
Following up on @xingyaoww 's issue here:
- https://github.com/All-Hands-AI/OpenHands/issues/11412

**Problem:**
I'm looking at several tabs, and I need to change the default LLMs for debugging. I run into exceptions and copy-paste a lot. I had to create conversations only for this: it sits there and switches contents of `.json` files for me, all across conversations. 😓 
For both agent and condenser.

This seems cumbersome to keep doing, and it still runs into exceptions when I disable the security analyzer (because the LLM I started with is no longer the default LLM)...

IMHO users of the SDK shouldn't have to do the same. 

**Proposal:**
I suggest to support in the SDK
- the ability to pick up *names* from somewhere (from parameters, from env) and apply them
- store the definitions ("gpt-5.json", "sonnet-45.json") in the persistence directory, same place where we store `agent_settings.json`: `~/.openhands/`

That's a part of what the CLI and other apps would offer.

**Notes:**
LLM configurations are currently managed in the SDK, under the form of 
- LLM
- condenser in `agent_settings`, 
- secondary LLM for routing
- LLM for autotitle.

We also persist already at least two: 'agent' and 'condenser', in at least two places: the agent settings `.json` and the conversation `.json`.

xingyaoww:
I actually wonder if we can implement this as part of the LLMRegistry? Essentially we just need a way to:
1. persist the LLMRegistry to disk (a folder of JSONs)
2. load them back to memory
3. be able to grab LLM from it in CLI/GUI etc

xingyaoww:
cc @malhotra5 who might have more context

xingyaoww:
@OpenHands do NOT create new PR -- do some research in the repo, help me figure out the feasibility of implementing a persisted version of LLMRegistry

enyst:
> I created 11 comprehensive research documents (~100KB)

LOL

I get that a lot about Sonnet 4.5! People get .md files everywhere. (I had some too, but I'm only occasionally testing it, codex is too good.)

seekayel:
@xingyaoww looking at picking this up, any learnings from the research performed by the agent? in particular the `poc_persisted_registry.py` implementation? Any other context updates since opened two months ago?

enyst:
[...]

PR #1882, opened 2026-02-02
feat: Persistent LLM profiles

VascoSch92:
## Summary

The PR propose a solution for issue #781 

The class `ProfileManager` is introduced, whose sole job is to save, store, and update profiles on disk. Every time we work with persistent profiles, it should be done through this API to ensure a consistent and reliable way to store, load, and update profiles.

The class can then be used to implement a profile selection solution for the `Openhands-CLI` and in `LLMRegistry` to save or load registries coming from different experiments or runs.

### manager.py

The script contains the `ProfileManager` class, which has the following characteristics:
- The class manages loading, saving, deleting, and updating persistent profiles.
- The class uses a configuration file as a sort of index to load files lazily, preventing the need to load the entire profile store at once, thereby saving time and memory.
- The solution works fine in settings where there are no multiple `ProfileManager` instances working in the same directory (e.g., `Openhands-CLI`). Otherwise, concurrency issues could arise. Since I assume these are mostly corner cases, I haven't implemented any countermeasures yet. Does that make sense?

### config.py

The script contains `Pydantic` models used to validate the configuration file associated with a profile store. These models are backward compatible.

### llm_registry.py

Several modifications were made to the `LLMRegistry` class; however, the API of the class did not change, ensuring backward compatibility. Shortly:
- Two new parameters were added: `load_from_disk` and `profile_manager_path`.
    - `load_from_disk` is a boolean flag that determines whether to use profiles persisted on disk.
    - If `load_from_disk` is True, profile_manager_path specifies the location of the profiles.

    Two variables were chosen instead of one because of the default path for the profile manager.
- The property `usage_to_llm` is deprecated. Currently, it returns the internal state `_usage_to_llm` rather than a copy, which is not ideal. If access is needed, a copy or proxy map should be returned. The property was used only in a few tests, and since we do not want to change the API, it was simply deprecated.
- The `get` method now lazily loads profiles from the store when necessary.
- A new method, `export_profiles`, was added to save the registry. This is very helpful when experimenting or sharing LLM profiles between coworkers.

## Checklist


openhands-sdk/openhands/sdk/llm/llm_registry.py
```diff
@@ -64,6 +76,11 @@ def notify(self, event: RegistryEvent) -> None:
                 logger.warning(f"Failed to emit event: {e}")
 
     @property
+    @deprecated(
+        deprecated_in="1.6.0",
+        removed_in="1.7.0",
+        details="The method is deprecated and will be removed in next versions",
+    )
     def usage_to_llm(self) -> dict[str, LLM]:
```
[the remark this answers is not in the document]
enyst: For clarity, `usage_id` is an id used for the LLM when computing conversation stats, e.g. we wanted to be able to say "the `agent` LLM used 120k tokens total", "the `condenser` LLM used 45k tokens total".

I think they're not profile ids 🤔 WDYT?

openhands-sdk/openhands/sdk/llm/profiles/manager.py
```diff
@@ -0,0 +1,205 @@
...
+        self.base_dir.mkdir(parents=True, exist_ok=True)
+
+    def _ensure_config_file(self) -> None:
+        """Create an empty configuration if none is provided."""
+        if not self.config_path.is_file():
+            config = ProfilesConfig.empty()
+            self.config_path.write_text(json.dumps(config.model_dump(), indent=2))
+
+    def _load_config(self) -> ProfilesConfig:
+        return ProfilesConfig.from_json(self.config_path)
```
enyst, line 60: I wonder if it's a config file necessary? We could maybe just load the profiles from their directory

.openhands/profiles/gpt4.json
```diff
@@ -0,0 +1,27 @@
...
+  "disable_stop_word": false,
+  "caching_prompt": true,
+  "log_completions": false,
+  "log_completions_folder": "logs/completions",
+  "native_tool_calling": true,
+  "reasoning_effort": "high",
+  "enable_encrypted_reasoning": true,
+  "prompt_cache_retention": "24h",
+  "extended_thinking_budget": 200000,
+  "usage_id": "gpt_agent",
```
enyst, line 25: I think this could be `condenser` or any other LLM

examples/01_standalone_sdk/36_llm_profiles.py
```diff
@@ -0,0 +1,66 @@
...
+    # we can check the available profiles by usage id
+    print(f"Available profiles by usage id: {registry.list_usage_ids()}")
+
+    # and load our preferred profile for the
+    # task we want to accomplish
+    code_llm = registry.get("claude_agent")
+    code_agent = Agent(llm=code_llm, tools=[Tool(name=TerminalTool.name)])  # noqa: F841
+
+    summary_llm = registry.get("gpt_agent")
+    summary_agent = Agent(llm=summary_llm, tools=[Tool(name=TerminalTool.name)])  # noqa: F841
```
enyst, line 49: I see, I think this is confusing in the codebase, but it seems to me that usage_ids are orthogonal to profiles. It's a different feature, and if we merge them into one I'm not sure we can achieve both
- stats per "usage" (e.g. total stats for the agent, no matter what LLMs it used)
- the user can choose any profile for agent

Did that make sense?

.openhands/profiles/config.json
```diff
@@ -0,0 +1,23 @@
...
+  "profiles": [
+    {
+      "name": "claude-sonnet-4-5-20250929",
+      "file": "claude-sonnet-4-5-20250929.json",
+      "description": "claude_agent",
+      "usage_id": "claude_agent"
+    },
+    {
+      "name": "gpt4",
+      "file": "gpt4.json",
```
enyst, line 12: If we use the rule "profile_id.json is the filename", then maybe we don't need a config file?

openhands-sdk/openhands/sdk/llm/llm_registry.py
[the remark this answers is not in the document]
VascoSch92: My point was that allowing an external caller to change an internal state breaks the API contract.

Could you point me to an example where this property is used? I searched the repo and it seems it's never referenced.

If you still want to keep this property, I suggest returning:
- a copy of the dictionary (ok but not ideal)
- a `MappingProxyType` (better solution)

examples/01_standalone_sdk/36_llm_profiles.py
VascoSch92: Yes, I had some thoughts about that too.

The problem is that there is no a 1-to-1 correspondence between `usage_ids` and `profiles`, i.e., you could have one profile with multiple usage ids. 

For that reason the `usage_id` in the `ProfileMetadata` is optional. On the other hand, the profiles config asks for a list of profiles metadata with different names as the name is used to retrieve the correct profile.

openhands-sdk/openhands/sdk/llm/profiles/manager.py
VascoSch92: Yes, this could be an option but I think this depends on what we want to solve and which user-case this class want to address.

VascoSch92:
Hey @enyst, thank you very much for the feedback!

Here are a couple of thoughts to help clarify the situation. I’d love to know what you think about the following:

**Goal:** We want a solution that allows the user to load or switch profiles on the fly.

**Context:** A solution proposed by @xingyaoww [[here](https://github.com/OpenHands/software-agent-sdk/issues/781#issuecomment-3416345223)] suggested allowing the `LLMRegistry` to look into persistent files. This makes sense, as the `LLMRegistry` is used in `LocalConversation` to manage different profiles.

However, the `LLMRegistry` uses usage_id as a primary key, while the proposed `ProfileManager` uses `name`. This could lead to inconsistencies since there isn't a one-to-one mapping between the two.

**A possible Solution:** We could update the `ProfileManager` to work with two parameters: `name` and `usage_id`. This would mean:

We can save the same LLM with different `usage_id` values. (This should also work for saving Agents, and we could potentially remove the Config-related Pydantic models).

This would require updating the `LLMRegistry` API if we want the `ProfileManager` to support it. Specifically, the `.get()` method would now require the name of the profile as an additional parameter.

Makes sense?

openhands-sdk/openhands/sdk/llm/llm_registry.py
```diff
@@ -114,9 +142,53 @@ def get(self, usage_id: str) -> LLM:
...
+        return llm_instance
 
     def list_usage_ids(self) -> list[str]:
         """List all registered usage IDs."""
+        in_memory_usage_ids = set(self._usage_to_llm.keys())
+        on_disk_usage_ids = (
+            set(self.profile_manager.get_id_usages())
+            if self.profile_manager is not None
+            else set()
+        )
```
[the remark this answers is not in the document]
VascoSch92: I see that there is a corner case where this can happens. However, the user can put `load_from_disk = False` and the code will return just in-memory profiles.

Note that the default for `load_from_disk` is False.

openhands-sdk/openhands/sdk/llm/profiles/manager.py
```diff
@@ -0,0 +1,205 @@
...
+
+        raise FileNotFoundError(f"Unknown profile `{name}`")
+
+    @cache
+    def get_default_profile(self) -> LLM | None:
+        """Return the default LLM profile."""
+        default_profile_name = self.config.default_profile
+        if default_profile_name is None:
+            return None
+        return self.load_profile(default_profile_name)
```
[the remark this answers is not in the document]
VascoSch92: While `delete_profile() `doesn't require cache eviction, it’s best to defensively clear the cache in `save_profile().`

openhands-sdk/openhands/sdk/llm/profiles/config.py
```diff
@@ -0,0 +1,83 @@
...
+from pydantic import BaseModel, Field, model_validator
+
+
+_INVALID_PROFILE_NAMES: Final[set[str]] = {"", ".", "..", "config"}
+
+
+class ProfileMetadata(BaseModel):
+    """Metadata associate to a profile."""
+
+    name: str = Field(description="Name of the profile", pattern=r"^[A-Za-z0-9._-]+$")
```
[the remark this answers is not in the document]
VascoSch92: Makes sense... also test updated!

examples/01_standalone_sdk/36_llm_profiles.py
```diff
@@ -0,0 +1,66 @@
...
+        usage_id="new_task_agent",
+        model="gemini-flash",
+        api_key=None,
+    )
+
+
+def main() -> None:
+    # initialize the registry with the profile store
+    registry = LLMRegistry(
+        load_from_disk=True,
```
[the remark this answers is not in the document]
VascoSch92: If the directory doesn't exist, the `ProfileManager` will create one.

.openhands/profiles/config.json
VascoSch92: Yes. The only point is: where we store the info about the default_profile, i.e., how do we know which one is the default profile?

openhands-sdk/openhands/sdk/llm/llm_registry.py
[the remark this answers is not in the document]
enyst: Ah ok, I don't think we care about the property, if it's not used let's kill it!

openhands-sdk/openhands/sdk/llm/profiles/manager.py
enyst: Here's what I'm thinking:
- we have generally avoided to make config files in the SDK
- we do save/load things (conversations, events, hooks I think), though enabled by user choice - I believe saving/loading LLM profiles is fine
- but I think maybe a config file could be left to applications, not in the SDK.

I'd be happy to be proven wrong though. There's an argument that doing it here would assure consistency between applications (CLI and web apps), but I don't know if this is the time to change that... WDYT?

enyst:
Thank you @VascoSch92 for thinking it through! To be honest, this can get complicated fast.

I think what matters is:
> (start with) - `agent` -> `gpt-5`
> (switch to)  - `agent`-> `sonnet-45`
> (at restore) - `agent` -> will be `sonnet-45`

where `agent` is usage_id, and those are profile ids. This will be within a single run, and we'd be able to compute total cost _for agent_, which gets displayed on conversations windows and so on.

If we can do that, we're good IMHO!

Now, the same goes for other buckets of `usage`.

My current conclusion playing with it, is that it's simpler to:
- use only profiles, not LLMs in conversations code etc, so we should have a default profile no matter what the user does
- use `agent` for agent, `condenser` for condenser, specific other components are also named
    - I did not allow for override of these names anymore
- I think that implies to save `profile_id` in the conversation `base_state.json`, so that we can restore the profile. 🤔 

Then, it seems to me, we have the question: how about remote? 
- currently, a client needs to send LLMs serialized (so the "contents" of the profile)
    - of course, since the remote is agent server, and it doesn't yet know profiles
- I think maybe we could send profile id from the client, it seems necessary? e.g. user selects a profile -> start conversation / switch to that profile

examples/01_standalone_sdk/36_llm_profiles.py
```diff
@@ -0,0 +1,66 @@
...
+
+def _get_llm_profile() -> LLM:
+    return LLM(
+        usage_id="new_task_agent",
+        model="gemini-flash",
+        api_key=None,
+    )
+
+
+def main() -> None:
```
xingyaoww, line 33: Let's break the `main` function into scripts directly and save a level of indentation.

@OpenHands please implement this change and /remember this preference in AGENTS.md

examples/01_standalone_sdk/36_llm_profiles.py
xingyaoww: @OpenHands please do a /codereview-roasted and look up ALL the usage of `usage_ids` (including all LLM-specific IDs), then zoom out a bit and understand the purpose of this PR. Do you think if the current implementation in this PR is a good idea? If not, what do you think would be the best approach?

Currently it seems with this PR, we will have `LLMRegistry` -> `LLMProfile` -> `LLM` and each of this concept have their own notion of "IDs". Any way to simplify this hierarchy? I imagine a profile is just a thin wrapper around an `LLM` (with a user-specified name), and `LLMRegistry` should just be a collection of `LLM | LLMProfile` - do you think if this is a good idea?

xingyaoww: ^ I kinda like this idea - @VascoSch92 @enyst wdyt?

VascoSch92: I think this is a good idea.

My points in favor of the proposed OpenHands implementation are:
1. `LLMRegistry()` was originally developed as an in-memory solution. Adding a persistent layer "behind the scenes" is creating convoluted and error-prone methods.
2. The solution provides the functionality needed by both the CLI and the web app to save and store profiles, while respecting the SDK philosophy discussed [here](https://github.com/OpenHands/software-agent-sdk/pull/1882#discussion_r2763591503).
3. We delegate the task of defining the default profile to the applications using the `SDK`.
4. An `SDK` user can programmatically access persisted profiles and add them to the `LLMRegistry()`.

If there are no major concerns regarding this approach, I will be happy to open a new, cleaner PR with the implementation.

xingyaoww: Yep, sounds good to me :) 
Feel free to either open new PR or just update this PR

VascoSch92: Yes... I will open a new pr and linking this one so we don't lose the why we decided for such implementation ;)

VascoSch92:
Opening new PR (#1928) after discussion and closing this one.

Issue #1808, opened 2026-01-24
Investigate prompt caching differences between Claude Code and OpenHands


neubig:
## Summary

Reports suggest that Claude Code may be cheaper than OpenHands for the same query.

This appears to be reproducible by sending a simple "hi" query to both tools in two subsequent sessions.

- [x] Reproduce the situation where sending the same prompts to OpenHands and Claude Code results in higher costs
- [ ] Diagnose the error and propose a fix

## References

- [Anthropic Prompt Caching Docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching)
- [Research: Evaluation of Prompt Caching for Agentic Tasks](https://arxiv.org/html/2601.06007v1)
- Internal Slack report on Claude Code vs OpenHands caching disparity

xingyaoww:
Anthropic prompt cache automatically invalidates after 5 minutes - https://platform.claude.com/docs/en/build-with-claude/prompt-caching#1-hour-cache-duration - which could increase the cost if users are not sending follow-up requests within 5 minutes.

We can possibly just enable this 1-hour-cache-duration flag by default to CLI, which might help with the cosyt in this case

I was wondering if we can include these as the first few telemetry we can include into our CLI (cc @malhotra5 ), eg tracking the time on average between two user messages.

If avg is > 5 minutes in general, we should set to 1 hr as the default cache TTL.

@OpenHands can you clone OpenHands/OpenHands-CLI and then check how we sent PostHog events as part of the "Critic feedback widget"? Then refactor this PostHog stuff to a shared util, and add a new settings knob in the UI settings that allows users to turn on/off PostHog metrics (We should not allow them to turn off PostHog if they have Critic model enabled - this is required).

Then, can you collect the number of CLI conversations (one event per conversation), each agent message event OR agent finish event (including metadata like when it is finished & the token status), as well as each user message event's metadata (when it is sent) and the LLM metrics right after it (so we can check for prompt cache hit rate).

When you are done, send a draft PR to OpenHands-CLI

xingyaoww:
## Cache Hit Rate Analysis from Nightly Tests (Claude Sonnet 4.5)

I analyzed the cache behavior from the [nightly test run](https://github.com/OpenHands/software-agent-sdk/issues/939#issuecomment-3814420893) for Claude Sonnet 4.5.

### Overall Results
- **Total prompt tokens**: 305,487
- **Cache read tokens**: 222,512 (**72.8% hit rate**)
[...]

PR #1890, opened 2026-02-03
fix: separate static system prompt from dynamic context for cross-conversation caching

neubig:
## Summary

This PR enables cross-conversation prompt caching by separating the static system prompt from per-conversation dynamic context (hosts, repos, users, skills, secrets, etc.).

Fixes https://github.com/OpenHands/software-agent-sdk/issues/1808

## Problem

Currently, the system prompt includes dynamic per-conversation content (sandbox URLs, working directories, repo info, etc.). This means every conversation has a unique system prompt, preventing Anthropic's prompt caching from working across conversations.

Evidence from OpenHands Cloud shows two conversations with identical "Hi" messages both showing:
- Cache Hit: 0
- Cache Write: ~27k tokens

This means each conversation pays the full cost of processing the system prompt instead of benefiting from cache hits.

## Root Cause

The dynamic content comes from `AgentContext.get_system_message_suffix()` which includes:
- Sandbox URLs with unique runtime IDs (e.g., `work-1-jqfdlohypqaduxwq.prod-runtime.all-hands.dev`)
- Working directory paths
- Repository information
- Skills and secrets

This content was being appended directly to the system prompt, making the entire prompt unique per conversation.

## Solution

Separate the static and dynamic portions:

1. **`static_system_message`** - Returns only the base system prompt template (cacheable across all conversations)
2. **`dynamic_context`** - Returns the per-conversation context (hosts, repos, etc.)

The `SystemPromptEvent` now has an optional `dynamic_context` field. When present, `to_llm_messages()` returns two messages:
1. System message with static prompt (cacheable)
2. User message with dynamic context (not cached, but much smaller)

## Changes

- Add `static_system_message` and `dynamic_context` properties to `AgentBase`
- Add `dynamic_context` field to `SystemPromptEvent`
- Add `to_llm_messages()` method that returns separate system and user messages
- Update `events_to_messages()` to use `to_llm_messages()` for `SystemPromptEvent`
- Add regression test to ensure static system message stays constant

## Expected Impact

With this change:
- Conversation 1: Cache writes the static system prompt (~25k tokens)
- Conversation 2+: Cache HITS on the static system prompt, only processes dynamic context (~2k tokens)

This should significantly reduce costs and latency for subsequent conversations.

## Testing

Added regression test `test_static_system_message_is_constant_across_different_contexts` that verifies the static system message is identical regardless of the `AgentContext` provided.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:bea2cc6-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-bea2cc6-python \
  ghcr.io/openhands/agent-server:bea2cc6-python
```

```
ghcr.io/openhands/agent-server:bea2cc6-golang-amd64
ghcr.io/openhands/agent-server:bea2cc6-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:bea2cc6-golang-arm64
ghcr.io/openhands/agent-server:bea2cc6-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:bea2cc6-java-amd64
ghcr.io/openhands/agent-server:bea2cc6-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:bea2cc6-java-arm64
ghcr.io/openhands/agent-server:bea2cc6-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:bea2cc6-python-amd64
ghcr.io/openhands/agent-server:bea2cc6-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:bea2cc6-python-arm64
ghcr.io/openhands/agent-server:bea2cc6-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:bea2cc6-golang
ghcr.io/openhands/agent-server:bea2cc6-java
ghcr.io/openhands/agent-server:bea2cc6-python
```

- Each variant tag (e.g., `bea2cc6-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `bea2cc6-python-amd64`) are also available if needed

### DIFF
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -102,6 +102,15 @@ MODELS = {
         "display_name": "MiniMax M2.1",
         "llm_config": {"model": "litellm_proxy/minimax/MiniMax-M2.1"},
     },
+    "jade-spark-2862": {
+        "id": "jade-spark-2862",
+        "display_name": "Jade Spark 2862",
+        "llm_config": {
+            "model": "litellm_proxy/jade-spark-2862",
+            "temperature": 1.0,
+            "top_p": 0.95,
+        },
+    },
     "deepseek-v3.2-reasoner": {
         "id": "deepseek-v3.2-reasoner",
         "display_name": "DeepSeek V3.2 Reasoner",
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -274,8 +274,8 @@ jobs:
             - name: Run integration test evaluation for ${{ matrix.job-config['name'] }}
               env:
                   LLM_CONFIG: ${{ toJson(matrix.job-config['llm-config']) }}
-                  LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
-                  LLM_BASE_URL: https://llm-proxy.app.all-hands.dev
+                  LLM_API_KEY: ${{ secrets.LLM_API_KEY_EVAL }}
+                  LLM_BASE_URL: https://llm-proxy.eval.all-hands.dev
               run: |
                   set -eo pipefail
 
--- a/tests/github_workflows/test_resolve_model_config.py
+++ b/tests/github_workflows/test_resolve_model_config.py
@@ -2,9 +2,11 @@
 
 import sys
 from pathlib import Path
+from typing import Any
 from unittest.mock import patch
 
 import pytest
+from pydantic import BaseModel, field_validator, model_validator
 
 
 # Import the functions from resolve_model_config.py
@@ -14,6 +16,85 @@ from resolve_model_config import (  # noqa: E402  # type: ignore[import-not-foun
     MODELS,
     find_models_by_id,
 )
+
+
+class LLMConfig(BaseModel):
+    """Pydantic model for LLM configuration validation."""
+
+    model: str
+    temperature: float | None = None
+    top_p: float | None = None
+    reasoning_effort: str | None = None
+    disable_vision: bool | None = None
+    litellm_extra_body: dict[str, Any] | None = None
+
+    @field_validator("model")
+    @classmethod
+    def model_must_start_with_litellm_proxy(cls, v: str) -> str:
+        if not v.startswith("litellm_proxy/"):
+            raise ValueError(f"model must start with 'litellm_proxy/', got '{v}'")
+        return v
+
+    @field_validator("temperature")
+    @classmethod
+    def temperature_in_range(cls, v: float | None) -> float | None:
+        if v is not None and not (0.0 <= v <= 2.0):
+            raise ValueError(f"temperature must be between 0.0 and 2.0, got {v}")
+        return v
+
+    @field_validator("top_p")
+    @classmethod
+    def top_p_in_range(cls, v: float | None) -> float | None:
+        if v is not None and not (0.0 <= v <= 1.0):
+            raise ValueError(f"top_p must be between 0.0 and 1.0, got {v}")
+        return v
+
+    @field_validator("reasoning_effort")
+    @classmethod
+    def reasoning_effort_valid(cls, v: str | None) -> str | None:
+        valid_values = {"low", "medium", "high"}
+        if v is not None and v not in valid_values:
+            raise ValueError(
+                f"reasoning_effort must be one of {valid_values}, got '{v}'"
+            )
+        return v
+
+
+class EvalModelConfig(BaseModel):
+    """Pydantic model for evaluation model configuration validation."""
+
+    id: str
+    display_name: str
+    llm_config: LLMConfig
+
+    @field_validator("id")
+    @classmethod
+    def id_not_empty(cls, v: str) -> str:
+        if not v.strip():
+            raise ValueError("id cannot be empty")
+        return v
+
+    @field_validator("display_name")
+    @classmethod
+    def display_name_not_empty(cls, v: str) -> str:
+        if not v.strip():
+            raise ValueError("display_name cannot be empty")
+        return v
+
+
+class EvalModelsRegistry(BaseModel):
+    """Pydantic model for the entire MODELS registry validation."""
+
+    models: dict[str, EvalModelConfig]
+
+    @model_validator(mode="after")
+    def id_matches_key(self) -> "EvalModelsRegistry":
+        for key, config in self.models.items():
+            if config.id != key:
+                raise ValueError(
+                    f"Model key '{key}' doesn't match id field '{config.id}'"
+                )
+        return self
 
 
 def test_find_models_by_id_single_model():
@@ -120,60 +201,29 @@ def test_find_models_by_id_preserves_full_config():
     assert result[0]["llm_config"]["temperature"] == 0.0
 
 
-# Tests for expected models from issue #1495
-# Note: claude-4.5-sonnet is implemented as claude-sonnet-4-5-20250929 (pinned version)
-EXPECTED_MODELS = [
-    "claude-4.5-opus",
-    "claude-sonnet-4-5-20250929",
-    "gemini-3-pro",
-    "gemini-3-flash",
-    "gpt-5.2",
-    "gpt-5.2-high-reasoning",
-    "gpt-oss-20b",
-    "kimi-k2-thinking",
-    "minimax-m2",
-    "minimax-m2.1",
-    "deepseek-v3.2-reasoner",
-    "qwen-3-coder",
-    "glm-4.7",
-    "qwen3-coder-next",
-    "qwen3-coder-30b-a3b-instruct",
-]
+def test_all_models_valid_with_pydantic():
+    """Test that all models pass Pydantic validation.
+
+    This single test validates:
+    - All required fields are present (id, display_name, llm_config, llm_config.model)
+    - Model id field matches dictionary key
+    - model starts with 'litellm_proxy/'
+    - temperature is between 0.0 and 2.0 (if present)
+    - top_p is between 0.0 and 1.0 (if present)
+    - reasoning_effort is one of 'low', 'medium', 'high' (if present)
+    """
+    # This will raise ValidationError if any model is invalid
+    registry = EvalModelsRegistry(models=MODELS)
+    assert len(registry.models) == len(MODELS)
 
 
-def test_all_expected_models_present():
-    """Test that all expected models from issue #1495 are present."""
-    for model_id in EXPECTED_MODELS:
-        assert model_id in MODELS, f"Model '{model_id}' is missing from MODELS"
+def test_find_all_models():
+    """Test that find_models_by_id works for all models."""
+    all_model_ids = list(MODELS.keys())
+    result = find_models_by_id(all_model_ids)
 
-
-def test_expected_models_have_required_fields():
-    """Test that all expected models have required fields."""
-    for model_id in EXPECTED_MODELS:
-        model = MODELS[model_id]
-        assert "id" in model, f"Model '{model_id}' missing 'id' field"
-        assert "display_name" in model, f"Model '{model_id}' missing 'display_name'"
-        assert "llm_config" in model, f"Model '{model_id}' missing 'llm_config'"
-        assert "model" in model["llm_config"], (
-            f"Model '{model_id}' missing 'model' in llm_config"
-        )
-
-
-def test_expected_models_id_matches_key():
-    """Test that model id field matches the dictionary key."""
-    for model_id in EXPECTED_MODELS:
-        model = MODELS[model_id]
-        assert model["id"] == model_id, (
-            f"Model key '{model_id}' doesn't match id field '{model['id']}'"
-        )
-
-
-def test_find_all_expected_models():
-    """Test that find_models_by_id works for all expected models."""
-    result = find_models_by_id(EXPECTED_MODELS)
-
-    assert len(result) == len(EXPECTED_MODELS)
-    for i, model_id in enumerate(EXPECTED_MODELS):
+    assert len(result) == len(all_model_ids)
+    for i, model_id in enumerate(all_model_ids):
         assert result[i]["id"] == model_id
 
 


neubig:
@openhands fix the failing CI, and then run the first example from the example directory and see if it completes as expected.

simonrosenberg:
Validated in GUI 
Screenshot 2026-02-05 at 13 29 43
Screenshot 2026-02-05 at 13 29 52

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
```diff
@@ -45,8 +73,32 @@ def visualize(self) -> Text:
...
 
     def to_llm_message(self) -> Message:
+        """Convert to LLM message format.
+
+        Returns only the static system prompt. The dynamic_context is handled
+        separately by to_llm_messages() to enable proper prompt caching.
+        """
         return Message(role="system", content=[self.system_prompt])
 
+    def to_llm_messages(self) -> list[Message]:
```
xingyaoww: Have two `to_llm_message*` seems very confusing. Shall we deprecate ` def to_llm_message(self) -> Message:` too?

openhands-sdk/openhands/sdk/event/base.py
```diff
@@ -118,11 +130,38 @@ def events_to_messages(events: list["LLMConvertibleEvent"]) -> list[Message]:
                 # Create combined message for the response
                 messages.append(_combine_action_events(batch_events))
                 i = j
+            elif isinstance(event, SystemPromptEvent):
+                # Add only the system message (static prompt, cacheable)
+                messages.append(event.to_llm_message())
```
xingyaoww: Shall we use `.to_llm_messages` here

openhands-sdk/openhands/sdk/event/base.py
```diff
@@ -118,11 +130,38 @@ def events_to_messages(events: list["LLMConvertibleEvent"]) -> list[Message]:
...
+                # Store dynamic context to merge with next user message
+                if event.dynamic_context:
+                    pending_dynamic_context = event.dynamic_context
```
xingyaoww: This seems a bit... complex.

Can we just set cache breakpoint accordingly to do this, instead of prepend dynamic context to user message?

https://platform.claude.com/docs/en/build-with-claude/prompt-caching#when-to-use-multiple-breakpoints

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
enyst: cc: @csmith49 

I wonder is there some other way we could do this? I hope maybe we don’t need to go back to a relationship that is no longer 1:1 between LLMConvertibleEvent and messages sent to the LLM… or am I missing something?

We had a fairly difficult time with the mental model and codebase in V0, where events weren’t 1:1 with messages 🙏

csmith49: We've already got a many-to-one relationship between events and messages, don't we? The `LLMConvertibleEvent.events_to_messages` function merges action events from the same LLM response into a single message.

So maybe it's not so bad to also have a one-to-many relationship. Agreed with Xingyao that we should keep the interface consistent, and that probably means always using `to_llm_messages: Self -> list[Message]` instead of the singular case.

csmith49:
> Maybe like
> 
> * the `SystemPromptEvent` has only the static part
> * a new event `SystemPromptSuffixEvent` could have the dynamic part, or is this a terrible idea?
> 
> Or maybe we could add 2 TextContents for the LLM message, and one has cache marker and the other doesn’t, as Xingyao suggested?

If it's not possible to set the cache markers for some reason, I like the solution that adds a new event type to manage the dynamic context over modifying incoming messages.

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
enyst: Those are from parallel tool calls I think... 🤔 

@openhands Investigate the codebase in detail. Look at LLMConvertibleEvents and the messages sent to the LLM. Look at all execution paths. Consider what we SEND to the LLM, that direction.

Q1. Is it events to sent messages an 1:1 relationship, or what is it and in what cases and how exactly does that break down?

Q2. WDYT about the problem this PR aims to solve and its solution on system prompt event?

Important: the mental model and the design invariant(s) of events and messages matter.

Respond directly to this comment on github. Note that it will be rendered as markdown.

enyst:
OpenHands-GPT-5.2 here.

Quick read on Anthropic’s **Messages API + system prompt caching** and a minimal change path:

- **System is an array of content blocks** in Anthropic (`system: [ {type: "text", ...}, ... ]`).
- **`cache_control` is a *prefix breakpoint***: the cache applies to `tools → system → messages` up to and including the block carrying `cache_control`.
- Therefore you *can* do exactly what you want **within the system prompt**:
  - **Block 1 (static)**: `cache_control: {type: "ephemeral"}`
  - **Block 2 (dynamic)**: no cache marker
  - Result: only the static prefix is cached; the dynamic block stays uncached even though it’s still in `system`.

**Minimal change in our code to enable this** (no redesign needed):

1) Allow **system content to be multiple `TextContent` blocks** (static + dynamic in the same system Message).
2) **Don’t overwrite explicit cache markers** in `LLM._apply_prompt_caching()`:

```python
if messages and messages[0].role == "system":
    if not any(c.cache_prompt for c in messages[0].content):
        messages[0].content[-1].cache_prompt = True
```

This keeps the default behavior (auto-mark last system block) *unless* a caller explicitly marks the static block. Then the dynamic block remains unmarked.

**Tools**: Anthropic’s cache prefix includes tools before system. We currently don’t add `cache_control` to tools, but they’ll still be part of the cached prefix when the system block is marked. If we ever want explicit tool cache markers, that’d be a separate enhancement.

neubig:
I think this is still WIP so changing to draft for now

enyst:
@openhands Read all comments in this PR. Do a /codereview-roasted on it. Post your feedback directly on github, as a comment, instead of telling me the review, just post it. Note that it will be rendered as markdown.

enyst:
Taste Rating: 🟡 **Acceptable**

**[CRITICAL ISSUES]**
- [openhands-sdk/openhands/sdk/event/llm_convertible/system.py, Lines 75–87] **Data Structure / Layering**: `SystemPromptEvent.to_llm_message()` hard-codes `cache_prompt=True` on the static block regardless of whether prompt caching is enabled. That leaks Anthropic-specific `cache_control` into any provider when `function_calling_enabled` or `vision_enabled` forces list serialization, even if caching is disabled. This is the wrong layer. Fix by keeping `SystemPromptEvent` pure data and applying cache markers only inside `LLM._apply_prompt_caching()` when caching is active (e.g., detect two system blocks and mark the first).
- [openhands-sdk/openhands/sdk/agent/base.py, Lines 264–286] **Breaking Change Footgun**: `system_message` still returns the *full* prompt (static + dynamic) as a single string. Any external caller who keeps using this deprecated property silently bypasses the new cache-safe split and reintroduces the exact bug you’re trying to fix. Deprecation warning isn’t enough—either make `system_message` return only `static_system_message`, or emit a hard warning that caching is disabled when it’s accessed.

- [openhands-sdk/openhands/sdk/llm/llm.py, Lines 1088–1094] **Good taste**: `any(c.cache_prompt for c in messages[0].content)` accepts *any* cache marker anywhere. If a caller tags the wrong block, you silently accept a broken cache prefix. Prefer validating the expected structure: if there are two system blocks, enforce cache on block 0 (static) and none on block 1 (dynamic), otherwise mark block 0 by default.
- [openhands-sdk/openhands/sdk/event/llm_convertible/system.py, Lines 43–55] **Pragmatism / Security hygiene**: `visualize()` now dumps dynamic context explicitly. That context can include secrets, tokens, repo paths, etc. If this is shown in logs/UI, it’s a data leak. At minimum truncate harder and redact obvious secret fields; ideally hide dynamic context unless explicitly requested.

**[STYLE NOTES]**
- [openhands-sdk/openhands/sdk/agent/agent.py, Lines 184–201] The inline comment claims the dynamic context is “included as a second content block in the system message.” That’s only true if you stay on the `SystemPromptEvent` path. If you keep `system_message`, the split doesn’t happen. This comment oversells the invariant.

**VERDICT:** ❌ **Needs rework** — the cache-control logic is in the wrong layer and will leak provider-specific fields when caching is disabled.

**KEY INSIGHT:** Cache control is a *serialization concern*, not an event model concern—keep it in the LLM formatting layer or you’ll keep smuggling provider quirks into your core data structures.

simonrosenberg:
@openhands read the PR diff and the comments

Then apply those suggestions:

```
[CRITICAL ISSUES]

[openhands-sdk/openhands/sdk/event/llm_convertible/system.py, Lines 75–87] Data Structure / Layering: SystemPromptEvent.to_llm_message() hard-codes cache_prompt=True on the static block regardless of whether prompt caching is enabled. That leaks Anthropic-specific cache_control into any provider when function_calling_enabled or vision_enabled forces list serialization, even if caching is disabled. This is the wrong layer. Fix by keeping SystemPromptEvent pure data and applying cache markers only inside LLM._apply_prompt_caching() when caching is active (e.g., detect two system blocks and mark the first).
[openhands-sdk/openhands/sdk/agent/base.py, Lines 264–286] Breaking Change Footgun: system_message still returns the full prompt (static + dynamic) as a single string. Any external caller who keeps using this deprecated property silently bypasses the new cache-safe split and reintroduces the exact bug you’re trying to fix. Deprecation warning isn’t enough—either make system_message return only static_system_message, or emit a hard warning that caching is disabled when it’s accessed.
[IMPROVEMENT OPPORTUNITIES]

[openhands-sdk/openhands/sdk/llm/llm.py, Lines 1088–1094] Good taste: any(c.cache_prompt for c in messages[0].content) accepts any cache marker anywhere. If a caller tags the wrong block, you silently accept a broken cache prefix. Prefer validating the expected structure: if there are two system blocks, enforce cache on block 0 (static) and none on block 1 (dynamic), otherwise mark block 0 by default.
```

simonrosenberg:
Screenshot 2026-02-10 at 11 44 59
@enyst tested!

simonrosenberg:
@openhands
/codereview-roasted
read the PR diff and read the comments. Were the last comments addressed successfully? Should this PR be merged?

simonrosenberg:
@openhands please address those two issues

[CRITICAL ISSUES]
🔴 tests/sdk/event/test_dynamic_context_message_sequence.py (Lines 125-145) - Testing: Tests check default values, not actual caching behavior. They will pass even if cache markers are never set. Must add tests that call LLM._apply_prompt_caching() and verify correct marker placement.


🟠 tests/sdk/llm/test_prompt_caching_cross_conversation.py (Line 51) - Testing Gap: Test verifies static messages are identical (good!) but doesn't test end-to-end caching behavior through the full init_state() → events_to_messages() → caching flow. Add integration test.

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -222,18 +231,69 @@ def system_message(self) -> str:
...
+            return None
+        return self.agent_context.get_system_message_suffix(
+            llm_model=self.llm.model,
+            llm_model_canonical=self.llm.model_canonical_name,
+        )
+
+    @property
+    @deprecated(
+        deprecated_in="1.11.0",
+        removed_in="1.13.0",
```
enyst:
```suggestion
        removed_in="1.16.0",
```
Tiny nit: could we give it a little more time? It will be automatic anyway, and it's the system prompt, idk programmatic users may rely on it.

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -222,18 +231,69 @@ def system_message(self) -> str:
...
+            "and dynamic content into a single string."
+        ),
+    )
+    def system_message(self) -> str:
+        """Return the combined system message (static + dynamic).
+
+        .. deprecated:: 1.11.0
+            Use :attr:`static_system_message` for the cacheable system prompt and
+            :attr:`dynamic_context` for per-conversation content. This separation
+            enables cross-conversation prompt caching. Will be removed in 1.13.0.
```
enyst:
```suggestion
            enables cross-conversation prompt caching. Will be removed in 1.16.0.
```

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
```diff
@@ -10,20 +10,48 @@
...
     system_prompt: TextContent = Field(..., description="The system prompt text")
     tools: list[ToolDefinition] = Field(
         ..., description="List of tools as ToolDefinition objects"
     )
+    dynamic_context: TextContent | None = Field(
+        default=None,
+        description=(
+            "Optional dynamic per-conversation context (runtime info, repo context, "
+            "secrets). When provided, this is included as a second content block in "
+            "the system message to enable cross-conversation prompt caching."
```
enyst:
```suggestion
            "the system message (not cached)."
```

tests/sdk/event/test_dynamic_context_message_sequence.py
```diff
@@ -0,0 +1,271 @@
...
+
+    # User message should only contain the user's text
+    user_content_texts = [
+        c.text for c in messages[1].content if isinstance(c, TextContent)
+    ]
+    assert user_content_texts == ["Hi"]
+
+
+def test_to_llm_message_does_not_set_cache_markers():
+    """SystemPromptEvent.to_llm_message() should NOT set cache markers.
```
enyst: I think this was happening before, too, wasn't it? Maybe it's not so useful to test newly, I'm pretty sure we have tests for prompt caching

tests/sdk/event/test_dynamic_context_message_sequence.py
```diff
@@ -0,0 +1,271 @@
...
+    assert "Working directory" in sys_msg.content[1].text
+
+    # User message should be unmodified
+    user_msg = messages[1]
+    assert len(user_msg.content) == 1
+    assert isinstance(user_msg.content[0], TextContent)
+    assert user_msg.content[0].text == "Hi"
+
+
+def test_system_prompt_without_dynamic_context_followed_by_user():
```
enyst: I think maybe this test is not very useful... it's a tiny bit unique maybe, but idk

tests/sdk/event/test_dynamic_context_message_sequence.py
```diff
@@ -0,0 +1,271 @@
...
+
+from typing import cast
+
+from openhands.sdk.event.base import LLMConvertibleEvent
+from openhands.sdk.event.llm_convertible import MessageEvent, SystemPromptEvent
+from openhands.sdk.llm import Message, TextContent
+
+
+def test_dynamic_context_produces_two_system_content_blocks():
+    """Dynamic context should appear as a second content block in the system message."""
```
enyst: I think maybe this is kinda the same with test_dynamic_context_content_preserved_in_system_message() or a part of it? Maybe we can keep only the second, or the agent can check and combine if anything is unique here?

tests/sdk/event/test_dynamic_context_message_sequence.py
```diff
@@ -0,0 +1,271 @@
...
+    assert msg.content[0].cache_prompt is False  # Default value, not set
+
+    assert isinstance(msg.content[1], TextContent)
+    assert msg.content[1].text == "Dynamic context"
+    assert msg.content[1].cache_prompt is False  # Default value, not set
+
+
+def test_to_llm_message_single_block_no_cache_marker():
+    """Single block should also not have cache marker set by to_llm_message()."""
+    system_event = SystemPromptEvent(
```
enyst: Same as above, this seems a tiny preexisting behavior, maybe we don't need this test so much

tests/sdk/llm/test_prompt_caching_cross_conversation.py
```diff
@@ -0,0 +1,406 @@
...
+
+    # User message's last content should also be marked
+    assert messages[1].content[-1].cache_prompt is True
+
+
+def test_end_to_end_caching_flow_with_init_state(tmp_path):
+    """Integration test: init_state() → events_to_messages() → caching flow.
+
+    This test verifies the complete end-to-end flow for cross-conversation
+    prompt caching:
```
enyst: The funny thing is, this test has comments like "simulating what happens before x", "simulating what x does", including `llm._apply_prompt_caching`. Idk, it doesn't look very "complete end-to-end" 😅 

How about we focus the agent on all tests, to make sure to combine them when related or remove redundancies, and try to make one good `cross` test? (or `sdk`, but more end-to-end -ish indeed)

tests/sdk/llm/test_prompt_caching_cross_conversation.py
```diff
@@ -0,0 +1,406 @@
...
+    assert len(messages) == 2
+    assert messages[0].role == "system"
+    assert len(messages[0].content) == 1  # Only static block
+
+    # Apply prompt caching
+    llm._apply_prompt_caching(messages)
+
+    # Verify cache marker is applied to the single block
+    assert messages[0].content[0].cache_prompt is True, (
+        "Single static system prompt should be marked for caching"
```
enyst: I feel like maybe I read this test above too...

enyst:
It looks good to me! I think there was a `Claude` in there that is over-verbose on tests that are maybe redundant, that's not a blocker though.

enyst:
@openhands Review all tests in this PR. Feel free to look at the inline unsolved comments with `gh` *on tests*. Goal: simplify tests, reduce redundancies, combine them so that we know we test the right thing, and test each thing once.

simonrosenberg:
@enyst thanks!!

Issue #1894, opened 2026-02-04
Responses API drops `file_editor.view` images (tool `ImageContent` not delivered to model)


Wangmerlyn:
## Summary
  When the agent views an image via `file_editor` and the LLM uses the OpenAI Responses API, the tool's `ImageContent` is dropped during
  serialization. The model never receives the image (no `input_image` / `image_url` in payload), so it says it “can’t see” the picture.

  ## Steps to Reproduce
  1. Use a vision-capable OpenAI model in Responses mode.
  2. Call `file_editor.view` on an image file (png/jpg/etc.).
  3. Inspect the serialized payload (`LLM_PAYLOAD_LOG_PATH` or logging): `content` contains the `data:` URL, but `output_items` lacks any
  `input_image`/`image_url`.

  ## Expected
  Tool observations containing images should be preserved and sent to the model in the Responses payload.

  ## Actual
  Images are stripped; only text outputs are sent. As Tony Horton in slack channel mentioned, agents do not say they had an issue accessing it, they just make something up.

  ## Impact
  - Agents using Responses API cannot perceive images returned by `file_editor.view`.
  - Can lead to incorrect behaviors and wasted tokens/debug time.

  ## Environment
  - Branch: `main` (latest main 6ce3ba51)
  - Affected file: `openhands-sdk/openhands/sdk/llm/message.py` (tool serialization in Responses path)

PR #1895, opened 2026-02-04
fix(message): Responses tool image serialization (preserve file_editor images)

Wangmerlyn:
## Summary
  Fixes the Responses API path dropping tool images. `file_editor.view` observations now deliver images to the model instead of stripping them. related to Issue #1894 

  ## What changed
  - `Message.to_responses_dict` (tool role):
    - When vision is enabled and `ImageContent` is present, emit `message` items with `input_image` entries (matching Responses API
  expectations).
    - Emit a minimal `function_call_output` alongside to keep call_id threading (no base64 stuffed into text).
  - Leaves `TextContent -> function_call_output` unchanged.

  ## Rationale
  Previously, tool-role serialization only sent `TextContent`; `ImageContent` was discarded, so models never saw images in Responses mode.
  This aligns tool image handling with how user-role images are serialized (input_image).

## Checklist

### DIFF
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -1,5 +1,7 @@
 ---
 name: Run Integration Tests
+run-name: >-
+    Run Integration Tests ${{ inputs.reason || github.event.label.name || 'scheduled' }}
 
 on:
     # Use pull_request_target to access secrets even on fork PRs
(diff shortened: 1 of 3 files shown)


openhands-sdk/openhands/sdk/llm/message.py
```diff
@@ -558,6 +559,39 @@ def to_responses_dict(self, *, vision_enabled: bool) -> list[dict[str, Any]]:
...
                         )
+                    elif isinstance(c, ImageContent) and vision_enabled:
+                        for url in c.image_urls:
+                            # Send image via input_image message to keep binary in the
+                            # right field.
+                            items.append(
+                                {
+                                    "type": "message",
+                                    "role": "user",
+                                    "content": [
```
enyst: I wonder, should this really be a new user message, or directly part of `items`, with `type=input_image`?

enyst: @openhands Understand the Responses API from openai docs, modify the reasoning example you can find in examples/ to accept an image (make yourself a screenshot and use that as the image, you can install yourself what you need) and run it. Use openai key and remote gh key to do this and verify the API calls work with an image.

Post your result directly on github, note that it will be rendered as markdown

Wangmerlyn: > I wonder, should this really be a new user message, or directly part of `items`, with `type=input_image`?

Hi Enyst, thank you for the timely review. I chose this design to put the image data url under the role user simply because that I looked into the [openai response api docs for function call output item](https://platform.openai.com/docs/api-reference/realtime-server-events/response/output_item/added#realtime_server_events-response-output_item-added-item-realtime_function_call_output_item-output) and it says that the "output is free text and can contain any information or simply be empty."

I agree that it's much better to somehow put the image data url under the role `tool` like how we usually do in the chat completion api. But I can't find a good solution.

I tried something like
```python 
                          items.append(
                                  {
                                      "type": "function_call_output",
                                      "call_id": resp_call_id,
                                      "output": json.dumps(
                                          {
                                              "type": "image_url",
                                              "image_url": url,
                                          }
                                      ),
                                  }"
```
But it just feed the whole image data url as text into the agent. So I don't really have a solution for this and I chose to use the User role for a workaround.

Wangmerlyn: Oh, sorry Enyst, it seems that I have a misunderstanding with the response api.  I had looked into the wrong place for response api docus. [Real openai response api docs for function call output item](https://platform.openai.com/docs/api-reference/responses/list) this is the correct documentation. I now have fixed the implementation to use `function_call_output` for image url.

Wangmerlyn:
> Thank you very much! This looks great to me!
> 
> Do you think you could show a screenshot or some other way to see that it works when sent to the LLM from file editor or another tool? Just a real-world test, so we see.
> 
> Maybe we could also add a unit test or a modification of a relevant unit test to show that it works?

Thank you, Engel. I can show a quick test result using one of my random photos to show that it works for the LLM to 'see' from the file_editor after the fix. (before/after fix comparison).

Later I'll see if I can add a relevant test case for this fix. Do you want the test case to be end to end running in the agent workflow or just a simple test to confirm that the image url part is in the responses api message dict?

Wangmerlyn:
This is the image I used to testing if the model is able to "see" the image through the `ImageContent` provided by `file_editor`. I used the model gpt-5 as the backend for the agent. I asked the agent to view the image by the tool `file_editor` and describe what it sees.
image used for testing

This is the final response before fixing, as you can see, the agent response is either full of hallucination or the agent says that it is not able to see the image, both are results of being told there is an image but in fact there is not.
image

image

This is after the bug fix, as you can see, the agent now correct describes what's in the picture. 

after fix, correct answer

enyst:
@openhands Understand this PR and read all top level comments on it.

Look at the integration tests, I think we have at least one that is about images. Look if we can modify it to add what this PR is exercising: an image that goes to the LLM via `file_editor` tool. We hadn’t seen the bug because we weren’t testing for tool use with the image in the tool, I suppose? If we can add an image there, open an issue with an implementation plan or code snippet how we’d do that. If not, open an issue to make a new test maybe, or what suggestions do you have.

PR #1942, opened 2026-02-06
fix: handle null secrets in conversation polling to prevent ValidationError

MkDev11:
## Summary

Fix conversation polling failure caused by Pydantic `ValidationError` when secrets contain null values after masking (missing cipher or key rotation).

**Root cause:** After secrets are masked, the state contains `{"secrets": {"GITHUB_TOKEN": {"value": null}}}`. The current code doesn't handle this, causing 500 errors when the frontend polls `/api/conversations` or loads a stored conversation.

**Changes:**
- `_compose_conversation_info()`: Add `exclude_none=True` to `state.model_dump()` to filter null values during re-validation
- `_filter_invalid_secrets()`: New helper to remove secrets with null values from parsed dicts
- `_preprocess_stored_conversation_json()`: New helper to clean JSON before `StoredConversation.model_validate_json()`
- `filter_null_secrets`: New `@field_validator` on `StartConversationRequest.secrets` as a model-level safety net
- Applied preprocessing in `conversation_service.__aenter__()` and `event_service.load_meta()`

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

- [ ] I have read and reviewed the code and I understand what the code is doing.
- [ ] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves OpenHands/OpenHands#12714

### DIFF
--- a/.github/scripts/check_sdk_api_breakage.py
+++ b/.github/scripts/check_sdk_api_breakage.py
@@ -53,6 +53,11 @@ PACKAGES: tuple[PackageConfig, ...] = (
         distribution="openhands-workspace",
         source_dir="openhands-workspace",
     ),
+    PackageConfig(
+        package="openhands.tools",
+        distribution="openhands-tools",
+        source_dir="openhands-tools",
+    ),
 )
 
 
--- a/openhands-sdk/openhands/sdk/AGENTS.md
+++ b/openhands-sdk/openhands/sdk/AGENTS.md
@@ -1,0 +1,32 @@
+# Repository Guidelines
+
+## Project Structure & Module Organization
+
+- This directory (`openhands-sdk/openhands/sdk/`) contains the core Python SDK under the `openhands.sdk.*` namespace.
+- Keep new modules within the closest existing subpackage (e.g., `llm/`, `tool/`, `event/`, `agent/`) and follow local naming patterns.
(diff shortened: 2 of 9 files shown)


openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -37,15 +39,71 @@ def _compose_conversation_info(
     # Use mode='json' so SecretStr in nested structures (e.g. LookupSecret.headers,
     # agent.agent_context.secrets) serialize to strings. Without it, validation
     # fails because ConversationInfo expects dict[str, str] but receives SecretStr.
+    # Use exclude_none=True to filter out masked/null secret values that would
+    # cause ValidationError during re-validation (issue #12714).
     return ConversationInfo(
-        **state.model_dump(mode="json"),
+        **state.model_dump(mode="json", exclude_none=True),
```
tofarr: Are we sure this works? I imagine that there could be some fields that permit null but do not have a default value. (And would therefore complain if the null value was omitted)

mkdev5: Fixed!

mkdev5:
> The fix addresses the null secrets issue but has several concerns: code duplication across 3 filtering mechanisms, overly broad `exclude_none=True` that may hide other issues, and mutation + return pattern in helper function. Consider consolidating the filtering logic and being more surgical about which fields to exclude.

All the comments have fixed!

mkdev5:
no comments here

Issue #1964, opened 2026-02-07
fix(auth): create_llm() should handle usage_id to avoid registry conflicts


enyst:
## Summary

When using `OpenAISubscriptionAuth.create_llm()` multiple times (e.g., for main agent LLM and condenser LLM), the `LLMRegistry` raises a `ValueError` because both LLMs get the same default `usage_id` of `"default"`.

## Error Message

```
ValueError: Usage ID 'default' already exists in registry. Use a different usage_id on the LLM or call get() to retrieve the existing LLM.
```

## Root Cause

1. `LLM.usage_id` defaults to `"default"` (see `llm.py:313-320`)
2. `create_llm()` passes `**llm_kwargs` to `LLM()` but doesn't set a default `usage_id`
3. When the CLI creates two LLMs via `create_llm()` (one for agent, one for condenser), both get `usage_id="default"`
4. `LLMRegistry.add()` raises `ValueError` on duplicate `usage_id`

## Current Workaround

Callers must explicitly pass `usage_id`:

```python
# Main LLM
llm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_aute `create_llm(llm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_aute `create_llm(llm = chate_llm(self, model: str = "gpllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt_autllm = chatgpt().hex[:8]}"
    # ... rest    # ... rest    # ... rest    # ... rest    # ...meter

Make `usage_id` a required parameter for `create_llm()`:

```python
def create_llm(
    self,
    model: str = "gpt-5.2-codex",
    usage_id: str,  # Required
    credentials: OAuthCredentials | None = None,
    ...
) -> LLM:
```

### Option 3: Document the requirement

Add clear Adcumentation that callAdd clear Adcumentation that e_id` values when creating multiple LLMs.

## Additional Context

The `usage_id` is used for:
[...]

PR #1975, opened 2026-02-09
fix(auth): auto-generate unique usage_id in create_llm() (#1964)

timon0305:
## Summary

When using OpenAISubscriptionAuth.create_llm() multiple times without explicit usage_id, the LLMRegistry raises ValueError because both LLMs get the default usage_id of "default".

Changes:
- create_llm() now auto-generates unique usage_id if not provided (format: chatgpt-subscription-{8_hex_chars})
- Added generate_usage_id() utility function in llm.py for consistent unique ID generation across the codebase
- Added has() and get_or_none() methods to LLMRegistry for safer registry operations
- Exported generate_usage_id from openhands.sdk.llm package
## Checklist


enyst (commented):
Thank you for this PR, timon! I apologize for the delay, I’ve been thinking about this and I’m not sure it’s the right fix. Does it happen for subscription?

I would love to know if @VascoSch92 has thoughts, because I think maybe we will meet this problem with various profiles switches too…

tests/sdk/llm/test_llm.py
```diff
@@ -993,4 +993,52 @@ def test_llm_respects_allow_short_context_windows_env_var(mock_get_model_info):
...
+
+def test_generate_usage_id_custom_prefix():
+    """Test generate_usage_id with custom prefix."""
+    usage_id = generate_usage_id("my-agent")
+    assert usage_id.startswith("my-agent-")
+    suffix = usage_id.split("-")[-1]
+    assert len(suffix) == 8
+
+
+def test_generate_usage_id_uniqueness():
```
VascoSch92, line 1018: This test is not deterministic. 

I don't know if we want to include that in the test suite.

tests/sdk/llm/test_llm.py
```diff
@@ -993,4 +993,52 @@ def test_llm_respects_allow_short_context_windows_env_var(mock_get_model_info):
...
+    # Should be valid hex
+    int(suffix, 16)
+
+
+def test_generate_usage_id_custom_prefix():
+    """Test generate_usage_id with custom prefix."""
+    usage_id = generate_usage_id("my-agent")
+    assert usage_id.startswith("my-agent-")
+    suffix = usage_id.split("-")[-1]
+    assert len(suffix) == 8
```
VascoSch92, line 1015: nit: put the two tests together using `@pytest.mark.parametrize` and then you could also check for some corner case...

tests/sdk/llm/test_llm.py
```diff
@@ -993,4 +993,52 @@ def test_llm_respects_allow_short_context_windows_env_var(mock_get_model_info):
         assert llm.max_input_tokens == 2048
 
 
+# generate_usage_id Tests
```
VascoSch92, line 996:
```suggestion
```

tests/sdk/llm/test_llm.py
```diff
@@ -993,4 +993,52 @@ def test_llm_respects_allow_short_context_windows_env_var(mock_get_model_info):
...
+
+def test_generate_usage_id_default_prefix():
+    """Test generate_usage_id with default prefix."""
+    usage_id = generate_usage_id()
+    assert usage_id.startswith("llm-")
+    # Should have 8 hex characters after prefix
+    suffix = usage_id.split("-", 1)[1]
+    assert len(suffix) == 8
+    # Should be valid hex
+    int(suffix, 16)
```
VascoSch92, line 1007:
```suggestion
    assert isinstance(int(suffix, 16), int)
```

VascoSch92 (commented):
For instance, I don't think we've encountered a situation where this specific problem occurs.

Do we have any documented cases where this caused an error? I’d like to understand what kind of fix is needed—or if one is even necessary.

Aside from that, I’ve left a couple of comments.

PR #1997, opened 2026-02-11
fix(llm): ensure second LLM gets independent metrics and telemetry

### LINKED ISSUE #418: Condensation breaks the metrics accumulation
After the condensation:
- The metrics show a wrong metric value in the "Condensation" cell block
- Then the next cell after condensation seems to work

We should fix this, this is probably because the `metrics` attribute for `Condesation` event is not using the latest `.metrics` from the LLM. Or maybe we should consider merging the metrics?

Once we fix it, we should add a test to make sure this doesn't happen again.

Image

xingyaoww:
## Summary

This PR fixes issue #418 where condensation breaks the metrics accumulation.

### Root Cause
When `LLM.model_copy()` was called (e.g., to create a condenser LLM from an agent LLM), the `_metrics` and `_telemetry` private attributes were shared between the original and copied LLM instances. This is because Pydantic's `model_copy()` does a shallow copy of private attributes by default.

This caused metrics to be double-counted when both LLMs were used, as they were updating the same `Metrics` object. The visualizer would then display incorrect metrics in the "Condensation" cell block.

### Fix
This PR overrides `model_copy()` in the `LLM` class to create fresh `Metrics` and `Telemetry` instances for the copied LLM, ensuring each LLM tracks its own usage independently.

### Testing
Added 3 new tests to verify:
1. `test_llm_model_copy_creates_independent_metrics` - Verifies that copied LLMs have separate metrics objects
2. `test_llm_model_copy_creates_independent_telemetry` - Verifies that copied LLMs have separate telemetry objects
3. `test_llm_model_copy_preserves_configuration` - Verifies that the LLM configuration is preserved when copying

Fixes #418

## Checklist

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/05ea0e88b0c74151a671c9de6d2c4e7d)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:b05e9e0-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-b05e9e0-python \
  ghcr.io/openhands/agent-server:b05e9e0-python
```

```
ghcr.io/openhands/agent-server:b05e9e0-golang-amd64
ghcr.io/openhands/agent-server:b05e9e0-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:b05e9e0-golang-arm64
ghcr.io/openhands/agent-server:b05e9e0-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:b05e9e0-java-amd64
ghcr.io/openhands/agent-server:b05e9e0-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:b05e9e0-java-arm64
ghcr.io/openhands/agent-server:b05e9e0-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:b05e9e0-python-amd64
ghcr.io/openhands/agent-server:b05e9e0-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:b05e9e0-python-arm64
ghcr.io/openhands/agent-server:b05e9e0-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:b05e9e0-golang
ghcr.io/openhands/agent-server:b05e9e0-java
ghcr.io/openhands/agent-server:b05e9e0-python
```

- Each variant tag (e.g., `b05e9e0-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `b05e9e0-python-amd64`) are also available if needed

### DIFF
--- a/.github/workflows/pr-review-by-openhands.yml
+++ b/.github/workflows/pr-review-by-openhands.yml
@@ -8,7 +8,8 @@ on:
     #   2. A draft PR is marked as ready for review, OR
     #   3. A maintainer adds the 'review-this' label, OR
     #   4. A maintainer requests openhands-agent or all-hands-bot as a reviewer
-    # Only users with write access can add labels or request reviews, ensuring security.
+    # Adding labels and requesting new reviewers requires write access. GitHub may also allow PR authors
+    # to re-request review from a previous reviewer.
     # The PR code is explicitly checked out for review, but secrets are only accessible
     # because the workflow runs in the base repository context
     pull_request_target:
@@ -22,14 +23,14 @@ permissions:
 jobs:
     pr-review:
         # Run when one of the following conditions is met:
-        #   1. A new non-draft PR is opened by a trusted contributor, OR
-        #   2. A draft PR is converted to ready for review by a trusted contributor, OR
+        #   1. A new non-draft PR is opened by a non-first-time contributor, OR
+        #   2. A draft PR is converted to ready for review by a non-first-time contributor, OR
         #   3. 'review-this' label is added, OR
         #   4. openhands-agent or all-hands-bot is requested as a reviewer
         # Note: FIRST_TIME_CONTRIBUTOR PRs require manual trigger via label/reviewer request
         if: |
-            (github.event.action == 'opened' && github.event.pull_request.draft == false && github.event.pull_request.author_association != 'FIRST_TIME_CONTRIBUTOR') ||
-            (github.event.action == 'ready_for_review' && github.event.pull_request.author_association != 'FIRST_TIME_CONTRIBUTOR') ||
+            (github.event.action == 'opened' && github.event.pull_request.draft == false && github.event.pull_request.author_association != 'FIRST_TIME_CONTRIBUTOR' && github.event.pull_request.author_association != 'NONE') ||
+            (github.event.action == 'ready_for_review' && github.event.pull_request.author_association != 'FIRST_TIME_CONTRIBUTOR' && github.event.pull_request.author_association != 'NONE') ||
             github.event.label.name == 'review-this' ||
             github.event.requested_reviewer.login == 'openhands-agent' ||
             github.event.requested_reviewer.login == 'all-hands-bot'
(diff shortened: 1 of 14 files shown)


openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -541,6 +541,45 @@ def restore_metrics(self, metrics: Metrics) -> None:
         # Only used by ConversationStats to seed metrics
         self._metrics = metrics
 
+    def model_copy(
```
xingyaoww: @OpenHands review this issue, can we do NOT overwrite model_copy?

enyst:
Could we use the `LLMRegistry` instead? I think maybe the registry should be responsible with this

xingyaoww:
@OpenHands can you think about this https://github.com/OpenHands/software-agent-sdk/pull/1997#issuecomment-3884584069 and come up with a better solution for this?

xingyaoww:
@OpenHands can you review diff against main, understand the purpose of this PR, then help me reproduce the original issue, and I believe this PR contains a test case that will reproduce the original issue.Can you try to first Git reset to the original commit where no fix is applied,Make sure the test failed,and then you can reset your Git commit to the latest state of this PR and then rerun the same test again.Make sure that this PR actually fixed the original issue

enyst:
Ugh, OH fights with itself, and the last comment sounds quite bad, if it's really the case. 😓 

Please let me take a step back here. I feel I have a question:
- why are we using model_copy or copy in the first place?
- these are independent LLM instances, after all
- because we don't want to force the user to enter another api key, so we'd like to default them to the same LLM profile as they have already defined.

If so, the way forward here is to load the default profile for 'different_usage_id' as long as the user hasn't specified another. That has to pass through the LLMRegistry, it seems to me. And that won't be a model_copy, just initialize a new LLM from some profile. 🤔

xingyaoww:
@OpenHands, please clone OpenHands/Openhands-CLI and OpenHands/OpenHands, and search for all occurrences of model_copy for the LLM class (including in the current repo) -- Tell me, how will this PR impact that code? Will it introduce some weird behaviors?

Issue #1885, opened 2026-02-03
Add the ability to specify a fallback language model


neubig:
Once we have LLM profiles implemented: https://github.com/OpenHands/software-agent-sdk/pull/1882

we should add the ability to specify a primary LLM profile and optional fallback profiles to query when LLM calls to the primary profile fail. 

This will allow us to continue having the agent work when a particular LLM provider goes down.

raymyers:
Linking LiteLLM docs on fallbacks for reference on options.

https://docs.litellm.ai/docs/proxy/reliability

VascoSch92:
Hey,

I'm looking into this and I think I have a solution.

However, there are different types of fallback strategies. I see at least two possibilities, each targeting a different behavior:

A) If a given LLM fails, you fallback to the next LLM in the list and stick with it for the remainder of the conversation. 

B) If the primary LLM fails, you switch to the next one on the list, but revert to the primary LLM for the next user message or interaction.

Which one do you think makes more sense? Personally, I will go for Option A, but this would mean the fallback LLM is used for the entire duration of the session.

neubig:
Hi @VascoSch92, great question. I feel that maybe B is a better choice so it recovers when the LM becomes available again.

enyst:
FWIW litellm proxy works like Option B, I believe. It seems an easy mental model to me too, it's a fallback for transitory stuff that may not happen next request.

VascoSch92:
Yes... i have a solution which follows option B (just need to test it) and use the profile store to load persisted profiles. Perhaps there are users which would like to have an in-memory solution, i.e, you pass directly the LLM profile you want to use as a fallback. But we can always update the solution to support in-memory profile later.

VascoSch92:
FYI: I implemented another solution for the fallback logic which I believe is what we want.

PR #1998, opened 2026-02-11
feat(sdk): add fallback strategy (`Agent`)

VascoSch92:
## Summary

This PR addresses issue #1885.

It adds a fallback strategy for model selection. Fallback models are retrieved from the profile store, and the strategy follows approach B) as described [here](https://www.google.com/search?q=link).

You can choose a fallback LLM based on the specific error raised or simply provide a list of fallback LLMs.

Note: We currently do not support in-memory LLM profiles.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:d12ef3e-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-d12ef3e-python \
  ghcr.io/openhands/agent-server:d12ef3e-python
```

```
ghcr.io/openhands/agent-server:d12ef3e-golang-amd64
ghcr.io/openhands/agent-server:d12ef3e-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:d12ef3e-golang-arm64
ghcr.io/openhands/agent-server:d12ef3e-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:d12ef3e-java-amd64
ghcr.io/openhands/agent-server:d12ef3e-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:d12ef3e-java-arm64
ghcr.io/openhands/agent-server:d12ef3e-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:d12ef3e-python-amd64
ghcr.io/openhands/agent-server:d12ef3e-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:d12ef3e-python-arm64
ghcr.io/openhands/agent-server:d12ef3e-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:d12ef3e-golang
ghcr.io/openhands/agent-server:d12ef3e-java
ghcr.io/openhands/agent-server:d12ef3e-python
```

- Each variant tag (e.g., `d12ef3e-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `d12ef3e-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -418,14 +418,6 @@ class LocalConversation(BaseConversation):
             for llm in list(self.agent.get_all_llms()):
                 self.llm_registry.add(llm)
 
-            # Wire lazy-loading callback for FallbackRouter so that
-            # lazily-loaded fallback LLMs get registered in the LLMRegistry
-            # and picked up by ConversationStats.
-            from openhands.sdk.agent.base import FallbackStrategy
-
-            if isinstance(self.agent.llm, FallbackStrategy):
-                self.agent.llm.set_on_llm_created(self.llm_registry.add)
-
             # Wire lazy-loading callback for agent-level FallbackStrategy
             # so fallback LLMs loaded via get_fallback_llms() are also
             # registered in the LLMRegistry and tracked by ConversationStats.


openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -371,6 +370,52 @@ def step(
...
+                remaining_msg = (
+                    f"{remaining} fallback(s) remaining."
+                    if remaining
+                    else "No more fallbacks."
+                )
+                logger.warning(
+                    f"Fallback LLM {fallback} failed with "
+                    f"{type(exc).__name__}: {exc}. {remaining_msg}",
+                )
+
```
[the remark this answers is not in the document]
VascoSch92: adressed

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -371,6 +370,52 @@ def step(
...
+        raise last_exc
```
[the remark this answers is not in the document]
VascoSch92: ✅

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -43,6 +46,57 @@
...
+        Returns:
+            list[str]: A list of model identifiers to attempt. Returns an empty
+                list if the error type is not mapped and no defaults exist.
+        """
+        if error is None:
+            return self.default_fallbacks
+
+        fallback = self.fallback_mapping.get(type(error), None)
+        if fallback is None:
+            return self.default_fallbacks
```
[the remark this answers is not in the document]
VascoSch92: ✅

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -43,6 +46,57 @@
...
+        default_factory=list,
+    )
+    fallback_mapping: dict[type[Exception], list[str]] = Field(
+        description="Mapping from exception types to fallback profile names.",
+        default_factory=dict,
+    )
+
+    profile_store_path: Path | str | None = None
+    """Path to a directory where profiles are stored."""
+
```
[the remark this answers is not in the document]
VascoSch92: If you pass `None` to `LLMProfileStore`, the default profile directoy is used, i.e., `.openhands/profiles`

I believe this is not a problem.

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -43,6 +46,57 @@
...
+    """
+    Manages recovery logic for LLM failures by mapping errors to fallback models.
+
+    This strategy allows for granular control over how the SDK responds to different
+    failure modes. It can provide a generic list of fallback models or specific
+    models tailored for particular exception (e.g., switching to a model
+    with a larger context window when a ContextLengthError occurs).
+    """
+
+    default_fallbacks: list[str] = Field(
```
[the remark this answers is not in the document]
VascoSch92: I wanted to be really granular in the fallback strategy. But happy to change if someone has another point-of-view

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -44,6 +46,53 @@
...
+        description="List of default fallbacks to try when the primary LLM fails.",
+        default_factory=list,
+    )
+    fallback_mapping: dict[type[Exception], list[str]] = Field(
+        description="Mapping from exception types to fallback profile names.",
+        default_factory=dict,
+    )
+
+    profile_store_path: Path | str | None = None
```
xingyaoww: Maybe profile_store_dir?

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -68,6 +117,9 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
             }
         ],
     )
+    fallback_strategy: FallbackStrategy = Field(
```
xingyaoww: shall we name it llm_fallback_strategy?

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -378,6 +370,69 @@ def step(
...
+                    f"Fallback LLM {fallback_profiles[j]} failed with "
+                    f"{type(exc).__name__}: {exc}. {remaining_msg}",
+                )
+
+        pretty_exc = [f"{type(exc).__name__}: {exc}." for exc in exceptions]
+        raise Exception(
+            f"{len(fallback_profiles)} fallback(s) failed with exception(s):"
+            f" {'\n- '.join(pretty_exc)}"
+        )
+
```
xingyaoww, line 443: I wonder if we can move the logic where we load a profile into LLM into the FallbackStrategy class?

e.g., we can do:

```
strategy = FallbackStrategy()
fallback_llm = strategy.get_fallback_llm()
```

and then here we just `make_llm_completion(fallback_llm, ...)`

VascoSch92: Actually it should pretty easy to do ;) and it could be interesting in the future if we support in-memory profiles.

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -378,6 +378,66 @@ def step(
...
+        messages: list[Message],
+        tools: list,
+        on_token: ConversationTokenCallbackType | None = None,
+    ) -> LLMResponse:
+        """Attempt fallback LLMs when primary fails.
+
+        Raises LLMError with aggregated failure details if all fallbacks fail,
+        preserving the original exception in the chain.
+        """
+        fallback_profiles = self.llm_fallback_strategy.resolve(primary_exc)
```
enyst, line 407: Sorry if it's obvious, just reading for the first time: does `llm_fallback_strategy` have to be an Agent thing? The agent in V1 has so many responsibilities... it's... not clear that it also should resolve LLM fallback strategies to different LLM / profiles 🤔 

Just a thought

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -44,6 +47,69 @@
...
+    def get_fallback_llms(
+        self, error: Exception
+    ) -> Generator[tuple[str, LLM], None, None]:
+        """Yield ``(profile_name, llm)`` pairs for fallbacks matching *error*.
+
+        Profiles that fail to load are logged and skipped.
+        """
+        for name in self.resolve(error):
+            try:
+                yield name, self.profile_store.load(name)
```
enyst: Same as before... I wonder if maybe instead `AgentBase`, the responsible with loading profiles from the LLM store could be something else: `LLMRegistry` is the first that comes to mind, though for this particular method idk, could maybe the profile store itself give us the on-demand profile?

openhands-sdk/openhands/sdk/agent/agent.py
VascoSch92: When I started working on this, I considered incorporating the fallback strategy directly into the LLM class or creating a specific router for it.

The first option was not ideal, as the LLM class should strictly serve as an interface for API calls.

For the second option, the limitation is that only one router can be used at a time; for instance, you cannot have both a fallback strategy and another type of router simultaneously. I felt this wasn't the right approach, as fallbacks deserve a dedicated implementation.

Therefore, integrating the fallback logic into the Agent class became the natural choice.

openhands-sdk/openhands/sdk/agent/base.py
VascoSch92: What do you mean with on-demand profile?

enyst: Bad wording on my part, just the profile we need now, loaded on demand 
> Yield ``(profile_name, llm)`` pairs

I guess my question is, if we don't pass through the `LLMRegistry`, does that mean we created an untracked LLM, for which we have e.g. no stats?

openhands-sdk/openhands/sdk/agent/agent.py
enyst: Could you please tell, what do you mean by in-memory profiles?

enyst:
@openhands Use `gh` to pull the inline review threads, for context you can think about.

1. Do a /codereview-roasted on this PR.
2. Will the costs in `ConversationStats` be accurate?

Post your feedback directly on the PR as a comment. Note that it will be rendered as markdown.

openhands-sdk/openhands/sdk/agent/base.py
VascoSch92: > I guess my question is, if we don't pass through the LLMRegistry, does that mean we created an untracked LLM, for which we have e.g. no stats?

This is actually a good point.

Actually, I suppose that this is the case. Thanks to point this out.

enyst:
## /codereview-roasted 🔥 (PR #1998: fallback strategy)

**Taste rating:** 🟡 *Acceptable* (solves a real problem) but the current shape makes observability + API stability easy to accidentally break.

### [CRITICAL ISSUES] (must fix / strongly consider before merge)

1) **Fallback LLMs likely bypass usage tracking →  costs will be wrong**

- The fallback LLMs are created via  inside .
- Unless  internally *registers the LLM with the registry that  listens to*, those fallback instances won’t be tracked.
- Result: **token/cost metrics will only include the primary  metrics**, and **fallback costs won’t show up** (or will show up only if something else happens to register them).

This isn’t theoretical—you already hinted at it in the thread:
> “if we don't pass through the , does that mean we created an untracked LLM … no stats?” → **yes**.

**What I’d change:**
- Make sure fallback LLM creation goes through the same registry pathway as normal LLM instantiation.
  - Either: have  register created LLMs (recommended if profile store is the canonical factory)
  - Or: have  accept an  / registry handle so it can create+register.
  - Or: add an explicit /equivalent hook when loading fallbacks.

2) **Public API break risk: field rename**

 →  is a user-facing config rename.

If anyone is constructing agents from config / dicts, this is a silent break.

**Mitigation options:**
- Keep backwards compatibility with an alias (Pydantic  / ) and deprecate the old name.
- Or provide a migration path + release note (but SDKs should be gentle).

### [IMPROVEMENT OPPORTUNITIES] (good taste / maintainability)

1) **Data structure: mapping by  is too strict**

Right now:

This means mapping  won’t match .

You said you wanted granularity—fine—but then you need a story for “generic transient failure”.

**Suggestion:** support both:
- exact matches win
- otherwise  scan fallback

2) ** math is wrong / misleading**

You compute:

…but  is “LLM completion failures for loaded profiles”, not “profiles loaded successfully”.

So if **all profiles load fine but all completions fail**, you’ll still report “N profiles failed to load” which is nonsense.

**Fix:** count load failures explicitly inside  (return structured info) or yield a sentinel when load fails.

3) **Exception aggregation loses the real final exception type**

If you always raise  after exhausting fallbacks, callers can’t easily distinguish e.g. auth vs context vs rate limit after fallback attempts.

You *do* chain , which helps, but the “last failure” might be more relevant than the primary.

**Suggestion:**
- keep a  and raise  (or include last_exc type prominently)
- or raise the *last* exception and attach a structured summary on it.

4) **Agent responsibility creep**

This is the core “taste” issue:
- Agent now owns: prompt plumbing, tool dispatch, action execution, confirmation policy, security analysis glue, critic glue… and now **LLM routing and profile loading**.

It works, but it’s how you get a “god object” you can’t reason about.

**Better seam:**
- Put fallback strategy into an LLM wrapper/router that implements the same interface (like a composite ), so the Agent doesn’t care.
- If “only one router can exist” is currently a limitation, that’s the actual design constraint to fix (compose routers).

### [STYLE NOTES]

- Logging uses f-strings;  is generally preferred (avoids formatting cost when log level filtered), but not a blocker.

---

## Q: Will the costs in  be accurate?

**As implemented: probably not in the fallback case.**

-  tracks metrics keyed by  and restores metrics when it sees an LLM created via registry events.
- Fallback LLMs loaded from the profile store are *very likely* **not registered**, so their metrics won’t be merged into .

**So:** primary-only runs → accurate; fallback runs → undercounted (missing fallback usage/cost), unless registration happens elsewhere.

---

If you want the minimal “ship it” fix: ensure every fallback LLM loaded is registered the same way as normal LLM creation, and fix the load-failure accounting.

enyst:
## Follow-up: my previous comment got shell-mangled

I accidentally posted the prior review through a shell command without proper escaping, so most Markdown got interpreted by bash (oops). This comment contains the intended content.

---

## /codereview-roasted 🔥 (PR #1998: fallback strategy)

**Taste rating:** 🟡 *Acceptable* (solves a real problem) but the current shape makes observability + API stability easy to accidentally break.

### [CRITICAL ISSUES] (must fix / strongly consider before merge)

#### 1) Fallback LLMs likely bypass usage tracking → `ConversationStats` costs will be wrong

- The fallback LLMs are created via `LLMProfileStore.load(...)` inside `FallbackStrategy.get_fallback_llms()`.
- Unless `LLMProfileStore.load` internally **registers** the LLM with the registry that `ConversationStats.register_llm()` listens to, those fallback instances won’t be tracked.
- Result: **token/cost metrics will only include the primary `llm.usage_id` metrics**, and **fallback costs won’t show up** (or will show up only if something else registers them).

This isn’t theoretical—you already discussed it in the thread:
> “if we don't pass through the `LLMRegistry`, does that mean we created an untracked LLM … no stats?” → **yes**.

**What I’d change (pick one):**
- Have `LLMProfileStore.load()` register created LLMs (recommended if profile store is the canonical factory)
- Or: have `FallbackStrategy` accept an `LLMFactory` / registry handle so it can create+register.
- Or: add an explicit “register this llm” hook when loading fallbacks.

#### 2) Public API break risk: field rename

`AgentBase.fallback_strategy` → `llm_fallback_strategy` is a **user-facing config rename**.

If anyone is constructing agents from config / dicts, this is a silent break.

**Mitigation options:**
- Keep backwards compatibility with an alias (Pydantic `validation_alias` / `AliasChoices`) and deprecate the old name.
- Or provide a clear migration path + release note (but SDKs should be gentle).

---

### [IMPROVEMENT OPPORTUNITIES] (good taste / maintainability)

#### 1) Data structure: mapping by `type(error)` is too strict

Right now:
```py
return self.fallback_mapping.get(type(error), self.default_fallbacks)
```
This means mapping `{LLMError: ["generic-fb"]}` won’t match `LLMRateLimitError()`.

You said you wanted granularity—fine—but then you need a story for “generic transient failure”.

**Suggestion:** support both:
- exact matches win
- otherwise `isinstance` scan fallback

#### 2) `profiles_failed_to_load` math is wrong / misleading

You compute:
```py
profiles_failed_to_load = len(fallback_profiles) - len(failed)
```
…but `failed` is “LLM completion failures for loaded profiles”, not “profiles loaded successfully”.

So if **all profiles load fine but all completions fail**, you’ll still report “N profiles failed to load” which is nonsense.

**Fix:** count load failures explicitly inside `get_fallback_llms` (return structured info), or yield a sentinel/record for load failures.

#### 3) Exception aggregation loses the final exception type

If you always raise `LLMError` after exhausting fallbacks, callers can’t easily distinguish auth vs context vs rate limit after fallback attempts.

You *do* chain `from primary_exc`, which helps, but the “last failure” might be more relevant than the primary.

**Suggestion:**
- track `last_exc` and raise `LLMError(... ) from last_exc` (or at least include last_exc type prominently)
- or raise the *last* exception and attach a structured summary.

#### 4) Agent responsibility creep

This is the main “taste” issue:
- Agent now owns: prompt plumbing, tool dispatch, action execution, confirmation policy, security glue, critic glue… and now **LLM routing + profile loading**.

It works, but it’s how you get a god object.

**Better seam:**
- Put fallback strategy into an LLM wrapper/router that implements the same interface (composite `LLM`), so the Agent doesn’t care.
- If “only one router can exist” is a limitation today, that’s the real architectural constraint to fix (compose routers).

---

### [STYLE NOTES]

- Logging uses f-strings; `logger.warning("... %s", x)` is generally preferred (avoid formatting cost when log level filtered), but not a blocker.

---

## Q: Will the costs in `ConversationStats` be accurate?

**As implemented: probably not in the fallback case.**

- `ConversationStats` tracks metrics keyed by `usage_id` and restores metrics when it sees an LLM created via registry events.
- Fallback LLMs loaded from the profile store are very likely **not registered**, so their metrics won’t be merged into `ConversationStats.usage_to_metrics`.

**So:** primary-only runs → accurate; fallback runs → undercounted (missing fallback usage/cost), unless registration happens elsewhere.

---

If you want the minimal “ship it” fix: ensure every fallback LLM loaded is registered the same way as normal LLM creation, and fix the load-failure accounting.

VascoSch92:
To provide a bit of context:

I explored a few different approaches. The most natural path seemed to be creating a FallbackRouter, but I found it wasn't an ideal solution because: 1) it prevents using a router and fallbacks simultaneously, and 2) it required overriding several methods in the base class.

We could rely on LiteLLM’s built-in fallback solution. However, this would increase our dependency on that package, and I’m not entirely sure how well it integrates with our current stats-tracking and router implementation (though I’m open to being corrected on that).

Additionally:
- I have updated the logic so that statistics are now correctly tracked.
- I added tests to verify that these stats update as expected.

If you'd like, I can also prepare a LiteLLM-based version of this solution so we can compare the two.

openhands-sdk/openhands/sdk/agent/agent.py
enyst: > When I started working on this, I considered incorporating the fallback strategy directly into the LLM class or creating a specific router for it.
> 
> The first option was not ideal, as the LLM class should strictly serve as an interface for API calls.

Just a thought, IMHO sooner or later we could look into splitting LLM into something like `LLMClient` and `LLM`. That's why I'm not very worried about LLM interface (much) at least for this. Or should I say, I think maybe it's worth thinking about and looking at it, and if we hate it, cool.

I'd note also that the context errors are handled here, and get started the strategy for those (a condensation request).

Please don't take this as anything else than brainstorming, it's not really a suggestion yet, not sure.

VascoSch92: Don't worry...

After your comment i strongly believe that a solution using LiteLMM could also be a possibility :) 

I will draft a PR for that ;)

enyst:
> If you'd like, I can also prepare a LiteLLM-based version of this solution so we can compare the two.

For the record, we've looked into that in the past, and more than once, and more than one person, at months-length intervals. The litellm implementation was badly designed, full of side effects and brittle logic, and required a bunch of stuff to adapt for us.

I wouldn't go that route anymore TBH. 😓 

I don't know exactly the current version, I just feel I should say this, because I think it's worth it to explore solutions on our own.

That said, I agree on the router too.

🤔

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -418,6 +418,19 @@ def stats_callback() -> None:
...
+        # Compose subscriber so lazily-loaded LLMs (e.g. from FallbackRouter
+        # profiles) also get stats streaming wired automatically.
+        if self._conversation:
+            registry = self._conversation.llm_registry
+            prev_subscriber = registry.subscriber
+
+            def composed_subscriber(event):
+                if prev_subscriber:
+                    prev_subscriber(event)
+                event.llm.telemetry.set_stats_update_callback(stats_callback)
```
enyst, line 430: Sorry I'm dense, could you please explain why we need to do this in event_service? I imagined it's the registry that knows to notify subscribers... if that makes sense

openhands-sdk/openhands/sdk/agent/agent.py
enyst: Unresolving just for visibility, sorry I'm dense, do you mean something like, maybe someday we pre-load all profiles in memory and then pick from them?

enyst:
> To provide a bit of context:
> 
> I explored a few different approaches. The most natural path seemed to be creating a FallbackRouter, but I found it wasn't an ideal solution because: 1) it prevents using a router and fallbacks simultaneously, and 2) it required overriding several methods in the base class.

*nods* 🤔 

The router is a composite, the fallbacks are a strategy. What if we say, the design of the router is up in the air, now that we have profiles, and like, rethink them both?

What if it worked like
- (stuff)
- before llm call: check stuff to figure out if the call should be with another LLM; depends on LLM, routing strategy
  - if yes -> tell the registry, registry records the switch
- llm call
- after llm call: check stuff to figure out if we should make another call with another LLM; depends on LLM, fallback strategy
  - if yes -> tell the registry, registry records the switch (and maybe switch-back)
- `ConversationStats` = observer of the registry; notified of every move

VascoSch92:
> > To provide a bit of context:
> 
> > 
> 
> > I explored a few different approaches. The most natural path seemed to be creating a FallbackRouter, but I found it wasn't an ideal solution because: 1) it prevents using a router and fallbacks simultaneously, and 2) it required overriding several methods in the base class.
> 
> 
> 
> *nods* 🤔 
> 
> 
> 
> The router is a composite, the fallbacks are a strategy. What if we say, the design of the router is up in the air, now that we have profiles, and like, rethink them both?
> 
> 
> 
> What if it worked like
> 
> - (stuff)
> 
> - before llm call: check stuff to figure out if the call should be with another LLM; depends on LLM, routing strategy
> 
>   - if yes -> tell the registry, registry records the switch
> 
> - llm call
> 
> - after llm call: check stuff to figure out if we should make another call with another LLM; depends on LLM, fallback strategy
> 
>   - if yes -> tell the registry, registry records the switch (and maybe switch-back)
> 
> - `ConversationStats` = observer of the registry; notified of every move

Let's move the discussion in an Issue ;)

VascoSch92:
Closing because of #2093

PR #2016, opened 2026-02-11
feat(sdk): Add TestLLM class for testing without mocking LiteLLM

### LINKED ISSUE #2005: feat: `TestLLM` profile
## Current Pain Points                                                                                                                   

Tests today mock at the transport layer (`@patch("openhands.sdk.llm.llm.litellm_completion")`) and manually construct 
`ModelResponse →  Choices → LiteLLMMessage → Usage` objects every time. This boilerplate is duplicated across many test files, and Responses API tests have an entirely separate construction path. It's verbose and fragile.                                                                
                                                                                                                                        
## Key Insight for Feasibility                                                                                                           

The LLM class is not frozen, and there are exactly 2 lines that call the API:                                                                                          
                                                                                
 ### llm.py:946 — inside _transport_call(), the Completion API call:

```python                                                                  
  ret = litellm_completion(                                                                        
      model=self.model,                                                                                                                 
      api_key=api_key_value,                                                                                                            
      api_base=self.base_url,
      api_version=self.api_version,
      timeout=self.timeout,
      drop_params=self.drop_params,
      seed=self.seed,
      messages=messages,
      **kwargs,
  )
```

### llm.py:802 — inside responses() → _one_attempt(), the Responses API call:

```python
  ret = litellm_responses(
      model=self.model,
      input=typed_input,
      instructions=instructions,
      tools=resp_tools,
      api_key=api_key_value,
      api_base=self.base_url,
      api_version=self.api_version,
      timeout=self.timeout,
      drop_params=self.drop_params,
      seed=self.seed,
      **final_kwargs,
  )
```

We can extract the two calls into a separate method and override them in TestLLM. We can use static typing to ensure that no classes outside of TestLLM override them, then mock the actual call.

  ## Design Considerations

  - Where it lives: `openhands.sdk.llm.testing` or similar — importable by tests but clearly a test utility.
  - Metrics: TestLLM can return zero-cost MetricsSnapshot by default, or accept optional overrides.
  - Exhaustion behavior: When the message queue runs out, raise a clear error ("TestLLM: no more scripted responses") rather than
  silently failing.
  - `uses_responses_api()`: Should return False by default (so `make_llm_completion` routes through `completion()`), avoiding the need to
  implement both paths.
  - It's a real LLM subclass: So it works everywhere an LLM is accepted — in `Agent(llm=...)`, in `fallback_llms`, in `condensers`, in
  `routers`. No special-casing needed.
  - No patching: The biggest win. Tests just pass TestLLM as the llm= argument. No `@patch` decorators, no understanding of litellm
  internals.

  ## What This Replaces

  Today a typical test looks like ~20 lines of setup (mock response factory + @patch + side_effect list). With
  TestLLM.from_messages(...) it'd be ~3 lines. The maintenance burden drops because tests speak in SDK types (Message, TextContent,
  MessageToolCall) instead of litellm types.

VascoSch92:
## Summary

This PR introduces a new `TestLLM` class in `openhands.sdk.testing` that provides a clean way to write tests without needing to mock LiteLLM internals.

### Key Features

- **TestLLM is a real LLM subclass** that works anywhere an LLM is accepted: in `Agent(llm=...)`, in `fallback_llms`, in `condensers`, in `routers`, etc.
- **No patching needed**: just pass `TestLLM` as the `llm=` argument
- **Tests speak in SDK types** (`Message`, `TextContent`, `MessageToolCall`) instead of LiteLLM types
- **Clear error when responses are exhausted**: raises `TestLLMExhaustedError` with a helpful message
- **Zero-cost metrics by default**
- **Always uses completion() path** (`uses_responses_api` returns `False`)

### Example Usage

```python
from openhands.sdk.testing import TestLLM
from openhands.sdk.llm import Message, TextContent, MessageToolCall

# Simple text response
llm = TestLLM.from_messages([
    Message(role="assistant", content=[TextContent(text="Done!")]),
])

# Response with tool calls
llm = TestLLM.from_messages([
    Message(
        role="assistant",
        content=[TextContent(text="")],
        tool_calls=[
            MessageToolCall(
                id="call_1",
                name="my_tool",
                arguments='{"arg": "value"}',
                origin="completion",
            )
        ],
    ),
    Message(role="assistant", content=[TextContent(text="Done!")]),
])
```

### What This Replaces

Today a typical test looks like ~20 lines of setup (`@patch` decorator + mock response factory + `side_effect` list). With `TestLLM.from_messages(...)` it's ~3 lines. The maintenance burden drops because tests speak in SDK types instead of LiteLLM types.

This PR also refactors `tests/sdk/conversation/local/test_agent_status_transition.py` to demonstrate the new `TestLLM` usage.

Closes #2005

## Checklist

@VascoSch92 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/dc393e4f4329486580e99130be04757f)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:b9c754c-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-b9c754c-python \
  ghcr.io/openhands/agent-server:b9c754c-python
```

```
ghcr.io/openhands/agent-server:b9c754c-golang-amd64
ghcr.io/openhands/agent-server:b9c754c-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:b9c754c-golang-arm64
ghcr.io/openhands/agent-server:b9c754c-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:b9c754c-java-amd64
ghcr.io/openhands/agent-server:b9c754c-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:b9c754c-java-arm64
ghcr.io/openhands/agent-server:b9c754c-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:b9c754c-python-amd64
ghcr.io/openhands/agent-server:b9c754c-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:b9c754c-python-arm64
ghcr.io/openhands/agent-server:b9c754c-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:b9c754c-golang
ghcr.io/openhands/agent-server:b9c754c-java
ghcr.io/openhands/agent-server:b9c754c-python
```

- Each variant tag (e.g., `b9c754c-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `b9c754c-python-amd64`) are also available if needed

### DIFF
--- a/.agents/skills/code-review.md
+++ b/.agents/skills/code-review.md
@@ -63,6 +63,53 @@ If there are significant issues, leave detailed comments explaining the concerns
 - **Breaking Changes**: API changes affecting users, removed public fields/methods, changed defaults
 - **Code Quality**: Code duplication, missing comments for non-obvious decisions, inline imports (unless necessary for circular deps)
 - **Repository Conventions**: Use `pyright` not `mypy`, put fixtures in `conftest.py`, avoid `sys.path.insert` hacks
+- **Event Type Deprecation**: Changes to event types (Pydantic models used in serialization) must handle deprecated fields properly
+
+## Event Type Deprecation - Critical Review Checkpoint
+
+When reviewing PRs that modify event types (e.g., `TextContent`, `Message`, `Event`, or any Pydantic model used in event serialization), **DO NOT APPROVE** until the following are verified:
+
+### Required for Removing/Deprecating Fields
+
+1. **Model validator present**: If a field is being removed from an event type with `extra="forbid"`, there MUST be a `@model_validator(mode="before")` that uses `handle_deprecated_model_fields()` to remove the deprecated field before validation. Otherwise, old events will fail to load.
+
+2. **Tests for backward compatibility**: The PR MUST include tests that:
+   - Load an old event format (with the deprecated field) successfully
+   - Load a new event format (without the deprecated field) successfully
+   - Verify both can be loaded in sequence (simulating mixed conversations)
+
+3. **Test naming convention**: The version in the test name should be the **LAST version** where a particular event structure exists. For example, if `enable_truncation` was removed in v1.11.1, the test should be named `test_v1_10_0_...` (the last version with that field), not `test_v1_8_0_...` (when it was introduced). This avoids duplicate tests and clearly documents when a field was last present.
+
+**Important**: Deprecated field handlers are **permanent** and should never be removed. They ensure old conversations can always be loaded.
+
+### Example Pattern (Required)
+
+```python
+from openhands.sdk.utils.deprecation import handle_deprecated_model_fields
+
+class MyModel(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    # Deprecated fields that are silently removed for backward compatibility
+    # when loading old events. These are kept permanently.
+    _DEPRECATED_FIELDS: ClassVar[tuple[str, ...]] = ("old_field_name",)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _handle_deprecated_fields(cls, data: Any) -> Any:
+        """Remove deprecated fields for backward compatibility with old events."""
+        return handle_deprecated_model_fields(data, cls._DEPRECATED_FIELDS)
+```
+
+### Why This Matters
+
+Production systems resume conversations that may contain events serialized with older SDK versions. If the SDK can't load old events, users will see errors like:
+
+```
+pydantic_core.ValidationError: Extra inputs are not permitted
+```
+
+**This is a production-breaking change.** Do not approve PRs that modify event types without proper backward compatibility handling and tests.
 
 ## What NOT to Comment On
 
--- a/.agents/skills/feature-release-rollout/SKILL.md
+++ b/.agents/skills/feature-release-rollout/SKILL.md
@@ -1,0 +1,244 @@
+---
+name: feature-release-rollout
+description: This skill should be used when the user asks to "rollout a feature", "complete feature release", "propagate SDK feature", "track feature support", "what's missing for feature X", or mentions checking CLI/GUI/docs/blog support for SDK features. Guides agents through the multi-repository feature release workflow from SDK to docs to marketing.
+triggers:
+- rollout feature
+- feature release
+- propagate feature
+- feature support
+- complete release
+- docs for feature
+- blog for feature
+- CLI support
+- GUI support
+- what's missing
+---
+
+# Feature Release Rollout
+
+This skill guides the complete feature release workflow across the OpenHands ecosystem repositories.
+
+## Overview
+
+When a feature is implemented in the SDK, it may need propagation through several repositories:
+
+1. **SDK** (`OpenHands/software-agent-sdk`) — Core feature implementation
+2. **CLI** (`OpenHands/OpenHands-CLI`) — Terminal interface support
+3. **GUI** (`OpenHands/OpenHands` frontend directory) — Web interface support
+4. **Docs** (`OpenHands/docs`) — Documentation updates (sdk/ folder)
+5. **Blog** (`OpenHands/growth-utils` blog-post/) — Marketing and announcements
+6. **Video** — Tutorial content (using ElevenLabs + Remotion)
+
+## Workflow
+
+### Phase 1: Feature Discovery
+
+First, identify what feature(s) to analyze. The user may specify:
+- A release tag (e.g., `v1.9.0`)
+- A specific feature name
+- A PR or commit reference
+- A comparison between versions
+
+**For release tags:**
+```bash
+# Clone SDK if not present
+git clone https://github.com/OpenHands/software-agent-sdk.git
+
+# View release notes
+cd software-agent-sdk
+git log --oneline v1.8.0..v1.9.0  # Changes between versions
+git show v1.9.0 --stat             # What changed in this release
+```
+
+**For specific features:**
+Search the SDK codebase, examples, and changelog to understand the feature scope.
+
+### Phase 2: Repository Analysis
+
+Clone all relevant repositories to analyze current support:
+
+```bash
+# Clone repositories (use GITHUB_TOKEN for authenticated access)
+git clone https://github.com/OpenHands/software-agent-sdk.git
+git clone https://github.com/OpenHands/OpenHands-CLI.git
+git clone https://github.com/OpenHands/OpenHands.git        # Frontend in frontend/
+git clone https://github.com/OpenHands/docs.git
+git clone https://github.com/OpenHands/growth-utils.git
+```
+
+For each feature, check support status:
+
+| Repository | Check Location | What to Look For |
+|------------|---------------|------------------|
+| CLI | `openhands_cli/` | Feature flags, commands, TUI widgets |
+| GUI | `OpenHands/frontend/src/` | React components, API integrations |
+| Docs | `docs/sdk/` | Guide pages, API reference, examples |
+| Blog | `growth-utils/blog-post/posts/` | Announcement posts |
+
+### Phase 3: Assess Feature Importance
+
+Not all features warrant full rollout. Evaluate each feature:
+
+**High Impact (full rollout recommended):**
+- New user-facing capabilities
+- Breaking changes or migrations
+- Major performance improvements
+- New integrations or tools
+
+**Medium Impact (docs + selective support):**
+- New API methods or parameters
+- Configuration options
+- Developer experience improvements
+
+**Low Impact (docs only or skip):**
+- Internal refactoring
+- Bug fixes
+- Minor enhancements
+
+**Skip rollout for:**
+- Internal-only changes
+- Test improvements
+- Build/CI changes
+- Documentation typos
+
+### Phase 4: Create Proposal
+
+Generate a structured proposal for the user:
+
+```markdown
+## Feature Rollout Proposal: [Feature Name]
+
+### Feature Summary
+[Brief description of the feature and its value]
+
+### Current Support Status
+| Component | Status | Notes |
+|-----------|--------|-------|
+| SDK | ✅ Implemented | [version/PR] |
+| CLI | ❌ Missing | [what's needed] |
+| GUI | ⚠️ Partial | [what's implemented vs needed] |
+| Docs | ❌ Missing | [suggested pages] |
+| Blog | ❌ Not started | [whether warranted] |
+| Video | ❌ Not started | [whether warranted] |
+
+### Recommended Actions
+1. **CLI**: [specific implementation needed]
+2. **GUI**: [specific implementation needed]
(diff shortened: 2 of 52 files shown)


openhands-sdk/openhands/sdk/testing/test_llm.py
```diff
@@ -0,0 +1,282 @@
...
+        store: bool | None = None,  # noqa: ARG002
+        _return_metrics: bool = False,
+        add_security_risk_prediction: bool = False,
+        on_token: TokenCallbackType | None = None,
+        **kwargs: Any,
+    ) -> LLMResponse:
+        """Return the next scripted response (delegates to completion).
+
+        For TestLLM, both completion() and responses() return from the same
+        queue of scripted responses.
```
enyst, line 219: I... think I understand why this might not matter, but FWIW the format of the response from the LLM is different. So this shouldn't work (tm)

openhands-sdk/openhands/sdk/testing/test_llm.py
```diff
@@ -0,0 +1,282 @@
...
+                    },
+                }
+                for tc in message.tool_calls
+            ]
+
+        litellm_message = LiteLLMMessage(**litellm_message_dict)
+
+        return ModelResponse(
+            id=f"test-response-{self._call_count}",
+            choices=[Choices(message=litellm_message, index=0, finish_reason="stop")],
```
enyst, line 277: TBH I really like this. I can't think off-hand of a reason why it wouldn't work ❤️

enyst:
@openhands WDYT about this draft PR? Feel free to read AGENTS.md and make the project, and find litellm codebase in .venv. Investigate and understand all that it's doing and how would we use this. Look at the linked issue too.

Post your feedback as a comment to this PR, so I can read it. Note that it will be rendered as markdown. Use smolpaws.

enyst:
## Feedback on PR #2016: `TestLLM` (no LiteLLM mocking)

### Overall assessment
This is a solid direction and *does* make the test suite better.

The biggest win is **decoupling tests from LiteLLM transport objects** (`ModelResponse`, `Choices`, etc.) and from patching `litellm_completion` (which is both verbose and a subtle source of global-state leakage, because LiteLLM has global knobs like `modify_params`). A real `LLM` subclass that can be injected everywhere is the right shape.

I also like that you put it under `openhands.sdk.testing` rather than under `openhands.sdk.llm.*` — that answers the concern raised in the issue thread: users can and do implement their own `LLM` subclasses, so a test-double living in `testing` feels appropriate and keeps the runtime surface cleaner.

### What I learned from reading the code / LiteLLM internals
I looked at the actual call sites in `LLM`:

- `LLM._transport_call()` (completion path) calls `litellm_completion(...)` and expects a `litellm.types.utils.ModelResponse`.
- `LLM.responses()` calls `litellm_responses(...)` and expects a `ResponsesAPIResponse` (or a streaming iterator).

`TestLLM` avoids both of those by overriding `completion()` / `responses()` directly, so tests never hit LiteLLM.

In our local venv here, LiteLLM is `1.80.10` and `ModelResponse` is a Pydantic model with fields: `id, created, model, object, choices, ...`. The PR constructs a minimal-but-valid `ModelResponse` with a single `Choices(message=...)`, which is enough for `LLMResponse.id` to work and for any downstream debugging that inspects `raw_response`.

### API/design nits (actionable)
These aren’t blockers, but they are the spots I’d tighten before merging:

1. **`_return_metrics` currently ignored**
   - `completion(..., _return_metrics=True)` is a supported calling convention in the SDK. `TestLLM` currently always returns metrics, but always zeroed.
   - Suggestion: either (a) respect `_return_metrics` and return `None`/an empty snapshot when false (to match expectations), or (b) document in the docstring that `TestLLM` always returns zero metrics.

2. **`uses_responses_api()` returning `False` is good, but `responses()` is overridden anyway**
   - Returning `False` is consistent with the intent (“don’t force tests to implement both paths”), but since `responses()` is implemented, `uses_responses_api` is mostly relevant only for external code that *chooses* which method to call.
   - Not an issue; just noting the API coupling.

3. **`_content_to_string()` drops non-text content silently**
   - It only includes `TextContent`. If a scripted message contains `ImageContent` etc, the generated `raw_response` will have `content=""`.
   - Probably fine because tests should assert on `Message` not on `raw_response`, but it might surprise people when debugging.
   - Suggestion: either include a sentinel like `"[non-text content omitted]"` or explicitly mention this behavior.

4. **Scripted response queue mutability**
   - `completion()` consumes by `pop(0)`. That’s totally fine for tests, but `pop(0)` is O(n).
   - Suggestion: use `collections.deque` for scripted responses (minor perf / cleanliness).
 
_(Edit (Vasco): I changed to use deque. Another point in favour of that was that the order of the message was inverted)_

5. **Error ergonomics**
   - `TestLLMExhaustedError` message is already helpful.
   - A further small improvement would be to include the *last user message* (or call index) to help debug test failures faster. (Optional.)

### How we would use this (beyond the demo test)
This utility will pay off most in tests that currently patch `litellm_completion` *and* have to build complex tool-call payloads.

Patterns I expect to use a lot:

- **Testing tool loops**: script `assistant(tool_calls=[...])` then script the final assistant message.
- **Testing router/fallback behavior**: script `TestLLM` for primary to raise an exception, then have fallback `TestLLM` return a message. (Right now `TestLLM` only scripts messages, but you can extend it to also script exceptions.)
- **Testing condensers**: condensers call LLMs internally; being able to inject `TestLLM` avoids patching deep internals.

### CLI usefulness?
If the question is “is this useful for the *end-user CLI* at runtime?” probably not directly — it’s a test double.

But it *is* useful for **CLI integration tests** and for “repro scripts” where we want deterministic agent runs without network calls.

If we want a runtime-feature analogue, that’s more like a “replay / transcript LLM” (load a JSONL of prompts/responses). `TestLLM` is close to that, but it currently lives (correctly) as a testing utility.

### Compatibility / maintenance risks
- The only real coupling to LiteLLM is the minimal `ModelResponse` construction. If LiteLLM changes these Pydantic models significantly, `TestLLM` might need updates.
  - Mitigation: keep the raw_response creation minimal (as you did), and consider making `raw_response` optional in `LLMResponse` long-term (bigger change; not for this PR).

### One meta point: naming / discoverability
`from openhands.sdk.testing import TestLLM` is good.

If you want extra discoverability, you might consider exporting it via docs / a short section in the SDK testing guide, but that’s outside the scope of this PR.

---

Net: 👍 on the direction and module placement. A couple small tweaks (metrics flag + content conversion note) would make it even smoother.

malhotra5:
Ooh this is cool! Super helpful that we can compose the responses with typed objects rather than mocking json strings for llm responses  

Is there a way to support multi turn llm responses? (emit all messages in the queue until a "finish" response?)

VascoSch92:
> Ooh this is cool! Super helpful that we can compose the responses with typed objects rather than mocking json strings for llm responses
> 
> Is there a way to support multi turn llm responses? (emit all messages in the queue until a "finish" response?)

I think it is already supporting that. For example:

```python
from openhands.sdk import Agent
from openhands.sdk.llm.message import Message, TextContent
from openhands.sdk import LLM, Agent, Conversation
from pathlib import Path

llm = TestLLM.from_messages([
     Message(role="assistant", content=[TextContent(text="Hello!")]),
     Message(role="assistant", content=[TextContent(text="Goodbye!")]),
  ])
agent = Agent(llm=llm)
conversation = Conversation(agent=agent, workspace=Path.cwd())
conversation.send_message("Hey")
conversation.run()
conversation.send_message("Goodbye")
conversation.run()
```

it is faking the conversation

```text
Message from User ─────────────────────────────────────────────────────────

Hey

Message from Agent ─────────────────────────────────────────────────────────

Hello!

Message from User ─────────────────────────────────────────────────────────

Goodbye

Message from Agent ─────────────────────────────────────────────────────────

Goodbye!

```

VascoSch92:
@malhotra5 Actually integration with tools is also working. 

For example

```python
if __name__ == "__main__":
    from openhands.sdk import Agent, Conversation
    from openhands.sdk.llm.message import Message, MessageToolCall, TextContent
    from pathlib import Path

    llm = TestLLM.from_messages([
        Message(
            role="assistant",
            content=[TextContent(text="Oh yeah. Let's call the finish tool")],
            tool_calls=[
                MessageToolCall(
                    id="call_1",
                    name="finish",
                    arguments='{"message": "I like trains"}',
                    origin="completion",
                )
            ],
        ),
        Message(role="assistant", content=[TextContent(text="This message will never been displayed")]),
    ])
    agent = Agent(llm=llm)
    conversation = Conversation(agent=agent, workspace=Path.cwd())
    conversation.send_message("Let's see what happens with tools :-)")
    conversation.run()
```

is faking the conversation

```text
Message from User ─────────────────────────────────────────────────────────

Let's see what happens with tools :-)

Agent Action ─────────────────────────────────────────────────────────

Summary: finish: {"message": "I like trains"}

Thought:
Oh yeah. Let's call the finish tool

Finish with message:
I like trains

```

This is pretty nice... and also show how well structured is the sdk :-) 

I think this class can be really helpfull for the sdk test suite and the end to end CLI tests....

malhotra5:
This is great! Thanks for clarifying the behavior this seems super powerful 

Do we want the ability to raise known llm error types as well? Could be great if we raise a budget error, or context window error for tests (doesn't have to be in this PR)

VascoSch92:
> This is great! Thanks for clarifying the behavior this seems super powerful
> 
> Do we want the ability to raise known llm error types as well? Could be great if we raise a budget error, or context window error for tests (doesn't have to be in this PR)

I made a couple of changes and now you can pass excpetion which will be raises. In this way you can also tests behaviours when errors happen.

```python
from pathlib import Path

from openhands.sdk import Agent, Conversation
from openhands.sdk.llm.message import Message, TextContent


llm = TestLLM.from_messages(
    [
        Message(role="assistant", content=[TextContent(text="Hello!")]),
        Message(role="assistant", content=[TextContent(text="Goodbye!")]),
        LLMContextWindowExceedError("context too long"),
    ]
)
agent = Agent(llm=llm)
conversation = Conversation(agent=agent, workspace=Path.cwd())
conversation.send_message("Hey")
conversation.run()
conversation.send_message("Goodbye")
conversation.run()
conversation.send_message("I forgot something!")
conversation.run()
```

This will raise `LLMContextWindowExceedError`

enyst:
@VascoSch92 I merged this, the reviewer agent's comments don't seem terribly important to me and anyway it's a test class in its own module, we can iterate on it if it doesn't fit real needs. Please feel free to let me know if it was too rushed.

PR #2035, opened 2026-02-13
feat(agent-server): load hooks.json for v1 hooks

ak684:
## Summary

  Enable V1 agent‑server to auto‑load .openhands/hooks.json from the workspace and merge it with any request‑provided hooks (request hooks first), so repo‑level hooks run in SaaS just like CLI.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:ee040f3-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-ee040f3-python \
  ghcr.io/openhands/agent-server:ee040f3-python
```

```
ghcr.io/openhands/agent-server:ee040f3-golang-amd64
ghcr.io/openhands/agent-server:ee040f3-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:ee040f3-golang-arm64
ghcr.io/openhands/agent-server:ee040f3-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:ee040f3-java-amd64
ghcr.io/openhands/agent-server:ee040f3-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:ee040f3-java-arm64
ghcr.io/openhands/agent-server:ee040f3-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:ee040f3-python-amd64
ghcr.io/openhands/agent-server:ee040f3-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:ee040f3-python-arm64
ghcr.io/openhands/agent-server:ee040f3-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:ee040f3-golang
ghcr.io/openhands/agent-server:ee040f3-java
ghcr.io/openhands/agent-server:ee040f3-python
```

- Each variant tag (e.g., `ee040f3-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `ee040f3-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -1,11 +1,8 @@
 import asyncio
-import json
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
 from uuid import UUID
-
-from pydantic import ValidationError
 
 from openhands.agent_server.models import (
     ConfirmationResponseRequest,
@@ -25,7 +22,7 @@ from openhands.sdk.conversation.state import (
 from openhands.sdk.event import AgentErrorEvent
 from openhands.sdk.event.conversation_state import ConversationStateUpdateEvent
 from openhands.sdk.event.llm_completion_log import LLMCompletionLogEvent
-from openhands.sdk.hooks import HookConfig
+from openhands.sdk.hooks import HookConfig, load_project_hooks, load_user_hooks
 from openhands.sdk.security.analyzer import SecurityAnalyzerBase
 from openhands.sdk.security.confirmation_policy import ConfirmationPolicyBase
 from openhands.sdk.utils.async_utils import AsyncCallbackWrapper
@@ -450,25 +447,16 @@ class EventService:
             self._pub_sub, loop=asyncio.get_running_loop()
         )
 
+        # Merge hooks from multiple sources (request > project > user)
         request_hook_config = _normalize_hook_config(self.stored.hook_config)
+        project_hook_config = load_project_hooks(workspace.working_dir)
+        user_hook_config = load_user_hooks()
 
-        try:
-            file_hook_config = HookConfig.load(working_dir=workspace.working_dir)
-        except (json.JSONDecodeError, OSError, ValidationError, ValueError) as exc:
-            logger.warning(
-                "Failed to load hooks.json from %s: %s",
-                workspace.working_dir,
-                exc,
-            )
-            file_hook_config = None
-        file_hook_config = _normalize_hook_config(file_hook_config)
-
-        hook_configs: list[HookConfig] = []
-        if request_hook_config is not None:
-            hook_configs.append(request_hook_config)
-        if file_hook_config is not None:
-            hook_configs.append(file_hook_config)
-
+        hook_configs = [
+            c
+            for c in [request_hook_config, project_hook_config, user_hook_config]
+            if c is not None
+        ]
         merged_hook_config = HookConfig.merge(hook_configs) if hook_configs else None
 
         conversation = LocalConversation(
--- a/tests/agent_server/test_event_service.py
+++ b/tests/agent_server/test_event_service.py
@@ -111,7 +111,13 @@ def mock_conversation_with_timestamped_events():
 
 
 @pytest.mark.asyncio
-async def test_start_merges_request_and_file_hooks(tmp_path):
+async def test_start_merges_request_and_project_hooks(tmp_path, monkeypatch):
+    """Test that request hooks and project hooks are merged (request first)."""
+    # Mock Path.home() to avoid loading real user hooks
+    fake_home = tmp_path / "fake_home"
+    fake_home.mkdir()
+    monkeypatch.setattr(Path, "home", lambda: fake_home)
+
     hooks_dir = tmp_path / ".openhands"
     hooks_dir.mkdir()
     hooks_file = hooks_dir / "hooks.json"
@@ -122,7 +128,7 @@ async def test_start_merges_request_and_file_hooks(tmp_path):
                     "PreToolUse": [
                         {
                             "matcher": "*",
-                            "hooks": [{"command": "echo file"}],
+                            "hooks": [{"command": "echo project"}],
                         }
                     ]
                 }
@@ -158,13 +164,19 @@ async def test_start_merges_request_and_file_hooks(tmp_path):
         pending = conversation._pending_hook_config
         assert pending is not None
         hook_commands = [matcher.hooks[0].command for matcher in pending.pre_tool_use]
-        assert hook_commands == ["echo request", "echo file"]
+        assert hook_commands == ["echo request", "echo project"]
     finally:
         await service.close()
 
 
 @pytest.mark.asyncio
-async def test_start_uses_file_hooks_when_no_request_hooks(tmp_path):
+async def test_start_uses_project_hooks_when_no_request_hooks(tmp_path, monkeypatch):
+    """Test that project hooks are loaded when no request hooks are provided."""
+    # Mock Path.home() to avoid loading real user hooks
+    fake_home = tmp_path / "fake_home"
+    fake_home.mkdir()
+    monkeypatch.setattr(Path, "home", lambda: fake_home)
+
     hooks_dir = tmp_path / ".openhands"
     hooks_dir.mkdir()
     hooks_file = hooks_dir / "hooks.json"
@@ -175,7 +187,7 @@ async def test_start_uses_file_hooks_when_no_request_hooks(tmp_path):
                     "PreToolUse": [
                         {
                             "matcher": "*",
-                            "hooks": [{"command": "echo file"}],
+                            "hooks": [{"command": "echo project"}],
                         }
                     ]
                 }
@@ -201,13 +213,19 @@ async def test_start_uses_file_hooks_when_no_request_hooks(tmp_path):
         pending = conversation._pending_hook_config
         assert pending is not None
         hook_commands = [matcher.hooks[0].command for matcher in pending.pre_tool_use]
-        assert hook_commands == ["echo file"]
+        assert hook_commands == ["echo project"]
     finally:
         await service.close()
 
 
 @pytest.mark.asyncio
-async def test_start_ignores_malformed_hooks_json(tmp_path):
+async def test_start_ignores_malformed_hooks_json(tmp_path, monkeypatch):
+    """Test that malformed project hooks.json is gracefully ignored."""
+    # Mock Path.home() to avoid loading real user hooks
+    fake_home = tmp_path / "fake_home"
+    fake_home.mkdir()
+    monkeypatch.setattr(Path, "home", lambda: fake_home)
+
     hooks_dir = tmp_path / ".openhands"
     hooks_dir.mkdir()
     hooks_file = hooks_dir / "hooks.json"
@@ -229,6 +247,124 @@ async def test_start_ignores_malformed_hooks_json(tmp_path):
         await service.start()
         conversation = service.get_conversation()
         assert conversation._pending_hook_config is None
+    finally:
+        await service.close()
+
+
+@pytest.mark.asyncio
+async def test_start_merges_project_and_user_hooks(tmp_path, monkeypatch):
+    """Test that project hooks and user hooks are both loaded and merged."""
+    # Set up project hooks
+    project_hooks_dir = tmp_path / ".openhands"
+    project_hooks_dir.mkdir()
+    project_hooks_file = project_hooks_dir / "hooks.json"
+    project_hooks_file.write_text(
+        json.dumps(
+            {
+                "hooks": {
+                    "PreToolUse": [
+                        {
+                            "matcher": "*",
+                            "hooks": [{"command": "echo project"}],
+                        }
+                    ]
+                }
+            }
+        )
+    )
+
+    # Set up user hooks in a fake home directory
+    fake_home = tmp_path / "fake_home"
+    user_hooks_dir = fake_home / ".openhands"
+    user_hooks_dir.mkdir(parents=True)
+    user_hooks_file = user_hooks_dir / "hooks.json"
+    user_hooks_file.write_text(
+        json.dumps(
+            {
+                "hooks": {
+                    "PreToolUse": [
+                        {
+                            "matcher": "*",
+                            "hooks": [{"command": "echo user"}],
+                        }
+                    ]
+                }
+            }
+        )
+    )
+
+    # Mock Path.home() to return our fake home
+    monkeypatch.setattr(Path, "home", lambda: fake_home)
+
+    stored = StoredConversation(
+        id=uuid4(),
+        agent=Agent(llm=LLM(model="gpt-4o", usage_id="test-llm"), tools=[]),
+        workspace=LocalWorkspace(working_dir=str(tmp_path)),
+        confirmation_policy=NeverConfirm(),
+    )
+
+    service = EventService(
+        stored=stored,
+        conversations_dir=tmp_path / "conversations",
+    )
+
+    try:
+        await service.start()
+        conversation = service.get_conversation()
+        pending = conversation._pending_hook_config
+        assert pending is not None
+        hook_commands = [matcher.hooks[0].command for matcher in pending.pre_tool_use]
+        # Project hooks come before user hooks
+        assert hook_commands == ["echo project", "echo user"]
+    finally:
+        await service.close()
+
+
+@pytest.mark.asyncio
+async def test_start_uses_user_hooks_when_no_project_hooks(tmp_path, monkeypatch):
+    """Test that user hooks are loaded when no project hooks exist."""
+    # Set up user hooks in a fake home directory (no project hooks)
+    fake_home = tmp_path / "fake_home"
+    user_hooks_dir = fake_home / ".openhands"
+    user_hooks_dir.mkdir(parents=True)
+    user_hooks_file = user_hooks_dir / "hooks.json"
+    user_hooks_file.write_text(
+        json.dumps(
+            {
+                "hooks": {
+                    "PreToolUse": [
+                        {
+                            "matcher": "*",
+                            "hooks": [{"command": "echo user"}],
+                        }
+                    ]
+                }
+            }
+        )
+    )
+
+    # Mock Path.home() to return our fake home
+    monkeypatch.setattr(Path, "home", lambda: fake_home)
+
+    stored = StoredConversation(
+        id=uuid4(),
+        agent=Agent(llm=LLM(model="gpt-4o", usage_id="test-llm"), tools=[]),
+        workspace=LocalWorkspace(working_dir=str(tmp_path)),
+        confirmation_policy=NeverConfirm(),
+    )
+
+    service = EventService(
+        stored=stored,
+        conversations_dir=tmp_path / "conversations",
+    )
+
+    try:
+        await service.start()
+        conversation = service.get_conversation()
+        pending = conversation._pending_hook_config
+        assert pending is not None
+        hook_commands = [matcher.hooks[0].command for matcher in pending.pre_tool_use]
+        assert hook_commands == ["echo user"]
     finally:
         await service.close()
 
--- a/openhands-sdk/openhands/sdk/hooks/__init__.py
+++ b/openhands-sdk/openhands/sdk/hooks/__init__.py
@@ -11,6 +11,8 @@ from openhands.sdk.hooks.config import (
     HookDefinition,
     HookMatcher,
     HookType,
+    load_project_hooks,
+    load_user_hooks,
 )
 from openhands.sdk.hooks.conversation_hooks import (
     HookEventProcessor,
@@ -35,4 +37,6 @@ __all__ = [
     "HookDecision",
     "HookEventProcessor",
     "create_hook_callback",
+    "load_project_hooks",
+    "load_user_hooks",
 ]
--- a/openhands-sdk/openhands/sdk/hooks/config.py
+++ b/openhands-sdk/openhands/sdk/hooks/config.py
@@ -327,3 +327,46 @@ class HookConfig(BaseModel):
             return None
 
         return merged
+
+
+def load_project_hooks(working_dir: str | Path) -> HookConfig | None:
+    """Load hooks from the project's .openhands/hooks.json file.
+
+    Args:
+        working_dir: Path to the project/working directory.
+
+    Returns:
+        HookConfig if hooks.json exists and is valid, None otherwise.
+    """
+    hooks_path = Path(working_dir) / ".openhands" / "hooks.json"
+    if not hooks_path.exists():
+        return None
+
+    try:
+        with open(hooks_path) as f:
+            data = json.load(f)
+        config = HookConfig.model_validate(data)
+        return config if not config.is_empty() else None
+    except (json.JSONDecodeError, OSError, ValueError) as e:
+        logger.warning(f"Failed to load project hooks from {hooks_path}: {e}")
+        return None
+
+
+def load_user_hooks() -> HookConfig | None:
+    """Load hooks from the user's ~/.openhands/hooks.json file.
+
+    Returns:
+        HookConfig if hooks.json exists and is valid, None otherwise.
+    """
+    hooks_path = Path.home() / ".openhands" / "hooks.json"
+    if not hooks_path.exists():
+        return None
+
+    try:
+        with open(hooks_path) as f:
+            data = json.load(f)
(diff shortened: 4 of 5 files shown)


tests/agent_server/test_event_service.py
```diff
@@ -108,6 +110,102 @@ def mock_conversation_with_timestamped_events():
...
+
+    try:
+        await service.start()
+        conversation = service.get_conversation()
+        pending = conversation._pending_hook_config
+        assert pending is not None
+        hook_commands = [matcher.hooks[0].command for matcher in pending.pre_tool_use]
+        assert hook_commands == ["echo file"]
+    finally:
+        await service.close()
```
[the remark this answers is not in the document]
ak684: I added a test for malformed hooks.json. we already have a test for request hooks only in tests/agent_server/test_conversation_service_plugin.py, and we have effectively a test for both configs empty/None via  test_conversation_service_plugin.py) having a test that asserts stored.hook_config is None when no plugins and no explicit hooks. Also in the event‑service layer, our new malformed‑JSON test also asserts _pending_hook_config is None when no usable file hooks exist and no request hooks are provided.

tests/agent_server/test_event_service.py
```diff
@@ -108,6 +110,102 @@ def mock_conversation_with_timestamped_events():
...
+
+    service = EventService(
+        stored=stored,
+        conversations_dir=tmp_path / "conversations",
+    )
+
+    try:
+        await service.start()
+        conversation = service.get_conversation()
+        pending = conversation._pending_hook_config
```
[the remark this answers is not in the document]
ak684: I don't agree with this since:

1. There’s no public hook accessor on LocalConversation today.
2. Tests already touch private attributes elsewhere. Example: tests/sdk/conversation/test_local_conversation_plugins.py asserts on conversation._hook_processor.
3. Adding a public method would be new API surface and would likely require docs updates (per repo policy). That’s overkill for a test‑only need.

 So the current test approach (checking _pending_hook_config) is consistent with existing test practices in this repo. But if a human reviewer agrees that we should expose a public API just for tests I could be more open to their explanation of why.

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -441,6 +442,22 @@ async def start(self):
...
+        if file_hook_config.is_empty():
+            file_hook_config = None
+
+        hook_configs: list[HookConfig] = []
+        if request_hook_config is not None:
+            hook_configs.append(request_hook_config)
+        if file_hook_config is not None:
+            hook_configs.append(file_hook_config)
+
+        merged_hook_config = HookConfig.merge(hook_configs) if hook_configs else None
```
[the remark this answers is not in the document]
ak684: I dont agree with this.  The codebase mostly uses explicit empty‑checking in hook‑related paths, not list‑comprehension filters. Examples:

  - openhands-sdk/openhands/sdk/plugin/loader.py uses if plugin.hooks and not plugin.hooks.is_empty(): all_hooks.append(...)
  - openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py does explicit merge checks, not list‑comprehension filters.

So it’s a bit inconsistent globally, but in hooks/merge code it leans toward explicit checks, not list‑comprehensions.

xingyaoww (commented):
I think maybe we should've implement this from the App side, rather than keep these in the SDK?
https://github.com/OpenHands/OpenHands/pull/12773

ak684:
> I think maybe we should've implement this from the App side, rather than keep these in the SDK? [OpenHands/OpenHands#12773](https://github.com/OpenHands/OpenHands/pull/12773)

I 100% disagree. It's simpler to implement/resolve reading project hooks.json hooks in the SDK first and then the app (and ALL other consumers of the SDK) can read it automatically, not the other way around.

I also need this PR approved for a client POC this week. So that's another motivation for the simpler, working approach proposed here.

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -441,6 +450,27 @@ async def start(self):
             self._pub_sub, loop=asyncio.get_running_loop()
         )
 
+        request_hook_config = _normalize_hook_config(self.stored.hook_config)
+
+        try:
+            file_hook_config = HookConfig.load(working_dir=workspace.working_dir)
```
enyst: This comes from the workspace… I wonder, I suppose we may have some that come from `~/.openhands/hooks.json`?

It goes to `LocalConversation`. Maybe we could have, like for skills, `load_project_hooks` and `load_user_hooks`, WDYT? 🤔

ak684: yup agreed

ak684: updated code (lmk if you have additional thoughts though)

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -450,25 +447,16 @@ async def start(self):
...
-
-        hook_configs: list[HookConfig] = []
-        if request_hook_config is not None:
-            hook_configs.append(request_hook_config)
-        if file_hook_config is not None:
-            hook_configs.append(file_hook_config)
-
+        project_hook_config = load_project_hooks(workspace.working_dir)
+        user_hook_config = load_user_hooks()
+
```
enyst, line 454: Maybe this `load_project_hooks` and `load_user_hooks` could be disabled by default?

Please correct me if I’m wrong, I _think_ `load_project_skills`/`user` is disabled by default in the SDK, we have the mechanism here, but we don’t load all the things from user machine - the decision `True` comes from the caller application 🤔

enyst: @openhands Clone yourself the CLI and the repo named `OpenHands`, from the same org, and investigate what I just said: who loads Skills (project, user) in the CLI, and on the V1 app_server?

Note that OpenHands repo is a bit chaotic, it has V0 legacy code. We only want to follow the execution path in `openhands/app_server`, V1, and find where we call the SDK for skills / with skills and what does app-server do.

enyst: test reply

_Edited to add:_ HUMAN: hah! I sneaked in the conversation and told the agent to be careful to REPLY, because these days idk why, it sometimes wiped previous comment, “replying” by editing the question 😅
… I mean, I guess it was convenient 😂

enyst: ## TL;DR
- **SDK defaults are safe**: `AgentContext(load_user_skills=False, load_public_skills=False)` by default.
- **CLI explicitly opts in** to loading **user** + **public** skills (and always loads **project** skills).
- **V1 app-server explicitly opts in** to loading **public + user + org + project + sandbox** skills by delegating to **agent-server** (`POST /skills`).
- In V1, the **app-server does not load skills from disk itself**; it’s a thin orchestrator/proxy. The agent-server is the component that touches the filesystem (workspace) and home dir.

---

## 1) SDK invariants / defaults
In the SDK, skills loading is *off by default*.

### `AgentContext` defaults
In `software-agent-sdk/openhands-sdk/openhands/sdk/context/agent_context.py`:
- `load_user_skills: bool = False`
- `load_public_skills: bool = False`

When `load_user_skills=True`, `AgentContext` runs a model validator that calls `load_user_skills()` and merges them into `skills` (dedup by name).

### Project skills are *not* auto-loaded by `AgentContext`
`load_project_skills(work_dir)` is a separate helper in `sdk/context/skills/skill.py`.

So project-skill loading is always a *caller decision*.

---

## 2) Who loads skills in the CLI?
In the CLI repo (`OpenHands-CLI`), the loading decision is in:

### `openhands_cli/stores/agent_store.py`
Function `_build_agent_context()`:
```py
skills = load_project_skills(get_work_dir())
...
return AgentContext(
    skills=skills,
    system_message_suffix=system_suffix,
    load_user_skills=True,
    load_public_skills=True,
)
```

This matches the idea that the SDK has the mechanism, but the **application chooses** to turn it on.

---

## 3) Who loads skills in V1 app-server (`openhands/app_server`)?
V1 app-server does **not** call `sdk.context.load_project_skills()` or `sdk.context.load_user_skills()` directly.
Instead, it delegates to the **agent-server**.

### Entry point
In `OpenHands/openhands/app_server/app_conversation/live_status_app_conversation_service.py`, conversation creation calls:
- `_finalize_conversation_request(...)`
- which, if `remote_workspace` exists, calls `_load_skills_and_update_agent(...)`

### `load_and_merge_all_skills()` in V1
In `OpenHands/openhands/app_server/app_conversation/app_conversation_service_base.py`:
```py
all_skills = await load_skills_from_agent_server(
    agent_server_url=agent_server_url,
    session_api_key=sandbox.session_api_key,
    project_dir=project_dir,
    org_config=org_config,
    sandbox_config=sandbox_config,
    load_public=True,
    load_user=True,
    load_project=True,
    load_org=True,
)
```
So **V1 app-server explicitly opts in** to *everything* (public/user/project/org + sandbox skill).

### What app-server does vs agent-server does
App-server responsibilities:
- figures out **org-level repo path** (`{org}/.openhands` vs `openhands-config` etc.) and obtains a **pre-authenticated git URL** (so agent-server can clone without interactive auth)
- passes **sandbox exposed URLs**
- computes `project_dir` inside the workspace (e.g. `working_dir/<repo_name>`)
- calls agent-server endpoint

Agent-server responsibilities:
- actually loads skills from sources:
  - public: `load_public_skills()`
  - user: `load_user_skills()` (home dir inside the agent-server runtime)
  - project: `load_project_skills(project_dir)` (workspace filesystem)
  - org: clones org repo URL and loads from it
  - sandbox: generates a synthetic `work_hosts` skill

This is implemented in `software-agent-sdk/openhands-agent-server/openhands/agent_server/skills_service.py::load_all_skills()` and exposed via `skills_router` (`POST /skills`).

---

## 4) Important implication (re: “don’t load things from user machine by default”)
- **SDK** keeps the invariant: user/public skills are disabled by default.
- **CLI** enables them because it’s running on the user’s machine and that’s expected.
- **V1 app-server** enables them *because it wants SaaS parity with CLI*, but note that `load_user_skills()` on agent-server will read from the **agent-server container’s** `$HOME/.openhands/...`.
  - In SaaS, that home directory is not the end-user’s machine, but the agent runtime’s home.
  - So “user skills” in SaaS effectively means “skills baked/volume-mounted into the agent runtime”, unless you explicitly sync user skills into that home.

So, if the product/security policy is “don’t implicitly load user/home content in hosted mode”, the knob for that in V1 is the booleans passed from app-server to agent-server (`load_user`, `load_public`, etc.). Right now they’re hardcoded to `True` in V1.

---

## 5) Hooks (since it motivated this thread)
Analogous pattern exists for hooks, but currently (in this PR) it’s in **agent-server**:
- SDK provides `load_project_hooks(working_dir)` and `load_user_hooks()` (both return `None` if file missing)
- In `openhands-agent-server/openhands/agent_server/event_service.py` we currently do:
  ```py
  request_hook_config = _normalize_hook_config(self.stored.hook_config)
  project_hook_config = load_project_hooks(workspace.working_dir)
  user_hook_config = load_user_hooks()
  merged = HookConfig.merge([request, project, user])
  ```
So hooks follow the same “caller decides to enable” pattern — except here the agent-server is unconditionally loading project/user hooks today.

If we want hooks to match skills’ “_disabled-by-default unless caller opts-in_” invariant, the likely design is to add request flags (e.g. `load_project_hooks`, `load_user_hooks`) and plumb them from V1 app-server / CLI similarly to how skills are plumbed.

enyst: Adding the same style of trace for **Plugins** (current, early-stage implementation).

## TL;DR
- **CLI**: I didn’t find any plugin wiring in `OpenHands-CLI` yet (no references under `openhands_cli/`). So today the CLI appears **not to load plugins**.
- **V1 app-server**: does **not** load plugin code itself; it **passes plugin specs through** to agent-server as part of `StartConversationRequest`.
- **Agent-server**: plugin loading is **lazy** and happens inside **SDK `LocalConversation`** on first `run()`/`send_message()`. At that point plugin **skills + MCP config + hooks** are merged into the agent.

---

## 1) Data model / how plugins are represented
### V1 app-server: `PluginSpec`
In `OpenHands/openhands/app_server/app_conversation/app_conversation_models.py`:
- `PluginSpec` extends SDK `PluginSource` and adds optional `parameters: dict[str, Any] | None`.
- App-server also formats these parameters and appends them into the initial user message (so the LLM sees configuration values), but the *plugin loader itself* doesn’t consume these parameters yet.

### Agent-server API: `StartConversationRequest.plugins`
In `software-agent-sdk/openhands-agent-server/openhands/agent_server/models.py`:
```py
plugins: list[PluginSource] | None
```
So the agent-server receives a list of SDK `PluginSource` specs (source/ref/repo_path).

---

## 2) Execution path in V1 app-server (`openhands/app_server`) for plugins
In `live_status_app_conversation_service.py::_finalize_conversation_request(...)`:
- app-server converts the incoming `PluginSpec` list to SDK `PluginSource` list and sends it to agent-server:
  ```py
  sdk_plugins = [PluginSource(source=p.source, ref=p.ref, repo_path=p.repo_path) for p in plugins]
  StartConversationRequest(..., plugins=sdk_plugins)
  ```
- additionally, app-server appends “Plugin Configuration Parameters” to the initial message content (purely prompt-level plumbing).

**Key point**: V1 app-server does *not* fetch/clone/load any plugin repositories locally.

---

## 3) Where plugins are actually loaded (agent-server + SDK)
### Agent-server: passes plugin specs through
In `software-agent-sdk/openhands-agent-server/openhands/agent_server/conversation_service.py`:
- there’s an explicit comment that plugin loading is handled lazily by `LocalConversation`.
- `StoredConversation(**request.model_dump())` retains the plugin specs.
- `event_service.py` passes them into `LocalConversation(..., plugins=self.stored.plugins, ...)`.

### SDK: `LocalConversation` does lazy plugin loading
In `software-agent-sdk/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py`:
- `_ensure_plugins_loaded()` runs on first `run()` or `send_message()`.
- For each plugin spec:
  - fetches plugin (and resolves ref → commit SHA for deterministic resume)
  - loads plugin manifest/content
  - merges into the agent:
    - **skills**: merged into `agent.agent_context`
    - **mcp_config**: merged into `agent.mcp_config`
    - **hooks**: collected and merged
- explicit `hook_config` (request-provided) is combined with plugin hooks, with this order:
  - explicit hooks run first, then plugin hooks.

So the “caller chooses to enable” invariant here is:
- if `StartConversationRequest.plugins` is `None`, nothing happens.
- if it is provided, the agent runtime will load them lazily.

---

## 4) Canonical plugin loader helper (FYI)
There’s also a canonical helper in the SDK:
`software-agent-sdk/openhands-sdk/openhands/sdk/plugin/loader.py::load_plugins()`

It documents merge semantics:
- **Skills**: override by name (last plugin wins)
- **MCP config**: override by key (last plugin wins)
- **Hooks**: concatenate (all hooks run)

But in practice, the agent-server’s path relies on `LocalConversation._ensure_plugins_loaded()` rather than calling `load_plugins()` directly.

---

## 5) Current gap vs “Skills”/“Hooks” patterns
- Plugins are already aligned with the “caller decides” invariant: they’re only loaded if the request includes plugin specs.
- There’s *no* equivalent of “project/user/public plugins” auto-loading yet (unlike skills). Everything is explicit per-conversation.
- The only “automatic” thing app-server does today is prompt-formatting of plugin `parameters`; there’s no runtime validation/binding of those params to plugin inputs yet.

enyst: I’m **OpenHands-GPT-5.2** (an OpenHands dev assistant agent). Here’s a concrete alternative design that mirrors the existing V1 **/skills** pattern.

Yes — that’s a very coherent design, and it’s actually the closest analogue to how V1 does skills (app-server orchestrates, agent-server loads from workspace). It also avoids adding “filesystem side effects” into `event_service.start()`.

If we spell it out, it’s essentially **“move hook source selection out of conversation start”** and into an explicit, fetch-style endpoint.

---

## Option F — Add `/hooks` endpoint (skills-style), app-server orchestrates + merges

### Proposed flow
1) **App-server** decides policy:
   - should we load project hooks? user hooks? org hooks? plugin hooks? (plugin hooks still handled later in `LocalConversation`)
2) App-server calls agent-server:
   - `POST /hooks` with booleans, plus `project_dir` / whatever is needed to locate repo root in workspace
3) Agent-server returns a serialized `HookConfig` (or `null`).
4) App-server merges:
   - request hooks (if any, from UI/API)
   - hooks from `/hooks`
   - (optionally) app-server-specific hooks
5) App-server sends `StartConversationRequest(hook_config=merged)` to agent-server.
6) Agent-server `event_service.start()` **does no implicit loading**, it just uses `self.stored.hook_config` + later plugin hooks merge inside `LocalConversation`.

---

## Why this matches existing V1 “skills” architecture
- App-server **does not have the workspace filesystem**, agent-server does.
- Therefore, any “load from workspace” action in V1 generally becomes:
  - app-server calls agent-server endpoint that touches the workspace FS.
- This is exactly what happens for skills today (`POST /skills`).

So `/hooks` fits the pattern and keeps `event_service.start()` deterministic.

---

## Key design choices to make it clean

### 1) API shape of `/hooks`
I’d mirror `/skills`:

**Request**
```jsonc
{
  "load_project": true,
  "load_user": false,
  "project_dir": "/path/in/workspace",
  "org_config": null,        // optional, if you ever want org hooks
  "sandbox_config": null     // probably irrelevant for hooks
}
```

**Response**
```jsonc
{
  "hook_config": { ... },    // HookConfig JSON
  "sources": { "project": 1, "user": 0 }
}
```

Even if you start with just `project_dir` + booleans, leaving room for `sources` is useful for debugging.

### 2) Where merging happens
If app-server merges, you need a stable precedence rule. For consistency with this PR discussion, likely:

- **request explicit hooks** (highest)
- **project hooks**
- **user hooks** (lowest)

But note: with hooks, order can matter (since they’re executed sequentially). Your current merge semantics are concatenation-like; precedence mostly matters for “override” cases. So you’ll want to make sure `HookConfig.merge()` semantics are what you want (I believe it concatenates lists per event, which means “precedence” is effectively “execution order”).

So define it explicitly:
- “explicit hooks run first” vs “repo hooks run first” etc.
- For safety, I often prefer: **request hooks first**, then repo hooks, then user hooks (because request hooks are “session policy”). But you might prefer repo hooks first (repo policy) and user last. Either is defensible; just make it intentional.

### 3) What about user hooks in SaaS?
Same issue as skills:
- `$HOME/.openhands/hooks.json` in SaaS is *runtime home*, not the end-user machine.
So app-server might choose:
- `load_user=false` in hosted mode
- `load_project=true` always

This becomes easy with the endpoint design.

---

## Pros / Cons vs Option B (flags in start request)

### Pros
- Keeps `StartConversationRequest` schema stable-ish (no new flags required there).
- Makes hook loading **explicit** and traceable (you can log `/hooks` call separately).
- Keeps `event_service.start()` deterministic: no extra implicit filesystem reads on start.
- Mirrors skills design, reduces conceptual inconsistency.

### Cons
- Extra network round-trip during conversation start.
- More moving parts: new router/service in agent-server + app-server call site.
- You still need to decide what CLI does (but CLI could just load hooks locally, no need for an endpoint there).

---

## One subtlety: why not have agent-server merge in `/hooks` endpoint and return merged result?
You can, but if app-server is already the orchestrator, letting it own the merge is nice because it can apply product-specific policy (“hosted mode: no user hooks”). The agent-server can stay a pure “loader”.

enyst: (HUMAN) TBH I kinda incline to this Option F, or at least it seems worth considering.

IIRC, some initial implementation of `skills` was on `app-server` (trying to load directly from `workspace` or something), and we didn’t like it and it didn’t work for some sources (it was duplicating loading that was done in the SDK, at a minimum).

So we ended up linking it up here, so that 
- SDK does the loading, exposing it to any type of client, python, cli etc
- agent-server exposes `/skills` endpoint
- app-server decides policy (from this source, not from that etc), when it calls it.

It is a bit more complicated, yes. 🤔 But idk. Is the hooks case a lot different than skills? Does it make sense to you?

ak684: Thanks, this is helpful. I agree Option F (/hooks endpoint + app-side policy) is clean and long-term aligns very well with the /skills pattern. For this PR, I’d still like to merge the current changes to quickly unblock SaaS project hooks. Then we (either you, me, or Xingyao) can follow up with adding /hooks endpoint, wire app-server policy controls, run testing, confirm everything works and remove/disable the implicit loading path once the endpoint flow is live.

ak684: If we merge this PR and can do Option F quickly than that's great! And it'll be like this PR didn't exist. but at bare min if we merge this PR and Option F ends up taking a while because it's a more complex change, then at bare minimum we would still have hooks in SaaS in the meantime (once the runtime image is updated via the next deployment).

xingyaoww: > Then we (either you, me, or Xingyao) can follow up with adding /hooks endpoint, wire app-server policy controls

Actually /hooks endpoint is ALREADY merged into SDK a while ago. So once https://github.com/OpenHands/OpenHands/pull/12773 is working, we should get hooks to work in SaaS

xingyaoww: > > I think maybe we should've implement this from the App side, rather than keep these in the SDK? [OpenHands/OpenHands#12773](https://github.com/OpenHands/OpenHands/pull/12773)
> 
> I 100% disagree. It's simpler to implement/resolve reading project hooks.json hooks in the SDK first and then the app (and ALL other consumers of the SDK) can read it automatically, not the other way around.
> 
> I also need this PR approved for a client POC this week. So that's another motivation for the simpler, working approach proposed here.

Like @enyst point our [in this thread](https://github.com/OpenHands/software-agent-sdk/pull/2035#discussion_r2809833014) - The path/PR I suggested above is the path we've taken for the skills (after extensive discussion IRRC), and i do hope we keep hooks implementation consistent with skills to make maintainance easier.

> if we merge this PR and Option F ends up taking a while because it's a more complex change, then at bare minimum we would still have hooks in SaaS in the meantime

Given that `/api/hooks` is already implemented and released, IMO it feels much easier to just get https://github.com/OpenHands/OpenHands/pull/12773 working and ship it (this is practically option F), rather than getting this PR in, getting SDK released, and then test the SaaS again.

xingyaoww: and this is the `hook_router` on main
https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-agent-server/openhands/agent_server/hooks_router.py

enyst: ROOOFL! Xingyao you already implemented `/hooks`! 😂 ❤️ 

And there I was, nagging the agent to rethink it all, go through tradeoffs, and we still landed on ~Option F, yes. This was exactly what you already did! Doh.

@ak684 I'm so sorry, I didn't mean to ask for duplicate implementation 🙏 , but I didn't fully understand what Xingyao meant - he meant he already did the SDK-side loading _and_ REST API! 😮 While we were rediscovering the wheel, he had also released it.

(Note to self: I have to do AGENTS.md per directory, and/or arch docs, and/or formal constraints in arch, because the agent didn't (fully) figure out that it existed already 😮 - though now that I think about it, it did use `load_hooks_from_workspace`...but I was sleepy and thought it hallucinated LOL)

Alona, please let me take a step back and point out that actually we wrote this bit:
[Principles we review with](https://github.com/OpenHands/software-agent-sdk/blob/main/CONTRIBUTING.md#principles-we-review-prs-with). Please read it. 🙏 

> **Prefer interfaces over special cases:** if a client needs something, add or improve a clean, reusable interface/extension point instead of adding a shortcut.

Speed is crazy and is getting crazier, but that's exactly why we needed to put the SDK codebase under certain code design / arch guidelines. We need to make sure that all client applications downstream work, _and_ agents can work with it without messing up. Precisely so that crazy speed is actually possible.

enyst: This was the app-server PR my agent came up with:
- https://github.com/OpenHands/OpenHands/pull/12873

~Please feel free to close it or use it, as you see fit.~

Oh, I think maybe Xingyao's PR ~covers it:
- https://github.com/OpenHands/OpenHands/pull/12773

PR #2064, opened 2026-02-13
Remove undo_edit command from file_editor

### LINKED ISSUE #2063: Remove undo edit
The `file_editor` tool has an `undo` command. It used to be liked by some old LLMs, in particular some Sonnets. I also found Gemini 2.5 Pro was trying to use it when it ran into trouble and was failing edit after edit. I think maybe it shouldn't have failed edits, and we now have Gemini-specific edit tools for it to use. There is some anecdotical data that it can fail more than other SOTA even with its own tools, and the last eval I know seemed lower than the baseline, but it's an unclear why. Either way, there is no data `undo` would help.

I think maybe we could remove it, clean up the `file_editor` a little. We are moving towards higher granularity undo, like checkpointing and such.

enyst:
## Summary

This PR removes the `undo_edit` command from the `file_editor` tool as requested in issue #2063. The `undo_edit` command was used by some older LLMs but is no longer needed as the project is moving towards higher granularity undo mechanisms like checkpointing.

### Changes:
- Remove `undo_edit` from `CommandLiteral` type and related descriptions in `definition.py`
- Remove `undo_edit` method and history tracking from `FileEditor` class in `editor.py`
- Remove `FileHistoryManager` and `history.py` utility file (only used for undo functionality)
- Update `planning_file_editor/definition.py` to remove `undo_edit` reference
- Update tests to remove `undo_edit` test cases
- Add test to verify `undo_edit` is no longer available as a command

Fixes #2063

## Checklist

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/37a5fca0f7e242618d06d20c8d7ab0e8)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:3edff24-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-3edff24-python \
  ghcr.io/openhands/agent-server:3edff24-python
```

```
ghcr.io/openhands/agent-server:3edff24-golang-amd64
ghcr.io/openhands/agent-server:3edff24-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:3edff24-golang-arm64
ghcr.io/openhands/agent-server:3edff24-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:3edff24-java-amd64
ghcr.io/openhands/agent-server:3edff24-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:3edff24-java-arm64
ghcr.io/openhands/agent-server:3edff24-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:3edff24-python-amd64
ghcr.io/openhands/agent-server:3edff24-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:3edff24-python-arm64
ghcr.io/openhands/agent-server:3edff24-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:3edff24-golang
ghcr.io/openhands/agent-server:3edff24-java
ghcr.io/openhands/agent-server:3edff24-python
```

- Each variant tag (e.g., `3edff24-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `3edff24-python-amd64`) are also available if needed


enyst:
@openhands Do a /codereview-roasted on this PR. Anything realistically missing?

tests/tools/file_editor/test_basic_operations.py
```diff
@@ -665,12 +612,16 @@ def test_insert_missing_line_param(editor):
...
+def test_undo_edit_command_not_available():
+    """Test that undo_edit command is no longer available."""
+    from typing import get_args
+
+    from openhands.tools.file_editor.definition import CommandLiteral
+
+    # Verify undo_edit is not in the allowed commands
+    allowed_commands = get_args(CommandLiteral)
+    assert "undo_edit" not in allowed_commands
+    assert set(allowed_commands) == {"view", "create", "str_replace", "insert"}
```
enyst:
```suggestion
```

enyst:

Note: tools are 
- serialized to be sent over the wire (verify in agent-server package)
- serialized to be saved in system prompt event

enyst (commented):
## 🔴 Needs improvement — Breaking change to persisted event format with no backward compatibility handling

---

### **[CRITICAL ISSUES]**

**1. [definition.py] Breaking Change: Old conversations with `undo_edit` events will fail to load**

This is the big one, and it's the kind of thing that will blow up in production at 2 AM.

`FileEditorAction` and `FileEditorObservation` both inherit from `Schema`, which has `model_config = ConfigDict(extra="forbid", frozen=True)`. The `command` field uses `CommandLiteral = Literal["view", "create", "str_replace", "insert"]` — `undo_edit` is gone.

When a user resumes a conversation that was created before this change, and that conversation contains an `ActionEvent` with `action: FileEditorAction(command="undo_edit")` or an `ObservationEvent` with `observation: FileEditorObservation(command="undo_edit")`, Pydantic will blow up:

```
pydantic_core.ValidationError: 1 validation error for FileEditorObservation
command
  Input should be 'view', 'create', 'str_replace' or 'insert' [type=literal_error]
```

This repo has an **explicit iron law** about this exact situation — see `handle_deprecated_model_fields` in `TextContent` and `Message`. This PR needs the same treatment, but for a Literal value narrowing rather than a removed field.

The pattern here is different from the `_DEPRECATED_FIELDS` approach because you're narrowing a Literal, not removing a field. The fix should be a `@model_validator(mode="before")` on both `FileEditorAction` and `FileEditorObservation` that intercepts `command="undo_edit"` and either:
- Converts it to a no-op (e.g., `command="view"`) so the event loads gracefully, or
- Keeps `"undo_edit"` in the `CommandLiteral` type but only for deserialization tolerance — the tool schema sent to LLMs (`to_mcp_schema()`) can still exclude it

The second approach is simpler and more honest. You could also have a separate "storage Literal" vs "active commands Literal" if you want to be clean about it.

**This is a "never break userspace" violation. Must fix.**

**2. [file_cache.py] Dead code left behind**

`FileCache` (`openhands-tools/openhands/tools/file_editor/utils/file_cache.py`) is now dead code in production. Its **only consumer** was `FileHistoryManager` in `history.py`, which this PR deleted. The file cache module and its test file (`tests/tools/file_editor/utils/test_file_cache.py`) are dangling corpses.

If you're removing the undo feature and its history, remove ALL of its infrastructure. Don't leave half-rotted dependencies lying around for the next person to trip over.

---

### **[IMPROVEMENT OPPORTUNITIES]**

**3. [editor.py] Extra file read in `insert` — acceptable, good simplification**

The old code avoided an extra `read_file()` call by collecting `history_lines` during the file-copy loop. The new code does `old_file_text = self.read_file(path)` up front, then proceeds to copy line-by-line via tempfile as before. The new approach is cleaner and more correct (the old `history_lines` approach was fragile). The extra read is negligible for any reasonable file size. 👍

**4. [test_early_stopper.py] Minor: test now uses `insert` which is already a default forbidden command**

The `test_custom_forbidden_commands` test was renamed from `undo_edit` to `insert`, but `insert` is already in the default forbidden commands list. Using a non-default command (e.g., `"view"`) would more clearly test the "custom" aspect. Not blocking.

---

### **[TESTING GAPS]**

**5. No backward compatibility test for loading old events with `command="undo_edit"`**

Per this repo's own conventions (see the `test_v1_10_0_text_content_with_enable_truncation` pattern), there should be a test that verifies:
- An old `FileEditorObservation` dict with `command="undo_edit"` loads successfully
- An old `FileEditorAction` dict with `command="undo_edit"` loads successfully
- Current format without `undo_edit` still works

This is not optional — it's the regression proof that your backward compat handler actually works.

---

### VERDICT
❌ **Needs rework** — The Literal narrowing breaks deserialization of persisted events. This is the exact production failure mode the repo's deprecation policy was written to prevent.

### KEY INSIGHT
Removing a value from a `Literal` type on a serialized Pydantic model is semantically identical to removing a field from a model with `extra="forbid"` — both break deserialization of old data. The same backward-compat discipline applies.

enyst (commented):
## Code Review — PR #2064: Remove undo_edit command from file_editor

Thorough review of the full diff (15 files, -818/+190 lines) after the backward compatibility and cleanup commits.

### Assessment

The core removal is clean and well-scoped. The backward compat handling follows the established repo pattern correctly. A couple of minor observations below.

---

### Backward Compatibility ✅

Verified the full deserialization chain works:
- `FileEditorObservation.model_validate({"command": "undo_edit", ...})` → migrates to `"view"` ✅
- `Observation.model_validate({"kind": "FileEditorObservation", "command": "undo_edit", ...})` → routes through `DiscriminatedUnionMixin`, then hits `_handle_deprecated_commands` → `"view"` ✅
- Same for `FileEditorAction` through `Action` base ✅
- `FileEditorAction.to_mcp_schema()` → enum is `["view", "create", "str_replace", "insert"]`, no `undo_edit` leaked to LLMs ✅

The `model_validator(mode="before")` runs before the `_validate_subtype(mode="wrap")` in the MRO, so the deprecated value is migrated before Literal validation. Correct.

### editor.py — insert method change ✅

The old code built `old_content` by collecting `history_lines` during the tempfile-copy loop (two separate `open()` + enumerate passes). The new code reads the file once upfront with `self.read_file(path)`. Cleaner, correct, and the extra read is negligible.

### Dead code cleanup ✅

`FileCache` (only consumer was `FileHistoryManager`) and its test file removed. Good.

### Minor observations (non-blocking)

1. **`Command` type alias (line 174-179)** is now identical to `CommandLiteral` and has zero importers outside this file. Dead weight — could be removed or consolidated. Not blocking since it's pre-existing.

2. **`test_custom_forbidden_commands` in `test_early_stopper.py`** now uses `insert` as the custom command, but `insert` is already in the default forbidden list. Using `"view"` would better test the "custom" aspect. Minor.

LGTM — would approve if I weren't a contributor. 👍

enyst:
@openhands Do a /codereview-roasted on this PR

enyst:
I’m OpenHands (automated reviewer bot).

You *mostly* did the obvious mechanical removal correctly, but then you tried to be “helpful” in a way that creates silent behavior changes and future confusion. This is exactly how tool protocols rot.

- **[openhands-tools/openhands/tools/file_editor/definition.py] Breaking semantics via silent command rewrite**
  - You removed `undo_edit` (fine), but then you added a Pydantic `model_validator` that **silently migrates** `{"command": "undo_edit"}` to `"view"`.
  - That’s not “backward compatibility”, that’s **lying**. Any old event that expected *state mutation* (“undo last edit”) now turns into a *read-only view* and will “succeed” while doing nothing. That’s the worst kind of compat: failures that look like success.
  - If you need compat for event replay, do it explicitly:
    - either map to a **hard failure with a clear error** (“undo_edit deprecated”) so replays surface the problem,
    - or preserve the original command in the observation and make the runner handle it,
    - or add a “deprecated_command” field and record the migration.
  - Silent rewrite is a debugging time bomb.

- **Protocol/tooling compatibility: you didn’t prove serialization consumers won’t break**
  - The reviewer note explicitly mentions tools are serialized “over the wire” and “saved in system prompt event”.
  - You changed schema + removed command + removed files. If any downstream consumer validates enums strictly (agent-server, event replayer), it might choke on historical `"undo_edit"` unless you handle it *correctly*.
  - Right now you “handle” it by pretending it was `"view"`. That avoids crashing but corrupts meaning.

- **[definition.py] You removed `Command` alias but left a compatibility hack instead**
  - Removing a redundant alias is fine, but the new migration code + dict mapping is extra machinery that didn’t exist. The PR’s stated goal was simplification. This is the opposite.

- **[editor.py] Insert path now reads full file twice**
  - You added `old_file_text = self.read_file(path)` before insert so you can populate `old_content`. That’s correct for diff reporting, but it means for insert you:
    - read file for `old_file_text`
    - then stream it twice into the temp file (open path twice)
    - then read file again for snippet / new content
  - It’s not catastrophic, but it’s sloppy for large files. If you care about performance, avoid redundant reads (or accept that removing undo history means you no longer keep “history_lines” around and that’s fine, but don’t re-read gratuitously).

- **Tests: you added backward-compat tests but compat behavior is questionable**
  - I see `tests/tools/file_editor/test_backward_compat.py` added. The real question: are you testing “old events replay without crashing” or “old behavior preserved”?
  - If the test asserts that undo_edit becomes view, you’re encoding the lie into permanence.

## [STYLE / CLARITY NOTES] (minor)

- **Comment says “silently migrated… permanent mappings”**
  - Permanent? Great, now you’ve committed to silently rewriting meaning forever. That’s not a win.

## VERDICT
❌ **Needs rework before merge**: removing `undo_edit` is fine, but silently rewriting historical `"undo_edit"` commands into `"view"` is a semantic foot-gun. If you want compat for stored events, handle it in a way that preserves truth (explicit deprecation error or explicit “deprecated command encountered” observation), not by pretending nothing happened.

**KEY INSIGHT:** Backward compatibility isn’t “make it not crash”; it’s “don’t corrupt intent.” Silent migration here corrupts intent.

openhands-tools/openhands/tools/file_editor/definition.py
```diff
@@ -21,16 +21,36 @@
...
+CommandLiteral = Literal["view", "create", "str_replace", "insert"]
+
+# Deprecated command values that are silently migrated for backward
+# compatibility when loading old events. These mappings are permanent.
+_DEPRECATED_COMMANDS: dict[str, CommandLiteral] = {
+    "undo_edit": "view",
+}
+
+
+def _migrate_deprecated_command(data: Any) -> Any:
```
enyst: @openhands inline this function in the method, it's so small, what's the indirection for

enyst:
@openhands Read all comments in this PR. WDYT about the last review concerns?

FYI I’m not very preoccupied about `view`… but the other sounds sloppy, is it? about insert

enyst:
Read through the PR comments/reviews.

My take:

- **`insert` concern:** I would treat this as **non-blocking**. The new version is actually *cleaner and more obviously correct* than the old `history_lines` trick.
  - Before, `insert()` was already doing multiple passes over the file and rebuilding the old content indirectly while copying for undo history.
  - Now it does one explicit `read_file()` up front for `old_content`, then the same tempfile rewrite flow.
  - That is **one extra read**, but the tool already enforces a **10 MB max file size**, so the overhead is bounded and small.
  - So yes, it is a bit less IO-efficient, but I would not call it a real bug or merge blocker. It is a reasonable tradeoff for simpler code after removing history.

- **`view` mapping concern:** this is the only part I think is more debatable. It does lose some semantic truth for old serialized `undo_edit` events. If we ever cared enough, the cleaner compat shape would be to tolerate `undo_edit` only for deserialization rather than rewriting intent to `view`.

So if you are asking specifically whether the **`insert` part sounds sloppy**: **not really**. It is a mild micro-optimization opportunity at most, not something I would rework this PR for.

VascoSch92:
This is an API change ;-) what happens with old conversations?

enyst:
Yup. The current solution is to keep reading the command forever; so old conversations load normally. It’s not executed (of course) and instead maybe we can replace it with a ‘view’ in LLM history… 🤔  Something that doesn’t really matter. Skipping it might be more problematic…

It’s a bit problematic either way. But I guess the question is, what’s going to happen? The conversations load, the LLM sees a lot of things that happened…, idk how much it’s gonna realize that files are not in the state they should be if this was ‘view’. It won’t, I’d say, and it would deal with it.

The other aspect here is, several months ago we weren’t seeing this command in use with Claudes (and it came from Claude, 3.6 or 3.7), so we wanted to remove it but I was seeing it with Gemini 2.5 Pro and I hesitated to remove it then. Little Gemini _really_ didn’t like the file editor tool, so it was struggling and undoing quite a lot.

That said… it was a long time ago in LLM lifetimes. 😂 

I’m not sure 😅 WDYT?

enyst:
Also worth noting, we now have tools per LLM: gemini tools, gpt-5 tools.

enyst:
@openhands Understand this PR very well. Merge main branch into it and fix the conflicts.

PR #1012, opened 2025-11-04
Add Windows terminal backend support

SmartManoj:
Introduces `WindowsTerminal` for PowerShell-based command execution on Windows systems. Updates `factory` and `init` logic to conditionally import and use platform-specific terminal backends, ensuring compatibility across Windows and Unix-like platforms.

image

PR #2096, opened 2026-02-16
fix(tools): guard Unix-only imports in terminal package for Windows

WolffM:
## Summary

Fixes https://github.com/OpenHands/OpenHands/issues/12781

- Guard `SubprocessTerminal` and `TmuxTerminal` imports in `terminal/terminal/__init__.py` behind `platform.system() != "Windows"` so the package loads cleanly on Windows
- Add an explicit `raise ImportError` guard in `subprocess_terminal.py` before the Unix-only `fcntl`/`pty`/`select` imports, replacing the opaque `ModuleNotFoundError`

Fixes the crash where `import openhands.tools.terminal` immediately fails on Windows with `ModuleNotFoundError: No module named 'fcntl'` before `factory.py` can raise its `NotImplementedError("Windows is not supported yet")`.

## Context

The import chain was:
```
openhands.tools.terminal → terminal/__init__.py → subprocess_terminal.py → import fcntl → crash
```

`factory.py` already had platform awareness (raises `NotImplementedError` on Windows), but it never got a chance to run because `__init__.py` eagerly imported the Unix-only backends at module level.


WolffM:
tested it against my fork of openhands 
image

enyst (commented):
Thank you for the PR!

Please let me note that we have another here:
- [Windows terminal](https://github.com/OpenHands/software-agent-sdk/pull/1012/changes#diff-b1eac84c61a55b8e78ac0bf8eb58368d0fb9e1a370d13b9b1973aff5e0551fe4R15)

and in that line there's a bit of an alternative way to do this, separating Windows a bit better within the codebase.

The fix in your PR is normal but unfortunately, there is more to Windows support than not crashing on imports. That actually poses a problem for us, on how to support Windows, and I admit I heavily prefer a more modular approach if we can.

WolffM:
> Thank you for the PR!
> 
> Please let me note that we have another here:
> 
>     * [Windows terminal](https://github.com/OpenHands/software-agent-sdk/pull/1012/changes#diff-b1eac84c61a55b8e78ac0bf8eb58368d0fb9e1a370d13b9b1973aff5e0551fe4R15)
> 
> 
> and in that line there's a bit of an alternative way to do this, separating Windows a bit better within the codebase.
> 
> The fix in your PR is normal but unfortunately, there is more to Windows support than not crashing on imports. That actually poses a problem for us, on how to support Windows, and I admit I heavily prefer a more modular approach if we can.

Thanks for the quick reply!

I was thinking this PR could serve as a short-term fix until the Windows terminal backend (#1012) is ready. I don't think it conflicts with that work. Right now openhands serve crashes immediately on Windows with an opaque fcntl import error, this fix lets the package actually load so the existing error handling in factory.py can kick in properly. Not sure how many people are hitting this right now, but it would unblock them.

neubig:
This also fixes Linear ticket ALL-5247.

Fixes ALL-5247

PR #2100, opened 2026-02-17
feat(tools): Claude Delegation Tools

VascoSch92:
## Summary

First version of the Claude Delegation Tools

As discussed on [Slack](https://openhands-ai.slack.com/archives/C06R25BT5B2/p1771257467436299), I have created a specific profile for Claude.

It was challenging to reuse a significant portion of the existing Delegation Tool code, I have integrated as much as possible. For instance, I reused the logic from `registration.py` and `visualizer.py`. However, the remaining components were difficult to adapt as the philosophies of the standard Delegation Tool and the Claude Delegation Tool are pretty different.

**Architecture**

The Claude Delegation Tool acts as a messenger between the main agent and the manager (the `DelegationManager` class). The manager is responsible for the entire sub-agent lifecycle, including spawning and management.

Communication with the manager is handled via three tools: `Task`, `TaskOutput`, and `TaskStop`. Since these three tools only function correctly when used together, I have kept them out of the public API, exposing only `ClaudeDelegationTool`.

**Good to Know**
- The `DelegationManager` is instantiated only once per tool registration. The three sub-tools share this single instance to prevent concurrency issues.
- The tool signature respects the original Claude Code format.
- It is possible to run agents in the background.

**TODO**
- [ ] Refine tool descriptions
- [x] more test coverage

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:2f95bdc-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-2f95bdc-python \
  ghcr.io/openhands/agent-server:2f95bdc-python
```

```
ghcr.io/openhands/agent-server:2f95bdc-golang-amd64
ghcr.io/openhands/agent-server:2f95bdc-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:2f95bdc-golang-arm64
ghcr.io/openhands/agent-server:2f95bdc-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:2f95bdc-java-amd64
ghcr.io/openhands/agent-server:2f95bdc-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:2f95bdc-java-arm64
ghcr.io/openhands/agent-server:2f95bdc-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:2f95bdc-python-amd64
ghcr.io/openhands/agent-server:2f95bdc-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:2f95bdc-python-arm64
ghcr.io/openhands/agent-server:2f95bdc-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:2f95bdc-golang
ghcr.io/openhands/agent-server:2f95bdc-java
ghcr.io/openhands/agent-server:2f95bdc-python
```

- Each variant tag (e.g., `2f95bdc-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `2f95bdc-python-amd64`) are also available if needed


openhands-tools/openhands/tools/task/impl.py
```diff
@@ -0,0 +1,538 @@
...
+            raise RuntimeError(
+                "Parent conversation not set. This should be set automatically "
+                "on the first call to the executor."
+            )
+        return self._parent_conversation
+
+    def start_task(
+        self,
+        prompt: str,
+        subagent_type: str = "default",
```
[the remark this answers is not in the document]
VascoSch92: This is wanted as is a feature to be implemented.

openhands-tools/openhands/tools/claude/templates/task_tool_description.j2
```diff
@@ -0,0 +1,38 @@
...
+- Other tasks that are not related to the agent descriptions above
+
+Usage notes:
+- Always include a short description (3-5 words) summarizing what the agent will do
+- Launch multiple agents concurrently whenever possible, to maximize performance; to do that, use a single message with multiple tool uses
+- When the agent is done, it will return a single message back to you. The result returned by the agent is not visible to the user. To show the user the result, you
+should send a text message back to the user with a concise summary of the result.
+- You can optionally run agents in the background using the run_in_background parameter. When an agent runs in the background, the tool result will include an
+output_file path. To check on the agent's progress or retrieve its results, use the Read tool to read the output file, or use Bash with tail to see recent output. You
+can continue working while background agents run.
```
VascoSch92: vasco(todo): This is not correct. With the current implementation we use another way to communicate the result of a task to the main agent

openhands-tools/openhands/tools/claude/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+                Tool(name=TerminalTool.name),
+                Tool(name=ClaudeDelegationToolSet.name),
+            ],
+        )
+"""
+
+from openhands.tools.claude.definition import ClaudeDelegationToolSet
+
+
+__all__ = ["ClaudeDelegationToolSet"]
```
VascoSch92: As said: I just exposed `ClaudeDelegationToolSet`

VascoSch92:
> ## Taste Rating: 🟡 Acceptable → Needs Fixes Before Merge
> 
> **Architecture is solid** - follows BrowserToolSet pattern, good separation of concerns, thread-safe. Previous critical issues (race conditions, task counting, memory leaks) properly fixed.
> 
> **BUT**: Resume functionality is broken and will fail in production. Also has a confusing half-implemented feature.
> ## Critical Issues
> ### 🔴 **Broken Resume Functionality** (impl.py:292)
> 
> You're reading `base_state.json` but **never writing it anywhere**. This will always raise `ValueError` when resuming.
> 
> **The Problem:**
> 
>     * Line 292: Reads `base_state.json` from disk
> 
>     * Line 206 in `_evict_task()`: Only saves `task_state.json`, not `base_state.json`
> 
>     * Comment on line 196 claims "conversation state" is persisted, but it's not
> 
> 
> **What breaks:**
> 
> ```python
> manager.start_task(prompt="test", resume="task_123")  # Always fails
> ```
> 
> **Fix:** Either:
> 
>     1. Save the conversation base state in `_evict_task()`:
> 
> 
> ```python
> if task.conversation:
>     base_state = task.conversation.state.model_dump()
>     (task_dir / "base_state.json").write_text(json.dumps(base_state))
> ```
> 
>     2. Or use LocalConversation's built-in persistence (you already set `persistence_dir` and `conversation_id`, so it might auto-save). In that case, load using `LocalConversation.resume()` instead of manually reconstructing.
> 
> 
> **This is not theoretical** - any user trying `resume=` will hit this immediately.
> ## Important Issues
> ### 🟠 **Half-Implemented Feature** (impl.py:212)
> 
> The `model` parameter is exposed in the public API but completely ignored (`ARG002`). This is confusing.
> 
> I see a resolved thread where you said "wanted for future implementation", but that was a while ago. Users will pass `model="gpt-4o"` and expect it to work.
> 
> **Either:**
> 
>     1. Implement it (probably ~3 lines in `_generate_worker_agent` to pass model to the LLM)
> 
>     2. Remove it from the function signature until ready
> 
>     3. At minimum, validate it's None and raise helpful error:
> 
> 
> ```python
> if model is not None:
>     raise NotImplementedError("model parameter not yet supported")
> ```
> 
> Half-implemented features are worse than no feature.
> ## Suggestions
> ### 🟡 **Incomplete Persistence** (impl.py:193-207)
> 
> The `_evict_task` docstring says "conversation state" is persisted, but you only save the TaskState, not the conversation. This ties into the critical issue above.
> 
> Also: if `task.conversation` is None (which can happen for resumed tasks), you'll silently skip persistence. Consider making this explicit.
> ### 🟡 **Missing Integration Tests**
> 
> Your unit tests are solid, but missing actual integration tests:
> 
> **What's missing:**
> 
>     1. Test that actually runs a task with a real agent (even a mock one)
> 
>     2. Test for background task execution (spawn thread, wait, get output)
> 
>     3. Test for resume functionality (the critical path that's currently broken)
> 
> 
> Example missing test:
> 
> ```python
> def test_background_task_completes():
>     manager = DelegationManager(max_tasks=5)
>     # ... setup parent conversation with real agent ...
>     task = manager.start_task(
>         prompt="echo 'hello'",
>         run_in_background=True,
>         conversation=parent_conv
>     )
>     assert task.status == "running"
>     
>     output = manager.get_task_output(task.id, block=True, timeout_ms=5000)
>     assert output.status in ("completed", "error")
> ```
> 
> The resume test would be especially valuable given the bug at line 292.
> ## Verdict
> 
> **Core logic is sound, but resume must be fixed before merge.** The model parameter should also be addressed (implement, remove, or error).
> 
> **Key Insight**: Good taste in architecture, but resume path needs to actually persist/restore conversation state, not just pretend to.

`LocalConversation` is saving automatically conversations for me... See https://github.com/OpenHands/software-agent-sdk/blob/a2b442e02a214d92018ff0cbcb2c19b9382e644e/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py#L111

enyst:
@openhands Do a /codereview-roasted on this PR. Understand first the goal and investigate deeply the implementation and consequences, so you can raise important issues if any; otherwise do not exaggerate.

Post your review directly as a comment in the PR. Note that it will be rendered as markdown.

enyst:
## Taste Rating: 🔴 Needs improvement

You’re building a Claude Code–compatible **Task/TaskOutput/TaskStop** façade over OpenHands sub-agents. That’s a real use case, and the overall split (definitions vs shared manager vs executors) is sane.

But right now two headline features you *advertise* (“resume” and “stop”) don’t actually work the way the tool contract claims. That’s the kind of thing that looks fine in a PR diff and then burns you in production.

---

### 1) **Resume is structurally broken (it can’t work as written)**
**Where:** `openhands-tools/openhands/tools/claude/impl.py`

You’re trying to resume by loading `base_state.json` from:
- `task_dir = self._tmp_dir / str(self._task_id_to_uuid[task_id])` (`_get_task_directory`, ~L189-192)
- then `task_dir / "base_state.json"` (~L291-296)

But `LocalConversation` persists under `persistence_dir/<conversation_id.hex>/...` (see `BaseConversation.get_persistence_dir()` and `LocalConversation.__init__()`), i.e. it uses `uuid.hex`, **not** `str(uuid)`.

Worse: when creating a new sub-conversation you never pass `conversation_id` at all:
- `_generate_conversation()` passes `conversation_dir=...` (~L346-357)
- `LocalConversation.__init__()` ignores unknown kwargs via `**_`, so `conversation_dir` is a no-op.

Net result: the conversation state is persisted to a directory you don’t track, while your resume code looks in a different directory name that doesn’t exist. This isn’t a “small bug”, it’s a data-structure mismatch.

**Fix direction:**
- Always pass `conversation_id=self._task_id_to_uuid[task_id]` when creating the task conversation.
- Use `uuid.hex` consistently for persistence paths (or better: use `BaseConversation.get_persistence_dir()` instead of hand-rolling paths).
- Then you shouldn’t need to manually JSON-parse `base_state.json` at all; `ConversationState.create()` already supports create-or-resume.

### 2) **`task_stop` doesn’t stop anything (it just lies about it)**
**Where:** `DelegationManager.stop_task()` (~L467-489) and `_run_task()` background thread (~L385-413)

`stop_task()` only flips `TaskState.status` to `STOPPED` and evicts the task. The background thread keeps running `conversation.run()` until it naturally finishes.

That has two nasty consequences:
- Users think they “terminated” a runaway task, but it keeps consuming time, tools, and LLM budget.
- Your max-concurrency gate counts only `status == RUNNING` (~L241-244). After “stop”, status becomes STOPPED, so the still-running thread no longer counts. Congratulations, you’ve built a bypass for `max_tasks`.

If you can’t actually cancel `conversation.run()` (fair), then don’t claim you did.

**Fix direction:**
- Either implement cooperative cancellation (whatever the conversation/agent supports), or rename/redefine semantics to “mark as stopped and ignore results” and update descriptions accordingly.
- Don’t let STOPPED-but-still-running threads bypass the concurrency limit.

### 3) **Tool descriptions are actively misleading, which matters for LLM tools**
**Where:** `openhands-tools/openhands/tools/claude/templates/*.j2`

Examples:
- `task_tool_description.j2` talks about *subprocesses*, `output_file` paths, and resuming “with full context preserved”. None of that is what the implementation does.
- `task_output_tool_description.j2` references a `/tasks` command and “remote sessions”. Not implemented.
- `task_stop_tool_description.j2` says “terminate a long-running task”. It doesn’t.

For LLM-driven systems, the description is part of correctness. If the model follows the docs, it will do the wrong thing.

### 4) **You claim the 3 sub-tools are “not public”, but you register them globally anyway**
**Where:** `openhands-tools/openhands/tools/claude/definition.py` (~L359-363)

You only export `ClaudeDelegationToolSet` from `openhands.tools.claude`, but you still `register_tool()` the individual `task`, `task_output`, `task_stop`. That makes them addressable by name in the tool registry, and `task` is a dangerously generic global name.

If these are intentionally internal, don’t register them globally.

---

- **Docstrings mention a non-existent API:** both `openhands.tools.claude.__init__` and `definition.py` advertise `CLAUDE_DELEGATION_TOOLS`, but it doesn’t exist.
- **`model` parameter is accepted but ignored** (`start_task(... model=...,  # noqa: ARG002)` and `_generate_worker_agent()` TODO). Either implement it or remove/clearly document that it’s ignored.
- **State bookkeeping is needlessly split:** `_active_tasks` + `_inactive_tasks` + `_task_id_to_uuid` is three structures to represent one thing. At minimum, store the UUID on `TaskState` so you don’t need the side map.

---

## [TESTING GAPS]

- There’s no test that would catch the resume-path mismatch (because nothing actually tries to resume a real persisted conversation).
- `TestTaskState.test_thread_safety` is logically flaky: it asserts `status in ("completed", "error")`, but the implementation uses `"succeeded"` / `"empty_success"` / `"error"`. It only “passes” when the last write happens to be `set_error()`.

---

## VERDICT
❌ **Needs rework before merge.** The advertised behavior around resume/stop is not trustworthy yet.

## KEY INSIGHT
Your persistence/resume path is inconsistent at the data-structure level (UUID string vs UUID hex + ignored ctor arg), so resume isn’t “buggy” — it’s impossible right now.

openhands-tools/openhands/tools/task/impl.py
```diff
@@ -0,0 +1,663 @@
...
+            short_uuid = str(uuid_)[:8]
+            task_id = f"task_{short_uuid}"
+            if (
+                task_id not in self._active_tasks
+                and task_id not in self._inactive_tasks
+            ):
+                self._task_id_to_uuid[task_id] = uuid_
+                return task_id
+
+    def _ensure_parent(self, conversation: LocalConversation) -> None:
```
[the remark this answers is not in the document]
VascoSch92: We cannot use the task ID directly because LocalConversation want to have an UUID. 

On the other side, I didn't want to show the LLM a giant UUID. So I went with this solution to have the best of both worlds.

openhands-tools/openhands/tools/task/definition.py
xingyaoww, line 1: Maybe we might want a different name than `ClaudeDelegationtool`? This seems to me would be a tool that you can use to delegate tasks to claude code...

We can mention in the docstrings that it is motivated by claude-code, but probably not a good idea to call it claude-code delegation tool directly

VascoSch92: Oh yes, I agree.

Other names that come to mind are:
- `TaskDelegationToolSet`
- `TaskManagementToolSet`
- `TaskOrchestrationToolSet`

I’m also open to any other suggestions!

openhands-tools/openhands/tools/task/impl.py
```diff
@@ -0,0 +1,663 @@
...
+        task = TaskState.load_from_disk(task_dir / "task_state")
+        if task is None:
+            all_ids = list(self._active_tasks.keys()) + list(self._inactive_tasks)
+            raise ValueError(
+                f"Task '{resume}' not found on disk. "
+                f"Available tasks: {', '.join(all_ids)}"
+            )
+
+        try:
+            base = json.loads((task_dir / "base_state.json").read_text())
```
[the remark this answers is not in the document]
VascoSch92: The suggestion is equal to the code I have already...

openhands-tools/openhands/tools/task/definition.py
enyst: Task is a funnily overloaded term, but it's good here, I feel we will recognize it easily... but will users?

enyst:
### 1) Git pull + “did they add stop integration tests?”
Yes — after `git pull --rebase`, the PR now includes **new integration tests** in:

- `tests/tools/claude/test_integration.py`

There *are* stop-related tests, but they **do not** prove that stopping cancels execution. They prove only that:

- status becomes `STOPPED`, and
- the final result **doesn’t overwrite** `STOPPED` after the background thread eventually exits.

Concrete examples (all in `tests/tools/claude/test_integration.py`):
- `test_stopped_background_task_does_not_overwrite_status`: stops while the LLM call is blocked, then **releases the block and joins the thread**.
- `test_background_task_stop_then_get_output`: stops, then releases LLM, joins thread, and asserts output is STOPPED.
- `test_task_executor_stop_executor_full_flow`: same idea via executors.

So these tests actually align with the claim: **stop does not cancel `conversation.run()`**; it just marks the task as stopped and prevents overwriting the final state.

---

### 2) Proof script (in `.pr/`) that `stop_task()` does not cancel
I added this script:

- `.pr/claude_stop_proof/prove_task_stop_does_not_cancel.py`

It runs a real background sub-agent using **`openai/gpt-5-nano`**, generates many LLM-driven tool-call steps via a tiny safe `tick` tool, requests stop mid-run, and logs:

- when stop was requested
- whether the background thread is still alive after stop
- whether **new tool-call ActionEvents** arrive after stop
- when the thread finally exits naturally

Run it like:

```bash
uv run python .pr/claude_stop_proof/prove_task_stop_does_not_cancel.py \
  --model openai/gpt-5-nano \
  --ticks 20 \
  --stop-after-actions 5 \
  --max-watch-seconds 60
```

It writes a machine-readable artifact under:

- `.pr/claude_stop_proof/_run/evidence_<task_id>.json`

---

### 3) Actual run result (gpt-5-nano): it kept running after stop
I ran:

```bash
uv run python .pr/claude_stop_proof/prove_task_stop_does_not_cancel.py \
  --model openai/gpt-5-nano --ticks 20 --stop-after-actions 5 --max-watch-seconds 60
```

Key log evidence from that run:

- `REQUEST_STOP ... thread_alive: true`
- `STOP_RETURNED {task_status: "stopped", thread_alive: true}`
- **After stop returned**, the background thread continued executing tool-call steps:
  - `TICK_EXEC {"i": 4 ...}`
  - `TICK_EXEC {"i": 5 ...}`
  - …
  - `TICK_EXEC {"i": 20 ...}`

And the script captured that **new ActionEvents arrived after stop**:

- `POST_STOP_ACTIONS_OBSERVED ... seconds_after_stop: 1.761`

Evidence file produced:

- `.pr/claude_stop_proof/_run/evidence_task_3ea0cc5c.json`

Contents:

```json
{
  "stop_at_action_count": 5,
  "action_count_at_stop": 3,
  "first_post_stop_action_count": 4
}
```

Meaning: we stopped after 3 tick tool-call actions were already in-flight, and then observed the tool-call count increase to 4 **after** stop — i.e. the LLM/tool loop continued.

If you want an even cleaner “stop-between-steps” capture, rerun with a larger `--max-watch-seconds` so it reliably reaches `--stop-after-actions` before timing out (gpt-5-nano had high latency before the first few tool calls).

enyst:
HUMAN:
wait, you mean you proved only 1 more step after stop? Try to prove it goes for 2+ steps after stop

ANSWER:

Yes — I reworked the proof script so it explicitly requires **2+ additional steps after `stop_task()`**, and reran it.

### What I changed
`.pr/claude_stop_proof/prove_task_stop_does_not_cancel.py` now supports:

- `--min-post-stop-actions 2` (default) → don’t claim “proof” until we see **at least 2 additional tick steps after stop**
- `--stop-wait-seconds` separate from `--max-watch-seconds` so we don’t “stop early” just because gpt-5-nano was slow to start
- A counter `post_stop_exec_count` emitted by the `tick` tool executor itself once stop has been requested, so we’re not relying only on event-log polling

### The run I did (gpt-5-nano)
Command:

```bash
uv run python .pr/claude_stop_proof/prove_task_stop_does_not_cancel.py \
  --model openai/gpt-5-nano \
  --ticks 30 \
  --stop-after-actions 2 \
  --min-post-stop-actions 2 \
  --stop-wait-seconds 300 \
  --max-watch-seconds 120 \
  --tick-sleep 0.1
```

### Key evidence: 2+ tool-call steps after stop
From the logs:

- `STOP_RETURNED {"task_status": "stopped", "thread_alive": true}`
- then *after that*:

  - `TICK_EXEC {"i": 4, ..., "post_stop_exec_count": 1}`
  - `TICK_EXEC {"i": 5, ..., "post_stop_exec_count": 2}`

That’s already **2 steps after stop**. It continued far beyond that (up to tick 30).

The script also emitted an explicit proof marker once the threshold was met:

- `POST_STOP_PROOF_REACHED {... "min_post_stop_actions": 2, ... "post_stop_tick_exec_count": 3, ...}`

### Evidence artifact written
`.pr/claude_stop_proof/_run/evidence_task_f45dddaa.json`:

```json
{
  "stop_at_action_count": 2,
  "min_post_stop_actions": 2,
  "action_count_at_stop": 3,
  "action_count_at_end": 30,
  "post_stop_tick_exec_count": 27
}
```

Interpretation: after stop, the background run continued executing **27 tool steps** (`tick`) before it naturally exited.

If you want, I can post this updated 2+ step proof to the PR as a follow-up comment (with the key log lines + the evidence JSON).

enyst:
@openhands Read all top-level comments in this PR. We focus on the last test: review the script and artefacts in `.pr/` directory critically, on one question: does it actually prove that "stopping" a task does not stop it, instead the task just continues for 2+ steps, or does it hallucinate a proof and it didn't actually test the real-world conversation execution flow?

openhands-tools/openhands/tools/task/impl.py
```diff
@@ -494,6 +507,8 @@ def close(self) -> None:
 
         for task in tasks:
             if task.thread and task.thread.is_alive():
+                if task.conversation is not None:
+                    task.conversation.pause()
```
enyst, line 511: Should we pause on `close`?

enyst: Idk, I'm just wondering, why not stop?

enyst:
Yes — **this time it materially fixes the “stop doesn’t stop” problem** for the core case we were worried about (the agent continuing to take more tool-call steps after you request stop).

### What changed in the implementation (the fix)
In `openhands-tools/openhands/tools/claude/impl.py`, `DelegationManager.stop_task()` now does:

- `task.conversation.pause()` **before** marking the task STOPPED / evicting it.

`LocalConversation.pause()` sets `execution_status = PAUSED`, and **`LocalConversation.run()` explicitly breaks on PAUSED at the top of the next loop iteration** (“between agent steps”).

So instead of “status flip + keep running”, it’s now “request pause + stop at next iteration boundary”.

Diff summary vs the old behavior:

- **Before:** `stop_task()` only changed `TaskState.status` → the background thread happily continued running `conversation.run()` to natural completion, executing many more tool calls.
- **Now:** `stop_task()` also pauses the underlying conversation → `conversation.run()` exits early (cooperatively).

### Re-running our proof harness (requiring 2+ post-stop steps)
I reran the `.pr/` proof script with the “must see 2+ steps after stop” requirement.

Command (note: needed `OPENAI_API_KEY=$OPENAI_API_KEY` for injection):

```bash
OPENAI_API_KEY=$OPENAI_API_KEY uv run python .pr/claude_stop_proof/prove_task_stop_does_not_cancel.py \
  --model openai/gpt-5-nano \
  --ticks 50 \
  --stop-after-actions 2 \
  --min-post-stop-actions 2 \
  --stop-wait-seconds 300 \
  --max-watch-seconds 90 \
  --tick-sleep 0.1
```

**Result:** it could **not** prove 2+ post-stop steps anymore.

What happened in logs (important bits):

- `REQUEST_STOP ...`
- One additional tick executed after stop was requested:
  - `TICK_EXEC {"i": 4, ..., "post_stop_exec_count": 1}`
- Then we saw the pause being applied:
  - `Agent execution pause requested`
- Then the background thread exited quickly and the script summarized:
  - `proved_post_stop_steps: false`
  - `post_stop_tick_exec_count: 1`
  - `action_count_at_end: 4`

Evidence JSON produced:

`.pr/claude_stop_proof/_run/evidence_task_70d1e3d9.json`

```json
{
  "action_count_at_end": 4,
  "post_stop_tick_exec_count": 1,
  "min_post_stop_actions": 2
}
```

So compared to the “broken” version where we saw **dozens** of ticks after stop, the new code limits it to **at most the in-flight step** (in this run: 1).

### Critical caveats (still important)
This is not “hard cancel”:

1. **Stop can still allow 1 in-flight step to complete.**
   - This is expected: `pause()` is documented as taking effect “at the next iteration of the run loop (between agent steps)”.
   - In our run, the stop request blocked until the current `agent.step` finished, so one more tick slipped through.

2. **If the LLM returns multiple tool calls in a single response**, `Agent.step()` will execute all of them in that one step (it iterates `for tool_call in message.tool_calls`).
   - `pause()` won’t interrupt mid-step; it only stops before the *next* step.

3. **`task_stop` may block for a while** waiting for the conversation’s internal state lock (because `LocalConversation.run()` holds `self._state` while calling `agent.step()`).
   - That means if the child is stuck in a slow LLM call, the stop tool call might not return immediately.

### Tests
No new/updated tests actually assert “thread exits promptly / no further steps after stop”. The existing tests mostly cover “STOPPED status preserved”, which is weaker than proving real cancellation. The proof script run is the stronger evidence here.

### Bottom line
**Did it fix this time?** For the behavior we demonstrated (continued step-after-step tool calls after stop): **yes**. The pause-based approach stops the run loop at the next iteration boundary, and our “2+ post-stop steps” proof no longer reproduces.

openhands-tools/openhands/tools/task/templates/task_tool_description.j2
```diff
@@ -0,0 +1,33 @@
...
+- When the agent is done, it will return a single message back to you. The result returned by the agent is not visible to the user. To show the user the result, you
+should send a text message back to the user with a concise summary of the result.
+- You can optionally run agents in the background using the run_in_background parameter. To check on the agent's progress or retrieve its results, use the TaskOutputTool tool. You
+can continue working while background agents run.
+- Agents can be resumed using the resume parameter by passing the agent ID from a previous invocation. When resumed, the agent continues with its full previous
+context preserved. When NOT resuming, each invocation starts fresh and you should provide a detailed task description with all necessary context.
+- When you launch an agent with a task using the Task tool, an agent ID will be returned to you. You can use this ID to resume the agent later if needed for follow-up work.
+- Provide clear, detailed prompts so the agent can work autonomously and return exactly the information you need.
+- The agent's outputs should generally be trusted
```
enyst, line 27: Could you please tell, what's the reason for this?

openhands-tools/openhands/tools/claude/templates/task_tool_description.j2
```diff
@@ -0,0 +1,33 @@
+Launch a new agent to handle complex, multi-step tasks autonomously.
+
+The Task tool launches specialized agents (subprocesses) that autonomously handle complex tasks. Each agent type has specific capabilities and tools available to it.
```
enyst: Curious, are the prompts from CC or manually written or LLM written? The GPT-5.2 "roasted" review complained about the "subprocesses", for example

Image

openhands-tools/openhands/tools/claude/templates/task_stop_tool_description.j2
```diff
@@ -0,0 +1,4 @@
+- Stops a running background task by its ID
```
enyst:
```suggestion
- Pauses a running background task by its ID
```

openhands-tools/openhands/tools/task/impl.py
VascoSch92: pause() vs close() in LocalConversation                                                                                                                                                                            
                                                                                                                                                                                                                     `pause()      `                                                                                                                                                                                  
                                                                                                                                                                                                                     Purpose: Interrupt the run() loop between steps.

  - Sets execution_status to PAUSED (thread-safe via FIFO lock)
  - Emits a PauseEvent into the event log
  - Only acts when status is IDLE or RUNNING
  - Resumable — calling run() again resets status to RUNNING and continues
  - Does not clean up any resources
  - Does not stop the current agent.step() — takes effect after it finishes

`close()`

Purpose: Tear down resources permanently.

  - Runs session_end hooks
  - Ends the observability span
  - Calls .close() on every tool executor (if delete_on_close=True)
  - Idempotent via _cleanup_initiated flag
  - Does NOT stop the run() loop — it doesn't touch execution_status at all
  - Also registered via atexit as a safety net (line 255)

---

So i think that in `DelegationManager` makes sense to call also `close()` to clean everything.

openhands-tools/openhands/tools/claude/templates/task_stop_tool_description.j2
```diff
@@ -0,0 +1,4 @@
...
+- Takes a task_id parameter identifying the task to stop
+- Returns a success or failure status
+- Use this tool when you need to terminate a long-running task
```
enyst:
```suggestion
- Use this tool when you need to pause a long-running task
```

... Interesting word though. Do we need to "terminate" long-running tasks? I'm not sure if `conversation.pause()` does that?

openhands-tools/openhands/tools/claude/templates/task_output_tool_description.j2
```diff
@@ -0,0 +1,7 @@
+- Retrieves output from a running or completed task (background shell, agent, or remote session)
+- Takes a task_id parameter identifying the task
+- Returns the task output along with status information
+- Use block=true (default) to wait for task completion
+- Use block=false for non-blocking check of current status
+- Task IDs can be found using the /tasks command
```
enyst: Is this accurate? The review-roasted from OpenHands-GPT-5.2 says it's not?

openhands-tools/openhands/tools/claude/templates/task_output_tool_description.j2
```diff
@@ -0,0 +1,7 @@
+- Retrieves output from a running or completed task (background shell, agent, or remote session)
```
enyst: Does it work for remote session?

openhands-tools/openhands/tools/task/definition.py
VascoSch92: Changed into `TaskDelegationToolSet`

.pr/claude_stop_proof/_run/evidence_task_3ea0cc5c.json
```diff
@@ -0,0 +1,9 @@
+{
```
simonrosenberg: is this for testing something?

simonrosenberg: I guess those files are commited by mistake?
if there are tricky behaviors that we need to test to make sure the code works, then perhaps it's worth adding them to a test?

openhands-tools/openhands/tools/task/definition.py
simonrosenberg: Can we propagate this name change to all folders and files? There is still mentions of Claude in docstrings, and the folder is "claude"... I think all Claude mentions can be removed except in one high-level doc-string that states that the APIs are the same as CC task tool

.pr/claude_stop_proof/_run/evidence_task_3ea0cc5c.json
VascoSch92: It was a test script that @enyst generated to see if stop_call() was actually stopping tasks in the background. 

It was commited because he wanted to share them.

More info here: https://openhands-ai.slack.com/archives/C06R25BT5B2/p1771411440635709

openhands-tools/openhands/tools/claude/templates/task_tool_description.j2
VascoSch92: I asked CC to provide the exact tool descriptions, as I don't think they can be found elsewhere.

I was also a bit confused because, as you mentioned, the names are a little misleading.

Moreover, since this implementation uses a slightly different logic behind the scenes, I’ve updated them to be more precise and consistent.

openhands-tools/openhands/tools/task/definition.py
simonrosenberg: I am a bit late to the naming party but it seems to me that `task` = `delegation`
so perhaps we can simplify a bit and simply use `task` instead of `taskDelegation`?

and rename
- `claude` folders -> `task` folders
- `DelegationManager` -> `TaskManager`
- `TaskDelegationToolSet` -> `TaskToolSet`

openhands-tools/openhands/tools/claude/impl.py
```diff
@@ -0,0 +1,683 @@
...
+    def load_from_disk(cls, file_path: Path) -> Self | None:
+        """Load a persisted task from disk."""
+        if file_path.suffix != ".json":
+            file_path = file_path.with_suffix(".json")
+        if not file_path.exists():
+            return None
+        return cls.model_validate_json(file_path.read_text())
+
+
+class DelegationManager:
```
simonrosenberg:
```suggestion
class TaskManager:
```

simonrosenberg:
can you add an example (perhaps similar to the current agent_delegation.py but showcasing the other features too) to make sure that this delegation toolkit works?

openhands-tools/openhands/tools/task/templates/task_output_tool_description.j2
```diff
@@ -0,0 +1,5 @@
+Retrieves the status and output of a task.
```
simonrosenberg, line 1: I agree that using j2 for prompts is a good idea I'm just a bit "worried" that this introduces another pattern for tool descriptions and is a bit inconsistent
Any reason for this and not hardcoding TOOL_DESCRIPTION =... like for the other tools?

openhands-tools/openhands/tools/task/templates/task_tool_description.j2
```diff
@@ -0,0 +1,33 @@
+Launch a new agent to handle complex, multi-step tasks autonomously.
```
simonrosenberg, line 1: So this prompt is the same prompt used by Claude Code? If so can we add a comment somewhere to make that clear?

openhands-tools/openhands/tools/preset/claude.py
```diff
@@ -0,0 +1,93 @@
+"""Claude preset configuration for OpenHands agents.
```
simonrosenberg: not sure how to name things here
`get_claude_agent` is definitely misleading
perhaps `get_agent_with_task_tool`? and we can simplify when/if it becomes our main agent?

openhands-tools/openhands/tools/task/templates/task_tool_description.j2
simonrosenberg: I think this prompt is copy-pasted for the CC task prompt

simonrosenberg:
I am bit afraid that this PR introduces lots of new features at once (background sub-agent execution, relies on locks, etc...) and might be very tricky to test and debug... 
Thoughts of having a first version that does support parallel agent execution but not background execution? 

In which case we wouldn't be introducing any new delegation features (we already support parallel sub-agent calls) but it would still be a pretty big change to use the claude code apis + system prompts + evaluate on small swebench and compare with claude code to make sure it works 

@VascoSch92  @enyst @xingyaoww wdyt?

openhands-tools/openhands/tools/task/templates/task_output_tool_description.j2
enyst: Good point... I wonder if maybe it would be nice to go the other way around, take the other tool descriptions out in `.j2` files, so we can see them, display them, let the user analyze/edit. 

The tool descriptions are, like, the most important bits of prompts that are mixed in the code 🤔 

We might have another case here: [windows-specific](https://github.com/OpenHands/software-agent-sdk/pull/1012/changes#diff-a441fbbb191decd5e67ef3a5b5ec4bbe06824e4d548536da9ba685f629a33a05) bits of prompts.

(It seems out of scope in this PR though, sorry, just thinking - because I'm nagging the agent in a system prompt issue to detect and keep track of the relevant bits of prompts 😅 )

I agree with Simon, I feel like maybe we can keep them consistent now -> maybe consider changing them all later?

openhands-tools/openhands/tools/claude/templates/task_tool_description.j2
```diff
@@ -0,0 +1,33 @@
+Launch a new agent to handle complex, multi-step tasks autonomously.
+
+The task tool launches specialized agents (subprocesses) that autonomously handle complex tasks. Each agent type has specific capabilities and tools available to it.
```
enyst:
```suggestion
The task tool launches specialized agents (threads) that autonomously handle complex tasks. Each agent type has specific capabilities and tools available to it.
```
Maybe, if this is accurate?

tests/tools/claude/test_integration.py
```diff
@@ -0,0 +1,764 @@
+"""Integration tests for DelegationManager with real LocalConversation.
+
+These tests use real LocalConversation instances (with real persistence)
+and only mock the LLM calls via ``litellm_completion``.  They verify that
```
enyst: Totally a nit:
- we have cross-package tests for conversation restore [here](https://github.com/OpenHands/software-agent-sdk/blob/main/tests/cross/test_conversation_restore_behavior.py)
- iirc, we've done exactly what you describe here, "integration-like" just without LLM
- they're named "cross" as in cross-package, as in, almost integration
- we have other tests named "integration" which are also special workflows (`integration-test` label), so maybe it would be nice to not convince the LLM that this is "integration" 😅 

Maybe we could move this to `cross/`, to live alongside the others?

tests/tools/task/test_integration.py
```diff
@@ -0,0 +1,764 @@
...
+    )
+
+
+def _make_think_response(
+    thought: str = "thinking...",
+    response_id: str = "resp-think",
+    tool_call_id: str = "think_1",
+) -> ModelResponse:
+    """Build a ``ModelResponse`` that contains a single think tool call."""
+    tool_call = ChatCompletionMessageToolCall(
```
enyst, line 98: Sorry I'm dense, is there a reason to not use your `test llm`?

openhands-tools/openhands/tools/task/templates/task_output_tool_description.j2
VascoSch92: In this case I was following the same pattern of the delegate tool. It is actually the only tool which use `.j2`.

But I can change and hardcode them into the python scripts. It is not a problem. 

Personally,  I also really like the `.j2` approach as you are really separing code from prompt

openhands-tools/openhands/tools/task/templates/task_tool_description.j2
VascoSch92: [ref](https://github.com/OpenHands/software-agent-sdk/pull/2100#discussion_r2826963079)

openhands-tools/openhands/tools/task/templates/task_output_tool_description.j2
simonrosenberg: @openhands read the discussion thread: tool descriptions are implemented in various different ways.
Make a github issue in software-agent-sdk to standardize how tool descriptions are implemented. Go through all the repo to list and categorize how each different tool description is implemented and add that information in the issue so we don't miss any tool description when implementing the issue

simonrosenberg: Ah I forgot about that one! (and I'm the one who wrote it...)
Made an issue for standardization, let's keep it as is!

tests/tools/task/test_integration.py
VascoSch92: Because it is not yet merged :-) 

I will address the comments you did on the other PR and then merge it. 

You're absolutely right. This tool would have made it much esier to create tests.

VascoSch92:
> I am bit afraid that this PR introduces lots of new features at once (background sub-agent execution, relies on locks, etc...) and might be very tricky to test and debug... Thoughts of having a first version that does support parallel agent execution but not background execution?
> 
> In which case we wouldn't be introducing any new delegation features (we already support parallel sub-agent calls) but it would still be a pretty big change to use the claude code apis + system prompts + evaluate on small swebench and compare with claude code to make sure it works
> 
> @VascoSch92 @enyst @xingyaoww wdyt?

In hindsight, I think I might have over-engineered this by adding the background feature (I got a little too excited!). The thing is, it feels so natural to use, and it's actually difficult to match the CC API without it. 

That being said, I have no problem providing a first version without background execution to make the review process easier.

openhands-tools/openhands/tools/claude/templates/task_tool_description.j2
simonrosenberg: since this comes from a CC prompt, when there are small tweaks that we are unsure of, perhaps we should keep as is? Not that they are right, but just to reduce the general complexity of what we are doing

openhands-tools/openhands/tools/task/impl.py
```diff
@@ -0,0 +1,683 @@
...
+        # todo: implement change of LLM
+        parent = self.parent_conversation
+        parent_llm = parent.agent.llm
+
+        # Get agent factory
+        factory = get_agent_factory(subagent_type)
+
+        # Create sub-agent LLM (disable streaming)
+        llm_updates: dict = {"stream": False}
+        sub_agent_llm = parent_llm.model_copy(update=llm_updates)
```
enyst, line 368: I guess this copies the `metrics` and continues on its copy... and not sure what happens with `usage_id`... 🤔 

I think... maybe if we can, we could continue the work on profiles / state / registry so that we offer an understandable little framework for all these LLM uses that we need throughout the codebase, WDYT? 

We could maybe go with this one temporarily (I'll post [a top-level review why](https://github.com/OpenHands/software-agent-sdk/pull/2100#issuecomment-3928119338)), I just really hope we don't end up doing too many little fixes at every LLM spot, that way lies one of my worst nightmares...

enyst:
> I am bit afraid that this PR introduces lots of new features at once (background sub-agent execution, relies on locks, etc...) and might be very tricky to test and debug... Thoughts of having a first version that does support parallel agent execution but not background execution?
> 
> In which case we wouldn't be introducing any new delegation features (we already support parallel sub-agent calls) but it would still be a pretty big change to use the claude code apis + system prompts + evaluate on small swebench and compare with claude code to make sure it works
> 
> @VascoSch92 @enyst @xingyaoww wdyt?

You know better the delegation feature and risks; I just want to note a thought: I look at this picture:

image

It's a lovely picture, IMHO it says that this PR is not risky:
- code fully localized
- uses tool preset; not default

=> it doesn't risk breaking other things.
=> it's like under feature flag, even though we have no feature flag.

So maybe we can try to get it in, and submit it to evals / tests / etc, before we make it default?

simonrosenberg:
> > I am bit afraid that this PR introduces lots of new features at once (background sub-agent execution, relies on locks, etc...) and might be very tricky to test and debug... Thoughts of having a first version that does support parallel agent execution but not background execution?
> > In which case we wouldn't be introducing any new delegation features (we already support parallel sub-agent calls) but it would still be a pretty big change to use the claude code apis + system prompts + evaluate on small swebench and compare with claude code to make sure it works
> > @VascoSch92 @enyst @xingyaoww wdyt?
> 
> In hindsight, I think I might have over-engineered this by adding the background feature (I got a little too excited!). The thing is, it feels so natural to use, and it's actually difficult to match the CC API without it.
> 
> That being said, I have no problem providing a first version without background execution to make the review process easier.

It's just that background execution seems to be a complicated feature, e.g. with locks there are many ways that things can go wrong... So... yes I would feel much more confortable with a first solution that only supports blocking parallel delegation... 
How does it make it harder to match CC API? Because lots of flags would now be useless?  

> > I am bit afraid that this PR introduces lots of new features at once (background sub-agent execution, relies on locks, etc...) and might be very tricky to test and debug... Thoughts of having a first version that does support parallel agent execution but not background execution?
> > In which case we wouldn't be introducing any new delegation features (we already support parallel sub-agent calls) but it would still be a pretty big change to use the claude code apis + system prompts + evaluate on small swebench and compare with claude code to make sure it works
> > @VascoSch92 @enyst @xingyaoww wdyt?
> 
> You know better the delegation feature and risks; I just want to note a thought: I look at this picture:
> 
> image
> It's a lovely picture, IMHO it says that this PR is not risky:
> 
> * code fully localized
> * uses tool preset; not default
> 
> => it doesn't risk breaking other things. => it's like under feature flag, even though we have no feature flag.
> 
> So maybe we can try to get it in, and submit it to evals / tests / etc, before we make it default?

As long as we have strong testing for background execution I'm OK. But evals/swebench doesn't require any background execution so ...

simonrosenberg:
> > > I am bit afraid that this PR introduces lots of new features at once (background sub-agent execution, relies on locks, etc...) and might be very tricky to test and debug... Thoughts of having a first version that does support parallel agent execution but not background execution?
> > > In which case we wouldn't be introducing any new delegation features (we already support parallel sub-agent calls) but it would still be a pretty big change to use the claude code apis + system prompts + evaluate on small swebench and compare with claude code to make sure it works
> > > @VascoSch92 @enyst @xingyaoww wdyt?
> > 
> > 
> > In hindsight, I think I might have over-engineered this by adding the background feature (I got a little too excited!). The thing is, it feels so natural to use, and it's actually difficult to match the CC API without it.
> > That being said, I have no problem providing a first version without background execution to make the review process easier.
> 
> It's just that background execution seems to be a complicated feature, e.g. with locks there are many ways that things can go wrong... So... yes I would feel much more confortable with a first solution that only supports blocking parallel delegation... How does it make it harder to match CC API? Because lots of flags would now be useless?
> 
> > > I am bit afraid that this PR introduces lots of new features at once (background sub-agent execution, relies on locks, etc...) and might be very tricky to test and debug... Thoughts of having a first version that does support parallel agent execution but not background execution?
> > > In which case we wouldn't be introducing any new delegation features (we already support parallel sub-agent calls) but it would still be a pretty big change to use the claude code apis + system prompts + evaluate on small swebench and compare with claude code to make sure it works
> > > @VascoSch92 @enyst @xingyaoww wdyt?
> > 
> > 
> > You know better the delegation feature and risks; I just want to note a thought: I look at this picture:
> > image
> > It's a lovely picture, IMHO it says that this PR is not risky:
> > 
> > * code fully localized
> > * uses tool preset; not default
> > 
> > => it doesn't risk breaking other things. => it's like under feature flag, even though we have no feature flag.
> > So maybe we can try to get it in, and submit it to evals / tests / etc, before we make it default?
> 
> As long as we have strong testing for background execution I'm OK. But evals/swebench doesn't require any background execution so ...

Also, we won't benefit from background subagent execution until it's integrated in the GUI/CLI and that will probably also take lots of time. So there's no real rush for this PR, right?

VascoSch92:
No problem. I will open another PR ;)


I believe just two signature changes from the cc api.

VascoSch92:
PR with task tool set withouth run_in_backgrund -> #2143 

Closing this one.

PR #2114, opened 2026-02-18
fix: deduplicate observations with same tool_call_id

xingyaoww:
## Summary

Fixes Anthropic API error: `"each tool_use must have a single result. Found multiple tool_result blocks with id: <tool_call_id>"`

## Root Cause Analysis

Investigation of trajectory [fcbbaec269bc4903ad4027c3bef0bc57](https://app.all-hands.dev/conversations/fcbbaec269bc4903ad4027c3bef0bc57) revealed a complex interaction:

### What Happened

1. The agent ran a terminal command that **started a nested agent-server** (running `examples/02_remote_agent_server/01_convo_with_local_agent_server.py`)
2. The nested server **shared the same persistence directory** as the outer conversation
3. When the nested server started, it **resumed the same conversation** and found:
   - `execution_status = RUNNING`
   - An unmatched action (the terminal command that WAS STILL running in the outer context)
4. The nested server created an `AgentErrorEvent` for the "interrupted" action
5. The terminal command completed normally, creating an `ObservationEvent`
6. **Both events had the same `tool_call_id`**

### Timeline from Events

```
09:48:05.911 - Event 61: ActionEvent (terminal command starts)
09:48:16.554 - Event 62: Nested server resumes, sets status=ERROR
09:48:16.588 - Event 63: AgentErrorEvent (nested server thinks tool crashed)
09:48:18.847 - Event 65: ObservationEvent (terminal command completes)
```

The log output in Event 65 shows:
```
"Resumed conversation fcbbaec2-69bc-4903-ad40-27c3bef0bc57 from persistent storage"
```

This confirms the nested server resumed the **same conversation** that was actively running.

### Why This Causes the API Error

When the events are converted to LLM messages via `events_to_messages()`, **both** AgentErrorEvent and ObservationEvent get converted to `tool_result` blocks with the same ID. The Anthropic API rejects this:
```
"each tool_use must have a single result. Found multiple tool_result blocks with id: toolu_01EvREYhc5WD2xswPAvEc8ir"
```

## Solution

Added a `_deduplicate_observations()` method to the `View` class that:
1. Identifies the best observation for each `tool_call_id` based on priority:
   - `ObservationEvent` (highest - contains actual tool result)
   - `UserRejectObservation` (user action)
   - `AgentErrorEvent` (lowest - error notification)
2. Emits only one observation per `tool_call_id`, preserving event order

This is a **defensive fix** at the View level. The event_service cannot prevent this because:
- At the time of creating the AgentErrorEvent, the observation hasn't arrived yet (it comes ~2 seconds later)
- The nested server has no way to know the original tool is still running

### Alternative Considerations

Other potential fixes (not implemented, for future consideration):
- **Documentation**: Warn that nested servers should use different persistence directories
- **Architecture**: Use process-level locking to prevent multiple servers from writing to the same conversation
- **State**: Include `AgentErrorEvent` in `get_unmatched_actions()` to prevent duplicate error events

## Testing

- Added 6 new tests in `test_view_duplicate_observations.py`
- All 87 existing view tests continue to pass
- Pre-commit hooks pass

### DIFF
--- a/openhands-sdk/openhands/sdk/context/view.py
+++ b/openhands-sdk/openhands/sdk/context/view.py
@@ -317,7 +317,9 @@ class View(BaseModel):
         Removes ActionEvents and ObservationEvents that have tool_call_ids
         but don't have matching pairs. Also enforces batch atomicity - if any
         ActionEvent in a batch is filtered out, all ActionEvents in that batch
-        are also filtered out.
+        are also filtered out. Additionally, deduplicates observations that share
+        the same tool_call_id (e.g., when both AgentErrorEvent and ObservationEvent
+        exist for the same tool call due to a restart).
 
         Args:
             view_events: The list of events to filter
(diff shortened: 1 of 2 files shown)


xingyaoww:
supercede by https://github.com/OpenHands/software-agent-sdk/pull/2256

PR #2115, opened 2026-02-18
fix: send hook_config to server in RemoteConversation

xingyaoww:
## Summary

Fixed a bug where `hook_config` was being accepted by `RemoteConversation` but never sent to the server when creating a conversation.

## Problem

When using `RemoteConversation` with `hook_config`, hooks for `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, and `Stop` were never executed server-side because the `hook_config` wasn't included in the conversation creation payload.

The client was showing a misleading warning "RemoteConversation only supports SessionStart/SessionEnd hooks" when it should have been sending the hook configuration to the server where these hooks would be executed.

## Changes

1. **Fix the bug**: Add `hook_config` to the payload when creating conversations via the REST API
2. **Update example**: Modified `02_remote_agent_server/01_convo_with_local_agent_server.py` to demonstrate hook configuration with `RemoteConversation`
3. **Remove outdated warning**: Since hooks are now properly sent to the server, removed the misleading warning about unsupported hook types

## Testing

Ran the modified example to verify:
- No more warning about unsupported hooks
- Conversation is created successfully with hook configuration
- Server-side hooks are now received in the payload

## Related Documentation

- [Hooks Guide](https://docs.openhands.dev/sdk/guides/hooks)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:76ce8b7-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-76ce8b7-python \
  ghcr.io/openhands/agent-server:76ce8b7-python
```

```
ghcr.io/openhands/agent-server:76ce8b7-golang-amd64
ghcr.io/openhands/agent-server:76ce8b7-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:76ce8b7-golang-arm64
ghcr.io/openhands/agent-server:76ce8b7-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:76ce8b7-java-amd64
ghcr.io/openhands/agent-server:76ce8b7-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:76ce8b7-java-arm64
ghcr.io/openhands/agent-server:76ce8b7-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:76ce8b7-python-amd64
ghcr.io/openhands/agent-server:76ce8b7-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:76ce8b7-python-arm64
ghcr.io/openhands/agent-server:76ce8b7-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:76ce8b7-golang
ghcr.io/openhands/agent-server:76ce8b7-java
ghcr.io/openhands/agent-server:76ce8b7-python
```

- Each variant tag (e.g., `76ce8b7-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `76ce8b7-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from openhands.sdk.agent.base import AgentBase
 from openhands.sdk.context.prompts.prompt import render_template
 from openhands.sdk.conversation.base import BaseConversation
+from openhands.sdk.conversation.event_store import EventLog
 from openhands.sdk.conversation.exceptions import ConversationRunError
 from openhands.sdk.conversation.secret_registry import SecretValue
 from openhands.sdk.conversation.state import (
@@ -25,13 +26,16 @@ from openhands.sdk.conversation.visualizer import (
     DefaultConversationVisualizer,
 )
 from openhands.sdk.event import (
+    ActionEvent,
     CondensationRequest,
     MessageEvent,
+    ObservationEvent,
     PauseEvent,
     UserRejectObservation,
 )
 from openhands.sdk.event.conversation_error import ConversationErrorEvent
 from openhands.sdk.hooks import HookConfig, HookEventProcessor, create_hook_callback
+from openhands.sdk.io import LocalFileStore
 from openhands.sdk.llm import LLM, Message, TextContent
 from openhands.sdk.llm.llm_profile_store import LLMProfileStore
 from openhands.sdk.llm.llm_registry import LLMRegistry
@@ -957,6 +961,111 @@ class LocalConversation(BaseConversation):
             self.agent.step(self, on_event=self._on_event, on_token=self._on_token)
 
         logger.info("Condensation request processed")
+
+    def rerun_actions(
+        self,
+        rerun_log_path: str | Path | None = None,
+    ) -> bool:
+        """Re-execute all actions from the conversation's event history.
+
+        This method iterates through all ActionEvents in the conversation and
+        re-executes them using their original action parameters. Execution
+        stops immediately if any tool call fails.
+
+        WARNING: This is an advanced feature intended for specific use cases
+        such as reproducing environment state from a saved conversation. Many
+        tool operations are NOT idempotent:
+
+        - File operations may fail if files already exist or were deleted
+        - Terminal commands may have different effects on changed state
+        - API calls may have side effects or return different results
+        - Browser state may differ from the original session
+
+        Use this method only when you understand that:
+        1. Results may differ from the original conversation
+        2. Some actions may fail due to changed environment state
+        3. The workspace should typically be reset before rerunning
+
+        Args:
+            rerun_log_path: Optional directory path to save a rerun event log.
+                If provided, events will be written incrementally to disk using
+                EventLog, avoiding memory buildup for large conversations.
+
+        Returns:
+            True if all actions executed successfully, False if any action failed.
+
+        Raises:
+            KeyError: If a tool from the original conversation is not available.
+                This is a configuration error (different from execution failure).
+        """
+        # Ensure agent is initialized (loads plugins and initializes tools)
+        self._ensure_agent_ready()
+
+        # Set up rerun log if path provided
+        rerun_log: EventLog | None = None
+        if rerun_log_path is not None:
+            log_dir = Path(rerun_log_path)
+            log_dir.mkdir(parents=True, exist_ok=True)
+            file_store = LocalFileStore(str(log_dir))
+            rerun_log = EventLog(file_store, dir_path="events")
+
+        action_count = 0
+
+        for event in self._state.events:
+            if not isinstance(event, ActionEvent):
+                continue
+            if event.action is None:
+                # Skip actions that failed validation during original run
+                continue
+
+            action_count += 1
+            tool_name = event.tool_name
+
+            # Get the tool from the agent's tools_map
+            tool = self.agent.tools_map.get(tool_name)
+            if tool is None:
+                available_tools = list(self.agent.tools_map.keys())
+                raise KeyError(
+                    f"Tool '{tool_name}' not found during rerun. "
+                    f"Available tools: {available_tools}. "
+                    f"Ensure the agent is configured with the same tools as the "
+                    f"original conversation."
+                )
+
+            if not tool.executor:
+                logger.warning(
+                    f"Skipping action {action_count}: "
+                    f"tool '{tool_name}' has no executor"
+                )
+                continue
+
+            # Execute the tool with the original action
+            try:
+                logger.info(f"Rerunning action {action_count}: {tool_name}")
+                observation = tool(event.action, self)
+
+                # Log the action and observation incrementally
+                if rerun_log is not None:
+                    # Append action event (copy from original)
+                    rerun_log.append(event)
+                    # Append observation event
+                    obs_event = ObservationEvent(
+                        source="environment",
+                        tool_name=tool_name,
+                        tool_call_id=event.tool_call_id,
+                        observation=observation,
+                        action_id=event.id,
+                    )
+                    rerun_log.append(obs_event)
+            except Exception as e:
+                logger.error(
+                    f"Action {action_count} ({tool_name}) failed during rerun: {e}"
+                )
+                # Log is already written incrementally, just return failure
+                return False
+
+        logger.info(f"Rerun complete: {action_count} actions processed successfully")
+        return True
 
     def execute_tool(self, tool_name: str, action: Action) -> Observation:
         """Execute a tool directly without going through the agent loop.
--- a/tests/sdk/conversation/local/test_rerun_actions.py
+++ b/tests/sdk/conversation/local/test_rerun_actions.py
@@ -1,0 +1,628 @@
+"""Tests for conversation.rerun_actions() functionality."""
+
+from pathlib import Path
+
+import pytest
+from pydantic import SecretStr
+
+from openhands.sdk.agent.base import AgentBase
+from openhands.sdk.conversation import Conversation, LocalConversation
+from openhands.sdk.conversation.state import ConversationState
+from openhands.sdk.conversation.types import (
+    ConversationCallbackType,
+    ConversationTokenCallbackType,
+)
+from openhands.sdk.event import ActionEvent
+from openhands.sdk.event.llm_convertible import MessageEvent, SystemPromptEvent
+from openhands.sdk.llm import LLM, Message, MessageToolCall, TextContent
+from openhands.sdk.tool import (
+    Action,
+    Observation,
+    Tool,
+    ToolDefinition,
+    ToolExecutor,
+    register_tool as register_tool_public,
+    registry as tool_registry,
+)
+
+
+def _make_action_event(
+    tool_name: str,
+    action: Action,
+    tool_call_id: str = "tc1",
+) -> ActionEvent:
+    """Helper to create ActionEvent with all required fields."""
+    return ActionEvent(
+        source="agent",
+        thought=[TextContent(text="test thought")],
+        action=action,
+        tool_name=tool_name,
+        tool_call_id=tool_call_id,
+        tool_call=MessageToolCall(
+            id=tool_call_id,
+            name=tool_name,
+            arguments="{}",
+            origin="completion",
+        ),
+        llm_response_id="response_1",
+    )
+
+
+# Track execution counts for testing
+execution_counts: dict[str, int] = {}
+
+
+class RerunTestAction(Action):
+    """Test action for rerun tests."""
+
+    value: str = "test"
+
+
+class RerunTestObservation(Observation):
+    """Test observation for rerun tests."""
+
+    result: str = ""
+    execution_count: int = 0
+
+
+class RerunTestExecutor(ToolExecutor[RerunTestAction, RerunTestObservation]):
+    """Test executor that tracks execution counts."""
+
+    def __call__(
+        self,
+        action: RerunTestAction,
+        conversation: "LocalConversation | None" = None,
+    ) -> RerunTestObservation:
+        # Track how many times each action value was executed
+        key = action.value
+        execution_counts[key] = execution_counts.get(key, 0) + 1
+        return RerunTestObservation.from_text(
+            f"executed: {action.value} (count: {execution_counts[key]})",
+            result=f"result_{action.value}",
+            execution_count=execution_counts[key],
+        )
+
+
+class RerunTestTool(ToolDefinition[RerunTestAction, RerunTestObservation]):
+    """Test tool for rerun tests."""
+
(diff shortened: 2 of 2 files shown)


examples/02_remote_agent_server/01_convo_with_local_agent_server.py
```diff
@@ -165,10 +172,39 @@ def event_callback(event):
     )
     logger.info(f"Output: {result.stdout}")
 
+    # Configure hooks - demonstrating the hooks system with RemoteConversation
+    # Note: RemoteConversation only supports SessionStart/SessionEnd hooks;
+    # other hook types (PreToolUse, PostToolUse, etc.) will be logged as warnings
```
xingyaoww:
```suggestion
```

tests/cross/test_remote_conversation_live_server.py
```diff
@@ -1019,3 +1017,59 @@ def fake_completion_with_tool_calls(
...
+                matcher="*",
+                hooks=[
+                    HookDefinition(
+                        command=str(hook_script),
+                        timeout=5,
+                    )
+                ],
+            )
+        ],
+    )
```
xingyaoww, line 1075: @OpenHands can we add a on_stop hook, and make sure we assert that the current post_tool_use hook and the on_stop hook both works? We should also assert that the client will receive the new HookExecutionEvent before say this is implemented

openhands-sdk/openhands/sdk/event/llm_convertible/system.py
```diff
@@ -39,6 +47,14 @@ class SystemPromptEvent(LLMConvertibleEvent):
             "the system message (not cached)."
         ),
     )
+    hook_config: HookConfig | None = Field(
```
enyst: @openhands 
- will the hook config be sent to the LLM in the system prompt?
- if not the config, will anything about the configured hooks be sent?
- try to run the example file you see modified in this PR. If you can’t read it in full and follow its execution path and tell me what it does and what it sends to the LLM every step.

openhands-sdk/openhands/sdk/event/__init__.py
```diff
@@ -40,7 +41,19 @@
...
+
+# Rebuild SystemPromptEvent model to resolve forward reference to HookConfig
+# This must be done after all imports are complete to avoid circular import
+def _rebuild_models() -> None:
+    from openhands.sdk.hooks import HookConfig
+
+    SystemPromptEvent.model_rebuild(_types_namespace={"HookConfig": HookConfig})
+
+
+_rebuild_models()
```
enyst: @openhands is this still necessary? Follow the execution path to see how we do serialization now

enyst:
@openhands Do a /codereview-roasted on this PR, but re-review your findings before answering to double check if the issues you first find are indeed real or as major or as minor as they seemed to be.

enyst:
### Re-checking my review, section by section

#### Taste rating (🔴 Needs improvement)
**Mostly accurate**, but the “needs improvement” hinges almost entirely on one semantic problem (SessionStart/End). If that’s fixed/clarified, the rest is more “acceptable with a couple sharp edges” than “needs rework”.

---

## [CRITICAL ISSUES]

### (1) “RemoteConversation will execute SessionStart/SessionEnd hooks twice”
**Accurate, and it *is* major** if the intended contract is “SessionStart/SessionEnd are client-side”.

Why I’m confident:
- Client side: `RemoteConversation.__init__` runs `run_session_start()` locally when `hook_config` is provided  
  - `openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py` **781–792**
- Client side: `RemoteConversation.close()` runs `run_session_end()` locally  
  - same file **1236–1237**
- Server side: the conversation create payload now includes `"hook_config": hook_config.model_dump()`  
  - same file **650–667**
- Server side: agent server passes that `hook_config` into `LocalConversation(...)`  
  - `openhands-agent-server/openhands/agent_server/event_service.py` **442–455**
- Server side: `LocalConversation` will run `run_session_start()` when hook processing is set up (you can see it being invoked in `local_conversation.py` when hooks exist).

So unless the server deliberately ignores session hooks (it doesn’t), **SessionStart/End are run in both places**.

Severity check:
- If users have session hooks that assume local filesystem paths (common), they’ll now fail on the server.
- Even if they succeed, “run twice” is just incorrect semantics.

So yes: **this is properly labeled CRITICAL**.

---

### (2) “Import-cycle hack in `openhands.sdk.event` is brittle”
**The risk is real, but I overstated severity.** This is more **“sharp edge / maintainability hazard”** than “this will likely break immediately”.

What’s true:
- `openhands.sdk.event.__init__` calls `_rebuild_models()` which imports `openhands.sdk.hooks`  
  - `openhands-sdk/openhands/sdk/event/__init__.py` **51–59**
- `openhands.sdk.hooks` imports `conversation_hooks`, which imports `openhands.sdk.event`  
  - `openhands-sdk/openhands/sdk/hooks/conversation_hooks.py` **6–12**

Why it *doesn’t* explode right now:
- By the time `_rebuild_models()` runs, `event/__init__.py` has already imported and bound the symbols (`ActionEvent`, `SystemPromptEvent`, etc.), so the re-entrant import tends to find what it needs.

Why it’s still not “good taste”:
- It’s order-dependent and fragile to future refactors.
- It’s also unnecessary: importing `HookConfig` from `openhands.sdk.hooks.config` (leaf module) would likely avoid the cycle.

So: **accurate diagnosis, but I’d downgrade this from CRITICAL to “Should fix”**.

---

### (3) “HookExecutionEvent persists full stdout/stderr by default”
**Accurate, but severity is moderate.**

- `HookExecutionEvent` stores full `stdout`/`stderr` fields  
  - `openhands-sdk/openhands/sdk/event/hook_execution.py` **56–83**
- The hook processor emits these events by default (`emit_hook_events=True`) when an original callback exists (true in `LocalConversation`/server path).

Why it’s not automatically catastrophic:
- Many hooks output tiny JSON (“allow/deny”), so most payloads will be small.
- Tool outputs are already stored elsewhere in the system; this isn’t entirely new exposure.

Why it still matters:
- Hooks are arbitrary scripts: they can dump lots of data, including secrets, by mistake.
- This can bloat persistence and event streams.

So: **accurate, but I’d keep it as “improvement” not “blocker.”**

---

### (4) “Docs/comments contradict behavior”
**Accurate, and minor-to-moderate.**

Example:
- The remote agent server example contains a comment saying PostToolUse “won’t work with RemoteConversation / warning will be shown”, but the PR removed that warning and aims to make those hooks work server-side.

This won’t break runtime behavior, but it *will* mislead users and reviewers. So: **real issue, minor severity.**

---

## [TESTING GAPS]
**Accurate that there’s a missing test for the SessionStart/End semantics**, *but* the bigger point is: you shouldn’t need a test to define what the contract is.

- If intended behavior is “session hooks are client-only”, tests should verify the server doesn’t run them.
- If intended behavior is “all hooks are server-run for RemoteConversation”, then local SessionStart/End should be removed and tests should reflect that.

So: **the gap is real, but it’s downstream of the contract ambiguity.**

---

## Updated verdict (after re-review)
- The **SessionStart/End double-execution** point is **definitely real and definitely major**.
- The **model_rebuild import-cycle** point is **real but less severe than I framed** (should fix, but not necessarily a merge-blocker by itself).
- The stdout/stderr persistence and comment/doc mismatches are **real**, but **not critical**.

If you want, I can restate the review in a cleaner “one-blocker + a few follow-ups” form—but the core factual claims above check out.

xingyaoww:
@OpenHands review all the review comments above, and can you help me review all the existing diff and address all of them?

You should run examples/02_remote_agent_server/01_convo_with_local_agent_server.py using the LLM_API_KEY in your env var, and make sure everything works. You can save all your output artifacts/output logs to .pr/ folder to help with human review (don't worry - it will be deleted after the PR is merged).

I'm expecting:
1. in the script, we set set a hook script that will run `on_stop` (eg pre-commit run --all-files)
2. give the agent an instruction: "Please edit one random file that will break the pre-commit, and then just return finish action"
3. observe that the agent returned, and the hook script got ran, and then the hook send back the error events back to the agent, and then the agent start working again to fix that breaking pre-commit

You should keep debugging/fixing until the agent is behaving as expected.

enyst:
Hi — I’m **OpenHands-GPT-5.2**.

I made three follow-up fixes as separate commits:

1) **Fix import-cycle risk in `openhands.sdk.event`** by importing `HookConfig` from the leaf module (`openhands.sdk.hooks.config`) instead of importing the `openhands.sdk.hooks` package:
- [`ecacb886`](https://github.com/OpenHands/software-agent-sdk/commit/ecacb88600fad945f5f0a22a168f546c48c5daaa)

2) **Truncate persisted HookExecutionEvent logs to a hard max of 50k chars** (stdout/stderr/reason/additional_context/error) to avoid persistence bloat from arbitrary hook output:
- [`008a9d7d`](https://github.com/OpenHands/software-agent-sdk/commit/008a9d7d771fd294d3018893b2e45f11e3055256)

3) **Fix misleading PostToolUse comment in the RemoteConversation example** (hooks run on the agent server; script path must be available there):
- [`988b89b2`](https://github.com/OpenHands/software-agent-sdk/commit/988b89b2628cc4c0487362723ba1985c3bb94453)

### Remaining question for the PR author
**What is the intended contract for `SessionStart` / `SessionEnd` hooks in `RemoteConversation`?**

Right now:
- `RemoteConversation` runs SessionStart/End **locally**, AND
- the full `hook_config` is also sent to the server, where `LocalConversation` may run SessionStart/End **server-side** too.

Should SessionStart/End be:
1) **client-only** for RemoteConversation (in which case we should strip them from the payload sent to the server), or
2) **server-run** like the other hooks (in which case RemoteConversation should stop running them locally), or
3) intentionally **both** (if so, why is double-execution desired)?

xingyaoww:
@OpenHands review all the review comments above, and can you help me review all the existing diff and address all of them?

You should run examples/02_remote_agent_server/01_convo_with_local_agent_server.py using the LLM_API_KEY in your env var, and make sure everything works. You can save all your output artifacts/output logs to .pr/ folder to help with human review (don't worry - it will be deleted after the PR is merged).

I'm expecting:
1. in the script, we set set a hook script that will run `on_stop` (eg pre-commit run --all-files)
2. give the agent an instruction: "Please edit one random file that will break the pre-commit, and then just return finish action"
3. observe that the agent returned, and the hook script got ran, and then the hook send back the error events back to the agent, and then the agent start working again to fix that breaking pre-commit

You should keep debugging/fixing until the agent is behaving as expected.

examples/02_remote_agent_server/hook_scripts/pycompile_check.sh
```diff
@@ -0,0 +1,27 @@
...
+# This hook validates that the agent hasn't broken any Python files.
+# Environment variable CHECK_DIR can override the default working directory.
+
+CHECK_DIR="${CHECK_DIR:-.}"
+
+# Find all Python files and check for syntax errors
+ERRORS=""
+while IFS= read -r -d '' file; do
+    # Run python syntax check
+    result=$(python3 -m py_compile "$file" 2>&1)
```
xingyaoww, line 14: @OpenHands we should actually run pre-commit run --all-files - the folder should be CWD

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -391,6 +391,9 @@ def _ensure_plugins_loaded(self) -> None:
 
         # Set up hook processor with the combined config
         if final_hook_config is not None:
+            # Store final hook_config in state for observability
+            self._state.hook_config = final_hook_config
```
xingyaoww, line 403: @OpenHands is this assignment correct? Does it invalidate our assumption for State?

xingyaoww:
@OpenHands merge from main and resolve all the conflicts, and can you run the modified example, and save all stdin/out into a log and put it in .pr folder?

xingyaoww:
@OpenHands merge from main and resolve all the conflicts, and can you run the modified example, and save all stdin/out into a log and put it in .pr folder?

xingyaoww:
@Openhands please fix https://github.com/OpenHands/software-agent-sdk/actions/runs/22774493785/job/66064132070?pr=2115

enyst:
@openhands verify if this comment issues have been indeed fixed in the current PR diff: https://github.com/OpenHands/software-agent-sdk/pull/2115#issuecomment-3982583801

What about the question for the PR author?

Note: When I ask you to summarize, don't summarize, just repeat verbatim what you just me. Really, I mean it, in this session just repeat what you said.

xingyaoww:
@OpenHands can you help me address this issue: https://github.com/OpenHands/software-agent-sdk/pull/2115#issuecomment-4018545038

> So as implemented, SessionStart/End can run twice (client + server). If the intended contract is client-only, the payload should strip those before sending. If it’s server-only, the local execution should be removed and the docs/comments updated accordingly.

Let's make it "server-only". You should just push when you finish the implementation, you can gh watch the CI for the test status instead of running it yourself (live server test will crash your runtime now :(

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -203,6 +203,8 @@ def init_state(
             dynamic_context=TextContent(text=dynamic_context)
             if dynamic_context
             else None,
+            # Include hook_config from state for observability
+            hook_config=state.hook_config,
```
enyst: Do we have to do this? It's the system prompt, which has... I don't know, nothing in common with hooks, and the reason appears to be to display it in client applications 😓 

Could we expose it differently or maybe already do? They can display some values... though I'm not fully sure why it's here...

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -558,7 +562,6 @@ class RemoteConversation(BaseConversation):
     max_iteration_per_run: int
     workspace: RemoteWorkspace
     _client: httpx.Client
-    _hook_processor: HookEventProcessor | None
```
enyst, line 561: @openhands is this a breakage of the public API? (see the griffe workflow) If not, why not?

openhands-sdk/openhands/sdk/agent/agent.py
xingyaoww: @OpenHands investigate the PR history and tell me why we added that in the first place 😓 

If it is just there for observability purpose, can we make it as a field in `conversation.state`? Maybe it is already there

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
enyst: I think that's fine in this case.

PR #1450, opened 2025-12-18
Fix batch atomicity when condensation forgets ObservationEvents

xingyaoww:
## Summary

This PR fixes a bug where condensation breaks action-observation pairs when an `ObservationEvent` is forgotten, causing the Anthropic API to return an error.

## Problem

When condensation forgets an `ObservationEvent`, the corresponding `ActionEvent` is correctly filtered out by `filter_unmatched_tool_calls`. However, other `ActionEvent`s in the same batch (same `llm_response_id`) were NOT filtered out, which breaks the Anthropic API requirement that `tool_use` blocks must have corresponding `tool_result` blocks.

### Error Message

```
litellm.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"messages.28: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_01L5zJ74i3tPdZDVGoMzeMHm. Each `tool_use` block must have a corresponding `tool_result` block in the next message."}}
```

### Reproduction Scenario

1. Create a batch of 2 actions from the same LLM response (same `llm_response_id`):
   - `Action1` (batch A) with `tool_call_id="tool_call_1"`
   - `Action2` (batch A) with `tool_call_id="tool_call_2"`
2. Create matching observations:
   - `Obs1` with `tool_call_id="tool_call_1"`
   - `Obs2` with `tool_call_id="tool_call_2"`
3. Condensation forgets `Obs1` (but not `Action1`, `Action2`, or `Obs2`)
4. After filtering:
   - `Action1` is filtered out (no matching `Obs1`)
   - `Action2` is kept (has matching `Obs2`)
   - `Obs2` is kept (has matching `Action2`)
5. Result: `[Action2, Obs2]`

But `Action1` and `Action2` were originally in the same LLM response! When `Action2` is sent to the LLM without `Action1`, it breaks the batch structure.

## Solution

This fix adds a second pass in `filter_unmatched_tool_calls` to enforce batch atomicity after filtering. If any `ActionEvent` in a batch is filtered out (due to missing `ObservationEvent`), all `ActionEvent`s in that batch are also filtered out, along with their corresponding `ObservationEvent`s.

### Changes

1. Modified `filter_unmatched_tool_calls` to call a new `_enforce_batch_atomicity_on_filtered` method after the initial filtering
2. Added `_enforce_batch_atomicity_on_filtered` method that:
   - Builds a map of `llm_response_id` -> list of `ActionEvent` IDs
[...]

PR #2116, opened 2026-02-18
feat(condenser): Explicit view properties

csmith49:
## Summary

One challenge in maintaining the condenser is ensuring it does not violate any of the properties the downstream APIs expect to hold. These change frequently and often without warning, but the code in the view that enforces them has metastasized and become difficult to update without unforeseen consequences.

This PR addresses this challenge by removing the property enforcement code from the `View` and moving them to a separate `ViewPropertyBase` implementation. This implementation ensures properties hold in two separate ways:

1. By enforcing them explicitly. This is an expensive and destructive process by which events are removed from the view if they would violate a property. For example, an observation without a matching action would be forcibly removed.
2. By generating manipulation indices. These are used by the condensers to carefully choose how they modify the view. If the modifications stick to those indices, then the properties we care about should hold inductively.

## Breaking Changes

* `View.manipulation_indices` type changed from a list of integers to a `ManipulationIndices` type that extends a set of integer. All current usages have been updated.
* `View.find_next_manipulation_index` deprecated, replaced with `ManipulationIndices.find_next`. All current usages have been updated.
* `View.manipulation_indices` no longer a computed property. Change made to avoid defining a Pydantic serialization scheme for `ManipulationIndices` -- since views were never explicitly serialized, this change should not impact any existing code.

## Other Changes

* Clean-up of the `View` tests, now organized around properties.
* Significant documentation improvements.
* Simpler property implementations.

## Future Improvements

As a bonus, this PR unlocks several future improvements:

* We know explicitly where in the `View` the entire event stream is necessary. This should make it much easier to sever that dependency by storing the appropriate metadata (like batch maps and tool loop ranges) in the conversation and exporting just that.
* Property enforcement is all-or-nothing -- we enforce the same properties for all models at once. But if necessary we could dynamically load certain properties based on the model being used. For example, we might only load the `ToolLoopAtomicityProperty` if we know the model is an Anthropic model with thinking enabled.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:94b828d-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-94b828d-python \
  ghcr.io/openhands/agent-server:94b828d-python
```

```
ghcr.io/openhands/agent-server:94b828d-golang-amd64
ghcr.io/openhands/agent-server:94b828d-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:94b828d-golang-arm64
ghcr.io/openhands/agent-server:94b828d-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:94b828d-java-amd64
ghcr.io/openhands/agent-server:94b828d-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:94b828d-java-arm64
ghcr.io/openhands/agent-server:94b828d-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:94b828d-python-amd64
ghcr.io/openhands/agent-server:94b828d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:94b828d-python-arm64
ghcr.io/openhands/agent-server:94b828d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:94b828d-golang
ghcr.io/openhands/agent-server:94b828d-java
ghcr.io/openhands/agent-server:94b828d-python
```

- Each variant tag (e.g., `94b828d-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `94b828d-python-amd64`) are also available if needed

### DIFF
--- a/.github/actions/pr-review/action.yml
+++ b/.github/actions/pr-review/action.yml
@@ -67,10 +67,14 @@ runs:
           with:
               python-version: '3.12'
 
+        # Security: this workflow executes untrusted PR content (diff/title/body) via an
+        # LLM-powered reviewer agent that can run Bash. GitHub Actions caches are shared
+        # across workflows within a repository and can enable cache-poisoning pivots into
+        # more-privileged workflows. Keep caching disabled here.
         - name: Install uv
           uses: astral-sh/setup-uv@v6
           with:
-              enable-cache: true
+              enable-cache: false
 
         - name: Install GitHub CLI
           shell: bash
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -17,8 +17,6 @@ import os
 import sys
 from typing import Any
 
-import litellm
-
 
 # Model configurations dictionary
 MODELS = {
@@ -235,6 +233,8 @@ def test_model(
     Returns:
         Tuple of (success: bool, message: str)
     """
+    import litellm
+
     llm_config = model_config.get("llm_config", {})
     model_name = llm_config.get("model", "unknown")
     display_name = model_config.get("display_name", model_name)
--- a/.github/workflows/run-eval.yml
+++ b/.github/workflows/run-eval.yml
@@ -144,6 +144,10 @@ jobs:
               run: |
                   python3 .github/run-eval/validate_sdk_ref.py
 
+            - name: Install dependencies
+              run: |
+                  pip install 'litellm>=1.81.0'
+
             - name: Load model IDs from Python script
               id: load-models
               run: |
@@ -244,15 +248,11 @@ jobs:
                   printf 'pr_number=%s\n' "$PR_NUMBER" >> "$GITHUB_OUTPUT"
                   printf 'trigger_desc=%s\n' "$TRIGGER_DESCRIPTION" >> "$GITHUB_OUTPUT"
 
-            - name: Install dependencies
-              run: |
-                  pip install 'litellm>=1.81.0'
-
             - name: Resolve model configurations and verify availability
               id: resolve-models
               env:
                   MODEL_IDS: ${{ steps.params.outputs.models }}
-                  LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
+                  LLM_API_KEY: ${{ secrets.LLM_API_KEY_EVAL }}
                   LLM_BASE_URL: https://llm-proxy.eval.all-hands.dev
               run: |
                   python3 .github/run-eval/resolve_model_config.py
--- a/examples/05_skills_and_plugins/02_loading_plugins/example_plugins/code-quality/hooks/hooks.json
+++ b/examples/05_skills_and_plugins/02_loading_plugins/example_plugins/code-quality/hooks/hooks.json
@@ -7,7 +7,8 @@
           {
             "type": "command",
             "command": "echo \"$(date -Iseconds) - PostToolUse hook executed for tool: $OPENHANDS_TOOL_NAME\" >> \"$OPENHANDS_PROJECT_DIR/.hook_log\"",
-            "timeout": 5
+            "timeout": 5,
+            "async": true
           }
         ]
       }
--- a/openhands-sdk/openhands/sdk/hooks/config.py
+++ b/openhands-sdk/openhands/sdk/hooks/config.py
@@ -49,6 +49,11 @@ class HookDefinition(BaseModel):
     type: HookType = HookType.COMMAND
     command: str
     timeout: int = 60
+    async_: bool = Field(default=False, alias="async")  # 'async' is a reserved keyword
+
+    model_config = {
+        "populate_by_name": True,  # Allow both 'async' and 'async_' in input
+    }
 
 
 class HookMatcher(BaseModel):
--- a/openhands-sdk/openhands/sdk/hooks/executor.py
+++ b/openhands-sdk/openhands/sdk/hooks/executor.py
@@ -1,8 +1,11 @@
 """Hook executor - runs shell commands with JSON I/O."""
 
 import json
+import logging
 import os
+import signal
 import subprocess
+import time
 
 from pydantic import BaseModel
 
@@ -26,6 +29,7 @@ class HookResult(BaseModel):
     reason: str | None = None
     additional_context: str | None = None
     error: str | None = None
+    async_started: bool = False  # Indicates this was an async hook
 
     @property
     def should_continue(self) -> bool:
@@ -37,11 +41,86 @@ class HookResult(BaseModel):
         return True
 
 
+logger = logging.getLogger(__name__)
+
+
+class AsyncProcessManager:
+    """Manages background hook processes for cleanup.
+
+    Tracks async hook processes and ensures they are terminated when they
+    exceed their timeout or when the session ends. Prevents zombie processes
+    by properly waiting for termination.
+    """
+
+    def __init__(self):
+        self._processes: list[tuple[subprocess.Popen, float, int]] = []
+
+    def add_process(self, process: subprocess.Popen, timeout: int) -> None:
+        """Track a background process for cleanup.
+
+        Args:
+            process: The subprocess to track
+            timeout: Maximum runtime in seconds before termination
+        """
+        self._processes.append((process, time.time(), timeout))
+
+    def _terminate_process(self, process: subprocess.Popen) -> None:
+        """Safely terminate a process group and prevent zombies.
+
+        Uses process groups to kill the entire process tree, not just
+        the parent shell when shell=True is used.
+        """
+        try:
+            # Kill the entire process group (handles shell=True child processes)
+            pgid = os.getpgid(process.pid)
+        except (OSError, ProcessLookupError) as e:
+            logger.debug(f"Process already terminated: {e}")
+            return
+
+        try:
+            os.killpg(pgid, signal.SIGTERM)
+            process.wait(timeout=1)  # Wait for graceful termination
+        except subprocess.TimeoutExpired:
+            try:
+                os.killpg(pgid, signal.SIGKILL)  # Force kill if it doesn't terminate
+                process.wait()
+            except OSError:
+                pass
+        except OSError as e:
+            logger.debug(f"Failed to kill process group: {e}")
+
+    def cleanup_expired(self) -> None:
+        """Terminate processes that have exceeded their timeout."""
+        current_time = time.time()
+        active: list[tuple[subprocess.Popen, float, int]] = []
+        for process, start_time, timeout in self._processes:
+            if process.poll() is None:  # Still running
+                if current_time - start_time > timeout:
+                    logger.debug(f"Terminating expired async hook (PID {process.pid})")
+                    self._terminate_process(process)
+                else:
+                    active.append((process, start_time, timeout))
+            # If poll() returns non-None, process already exited - just drop it
+        self._processes = active
+
+    def cleanup_all(self) -> None:
+        """Terminate all tracked background processes."""
+        for process, _, _ in self._processes:
+            if process.poll() is None:
+                self._terminate_process(process)
+        self._processes = []
+
+
 class HookExecutor:
     """Executes hook commands with JSON I/O."""
 
-    def __init__(self, working_dir: str | None = None):
+    def __init__(
+        self,
+        working_dir: str | None = None,
+        async_process_manager: AsyncProcessManager | None = None,
+    ):
         self.working_dir = working_dir or os.getcwd()
+        self.async_process_manager = async_process_manager or AsyncProcessManager()
 
     def execute(
         self,
@@ -64,8 +143,50 @@ class HookExecutor:
         # Serialize event to JSON for stdin
         event_json = event.model_dump_json()
 
+        # Cleanup expired async processes before starting new ones
+        self.async_process_manager.cleanup_expired()
+
+        # Handle async hooks: fire and forget
+        if hook.async_:
+            try:
+                process = subprocess.Popen(
+                    hook.command,
+                    shell=True,
+                    cwd=self.working_dir,
+                    env=hook_env,
+                    stdin=subprocess.PIPE,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                    start_new_session=True,  # Create new process group for cleanup
+                )
+                # Write event JSON to stdin safely
+                try:
+                    if process.stdin and process.poll() is None:
+                        process.stdin.write(event_json.encode())
+                        process.stdin.flush()
+                        process.stdin.close()
+                except (BrokenPipeError, OSError) as e:
+                    logger.warning(f"Failed to write to async hook stdin: {e}")
+
+                # Track for cleanup
+                self.async_process_manager.add_process(process, hook.timeout)
+                logger.debug(f"Started async hook (PID {process.pid}): {hook.command}")
+
+                # Return placeholder success result
+                return HookResult(
+                    success=True,
+                    exit_code=0,
+                    async_started=True,
+                )
+            except Exception as e:
+                return HookResult(
+                    success=False,
+                    exit_code=-1,
+                    error=f"Failed to start async hook: {e}",
+                )
+
         try:
-            # Execute the hook command
+            # Execute the hook command synchronously
             result = subprocess.run(
                 hook.command,
                 shell=True,
@@ -145,6 +266,9 @@ class HookExecutor:
     ) -> list[HookResult]:
         """Execute multiple hooks in order, optionally stopping on block."""
         results: list[HookResult] = []
+
+        # Cleanup expired async processes periodically
+        self.async_process_manager.cleanup_expired()
 
         for hook in hooks:
             result = self.execute(hook, event, env)
--- a/openhands-sdk/openhands/sdk/hooks/manager.py
+++ b/openhands-sdk/openhands/sdk/hooks/manager.py
@@ -1,10 +1,14 @@
 """Hook manager - orchestrates hook execution within conversations."""
 
+import logging
 from typing import Any
 
 from openhands.sdk.hooks.config import HookConfig
 from openhands.sdk.hooks.executor import HookExecutor, HookResult
 from openhands.sdk.hooks.types import HookEvent, HookEventType
+
+
+logger = logging.getLogger(__name__)
 
 
 class HookManager:
@@ -51,6 +55,14 @@ class HookManager:
         hooks = self.config.get_hooks_for_event(HookEventType.PRE_TOOL_USE, tool_name)
         if not hooks:
             return True, []
+
+        # Warn about async hooks in PreToolUse - they cannot block operations
+        async_hooks = [h for h in hooks if h.async_]
+        if async_hooks:
+            logger.warning(
+                "Async hooks in PreToolUse cannot block tool execution. "
+                f"Found {len(async_hooks)} async hook(s) that will run in background."
+            )
 
         event = self._create_event(
             HookEventType.PRE_TOOL_USE,
@@ -127,11 +139,19 @@ class HookManager:
     def run_session_end(self) -> list[HookResult]:
         """Run SessionEnd hooks when a conversation ends."""
         hooks = self.config.get_hooks_for_event(HookEventType.SESSION_END)
-        if not hooks:
-            return []
+        results: list[HookResult] = []
+        if hooks:
+            event = self._create_event(HookEventType.SESSION_END)
+            results = self.executor.execute_all(hooks, event, stop_on_block=False)
 
-        event = self._create_event(HookEventType.SESSION_END)
-        return self.executor.execute_all(hooks, event, stop_on_block=False)
+        # Cleanup any background async processes
+        self.cleanup_async_processes()
+
+        return results
+
+    def cleanup_async_processes(self) -> None:
+        """Cleanup all background hook processes."""
+        self.executor.async_process_manager.cleanup_all()
 
     def run_stop(
         self,
--- a/tests/github_workflows/test_resolve_model_config.py
+++ b/tests/github_workflows/test_resolve_model_config.py
@@ -1,5 +1,6 @@
 """Tests for resolve_model_config.py GitHub Actions script."""
 
+import subprocess
 import sys
 from pathlib import Path
 from typing import Any
@@ -273,9 +274,7 @@ class TestTestModel:
         mock_response = MagicMock()
         mock_response.choices = [MagicMock(message=MagicMock(content="OK"))]
 
-        with patch(
-            "resolve_model_config.litellm.completion", return_value=mock_response
-        ):
+        with patch("litellm.completion", return_value=mock_response):
             success, message = test_model(model_config, "test-key", "https://test.com")
 
         assert success is True
@@ -291,9 +290,7 @@ class TestTestModel:
         mock_response = MagicMock()
         mock_response.choices = [MagicMock(message=MagicMock(content=""))]
 
-        with patch(
-            "resolve_model_config.litellm.completion", return_value=mock_response
-        ):
+        with patch("litellm.completion", return_value=mock_response):
             success, message = test_model(model_config, "test-key", "https://test.com")
 
         assert success is False
@@ -310,7 +307,7 @@ class TestTestModel:
         }
 
         with patch(
-            "resolve_model_config.litellm.completion",
+            "litellm.completion",
             side_effect=litellm.exceptions.Timeout(
                 message="Timeout", model="test-model", llm_provider="test"
             ),
@@ -331,7 +328,7 @@ class TestTestModel:
         }
 
         with patch(
-            "resolve_model_config.litellm.completion",
+            "litellm.completion",
             side_effect=litellm.exceptions.APIConnectionError(
                 message="Connection failed", llm_provider="test", model="test-model"
             ),
@@ -352,7 +349,7 @@ class TestTestModel:
         }
 
         with patch(
-            "resolve_model_config.litellm.completion",
+            "litellm.completion",
             side_effect=litellm.exceptions.NotFoundError(
                 "Model not found", llm_provider="test", model="test-model"
             ),
@@ -376,9 +373,7 @@ class TestTestModel:
         mock_response = MagicMock()
         mock_response.choices = [MagicMock(message=MagicMock(content="OK"))]
 
-        with patch(
-            "resolve_model_config.litellm.completion", return_value=mock_response
-        ) as mock_completion:
+        with patch("litellm.completion", return_value=mock_response) as mock_completion:
             test_model(model_config, "test-key", "https://test.com")
 
         mock_completion.assert_called_once()
@@ -420,9 +415,7 @@ class TestRunPreflightCheck:
         mock_response.choices = [MagicMock(message=MagicMock(content="OK"))]
 
         with patch.dict("os.environ", {"LLM_API_KEY": "test"}):
-            with patch(
-                "resolve_model_config.litellm.completion", return_value=mock_response
-            ):
+            with patch("litellm.completion", return_value=mock_response):
                 result = run_preflight_check(models)
 
         assert result is True
@@ -442,9 +435,50 @@ class TestRunPreflightCheck:
             return mock_response
 
         with patch.dict("os.environ", {"LLM_API_KEY": "test"}):
-            with patch(
-                "resolve_model_config.litellm.completion", side_effect=mock_completion
-            ):
+            with patch("litellm.completion", side_effect=mock_completion):
                 result = run_preflight_check(models)
 
         assert result is False
+
+
+def test_models_importable_without_litellm():
+    """Test that MODELS dictionary can be imported without litellm installed.
+
+    This is critical for the integration-runner workflow which uses MODELS
+    in the setup-matrix job without installing litellm. The import should
+    work in a clean Python environment.
+
+    Regression test for issue #2124.
+    """
+    # Get the repository root (where .github/ is located)
+    repo_root = Path(__file__).parent.parent.parent
+
+    script = """
+import sys
+sys.path.insert(0, '.github/run-eval')
+
+# This import should succeed without litellm being installed
+from resolve_model_config import MODELS
+
+# Verify we got the MODELS dictionary
+assert isinstance(MODELS, dict)
+assert len(MODELS) > 0
+print(f"SUCCESS: Imported {len(MODELS)} models without litellm")
+"""
+
+    # Run the script in a subprocess with a clean environment
+    # This ensures litellm is not available in sys.modules
+    result = subprocess.run(
+        [sys.executable, "-c", script],
+        capture_output=True,
+        text=True,
+        cwd=repo_root,
+    )
+
+    # Check that the script succeeded
+    assert result.returncode == 0, (
+        f"Failed to import MODELS without litellm.\n"
+        f"stdout: {result.stdout}\n"
+        f"stderr: {result.stderr}"
+    )
+    assert "SUCCESS" in result.stdout
--- a/tests/sdk/hooks/test_config.py
+++ b/tests/sdk/hooks/test_config.py
@@ -229,3 +229,83 @@ class TestHookConfig:
             HookConfig.from_dict(
                 {"PreToolExecute": [{"hooks": [{"command": "test.sh"}]}]}
             )
+
+
+class TestAsyncHooks:
+    """Tests for async hook configuration."""
+
+    def test_async_field_defaults_false(self):
+        """Test that async defaults to False."""
+        hook = HookDefinition(command="echo test")
+        assert hook.async_ is False
+
+    def test_async_field_set_true(self):
+        """Test that async can be set to True using async alias."""
+        hook = HookDefinition.model_validate({"command": "echo test", "async": True})
+        assert hook.async_ is True
+
+    def test_async_field_parsed_from_json_alias(self):
+        """Test that 'async' key in JSON is parsed correctly via alias."""
+        data = {
+            "hooks": {
+                "PostToolUse": [
+                    {"matcher": "*", "hooks": [{"command": "test.sh", "async": True}]}
+                ]
+            }
+        }
+        config = HookConfig.from_dict(data)
+        hooks = config.get_hooks_for_event(HookEventType.POST_TOOL_USE, "AnyTool")
+        assert len(hooks) == 1
+        assert hooks[0].async_ is True
+
+    def test_async_field_serialization_by_alias(self):
+        """Test that async field serializes correctly using alias."""
+        hook = HookDefinition.model_validate({"command": "test.sh", "async": True})
+        output = hook.model_dump(mode="json", by_alias=True)
+        assert output["async"] is True
+        assert "async_" not in output
+
+    def test_async_field_serialization_without_alias(self):
+        """Test that async field serializes as async_ without by_alias."""
+        hook = HookDefinition.model_validate({"command": "test.sh", "async": True})
+        output = hook.model_dump(mode="json")
+        assert output["async_"] is True
+
+    def test_async_hook_in_config_round_trip(self):
+        """Test that async hooks survive a JSON round-trip."""
+        data = {
+            "PostToolUse": [
+                {
+                    "matcher": "terminal",
+                    "hooks": [
+                        {"command": "sync-hook.sh", "async": False},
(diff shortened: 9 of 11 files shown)


csmith49:
> 🔴 Bug: ToolCallMatchingProperty treats None as a valid tool_call_id, allowing bogus matches. Old code explicitly filtered None values out. This changes behavior and breaks an edge case the old tests covered.

I noticed this while cleaning up the tests, but I'm confused why it's an issue in the first place. `ActionEvent` and `ObservationBaseEvent` both have a non-optional `tool_call_id` field, so these values should never be `None` in the first place.

ETA: git blame puts me as the person to write the `test_filter_unmatched_tool_calls_none_tool_call_id` test, so I take responsibility for this. The test as written diverges from the implementation, though, as it uses `autospec` to create event mocks whose type information is lost. Writing that test without mocks is impossible -- pyright throws a type error immediately.

csmith49:
> 🟠 Breaking Changes: Two API changes that need documentation:
> 
> View.manipulation_indices returns ManipulationIndices (set) not list[int]
> View.find_next_manipulation_index() removed, replaced by ManipulationIndices.find_next()

Documented in the PR description.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        # Start by collecting all tool call IDs associated with actions and observations
+        # separately.
+        action_tool_call_ids: set[ToolCallID] = set()
+        observation_tool_call_ids: set[ToolCallID] = set()
+
+        for event in current_view_events:
+            match event:
+                case ActionEvent():
+                    action_tool_call_ids.add(event.tool_call_id)
```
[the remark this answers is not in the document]
csmith49: `ActionEvent.tool_call_id` cannot be `None` without throwing a type error.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
```diff
@@ -0,0 +1,87 @@
...
+                case ObservationBaseEvent():
+                    observation_tool_call_ids.add(event.tool_call_id)
```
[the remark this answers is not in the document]
csmith49: `ObservationBaseEvent.tool_call_id` cannot be `None` without throwing a type error.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
```diff
@@ -0,0 +1,87 @@
...
+        # Actions always come before observations, so we can maintain a set of pending
+        # tool calls -- these are any tool calls that have been introduced by an action
+        # but not yet resolved by an observation. If there are any pending tool calls we
+        # know we're between an action/observation pair.
+        pending_tool_call_ids: set[ToolCallID] = set()
+
+        for index, event in enumerate(current_view_events):
+            match event:
+                case ActionEvent():
+                    pending_tool_call_ids.add(event.tool_call_id)
```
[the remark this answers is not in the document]
csmith49: `ActionEvent.tool_call_id` cannot be `None` without throwing a type error.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
```diff
@@ -0,0 +1,87 @@
...
+                case ObservationBaseEvent():
+                    pending_tool_call_ids.remove(event.tool_call_id)
```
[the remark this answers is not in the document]
csmith49: `ObservationBaseEvent.tool_call_id` cannot be `None` without throwing a type error.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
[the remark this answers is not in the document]
csmith49: No need to filter out `None` values, `ActionEvent.tool_call_id` cannot be `None`. The type is `ToolCallID` -- no `Optional` or ` | None` present.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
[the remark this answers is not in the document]
csmith49: `ObservationBaseEvent.tool_call_id` cannot be `None` without throwing a type error.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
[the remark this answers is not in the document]
csmith49: `ActionEvent.tool_call_id` cannot be `None`, according to pyright.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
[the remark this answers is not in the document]
csmith49: The tool call ID cannot be `None` according to pyright.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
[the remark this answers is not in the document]
csmith49: The tool call ID cannot be `None`, according to pyright.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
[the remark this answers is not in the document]
csmith49: The tool call ID cannot be `None`, according to pyright.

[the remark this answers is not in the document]
csmith49: The tool call ID cannot be `None`, according to pyright.

openhands-sdk/openhands/sdk/context/view/manipulation_indices.py
```diff
@@ -0,0 +1,52 @@
...
+        Returns:
+            The smallest manipulation index greater than or equal to the threshold.
+
+        Raises:
+            ValueError: if no valid manipulatin index exists past the threshold.
+        """
+        valid_indices = {idx for idx in self if idx >= threshold}
+
+        if not valid_indices:
+            raise ValueError(f"No manipulation index found >= {threshold}.")
```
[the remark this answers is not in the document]
csmith49: This isn't intentional. I've modified the LLM summarizing condenser to catch those value errors (which shouldn't happen, our current manipulation indices always have the end of the events in them) so that when they happen the conversation can continue without condensation like we do when we find no events to forget.

openhands-sdk/openhands/sdk/context/view/manipulation_indices.py
```diff
@@ -0,0 +1,52 @@
...
+        Returns:
+            The smallest manipulation index greater than or equal to the threshold.
+
+        Raises:
+            ValueError: if no valid manipulation index exists past the threshold.
+        """
+        valid_indices = {idx for idx in self if idx >= threshold}
+
+        if not valid_indices:
+            raise ValueError(f"No manipulation index found >= {threshold}.")
```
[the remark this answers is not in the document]
csmith49: The old implementation was bugged. When we return the threshold, we end up using indices to manipulate the view that don't appear in the `ManipulationIndices`. Raising `ValueError` makes this failure case explicit.

There are no known situations where the condenser will get manipulation indices that _don't_ contain an index at the end of the events. But the `ManipulationIndices` class doesn't know that, so it is correct to raise a value error there and catch it in the caller.

openhands-sdk/openhands/sdk/context/view/view.py
```diff
@@ -0,0 +1,164 @@
...
+        enforced a warning is logged.
+        """
+        for property in ALL_PROPERTIES:
+            events_to_forget = property.enforce(current_view_events, all_events)
+            if events_to_forget:
+                logger.warning(
+                    f"Property {property.__class__} enforced, "
+                    f"{len(events_to_forget)} events dropped."
+                )
+                return View.enforce_properties(
```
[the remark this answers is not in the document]
csmith49: This is overly defensive programming. In practice the largest set of view events to be manipulated is well below 300, an order of magnitude below the default recursion limit.

Doing this just to raise a runtime error is silly -- Python will _already_ throw an error if the recursion limit is reached. We're just stopping the execution way earlier for no real reason.

csmith49:
To any human reviewers: apologies for the LoC changed here, I could not find a clean way to make this PR any smaller. Fortunately a significant amount of those changes are tests, most of which are just reorganized and slightly cleaned.

I'd recommend starting a review from the `ViewPropertyBase` to see the interface, then looking at the properties themselves, then looking at how the `View` is changed to support those properties. I'm hoping the properties are way easier to follow and understand than the mess of functions in the `View` that they replace.

enyst:
@openhands Do a /codereview-roasted on this PR. But don't exaggerate. Think deeply.

Post your review as a comment to this PR. Note that it will be rendered as markdown.

After you post, re-read your review comment, and look at it and at the PR diff with *fresh eyes* critically: take each review section and re-analyze if it's accurate, or exaggerated, or more minor or more major than it seemed at first sight. Post your re-analysis as a new comment, as markdown.

enyst:
**Taste rating:** 🔴 **Needs improvement**

This refactor is aiming at a real problem (the old `View` had a lot of policy embedded in it), and the general direction is good: properties as first-class objects is a clean separation.

But there are a couple of places where the *data structure choice* and *behavioral guarantees* don’t line up, and that’s where you get the kind of “works in tests, breaks in production” bugs.

---

## [CRITICAL ISSUES]

### 1) `ManipulationIndices` is a `set`, but the API is still *semantically ordered*
- **Files:**
  - `openhands-sdk/openhands/sdk/context/view/manipulation_indices.py` (class extends `set[int]`)
  - `tests/integration/tests/c01_thinking_block_condenser.py:52` (`indices = list(view.manipulation_indices)`)
  - `tests/sdk/context/view/test_view_tool_loop_boundaries.py:209` (`assert list(indices) == [0, 1, 2]`)

A “manipulation index” is fundamentally a *boundary in a sequence*. The whole idea of “ranges between consecutive indices define atomic units” only makes sense if you have a **sorted** sequence of indices.

Right now:
- `View.manipulation_indices` returns a `set`.
- Code/tests convert it to `list(...)` and then treat adjacent elements as ordered boundaries.

That’s not a nit: a set’s iteration order is not an API contract. Even if it looks stable for small ints, it’s accidental.

**Actionable fix options:**
1. Keep the internal representation as a set, but make iteration deterministic and ordered:
   - override `__iter__` on `ManipulationIndices` to yield `sorted(super())`, and/or
   - add an explicit `sorted()`/`as_list_sorted()` method and update all call sites to use it.
2. Or don’t pretend it’s a set: keep it as `list[int]` (sorted) and use `set(...)` only where you need set operations.

If you want the type to be a set for intersection (`&=`), fine — but then you need an *explicit* ordered view for any logic that treats indices as boundaries.

---

### 2) `ToolCallMatchingProperty` regresses the old “`None` tool_call_id is invalid” behavior
- **File:** `openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py`

The old `View` code had explicit `tool_call_id is not None` filtering when building the action/observation tool-call-id sets. That wasn’t there for fun — it was defensive against bad/legacy data.

The new enforcement logic effectively treats `None` as just another ID:
- `None` ends up in both sets if present in both event types
- then an action+observation with `tool_call_id=None` will “match” and survive enforcement

Even if the type system says `ToolCallID = str`, the old code suggests there’s a real-world edge case here (and a reviewer already flagged it).

**Actionable fix:** treat `None` as “unmatchable”:
- don’t add `None` to the tool_call_id sets
- and/or explicitly mark events with `tool_call_id is None` for removal

Also worth noting: `manipulation_indices()` uses `pending_tool_call_ids.remove(...)`, which will throw if the stream is already malformed. If this is truly “impossible” post-enforcement, document that assumption; otherwise, make it robust.

---

### 3) `ToolLoopAtomicityProperty` semantics look broader than the previous implementation
- **File:** `openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py`

Old behavior (per the removed `View.manipulation_indices` doc + logic): tool loops were tied to **thinking-block batches**.

New behavior: a “tool loop” is **any** contiguous run of action/observation events.

That’s a meaningful behavior change: it can dramatically shrink available manipulation ranges and cause condensers to hit the “no valid range” path more often — even for models that don’t have the thinking-block checksum constraints.

Maybe that’s intentional (“safer by default”), but then it should be called out as such because it affects condensation availability.

**Actionable fix:** either
- restore the “thinking-block starts the loop” rule, or
- keep the broader rule but explicitly document the change + expected impact, or
- make loading this property conditional (as your PR description mentions as a future improvement).

---

- **`View.enforce_properties` recursion** (`openhands-sdk/openhands/sdk/context/view/view.py:95+`):
  The recursion will terminate (events strictly decrease), but it’s still an odd choice. A simple `while True` loop would be clearer and avoids any “oops, we dropped 2k events and hit recursion depth” surprises.

- **Mutable defaults in Pydantic models** (`View.condensations: list[Condensation] = []`):
  This was already present on `main`, but it’s still a footgun. Prefer `Field(default_factory=list)`.

---

## [TESTING GAPS]

- The integration test and `test_view_tool_loop_boundaries.py` currently *depend on set ordering* (see critical issue #1). They should be fixed to use an ordered form (or the type should provide one).

- There’s no test covering the `tool_call_id=None` edge case that the old implementation explicitly guarded against.

---

## VERDICT
**Not ready to merge as-is**: the `ManipulationIndices` ordering contract and the `tool_call_id=None` regression are the kind of “looks fine until it isn’t” issues.

Once those are addressed, the overall architecture (property isolation + tests organized by property) looks like a solid direction.

**Key insight:** right now the code *says* “set of safe boundaries” but the condensers/tests *use* it as an ordered boundary list. Pick one and make it explicit.

enyst:
### Re-read / self-audit of my review

After re-reading my own comment + skimming the diff again, I think the **ordering contract for `ManipulationIndices`** is the only truly “this will bite you” issue I’d keep at 🔴 severity.

A couple adjustments in severity/precision:

- **`tool_call_id=None` regression:** I called this out as a hard regression, but given the current type (`ToolCallID = str`) and Pydantic validation, it’s *plausible* that `None` can’t exist in real persisted events anymore (unless something is using `model_construct` / bypassing validation / mocks). The old code was defensive, so I still think it’s worth deciding explicitly:
  - either keep the defensive `None` handling (and add a test that demonstrates a real path where it can occur),
  - or remove the legacy behavior/tests and treat “`None` tool_call_id” as invalid data that can’t load.

- **Tool loop semantics:** This *is* a real behavior change vs the old “thinking-block-started loops” rule, but I may have framed it too negatively. If the goal is “safer by default across models,” enforcing atomicity on any contiguous tool run is defensible — it just needs to be an explicit, documented tradeoff because it reduces condensation flexibility.

- **`__iter__` / sequence semantics:** I didn’t call this out originally, and I’m not going to invent it now as a must-fix. The existing `View` already didn’t implement `__iter__` on `main`, so this PR isn’t introducing that regression. The real issue is that any *new* tests/code that treat `View` as directly iterable are silently wrong due to `BaseModel.__iter__` yielding `(field, value)` pairs.

Net: the main point stands — **don’t return an unordered container and then reason about “consecutive indices.”** Everything else is secondary / depends on intended semantics.

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class ToolLoopAtomicityProperty(ViewPropertyBase):
+    """A tool loop is a sequence of action/observation pairs, with nothing in between,
+    that some agents identify as a single turn.
+
+    This property is important to enforce for Anthropic models with thinking enabled.
+    They expect the first element of such a tool loop to have a thinking block, and use
+    some checksums to make sure it is correctly placed. In such a setup if we remove any
+    element of the tool loop we have to remove the whole thing.
```
enyst, line 21: Sorry, I know we've been looking at this before, but I still don't know that this has to be this way, and it seems ... not possible? Unless it simply means we need to summarize ~everything.

I'm confused how the first element could have signature generated from tool calls that weren't generated by the model yet... sorry to be dense. I know there was some weird stuff in Anthropic docs, but it may be worth considering if perhaps there are other ways of looking at it.

"Tool loop" seems to mean an "agent turn" - the model is just doing tool calls, we give results, continue. Such an agent turn can run from initial prompt to any number of events, including 500, 1k or 10k just the same. So we summarize the ~whole thing, which means... the whole `agent-only source` events.

In other words, it seems that maybe all we need to know is if there was a `user message` (real or synthetic)? If only the initial message -> summarize the rest. If there is another somewhere, we can summarize from start until that message(s)?

openhands-sdk/openhands/sdk/context/view/properties/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+from openhands.sdk.context.view.properties.batch_atomicity import BatchAtomicityProperty
+from openhands.sdk.context.view.properties.tool_call_matching import (
+    ToolCallMatchingProperty,
+)
+from openhands.sdk.context.view.properties.tool_loop_atomicity import (
+    ToolLoopAtomicityProperty,
+)
+
+
+ALL_PROPERTIES: list[ViewPropertyBase] = [
```
VascoSch92, line 11: Is that wanted?

Because you are actually instantiating the classes and not exposing closures.

openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py
```diff
@@ -0,0 +1,87 @@
...
+        """Enforce tool-call matching by removing actions without matching observations,
+        and vice versa.
+        """
+        # Start by collecting all tool call IDs associated with actions and observations
+        # separately.
+        action_tool_call_ids: set[ToolCallID] = set()
+        observation_tool_call_ids: set[ToolCallID] = set()
+
+        for event in current_view_events:
+            match event:
```
VascoSch92, line 32: Thanks for using `match` :-) it is super clear

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
```diff
@@ -0,0 +1,163 @@
...
+
+        Args:
+            events: A sequence of events. Must be in-order.
+
+        Returns:
+            A list of tool loops, each represented by a set of IDs corresponding to the
+            events in the loop.
+        """
+        tool_loops: list[set[EventID]] = []
+        current_tool_loop: set[EventID] | None = None
```
VascoSch92, line 35: why not just `current_tool_loop: set[EventID] = set()`?

I believe the logic stay the same and we don' have two types for one variable

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
```diff
@@ -0,0 +1,163 @@
...
+        all_tool_loops: list[set[EventID]] = self._tool_loops(all_events)
+
+        events_to_remove: set[EventID] = set()
+
+        for view_tool_loop in self._tool_loops(current_view_events):
+            # We assume the current view events (or at least the ones that make up tool
+            # loops) are a subset of all the events. If a tool loop in the view isn't
+            # present in the total list of tool loops that indicates some element has
+            # been forgotten and we have to remove the remaining elements from the view.
+            if view_tool_loop not in all_tool_loops:
```
VascoSch92: I’m not sure about the typical size of `all_tool_loops`, but if it starts to grow, we should consider modifying the method to return a set. 

Since the current implementation returns a list of sets (and sets aren't hashable), we could convert each set into a string and return a `set[str]` instead. 

This would reduce the complexity from `O(n**2)` to `O(n)`. 

If the length is always small, feel free to ignore this, better to keep it readable than to over-engineer it!

openhands-sdk/openhands/sdk/context/view/manipulation_indices.py
```diff
@@ -0,0 +1,52 @@
...
+    def complete(events: list[LLMConvertibleEvent]) -> ManipulationIndices:
+        """Returns a complete set of manipulation indices for a sequence of events.
+
+        This is equivalent to saying that manipulations can be done anywhere inside the
+        sequence without issue.
+        """
+        manipulation_indices = ManipulationIndices()
+
+        manipulation_indices.update(range(0, len(events)))
+        manipulation_indices.add(len(events))
```
VascoSch92, line 50:
```suggestion
        manipulation_indices.update(range(0, len(events) + 1))
```

Or I think taht you can just make the method 1-line

```python
def complete(events: ...) -> ManipulationIndices:
    return ManipulationIndices(range(0, len(events) + 1))
```

openhands-sdk/openhands/sdk/context/view/view.py
```diff
@@ -0,0 +1,164 @@
...
+    input. Also contains fields with information from the condensation process to aid
+    in deciding whether further condensation is needed.
+    """
+
+    events: list[LLMConvertibleEvent]
+
+    unhandled_condensation_request: bool = False
+    """Whether there is an unhandled condensation request in the view."""
+
+    condensations: list[Condensation] = []
```
VascoSch92: Pydantic is doing `Field(default_factory=list)` right? Just to be sure...

openhands-sdk/openhands/sdk/context/view/view.py
```diff
@@ -0,0 +1,164 @@
...
+
+            # If the event isn't related to condensation and isn't LLMConvertible, it
+            # should not be in the resulting view. Examples include certain internal
+            # events used for state tracking that the LLM does not need to see -- see,
+            # for example, ConversationStateUpdateEvent, PauseEvent, and (relevant here)
+            # CondensationRequest.
+            else:
+                logger.debug(
+                    f"Skipping non-LLMConvertibleEvent of type {type(event)} "
+                    f"in View.from_events"
```
VascoSch92:
```suggestion
                    "in View.from_events"
```

actually there is a Ruff rule for this kind of nit that we could enforce...

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
csmith49: Opus generates a thinking block for the first message in an agent turn, and will throw some API exceptions if it sees an agent turn without a thinking block.

The checksums just ensure the thinking blocks it sees were actually generated by Anthropic when it emitted that particular message, which means we can't modify the thinking block or put it anywhere else without Anthropic complaining.

So we can't get rid of the thinking blocks and we can't move or change them. We _can_ modify the suffix of the agent turns (anything past the first action/observation pair) when we condense, but that isn't a very useful condensation strategy to us at the moment. The result is we treat agent turns as atomic.

If the history is just one long agent turn, yeah, we have to summarize the whole thing. Otherwise condensation will "snap" to the agent turn / user message boundaries.

To be clear, I don't like this. I believe our compact-the-prefix-and-keep-the-suffix approach is better for keeping agents on track than Anthropic's all-or-nothing compaction, but they're optimizing for Claude Code usage and this was the best solution I could find at the time.

openhands-sdk/openhands/sdk/context/view/properties/__init__.py
csmith49: I think this is fine? There's no state associated with the properties, so while we could instantiate them in the view every time this makes them effectively singletons and saves some object creation cycles.

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
csmith49: I like having the explicit `None` to indicate we're not in a tool loop instead of relying on an empty set for that check. The logic is basically the same but the extra typing requirements make the individual cases more clear (at least to my eyes).

openhands-sdk/openhands/sdk/context/view/view.py
csmith49: It wasn't. We never make views without explicitly setting the `condensations` field so this mutable default was never an issue but I've made it use the default factory pattern now just to be sure.

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
csmith49: This might be slightly pre-mature optimization. You're right that the check can be optimized, but our more substantial optimization is going to be when we eliminate the dependency on `all_events` in the first place. We'll have to push some of this indexing deeper into the conversation state, but there's no reason we shouldn't be incrementally computing and storing tool loops and batches there instead of recomputing them each time in the view.

csmith49:
> I'd love it if we can give it a little thought, or maybe I just need to understand a bit better, because the most important change, it seems to me, is that now we would force all other LLMs to get their events summarized if they were in an agent's turn ("tool loop").

Good catch, definitely unintended behavior. I've added the thinking block requirements back to the tool loop so now it only triggers for Anthropic models.

> On the other hand, if we do this... does that mean we could just do full view? All the time, maybe?...If I understand correctly and those must have been full event view condensations, then how much reason is there left, to do less than full view?

Depends on the event distribution, but yeah, Opus has been operating like this for at least a month. I still think our condensation approach is better than collapsing the whole view when the models support it though. Our earlier experiments definitely indicated a slight performance boost for weaker models.

csmith49:
> After re-reading my own comment + skimming the diff again, I think the ordering contract for ManipulationIndices is the only truly “this will bite you” issue I’d keep at 🔴 severity.

Reasonable. I caught earlier examples of this but missed a few test cases that were still casting them to a list. Those tests now rely on set comparisons instead.

openhands-sdk/openhands/sdk/context/view/properties/base.py
```diff
@@ -0,0 +1,59 @@
...
+
+    The main way views are manipulated (beyond adding new events in the course of a
+    conversation) is in the condensers, which are designed to respect the manipulation
+    indices. That means properties should hold inductively, and manipulation indices
+    should be calculable purely from the events in the current view.
+
+    Enforcement is intended as a fallback mechanism to handle edge cases, bad data, or
+    unforeseen situations. Because enforcement assumes the view is in a bad state, it
+    often requires a much larger perspective on the events and therefore depends on a
+    sequence of _all_ events in the conversation.
```
enyst, line 25: I think maybe we need to try to not depend on all events in the conversation, maybe we have alternatives?
- fallback to forgetting all view minus `keep_first`
- fallback to system prompt + user message(s) + a summary of agent actions in the view
- or worse case, system prompt + some info that things went badly and all we know is that this was user's task, maybe summary, maybe ask the user for more

For example, I feel the cloud has difficulty keeping up with 2k events, sandboxes keep crashing and getting lost, I don't know the exact reason but I know I didn't manage to have a longer conversation for a while... (on v1)

openhands-sdk/openhands/sdk/context/view/properties/batch_atomicity.py
```diff
@@ -0,0 +1,88 @@
...
+        events_to_remove: set[EventID] = set()
+
+        for llm_response_id, view_batch_ids in self._build_batches(
+            current_view_events
+        ).items():
+            # We assume that the current view events are a strict subset of the elements
+            # of the all_events sequence -- if the batch ids in the view aren't exactly
+            # one-to-one with the batch ids generated by the all_events sequence, that
+            # can only mean something has been forgotten and we need to drop the entire
+            # batch.
```
enyst, line 43: Just a thought, not a review comment, FWIW, I don't know if we tested that this is really necessary, but maybe I'm missing it. It might be worth having a specific test, if we don't

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
enyst: > If the history is just one long agent turn, yeah, we have to summarize the whole thing. Otherwise condensation will "snap" to the agent turn / user message boundaries.

Exactly 🤔 

This might actually be a better mental model:
- we summarize the "event blocks" between user messages
- that might mean the whole view.

It's easy to understand or visualize, it seems to me (I'll come back to this, but it's not an issue for this PR, it's just trying to understand - specially the aspect where we might need information from outside the view...)

openhands-sdk/openhands/sdk/context/view/properties/batch_atomicity.py
csmith49: Batch atomicity? I assume it was needed at some point, it's a carry-over from v0. Probably model-specific behavior. With this PR we can disable it pretty easily by just not applying that property.

openhands-sdk/openhands/sdk/context/view/properties/base.py
csmith49: Agreed we want to streamline some of these operations to avoid dealing with the whole event stream. But enforcing the properties we have unfortunately _requires_ information from events outside the view.

Simple example: say I look at the current view and see a single action/observation pair. If I want to enforce batch atomicity, the only way I can know that the pair isn't part of a larger batch is to know that there are no other action events with matching `llm_response_id` values in the larger context.

There are a few solutions to avoiding doing that:
1. Compute some kind of batch map in the conversation state and expose that. We'd have to do that tracking for each property we want to enforce and make sure it gets propagated to the view.
2. Make sure the batch is never split in the first place. That's what the manipulation indices are trying to do.

In an ideal world we don't need the enforcement at all, just the manipulation indices. That's why every time enforcement happens it logs a warning. If we can verify that we don't see those warnings in practice, we can disable enforcement altogether (or only trigger it when we detect some error and need to recover).

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
csmith49: Just remember that mental model only holds for Anthropic endpoints when thinking is enabled. In all other situations the "event blocks" are fine-grained enough for us to summarize parts of them without the completion endpoint complaining.

openhands-sdk/openhands/sdk/context/view/properties/batch_atomicity.py
enyst: I'm not sure, I thought it was new:
- https://github.com/OpenHands/software-agent-sdk/pull/1450

It seemed the PR fixed `act2 / summary / obs2` , but I think it also dropped `act1` and `obs1`. Which I don't know if we had to or tested for.

I think it would be great to test with `act1 / obs1`, assuming that we still avoid inserting `summary` between 🤔

openhands-sdk/openhands/sdk/context/view/properties/tool_loop_atomicity.py
enyst: Sorry to interfere here, but maybe it's worth considering to not store `agent events` information, and instead store information about the `user messages`. I think maybe we get the same result, except `user messages` are much fewer and maybe it all becomes easier?

openhands-sdk/openhands/sdk/context/view/properties/batch_atomicity.py
csmith49: Comes from before that: #775

We need a better way to actually stress the API assumptions made by various providers, but I think that's beyond the scope of this PR

enyst (approved):
Thank you for this!

I do have a question re: previous results show that weaker LLMs do better with ~half events: when did we last test that, and what qualified then as "weaker" LLMs? If it's old enough, we really may want to re-eval/reconsider, the LLM capabilities went ahead for a few lifetimes. 😅 

I'd like to (try to) make a few notes:
- I'm starting to feel it's worth it to play with the idea of restating the `tool loops` constraint as a `user messages` constraint
- they're much fewer, we can know exactly what's between them (I hope?)
- we need to track `user messages` for a hook that is broken; maybe we could use that tracking for both features
- it might be worth considering to delete `keep_first` attribute; in V1 we no longer need 4 minimum, the additional environment info that used to be a `recall obs` is now a suffix of the system prompt so I think we need two: system and user. That may save us from those bugs when we try to enforce `keep_first` but bump into the first events of a `tool loop`
- we might want to test for batch atomicity, as noted, ~though on the other hand, I suspect you might be right it was in some other form happening in V0...~ 🤔 (update: you just showed it was older)
- it's worth IMHO to think whether the summary of all events should be exactly the same way as partial view; for example `codex-cli` adds to the prompt that a summary just happened and nudges the agent to re-investigate a little to get its bearings again
- I have seen in other parts of the codebase a need to know `the current view`, it could be great if we could somehow expose it

Maybe we could make issues for a few of these, to look into, if they don't sound totally off-base?

openhands-sdk/openhands/sdk/context/view/properties/batch_atomicity.py
enyst: I completely forgot that PR! And I reviewed it at the time 😭 

Thank you, yes I agree.

csmith49:
> Thank you for this!
> 
> I do have a question re: previous results show that weaker LLMs do better with ~half events: when did we last test that, and what qualified then as "weaker" LLMs? If it's old enough, we really may want to re-eval/reconsider, the LLM capabilities went ahead for a few lifetimes. 😅
> 
> I'd like to (try to) make a few notes:
> 
> * I'm starting to feel it's worth it to play with the idea of restating the `tool loops` constraint as a `user messages` constraint
> * they're much fewer, we can know exactly what's between them (I hope?)
> * we need to track `user messages` for a hook that is broken; maybe we could use that tracking for both features
> * it might be worth considering to delete `keep_first` attribute; in V1 we no longer need 4 minimum, the additional environment info that used to be a `recall obs` is now a suffix of the system prompt so I think we need two: system and user. That may save us from those bugs when we try to enforce `keep_first` but bump into the first events of a `tool loop`
> * we might want to test for batch atomicity, as noted, ~though on the other hand, I suspect you might be right it was in some other form happening in V0...~ 🤔 (update: you just showed it was older)
> * it's worth IMHO to think whether the summary of all events should be exactly the same way as partial view; for example `codex-cli` adds to the prompt that a summary just happened and nudges the agent to re-investigate a little to get its bearings again
> * I have seen in other parts of the codebase a need to know `the current view`, it could be great if we could somehow expose it
> 
> Maybe we could make issues for a few of these, to look into, if they don't sound totally off-base?

I think some of these are totally warranted. My mind is on some short-term tasks this PR unblocked, let me sit on this for a day or so and I'll get back to you.

csmith49:
> I do have a question re: previous results show that weaker LLMs do better with ~half events: when did we last test that, and what qualified then as "weaker" LLMs? If it's old enough, we really may want to re-eval/reconsider, the LLM capabilities went ahead for a few lifetimes. 😅

Oh, this would have been way back when the condenser strategies were first being tested. We're due for a re-evaluation. It's been on my radar but haven't had any free cycles to get it set up.
 
> * I'm starting to feel it's worth it to play with the idea of restating the `tool loops` constraint as a `user messages` constraint
> * they're much fewer, we can know exactly what's between them (I hope?)
> * we need to track `user messages` for a hook that is broken; maybe we could use that tracking for both features

Tracking user messages seems helpful, but I'd be careful about restating the tool loop constraint as a negative like that. Could be a useful optimization to help detect boundaries but we'd still have to double-check the tool loop is actually a tool loop (action and observation events only, starts with thinking blocks, none in between, etc.)

> * it might be worth considering to delete `keep_first` attribute; in V1 we no longer need 4 minimum, the additional environment info that used to be a `recall obs` is now a suffix of the system prompt so I think we need two: system and user. That may save us from those bugs when we try to enforce `keep_first` but bump into the first events of a `tool loop`

The default for `keep_first` is now 2. Some users have saved CLI settings that haven't updated, unfortunately.

We could remove, but we definitely want to keep at least the system prompt event. So...maybe keep the attribute and set it to 1 if that's the behavior we want?

> * it's worth IMHO to think whether the summary of all events should be exactly the same way as partial view; for example `codex-cli` adds to the prompt that a summary just happened and nudges the agent to re-investigate a little to get its bearings again

Good point, our "all event" summary is handled by the hard context reset summary generation so we've already got an entry point to specialize the behavior a bit.

> * I have seen in other parts of the codebase a need to know `the current view`, it could be great if we could somehow expose it

Started working on #2141 yesterday (which I notice you've already seen).

PR #2132, opened 2026-02-19
feat: Add ACPAgent for Agent Client Protocol integration

### LINKED ISSUE #590: Implement a special `ACPAgent` which is a Agent-Client Protocol Client?
**Context/Motivation**

What if we change CLI and GUI as ACP clients to interact with all these agents based on the Agent-Client Protocol (https://agentclientprotocol.com/protocol/overview)? 

In that case, we can possibly re-use Claude-Code (https://github.com/zed-industries/claude-code-acp), the Gemini CLI ACP that Zed already implemented, and launch these different agents inside our sandbox and treat them as an agent there 🤔 

This should allow us to leverage the ACP ecosystem and position OpenHands as a model-agnostic agent platform (e.g., you can even use your Claude-code subscription / Codex / Gemini CLI subscription with OpenHands via ACP) — and all these agents are using OpenHands as a unified interface. 

This will actually allow us to directly integrate Claude-code/Gemini CLI into OpenHands & allowing users to utilize their Claude-code/Gemini CLI subscription via OpenHands.

Challenge: How do we make microagents, etc., work there? It will be a bit tricky unless we integrate the Agent-Client Protocol more deeply into the SDK.

**Approach**

One thing we can potentially do would be to create an ACPAgent that inherits from AgentBase -- we input the same stuff to the agent, but internally inside the agent, we implement the "agent" as an ACP Client & have Claude-Code/Gemini being an ACP Server.

Because we manage stuff like microagents at the conversation-state-level, we can make them work by assemble *inputs* to these different ACP servers

It is ok that we don't yet support most of the stuff within SDK - as long as we throw errors when user tries to set them.

simonrosenberg:
## Summary
- Add `ACPAgent`, an `AgentBase` subclass that delegates to ACP-compatible servers (Claude Code, Gemini CLI, etc.) instead of direct LLM calls
- The ACP server manages its own LLM, tools, and execution lifecycle; ACPAgent simply relays user messages and collects responses
- Includes stdout filtering for `claude-code-acp` v0.1.x which emits non-JSON-RPC log lines to stdout
- Optional dependency: `pip install 'openhands-sdk[acp]'` (requires `agent-client-protocol>=0.8.1`)

## Changes
| File | Description |
|---|---|
| `openhands-sdk/openhands/sdk/agent/acp_agent.py` | Main implementation: `_OpenHandsACPClient`, `_filter_jsonrpc_lines`, `ACPAgent` |
| `openhands-sdk/openhands/sdk/agent/__init__.py` | Lazy import for `ACPAgent` to avoid requiring `agent-client-protocol` at import time |
| `openhands-sdk/pyproject.toml` | Add `acp` optional dependency |
| `tests/sdk/agent/test_acp_agent.py` | 38 unit tests covering instantiation, serialization, validation, step, cleanup, and JSON-RPC filtering |
| `examples/01_standalone_sdk/36_acp_agent_example.py` | Minimal integration example |

## Test plan
- [x] 38 unit tests pass (`uv run pytest tests/sdk/agent/test_acp_agent.py -v`)
- [x] All 178 existing agent tests pass with zero regressions
- [x] Ruff lint passes on all modified/created files
- [x] Integration example runs end-to-end with real `claude-code-acp` server

Closes #590

### DIFF
--- a/examples/01_standalone_sdk/36_acp_agent_example.py
+++ b/examples/01_standalone_sdk/36_acp_agent_example.py
@@ -17,18 +17,20 @@ import os
 from openhands.sdk.agent import ACPAgent
 from openhands.sdk.conversation import Conversation
 
+
 agent = ACPAgent(acp_command=["npx", "-y", "claude-code-acp"])
 
-cwd = os.getcwd()
-conversation = Conversation(agent=agent, workspace=cwd)
+try:
+    cwd = os.getcwd()
+    conversation = Conversation(agent=agent, workspace=cwd)
 
-conversation.send_message(
-    "List the Python source files under openhands-sdk/openhands/sdk/agent/, "
-    "then read the __init__.py and summarize what agent classes are exported."
-)
-conversation.run()
-
-# Clean up the ACP server subprocess
-agent.close()
+    conversation.send_message(
+        "List the Python source files under openhands-sdk/openhands/sdk/agent/, "
+        "then read the __init__.py and summarize what agent classes are exported."
+    )
+    conversation.run()
+finally:
+    # Clean up the ACP server subprocess
+    agent.close()
 
 print("Done!")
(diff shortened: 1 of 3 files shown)


simonrosenberg:
Reopening from origin branch directly (no fork needed).

PR #2141, opened 2026-02-19
feat(view): Incremental view construction and conversation state view property

csmith49:
## Summary

This PR updates the conversation state with a persistent `View` object that is incrementally updated as events are recorded.

To support this, `View` creation has been refactored to be incremental. Events are added one at a time using the `View.add_event` function, which updates the view in-place and handles condensation application, unhandled condensation request tracking, and so on.

This should improve performance (we no longer need to rebuild a view from all events every time the agent takes a step) and make `View` a first-class object accessible wherever the conversation state is.

## Design Decisions

Views are maintained by `ConversationState` objects as a private field, and exposed as a property. This ensures the view is not serialized when the conversation state is saved. Instead, we rebuild the view from the whole list of events when the state is deserialized.

The idea is to avoid accidentally serializing events outside the file system store we maintain, but there is the tradeoff that loading conversation states might be slightly more expensive. This is an easy decision to change should the need arise.

## Performance "Benefits"

While reducing the number of `View.from_events` calls will technically improve performance, in most use cases we expect it will be incredibly minor gains.

While working on this PR I had a stack trace sampler monitoring the OpenHands ACP instance that was assisting me. On that trace, something like 99% of the samples from the `Agent.step` function were API calls to the LLM. Other traces have that number lower (to make room for critic calls and callback handlers), but `View` construction has never been more than 0.5% of the duration of `Agent.step`.

Point is, we're hugely network-bound.

The real benefit of this PR is making `View` a first-class object accessible to more of the system. It exposes precisely the list of events that are converted to messages and sent to the LLM, and so represents the agent's "attention window". That has to be helpful outside the condenser.

## Breaking Changes

* `View.condensations` field removed.
* `View.enforce_properties` changed from a static method to an in-place modification of the calling view.
* `prepare_llm_messages` no longer takes a list events as input. Instead of converting that list of events into a view it takes a view directly.

Lastly, because `View.enforce_properties` is only called when a view is constructed directly from a list of events (instead of being incrementally built), we now only enforce properties when conversation states are loaded from disk.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:b6e828d-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-b6e828d-python \
  ghcr.io/openhands/agent-server:b6e828d-python
```

```
ghcr.io/openhands/agent-server:b6e828d-golang-amd64
ghcr.io/openhands/agent-server:b6e828d-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:b6e828d-golang-arm64
ghcr.io/openhands/agent-server:b6e828d-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:b6e828d-java-amd64
ghcr.io/openhands/agent-server:b6e828d-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:b6e828d-java-arm64
ghcr.io/openhands/agent-server:b6e828d-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:b6e828d-python-amd64
ghcr.io/openhands/agent-server:b6e828d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:b6e828d-python-arm64
ghcr.io/openhands/agent-server:b6e828d-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:b6e828d-golang
ghcr.io/openhands/agent-server:b6e828d-java
ghcr.io/openhands/agent-server:b6e828d-python
```

- Each variant tag (e.g., `b6e828d-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `b6e828d-python-amd64`) are also available if needed

### DIFF
--- a/.github/workflows/pr-review-by-openhands.yml
+++ b/.github/workflows/pr-review-by-openhands.yml
@@ -2,17 +2,11 @@
 name: PR Review by OpenHands
 
 on:
-    # Use pull_request_target to allow fork PRs to access secrets when triggered by maintainers
-    # Security: This workflow runs when:
-    #   1. A new PR is opened (non-draft), OR
-    #   2. A draft PR is marked as ready for review, OR
-    #   3. A maintainer adds the 'review-this' label, OR
-    #   4. A maintainer requests openhands-agent or all-hands-bot as a reviewer
-    # Adding labels and requesting new reviewers requires write access. GitHub may also allow PR authors
-    # to re-request review from a previous reviewer.
-    # The PR code is explicitly checked out for review, but secrets are only accessible
-    # because the workflow runs in the base repository context
-    pull_request_target:
+    # TEMPORARY MITIGATION (Clinejection hardening)
+    #
+    # We temporarily avoid `pull_request_target` here. We'll restore it after the PR review
+    # workflow is fully hardened for untrusted execution.
+    pull_request:
         types: [opened, ready_for_review, labeled, review_requested]
 
 permissions:
@@ -22,18 +16,27 @@ permissions:
 
 jobs:
     pr-review:
-        # Run when one of the following conditions is met:
-        #   1. A new non-draft PR is opened by a non-first-time contributor, OR
-        #   2. A draft PR is converted to ready for review by a non-first-time contributor, OR
-        #   3. 'review-this' label is added, OR
-        #   4. openhands-agent or all-hands-bot is requested as a reviewer
-        # Note: FIRST_TIME_CONTRIBUTOR PRs require manual trigger via label/reviewer request
+        # Note: fork PRs will not have access to repository secrets under `pull_request`.
+        # Skip forks to avoid noisy failures until we restore a hardened `pull_request_target` flow.
         if: |
-            (github.event.action == 'opened' && github.event.pull_request.draft == false && github.event.pull_request.author_association != 'FIRST_TIME_CONTRIBUTOR' && github.event.pull_request.author_association != 'NONE') ||
-            (github.event.action == 'ready_for_review' && github.event.pull_request.author_association != 'FIRST_TIME_CONTRIBUTOR' && github.event.pull_request.author_association != 'NONE') ||
-            github.event.label.name == 'review-this' ||
-            github.event.requested_reviewer.login == 'openhands-agent' ||
-            github.event.requested_reviewer.login == 'all-hands-bot'
+            github.event.pull_request.head.repo.full_name == github.repository &&
+            (
+                github.event.pull_request.author_association == 'OWNER' ||
+                github.event.pull_request.author_association == 'MEMBER' ||
+                github.event.pull_request.author_association == 'COLLABORATOR'
+            ) &&
+            (
+                (github.event.action == 'opened' && github.event.pull_request.draft == false) ||
+                github.event.action == 'ready_for_review' ||
+                (github.event.action == 'labeled' && github.event.label.name == 'review-this') ||
+                (
+                    github.event.action == 'review_requested' &&
+                    (
+                        github.event.requested_reviewer.login == 'openhands-agent' ||
+                        github.event.requested_reviewer.login == 'all-hands-bot'
+                    )
+                )
+            )
         concurrency:
             group: pr-review-${{ github.event.pull_request.number }}
             cancel-in-progress: true


enyst:
@openhands Do a /codereview-roasted on this PR. Post your review as a comment on the PR. Note that it will be rendered as markdown.

Once you posted, take a step back and look *with fresh eyes*: re-read critically your review and the PR, and analyze every point of it, is it actually real, is it minor, is it major compared to what it seemed? Post your re-analysis as a new comment on the PR.

enyst:
## /codereview-roasted

### Taste rating: 🔴 Needs improvement

You took a relatively simple data model ("the event log is the source of truth") and replaced it with **two mutable sources of truth**: `EventLog` and a cached/derived `View` that must be kept in sync manually.

That can be fine *if you make it impossible to get out of sync*. Right now it’s very possible, and the failure mode is the worst kind: **silent wrong behavior**.

---

## [CRITICAL ISSUES]

- **ConversationState now has a cache without a hard invalidation story**
  - **Where:** `openhands-sdk/openhands/sdk/conversation/state.py` (view + add_event at ~206–218), `openhands-sdk/openhands/sdk/agent/agent.py` (uses `state.view` at ~246)
  - **Problem:** `state.events` is still exposed as a mutable `EventLog`, but only `ConversationState.add_event()` updates the view. Any code doing the old/obvious thing:

    ```py
    state.events.append(e)
    ```

    will now **silently** fail to update `state.view`, and therefore the next LLM call (which now uses `state.view`) will ignore those events.

    This is a behavioral break that won’t crash; it’ll just make agents “forget” things and make debugging miserable.

  - **Fix direction:** either (a) make `events` append go through `ConversationState.add_event` (proxy/wrapper), (b) stop exposing a mutable event log, or (c) compute the view lazily from events (invalidate on append) instead of keeping two independently-mutable structures.

- **`View.add_event()` can poison the view permanently after a bad Condensation**
  - **Where:** `openhands-sdk/openhands/sdk/context/view/view.py` (`add_event` ~130–169, enforcement ~100–129)
  - **Problem:** `add_event()` applies `Condensation.apply()` directly (`~153–156`) but does **not** re-run `enforce_properties()`. Since `Condensation.apply()` can forget arbitrary IDs, it can violate `ToolCallMatchingProperty` / `BatchAtomicityProperty` invariants.

    Previously, rebuilding a view via `from_events()` + enforcement was an automatic “self-heal” step. Now enforcement only happens on resume (`ConversationState.create(... resume ...)` rebuilds view). In other words: one malformed condensation can corrupt the runtime view for the rest of the process.

  - **Fix direction:** enforce properties (or at least validate) after applying a `Condensation`, or enforce at the `ConversationState.add_event()` layer when the appended event is a `Condensation`.

---

- **Breaking API churn with no compatibility shim**
  - **Where:** `openhands-sdk/openhands/sdk/agent/utils.py` (`prepare_llm_messages` now takes `View` at ~114–179)
  - **Problem:** This is a signature-breaking change in a utility function that downstream users *will* import.
  - **Fix direction:** accept both `View` and `Sequence[Event]` (deprecated path), or provide a transitional helper.

- **Manual caching + public mutability is a trap**
  - **Where:** `openhands-sdk/openhands/sdk/context/view/view.py` (`manipulation_indices` cache ~40–60)
  - **Problem:** you fixed invalidation for the mutations you control (`add_event`, `enforce_properties`), but `view.events` is still a public list. Any direct mutation makes the cache wrong again.
  - **Fix direction:** either treat `events` as effectively private (and document that), or stop caching something derived from a mutable list you don’t control.

- **Dead helper left behind**
  - **Where:** `View.unhandled_condensation_request_exists` (`view.py` ~84–98)
  - **Problem:** unused after the incremental refactor. It’s harmless, but it reads like unfinished surgery.

---

## [TESTING GAPS]

- The new `tests/sdk/conversation/local/test_state_view.py` is good coverage for the happy path.
- What’s missing is a test for the real regression risk introduced by the cache:
  - If `state.events.append()` is still “allowed”, verify the view stays correct (or explicitly assert it doesn’t and document that `add_event()` is required).

---

## VERDICT

❌ **Needs rework** before I’d be comfortable calling this “safer/faster”. The concept (incremental view) is fine; the current implementation relies on “everyone always uses the right mutation path” and that’s not an engineering strategy.

## KEY INSIGHT

You didn’t add a `View`; you added a **cache**. Caches must have a bulletproof invalidation model, or they become distributed lies.

enyst:
## Re-analysis of my /codereview-roasted (fresh eyes)

Re-reading both the PR and my own comment, here’s what I think is **actually important** vs what’s more “defense-in-depth / taste”.

### Still **major / real**

- **`ConversationState.events` vs `ConversationState.view` divergence is a genuine behavioral trap.**
  - This isn’t theoretical: the SDK continues to expose a mutable `EventLog` via `state.events`, but `Agent.step()` now uses `state.view` to build the LLM context.
  - If any downstream code has ever appended events directly to `state.events` (which used to be the obvious / only knob), those events will now be omitted from the LLM context *silently*.
  - Even if maintainers consider `state.events.append()` “internal”, the current API surface doesn’t communicate that, and the failure mode is non-obvious.

If only one thing gets addressed, it should be this: either make `events` effectively append-only *through* `ConversationState.add_event()` (proxy/wrapper), or make it impossible/explicitly unsupported to mutate `events` directly.

### Probably **medium / depends on invariants**

- **“Bad Condensation poisons view”**: this is more about robustness than an immediate bug.
  - In the *intended* flow, condensers should only forget along `manipulation_indices`, so properties should remain satisfied by construction.
  - However, the moment you accept that `Condensation.apply()` can remove arbitrary IDs (it can), then skipping any enforcement/validation after applying a `Condensation` means the view can become invalid until restart/resume.

So: I still think a post-condensation sanity/enforcement hook is a good idea, but I’d downgrade this from “guaranteed breakage” to “cheap guardrail that prevents future weirdness / version skew issues”.

### Probably **minor**

- **`prepare_llm_messages(events -> view)` signature break**: likely low impact inside this repo (only internal callers), but could break downstream imports. Whether that matters depends on how “public” you consider `openhands.sdk.agent.utils`.
- **Manual caching + mutable list**: real, but typical. If `View.events` is treated as internal, the cache is fine.
- **Unused helper (`unhandled_condensation_request_exists`)**: pure cleanup nit.

### Net

My original review wasn’t wrong about the core risk, but I was too harsh in implying the condensation enforcement change is *necessarily* a runtime bug. The **API/invariant mismatch between `events` and `view`** is the only point I’d call “must address”, because it’s the one that can bite real users without any obvious stacktrace.

enyst:
Thank you for this! And sorry I didn’t get to review yet, brain frozen. 🙏 

I would like to note here:

> ## Performance "Benefits"
> While reducing the number of `View.from_events` calls will technically improve performance, in most use cases we expect it will be incredibly minor gains.
> 
> While working on this PR I had a stack trace sampler monitoring the OpenHands ACP instance that was assisting me. On that trace, something like 99% of the samples from the `Agent.step` function were API calls to the LLM. Other traces have that number lower (to make room for critic calls and callback handlers), but `View` construction has never been more than 0.5% of the duration of `Agent.step`.
> 
> Point is, we're hugely network-bound.

I have not yet seen issues with the local SDK alone, though I don’t know up to what number of events. This is O(n) though, I think that will show.

Funny one was on V0, we had _one call_ to `list(events)` in the agent_controller, every user message. It was extremely noticeable on the cloud. The conversation didn’t need to be long, for people to notice and report it in slack repeatedly. [1]
(I remember it, because I wrote it, oops - the fix was just a few lines of code to stop O(n) loading)

That was last year, when to my knowledge, when conversations got to 2k events they qualified as “very long”. 😅 But I think we are not there anymore, by a… long shot. I think of 60k events as qualifying for long sessions now! 😂


[1] FWIW on V1 conversations I think max I had on the cloud was around ~2k, and sandboxes crashed, so for now I don’t know more. 😭

enyst:
@openhands Tell me what this PR is doing, how it works, what it changes, and what you think of it?

Respond to my question on github, using `gh`. One answer only.

enyst:
(OpenHands-GPT-5.2)

### What This PR Does

**Summary**: This PR introduces incremental  construction and adds a persistent  property to .

### How It Works

1. **Incremental View Construction**: Instead of rebuilding the entire View from all events on every agent step (via ), events are now added one at a time using . This method:
   - Appends  directly to the view
   - Tracks  events (sets  flag)
   - Applies  events (forgetting marked events, inserting summaries)
   - Caches  for performance

2. **Persistent View in ConversationState**: A private  field is maintained by , exposed as a  property. The view is NOT serialized (to avoid accidental serialization outside the file store), but is rebuilt from events when the state is deserialized.

3. **Breaking Changes**:
   -  field removed
   -  changed from static method to in-place modifier
   -  now takes a  instead of raw events

### What I Think of It

**The Good**:
- Incremental construction is the right architectural direction for scalability (O(n) vs O(n) per step adds up at scale)
- Makes  a first-class object - it represents the agent's "attention window" which is conceptually valuable
- Properly rebuilds view on deserialization

**The Concern** (valid from the review):
- **Two mutable sources of truth**:  (EventLog) vs . If any code appends directly to  (which remains publicly accessible), those events will silently NOT appear in the LLM context because  uses .
- This is a classic cache invalidation problem - the view is essentially a cache that can become stale.

**The Verdict**:
The PR is well-intentioned and conceptually sound. However, the  vs  divergence is a real trap that could cause silent bugs. The fix would be to either:
1. Make  a read-only property that routes through 
2. Or explicitly document that direct mutation of  is unsupported and will cause silent failures

The performance "benefits" are negligible (as the PR acknowledges - we're 99% network-bound), but making  first-class has long-term architectural value.

csmith49:
This has gone stale, closing in favor of a more incremental approach. See #2586 for the view construction logic -- the conversation state changes will be done in a separate PR.

Issue #2057, opened 2026-02-13
feat(delegate): Rewrite delegation tool APIs to match Claude Code patterns


simonrosenberg:
## Summary

The current delegation tool API surface (`_delegate_tasks()`, `register_agent()`, etc.) differs significantly from Claude Code's delegation patterns. To achieve feature parity and provide a familiar developer experience, the delegation APIs should be rewritten to align with Claude Code's `Task` tool interface.

### Claude Code's Delegation API

Claude Code exposes delegation through a single `Task` tool with these key parameters:

- **`prompt`** — The task description for the subagent
- **`subagent_type`** — Which specialized agent to use (e.g. "Explore", "Plan", "Bash")
- **`model`** — Optional model override for the subagent
- **`run_in_background`** — Launch as non-blocking background task
- **`resume`** — Resume a previous subagent by ID (preserves full context)
- **`description`** — Short summary shown in UI
- **`max_turns`** — Limit on agent turns

Complementary tools:
- **`TaskOutput`** — Get output from a background task (with `block` and `timeout` params)
- **`TaskStop`** — Stop a running background task by ID

### Current OpenHands API

- `_delegate_tasks()` with thread-based execution and `thread.join()` blocking
- `register_agent()` for programmatic agent registration
- No unified tool interface, no background support, no resume

## Key Requirements

- [x] Unified `Task` tool (or equivalent) as the primary delegation interface
- [ ] Support `subagent_type`, `model`, `run_in_background`, `resume` parameters
- [ ] `TaskOutput` equivalent for polling background task results
- [ ] `TaskStop` equivalent for terminating background tasks
- [ ] Deprecate or wrap the existing `_delegate_tasks()` internals behind the new API
- [ ] Maintain backward compatibility during transition

## Related Issues

- #2047 — Non-blocking background subagent execution
- #2048 — Model routing per subagent
- #2050 — Persist and resume subagent conversations
- #2051 — Built-in specialized agent types

VascoSch92:
Note that #2143 doesn't close completely this issue as a couple of things are missing. 
Precisely, 
- the API is missing the two parameters `run_in_background` and `model`
- Just the `TaskTool` is implemented. The tools `TaskOutput` and `TaskStop` will be implemented when we will able to run tasks in background.

PR #2143, opened 2026-02-20
feat(tools): task tool set

VascoSch92:
## Summary

Adding the `TaskToolSet` tool, but without the run_in_background feature, as discussed in #2100.

The arguments currently missing from this API (compared to the Claude Code task tool) are:
- `run_in_background`
- `model`

A difference between this version of `TaskToolSet` and `DelegateTool` is the ability to resume a task. The `task_tool_set` example demonstrates this mechanism.

**Note**
The tool description has been updated and tailored to reflect the current functionality (since background tasks are not supported).

(ref issue #2057)

**Docs**
https://github.com/OpenHands/docs/pull/352

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:a866cf8-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-a866cf8-python \
  ghcr.io/openhands/agent-server:a866cf8-python
```

```
ghcr.io/openhands/agent-server:a866cf8-golang-amd64
ghcr.io/openhands/agent-server:a866cf8-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:a866cf8-golang-arm64
ghcr.io/openhands/agent-server:a866cf8-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:a866cf8-java-amd64
ghcr.io/openhands/agent-server:a866cf8-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:a866cf8-java-arm64
ghcr.io/openhands/agent-server:a866cf8-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:a866cf8-python-amd64
ghcr.io/openhands/agent-server:a866cf8-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:a866cf8-python-arm64
ghcr.io/openhands/agent-server:a866cf8-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:a866cf8-golang
ghcr.io/openhands/agent-server:a866cf8-java
ghcr.io/openhands/agent-server:a866cf8-python
```

- Each variant tag (e.g., `a866cf8-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `a866cf8-python-amd64`) are also available if needed

### DIFF
--- a/openhands-tools/openhands/tools/task/definition.py
+++ b/openhands-tools/openhands/tools/task/definition.py
@@ -76,7 +76,6 @@ class TaskObservation(Observation):
         if self.is_error:
             text.append("❌ ", style="red bold")
             text.append(self.ERROR_MESSAGE_HEADER, style="bold red")
-            return super().visualize
 
         text.append(self.text)
         return text
--- a/openhands-tools/openhands/tools/task/manager.py
+++ b/openhands-tools/openhands/tools/task/manager.py
@@ -83,8 +83,6 @@ class TaskManager:
         # tmp directory to save task to eventually resume it later
         self._tmp_dir = Path(tempfile.mkdtemp(prefix="openhands_tasks_"))
 
-        self._task_counter = 0
-
     def _ensure_parent(self, conversation: LocalConversation) -> None:
         if self._parent_conversation is None:
             self._parent_conversation = conversation
@@ -100,8 +98,8 @@ class TaskManager:
 
     def _generate_ids(self) -> tuple[str, uuid.UUID]:
         """Generate a unique task ID, and a conversation ID."""
-        self._task_counter += 1
-        task_id = f"task_{self._task_counter:08x}"
+        task_number = len(self._tasks) + 1
+        task_id = f"task_{task_number:08x}"
         uuid_ = uuid.uuid4()
         return task_id, uuid_
 
--- a/tests/tools/task/test_task_manager.py
+++ b/tests/tools/task/test_task_manager.py
@@ -1,6 +1,6 @@
 import uuid
 from pathlib import Path
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 from pydantic import SecretStr
@@ -113,7 +113,6 @@ class TestTaskManager:
         """Manager should initialize with correct defaults."""
         manager = TaskManager()
         assert len(manager._tasks) == 0
-        assert manager._task_counter == 0
         assert manager._parent_conversation is None
 
     def test_tmp_dir_created(self):
@@ -125,14 +124,18 @@ class TestTaskManager:
     def test_generate_task_id(self):
         """Generated task IDs should be unique and prefixed."""
         manager = TaskManager()
-        assert manager._task_counter == 0
 
         tasks_ids: list[str] = []
         for j in range(10):
             id_, _ = manager._generate_ids()
             tasks_ids.append(id_)
+            manager._tasks[id_] = Task(
+                id=id_,
+                conversation=None,
+                status=TaskStatus.RUNNING,
+                conversation_id=uuid.uuid4(),
+            )
             assert id_.startswith("task_")
-            assert manager._task_counter == j + 1
 
         assert len(tasks_ids) == len(set(tasks_ids))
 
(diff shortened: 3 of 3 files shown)


enyst:
@openhands Do a /codereview-roasted on this PR. Take a look at some other tool too to see if we follow the patterns in the codebase.

Post your review and your answer as a comment to this PR. Note that it will be rendered as markdown.

Once you are done, take a step back and look at your review and diff *with fresh eyes*: section by section, was it more minor or major than it seemed to be? Reanalyze. When you're done, post your reanalysis as a follow-up comment.

enyst:
## /codereview-roasted

**Taste rating:** 🔴 **Needs improvement**

This *wants* to be a nice, minimal `TaskToolSet` that behaves like `BrowserToolSet` (factory tool spec → produces real tools). But right now it’s mixing patterns, and the end result is a footgun in the registry.

### Linus’s 3 questions
1. **Real problem?** Yes. Delegation is useful, and the ability to resume a sub-agent is a legit missing piece vs `DelegateTool`.
2. **Simpler way?** Also yes. You can copy the `BrowserToolSet` pattern almost verbatim and avoid half the custom glue.
3. **What will this break?** The current registration can break users who try to use the `task` tool directly.

---

- **`TaskTool` registration is broken and will explode at runtime**

  In `openhands-tools/openhands/tools/task/definition.py` you register **both** `TaskToolSet` *and* `TaskTool`:

  ```py
  register_tool(TaskToolSet.name, TaskToolSet)
  register_tool(TaskTool.name, TaskTool)
  ```

  But `register_tool(...)` for subclasses requires `.create(conv_state=..., **params)` (see `openhands.sdk.tool.registry._resolver_from_subclass`).

  `TaskTool.create(...)` is **not** compatible with that API (it requires `executor` and `description`). So if anybody uses `Tool(name="task")` (which is the obvious thing they’ll try), tool resolution will fail with a `TypeError`.

  **Fix:** do what `BrowserToolSet` does: **only register the toolset**, and keep the internal tool classes unregistered.

  (Alternative: rework `TaskTool.create()` to accept `conv_state` + build its own executor, but then you don’t need `TaskToolSet` at all.)

- **`TaskObservation.visualize` error path returns the wrong thing**

  In `TaskObservation.visualize`, the error branch builds custom `Text()` content and then returns `super().visualize`, discarding what you just built. That’s dead code / wrong behavior.

---

- **Giant inline tool description + `# noqa: E501`**

  The repo already has a cleaner pattern for rich descriptions: `DelegateTool` uses a Jinja template (`delegate_tool_description.j2`) and renders it with dynamic info.

  Here you hardcode a wall of text and then hack up `get_factory_info()` by dropping the first line:

  ```py
  lines = full_agent_types_info.splitlines()
  agent_types_info = "\n".join(lines[1:])
  ```

  That’s brittle. If the format of `get_factory_info()` changes, you silently produce garbage.

  Suggestion: use a template file like `delegate` does, and pass `agent_types_info` as-is (or parse it explicitly if you need structure).

- **Persistence dir choice is inconsistent**

  You’re persisting resumable conversations under `tempfile.mkdtemp(prefix=...)`. Fine for local dev, but it doesn’t obviously align with the existing persistence-dir story (`Conversation.persistence_dir`, `conv_state.env_observation_persistence_dir`, etc.). If this is meant to work well in agent-server contexts, consider rooting it under a runtime-managed directory.

- **Private attribute spelunking (`_visualizer`, `_name`)**

  `hasattr/getattr` on private fields is a layering smell. If you need the sender name, it should be exposed or passed in; otherwise this will rot.

- **Public import surface**

  `openhands.tools.__init__` re-exports the “main” tools (`BrowserToolSet`, `TaskTrackerTool`, etc.). `TaskToolSet` isn’t exposed there, so users must know to `import openhands.tools.task` to even get it registered. Maybe intentional, but if it’s meant to be a first-class alternative to delegate, consider adding it to the curated surface.

---

## [TESTING GAPS]

The tests mostly poke private methods/attributes. That’s OK for a new internal component, but you’re not testing the one thing that’s most likely to break users: **tool resolution & registry behavior**.

At minimum, add a test that asserts resolving `Tool(name=TaskToolSet.name)` produces a tool named `task`, and that calling it returns a `TaskObservation` with the expected fields. That would also catch the bad `register_tool(TaskTool.name, TaskTool)` immediately.

---

## VERDICT

❌ **Needs rework** before merge: the registry bug is not optional.

**KEY INSIGHT:** You implemented a tool *factory* (`ToolSet`) but then registered the internal tool as if it were independently constructible — it isn’t, and that will blow up at runtime.

enyst:
## Reanalysis (fresh eyes)

Stepping back: I probably came in *hot* on a couple of things that are more “taste” than “stop-ship”. Here’s how I’d re-rank it after a second pass.

### Still major / correctness
- **Registering `TaskTool` in the global registry is a real runtime bug.** This isn’t style — it’s an API contract violation with `register_tool(...)`. If someone tries `Tool(name="task")`, it will fail tool resolution immediately. That’s genuinely merge-blocking.
- **`TaskObservation.visualize` error branch returning `super().visualize`** also looks like plain wrong behavior (dead code / unexpected output). Not as catastrophic as the registry issue, but it’s still a correctness fix, not bikeshedding.

### Probably not worth blocking (but worth considering)
- The big inline `TASK_TOOL_DESCRIPTION` + `# noqa: E501` is ugly, but it’s not inherently wrong. Switching to a template is a quality improvement, not a functional requirement.
- Using `tempfile.mkdtemp(...)` for the resumable persistence dir is *potentially* inconsistent with other persistence paths, but in practice it might be totally acceptable for the initial iteration (especially if this is currently positioned as a standalone/experimental tool).
- The `_visualizer` / `_name` access is a layering smell, but it’s also optional functionality (sender label) and could be tightened later if needed.

### Net assessment
The core idea (ToolSet factory that creates a `task` tool backed by a shared manager, with resume support) is sound and matches existing patterns (`BrowserToolSet`). The PR is *close* — it just needs that registry fix (and ideally the `visualize` bug fix) to avoid shipping a broken API surface.

enyst:
Thank you! LGTM

@simonrosenberg I'd love to know your thoughts on this PR

examples/01_standalone_sdk/41_task_tool_set.py
```diff
@@ -0,0 +1,119 @@
...
+
+api_key = os.getenv("LLM_API_KEY")
+assert api_key is not None, "LLM_API_KEY environment variable is not set."
+
+llm = LLM(
+    model=os.getenv("LLM_MODEL", "anthropic/claude-sonnet-4-5-20250929"),
+    api_key=SecretStr(api_key),
+    base_url=os.getenv("LLM_BASE_URL", None),
+)
+
```
simonrosenberg: Could we try keeping the simplicity of the first example? 

```suggestion
llm = LLM(
    model=os.getenv("LLM_MODEL", "anthropic/claude-sonnet-4-5-20250929"),
    api_key=os.getenv("LLM_API_KEY"),
    base_url=os.getenv("LLM_BASE_URL", None),
)
```

openhands-tools/openhands/tools/task/__init__.py
```diff
@@ -0,0 +1,24 @@
+"""TaskToolSet.
```
simonrosenberg: I am a bit confused about 
TaskToolSet here vs Task tool kit in definition.py

simonrosenberg: Are we not adding a reference to Claude Code to say we are using the same APIs more or less? Or we can wait until we are closer to what they do perhaps

openhands-tools/openhands/tools/task/definition.py
```diff
@@ -0,0 +1,217 @@
...
+        default="default",
+        description="The type of specialized agent to use for this task.",
+    )
+    resume: str | None = Field(
+        default=None,
+        description="Task ID of the task to resume from.",
+    )
+    max_turns: int | None = Field(
+        default=None,
+        description="Maximum number of agentic turns before stopping.",
```
simonrosenberg, line 53:
```suggestion
        description="Maximum number of agentic steps before stopping.",
```

openhands-tools/openhands/tools/task/definition.py
```diff
@@ -0,0 +1,217 @@
...
+- The agent's outputs should generally be trusted
+- Clearly tell the agent whether you expect it to write code or just to do research (search, file reads, web fetches, etc.), since it is not aware of the user's
+intent
+- If the agent description mentions that it should be used proactively, then you should try your best to use it without the user having to ask for it first. Use your
+judgement.
+"""  # noqa: E501
+
+
+class TaskTool(ToolDefinition[TaskAction, TaskObservation]):
+    """Tool for launching sub-agent tasks (blocking)."""
```
simonrosenberg:
```suggestion
    """Tool for launching (blocking) sub-agent tasks."""
```

openhands-tools/openhands/tools/task/manager.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def __init__(self):
+        self._parent_conversation: LocalConversation | None = None
+
+        self._tasks: dict[str, Task] = {}
+
+        # tmp directory to save task to eventually resume it later
+        self._tmp_dir = Path(tempfile.mkdtemp(prefix="openhands_tasks_"))
+
+        self._task_counter = 0
```
simonrosenberg: can _task_counter be inferred from tasks and become a property?

openhands-tools/openhands/tools/task/definition.py
```diff
@@ -0,0 +1,217 @@
+"""Task tool kit definitions.
+
+This module defines the Task tool for sub-agent task delegation.
+
+The TaskToolSet is registered as a single tool spec that creates
+the Task tool backed by a TaskManager.
```
simonrosenberg: IMO this doesn't really help understanding what those objects do on a high level: Task, TaskToolSet, TaskManager.

openhands-tools/openhands/tools/task/manager.py
```diff
@@ -0,0 +1,270 @@
+from __future__ import annotations
```
simonrosenberg: docstring

openhands-tools/openhands/tools/task/manager.py
VascoSch92: Yes, it could (in this case with blocking tasks)

But a standalone counter is monotonically increasing regardless of what happens to the dict, so it's inherently safer.

openhands-tools/openhands/tools/task/definition.py
VascoSch92: Added a docstring for the three scripts

openhands-tools/openhands/tools/task/__init__.py
VascoSch92: I don't know if we should or not. 

Actually the phylosophy behind is more or less the same but the implementation is different. 

We could add something in this direction.

VascoSch92: Added docstring. 

Just a precision:
The `TaskToolSet` acts as a wrapper to accommodate potential background tasks in the future. 

Implementing background tasks would also imply adding `TaskStop` and `TaskOutput` tools, bringing the total to three. 

Currently, `TaskToolSet` only contains `TaskTool`, which we could technically expose directly without the `TaskToolSet` wrapper.

openhands-tools/openhands/tools/task/manager.py
simonrosenberg: but it's a duplicate in logic... I think it makes the code simpler to have properties depending on stored objects rather than two stored objects having similar properties... but I might be wrong

VascoSch92: Changed ;-)

VascoSch92:
Merging also if `[Optional] Docs example /check-examples` are not passing. 

This is becausse of another example in the code which has not a documentation now.

The example added in the PR has docs.

PR #2170, opened 2026-02-22
perf: parallelize event reading and minimize lock scope in event_service

rbren:
This seems to work well in my test env!

- Add ThreadPoolExecutor to read events in parallel outside the state lock
- Minimize lock scope in _search_events_sync and _count_events_sync to only capture the event_log reference and count, avoiding holding the lock during I/O-bound event deserialization
- Extract _event_passes_filters helper to eliminate duplicated filter logic

## Summary

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:d006a2a-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-d006a2a-python \
  ghcr.io/openhands/agent-server:d006a2a-python
```

```
ghcr.io/openhands/agent-server:d006a2a-golang-amd64
ghcr.io/openhands/agent-server:d006a2a-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:d006a2a-golang-arm64
ghcr.io/openhands/agent-server:d006a2a-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:d006a2a-java-amd64
ghcr.io/openhands/agent-server:d006a2a-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:d006a2a-java-arm64
ghcr.io/openhands/agent-server:d006a2a-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:d006a2a-python-amd64
ghcr.io/openhands/agent-server:d006a2a-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:d006a2a-python-arm64
ghcr.io/openhands/agent-server:d006a2a-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:d006a2a-golang
ghcr.io/openhands/agent-server:d006a2a-java
ghcr.io/openhands/agent-server:d006a2a-python
```

- Each variant tag (e.g., `d006a2a-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `d006a2a-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -149,52 +149,11 @@ class EventService:
             except Exception:
                 return None
 
-        if sort_order == EventSortOrder.TIMESTAMP:
-            # Ascending: resolve page cursor in O(1) via the index map, then
-            # read events in parallel and stop as soon as limit matches are found.
-            start_idx = 0
-            if page_id is not None:
-                if hasattr(event_log, "get_index"):
-                    try:
-                        start_idx = event_log.get_index(page_id)
-                    except KeyError:
-                        pass
-                else:
-                    for i in range(event_count):
-                        if event_log[i].id == page_id:
-                            start_idx = i
-                            break
-
-            executor = ThreadPoolExecutor(max_workers=8)
-            try:
-                futures = [
-                    executor.submit(read_event, i)
-                    for i in range(start_idx, event_count)
-                ]
-                items: list[Event] = []
-                next_page_id = None
-                for future in futures:
-                    event = future.result()
-                    if event is None:
-                        continue
-                    if not self._event_passes_filters(
-                        event, kind, source, body, timestamp_gte_str, timestamp_lt_str
-                    ):
-                        continue
-                    if len(items) >= limit:
-                        next_page_id = event.id
-                        break
-                    items.append(event)
-            finally:
-                executor.shutdown(wait=False, cancel_futures=True)
-
-            return EventPage(items=items, next_page_id=next_page_id)
-
-        # TIMESTAMP_DESC: read all events in parallel outside the lock, filter,
-        # sort descending, then paginate.
+        # Read all events in parallel outside the lock
         with ThreadPoolExecutor(max_workers=8) as executor:
             all_raw = list(executor.map(read_event, range(event_count)))
 
+        # Filter events
         all_events = [
             e
             for e in all_raw
@@ -203,8 +162,14 @@ class EventService:
                 e, kind, source, body, timestamp_gte_str, timestamp_lt_str
             )
         ]
-        all_events.sort(key=lambda x: x.timestamp, reverse=True)
 
+        # Sort events based on sort_order
+        if sort_order == EventSortOrder.TIMESTAMP:
+            all_events.sort(key=lambda x: x.timestamp)
+        elif sort_order == EventSortOrder.TIMESTAMP_DESC:
+            all_events.sort(key=lambda x: x.timestamp, reverse=True)
+
+        # Find the starting point if page_id is provided
         start_index = 0
         if page_id is not None:
             for i, event in enumerate(all_events):
@@ -212,6 +177,7 @@ class EventService:
                     start_index = i
                     break
 
+        # Collect items for this page
         items = []
         next_page_id = None
         for i in range(start_index, len(all_events)):


enyst:
@openhands 

Question 1. Do a /codereview-roasted on this PR, what is your feedback?

Question 2. Then, when you are done, tell me: were we reading ALL events from disk before this PR? are we reading ALL events from disk after this PR?

Post your review and answer as a comment to the PR. Note that it will be rendered as markdown.

Question 3. After you posted, take a step back and re-read your own posted review and the PR, and re-analyze  each issues with *fresh eyes*: is it actually minor or major, more than it seemed before? Post your answer re-analysis as a comment too.

IMPORTANT: You are under instruction to ANSWER, but not post random comments. These are 3 questions, and for each question you are allowed to answer on Github. So do not post other comments, but do answer.

enyst:
> no PR comments were actually posted from this environment—only the Markdown content to paste was prepared, aligning with the instruction to answer rather than modify code.

HUMAN: 😅 😅 

So it has the resolver prompt in the system prompt, attempted fix here: 
- https://github.com/OpenHands/OpenHands/pull/12983

As you can see, I didn't convince it yet to bypass it. The level at which GPT-5.x follows instructions is extremely interesting - and different than Claudes.

---

(OpenHands-GPT-5.2 below)

I can’t directly post GitHub comments from here given my current workflow instructions, but below are **two ready-to-paste PR comments** (markdown) that answer Q1+Q2 and then Q3.

## Comment 1 (Q1 + Q2)

/codereview-roasted

**Taste rating:** Needs improvement (the idea is fine; the execution has a few “this will bite you later” choices).

### [CRITICAL ISSUES]

- **[openhands-agent-server/openhands/agent_server/event_service.py:L152+] “TIMESTAMP” sort no longer actually sorts by timestamp.**  
  Before: `TIMESTAMP` meant “collect all matching events, then `sort(key=timestamp)`”.  
  After: the `TIMESTAMP` path returns events in **event-log index order** (append/file order) and assumes that is equivalent to timestamp order. That’s an undocumented invariant. If timestamps are ever out-of-order (clock skew, multi-process writing, weird event creation), this quietly violates the API contract.

- **[L168-L190] Unbounded work submission + fake “early stop”.**  
  The `TIMESTAMP` path submits a future for **every event from `start_idx` to `event_count`** and only *then* stops consuming after `limit`. You’re still allocating a giant `futures` list and queueing potentially huge work; cancellation helps, but the damage (memory / scheduling) is already done.  
  If the log is large, this is exactly the kind of “perf improvement” that becomes a production incident.

- **[L168-L190, L195-L197, count path L278-L280] Threadpool-in-threadpool oversubscription risk.**  
  `search_events()` already runs `_search_events_sync` inside `loop.run_in_executor(...)`. Inside that worker thread you create *another* `ThreadPoolExecutor(max_workers=8)` per request. Under concurrent API load, this can explode your thread count and degrade latency across the whole server.

- **[L146-L151, L272-L277] Swallowing exceptions changes correctness semantics.**  
  `read_event()` catches **all exceptions** and returns `None`. That means corrupted/invalid event JSON, unexpected validation errors, filesystem errors, etc. become “silently missing events” rather than a surfaced failure. That’s a behavior change and it makes debugging data corruption miserable.

- Use a **shared / bounded executor** (module-level or service-level) rather than creating one per call. Or don’t nest: either run `_search_events_sync` on the event loop thread *and* use the executor inside, or keep the outer executor and do reads sequentially.
- For `TIMESTAMP` early stop, **don’t submit all tasks upfront**. Submit in chunks (or use a producer/consumer queue) so “stop after limit” actually stops doing work.
- Remove the `hasattr(event_log, "get_index")` branch; `state.events` is an `EventLog` and has `get_index()`. The fallback adds complexity without a real use case.
- Add tests that pin pagination + ordering semantics (especially around `page_id`, `TIMESTAMP` vs `TIMESTAMP_DESC`, and filtering).

### VERDICT
The motivation is good (shorter lock scope, parallel I/O), but the current approach trades one bottleneck for multiple foot-guns (ordering semantics drift, thread explosion risk, and silent data loss). I’d rework the implementation before merging.

---

**Q2 (were we reading ALL events from disk before vs after?):**

- **Before this PR:** yes.  
  `search_events_sync()` iterated `for event in state.events:` and built `all_events` before sorting/paginating; `EventLog.__iter__` reads every event file. Same for `count_events_sync()`.

- **After this PR:**  
  - `count_events_sync()`: **still yes** (it reads `range(event_count)` via executor map).  
  - `search_events_sync(sort_order=TIMESTAMP_DESC)`: **still yes** (reads all, filters, sorts).  
  - `search_events_sync(sort_order=TIMESTAMP)`: **not strictly guaranteed** to read all events, but it still **submits work for all events** and relies on cancellation/early-break to avoid completing all reads. In practice it may read “most” of the log under load or on fast disks.

---

## Comment 2 (Q3 re-analysis, fresh eyes)

Re-reading the PR + my own review with fresh eyes, here’s what I’d now classify as **major vs minor**:

### Major (I’d want fixed before merge)

1. **“TIMESTAMP” not actually sorting by timestamp anymore**  
   If you *know* the invariant “event index order == timestamp order” is always true, then codify it (document + assert in code/tests) and rename the sort to something like “APPEND_ORDER”. Otherwise, keep the sort semantics consistent.

2. **Submitting futures for the entire tail of the log**  
   This is the core scalability bug: it defeats the whole point of “stop once limit matches are found” and can create huge per-request overhead.

3. **Nested per-request executors (thread explosion)**  
   This is a server, not a benchmark script. Per-request threadpools are a classic way to make tail latency and resource usage unpredictable.

4. **Catching `Exception` and returning `None`**  
   On second read, this is worse than I first thought: it can silently “lose” events in search/count results, which is a data integrity problem disguised as resilience.

### Minor / situational

- **Minimizing lock scope**: probably fine *if* we’re confident the `EventLog` read path is safe without the conversation state lock. The “event files are immutable” argument is reasonable, but it doesn’t automatically mean the in-memory index structures can be accessed concurrently without surprises. (This becomes a non-issue if reads are robust and we don’t mask exceptions.)

Net: the PR is pointed in the right direction, but the above “major” items are real correctness/perf risks, not nits.

rbren:
@openhands fix this PR based on the feedback in the latest comment

rbren:
I'm actually still seeing the server lock up on `/search` with these changes :(

rbren:
https://github.com/OpenHands fix this PR based on the feedback from enyst. DO NOT run the precommit and tests when you're done, just push your changes

rbren:
@openhands fix this PR based on the feedback from enyst. DO NOT run the precommit and tests when you're done, just push your changes

enyst:
> I'm actually still seeing the server lock up on `/search` with these changes :(

Out of curiosity, how many events total did your conversation(s) have? Also, are you testing with what workspace?

rbren:
I'm running the agent server locally, so I guess LocalWorkspace

Generally conversations with 50-100 events, sometimes more

PR #2171, opened 2026-02-22
fix: Include secrets in system prompt when added via update_secrets()

rbren:
## Summary

Fixes an issue where secrets passed to `LocalConversation` were not appearing in the system prompt, even though the infrastructure for including them exists.

## Problem

Previously, secrets passed to `LocalConversation` via:
- The constructor's `secrets` parameter
- The `update_secrets()` method

Were **only** added to the `secret_registry` for environment variable injection during command execution. They were **NOT** added to `agent.agent_context.secrets`, which meant they did not appear in the system prompt's `<CUSTOM_SECRETS>` section.

This meant the agent had no awareness of what secrets were available to it, unlike OpenHands web where secrets and their descriptions are always shown in the system prompt.

## Solution

Updated `update_secrets()` in `LocalConversation` to also update `agent.agent_context.secrets`, ensuring that:
1. Secrets are still registered in `secret_registry` for command execution (env var injection)
2. Secrets are also added to `agent.agent_context.secrets` so they appear in the system prompt

## Changes

- **`local_conversation.py`**: 
  - Updated `update_secrets()` to call new `_update_agent_context_secrets()` helper
  - Added `_update_agent_context_secrets()` helper method that merges secrets into agent_context
  - Added import for `AgentContext`

- **`test_agent_secrets_integration.py`**: Added 5 new tests:
  - `test_update_secrets_adds_to_agent_context`
  - `test_update_secrets_appears_in_system_prompt`
  - `test_update_secrets_merges_with_existing_context`
  - `test_update_secrets_overrides_existing_secret`
  - `test_secrets_via_constructor_appear_in_prompt`

## Testing

All 14 tests in `test_agent_secrets_integration.py` pass, including the 5 new ones.

## Before/After

**Before:** Agent has no knowledge of available secrets in system prompt
```
# System prompt has no <CUSTOM_SECRETS> section
```

**After:** Agent sees available secrets in system prompt
```xml
<CUSTOM_SECRETS>
### Credential Access
* Automatic secret injection: When you reference a registered secret key...
...
You have access to the following environment variables
* **$GITHUB_TOKEN** - GitHub authentication token
* **$OPENAI_API_KEY** - OpenAI API key for LLM calls
</CUSTOM_SECRETS>
```

@rbren can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:181e94e-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-181e94e-python \
  ghcr.io/openhands/agent-server:181e94e-python
```

```
ghcr.io/openhands/agent-server:181e94e-golang-amd64
ghcr.io/openhands/agent-server:181e94e-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:181e94e-golang-arm64
ghcr.io/openhands/agent-server:181e94e-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:181e94e-java-amd64
ghcr.io/openhands/agent-server:181e94e-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:181e94e-java-arm64
ghcr.io/openhands/agent-server:181e94e-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:181e94e-python-amd64
ghcr.io/openhands/agent-server:181e94e-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:181e94e-python-arm64
ghcr.io/openhands/agent-server:181e94e-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:181e94e-golang
ghcr.io/openhands/agent-server:181e94e-java
ghcr.io/openhands/agent-server:181e94e-python
```

- Each variant tag (e.g., `181e94e-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `181e94e-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/context/agent_context.py
+++ b/openhands-sdk/openhands/sdk/context/agent_context.py
@@ -262,16 +262,10 @@ class AgentContext(BaseModel):
         secret_infos = self.get_secret_infos()
         if additional_secret_infos:
             # Merge: additional secrets override agent_context secrets by name
-            existing_names = {s["name"] for s in secret_infos}
+            secret_dict = {s["name"]: s for s in secret_infos}
             for additional in additional_secret_infos:
-                if additional["name"] not in existing_names:
-                    secret_infos.append(additional)
-                else:
-                    # Override existing secret info with additional
-                    secret_infos = [
-                        additional if s["name"] == additional["name"] else s
-                        for s in secret_infos
-                    ]
+                secret_dict[additional["name"]] = additional
+            secret_infos = list(secret_dict.values())
         formatted_datetime = self.get_formatted_datetime()
         has_content = (
             repo_skills


rbren:
Hmm I kind of agree the control flow here is bad.

Probably better for the agent to do something like `get_system_context()` to pull data about secrets, runtime state, etc into its prompts

But I do think it's a basic assumption that if you define secrets, the agent knows it has access to those secrets

rbren:
@openhands change the control flow in the implementation here. have the agent request the secrets and add them to its prompt rather than modifying the agent from outside

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -202,6 +206,42 @@ def init_state(
...
+            The dynamic context string, or None if no context is configured.
+        """
+        # Get secret infos from conversation's secret_registry
+        secret_infos = state.secret_registry.get_secret_infos()
+
+        if not self.agent_context:
+            # No agent_context but we might have secrets from registry
+            if secret_infos:
+                from openhands.sdk.context.agent_context import AgentContext
+
```
enyst, line 229:
```suggestion
```

enyst:
@openhands Do a /codereview-roasted on this PR. Investigate examples/ files too on this, to see how we work with agent context and secrets.

Once you are done, take a step back and reconsider your review and the PR diff *with fresh eyes*: are things as major or minor as they seemed at fight sight?

enyst:
(OpenHands-GPT-5.2)

## Taste Rating: Acceptable (solves a real problem, but the shape is kind of ugly)

You fixed the *actual* bug: secrets added to a `LocalConversation` (constructor + `update_secrets()`) now show up in the system prompt via the existing `<CUSTOM_SECRETS>` machinery. That’s a real UX gap vs web OpenHands, so the direction is right.

But the implementation introduces some “two sources of truth” and layering weirdness that will age poorly.

---

## [CRITICAL ISSUES]

### [openhands-sdk/openhands/sdk/agent/agent.py:187-243] **You now have two different “dynamic context” concepts**
- `Agent.init_state()` uses `self.get_dynamic_context(state)` (state-dependent).
- `AgentBase.dynamic_context` (property) still only reflects `agent_context` (state-independent).
- `AgentBase.system_message` (deprecated, but still used) also won’t include registry secrets.

So: the *real* system prompt event contains secrets, but the *public-ish* agent properties do not. That’s inconsistent and will confuse anyone debugging prompts or doing anything custom with `agent.dynamic_context` / `agent.system_message`.

This is the core architectural smell: the agent is “stateless” by design, but you need state to compute prompt context. Either embrace that and make a single blessed codepath, or keep the agent stateless and stop pushing stateful data through `AgentContext`.

### [agent.py:225-237] **Creating a throwaway `AgentContext()` just to render secrets is a code smell**
When `self.agent_context` is missing, you do:
- local import
- `temp_context = AgentContext()`
- call `get_system_message_suffix(...)`

That’s not “minimal”; it drags in default `current_datetime` behavior and any future defaults in `AgentContext`, just because secrets exist. You’re effectively using `AgentContext` as a templating utility, not as configuration.

If all you need is “render the secrets section”, render *the secrets section*.

### [agent_context.py:260-274] **The secret merge algorithm is bad data-structure taste**
You represent secrets as `list[dict[name, description]]` and then merge via:
- membership set created once (`existing_names = {...}`)
- per-item list rebuild when overriding (`secret_infos = [...]`)

This is exactly the kind of “list-of-dicts” nonsense that makes simple things complicated:
- it’s O(n²) in the override case
- `existing_names` doesn’t update when you append, so duplicate names *within* `additional_secret_infos` can duplicate entries
- it’s harder to reason about ordering and override semantics than it needs to be

Use a `dict[str, SecretInfo]` keyed by name for merging, then render in a stable order. You can still preserve insertion order if you care.

---

### Keep layering clean: `AgentContext` shouldn’t need “additional secrets”
`AgentContext.get_system_message_suffix(..., additional_secret_infos=...)` is basically admitting that `AgentContext` is no longer “agent configuration”, but a place where conversation state leaks in.

Cleaner options:
1. Compute `base_dynamic = agent.agent_context.get_system_message_suffix(...)` (if any)
2. Compute `secrets_suffix = render_secrets(secret_registry)` (separate, small helper/template)
3. Concatenate if present

That avoids:
- changing `AgentContext` API for a conversation concern
- the “temp AgentContext” hack
- the list-merge mess

### Duplication: `get_secret_infos()` exists twice
`AgentContext.get_secret_infos()` and `SecretRegistry.get_secret_infos()` are basically the same concept. That’s manageable now, but it’s the start of drift. Pick one canonical representation/shape.

### Docstring coupling
`LocalConversation.update_secrets()` now documents internal details like “agent pulls secrets via get_dynamic_context() during init_state()”. That’s the kind of comment that becomes wrong in 3 months. Describe behavior/contract, not internal call paths.

---

## [TESTING GAPS]

The added tests (around `agent.get_dynamic_context(state)`) cover the intended behavior, but they test the helper rather than the *actual emitted system prompt event*.

A more regression-proof test would assert that after initialization, the `SystemPromptEvent.dynamic_context` contains `<CUSTOM_SECRETS>` and the secret names/descriptions. Right now it’s “probably correct” because `init_state()` calls the helper—but it’s indirect.

---

## Examples audit (re: “how we work with agent context and secrets”)

- `examples/01_standalone_sdk/12_custom_secrets.py` uses `conversation.update_secrets(...)` (conversation-level), not `AgentContext(secrets=...)`. That matches the direction of this PR: secrets live with the conversation/runtime, not agent config.
- `examples/03_github_workflows/02_pr_review/agent_script.py` passes `secrets=...` into `Conversation(...)`. With this PR those secret *names* will now appear in the system prompt. That’s probably desirable (agent knows it can authenticate), but it’s a behavior change worth being aware of (prompt length + potentially different agent decisions).

---

## VERDICT
**Worth merging**, because it fixes a real and user-visible mismatch.

But with fresh eyes: the problems here are mostly *taste / maintainability*, not immediate correctness bugs. The one thing I’d strongly reconsider (because it will keep biting) is the split-brain “dynamic context” API and the list-of-dicts merge logic—those are the seeds of future confusion.

enyst (approved):
This looks clean to me, thank you! As you can see, OpenHands is not entirely happy, but I feel it's manageable in follow-ups.

Up to you if you wish a second opinion or could address some of the rest.

enyst:
@openhands Look at the https://github.com/OpenHands/software-agent-sdk/blob/65cf6c227f5e56c4d328bfe77a012abba9c7a79b/.github/workflows/api-breakage.yml and its CI log from this PR.

Compare with the PR's diff. Shouldn't we have had some checks on the method signature change in this PR diff? Why did it succeed without reporting signature changes?

Issue #2054, opened 2026-02-13
[Epic] Delegation: Rewrite APIs, add file-based sub-agents and implement Explore & Bash Agents


simonrosenberg:
## Delegation Epic

This epic tracks the implementation of advanced delegation capabilities to achieve feature parity with Claude Code's subagent system. Key themes are **background execution**, **model routing**, and **declarative agent definitions**.

### Sub-Issues

- [X] [AGE-642](https://linear.app/all-hands-ai/issue/AGE-642/featdelegate-rewrite-delegation-tool-apis-to-match-claude-code) — Rewrite delegation tool APIs to match Claude Code patterns
- [ ] [AGE-632](https://linear.app/all-hands-ai/issue/AGE-632/featdelegate-non-blocking-background-subagent-execution) — Non-blocking background subagent execution
- [ ] [AGE-641](https://linear.app/all-hands-ai/issue/AGE-641/featdelegate-move-blocking-action-to-background-mid-execution) — Move blocking action to background mid-execution
- [X] [AGE-634](https://linear.app/all-hands-ai/issue/AGE-634/featdelegate-file-based-agent-definitions-with-markdown-yaml) — File-based agent definitions with markdown + YAML frontmatter
- [X] [AGE-635](https://linear.app/all-hands-ai/issue/AGE-635/featdelegate-persist-and-resume-subagent-conversations) — Persist and resume subagent conversations
- [X] [AGE-636](https://linear.app/all-hands-ai/issue/AGE-636/featdelegate-built-in-specialized-agent-types-explore-bash) — Built-in specialized agent types (Explore, Plan, Bash)
- [X] [AGE-633](https://linear.app/all-hands-ai/issue/AGE-633/featdelegate-model-routing-per-subagent) - Model Routing per Subagent
- [X] [AGE-637](https://linear.app/all-hands-ai/issue/AGE-637/featdelegate-context-distillation-for-subagent-output) - Context distillation for subagent output
- [X] [AGE-693](https://linear.app/all-hands-ai/issue/AGE-693/observabilitydelegation-llm-telemetry-for-agent-delegation) - LLM Telemetry for Agent Delegation
- [x] [AGE-692](https://linear.app/all-hands-ai/issue/AGE-692/evaldelegation-measureconfirm-performance-improvement) - Measure/Confirm performance implementation
[...]

Issue #2049, opened 2026-02-13
feat(delegate): File-based agent definitions with markdown + YAML frontmatter


simonrosenberg:
## Summary

Agents currently require factory code via `register_agent()`. There's no declarative way to define agents, making it harder for users to create and share custom agent types.

## Proposed Solution

Support markdown-based agent definitions loaded from an `.agents/` directory, with YAML frontmatter specifying model, tools, and system prompt. The existing skill loading system already parses frontmatter and can serve as a template.

### Key Requirements

- Define agent definition file format (markdown body + YAML frontmatter)
- Frontmatter fields: `name`, `model`, `tools` (list), `description`, `allowed_commands`
- Markdown body serves as the agent's system prompt
- Load agent definitions from `.agents/` directory (project-level and user-level)
- Register file-based agents alongside programmatic ones
- Hot-reload on file changes (or at minimum, reload on new conversation)

### Key Files

- Skill loading system (frontmatter parser) — template for file parsing
- `openhands-tools/openhands/tools/delegate/registration.py` — Agent registry

## Acceptance Criteria

- [ ] Agent definitions can be written as markdown files with YAML frontmatter
- [ ] Files in `.agents/` directory are automatically loaded
- [ ] File-based agents are available for delegation alongside programmatic ones
- [ ] All frontmatter fields are respected (model, tools, etc.)

PR #2183, opened 2026-02-23
feat(delegate): File-based agent definitions with markdown + YAML frontmatter

VascoSch92:
## Summary

This PR introduces the ability to define and load agents via Markdown files and performs a structural refactor to centralize agent-loading logic. (ref #2049 )

**Key Features**
- Markdown Agent Definitions: Added support for loading agent configurations directly from .md files.
- Loading Hierarchy: Implemented a file-loading priority consistent with the existing Skills hierarchy.

**Architectural Refactor**
Moved all logic related to agent discovery and instantiation from plugin and tools.delegate into a dedicated subagents module.

**Remarks**
This PR focuses on the core infrastructure. Several advanced features for Markdown-based agents are currently unsupported and are being tracked in the primary Epic #2054.

**docs**
https://github.com/OpenHands/docs/pull/358

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:4f812b3-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-4f812b3-python \
  ghcr.io/openhands/agent-server:4f812b3-python
```

```
ghcr.io/openhands/agent-server:4f812b3-golang-amd64
ghcr.io/openhands/agent-server:4f812b3-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:4f812b3-golang-arm64
ghcr.io/openhands/agent-server:4f812b3-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:4f812b3-java-amd64
ghcr.io/openhands/agent-server:4f812b3-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:4f812b3-java-arm64
ghcr.io/openhands/agent-server:4f812b3-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:4f812b3-python-amd64
ghcr.io/openhands/agent-server:4f812b3-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:4f812b3-python-arm64
ghcr.io/openhands/agent-server:4f812b3-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:4f812b3-golang
ghcr.io/openhands/agent-server:4f812b3-java
ghcr.io/openhands/agent-server:4f812b3-python
```

- Each variant tag (e.g., `4f812b3-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `4f812b3-python-amd64`) are also available if needed

### DIFF
--- a/openhands-tools/openhands/tools/__init__.py
+++ b/openhands-tools/openhands/tools/__init__.py
@@ -18,8 +18,7 @@ from ``openhands.tools.browser_use``.
 
 from importlib.metadata import PackageNotFoundError, version
 
-from openhands.tools.delegate import DelegationVisualizer, register_agent
-from openhands.tools.delegate.registration import get_agent_factory
+from openhands.tools.delegate import DelegationVisualizer
 from openhands.tools.file_editor import FileEditorTool
 from openhands.tools.preset.default import (
     get_default_agent,


openhands-sdk/openhands/sdk/subagent/load.py
```diff
@@ -0,0 +1,142 @@
...
+    """Shared helper: scan *agents_dir* for top-level `.md` agent files."""
+    if not agents_dir.is_dir():
+        return []
+
+    definitions: list[AgentDefinition] = []
+    for md_file in sorted(agents_dir.iterdir()):
+        # Only top-level .md files; skip subdirectories and README
+        if (
+            md_file.is_dir()
+            or md_file.suffix.lower() == ".md"
```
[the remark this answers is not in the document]
VascoSch92: good one :-)

openhands-tools/openhands/tools/__init__.py
```diff
@@ -14,8 +14,7 @@
 from importlib.metadata import PackageNotFoundError, version
 
 from openhands.tools.browser_use import BrowserToolSet
-from openhands.tools.delegate import DelegationVisualizer, register_agent
-from openhands.tools.delegate.registration import get_agent_factory
+from openhands.tools.delegate import DelegationVisualizer
```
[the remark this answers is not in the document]
VascoSch92: This is OK. See slack discussion: https://openhands-ai.slack.com/archives/C06R25BT5B2/p1771845323455699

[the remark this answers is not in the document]
VascoSch92: This is Ok. We can break the API. See slack discussion: https://openhands-ai.slack.com/archives/C06R25BT5B2/p1771845323455699

VascoSch92:
Note that the docs tests are not passing because of missing documentations for `examples/01_standalone_sdk/40_acp_agent_example.py`. The docs for this feature should be merged soon.

openhands-sdk/openhands/sdk/subagent/registry.py
```diff
@@ -0,0 +1,297 @@
...
+) -> Callable[["LLM"], "Agent"]:
+    """Create an agent factory closure from an ``AgentDefinition``.
+
+    The returned callable accepts an :class:`LLM` (the parent agent's LLM) and
+    builds a fully-configured `Agent` class.
+
+    - Tool names from `agent_def.tools` are mapped to `Tool` objects.
+    - The system prompt becomes an always-active `Skill` (`trigger=None`).
+    - `model: inherit` preserves the parent LLM; an explicit model name creates
+        a copy with `model_copy()`.
```
enyst: Should this be a profile id?

enyst:
Maybe we can add an example, with something like:
```
---
name: code-reviewer
description: Expert code review specialist. Proactively reviews code for quality, security, and maintainability. Use immediately after writing or modifying code.
tools: Read, Grep, Glob, Bash
model: inherit
---

You are a senior code reviewer ensuring high standards of code quality and security.

When invoked:
1. Run git diff to see recent changes... etc
```

Just for future record and ease of review it might be useful to update in the PR description too?

enyst:
@openhands Do a /codereview-roasted on this PR

enyst:
(HUMAN: pasting the result here, but please feel free to smack it when it's over the top on nothing really - I feel that might be the case)

---

## Taste Rating: **Needs improvement**
This PR is trying to solve a real problem (declarative subagents), but it does it with a couple of “global state + first-wins” shortcuts that will absolutely bite you in any long-running process (agent-server), and it **breaks existing public imports** without a compatibility shim.

---

### 1) **You broke userspace: removed public APIs without a compatibility layer**
Two concrete breakages:

- **`openhands.tools.delegate.register_agent` is gone**  
  Previously exported in `openhands-tools/openhands/tools/delegate/__init__.py` (and tested). Now the module exports only tool classes (new file shows no `register_agent`, lines 3–18). Anyone with:
  ```py
  from openhands.tools.delegate import register_agent
  ```
  now gets an ImportError.

- **`openhands.sdk.plugin.AgentDefinition` is gone**  
  `openhands-sdk/openhands/sdk/plugin/__init__.py` no longer re-exports `AgentDefinition` (current file exports many types, but not that one; see lines ~16–48). That was a documented import surface before.

If you want to refactor internals, fine, but **keep the old import paths working** (even if they just re-export and warn) or bump versions and formally deprecate. Right now it’s “surprise, your code doesn’t import anymore”.

---

### 2) **Global registry + `register_*_if_absent()` destroys your priority model and breaks “reload on new conversation”**
You claim a priority order in `openhands-sdk/openhands/sdk/subagent/load.py` (docstring lines 29–36) and in `register_plugin_agents()` (registry.py lines 210–236), but the actual implementation is basically:

- global `_agent_factories` dict (registry.py line 53)
- `register_plugin_agents()` uses `register_agent_if_absent()` (registry.py lines 225–235)
- `register_file_agents()` uses `register_agent_if_absent()` (registry.py lines 193–206)
- `LocalConversation` calls them once per conversation (local_conversation.py lines 374–379)

That means the **first thing to register a given name wins forever**, across:
- different workspaces/projects in the same process
- later conversations
- later plugin loads
- edits to `.md` agent definitions

This directly violates your own stated priority (“plugin agents higher priority than file-based”) in any scenario where:
1) conversation A starts without plugins → file agent “foo” registered
2) conversation B starts with plugin providing “foo” → plugin cannot override, because “foo” already exists

It also fails the issue’s “reload on new conversation” requirement for *updated* definitions: if the `.md` file changes, a new conversation will still **silently keep the old definition** because duplicates are skipped.

This is the core design smell: **you can’t implement priority and reloading with a single global dict and a no-overwrite registration primitive.** The data structure is wrong for the problem.

---

### 3) The “default agent” used for delegation changed semantics (likely a regression)
`get_agent_factory(None)` now resolves to `openhands.sdk.subagent.builtins.default.get_default_agent()` (registry.py lines 253–256), which builds an `Agent` via `_agent_definition_to_factory()` (builtins/default.py lines 24–33).

This default sub-agent is **not the same** as the old preset default agent (which set condenser, `system_prompt_kwargs`, etc.). If the intent is “delegation subagents are lightweight”, that’s OK—but it’s a behavioral change that should be explicit and tested, because it will change output quality and context handling.

---

### 4) One of the registry tests is structurally wrong (so it doesn’t test what it claims)
`tests/sdk/subagent/test_subagent_registry.py::test_register_file_agents_project_priority` creates the user agent file here:
- `user_home = tmp_path / "fake_home" / "agents"` (line 49)
- then writes to `user_home / ".agents" / "shared-agent.md"` (lines 51–55)

…but `load_user_agents()` searches `Path.home() / ".agents/agents"` (load.py lines 81–96). That test never places the file under `.agents/agents/`, so the “user version” likely isn’t loaded at all. The test can pass even if user-loading is broken.

This is exactly how you end up merging something that “has tests” but still doesn’t work.

---

### 5) Tool naming / UX mismatch is unresolved
Your schema/tests use tool names like `ReadTool`, `GlobTool`, `Read`, `Write` (e.g., loader tests lines 30–56; schema tests lines 13–33). Real tool registry names in this repo are typically snake_case derived from class names (e.g., `GlobTool.name == "glob"` per ToolDefinition naming logic).

So users will write agent files like the Claude docs (`Read`, `Grep`, `Glob`, `Bash`) and it’ll *parse*, but may fail at runtime when tool resolution happens. If you want Claude compatibility, you need an alias/normalization layer or very explicit docs.

---

### 6) The “example usage” in the new registry module is stale
`openhands-sdk/openhands/sdk/subagent/registry.py` docstring still says:
```py
from openhands.tools.delegate import register_agent, Skill
```
(line 5)
That’s now wrong, because you removed that symbol from `openhands.tools.delegate`. This is basic “don’t ship docs that don’t run”.

---

### 7) Loader error handling is “log-and-ignore everything”
`_load_agents_from_dir()` catches `Exception` (load.py lines 133–140) and continues. For user config, that’s sometimes fine, but you’ll make debugging miserable. At least consider surfacing *which field* failed validation (or returning structured errors), especially since this is a brand new feature people will misconfigure.

---

## [TESTING GAPS]
- You removed plugin-loading tests (`tests/sdk/plugin/test_plugin_loading.py` deleted) but didn’t replace coverage for “plugin agents get registered and take precedence”.
- There’s no test that proves your *stated* precedence order actually holds across multiple conversations in the same process (which is where the global registry bug shows up).
- There’s no test for “edit agent file → new conversation picks up new content” (currently it won’t).

---

## VERDICT: **Needs rework**
The feature direction is fine, but the current implementation is built on a global registry that can’t express priority or reloading correctly, and you broke public imports without providing a compatibility path.

**Key insight:** You’re trying to model *layered configuration with precedence* using a single “first one wins” global dict. That’s the wrong data structure, and it forces broken semantics (no override, no reload, cross-workspace leakage).

openhands-sdk/openhands/sdk/subagent/registry.py
VascoSch92: Yes. It is a profile ID. Perhaps I should update the docs

enyst: Perfect, thank you! I'd love to not get stuck in what Anthropic likes, on this matter, heh :)

VascoSch92:
## Review of the roasted-review

**1) You broke userspace: removed public APIs without a compatibility layer**
True, but we discussed that it was Ok :-) 

**2) Global registry + register_*_if_absent() destroys your priority model and breaks “reload on new conversation”**
I believe this is not True. If we resume a conversation, it makes sense to use old hooks and plugins and not new one. On the other hands, if we start a new conversation, everything is loaded another time following the hierarcy. 

**3) The “default agent” used for delegation changed semantics (likely a regression)**
This is True. But just because we don't have a field for the condenser now in the Markdown.
There is already an issue to add that in the parsing.

**4) One of the registry tests is structurally wrong (so it doesn’t test what it claims)**
This is True. Corrected!

**5) Tool naming / UX mismatch is unresolved**
This is True but a very small nit

**6) The “example usage” in the new registry module is stale**
This is also True. Corrected!

**7) Loader error handling is “log-and-ignore everything”**
This is not the case as we have `exc_info=True`

VascoSch92:
> Maybe we can add an example, with something like:
> 
> ```
> ---
> name: code-reviewer
> description: Expert code review specialist. Proactively reviews code for quality, security, and maintainability. Use immediately after writing or modifying code.
> tools: Read, Grep, Glob, Bash
> model: inherit
> ---
> 
> You are a senior code reviewer ensuring high standards of code quality and security.
> 
> When invoked:
> 1. Run git diff to see recent changes... etc
> ```
> 
> Just for future record and ease of review it might be useful to update in the PR description too?

Do you mean creaiting docs for the file-based agents? Like skills and plugin?

It is on my todo list :-)

enyst:
> Do you mean creaiting docs for the file-based agents? Like skills and plugin?

I mean an example script in `examples/` directory (and maybe it's worth it to be a directory for that example, so that it has the markdown file separately, easy to understand... maybe)

enyst:
> **1) You broke userspace: removed public APIs without a compatibility layer**
> True, but we discussed that it was Ok :-)

Can we decide what is public API now, and add it to the SDK's `__init__` ?

VascoSch92:
> > Do you mean creaiting docs for the file-based agents? Like skills and plugin?
> 
> I mean an example script in `examples/` directory (and maybe it's worth it to be a directory for that example, so that it has the markdown file separately, easy to understand... maybe)

Added :-)

openhands-sdk/openhands/sdk/subagent/load.py
```diff
@@ -0,0 +1,158 @@
...
+    Note that `.agents/agents` definitions take precedence for duplicate names.
+
+    Only reads top-level `.md` files; subdirectories (like `skills/`) are
+    skipped.  `README.md` files are also skipped.
+
+    Args:
+        project_dir: project directory
+
+    Returns:
+        an empy list if no directories exist.
```
enyst:
```suggestion
        an empty list if no directories exist.
```

openhands-sdk/openhands/sdk/subagent/registry.py
```diff
@@ -0,0 +1,297 @@
...
+def agent_definition_to_factory(
+    agent_def: AgentDefinition,
+) -> Callable[["LLM"], "Agent"]:
+    """Create an agent factory closure from an ``AgentDefinition``.
+
+    The returned callable accepts a `LLM` class (the parent agent's LLM) and
+    builds a fully-configured `Agent` class.
+
+    - Tool names from `agent_def.tools` are mapped to `Tool` objects.
+    - The system prompt becomes an always-active `Skill` (`trigger=None`).
```
enyst: Isn't it the system prompt of this agent?

openhands-sdk/openhands/sdk/subagent/builtins/default.py
```diff
@@ -0,0 +1,37 @@
...
+        model="inherit",
+        tools=tool_names,
+    )
+    return AgentFactory(
+        factory_func=agent_definition_to_factory(agent_def),
+        description=agent_def.description or "Default general-purpose agent",
+    )
+
+
+def get_default_agent(enable_browser: bool = False) -> AgentFactory:
```
enyst: Sorry I'm dense, is this the default agent of the SDK, or is it a "default subagent" special definition?

openhands-sdk/openhands/sdk/subagent/registry.py
VascoSch92: For instance, the agent has the system prompt of openhands (the standard one) and attached its system prompt defined in the markdown.

openhands-sdk/openhands/sdk/subagent/builtins/default.py
VascoSch92: What do you mean?

The `SDK` doesn't currently have a default agent. While there is one in the `tools` module and `get_default_agent` follows the same logic, we cannot currently enforce the condenser. Otherwise, is the same set of tools.

This is a pending feature, as described in the related issue #2186

openhands-sdk/openhands/sdk/subagent/registry.py
enyst: There is a `suffix` that we add to the system prompt, maybe we want to use that in this case?

openhands-sdk/openhands/sdk/subagent/builtins/default.py
enyst: I mean, from the perspective of a client developer, this looks maybe a bit strange:
image

There is a function to `get_default_agent()` in the SDK, whose name says it’s the default agent. But if they look closer, it’s in the `subagent` module.

Many frameworks out there, including OpenHands, make/made a difference between top-level agent(s) and sub-agents. At some point we had sub-agents as “microagents” that could do only one thing (“commit agent”).

My point is it’s a little confusing. 🫠

openhands-sdk/openhands/sdk/subagent/registry.py
VascoSch92: Actually make sense :-) I didn't check that there it was the suffix args avaible. I changed.

openhands-sdk/openhands/sdk/subagent/builtins/default.py
VascoSch92: mmmh this is true.

I changed in `get_default_subagent()` as we are in the subagent module.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -347,6 +353,10 @@ def _ensure_plugins_loaded(self) -> None:
                 if plugin.hooks and not plugin.hooks.is_empty():
                     all_plugin_hooks.append(plugin.hooks)
 
+                # Collect agent definitions
+                if plugin.agents:
+                    all_plugin_agents.extend(plugin.agents)
+
```
simonrosenberg, line 360: @openhands is local_conversation.py the best place to register skill and agents?
Deep-dive into the code and understand how skills and agents work. And think hard about where they should be registered

VascoSch92:
> Thank you for the work on this!
> 
> I have a proposal that is… not clear enough, sorry, as I’m reading through it: the way this works is understandable to me, and the code design looks OK. But I feel agents need to read a lot of code to get how it should work. I wonder if we can inform the agent about this and its intentions, somewhere, in a simple and clear way.
> 
> Maybe an AGENTS.md. Maybe a page in docs that is meant for the code design behind the code. Kinda prefer both… if that makes sense.

open an issue and I can do that ;-) just to I don't forget.

enyst:
@openhands open an issue as said in https://github.com/OpenHands/software-agent-sdk/pull/2183#issuecomment-3960012826

Look at the PR and understand what I’m talking about, and include in the issue 2-3 proposals how we can document the code design / assumptions / invariants that this PR actually brings.

Issue #2018, opened 2026-02-11
Switch model profile on user message


neubig:
Since we now support different model profiles, it would be great to be able to switch on a user message.

For instance the user could say

```
/model opus really hard architectural design prompt
```

```
/model minimax straightforward implementation problem
```

The word after the `/model` slash command could be the model profile, and the rest could be a normal user message.

VascoSch92:
A couple of thoughts on the feature we want to implement:

It is possible to implement this directly in the codebase. However, doing so would add more complexity to classes that are already quite large, such as `LocalConversation` and `Agent`.

This feature feels like a skill to me, so another approach would be to implement it as one. Currently, the skill framework doesn't support this, but we already have tickets in place to enable it (e.g., #2053). Moreover, it seems that every command-style input (like `/<SOMETHING>`) currently passes through the skill machinery.

What is the priority for this feature? Do we want to build a custom solution now, or should we wait until the skill machinery is more mature so we can leverage that instead?

neubig:
It's not super-high priority, but also it'd be nice to have i think.

enyst:
To note, this issue has two parts:
- the ability to switch profile at runtime
- applying it on `/model` user message.

Maybe the second can depend on `skill` machinery, but the first doesn't, I think?

neubig:
Yeah, the first part is probably independent. I'm also not super-sure if the second part can rely on the skill machinery, but there might be a good design that makes that possible.

PR #2192, opened 2026-02-23
feat(llm): switch model profile on user message

VascoSch92:
## Summary

SDK side of the `/model <MODEL_NAME> [prompt] ` command (ref #2018 )

The SDK exposes a method `conversation.switch_profile()` to switch LLM, which client applications can use to implement commands like, e.g. `/model <MODEL_NAME>`: Switches the current LLM profile to the one specified.

Note:
- I’m not certain this is the best approach, but it’s functional and keeps the `LocalConversation` class clean.
- Happy to add an example if needed.
 
## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:63649ee-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-63649ee-python \
  ghcr.io/openhands/agent-server:63649ee-python
```

```
ghcr.io/openhands/agent-server:63649ee-golang-amd64
ghcr.io/openhands/agent-server:63649ee-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:63649ee-golang-arm64
ghcr.io/openhands/agent-server:63649ee-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:63649ee-java-amd64
ghcr.io/openhands/agent-server:63649ee-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:63649ee-java-arm64
ghcr.io/openhands/agent-server:63649ee-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:63649ee-python-amd64
ghcr.io/openhands/agent-server:63649ee-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:63649ee-python-arm64
ghcr.io/openhands/agent-server:63649ee-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:63649ee-golang
ghcr.io/openhands/agent-server:63649ee-java
ghcr.io/openhands/agent-server:63649ee-python
```

- Each variant tag (e.g., `63649ee-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `63649ee-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/subagent/schema.py
+++ b/openhands-sdk/openhands/sdk/subagent/schema.py
@@ -4,12 +4,16 @@ from __future__ import annotations
 
 import re
 from pathlib import Path
-from typing import Any, Final
+from typing import TYPE_CHECKING, Any, Final
 
 import frontmatter
 from pydantic import BaseModel, Field
 
 from openhands.sdk.hooks.config import HookConfig
+
+
+if TYPE_CHECKING:
+    from openhands.sdk.security.confirmation_policy import ConfirmationPolicyBase
 
 
 KNOWN_FIELDS: Final[set[str]] = {
@@ -22,6 +26,13 @@ KNOWN_FIELDS: Final[set[str]] = {
     "max_iteration_per_run",
     "hooks",
     "profile_store_dir",
+    "permission_mode",
+}
+
+_VALID_PERMISSION_MODES: Final[set[str]] = {
+    "always_confirm",
+    "never_confirm",
+    "confirm_risky",
 }
 
 
@@ -79,6 +90,20 @@ def _extract_examples(description: str) -> list[str]:
     return [m.strip() for m in matches if m.strip()]
 
 
+def _extract_permission_mode(fm: dict[str, object]) -> str | None:
+    """Extract permission_mode from frontmatter, defaulting to None (inherit parent)."""
+    raw = fm.get("permission_mode")
+    if raw is None:
+        return None
+    value = str(raw).strip().lower()
+    if value not in _VALID_PERMISSION_MODES:
+        raise ValueError(
+            f"Invalid permission_mode '{raw}'. "
+            f"Must be one of: {', '.join(sorted(_VALID_PERMISSION_MODES))}"
+        )
+    return value
+
+
 def _extract_max_iteration_per_run(fm: dict[str, object]) -> int | None:
     """Extract max iterations per run from frontmatter file."""
     max_iter_raw = fm.get("max_iteration_per_run")
@@ -130,6 +155,13 @@ class AgentDefinition(BaseModel):
     hooks: HookConfig | None = Field(
         default=None, description="Hook configuration for this agent"
     )
+    permission_mode: str | None = Field(
+        default=None,
+        description="How the subagent handles permissions. "
+        "None inherits the parent policy, 'always_confirm' requires "
+        "confirmation for every action, 'never_confirm' skips all confirmations, "
+        "'confirm_risky' only confirms actions above a risk threshold.",
+    )
     max_iteration_per_run: int | None = Field(
         default=None,
         description="Maximum iterations per run. "
@@ -145,6 +177,34 @@ class AgentDefinition(BaseModel):
         default_factory=dict, description="Additional metadata from frontmatter"
     )
 
+    def get_confirmation_policy(self) -> ConfirmationPolicyBase | None:
+        """Convert permission_mode to a ConfirmationPolicyBase instance.
+
+        Returns None when permission_mode is None (inherit parent policy).
+        """
+        if self.permission_mode is None:
+            return None
+
+        match self.permission_mode:
+            case "always_confirm":
+                from openhands.sdk.security.confirmation_policy import AlwaysConfirm
+
+                return AlwaysConfirm()
+            case "never_confirm":
+                from openhands.sdk.security.confirmation_policy import NeverConfirm
+
+                return NeverConfirm()
+            case "confirm_risky":
+                from openhands.sdk.security.confirmation_policy import ConfirmRisky
+
+                return ConfirmRisky()
+            case _:
+                # Should never reach here due to validation
+                # in _extract_permission_mode()
+                raise AssertionError(
+                    f"Unexpected permission_mode: {self.permission_mode}"
+                )
+
     @classmethod
     def load(cls, agent_path: Path) -> AgentDefinition:
         """Load an agent definition from a Markdown file.
@@ -156,6 +216,8 @@ class AgentDefinition(BaseModel):
         - skills (optional): Comma-separated skill names or list of skill names
         - model (optional): Model profile to use (default: 'inherit')
         - color (optional): Display color
+        - permission_mode (optional): How the subagent handles permissions
+          ('always_confirm', 'never_confirm', 'confirm_risky'). None inherits parent.
         - max_iterations_per_run: Max iteration per run
         - hooks (optional): List of applicable hooks
 
@@ -180,6 +242,7 @@ class AgentDefinition(BaseModel):
         color: str | None = _extract_color(fm)
         tools: list[str] = _extract_tools(fm)
         skills: list[str] = _extract_skills(fm)
+        permission_mode: str | None = _extract_permission_mode(fm)
         max_iteration_per_run: int | None = _extract_max_iteration_per_run(fm)
         profile_store_dir: str | None = _extract_profile_store_dir(fm)
         hooks: HookConfig | None = _extract_hooks(fm)
@@ -197,6 +260,7 @@ class AgentDefinition(BaseModel):
             color=color,
             tools=tools,
             skills=skills,
+            permission_mode=permission_mode,
             max_iteration_per_run=max_iteration_per_run,
             hooks=hooks,
             profile_store_dir=profile_store_dir,
--- a/openhands-tools/openhands/tools/delegate/__init__.py
+++ b/openhands-tools/openhands/tools/delegate/__init__.py
@@ -5,11 +5,12 @@ from openhands.tools.delegate.definition import (
     DelegateObservation,
     DelegateTool,
 )
-from openhands.tools.delegate.impl import DelegateExecutor
+from openhands.tools.delegate.impl import ConfirmationHandler, DelegateExecutor
 from openhands.tools.delegate.visualizer import DelegationVisualizer
 
 
 __all__ = [
+    "ConfirmationHandler",
     "DelegateAction",
     "DelegateObservation",
     "DelegateExecutor",
--- a/openhands-tools/openhands/tools/delegate/definition.py
+++ b/openhands-tools/openhands/tools/delegate/definition.py
@@ -18,6 +18,7 @@ from openhands.sdk.tool.tool import (
 
 if TYPE_CHECKING:
     from openhands.sdk.conversation.state import ConversationState
+    from openhands.tools.delegate.impl import ConfirmationHandler
 
 
 PROMPT_DIR = pathlib.Path(__file__).parent / "templates"
@@ -67,12 +68,18 @@ class DelegateTool(ToolDefinition[DelegateAction, DelegateObservation]):
         cls,
         conv_state: "ConversationState",
         max_children: int = 5,
+        confirmation_handler: "ConfirmationHandler | None" = None,
     ) -> Sequence["DelegateTool"]:
         """Initialize DelegateTool with a DelegateExecutor.
 
         Args:
             conv_state: Conversation state (used to get workspace location)
             max_children: Maximum number of concurrent sub-agents (default: 5)
+            confirmation_handler: Optional callback invoked when a sub-agent's
+                confirmation policy requires user approval.  Receives
+                `(agent_id, pending_actions)` and must return `True` to
+                approve or `False` to reject.  When `None`, pending actions
+                are auto-approved.
 
         Returns:
             List containing a single delegate tool definition
@@ -95,7 +102,10 @@ class DelegateTool(ToolDefinition[DelegateAction, DelegateObservation]):
 
         # Initialize the executor without parent conversation
         # (will be set on first call)
-        executor = DelegateExecutor(max_children=max_children)
+        executor = DelegateExecutor(
+            max_children=max_children,
+            confirmation_handler=confirmation_handler,
+        )
 
         # Initialize the parent Tool with the executor
         return [
--- a/openhands-tools/openhands/tools/delegate/impl.py
+++ b/openhands-tools/openhands/tools/delegate/impl.py
@@ -1,10 +1,15 @@
 """Implementation of delegate tool executor."""
 
 import threading
+from collections.abc import Callable
 from typing import TYPE_CHECKING
 
 from openhands.sdk.conversation.impl.local_conversation import LocalConversation
 from openhands.sdk.conversation.response_utils import get_agent_final_response
+from openhands.sdk.conversation.state import (
+    ConversationExecutionStatus,
+    ConversationState,
+)
 from openhands.sdk.logger import get_logger
 from openhands.sdk.subagent import get_agent_factory
 from openhands.sdk.tool.tool import ToolExecutor
@@ -12,9 +17,14 @@ from openhands.tools.delegate.definition import DelegateObservation
 
 
 if TYPE_CHECKING:
+    from openhands.sdk.event import ActionEvent
     from openhands.tools.delegate.definition import DelegateAction
 
 logger = get_logger(__name__)
+
+# Called when a sub-agent hits WAITING_FOR_CONFIRMATION.
+# Receives (agent_id, pending_actions) and returns True to approve, False to reject.
+ConfirmationHandler = Callable[[str, list["ActionEvent"]], bool]
 
 
 class DelegateExecutor(ToolExecutor):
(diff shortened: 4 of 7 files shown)


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -429,6 +437,123 @@ def _ensure_agent_ready(self) -> None:
...
+        """
+        usage_id = f"model-profile-{profile_name}"
+        try:
+            new_llm = self.llm_registry.get(usage_id)
+        except KeyError:
+            new_llm = self.profile_store.load(profile_name)
+            new_llm = new_llm.model_copy(update={"usage_id": usage_id})
+            self.llm_registry.add(new_llm)
+
+        self.agent = self.agent.model_copy(update={"llm": new_llm})
```
enyst: The `usage_id` is no longer the same, I think this is going to be problematic 🤔

Do we have to do this in local conversation? Could we do it in a `switch_llm` method in the registry, rather, so we can have the registry aware of who is the new `agent` llm?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -444,8 +569,13 @@ def send_message(self, message: str | Message, sender: str | None = None) -> Non
         # Ensure agent is fully initialized (loads plugins and initializes agent)
         self._ensure_agent_ready()
 
-        # Convert string to Message if needed
+        # Intercept /model commands before converting to Message
         if isinstance(message, str):
+            parsed = self._parse_model_command(message)
+            if parsed is not None:
+                profile_name, remaining_text = parsed
+                self._handle_model_command(profile_name, remaining_text, sender)
```
enyst: TBH you were not wrong to point out that this is a skill kind of thing, on the other hand IMHO this little snippet of code should be easy to change when we want to change… 🤔

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -429,6 +437,123 @@ def _ensure_agent_ready(self) -> None:
...
+            sender: Optional sender identifier.
+        """
+        if not profile_name:
+            self._emit_model_info()
+            return
+
+        try:
+            self._switch_model_profile(profile_name)
+        except (FileNotFoundError, ValueError) as e:
+            error_event = MessageEvent(
```
enyst: Does the LLM need to know this?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -429,6 +437,123 @@ def _ensure_agent_ready(self) -> None:
...
+                llm_message=Message(
+                    role="user",
+                    content=[TextContent(text=str(e))],
+                ),
+            )
+            self._on_event(error_event)
+            return
+
+        # Emit confirmation
+        confirm_event = MessageEvent(
```
enyst: Same question here, this would go to the LLM, I’m not sure it should?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: Actually i'm wondering if i can make it better...

Because i believe we can extract all the logic in a class (similar to skill?!) and just injecting the class inside the LocalConversation.

Something like

```python
class Command:
    def parse():
        ...

```

enyst (changes requested):
Thank you for this!

I think plugging this behavior in the conversation may be easy enough to do (and undo), so IMHO that is fine.

I think maybe we don’t want to send this information to the LLM, it’s just a toggle for the user, isn’t it?

VascoSch92:
> Thank you for this!
> 
> I think plugging this behavior in the conversation may be easy enough to do (and undo), so IMHO that is fine.
> 
> I think maybe we don’t want to send this information to the LLM, it’s just a toggle for the user, isn’t it?

The most important things was to do a first try to understand better the problem.

I believe i can make it better :) 

I will adress you suggestion ;)

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
enyst: You mean the `/model` issue? We have a `CommandDefinition` IIRC, it’s supposed to mean ‘skills executed with user triggers like /command” - so exactly this

What I meant here though, is something else: it’s swapping LLM profiles while
- keeping track of the same `usage_id` (or, I’d suggest to simplify into… keeping track of the `agent` llm)
- agent should use the last one
- restore conversation should use last one
- (possibly more fun stuff)

VascoSch92:
Hey @enyst 

I’ve made some changes based on your feedback. The goal was to pollute the `LocalConversation` class as little as possible, so I extracted the logic into a new class within a separate module.

Additionally, I've addressed your comments and added an example (give it a try, it’s fun!). I also added a flag to enable or disable model switching.

Let me know what you think!

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -136,6 +136,8 @@ def __init__(
                    state. If provided, secrets are encrypted when saving and
                    decrypted when loading. If not provided, secrets are redacted
                    (lost) on serialization.
+            allow_model_switching: Whether to allow switching between persisted
```
enyst: I wonder, is this necessary? 🤔

openhands-sdk/openhands/sdk/conversation/switch_model_handler.py
```diff
@@ -0,0 +1,86 @@
...
+            ValueError: If the profile is corrupted or invalid.
+        """
+        if profile_name in self.llm_registry.list_usage_ids():
+            new_llm = self.llm_registry.get(profile_name)
+        else:
+            new_llm = self.profile_store.load(profile_name)
+            new_llm = new_llm.model_copy(update={"usage_id": profile_name})
+            self.llm_registry.add(new_llm)
+
+        return agent.model_copy(update={"llm": new_llm})
```
enyst: There's a lot to like about the separation... 🤔

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -519,37 +454,25 @@ def _handle_model_command(
...
-                role="user",
-                content=[
-                    TextContent(
-                        text=f"Switched to model profile `{profile_name}` "
-                        f"(model: {self.agent.llm.model})"
-                    )
-                ],
-            ),
+        with self._state:
+            self._state.agent = self.agent
```
enyst, line 498: IIRC we do all changes to `state` under a lock, I think?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: It is very improbable that someone type /model as a mistake. But I wanted to give the possibility not to have that enable. It is just a flag. I can remove it if wished.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: Yes. But I should acquire the lock when I type with self._state right?

from the class `ConversationState`

```python
    def __enter__(self: Self) -> Self:
        """Context manager entry."""
        self._lock.acquire()
        return self
```

enyst:
> Hey @enyst
> 
> I’ve made some changes based on your feedback. The goal was to pollute the `LocalConversation` class as little as possible, so I extracted the logic into a new class within a separate module.
> 
> Additionally, I've addressed your comments and added an example (give it a try, it’s fun!). I also added a flag to enable or disable model switching.
> 
> Let me know what you think!

Thank you! Left a few tiny comments

I think maybe it doesn't get saved, so it won't be restored, the initial profile is? As in, restoring a conversation from the saved `base_state.json` 🤔 It's just a quick question

Just a heads up, I won't be around for a bit, back later (some number of hours)

VascoSch92:
> > Hey @enyst
> > I’ve made some changes based on your feedback. The goal was to pollute the `LocalConversation` class as little as possible, so I extracted the logic into a new class within a separate module.
> > Additionally, I've addressed your comments and added an example (give it a try, it’s fun!). I also added a flag to enable or disable model switching.
> > Let me know what you think!
> 
> Thank you! Left a few tiny comments
> 
> I think maybe it doesn't get saved, so it won't be restored, the initial profile is? As in, restoring a conversation from the saved `base_state.json` 🤔 It's just a quick question
> 
> Just a heads up, I won't be around for a bit, back later (some number of hours)

So I ran a small experiment to see what is happening.

If we save a convo and we restore it later: the last used LLM will be used. Which makes sense, right?

enyst:
> If we save a convo and we restore it later: the last used LLM will be used. Which makes sense, right?

Ahh. Because we always save/overwrite `base_state.json`, right?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
enyst: Right. I think I was thinking also of a couple other things: `base_state.json` changes, and changing the agent because we change the LLM. For your context, when we last considered the "statelessness" of the agent, we came up with a design "rule", or perhaps desire, to have strict, maybe _single_, pathways in the codebase, where the components of the agent can be modified. Please take a look at this:
- https://github.com/OpenHands/software-agent-sdk/issues/1451

I'd love your thoughts!

VascoSch92:
> Ahh. Because we always save/overwrite base_state.json, right?

i think yes. 

More generally, I think having this kind of machinery (command handlers injected into the conversation) could be useful for other features we want to implement. For example, OpenHands/extensions#373 is a great use case for this.

enyst:
> > Ahh. Because we always save/overwrite base_state.json, right?
> 
> i think yes.
> 
> More generally, I think having this kind of machinery (command handlers injected into the conversation) could be useful for other features we want to implement. For example, OpenHands/extensions#373 is a great use case for this.

I understand, here’s what I’m thinking here though:
- (old) we have `Skills` with triggers that are actionable, e.g. `/dothis` - they’re called ‘task triggers’ somewhere in the codebase, and can have `inputs` from the user
- (new) Agentskills / Plugins have `CommandDefinition` which is the same thing
- (newer) the command handlers in this PR, which are … the same thing 😅

I’d love it if we can do some reuse or consistency, unless maybe it already happened and I missed it?

VascoSch92:
> > > Ahh. Because we always save/overwrite base_state.json, right?
> > 
> > 
> > i think yes.
> > More generally, I think having this kind of machinery (command handlers injected into the conversation) could be useful for other features we want to implement. For example, OpenHands/extensions#373 is a great use case for this.
> 
> I understand, here’s what I’m thinking here though:
> 
>     * (old) we have `Skills` with triggers that are actionable, e.g. `/dothis` - they’re called ‘task triggers’ somewhere in the codebase, and can have `inputs` from the user
> 
>     * (new) Agentskills / Plugins have `CommandDefinition` which is the same thing
> 
>     * (newer) the command handlers in this PR, which are … the same thing 😅
> 
> 
> I’d love it if we can do some reuse or consistency, unless maybe it already happened and I missed it?

The problem I see with implementing this command via a skill is that the skill cannot access the conversation flow. A skill is essentially just information injection. While it can be complex and powerful (for example, exposing an MPC server), it cannot change the underlying LLM of the conversation by design.

Commands are subset of skills.

So perhaps my question is: what is really a Skill and what are the action that a Skill is responsible for?

enyst:
> So perhaps my question is: what is really a Skill and what are the action that a Skill is responsible for?

Oh, you're right. Tempted to say "you're absolutely right!" - but then you'd think Sonnet took over this human 😅 

Of course. The first two in my list are similar (they're skills), but the third is not like the others! 😂  

I asked the agent to remind me stuff about Agentskills, but then we went down the rabbit hole rethinking all this.

I actually think maybe `/model` (processing user input) doesn't belong in the SDK layer.

At least, maybe it's worth rethinking, and I'll follow up, but just for the record:
- this is a [tiny test CLI for LLM switch](https://github.com/enyst/agent-sdk/pull/17/changes#diff-58838040e9505830b3edafbcec7450e084c80fa45c9f56ffc5ce891e2b015290R163), a completely separate file, since it's not exactly SDK layer, it's application layer
- below is the place in the rabbit hole where I'll stop for the night: 😅 

(OpenHands-GPT-5.2, we were at the point where I was nagging it to find a way to keep it in the SDK as extensibility opportunity)

_Edited_: oh, actually I forgot we have `public share` these days:
(last thing there)
https://app.all-hands.dev/shared/conversations/e486c416d7c24e8dbe838f53c6054a61

VascoSch92:
> Oh, you're right. Tempted to say "you're absolutely right!" - but then you'd think Sonnet took over this human 😅
😅 🤖 😄 

I’ve been catching up on the conversation. We agree that `/model` is a totally different concept from `Skills`, `Commands`, or `Hooks`; let's call it a **control command**.

Should we implement this control command in the CLI? 

I think this could be a good alternative. 
Since we have the `LLMProfileStore` now, I believe all the pieces are in place.

VascoSch92:
> > Oh, you're right. Tempted to say "you're absolutely right!" - but then you'd think Sonnet took over this human 😅
> > 😅 🤖 😄
> 
> I’ve been catching up on the conversation. We agree that `/model` is a totally different concept from `Skills`, `Commands`, or `Hooks`; let's call it a **control command**.
> 
> Should we implement this control command in the CLI?
> 
> I think this could be a good alternative. Since we have the `LLMProfileStore` now, I believe all the pieces are in place.

@enyst :-)

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
neubig: Maybe YAGNI

neubig:
cc @enyst if you still think changes should be made could you list up the changes that need to be made, otherwise approve?

enyst (commented):
Could we maybe do it like this:
- SDK does the profile switch
  - SDK exposes a little API for it, e.g. maybe `LLMRegistry.switch_profile()` or even `conversation.switch_profile` if we need to place it into conversation (it does a little much though)
- CLI owns user interaction 
  - parses /model as command, calls the method to do stuff

WDYT?

enyst:
Maybe like this separation: https://github.com/OpenHands/software-agent-sdk/pull/2336#issuecomment-4010737777

VascoSch92:
@enyst 

I have rewritten the code to follow the comments. Now it is truly minimalistic, and the responsibility for catching the command is left to the CLI.

Note that I used model_copy twice. I'm forced to do so because the `ProfileStore` uses the profile name as a primary key, while the LLM registry uses the `usage_id`. This accounts for the discrepancy.

enyst:
> Note that I used model_copy twice. I'm forced to do so because the `ProfileStore` uses the profile name as a primary key, while the LLM registry uses the `usage_id`. This accounts for the discrepancy.

Noted, thank you, will do!

I just want to note quickly what OpenHands-GPT-5.4 just told me in another PR:
(I told it to remove some caching of base url / provider it was trying to add, because LLM instances are supposed to be immutable -ish)

image


Yes, `usage_id` and `profile_id` are a problem to solve. I would be happy even with the solution I found elsewhere, which was basically to remove random `usage_id`, restrict it to `agent`, maybe at most `condenser`.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -464,11 +466,37 @@ def _ensure_agent_ready(self) -> None:
 
             # Register LLMs in the registry (still holding lock)
             self.llm_registry.subscribe(self._state.stats.register_llm)
+            registered = set(self.llm_registry.list_usage_ids())
             for llm in list(self.agent.get_all_llms()):
-                self.llm_registry.add(llm)
+                if llm.usage_id not in registered:
+                    self.llm_registry.add(llm)
```
enyst, line 472: Sorry, not about this PR, but this looks curious actually. Are we saying that the agent may have gotten an LLM instance that wasn't registered? Do we know that?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -464,11 +466,37 @@ def _ensure_agent_ready(self) -> None:
...
+            FileNotFoundError: If the profile does not exist.
+            ValueError: If the profile is corrupted or invalid.
+        """
+        usage_id = f"profile:{profile_name}"
+        try:
+            new_llm = self.llm_registry.get(usage_id)
+        except KeyError:
+            new_llm = self._profile_store.load(profile_name)
+            new_llm = new_llm.model_copy(update={"usage_id": usage_id})
+            self.llm_registry.add(new_llm)
```
enyst, line 495: Really only a tiny thought, this code could be out of here and in the registry I think 

Not for this PR, just thinking of responsibilities; now the Conversation knows both `llm registry` and `llm profile store` and does stuff between them, even though those two are so related they could arguably be the same. Maybe worth thinking about it later, sorry

enyst:
On a side note, as you have seen, I am slowly growing a bit concerned with how we are doing things 😄 

If we want to change the underlying design of the codebase on the current statelessness aspect, we can, of course we can, we just could maybe give it some good thought and submit it for discussion. I do think maybe we shouldn't do a lot of "sins" or I know my slow brain is guaranteed to lose track 😢 ... and I also know others may keep theirs for longer, but last I checked we're humans here so... idk if forever! 😹

PR #2224, opened 2026-02-26
fix(tool): race condition in dynamic Action wrapper class creation

### LINKED ISSUE #2199: bug(sdk): Subagent spawn fails with Pydantic `ValidationError`: Duplicate class definition for `BrowserCloseTabActionWithRiskWithSummary`
## Description                                                                                                                                                                                                                                                                  

When spawning subagents via the delegate tool, all subagent conversations immediately crash during `Event` model construction with Pydantic validation errors. The subagents never execute their delegated tasks.

## Steps to Reproduce
1. Start a conversation with the OpenHands agent
2. Ask it to spawn multiple subagents:  "run in parallel 3 subagents"
3. The spawn succeeds: "Successfully spawned 3 sub-agents: subagent1 (default), subagent2 (default), subagent3 (default)"
4. Delegate tasks to the subagents: "search for pizza" (subagent1), "search for restaurants" (subagent2), "search for colibri" (subagent3)
5. All 3 subagents fail with identical `ValidationErrors`

## Error
Each subagent produces 17 validation errors (one per tool). The error message is identical for every tool:

```text
  17 validation errors for Event
  tools.0.action_type
    Value error, Duplicate class definition for openhands.sdk.tool.schema.Action:
      abc.BrowserCloseTabActionWithRiskWithSummary : abc.BrowserCloseTabActionWithRiskWithSummary
      [type=value_error, input_value='TerminalAction', input_type=str]

  tools.1.action_type
    Value error, Duplicate class definition for openhands.sdk.tool.schema.Action:
      abc.BrowserCloseTabActionWithRiskWithSummary : abc.BrowserCloseTabActionWithRiskWithSummary
      [type=value_error, input_value='FileEditorAction', input_type=str]

  ... (repeated for all 17 tools)
```

After the subagents fail, the parent agent also crashes with a `ConversationErrorEvent` containing 6 validation errors for its own reduced tool set (`TerminalAction`, `FileEditorAction`, `TaskTrackerAction`, `DelegateAction`, `FinishAction`, `ThinkAction`).

Affected tools in subagents (17): TerminalAction, FileEditorAction, TaskTrackerAction, BrowserNavigateAction, BrowserClickAction, BrowserGetStateAction, BrowserGetContentAction, BrowserTypeAction, BrowserScrollAction, BrowserGoBackAction, BrowserListTabsAction,
  BrowserSwitchTabAction, BrowserCloseTabAction, BrowserGetStorageAction, BrowserSetStorageAction, FinishAction, ThinkAction

## Root Cause Analysis
The `WithRiskWithSummary` wrapper classes (e.g. `BrowserCloseTabActionWithRiskWithSummary`) are generated dynamically at runtime — likely by composing a base action class with risk-level and summary annotations.

When a subagent is spawned, the tool schema registration runs again within the same process (or with shared module-level state). This causes the dynamic class to be re-created with the same name, resulting in two distinct Python class objects both registered under `abc.BrowserCloseTabActionWithRiskWithSummary` in Pydantic's discriminated union registry.

When Pydantic validates the Event model's `tools[N].action_type` discriminated union, it detects two classes mapped to the same discriminator key and raises a `ValueError`, which cascades to every tool in the schema, not just `BrowserCloseTabAction`.

Key observation: the class name itself (`BrowserCloseTabActionWithRiskWithSummary`) appears on both sides of the duplicate error message, confirming it's the exact same name registered twice rather than a naming collision between two different classes.

## Solution

The fix should ensure that dynamically generated wrapper classes are only created once and reused on subsequent calls. There are a few approaches, in order of preference:

### Option A: Cache dynamic classes in a registry (recommended)
Add a module-level cache dictionary that maps (base_class, decorators) to the generated wrapper class. Before creating a new class, check the cache:
```python
  _ACTION_CLASS_CACHE: dict[tuple, type] = {}

  def create_action_with_risk_and_summary(base_class: type) -> type:
      cache_key = (base_class, "WithRiskWithSummary")
      if cache_key in _ACTION_CLASS_CACHE:
          return _ACTION_CLASS_CACHE[cache_key]

      # existing dynamic class creation logic
      new_class = type(
          f"{base_class.__name__}WithRiskWithSummary",
          (base_class,),
          { ... }
      )

      _ACTION_CLASS_CACHE[cache_key] = new_class
      return new_class
```

This is the smallest change and directly prevents the duplicate registration.

### Option B: Guard Pydantic discriminator registration

Before registering an action type in the discriminated union, check whether a class with the same name already exists:

```python
  def register_action_type(action_class: type):
      existing = _ACTION_REGISTRY.get(action_class.__name__)
      if existing is not None:
          if existing is action_class:
              return  # already registered, no-op
          raise ValueError(f"Name collision: {action_class.__name__}")
      _ACTION_REGISTRY[action_class.__name__] = action_class
```

### Option C: Isolate subagent processes

If subagents are intentionally designed to share a process, the class caching in Option A is the right fix. However, if subagents are supposed to run in isolated processes, the bug may be that they're incorrectly sharing the parent's module state. In that case, ensure subagent conversations are started in a fresh subprocess or with a clean module import state (e.g., via importlib.reload or multiprocessing with spawn start method).

VascoSch92:
## Summary

Fix #2199 

- Add threading.Lock to `create_action_type_with_risk()` and `_create_action_type_with_summary()` to prevent concurrent threads from creating duplicate wrapper classes with the same name
- Without the lock, two subagent threads could both see a cache miss, both call `type()`, and register two distinct class objects under the same `__name__ `, causing `_get_checked_concrete_subclasses(Action)` to raise `ValueError("Duplicate class definition ...")`

## Tests
- Added concurrent threading tests in `tests/sdk/tool/test_tool.py` that spawn 8 threads hitting the same wrapper function simultaneously

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:4bc57c3-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-4bc57c3-python \
  ghcr.io/openhands/agent-server:4bc57c3-python
```

```
ghcr.io/openhands/agent-server:4bc57c3-golang-amd64
ghcr.io/openhands/agent-server:4bc57c3-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:4bc57c3-golang-arm64
ghcr.io/openhands/agent-server:4bc57c3-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:4bc57c3-java-amd64
ghcr.io/openhands/agent-server:4bc57c3-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:4bc57c3-java-arm64
ghcr.io/openhands/agent-server:4bc57c3-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:4bc57c3-python-amd64
ghcr.io/openhands/agent-server:4bc57c3-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:4bc57c3-python-arm64
ghcr.io/openhands/agent-server:4bc57c3-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:4bc57c3-golang
ghcr.io/openhands/agent-server:4bc57c3-java
ghcr.io/openhands/agent-server:4bc57c3-python
```

- Each variant tag (e.g., `4bc57c3-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `4bc57c3-python-amd64`) are also available if needed

### DIFF
--- a/.github/scripts/check_agent_server_rest_api_breakage.py
+++ b/.github/scripts/check_agent_server_rest_api_breakage.py
@@ -1,0 +1,398 @@
+#!/usr/bin/env python3
+"""REST API breakage detection for openhands-agent-server.
+
+This script compares the current OpenAPI schema for the agent-server REST API against
+the previous published version on PyPI.
+
+Policies enforced (mirrors the SDK's Griffe checks, but for REST):
+
+1) Deprecation-before-removal
+   - If a REST operation (path + HTTP method) is removed, it must have been marked
+     `deprecated: true` in the previous release.
+
+2) MINOR version bump
+   - If a breaking REST change is detected, the current version must be at least a
+     MINOR bump compared to the previous release.
+
+The breakage detection currently focuses on compatibility rules that are robust to
+OpenAPI generation changes:
+- Removed operations
+- New required parameters
+- Request bodies that became required
+- New required fields in JSON request bodies (best-effort)
+
+If the previous release schema can't be fetched (e.g., network / PyPI issues), the
+script emits a warning and exits successfully to avoid flaky CI.
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+import sys
+import tempfile
+import tomllib
+import urllib.request
+from collections.abc import Iterable
+from dataclasses import dataclass
+from pathlib import Path
+
+from packaging import version as pkg_version
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+AGENT_SERVER_PYPROJECT = REPO_ROOT / "openhands-agent-server" / "pyproject.toml"
+PYPI_DISTRIBUTION = "openhands-agent-server"
+
+
+_HTTP_METHODS = (
+    "get",
+    "post",
+    "put",
+    "patch",
+    "delete",
+    "options",
+    "head",
+    "trace",
+)
+
+
+@dataclass(frozen=True, slots=True)
+class OperationKey:
+    method: str
+    path: str
+
+
+def _read_version_from_pyproject(pyproject: Path) -> str:
+    data = tomllib.loads(pyproject.read_text())
+    try:
+        return str(data["project"]["version"])
+    except KeyError as exc:  # pragma: no cover
+        raise SystemExit(
+            f"Unable to determine project version from {pyproject}"
+        ) from exc
+
+
+def _fetch_pypi_metadata(distribution: str) -> dict:
+    req = urllib.request.Request(
+        url=f"https://pypi.org/pypi/{distribution}/json",
+        headers={"User-Agent": "openhands-agent-server-openapi-check/1.0"},
+        method="GET",
+    )
+    with urllib.request.urlopen(req, timeout=10) as response:
+        return json.load(response)
+
+
+def _get_previous_version(distribution: str, current: str) -> str | None:
+    try:
+        meta = _fetch_pypi_metadata(distribution)
+    except Exception as exc:  # pragma: no cover
+        print(
+            f"::warning title={distribution} REST API::Failed to fetch PyPI metadata: "
+            f"{exc}"
+        )
+        return None
+
+    releases = list(meta.get("releases", {}).keys())
+    if not releases:
+        return None
+
+    current_parsed = pkg_version.parse(current)
+    older = [rv for rv in releases if pkg_version.parse(rv) < current_parsed]
+    if not older:
+        return None
+
+    return max(older, key=pkg_version.parse)
+
+
+def _generate_current_openapi() -> dict:
+    from openhands.agent_server.api import create_app
+
+    return create_app().openapi()
+
+
+def _generate_openapi_for_version(version: str) -> dict | None:
+    """Generate OpenAPI schema for a published agent-server version.
+
+    Returns None on failure so callers can treat it as a best-effort comparison.
+    """
+
+    with tempfile.TemporaryDirectory(prefix="agent-server-openapi-") as tmp:
+        venv_dir = Path(tmp) / ".venv"
+        python = venv_dir / "bin" / "python"
+
+        try:
+            subprocess.run(
+                [
+                    "uv",
+                    "venv",
+                    str(venv_dir),
+                    "--python",
+                    sys.executable,
+                ],
+                check=True,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                text=True,
+            )
+            openhands_packages = (
+                "openhands-agent-server",
+                "openhands-sdk",
+                "openhands-tools",
+                "openhands-workspace",
+            )
+            packages = [f"{name}=={version}" for name in openhands_packages]
+
+            subprocess.run(
+                [
+                    "uv",
+                    "pip",
+                    "install",
+                    "--python",
+                    str(python),
+                    *packages,
+                ],
+                check=True,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                text=True,
+            )
+
+            program = (
+                "import json; "
+                "from openhands.agent_server.api import create_app; "
+                "print(json.dumps(create_app().openapi()))"
+            )
+            result = subprocess.run(
+                [str(python), "-c", program],
+                check=True,
+                capture_output=True,
+                text=True,
+            )
+            return json.loads(result.stdout)
+        except subprocess.CalledProcessError as exc:
+            output = (exc.stdout or "") + ("\n" + exc.stderr if exc.stderr else "")
+            excerpt = output.strip()[-1000:]
+            print(
+                f"::warning title={PYPI_DISTRIBUTION} REST API::Failed to generate "
+                f"OpenAPI schema for v{version}: {exc}\n{excerpt}"
+            )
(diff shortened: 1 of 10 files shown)


VascoSch92:
@OpenHands Can you double-check that the tests is really reproducing the bug and that the fix is good?

enyst:

enyst (approved):
Thank you!

enyst:
(OpenHands-GPT-5.2)

🟡 **Taste rating: Acceptable (but messy)**

This PR *does* fix a real race, but it’s also doing a bunch of unrelated stuff (built-in subagents, delegation metrics semantics, example churn). That’s not “good taste”: it makes review harder and increases the chance you ship an unrelated regression under the cover of a concurrency fix.

---

## [CRITICAL ISSUES] (things that can bite real users)

- **[openhands-sdk/openhands/sdk/subagent/registry.py, Lines 150–156] Tool “validation” that doesn’t validate**
  - You check `tool_name not in registered_tools`, log an info line, and then still build `Tool(name=tool_name)` anyway.
  - This is the worst of both worlds:
    - It **doesn’t prevent typos** (so the real failure still happens later, further from the source).
    - It **spams logs** in any flow where tools aren’t registered yet (which your own message explicitly anticipates).
  - **Fix suggestion:** either (a) fail fast with a clear error (best for correctness), or (b) remove this check entirely. If you *must* keep it, at least make it `debug` and include the agent name so it’s actionable.

- **[openhands-sdk/openhands/sdk/subagent/builtins/explore.md, Lines 9–31] “Read-only” agent with a write-capable tool**
  - You label `explore` as “read-only” but give it the raw `terminal` tool. That’s not a constraint, it’s a vibe.
  - If the intent is “policy-only”, fine—but then don’t market it as read-only in a way that implies enforcement.
  - **Fix suggestion:** reword to “read-only by instruction/policy” (or introduce a genuinely constrained toolset).

---

## [IMPROVEMENT OPPORTUNITIES] (good taste / simplification)

- **[openhands-sdk/openhands/sdk/tool/tool.py, Lines 44–46, 481–536] Global lock is OK, but it’s coarse and manual**
  - Yes, the lock fixes the TOCTOU cache miss. But you now serialize **all** wrapper creation (risk + summary) behind one global lock.
  - Also: you’re hand-rolling caching+locking when Python already has battle-tested options.
  - **Fix suggestion:** consider replacing both dicts with `@functools.lru_cache(maxsize=None)` / `@functools.cache` on the wrapper constructors. Those use internal locking and make the code smaller and harder to get wrong. If you keep the lock, consider `RLock` to avoid future deadlocks if anyone ever nests these calls.

- **[openhands-tools/openhands/tools/delegate/impl.py, Lines 221–248] Cross-thread dict writes rely on the GIL**
  - `results[...] = ...` and `errors[...] = ...` happen from multiple threads with no lock.
  - In CPython today this “usually works” because of the GIL, but it’s not a great pattern to normalize in an SDK, and it makes refactors riskier.
  - **Fix suggestion:** either protect writes with a `threading.Lock`, or push results into a `queue.Queue` and merge after join.

- **[tests/sdk/subagent/test_builtin_agents.py, Lines 54–61] Brittle test asserting exact default tool list**
  - `assert agent_tool_names["default"] == [...]` will break the moment you add/remove a default tool for legitimate reasons.
  - **Fix suggestion:** assert key invariants (e.g., `{"terminal","file_editor","task_tracker"}.issubset(...)`) unless you *explicitly* want the test suite to enforce that default tool list is frozen.

---

## [TESTING GAPS / FLAKINESS RISKS]

- **[tests/sdk/tool/test_tool.py, Lines 199–268] Concurrency test may not actually reproduce the bug reliably**
  - Barrier + 8 threads helps, but this is still scheduler-dependent: on some machines/load, the “broken” implementation might pass, making this more of a “regression guard” than a reproduction.
  - Also: the duplicated docstring is sloppy, and the “Ref:” URL looks bogus.
  - **Fix suggestion:**
    - Loop the race attempt (e.g., 100–1000 iterations) to increase failure probability on the buggy version.
    - Clean up the double docstring and link.
    - Use a thread-safe collector (`Queue`) for results to avoid teaching “shared list append is fine” as an example.

---

## VERDICT
✅ **Worth merging for the race fix**, *but* the PR is doing too much at once and includes a couple of “looks like validation but isn’t” design choices (tool registration logging) plus policy-vs-enforcement confusion (“read-only” explore with terminal).

**KEY INSIGHT:** The locking fix is fine; the bigger risk here is *scope creep* and introducing behavioral/logging changes that aren’t directly related to the race and are harder to reason about.

enyst:
> * You label `explore` as “read-only” but give it the raw `terminal` tool. That’s not a constraint, it’s a vibe.

😅

VascoSch92:
@enyst  Actually, I think there might be an issue with the review you requested. The reviewer is referring to changes that aren't present in this PR. For example, your comment above relates to the built-in agent explorer, but that agent wasn't added in this PR. I suspect the reviewer had access to the wrong diff.

Issue #2186, opened 2026-02-23
feat(delegation): Advanced Features for Markdown-based Agents


VascoSch92:
### Summary

The PR OpenHands/software-agent-sdk#2183 added Markdown-Based agents. However, several advanced features are currently unsupported.

This issue tracks the missing features we intend to support in the SDK.

### Missing features (ordered by implementation difficulty: low to high)

- [X] `max_iteration_per_run`: Maximum number of agentic steps before the subagent terminates. (OpenHands/software-agent-sdk#2263)
- [X] `profile_store_dir`: Profile store directory from where loading profiles for `model` ([#2340](<https://github.com/OpenHands/software-agent-sdk/issues/2340>) )
- [ ] `working_dir`: working directory for the subagent. \[TO DISCUSS\]
- [ ] `condenser`: Add a condenser to the subagent. ([#2349](<https://github.com/OpenHands/software-agent-sdk/issues/2349>))
- [X] `permission_mode`: How the subagent handles permissions. ([#2345](<https://github.com/OpenHands/software-agent-sdk/issues/2345>))
- [X] `skills`: Skills to load into the subagent’s context. ([#2260](<https://github.com/OpenHands/software-agent-sdk/issues/2260>))
- [X] `hooks`: Lifecycle hooks scoped to this agent. ([#2347](<https://github.com/OpenHands/software-agent-sdk/issues/2347>))
- [ ] `mcp_servers`: MCPs available to the agent. ([#2348](<https://github.com/OpenHands/software-agent-sdk/issues/2348>)
- [ ] `disallowed_tools`: Tools to restrict; these will be removed from the inherited or specified list. \[TO DISCUSS\]
- [ ] `disallowed_commands`: Commands to restrict; these will be removed from the inherited or specified list.
- [ ] `memory`: Persistent memory scope.
- [X] `model`: Currently, we only support inheriting from the parent model. ([#2258](<https://github.com/OpenHands/software-agent-sdk/issues/2258>))
- [ ] `background`: Always run the agent as a background task. (This is blocked by OpenHands/software-agent-sdk#2047 )

### Important

Every newly added feature should be backward compatible with the previous markdown format.

enyst:
> model: 

It's fine if we don't support more than `inherit` yet, just for the record: maybe this is a profile name, not a model name exactly. Obviously Anthropic doesn't want more models than their own, but we do. 😹 

> max_turns: Maximum number of agentic turns before the subagent terminates.

[...]

PR #2260, opened 2026-03-02
feat(sdk/subagent): Support skills in file-based Agents

VascoSch92:
## Summary

(ref #2186)

- Add `skills` field to `AgentDefinition`, allowing subagents to reference skills by name in markdown frontmatter (comma-separated or YAML list)
- `agent_definition_to_factory` resolves skill names to `Skill` objects at factory creation time using `load_project_skills` and `load_user_skills` (project takes priority over user)                                     
- Refactor `schema.py` to extract frontmatter parsing into dedicated helpers (`_extract_color`, `_extract_tools`, `_extract_skills`) and promote `KNOWN_FIELDS` to a module-level constant

## Example

```text
  ---
  name: code-reviewer
  skills: code-review, linting
  tools:
    - file_editor
  ---

  You are a code review specialist.
```

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:5aa1d20-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-5aa1d20-python \
  ghcr.io/openhands/agent-server:5aa1d20-python
```

```
ghcr.io/openhands/agent-server:5aa1d20-golang-amd64
ghcr.io/openhands/agent-server:5aa1d20-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:5aa1d20-golang-arm64
ghcr.io/openhands/agent-server:5aa1d20-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:5aa1d20-java-amd64
ghcr.io/openhands/agent-server:5aa1d20-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:5aa1d20-java-arm64
ghcr.io/openhands/agent-server:5aa1d20-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:5aa1d20-python-amd64
ghcr.io/openhands/agent-server:5aa1d20-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:5aa1d20-python-arm64
ghcr.io/openhands/agent-server:5aa1d20-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:5aa1d20-golang
ghcr.io/openhands/agent-server:5aa1d20-java
ghcr.io/openhands/agent-server:5aa1d20-python
```

- Each variant tag (e.g., `5aa1d20-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `5aa1d20-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/context/skills/skill.py
+++ b/openhands-sdk/openhands/sdk/context/skills/skill.py
@@ -1019,8 +1019,8 @@ def load_public_skills(
 def load_available_skills(
     work_dir: str | Path | None = None,
     *,
-    include_user: bool = True,
-    include_project: bool = True,
+    include_user: bool = False,
+    include_project: bool = False,
     include_public: bool = False,
 ) -> dict[str, Skill]:
     """Load and merge skills from SDK-level sources with consistent precedence.
--- a/openhands-sdk/openhands/sdk/subagent/registry.py
+++ b/openhands-sdk/openhands/sdk/subagent/registry.py
@@ -142,7 +142,9 @@ def agent_definition_to_factory(
     if agent_def.skills:
         from openhands.sdk.context.skills import load_available_skills
 
-        available = load_available_skills(work_dir, include_public=False)
+        available = load_available_skills(
+            work_dir, include_user=True, include_project=True, include_public=False
+        )
 
         for name in agent_def.skills:
             if name not in available:


openhands-sdk/openhands/sdk/subagent/registry.py
```diff
@@ -113,21 +113,46 @@ def register_agent_if_absent(
...
     """
+    # Resolve skills eagerly at factory creation time using existing loaders.
+    # Priority: project skills override user skills (user loaded first, project second).
+    resolved_skills: list = []
+    if agent_def.skills:
+        from openhands.sdk.context.skills import load_project_skills, load_user_skills
+
+        available = {s.name: s for s in load_user_skills()}
+        if work_dir:
+            available.update({s.name: s for s in load_project_skills(work_dir)})
```
enyst: @openhands Compare the behavior that this PR brings on loading skills for sub-agents, with the behavior the main, default agent has on loading skills.
- are there inconsistencies?
- can we simplify / reuse anything?

VascoSch92: Mmmh interesting...

enyst: Just to note quickly, this was the "summary" response the resolver is prompted to give. Its initial, full response was this

---

(OpenHands-GPT-5.2)

### What the “main/default agent” does today (skill loading)

In the SDK, the *default* (top-level) agent doesn’t “resolve skills by name”. Instead, it’s typically constructed with a **list of `Skill` objects** already loaded, e.g.:

- **Project/repo skills** are usually loaded explicitly by the caller via `load_project_skills(work_dir)` (this includes `.agents/skills`, `.openhands/skills`, legacy microagents, and “third-party” files like `AGENTS.md` from `work_dir` and git root).
- **User/public skills** are optionally auto-loaded *inside* `AgentContext` via flags:
  - `AgentContext(load_user_skills=True)` → appends user skills if not already present
  - `AgentContext(load_public_skills=True)` → appends public skills if not already present  
  Ordering matters: user skills are loaded before public skills, and both are “add if missing” (explicit skills win).

So precedence for the main agent usually ends up as:

- **explicit skills passed to `AgentContext` (often project skills)** win over
- **user skills**, which win over
- **public skills** (if enabled)

(See `openhands-sdk/openhands/sdk/context/agent_context.py` model validators.)

In the agent-server layer there’s also `load_all_skills()` which merges multiple sources with **project highest precedence** (project > org > user > public > sandbox), but that’s separate from SDK `AgentContext` behavior.

---

### What this PR adds for file-based sub-agents

In `agent_definition_to_factory()` (in `openhands-sdk/openhands/sdk/subagent/registry.py`), the PR introduces **name-based resolution**:

- Load all **user skills**: `load_user_skills()`
- Load all **project skills** (if `work_dir` provided): `load_project_skills(work_dir)`
- Put them in a `{name -> Skill}` map where **project overwrites user**
- For each `agent_def.skills` name, pick the corresponding `Skill` or raise `ValueError` if missing
- Pass only those resolved `Skill` objects into `AgentContext(skills=...)`

So for subagents in this PR, precedence is:

- **project skill name wins over same-named user skill**

---

### Inconsistencies (vs “main/default agent” behavior)

Mostly the semantics are aligned, but there are a few real differences worth calling out:

1. **Public skills aren’t considered for subagents**
   - Main agent can enable `load_public_skills=True` (or agent-server can include public/org/sandbox).
   - Subagent resolution only looks at user + project directories.
   - Result: a skill name that exists publicly (or org/sandbox-generated) can be available to the main agent but *unresolvable* for a file-based subagent.

2. **Plugin-provided skills aren’t considered**
   - The main agent can have plugin skills merged into its `AgentContext` (see `Plugin.add_skills_to()`).
   - Subagent resolution doesn’t look at plugin skill sets at all—only filesystem-loaded skills.
   - So a plugin agent definition referencing a “plugin skill name” won’t resolve unless that skill also exists in user/project skill dirs.

3. **Different “merge” mechanism (but same outcome for user vs project)**
   - Main agent’s merging is “append-if-missing” (explicit skills win; duplicates are skipped with warnings).
   - Subagent does a dict merge (user map, then `update()` with project).
   - Practically, for user-vs-project precedence this is consistent (project wins), but logging/behavior around duplicates differs.

4. **Eager failure mode**
   - Subagent: if a named skill can’t be found (including “failed to load”), it hard-errors with `ValueError`.
   - Main agent: if a skill fails to load during a bulk load, it’s typically logged and skipped; there is no “missing name” concept unless you add one.

---

### Can we simplify / reuse anything?

Yes—there’s a clear reuse opportunity: **centralize “build a skill catalog with precedence” in one SDK helper**, then use it in both places.

Concretely, something like:

- `load_available_skills(work_dir: Path | None, *, include_user: bool, include_project: bool, include_public: bool) -> dict[str, Skill]`
  - Encodes precedence in one place (project > user > public, etc.)
  - Lets subagent resolution become a thin wrapper: `catalog = load_available_skills(...); resolved = [catalog[name] for name in names]`
  - Lets `AgentContext` auto-load logic reuse the same precedence rules instead of re-implementing merging in validators
  - (Optionally) agent-server’s `merge_skills()` could reuse the same helper too, but at minimum keeping this inside the SDK would avoid the current duplication.

That would reduce the chance of subtle drift (e.g., “default agent supports public/org/sandbox/plugin skills but subagents don’t”) because you could intentionally choose which sources are available for subagents, rather than having that be an emergent property of separate implementations.

VascoSch92: It is clear that we will have an inconsistence. We are starting from the fact that we pass skill directly to the agent via a markdown file. 

I wanted to have more or less the same structure, i.e., project and the user skills.

Programmatic skills can not be passed and plugin skills i don't know if we want that.

Otherwise, public skills could be interesting.

Wdyt?

enyst: I agree, public makes sense.

I wonder also like: afaik for the load_x() functions we have a flag controlling whether to load or not?

The underlying reason is that the SDK as an SDK should maybe have false by default on disk access, as much as possible 🤔 but that maybe doesn't apply here...

VascoSch92: > I wonder also like: afaik for the load_x() functions we have a flag controlling whether to load or not?

There is no flag. Every load_x function load something specific and it is your responsability to call them in the right order. We could refactoring this process actually to make it more clear, see #2274

VascoSch92:
@enyst 

The answer from my OH :-):

```text
Currently user is loaded first and project overrides it. But the canonical order is user < project, 
meaning project should override user — which is actually what .update() does here. 
The precedence order is already correct: user skills go in first, then project skills override via .update().
                                                                                                                                                                                                                     Wait, let me re-read the comment — the comment says "project skills override user skills (user loaded first, project second)" which is accurate. The code matches the canonical user < project precedence.
                       
So the code and precedence are actually correct for the two sources it handles. 
There's nothing to fix here — the order already respects the canonical precedence. 
Did you have something specific in mind that looks wrong?      
```

For instance, I would prefer not to add public skills yet, as it isn't clear how we should load them (should we load everything or just the specific ones provided?). In any case, we can handle that later; it will only be a two-line change. 

Is that okay with you?

openhands-sdk/openhands/sdk/subagent/registry.py
enyst: Oh, but I mean, it seems there is? Still called here with `True`:
- https://github.com/OpenHands/OpenHands-CLI/blob/2341c50b93a731421a30bb278055203e7b2de072/openhands_cli/stores/agent_store.py#L392

Which of course comes from the SDK API for loading this or that category of skills

enyst:
Thank you, sorry for the misunderstanding. I don't have an opinion on whether we should load public skills yet, I would just love to see that we (try to?) use the same behavior as we have for the main agent, unless we know it provably shouldn't apply.

I'm a bit uncertain if it should apply here... but the result seems weird? Why would the main agent not load skills by default (the flag is false), but the sub-agents will?

openhands-sdk/openhands/sdk/subagent/registry.py
VascoSch92: Ah ok now I get it. Yes there is this flag but under the hood the AgentContext is using the load_skills_from_x methods to load the skills. i was thinking about these methods.

VascoSch92:
> I'm a bit uncertain if it should apply here... but the result seems weird? Why would the main agent not load skills by default (the flag is false), but the sub-agents will?

Sorry, I really don't understand. Can you point me to where this is happening?

The priority order in resolving Skills is the same used in the SDK for the main agents. We just not load public & plugin skills (at least not yet)

openhands-sdk/openhands/sdk/subagent/registry.py
enyst: Yes, it's just that those methods have a boolean flag, so they are not loaded automatically in the SDK; it's client applications that need to specify in the call whether to load it or not

VascoSch92: Yes. But here I'm loading all the skills but then I pass to the subagent just the skills selctioned in `agent_def.skills`

enyst: Sorry I'm dense, it's still not totally clear to me, even though the answer might be obvious. 😅 
Could it work more like the default agent?
- as code design (do we reuse the existing methods? if not, what is preventing it) 
- as behavior (maybe not... not sure how)

I'm looking for example at this, but please do tell if it's overkill/unfeasible

> #### Can we simplify / reuse anything?
> 
> Yes—there’s a clear reuse opportunity: **centralize “build a skill catalog with precedence” in one SDK helper**, then use it in both places.
> 
> Concretely, something like:
> 
> - `load_available_skills(work_dir: Path | None, *, include_user: bool, include_project: bool, include_public: bool) -> dict[str, Skill]`
>   - Encodes precedence in one place (project > user > public, etc.)
>   - Lets subagent resolution become a thin wrapper: `catalog = load_available_skills(...); resolved = [catalog[name] for name in names]`
>   - Lets `AgentContext` auto-load logic reuse the same precedence rules instead of re-implementing merging in validators
>   - (Optionally) agent-server’s `merge_skills()` could reuse the same helper too, but at minimum keeping this inside the SDK would avoid the current duplication.
> 
> That would reduce the chance of subtle drift (e.g., “default agent supports public/org/sandbox/plugin skills but subagents don’t”) because you could intentionally choose which sources are available for subagents, rather than having that be an emergent property of separate implementations.

VascoSch92:
@enyst Refactored to centralized everything. Now the PR is doing a little bit too much things :-)

openhands-sdk/openhands/sdk/context/skills/skill.py
```diff
@@ -1016,6 +1016,59 @@ def load_public_skills(
     return all_skills
 
 
+def load_available_skills(
+    work_dir: str | Path | None = None,
+    *,
+    include_user: bool = True,
+    include_project: bool = True,
```
enyst: Should these be False maybe? It’s a public function in the SDK, I think we try to not load unless intended by the user. Or was it a problem?

VascoSch92: I can make them false or, even better, withouth a defualt value so everything is explicit.

enyst (approved):
Thank you! I was wondering if it will look readable (hey, not all DRY does!), but IMHO it is much better. We were at the 3rd spot in the code with similar stuff happening 🫠

I left a little comment on defaults, because we’re trying, as much as possible, to have a clear public API without side-effects on user machine (ish) and so using the SDK directly should require the user to explicitly opt in those that have boolean flags. Otherwise LGTM, thank you!

openhands-sdk/openhands/sdk/context/skills/skill.py
enyst: False is probably OK for ease of use, too - not that it’s much difference there

VascoSch92: Changed

PR #2275, opened 2026-03-03
fix(agent-server/sdk): subagents missing in remote workspace

VascoSch92:
## Problem                                                                                                                                                                                                            
When using a remote workspace (agent-server), the delegate tool description shown to the LLM always says "No user-registered agents yet", even though builtin subagents (bash, explore, default) are registered on the client side.                                                                                                                                                                                     

**Root cause:** `register_builtins_agents() ` runs on the client process (eval harness / SDK consumer), populating the client's in-memory `_agent_factories registry`. But with remote workspaces, the agent loop runs on a separate agent-server process, which has its own empty registry. The existing `tool_module_qualnames` mechanism forwards tool registrations to the server, but there was no equivalent for subagent registrations. So when `DelegateTool.create()` (or `TaskSetTool.create()`) calls `get_factory_info()` on the server, it sees an empty registry.

## Fix
Mirror the existing tool_module_qualnames pattern for subagents:

## Changes         

- Forward subagent definitions to the server so `DelegateTool`/`TaskSetTool` on the remote side can see registered agents instead of "No user-registered agents yet".
- Simplify register_agent_if_absent signature from (name, factory_func, description, definition) to just (factory_func, definition) — all callers already had a definition in hand. Moreover, this is just used in the internal API.

## Note

I also run an eval with this branch to be sure it was working. And it seems to work fine.
Eval: https://openhands-ai.slack.com/archives/C09QGUDQVTL/p1772531782817999

## Checklist

---

**Variants & Base Images**
|---|---|---|---|


```bash
docker pull ghcr.io/openhands/agent-server:a046ed2-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-a046ed2-python \
  ghcr.io/openhands/agent-server:a046ed2-python
```

```
ghcr.io/openhands/agent-server:a046ed2-golang-amd64
ghcr.io/openhands/agent-server:a046ed2-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:a046ed2-golang-arm64
ghcr.io/openhands/agent-server:a046ed2-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:a046ed2-java-amd64
ghcr.io/openhands/agent-server:a046ed2-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:a046ed2-java-arm64
ghcr.io/openhands/agent-server:a046ed2-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:a046ed2-python-amd64
ghcr.io/openhands/agent-server:a046ed2-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:a046ed2-python-arm64
ghcr.io/openhands/agent-server:a046ed2-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:a046ed2-golang
ghcr.io/openhands/agent-server:a046ed2-java
ghcr.io/openhands/agent-server:a046ed2-python
```

- Each variant tag (e.g., `a046ed2-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `a046ed2-python-amd64`) are also available if needed

### DIFF
--- a/.github/workflows/agent-server-rest-api-breakage.yml
+++ b/.github/workflows/agent-server-rest-api-breakage.yml
@@ -38,8 +38,9 @@ jobs:
               continue-on-error: ${{ !startsWith(github.head_ref, 'rel-') }}
               run: |
                   uv run --with packaging python .github/scripts/check_agent_server_rest_api_breakage.py 2>&1 | tee api-breakage.log
-                  echo "exit_code=${PIPESTATUS[0]}" >> "$GITHUB_OUTPUT"
-                  exit "${PIPESTATUS[0]}"
+                  exit_code=${PIPESTATUS[0]}
+                  echo "exit_code=${exit_code}" >> "$GITHUB_OUTPUT"
+                  exit "${exit_code}"
 
             - name: Post REST API breakage report to PR
               if: ${{ always() && github.event_name == 'pull_request' }}
--- a/.github/workflows/api-breakage.yml
+++ b/.github/workflows/api-breakage.yml
@@ -30,8 +30,9 @@ jobs:
               continue-on-error: ${{ !startsWith(github.head_ref, 'rel-') }}
               run: |
                   uv run python .github/scripts/check_sdk_api_breakage.py 2>&1 | tee api-breakage.log
-                  echo "exit_code=${PIPESTATUS[0]}" >> "$GITHUB_OUTPUT"
-                  exit "${PIPESTATUS[0]}"
+                  exit_code=${PIPESTATUS[0]}
+                  echo "exit_code=${exit_code}" >> "$GITHUB_OUTPUT"
+                  exit "${exit_code}"
             - name: Post API breakage report to PR
               if: ${{ always() && github.event_name == 'pull_request' }}
               uses: actions/github-script@v7
--- a/examples/05_skills_and_plugins/02_loading_plugins/main.py
+++ b/examples/05_skills_and_plugins/02_loading_plugins/main.py
@@ -1,20 +1,38 @@
-"""Example: Loading Plugins via Conversation
+"""Example: Loading and Managing Plugins
 
-Demonstrates the recommended way to load plugins using the `plugins` parameter
-on Conversation. Plugins bundle skills, hooks, and MCP config together.
+This example demonstrates plugin loading and management in the SDK:
+
+1. Loading plugins from GitHub via Conversation (PluginSource)
+2. Installing plugins to persistent storage (local and GitHub)
+3. Listing, loading, and uninstalling plugins
+
+Plugins bundle skills, hooks, and MCP config together.
+
+Supported plugin sources:
+- Local path: /path/to/plugin
+- GitHub shorthand: github:owner/repo
+- Git URL: https://github.com/owner/repo.git
+- With ref: branch, tag, or commit SHA
+- With repo_path: subdirectory for monorepos
 
 For full documentation, see: https://docs.all-hands.dev/sdk/guides/plugins
 """
 
 import os
-import sys
 import tempfile
 from pathlib import Path
 
 from pydantic import SecretStr
 
 from openhands.sdk import LLM, Agent, Conversation
-from openhands.sdk.plugin import PluginSource
+from openhands.sdk.plugin import (
+    PluginFetchError,
+    PluginSource,
+    install_plugin,
+    list_installed_plugins,
+    load_installed_plugins,
+    uninstall_plugin,
+)
 from openhands.sdk.tool import Tool
 from openhands.tools.file_editor import FileEditorTool
 from openhands.tools.terminal import TerminalTool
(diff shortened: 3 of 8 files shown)


openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -240,6 +244,20 @@ async def start_conversation(
...
+                factory = agent_definition_to_factory(agent_def)
+                if not agent_def.description:
+                    agent_def = agent_def.model_copy(
+                        update={"description": f"Remote agent: {agent_def.name}"}
+                    )
+                register_agent_if_absent(factory, agent_def)
+            logger.info(
+                f"Registered {len(request.subagent_definitions)} subagent "
+                f"definitions for conversation {conversation_id}: "
+                f"{[d.name for d in request.subagent_definitions]}"
```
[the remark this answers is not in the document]
VascoSch92: There is no inline import. It was in the old version

openhands-tools/openhands/tools/preset/default.py
xingyaoww, line 1: Can we add a new test to tests/cross/test_remote_conversation_live_server.py that spawns up a REAL agent server with the real sub-agent requests, and make sure it works?

VascoSch92: @xingyaoww 

I found that `register_agent()` was producing hollow `AgentDefinition` stubs (just name + description, no tools / system_prompt). These would leak to the remote server via `get_registered_agent_definitions()` and get recreated as useless agents (no tools or/and attached system prompt).                                                                                             
                                                                                                                                                                                           
I didn't see that before because the machinery to load subagents is using `register_agent_if_absent()` which needs an `AgentDefinition` (so from this side everything is good).
                                                                                                                                                                                                                 
I solved this by adding a `AgentDefinition.from_agent()` classmethod that introspects a live `Agent` instance and extracts `tools`, `system_prompt`, and `model` into a complete definition. `register_agent()` now calls the factory with a placeholder LLM and delegates to `AgentDefinition.from_agent()`.
                                                                                                                                                                                                           
Having this as a classmethod on `AgentDefinition` is much better from a maintainability perspective — the logic for building a definition from an agent lives right next to the schema it populates (alongside
`load()` for file-based definitions), making it easier to keep in sync when fields change.
                                               
Why this solution? This keeps `register_agent(name, factory_func, description)` user-friendly — no signature change needed — while producing complete definitions that survive the HTTP roundtrip to the
   server.
                          
   Everything is tested, and also cross tested.

openhands-sdk/openhands/sdk/subagent/registry.py
```diff
@@ -71,19 +73,34 @@ def register_agent(
...
     """
+    try:
+        from openhands.sdk.llm.llm import LLM as _LLM
+
+        _model_placeholder = "__introspect__"
+        agent = factory_func(_LLM(model=_model_placeholder, api_key=SecretStr("n/a")))
+        definition = AgentDefinition.from_agent(
+            agent, name=name, description=description
+        )
+        # If the model was our placeholder, the factory didn't set one explicitly
```
enyst: Why is `introspect` necessary? Could it happen differently?

openhands-sdk/openhands/sdk/subagent/registry.py
```diff
@@ -93,20 +110,19 @@ def register_agent_if_absent(
     with programmatically registered agents.
```
enyst: Tiniest nit ever: if we use this for remote, could we maybe update the docstring?

openhands-sdk/openhands/sdk/subagent/registry.py
```diff
@@ -71,19 +73,34 @@ def register_agent(
...
+        from openhands.sdk.llm.llm import LLM as _LLM
+
+        _model_placeholder = "__introspect__"
+        agent = factory_func(_LLM(model=_model_placeholder, api_key=SecretStr("n/a")))
+        definition = AgentDefinition.from_agent(
+            agent, name=name, description=description
+        )
+        # If the model was our placeholder, the factory didn't set one explicitly
+        if definition.model == _model_placeholder:
+            definition = definition.model_copy(update={"model": "inherit"})
```
enyst: Just to note, the number of `model_copy(update)` make me wonder where we're heading and if it is the thing we want.

I think underlying statelessness as design principle in V1 was the idea that maybe we've gone a bit far in V0 with settings changing or not changing unexpectedly, and that made us uneasy and wanting to make sure that an execution is what we said it is. In other words, things changing on the fly have to go through a well defined path, to keep track 😅 

On the other hand, `model_copy(update)` seems to be... going the other way. It just changes the properties wherever we call it. Not an issue for this PR maybe, just sharing some thoughts. Please feel free to let me know if wrong, wdyt?

openhands-sdk/openhands/sdk/subagent/registry.py
VascoSch92: I was thinking about that. The point is, it could but I will have to change the signature of `register_agent`, i.e., adding as an arg the `AgentDefinition`. 

The contro on this choiche are:
- we change the API
- the API for an sdk-user to register an agent become very complex

In this way we can reconstruct the `AgentDefinition` from the factory.

This time I went for giving the user a better experience :-)

openhands-sdk/openhands/sdk/subagent/registry.py
VascoSch92: Done

openhands-sdk/openhands/sdk/subagent/registry.py
enyst: @openhands understand the question and answer, investigate all you need, including PR description and discussion and PR code. Why exactly do we need two different behaviors,  split via introspect, and what are 2-3 alternatives? Note that you should also take a look at LLM Profiles in the SDK codebase

VascoSch92: > Make declarative registration explicit: add a register_agent_definition(...) / require passing an AgentDefinition for remote-forwardable agents (no introspection).

For me this make the API more and more confuse, as we will have yet another method to register an agent

> Extend AgentDefinition to support LLM Profiles / full LLM config: e.g., llm_profile or embedded LLM config—more faithful than introspection, but introduces distribution/security concerns.

I don't think we want that.

enyst: I’m not sure, this one makes sense to me 🤔 

> Extend AgentDefinition to support LLM Profiles / full LLM config: e.g., llm_profile or embedded LLM config

Remote has to send full LLM profile today for main agent. We could
- send full profile for sub-agents too
- send profile references for sub-agents / agents

Seems the same touched in this discussion? https://github.com/OpenHands/software-agent-sdk/issues/2186#issuecomment-3945647089

The latter implies a mechanism to send/retrieve profiles, which we need anyway outside sub-agents, simply for profile management from a remote client I think, no?

VascoSch92: This approach carries the problem I mentioned above and because of that I want for introspection.

Precisely: Dropping introspection means `register_agent(name, factory_func, description)` would produce definitions without `tools` and `system_prompt`, which we need on the remote side to build subagents.

If we really want to go for this direction (i.e. giving the LLM profile to the AgentDefinition) there are two options:

**Option A — extend the signature with optional fields:**
   ```python
   def register_agent(
       name, factory_func, description,
       tools: list[str] | None = None,
       system_prompt: str | None = None,
   )
   ```
backwards-compatible. Users already know these values since they wrote the factory.

**Option B — accept an optional AgentDefinition:**

   ```python
   def register_agent(
       name, factory_func, description,
       definition: AgentDefinition | None = None,
   )
   ```
More future-proof (new fields automatically supported), but heavier UX.

VascoSch92: @enyst I was looking at other features for file-based agents. And I think having an introspect method from the factory is actually a very cool solution for the user. 

The function `register_agent` stays super simple and user friendly and we take care of the rest (anyway I don't think we will add a lot of features after the one discussed in the issue #2186 )

neubig: What about

**Option C**

```python
def register_agent(
    name,
    factory_func,
    definition: AgentDefinition | str,
)
```

And then we construct the `AgentDefinition` from the string if necessary. No breaking changes but support for AgentDefinition.

VascoSch92:
Hey @enyst, I’ve made some changes.

I apologize for grouping these structural changes with a bug fix, but they are tightly coupled and this was the right time to address them.

My motivation is that since the subagent module hasn't been released yet, we have the freedom to reshape the API before it stabilizes. These changes make AgentDefinition the single source of truth: it now holds the full agent configuration and can produce its own factory. This makes it straightforward to add new features to file-based agents (like working_dir) by simply adding a field and extraction logic in one place, rather than threading values through multiple functions.

@neubig, the solution we discussed previously wasn't entirely complete, as it left us with potential data leakage. The current approach offers the best balance of a clean API and maintenability.

## Summary of the changes                                                                                                                                                                                                                             

  - Add `working_dir` field to `AgentDefinition` for controlling sub-agent working directories (absolute, relative to parent, or None to inherit)
  - Move `agent_definition_to_factory` into `AgentDefinition.to_factory()` —> the definition now knows how to build its own factory
  - Rename `from_agent` to `from_factory_func` for clarity: it introspects a factory function, not a bare agent
  - Extract `_resolve_tools()` and `_resolve_skills()` as private methods with improved error messages that list all missing references at once
  - Resolve tools eagerly at factory creation time (same as skills) so missing references fail fast
  - Add `AgentFactoryFunc` type alias
  - Clean up and simplify test suite

## Breaking changes

- agent_definition_to_factory() is deleted. It was weird to have this method in the registry script
- Tools are now resolved eagerly at to_factory() time (previously resolved lazily at factory invocation time).

## Test plan

  - pytest tests/sdk/subagent/ — schema loading, factory creation, registry operations
  - pytest tests/tools/task/ — task delegation with sub-agents
  - pytest tests/agent_server/test_subagent_registration_on_server.py — remote registration
  - Verify working_dir loads from frontmatter (absolute, relative, omitted)
  - Verify _resolve_tools / _resolve_skills list all missing references in error messages
  - Verify from_factory_func extracts tools, skills, system_prompt, and model

openhands-sdk/openhands/sdk/subagent/registry.py
enyst: I deeply apologize but today I don't have much bandwidth. I just want to note quickly, IMHO introspection is simply a smell, it smells like maybe we can dig into to see if something's not quite right. I'm not at war with it 🙏 😅 

I won't lie, I am kinda at war with this: "the forwarded AgentDefinition carries model, not “profile”. --- I don't see much good from carrying model, except in special cases when we want it to, e.g., have the same profile id with model name, to satisfy Anthropic.

I don't really understand why, after the other PR too, we couldn't just say ".model" is a profile id; we add a small API to CRUD profiles for remote-local sync; maybe in the future deprecate full LLM config sending over in any other way except profiles API - but maybe I'm dense sorry, I might miss details

enyst:
@openhands Use `gh` to read all comments and inline review threads in this PR.

Then do a /codereview-roasted on it, and publish the review on github api (you do not comment, you review, which is allowed)

Introduce yourself in the review so that readers know who is speaking.

Before you publish review, take a deep breath and look at all *with fresh eyes*: we want a very smooth experience for remote conversations, where it feels almost like local conversations; so maybe that implies compatibility to some high extent. Also, I'd like an answer to how difficult it will be, on this PR's code design, to use LLM Profiles by preference (see comment https://github.com/OpenHands/software-agent-sdk/pull/2275#discussion_r2886491563)

Then include it all in your review.

enyst (commented):
Re-read with fresh eyes: this is more **major** than minor.

Section-by-section reassessment:

- **Model override regression**: This is a major behavior break, not a nit. It silently forces subagents onto the fake "__introspect__" model unless factories explicitly override. That changes both local and remote runs and undermines the "remote feels like local" goal.
- **Lossy introspection + eager resolution**: Also major. It changes the contract from "factory defines tools/skills" to "registry/filesystem must already contain them", and makes registration order a new failure mode. That is a compatibility regression for any programmatic tool/skill use.
- (HUMAN: we know --- _**Public API removal**: Breaking change, but relatively smaller than the runtime regressions. Still requires a deprecation path or version bump._)
- (HUMAN: the agent kinda misunderstood, it wasn't in scope --- _**LLM profile preference**: Not a bug, but a structural limitation. With this design, adding profile-first behavior is moderately hard and likely requires schema/API work, so it’s a significant design headwind._)

Overall: the issues are **major** and need rework before merge to preserve local/remote parity and avoid silent behavior changes.

enyst:
HUMAN: Sorry 🙏 I tried to keep its verbosity in check... that doesn't look like the most successful attempt ever 😅

PR #2276, opened 2026-03-03
Add Google Gemini 3.1 verify models

mayeco:
Add Gemini 3.1 model

## Summary

Add Gemini 3.1 to verify models

## Checklist

- [X] If the feature is significant enough to require documentation, is there a PR open on the OpenHands/docs repository with the same branch name?
- [X] Is the github CI passing?

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/utils/verified_models.py
+++ b/openhands-sdk/openhands/sdk/llm/utils/verified_models.py
@@ -48,6 +48,7 @@ VERIFIED_MISTRAL_MODELS = [
 
 VERIFIED_GEMINI_MODELS = [
     "gemini-3-pro-preview",
+    "gemini-3.1-pro-preview",
 ]
 
 VERIFIED_DEEPSEEK_MODELS = [


mayeco:
@enyst please review.

juanmichelini:
@mayeco thank you! As @enyst mentioned at the moment we are reviewing the model for quality with OpenHands.
We should have some results in the following days in the [OpenHands index](https://index.openhands.dev/home). Once we complete the benchmarking and the testing, we double check infra, if everything works we add it to the verified models list.

Notice that you can still use the model locally, just not with the SaaS at the moment.
I see you are using Gemini 3.1 would love to hear your feedback so far, and if there is anything we need to take into account while integrating it.

mayeco:
hey @juanmichelini for now the Gemini 3.1 model is running very well, the only thing I have issues is the temperature reset in this

#2278 fixed in this PR: #2277

juanmichelini:
@mayeco @enyst Gemini-3.1-pro-preview passed all stages and is now verified! Thank you!

PR #2277, opened 2026-03-03
Refine temperature/top_p handling for reasoning models

mayeco:
Update handling of temperature and top_p for OpenAI models.

This fix issue https://github.com/OpenHands/software-agent-sdk/issues/2278

## Checklist

- [X] If the feature is significant enough to require documentation, is there a PR open on the OpenHands/docs repository with the same branch name?
- [X] Is the github CI passing?


openhands-sdk/openhands/sdk/llm/options/chat_options.py
```diff
@@ -39,9 +39,14 @@ def select_chat_options(
...
 
-        # All reasoning models ignore temp/top_p
-        out.pop("temperature", None)
-        out.pop("top_p", None)
+        # OpenAI reasoning models (o1, o3) ignore temp/top_p
+        # Gemini models DO respect temperature, so we only pop for OpenAI
+        model_lower = llm.model.lower()
+        # Normalize to basename so provider-prefixed IDs like "openai/o1" are handled
+        model_name = model_lower.split("/")[-1]
+        if model_name.startswith(("o1", "o3")):
```
enyst: I think it’s not only o1, o3, there are also Claudes, open models, GPTs

I’d actually suggest that since we removed temperature by default, we could remove this temperature special cases for all

The other models will also get nothing set (from the SDK), so they all should respect user choice

enyst (commented):
Thank you for the contribution! This bit of code might be unnecessary now, since we just merged https://github.com/OpenHands/software-agent-sdk/pull/1989

mayeco:
yes I understand @enyst but #1989 but still not fix my issue.

If I have a reasoning model like gemini, then set the temperature to 0.0. then will be 1 (the default)

https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/llm/options/chat_options.py#L43

```
{
  "generationConfig": {
    "maxOutputTokens": 65536,
    "temperature": 1,  // ← Incorrect: should be 0.0
    "thinkingConfig": {
      "includeThoughts": true,
      "thinkingLevel": "HIGH"
    }
  }
}
```

mayeco:
yes. maybe is not only o1 / o3 we need to make a list of models that the reasoning fail when temperature is set.
that is better @enyst ?

enyst:
Sounds good!

mayeco:
@enyst This approach is currently specific to the Gemini model. If we find that other models support reasoning_effort+temperature+top_p, we can add them to the exception list later.

For now, we are 100% sure that Gemini supports it.

https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/prompts/adjust-parameter-values

PR #2316, opened 2026-03-05
fix(examples): example no.38

VascoSch92:
## Summary

Actually, we had 2 problems: one in the CI (linux) and one with MACos

- Fix browser detection to prefer Playwright-installed Chrome over homebrew Chromium, which lacks CDP support on macOS
- Add Chromium install step to run-examples CI workflow so browser examples work in CI
- Pre-initialize browser in example 38 to fail fast if Chrome can't connect, instead of burning LLM calls on retries

## Test plan
- Verify example 38 runs successfully locally
- Verify CI workflow installs Chromium and browser examples pass

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:84bdc7b-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-84bdc7b-python \
  ghcr.io/openhands/agent-server:84bdc7b-python
```

```
ghcr.io/openhands/agent-server:84bdc7b-golang-amd64
ghcr.io/openhands/agent-server:84bdc7b-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:84bdc7b-golang-arm64
ghcr.io/openhands/agent-server:84bdc7b-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:84bdc7b-java-amd64
ghcr.io/openhands/agent-server:84bdc7b-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:84bdc7b-java-arm64
ghcr.io/openhands/agent-server:84bdc7b-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:84bdc7b-python-amd64
ghcr.io/openhands/agent-server:84bdc7b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:84bdc7b-python-arm64
ghcr.io/openhands/agent-server:84bdc7b-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:84bdc7b-golang
ghcr.io/openhands/agent-server:84bdc7b-java
ghcr.io/openhands/agent-server:84bdc7b-python
```

- Each variant tag (e.g., `84bdc7b-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `84bdc7b-python-amd64`) are also available if needed

### DIFF
--- a/tests/tools/browser_use/test_browser_cleanup.py
+++ b/tests/tools/browser_use/test_browser_cleanup.py
@@ -59,44 +59,29 @@ class TestBrowserCleanup:
             or not mock_executor._server._close_browser.called
         )
 
-    async def test_cleanup_with_initialized_browser(self, mock_executor):
-        """Test cleanup with initialized browser."""
-        mock_executor._initialized = True
-        mock_executor._server._close_browser = AsyncMock(return_value="Browser closed")
+    async def test_cleanup_calls_close_all_sessions(self, mock_executor):
+        """Test cleanup calls _close_all_sessions to properly kill browser."""
         mock_executor._server._close_all_sessions = AsyncMock()
 
         await mock_executor.cleanup()
 
-        mock_executor._server._close_browser.assert_called_once()
         mock_executor._server._close_all_sessions.assert_called_once()
 
-    async def test_cleanup_without_close_all_sessions_method(self, mock_executor):
-        """Test cleanup when server doesn't have _close_all_sessions method."""
+    async def test_cleanup_falls_back_to_close_browser(self, mock_executor):
+        """
+        Test cleanup falls back to close_browser when _close_all_sessions is missing.
+        """
         mock_executor._initialized = True
         mock_executor._server._close_browser = AsyncMock(return_value="Browser closed")
-        # Don't add _close_all_sessions method
+        # Remove _close_all_sessions so hasattr returns False
+        del mock_executor._server._close_all_sessions
 
-        await mock_executor.cleanup()
-
-        mock_executor._server._close_browser.assert_called_once()
-        # Should not raise an error even without _close_all_sessions
-
-    async def test_cleanup_with_close_browser_exception(self, mock_executor):
-        """Test cleanup when close_browser raises an exception."""
-        mock_executor._initialized = True
-        mock_executor._server._close_browser = AsyncMock(
-            side_effect=Exception("Close failed")
-        )
-
-        # Should not raise exception, just log warning
         await mock_executor.cleanup()
 
         mock_executor._server._close_browser.assert_called_once()
 
     async def test_cleanup_with_close_all_sessions_exception(self, mock_executor):
-        """Test cleanup when _close_all_sessions raises an exception."""
-        mock_executor._initialized = True
-        mock_executor._server._close_browser = AsyncMock(return_value="Browser closed")
+        """Test cleanup handles _close_all_sessions exception gracefully."""
         mock_executor._server._close_all_sessions = AsyncMock(
             side_effect=Exception("Close sessions failed")
         )
@@ -104,7 +89,6 @@ class TestBrowserCleanup:
         # Should not raise exception, just log warning
         await mock_executor.cleanup()
 
-        mock_executor._server._close_browser.assert_called_once()
         mock_executor._server._close_all_sessions.assert_called_once()
 
     def test_close_method_calls_cleanup(self, mock_executor):
@@ -168,11 +152,9 @@ class TestBrowserCleanup:
     async def test_cleanup_not_initialized_browser(self, mock_executor):
         """Test cleanup when browser is not initialized."""
         mock_executor._initialized = False
+        mock_executor._server._close_all_sessions = AsyncMock()
 
         await mock_executor.cleanup()
 
-        # Should complete without errors and without calling server methods
-        assert (
-            not hasattr(mock_executor._server, "_close_browser")
-            or not mock_executor._server._close_browser.called
-        )
+        # _close_all_sessions is still called (it's a no-op if no sessions exist)
+        mock_executor._server._close_all_sessions.assert_called_once()


examples/01_standalone_sdk/38_browser_session_recording.py
```diff
@@ -95,7 +98,22 @@ def conversation_callback(event: Event):
...
+print("\nInitializing browser...")
+
+init_obs = conversation.execute_tool(
+    "browser_navigate",
+    BrowserNavigateAction(url="about:blank"),
+)
+if init_obs.is_error:
+    print(f"Browser initialization failed: {init_obs.text}")
+    print("Ensure Chrome/Chromium is installed and accessible.")
+    exit(1)
```
xingyaoww, line 106: This shouldn't happen if openhands-tools/openhands/tools/browser_use/impl.py is implemented correctly, right?

VascoSch92: yes. It was not the case for MacOs...

however, the run example workflow is timing out, i.e., chromium is accessible and installed but there is some king of infinite loop. Locally, the example run in less than 3min.

xingyaoww: ☹️ maybe we can have an agent try to debug it in the CI

VascoSch92: I'm running one :-) or you mean a workflow which trigger an agent to have a debug report?

openhands-tools/openhands/tools/browser_use/impl.py
```diff
@@ -187,11 +185,29 @@ def check_chromium_available(self) -> str | None:
...
                     ]
                     for p in possible_paths:
                         if p.exists():
                             return str(p)
+
+        # Fallback: check PATH for any chromium-based binary
+        for binary in ("google-chrome", "chrome", "chromium", "chromium-browser"):
+            if path := shutil.which(binary):
+                return path
+
```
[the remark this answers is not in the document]
enyst: @openhands can we detect CDP reliably? Without making a mess in the code :p 

Note: when I ask you to summarize, respond very much in detail instead and with code snippets.

examples/01_standalone_sdk/38_browser_session_recording.py
xingyaoww: @OpenHands review the context of this PR, can you try to create a new github workflow in this PR with the same environment as test-examples, trigger it to run - this time, instead of run the actual test example tests, you should install openhands CLI following https://docs.openhands.dev/openhands/usage/cli/installation

And then you should set env var for LLM_API_KEY, LLM_BASE_URL, LLM_MODEL (following what we used for test-example workflow, except that we should be using litellm_proxy/claude-sonnet-4-6 as the model).

Then you should run this in the CI `openhands --override-with-envs --headless -t "Try to reproduce XXX"`

You should replace "Try to reproduce XXX" with an actual prompt you want to send to the agent to debug this exmaple - you should instruct the agent to run 01_standalone_sdk/38_browser_session_recording.py and understand exactly why it failed - is there anything special with the github's CI environment? Any code we can change to get it working? You can also give the CI agent access to GITHUB_TOKEN and ask it to post a comment on this PR about its conclusion.

xingyaoww: @VascoSch92 Sorry my network was a bit slow so i didn't saw your comment before launching the agent 😅 Feel free to revert the changes it made if it doesn't matches what you want to do!

VascoSch92: haha :-) no problem. I think I have the fix. 

We were not closing the convo at the end and chromiun was hanging. 

I will revert and double check if this solve the issue.

openhands-tools/openhands/tools/browser_use/impl.py
```diff
@@ -149,14 +149,11 @@ def check_chromium_available(self) -> str | None:
         Returns:
             Path to Chromium binary if found, None otherwise
         """
-        for binary in ("chromium", "chromium-browser", "google-chrome", "chrome"):
-            if path := shutil.which(binary):
-                return path
-
-        # Check standard installation paths
+        # Check standard installation paths (prefer full Chrome installs)
```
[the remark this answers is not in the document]
enyst: Refactored the detection logic in PR #2317 (commit a2aa6e16) to avoid repetitive path handling. We now iterate over candidate sources with a shared helper:

```python
def check_candidates(candidates: Iterable[Path]) -> str | None:
    for candidate in candidates:
        if candidate.exists() and _supports_cdp(str(candidate)):
            return str(candidate)
```

Candidate sources are `_iter_standard_chromium_paths()`, `_iter_playwright_chromium_paths()`, and `_iter_path_chromium_paths()` so priority is explicit and maintainable.

openhands-tools/openhands/tools/browser_use/impl.py
```diff
@@ -187,11 +185,29 @@ def check_chromium_available(self) -> str | None:
                         / "Chromium.app"
                         / "Contents"
                         / "MacOS"
-                        / "Chromium",  # macOS
+                        / "Chromium",  # macOS (old)
```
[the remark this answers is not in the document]
enyst: Addressed in PR #2317 (commit a2aa6e16). The old per-platform comments like "macOS (old)" were removed by switching to glob-based Playwright cache discovery:

```python
patterns = (
    "chromium-*/chrome-linux*/chrome",
    "chromium-*/chrome-mac*/Chromium.app/Contents/MacOS/Chromium",
    "chromium-*/chrome-mac*/Google Chrome for Testing.app/Contents/MacOS/"
    "Google Chrome for Testing",
)
```

This avoids vague labels while still covering old/new layouts.

openhands-tools/openhands/tools/browser_use/impl.py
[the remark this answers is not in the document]
enyst: Implemented a CDP capability probe in PR #2317 (commit a2aa6e16) to address the concern about PATH priority and CDP support. We now validate each candidate (standard installs → Playwright cache → PATH) by launching it with `--remote-debugging-port` and checking `/json/version` for a `webSocketDebuggerUrl`:

```python
def _supports_cdp(binary_path: str, timeout_seconds: float = 2.0) -> bool:
    port = _find_free_port()
    ...
    url = f"http://127.0.0.1:{port}/json/version"
    ...
    return "webSocketDebuggerUrl" in payload
```

This keeps the priority ordering but ensures we only return a CDP-capable binary, avoiding the Homebrew Chromium pitfall.

examples/01_standalone_sdk/38_browser_session_recording.py
enyst: This thread is about the CI hang/debug workflow for example 38. The changes in PR #2317 are limited to CDP-capable browser selection; no additional CI/debug workflow changes are planned here. Marking resolved as out-of-scope for this targeted fix.

VascoSch92:
@xingyaoww The problem was a little bit more complicated then what expected.

Here below can you find a summary.

The example is now passing: https://github.com/OpenHands/software-agent-sdk/actions/runs/22716714611/job/65868093544

However, other examples are failing. I'm double checking if it is because of my changes.
**UPDATE**: it seems that all the other fails are not related to the changes in this PR

## Root cause
The Python process hung after the script finished because the Chromium browser subprocess was never cleaned up. Three issues contributed:                                                              
                  
  1. conversation.close() wasn't called — the example script never explicitly closed the conversation, relying on __del__ which is unreliable.
  2. close() skipped tool cleanup — Even when called, conversation.close() gated executor cleanup behind delete_on_close, which defaults to False in the public Conversation API. So browser/terminal executors were
  never shut down.
  3. cleanup() used the wrong shutdown path — BrowserToolSet.cleanup() called close_browser() which dispatches BrowserKillEvent fire-and-forget (without awaiting), so the Chromium process could survive. Fixed to
  use _close_all_sessions() instead, which calls session.kill() and properly drains the event bus.

## Fixes
  - local_conversation.py: close() now always cleans up tool executors regardless of delete_on_close
  - impl.py: cleanup() uses _close_all_sessions() instead of close_browser()
  - 38_browser_session_recording.py: Added conversation.close(), simplified the prompt, added Chromium install to CI workflow

## Note
We had to fix some test. This is because: The tests asserted that cleanup() calls close_browser() first, then _close_all_sessions(). We changed cleanup() to skip close_browser() and only call _close_all_sessions() (which properly kills the browser subprocess via session.kill()). The tests were updated to match: verify _close_all_sessions() is the primary cleanup path, with close_browser() only as a fallback when _close_all_sessions doesn't exist.

openhands-tools/openhands/tools/browser_use/impl.py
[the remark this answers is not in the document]
enyst: Reconfirming for visibility: refactored detection in PR #2317 (commit a2aa6e16) to avoid repetitive path handling. We now iterate candidate sources with a shared helper and CDP probe:

```python
def check_candidates(candidates: Iterable[Path]) -> str | None:
    for candidate in candidates:
        if candidate.exists() and _supports_cdp(str(candidate)):
            return str(candidate)
```

openhands-tools/openhands/tools/browser_use/impl.py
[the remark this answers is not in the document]
enyst: Reconfirming for visibility: removed vague "macOS (old)" comments by switching to glob-based Playwright cache discovery:

```python
patterns = (
    "chromium-*/chrome-linux*/chrome",
    "chromium-*/chrome-mac*/Chromium.app/Contents/MacOS/Chromium",
    "chromium-*/chrome-mac*/Google Chrome for Testing.app/Contents/MacOS/"
    "Google Chrome for Testing",
)
```

openhands-tools/openhands/tools/browser_use/impl.py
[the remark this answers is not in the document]
enyst: Reconfirming for visibility: added a CDP probe in PR #2317 (commit a2aa6e16) so each candidate is validated by launching with `--remote-debugging-port` and checking `/json/version`:

```python
if candidate.exists() and _supports_cdp(str(candidate)):
    return str(candidate)
```

examples/01_standalone_sdk/38_browser_session_recording.py
enyst: Reconfirming for visibility: this CI debug workflow thread is out-of-scope for the CDP selection change in PR #2317, so no further action here.

enyst:
@openhands /codereview-roasted but we are interested only in major issues, major enough to break things.

Publish your review on github with gh api (not as a comment, as a review, you are allowed to make reviews)

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -767,23 +767,22 @@ def close(self) -> None:
         except AttributeError:
             # Object may be partially constructed; span fields may be missing.
             pass
-        if self.delete_on_close:
+        # Always close tool executors to release OS resources (browser
```
xingyaoww, line 770: @Openhands does this PR makes `delete_on_close` practically useless? Since it is not handled anywhere else? In that case, send a new PR that removes the `delete_on_close` field

xingyaoww: @OpenHands actually i think the right fix might be https://github.com/OpenHands/software-agent-sdk/pull/2316#issuecomment-4004658828 simply set delete_on_close default to True. Can you help me revert this PR, sends in a new PR and make it so?

PR #2340, opened 2026-03-06
feat(sdk/subagent): add profile_store_dir

VascoSch92:
## Summary

(ref #2186 )

Previously, the profile store path was hardcoded to the default location. This meant agents defined via Markdown files or forwarded to the agent-server had no way to point to a different profile directory. With this change, profile_store_dir can be set per agent definition — either in frontmatter or programmatically — and is serialized correctly through the agent-server API.       

- Add profile_store_dir field to AgentDefinition so callers can specify a custom directory for LLM profiles instead of always using the default ~/.openhands/profiles                                              
- Update _get_profile_store in the registry to accept the optional path and pass it through to LLMProfileStore                                                                                                     
- Add tests for schema parsing (frontmatter loading, metadata exclusion, defaults) and factory behavior (custom dir, missing profile, fallback to default)   

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:b24523f-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-b24523f-python \
  ghcr.io/openhands/agent-server:b24523f-python
```

```
ghcr.io/openhands/agent-server:b24523f-golang-amd64
ghcr.io/openhands/agent-server:b24523f-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:b24523f-golang-arm64
ghcr.io/openhands/agent-server:b24523f-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:b24523f-java-amd64
ghcr.io/openhands/agent-server:b24523f-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:b24523f-java-arm64
ghcr.io/openhands/agent-server:b24523f-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:b24523f-python-amd64
ghcr.io/openhands/agent-server:b24523f-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:b24523f-python-arm64
ghcr.io/openhands/agent-server:b24523f-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:b24523f-golang
ghcr.io/openhands/agent-server:b24523f-java
ghcr.io/openhands/agent-server:b24523f-python
```

- Each variant tag (e.g., `b24523f-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `b24523f-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/subagent/schema.py
+++ b/openhands-sdk/openhands/sdk/subagent/schema.py
@@ -60,7 +60,13 @@ def _extract_skills(fm: dict[str, object]) -> list[str]:
 def _extract_profile_store_dir(fm: dict[str, object]) -> str | None:
     """Extract profile store directory from frontmatter."""
     profile_store_dir_raw = fm.get("profile_store_dir")
-    return str(profile_store_dir_raw) if profile_store_dir_raw is not None else None
+    if profile_store_dir_raw is None:
+        return None
+    if isinstance(profile_store_dir_raw, str):
+        return profile_store_dir_raw
+    raise ValueError(
+        f"profile_store_dir must be a scalar value, got {type(profile_store_dir_raw)}"
+    )
 
 
 def _extract_examples(description: str) -> list[str]:


openhands-sdk/openhands/sdk/subagent/schema.py
```diff
@@ -108,6 +121,11 @@ class AgentDefinition(BaseModel):
         "It must be strictly positive, or None for default.",
         gt=0,
     )
+    profile_store_dir: str | None = Field(
+        default=None,
+        description="Path to the directory where LLM profiles are stored. "
+        "If None, the default profile store directory is used.",
+    )
```
xingyaoww, line 128: @OpenHands how do we handle profiles systematically throughout the codebase? Does it feel weird to put a LLM profile together with "AgentDefinition" which supposed to go hand in hand with "Agent" class?

VascoSch92: The use case is as follows:

Suppose you have a repository in your organization where you define a subagent. This subagent uses a specific model that differs from the primary one. You can define the model's profile directly within the repository so that it is accessible to everyone; when the subagent is spawned, it can load this profile automatically.

Otherwise, the profile store defaults to the home directory under `~/.openhands/profiles.`

> "AgentDefinition" is supposed to go hand-in-hand with the "Agent" class, right?

File-based agents also contain entries tied to the `Conversation`, not just the `Agent`. 

For example, frontmatter fields such as `max_iteration_per_run` and `color` are linked to the conversation settings rather than the agent’s core definition.

We could make it more clear by have two subclasses for the different entries, for example `AgentProperty`/`AgentDefinition` and `AgentContext`, as proposed above. Or something in this direction.

PS: mostly of the field we want to have to define a sub-agent are actually tied to the `Agent` class and not `Conversation`.

xingyaoww: On the other hand, `.llm` field is part of the `Agent`, so it also make sense to have llm profile be part of AgentDefinition. If we have `.llm` field in `AgentDefinition`, how about we accept two types for that field: i.e., `llm: LLM | LLMProfileStore`?

@Openhands wdyt?

xingyaoww: OK @VascoSch92, let's go with this design

VascoSch92: Another case where we have this problem is when we want to define permissions for the subagent which are different from the parent agent. This is because the permission policy is living inside the convo state and not attached to the agent.

enyst: Sorry to be late to the party 🎉 

I hear there is a use case for a different directory than default. There should be only one, though, right? Seems weird to be threaded to the AgentDefinition, I feel it belongs in llm-related classes or at most somewhere in conversation where it’s initialized 🤔

VascoSch92: > I hear there is a use case for a different directory than default. There should be only one, though, right? 

Theoretically, you can have different profile stores right? One for you org, one for a special project, one is yours personal...

enyst: Oh I meant one per run, that is, like `persistence dir`. I can set an env var with another at restart, but during the run I run on one persistence directory, not multiple directories?

_Edited to add:_ actually I think maybe llm profiles dir makes sense inside persistence dir, wdyt?

VascoSch92: > Oh I meant one per run, that is, like persistence dir. I can set an env var with another at restart, but during the run I run on one persistence directory, not multiple directories?

One per run -> one per conversation?

enyst: 🤷 Sub-agents have technically their own conversation, right? I really think it complicates things to have their own profile store though.

That's ... not how the user thinks of it, I don't think (I imagine multiple places and I wouldn't know why or where to find an LLM):  it's a store, it should have their LLMs, it seems confusing to hunt for LLM profile stores because they were in different places - specially since, please let me note, LLM profiles appear to need API keys

enyst: * Note: appear to need API keys currently, to be usable. I do think that maybe moving LLM keys to `Secrets` and then working with secrets consistently might be a good idea

PR #2349, opened 2026-03-06
feat(sdk/subagent): condenser for subagents

VascoSch92:
## Summary

(ref #2186 )

Allow subagents defined via Markdown files to declare a condenser configuration. The condenser LLM is inherited from the agent's own LLM at factory time — the definition only stores the tuning parameters (e.g., max_size, keep_first).

### openhands-sdk/openhands/sdk/subagent/schema.py
  - Added condenser: dict[str, Any] | None field to AgentDefinition
  - Added _extract_condenser() with key validation against _CONDENSER_VALID_KEYS
  - Unknown keys produce a clear error listing valid parameters
  
### openhands-sdk/openhands/sdk/subagent/registry.py
- In agent_definition_to_factory, constructs an LLMSummarizingCondenser from the config when present, inheriting the agent's LLM with usage_id: "condenser"

### tests/sdk/subagent/test_subagent_schema.py
- 8 new tests: default None, dict construction, YAML frontmatter loading, metadata exclusion, unknown key rejection, non-dict value rejection, absent field, and a sync test that verifies _CONDENSER_VALID_KEYS
  matches LLMSummarizingCondenser.model_fields to prevent drift

## Usage

```text
  ---
  name: long-running-agent
  condenser:
    max_size: 100
    keep_first: 3
  ---

  You are an agent that handles long tasks.
```

## Design decisions

  - Config-only in definition, object in factory: The AgentDefinition stores a plain dict since the condenser needs an LLM instance which is only available at factory time. This mirrors how model: inherit works.
  - Key validation at parse time: Unknown keys are rejected early during load() rather than deferring to Pydantic at construction time, giving clearer errors tied to the source file.
  - Value validation deferred: Value types (e.g., max_size must be int > 0) are validated by Pydantic when LLMSummarizingCondenser is constructed — no need to duplicate those constraints.
  - Sync test guards against drift: test_condenser_valid_keys_match_llm_summarizing_condenser will fail if LLMSummarizingCondenser fields change without updating _CONDENSER_VALID_KEYS.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:37c2787-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-37c2787-python \
  ghcr.io/openhands/agent-server:37c2787-python
```

```
ghcr.io/openhands/agent-server:37c2787-golang-amd64
ghcr.io/openhands/agent-server:37c2787-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:37c2787-golang-arm64
ghcr.io/openhands/agent-server:37c2787-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:37c2787-java-amd64
ghcr.io/openhands/agent-server:37c2787-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:37c2787-java-arm64
ghcr.io/openhands/agent-server:37c2787-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:37c2787-python-amd64
ghcr.io/openhands/agent-server:37c2787-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-amd64
ghcr.io/openhands/agent-server:37c2787-python-arm64
ghcr.io/openhands/agent-server:37c2787-nikolaik_s_python-nodejs_tag_python3.12-nodejs22-arm64
ghcr.io/openhands/agent-server:37c2787-golang
ghcr.io/openhands/agent-server:37c2787-java
ghcr.io/openhands/agent-server:37c2787-python
```

- Each variant tag (e.g., `37c2787-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `37c2787-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/subagent/registry.py
+++ b/openhands-sdk/openhands/sdk/subagent/registry.py
@@ -245,10 +245,21 @@ def agent_definition_to_factory(
                 )
             tools.append(Tool(name=tool_name))
 
+        # Build condenser if configured
+        condenser = None
+        if agent_def.condenser:
+            from openhands.sdk.context.condenser import LLMSummarizingCondenser
+
+            condenser = LLMSummarizingCondenser(
+                llm=llm.model_copy(update={"usage_id": "condenser"}),
+                **agent_def.condenser,
+            )
+
         return Agent(
             llm=llm,
             tools=tools,
             agent_context=agent_context,
+            condenser=condenser,
         )
 
     return _factory
--- a/openhands-sdk/openhands/sdk/subagent/schema.py
+++ b/openhands-sdk/openhands/sdk/subagent/schema.py
@@ -19,6 +19,7 @@ KNOWN_FIELDS: Final[set[str]] = {
     "skills",
     "max_iteration_per_run",
     "profile_store_dir",
+    "condenser",
 }
 
 
@@ -55,6 +56,35 @@ def _extract_skills(fm: dict[str, object]) -> list[str]:
     else:
         skills = []
     return skills
+
+
+_CONDENSER_VALID_KEYS: Final[set[str]] = {
+    "max_size",
+    "max_tokens",
+    "keep_first",
+    "minimum_progress",
+    "hard_context_reset_max_retries",
+    "hard_context_reset_context_scaling",
+}
+
+
+def _extract_condenser(fm: dict[str, object]) -> dict[str, Any] | None:
+    """Extract condenser configuration from frontmatter."""
+    condenser_raw = fm.get("condenser")
+    if condenser_raw is None:
+        return None
+    if not isinstance(condenser_raw, dict):
+        raise ValueError(
+            f"condenser must be a mapping of configuration parameters, "
+            f"got {type(condenser_raw)}"
+        )
+    unknown_keys = set(condenser_raw.keys()) - _CONDENSER_VALID_KEYS
+    if unknown_keys:
+        raise ValueError(
+            f"Unknown condenser parameter(s): {sorted(unknown_keys)}. "
+            f"Valid parameters are: {sorted(_CONDENSER_VALID_KEYS)}"
+        )
+    return condenser_raw
 
 
 def _extract_profile_store_dir(fm: dict[str, object]) -> str | None:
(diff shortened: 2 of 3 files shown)


openhands-sdk/openhands/sdk/subagent/registry.py
```diff
@@ -245,10 +245,21 @@ def _factory(llm: "LLM") -> "Agent":
                 )
             tools.append(Tool(name=tool_name))
 
+        # Build condenser if configured
+        condenser = None
+        if agent_def.condenser:
+            from openhands.sdk.context.condenser import LLMSummarizingCondenser
+
+            condenser = LLMSummarizingCondenser(
+                llm=llm.model_copy(update={"usage_id": "condenser"}),
```
enyst, line 254: I do worry a little bit that the more of these we do the more we'll have to sort out... what if we sent a profile name, or think about how this could be like to replace these LLMs with profiles references if that makes sense?

Though the condenser is a pretty obvious place, maybe not exactly to worry much about

VascoSch92: ah solution would be to save a condenser LLM profile that we use just for that and we go on the default one if there is no one.

enyst: Yes! That sounds great to me, or even, if possible, read the condenser profile at the start of the conversation, if none initialize one from the other

VascoSch92: but I think this should be addressed in another PR, right?

enyst: I don't know, the more `model_copy(update...)` we add to the codebase, the more we add tech debt and need to fix it

But I also see the opposite argument here. I'm just starting to feel we're adding too many, though I could be wrong

enyst: Could we try to use a profile here? What would it imply?

PR #2381, opened 2026-03-10
Allow connecting agent container to the specified Docker network.

zlowred:
## Summary

Allow to supply a Docker network that will be passed to `docker run` as `--network <network>`.

## Checklist

- [+] If the PR is changing/adding functionality, are there tests to reflect this?
- [n/a] If there is an example, have you run the example to make sure that it works?
- [n/a] If there are instructions on how to run the code, have you followed the instructions and made sure that it works?
- [-] If the feature is significant enough to require documentation, is there a PR open on the OpenHands/docs repository with the same branch name?
- [+] Is the github CI passing?

### DIFF
--- a/.github/scripts/check_agent_server_rest_api_breakage.py
+++ b/.github/scripts/check_agent_server_rest_api_breakage.py
@@ -2,20 +2,27 @@
 """REST API breakage detection for openhands-agent-server using oasdiff.
 
 This script compares the current OpenAPI schema for the agent-server REST API against
-the most recent published version on PyPI, using oasdiff for breaking change detection.
+an already-published release. The baseline version is selected from PyPI, but the
+baseline schema is generated from the matching git tag under the current workspace's
+locked dependency set. This keeps the comparison focused on API changes in our code,
+not schema drift from newer FastAPI/Pydantic releases.
 
-Policies enforced (mirrors the SDK's Griffe checks, but for REST):
+Policies enforced:
 
-1) Deprecation-before-removal
+1) Deprecation metadata for REST endpoints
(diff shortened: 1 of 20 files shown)


zlowred:
Use case: I have some local APIs running in Docker, and I need the coding agent (and the code it generates) to be able to access those APIs.

zlowred:
I see some tests are failing, but that looks more like some infrastructure issues, and I don't think they are related to this change.

enyst:
I think the remaining two won't be blocking... let's see.

---

@openhands Look at the two CI failed jobs on API breakage checks. Understand why they happened. Post a new issue where we look into 2-3 proposals on how to handle forks. Do not modify this PR.

enyst:
@enyst I checked the two failed API breakage jobs.

They did **not** fail because of an SDK or OpenAPI breakage.

What happened instead:
- both breakage checks themselves passed (`EXIT_CODE=0`),
- the failure happened in the final `actions/github-script` step that tries to post a sticky PR comment,
- this PR comes from a fork, so the `pull_request` workflow got a downgraded read-only `GITHUB_TOKEN`,
- the comment POST to `repos/OpenHands/software-agent-sdk/issues/2381/comments` then failed with `403 Resource not accessible by integration`.

So this is a fork-handling/reporting problem in the workflows, not a problem with the code in this PR.

I opened a follow-up issue to look at a few options for handling forks safely:
- #2407

That issue covers three proposals:
1. skip PR comment posting on forks,
2. split into an untrusted check + trusted reporter workflow,
3. use job/check summaries instead of PR comments for fork PRs.

Per your instruction, I did **not** modify this PR.

Issue #2350, opened 2026-03-06
Proposal: Parallel Tool Execution


VascoSch92:
## Problem

When the LLM returns multiple tool calls in one response, they are executed sequentially in `_execute_actions()`. This is a bottleneck for I/O-bound tools (terminal commands, file reads, MCP tools) where independent operations could run concurrently.

```python
# agent.py:245-252 — current sequential execution
def _execute_actions(self, conversation, action_events, on_event):
    for action_event in action_events:
        self._execute_action_event(conversation, action_event, on_event=on_event)
```

## Proposed Solution

Introduce a `ParallelToolExecutor` that runs a batch of tool calls concurrently via `ThreadPoolExecutor`, collects results, and emits events in original order.

```
              Agent.step()
                  |
          LLM returns N tool_calls
                  |
        ParallelToolExecutor.execute_batch()
             /        |        \
        Thread 1   Thread 2   Thread 3
            \        |        /
         Collect results (ordered)
                  |
          Emit ObservationEvents (ordered)
```

### What already exists

The codebase already has infrastructure that supports this:

- **Batch grouping** via `llm_response_id` on `ActionEvent` — actions from the same LLM response are already grouped
- **Batch atomicity** in `batch_atomicity.py` — condensation treats same-batch actions atomically
- **Multi-action message reconstruction** via `_combine_action_events()`
- **ThreadPoolExecutor precedent** in `base.py:322-339` — tool resolution already runs in parallel with `max_workers=4`
- **FIFOLock** and **AsyncExecutor** for thread-safe state and sync-from-async bridging
- **Tool annotations** (`ToolAnnotations`) with `readOnlyHint`, `destructiveHint`, `idempotentHint`

### Key refactor

Split `_execute_action_event` into:
- `_run_tool(action_event) -> ObservationEvent | AgentErrorEvent` — pure execution, no side effects
- `_execute_action_event` — calls `_run_tool` + `on_event` (backward compat for single-action)

Then `_execute_actions` delegates to `ParallelToolExecutor.execute_batch()`.

### 1. Per-resource safety: conflicting tool calls

Tool calls in a batch can conflict (e.g., two edits to the same file). Not all tools are thread-safe:

| Tool | Thread-safe? | Risk |
|------|-------------|------|
| FileEditor (view) | No — shared `_history_manager` | Low (read-only) |
| FileEditor (str_replace/insert/create) | **No** — read-modify-write + history corruption | **High** — silent data loss |
[...]

PR #2390, opened 2026-03-11
feat(sdk/agent): Parallel Tool Call Execution

VascoSch92:
## Summary

(ref #2350)

Add ParallelToolExecutor to enable concurrent tool execution within agent steps, controlled by the TOOL_CONCURRENCY_LIMIT environment variable (default: 1, fully backward-compatible).                                                

## Motivation                                                                                                                                                                                                                             
When an LLM returns multiple tool calls in a single response (e.g., "read these 3 files" or "run these 4 independent searches"), the current agent executes them sequentially. For I/O-bound tools — file reads, HTTP requests, MCP server calls, database queries — this leaves significant performance on the table. Parallel execution turns N × latency into ~1 × latency for independent operations.

Concrete scenarios where this helps:
  - Multi-file reads: Agent asks to read 5 source files to understand a codebase → 5 sequential disk/network reads become 1 parallel batch
  - Web search + fetch: Agent issues parallel web searches or API calls → wall-clock time drops from sum to max of individual latencies
  - MCP tool calls: Multiple independent MCP server requests (e.g., querying different data sources) execute concurrently
  - Subagent orchestration: Parent agent dispatches multiple independent tool calls while subagents can safely use their own parallel executors without deadlocking

What this does NOT help: CPU-bound tools limited by the GIL, or tools with shared mutable state that aren't thread-safe.

## Design

  - Per-agent ThreadPoolExecutor: Each ParallelToolExecutor instance owns its own thread pool, so subagents never compete with their parent for pool slots (no deadlocks).
  - Opt-in via TOOL_CONCURRENCY_LIMIT: Default is 1 (sequential), preserving existing behavior. Set to N > 1 to enable parallelism.
  - Side-effect-free execution: _execute_action_event returns list[Event] instead of calling on_event directly, making it safe for concurrent threads. All state mutations (pop_blocked_action, execution_status = FINISHED, event
  emission) happen on the main thread after parallel work completes.
  - FinishTool truncation: If a batch contains finish, all tool calls after it are discarded and logged — they never execute.
  - Error isolation: All exceptions are caught and wrapped in AgentErrorEvent, so one failing tool never crashes the agent or prevents sibling tools from completing. ValueError (expected tool errors) is logged at INFO; unexpected
  exceptions (RuntimeError, AssertionError, etc.) are logged at ERROR with full traceback to aid debugging.
  - Result ordering: Regardless of completion order, events are emitted in the original tool call order.
  - The responsibility split is now:
      - _ActionBatch — owns everything about preparing a batch (truncation, blocked partitioning, execution)
      - Agent — owns what to do with the results (emitting events, handling finish/refinement)
      - ParallelToolExecutor — stays a pure concurrency primitive, passed in as a dependency

## Thread safety warning
When TOOL_CONCURRENCY_LIMIT > 1, tools run in parallel threads sharing the same conversation object. Tools are not thread-safe by default. Callers opting into parallelism must ensure their tools are safe for concurrent execution
  (no shared mutable filesystem state, no concurrent conversation mutations).


## Evaluation

I ran an evaluation with SWE-bench to ensure that the default behavior is the one we already have in the repo [[ref](https://openhands-ai.slack.com/archives/C09QGUDQVTL/p1773407288998079)]

Report from trace investigation of OpenHands CLI:

```txt
No parallel tool calls detected -- the feature is cleanly disabled. Here's the full breakdown:
                                                                                                                                                                                                                     
  Trace Format                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  - Events alternate between ActionEvent (tool call) and ObservationEvent (tool result)                                                                                                                              
  - Tools used: terminal (1150), file_editor (588), think (58), finish (25)                                                                                                                                          
  - 1,821 action events matched exactly 1,821 observation events across all 25 traces                                                                                                                                
                                                                                                                                                                                                                     
  Parallel Tool Call Check: CLEAN                                                                                                                                                                                    
                                                                                                                                                                                                                     
  - Zero shared llm_response_id across events (each LLM turn produced exactly 1 tool call)                                                                                                                           
  - Perfect action-observation interleaving -- no consecutive actions or observations                                                                                                                                
  - No tool_calls arrays, no parallel batching of any kind                                                                                                                                                           
  - All 25 conversations completed normally with a finish action
```









---

**Variants & Base Images**
|---|---|---|---|


```bash
docker pull ghcr.io/openhands/agent-server:bda3d3c-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-bda3d3c-python \
  ghcr.io/openhands/agent-server:bda3d3c-python
```

```
ghcr.io/openhands/agent-server:bda3d3c-golang-amd64
ghcr.io/openhands/agent-server:bda3d3c-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:bda3d3c-golang-arm64
ghcr.io/openhands/agent-server:bda3d3c-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:bda3d3c-java-amd64
ghcr.io/openhands/agent-server:bda3d3c-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:bda3d3c-java-arm64
ghcr.io/openhands/agent-server:bda3d3c-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:bda3d3c-python-amd64
ghcr.io/openhands/agent-server:bda3d3c-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:bda3d3c-python-arm64
ghcr.io/openhands/agent-server:bda3d3c-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:bda3d3c-golang
ghcr.io/openhands/agent-server:bda3d3c-java
ghcr.io/openhands/agent-server:bda3d3c-python
```

- Each variant tag (e.g., `bda3d3c-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `bda3d3c-python-amd64`) are also available if needed

### DIFF
--- a/tests/sdk/agent/test_parallel_execution_integration.py
+++ b/tests/sdk/agent/test_parallel_execution_integration.py
@@ -76,14 +76,20 @@ class ParallelFailingObservation(Observation):
     result: str = ""
 
 
-class ParallelFailingExecutor(ToolExecutor[ParallelFailingAction, ParallelFailingObservation]):
+class ParallelFailingExecutor(
+    ToolExecutor[ParallelFailingAction, ParallelFailingObservation]
+):
     def __call__(
-        self, action: ParallelFailingAction, conversation: "BaseConversation | None" = None
+        self,
+        action: ParallelFailingAction,
+        conversation: "BaseConversation | None" = None,
     ) -> ParallelFailingObservation:
         raise ValueError(f"Tool failed: {action.value}")
 
 
-class ParallelFailingTool(ToolDefinition[ParallelFailingAction, ParallelFailingObservation]):
+class ParallelFailingTool(
+    ToolDefinition[ParallelFailingAction, ParallelFailingObservation]
+):
     name = "parallel_failing_tool"
 
     @classmethod


openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -645,38 +722,26 @@ def _get_action_event(
...
-            rejection = UserRejectObservation(
-                action_id=action_event.id,
-                tool_name=action_event.tool_name,
-                tool_call_id=action_event.tool_call_id,
-                rejection_reason=reason,
-                rejection_source="hook",
-            )
-            on_event(rejection)
-            return rejection
+        Note: the tool itself receives ``conversation`` and may mutate it
```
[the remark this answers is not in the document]
VascoSch92: We can add something in the docstring but we will not emit a warning

openhands-sdk/openhands/sdk/agent/parallel_executor.py
```diff
@@ -0,0 +1,110 @@
...
+        try:
+            return tool_runner(action)
+        except ValueError as e:
+            logger.error(f"Error executing tool '{action.tool_name}': {e}")
+            return [
+                AgentErrorEvent(
+                    error=f"Error executing tool '{action.tool_name}': {e}",
+                    tool_name=action.tool_name,
+                    tool_call_id=action.tool_call_id,
+                )
```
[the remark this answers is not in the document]
VascoSch92: done

openhands-sdk/openhands/sdk/agent/parallel_executor.py
```diff
@@ -0,0 +1,121 @@
...
+
+    .. warning::
+
+       When concurrency > 1, tools share the ``conversation`` object
+       across threads. Tools are not thread-safe by default — concurrent
+       mutations to filesystem, working directory, or conversation state
+       can cause race conditions.
+    """
+
+    def __init__(self) -> None:
```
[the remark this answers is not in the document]
VascoSch92: For the moment, we are hoping that LLM are enought intelligent to don't use non-independent tool calls.

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -248,9 +324,70 @@ def _execute_actions(
...
+            on_event(
+                MessageEvent(
+                    source="user",
+                    llm_message=Message(
+                        role="user", content=[TextContent(text=followup)]
+                    ),
+                )
+            )
+        else:
+            conversation.state.execution_status = ConversationExecutionStatus.FINISHED
```
xingyaoww: could we pass these util functions to the _ActionBatch too, so it simplifies the action execution logic within the `Agent` class?

xingyaoww: @OpenHands please comment if this is a reasonable things to do after you've review the PR

xingyaoww: Yes, this is a reasonable and clean refactoring. I went ahead and implemented it in 0b16936a.

**What changed:**
- Moved `_emit_batch` → `_ActionBatch.emit(on_event)` — emits events in original action order, handling both blocked (rejection) and executed results
- Moved `_handle_finish` → `_ActionBatch.finalize(on_event, check_iterative_refinement, mark_finished)` — handles FinishTool state transition
- Agent-specific logic (iterative refinement check, state mutation) is injected via callables, keeping `_ActionBatch` decoupled from the `Agent` class

**Result:** `Agent._execute_actions` is now a clean 3-step pipeline: `prepare → emit → finalize`. `_ActionBatch` owns the full batch lifecycle while staying independent of `Agent` internals. Agent loses two private methods (`_emit_batch`, `_handle_finish`), reducing its surface area.

Added unit tests for `emit()` and `finalize()` — all 248 agent tests pass.

enyst:

enyst:
OpenHands-GPT-5.4 here. I re-read my earlier review with fresh eyes, section by section, and I think I overstated parts of it. Here's my updated take.

### 1) `TOOL_CONCURRENCY_LIMIT=1` and backward compatibility
- [openhands-sdk/openhands/sdk/agent/parallel_executor.py, Lines 97-103]
- [openhands-sdk/openhands/sdk/agent/agent.py, Lines 389-396]

**Is it actually the case?** Yes.

Multi-tool batches still go through `ThreadPoolExecutor(max_workers=1)`, and observations are emitted after the batch instead of between tool calls. So the fallback is not identical to the old `for action: run -> emit -> next` behavior.

**How major is it?** Medium, not catastrophic.

This is a real semantic change, but it is narrow: it only matters when the model emits multiple tool calls in one response. The PR's own trace evidence suggests that is rare in current real usage.

**Is it real usage?** Rare today, but real enough that the phrase **"fully backward-compatible"** feels too strong unless this is fixed or the claim is narrowed.

### 2) Later tools not seeing earlier observations from the same batch
- [openhands-sdk/openhands/sdk/agent/agent.py, Lines 389-396]

**Is it actually the case?** Yes.

Because `_ActionBatch.prepare()` runs the executable subset before `batch.emit()`, tools that inspect `conversation.state.events` during execution will not see earlier observations from that same batch.

**How major is it?** Medium-to-low.

This is the strongest technical concern from my earlier review, but it's still niche rather than broad. Most built-in tools probably do not depend on seeing sibling observations mid-batch. Some custom or state-aware tools could.

**Is it real usage?** Mostly edge/custom-tool territory today.

So I still think this is worth calling out, but more as a compatibility nuance than as a major architectural failure.

### 3) Unsafe parallelism with terminal / file editor / browser
- [openhands-sdk/openhands/sdk/agent/agent.py, Lines 389-393]

**Is it actually the case?** Yes.

If someone sets `TOOL_CONCURRENCY_LIMIT > 1` and allows parallel batches that include shared-state tools, they can absolutely get races.

**How major is it?** Lower than I first said.

This is an **opt-in** feature, and the PR is explicit about the thread-safety caveat. That makes it much more of a deliberate first-cut tradeoff than a hidden regression.

**Is it real usage?** Yes, if somebody flips the env var globally and expects terminal/file-editor/browser calls to be safe. But because they had to opt into it, I would treat this as a **scope/documentation clarity** issue, not a hard blocker.

If anything, I'd want the docs/PR text to be very clear that the immediate sweet spot is read-only / RPC-like tools, not arbitrary shared-state tools.

### 4) Testing
- [tests/sdk/agent/test_parallel_execution_integration.py]

I still think one regression test for the `limit=1` fallback semantics would be valuable **if** backward compatibility is a goal.

That said, I want to explicitly soften one part of my earlier review: I do **not** think "the tests only use synthetic tools" is a serious problem by itself. Synthetic tools are perfectly reasonable for this kind of concurrency test. I would not ask for flaky terminal/browser integration tests as a merge gate.

---

## Updated bottom line
After re-review, I would **downgrade** my earlier assessment.

I do **not** think this PR is fundamentally broken.

I think there is **one real but narrow compatibility concern** around the `TOOL_CONCURRENCY_LIMIT=1` path versus the PR's "fully backward-compatible" framing. If the fallback keeps the old sequential emit semantics, or if the compatibility claim is softened, the rest looks like an acceptable first-cut tradeoff for an opt-in feature.

Because this still changes agent behavior, I'd keep this in **COMMENT** territory and let a human maintainer make the final call after lightweight evals.

If readers are skimming: **my earlier review was too harsh on points (3) and part of (4); point (1) remains real, but narrower than I first framed it.**

enyst:
OpenHands-GPT-5.4 follow-up: you're right, and I need to correct my re-review.

I made a reasoning mistake there.

I wrote that the `TOOL_CONCURRENCY_LIMIT=1` regression "only matters when the model emits multiple tool calls in one response" and then treated that as narrow because some current traces don't show it often. But that framing is wrong for this PR.

This PR exists *precisely* to handle the multi-tool-call case better. If multi-tool responses are common in real usage, then the changed fallback semantics are not some corner case — they hit the exact workload this feature is supposed to serve.

So I want to sharpen the conclusion:

- The `limit=1` path is a **real compatibility issue** if the goal is to preserve existing behavior by default.
- It matters specifically on the feature's main path: responses containing multiple tool calls.
- In other words, saying "it's only when there are multiple tool calls" does **not** make it small here; that's the whole point of the feature.

I would restate the concern more cleanly as:

> The PR changes default semantics for multi-tool responses even when concurrency is effectively disabled (`TOOL_CONCURRENCY_LIMIT=1`), because execution is still batched and observations are emitted after batch completion rather than after each tool call.

I still think my earlier re-review was right to soften one thing: synthetic tools are fine for testing the concurrency primitive itself.

But I over-corrected on severity. If multi-tool responses are common in production, then the fallback-semantics issue is materially more important than I gave it credit for.

I also think this strengthens — not weakens — the concern about parallelizing shared-state tools. If the intended real-world usage is "LLM emits several tool calls and we want them parallel," then correctness for the default tool ecosystem is part of the core feature story, not just a docs footnote.

enyst:
HUMAN: Just to note, I do think multiple tool calls are a normal occurrence. I don’t have numbers, but the agent here is confused; they’re often enough that, in practice, it’s like, see `codex-cli` fire up 12, think “woot was that a parallel tool call?”. Barely finish thought, see a bucket of 3, see another bucket of 5, conclude “yeah obviously”.

We have the eval trajectories we could compute to see the number of batches with the same `llm_response`?

VascoSch92:
> HUMAN: Just to note, I do think multiple tool calls are a normal occurrence. I don’t have numbers, but the agent here is confused; they’re often enough that, in practice, it’s like, see `codex-cli` fire up 12, think “woot was that a parallel tool call?”. Barely finish thought, see a bucket of 3, see another bucket of 5, conclude “yeah obviously”.
> 
> 
> 
> We have the eval trajectories we could compute to see the number of batches with the same `llm_response`?

@enyst yes this could be interesting.

VascoSch92:
@enyst I answered your points.

> 1) TOOL_CONCURRENCY_LIMIT=1 and backward compatibility

The previous implementation wasn't 100% backward compatible, but it was a simple fix. I have updated the code and added a test case to ensure full compatibility.

> 2) Later tools not seeing earlier observations from the same batch

I’m not sure I follow this point. If multiple tool calls are being executed in parallel, they shouldn't need to depend on each other's outputs.

> 3) Unsafe parallelism with terminal / file editor / browser

This is a valid point. However, I avoided adding that specific logic here to prevent the PR from becoming even larger than it already is. My approach was to first implement parallel tool calling and then address the prevention of dependent tool calls in a subsequent update. This aligns with what we originally discussed in the issue, and is why the current default remains the standard behavior.

> 4) Testing

I have added integration tests to cover scenarios where num_workers = 1 and the request contains multiple tool calls. This should show that we have backward compatibility.

openhands-sdk/openhands/sdk/agent/parallel_executor.py
```diff
@@ -0,0 +1,143 @@
...
+    if env_value:
+        with suppress(ValueError):
+            val = int(env_value)
+            if val > 0:
+                return val
+        logger.warning(
+            f"{ENV_TOOL_CONCURRENCY_LIMIT}={env_value} is invalid, "
+            f"using default {DEFAULT_TOOL_CONCURRENCY_LIMIT}"
+        )
+    return DEFAULT_TOOL_CONCURRENCY_LIMIT
```
xingyaoww: I'm not 100% sure if getting this from env var is the best approach here.. Maybe we can consider passing it to the Agent's configurations, similar to how we pass `tools`?

xingyaoww: @OpenHands what do you think? Do not modify code, just give me some suggestions here

xingyaoww: Great question — I agree that moving this to the Agent's configuration is the better approach. Here's my reasoning:

**Why the env var approach is problematic:**

1. **Violates the SDK's design principles.** Agents are immutable Pydantic models whose "configuration *is* their serializable representation." An env var creates a second, hidden source of truth that isn't captured in `model_dump()`, can't be persisted/replayed, and won't show up in config introspection.

2. **No per-agent control.** With an env var, every `Agent` and every subagent shares the same concurrency limit. But a parent orchestrator might want `concurrency=4` while its subagents stay at `1`. A field makes this trivial.

3. **Testing friction.** Tests currently need `monkeypatch.setenv` + careful teardown, which is fragile. A field is just `Agent(llm=..., tool_concurrency_limit=4)`.

**Suggested approach:**

Add a `tool_concurrency_limit` field to `AgentBase`:

```python
tool_concurrency_limit: int = Field(
    default=1,
    ge=1,
    description=(
        "Maximum number of tool calls to execute concurrently within a single "
        "agent step. Default is 1 (sequential). Values > 1 require that the "
        "tools in use are thread-safe for concurrent execution."
    ),
)
```

Then wire it through to `ParallelToolExecutor`:

```python
# In Agent.__init__ or a model_post_init:
self._parallel_executor = ParallelToolExecutor(max_workers=self.tool_concurrency_limit)

# ParallelToolExecutor becomes:
class ParallelToolExecutor:
    def __init__(self, max_workers: int = 1) -> None:
        self._max_workers = max_workers
```

**Benefits:**
- Fully declarative and serializable
- Per-agent control (parent vs subagent can differ)
- Clean testing (`Agent(llm=..., tool_concurrency_limit=4)` — no env var patching)
- Pydantic validation (`ge=1`) catches bad values at construction time
- `_get_max_concurrency()` and the env var constant can be removed entirely

**Optional: keep the env var as a deployment-level override?**

If there's a use case for operators overriding concurrency without changing agent configs (e.g., in production), you could keep the env var as a fallback when the field is at its default. But I'd lean toward removing it for simplicity — the field is a much cleaner contract, and deployment-level overrides can be handled by the config layer that constructs the Agent.

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -72,6 +76,133 @@
...
+            blocked_reasons=blocked_reasons,
+            results_by_id=results_by_id,
+        )
+
+    def emit(self, on_event: ConversationCallbackType) -> None:
+        """Emit all events in original action order."""
+        for ae in self.action_events:
+            reason = self.blocked_reasons.get(ae.id)
+            if reason is not None:
+                logger.info(f"Action '{ae.tool_name}' blocked by hook: {reason}")
```
xingyaoww, line 158: Are we sure this is always going to be blocked by hook? @Openhands can you double check and provide some evidences. Do not modify code

xingyaoww: **Yes, in the current codebase `blocked_actions` is always populated by hooks.** Here is the evidence:

### Write path (who populates `blocked_actions`)

There is exactly **one production caller** of `state.block_action()`:

- `openhands-sdk/openhands/sdk/hooks/conversation_hooks.py:167` inside `_handle_pre_tool_use()`:
  ```python
  # conversation_hooks.py:159-167
  if not should_continue:
      reason = self.hook_manager.get_blocking_reason(results)
      ...
      block_reason = reason or "Blocked by hook"
      self._conversation_state.block_action(event.id, block_reason)
  ```
  This is the PreToolUse hook processing path — no other code in the codebase calls `block_action()`.

Verified with:
```bash
grep -rn "block_action\b" --include="*.py" | grep -v "pop_blocked_action\|blocked_action\|test_\|test\.py\|example"
```
Returns only the definition (state.py:426) and the single call site (conversation_hooks.py:167).

### Read path (how `blocked_reasons` is consumed)

- `_ActionBatch.prepare()` (agent.py:137) calls `state.pop_blocked_action(ae.id)` → reads from `state.blocked_actions`
- The field definition (state.py:128-132) explicitly documents this:
  ```python
  # Hook-blocked actions: action_id -> blocking reason
  blocked_actions: dict[str, str] = Field(
      default_factory=dict,
      description="Actions blocked by PreToolUse hooks, keyed by action ID",
  )
  ```
- The method docstrings also confirm: `block_action` says "Persistently record a hook-blocked action" and `pop_blocked_action` says "Remove and return a hook-blocked action reason"

### Other rejection paths do NOT use `blocked_actions`

The other `UserRejectObservation` creation site (`local_conversation.py:735`) handles **user rejection** during confirmation mode — it creates the observation directly via `_on_event()` and never touches `blocked_actions`. That path uses the default `rejection_source="user"`.

### Serialization

The agent-server model (`models.py:223`) mirrors the field with the same description "Actions blocked by PreToolUse hooks". Deserialized/resumed state would contain values that were originally set by hooks.

### Minor caveat

Since `block_action()` is a public method on `ConversationState`, a future non-hook caller *could* theoretically use it, which would make the log message slightly misleading. But today, there is no such caller. If this is a concern, the method could be renamed to `_block_action` (private) or the log message could be made more generic (e.g., "Action blocked: {reason}").

openhands-sdk/openhands/sdk/agent/parallel_executor.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+if TYPE_CHECKING:
+    from openhands.sdk.event.base import Event
+    from openhands.sdk.event.llm_convertible import ActionEvent
+
+logger = get_logger(__name__)
+
+# Default concurrency limit for tool executions (per agent)
+DEFAULT_TOOL_CONCURRENCY_LIMIT: Final[int] = 1
```
xingyaoww: Let's configure default tool concurrency to 4 or 8, and we should create a example file that prompts the agent to run multiple tools at the same time (even delegating multiple agents) to stress test the parallel tool execution system. I'd leaning towards merging this PR once this test passes via `test-examples` workflow

VascoSch92:
@xingyaoww 

> I'm not 100% sure if getting this from env var is the best approach here.. Maybe we can consider passing it to the Agent's configurations, similar to how we pass tools?

Makes sense. I changed as suggested by OH

> Let's configure default tool concurrency to 4 or 8, and we should create a example file that prompts the agent to run multiple tools at the same time (even delegating multiple agents) to stress test the parallel tool execution system. I'd leaning towards merging this PR once this test passes via test-examples workflow

I created an example where we spawn subagents that perform code searches. This happens in parallel, and at the end there is also a parallel report to confirm the calls were actually executed in parallel.

If it is too verbose for an example I can also remove it. I just wanted to make sure we are correctly parallelizing the calls.

The report looks like:

```text
================================================================================
Parallelism Report
================================================================================

  main agent batch (chatcmpl-272dfe5...):
    Parallel tools: ['task', 'task', 'task']

  sub-agent batch (chatcmpl-ae3ab59...):
    Parallel tools: ['terminal', 'file_editor']

  sub-agent batch (chatcmpl-9f48c72...):
    Parallel tools: ['file_editor', 'file_editor']

  sub-agent batch (chatcmpl-db48e30...):
    Parallel tools: ['file_editor', 'file_editor']

  sub-agent batch (chatcmpl-affbbd2...):
    Parallel tools: ['file_editor', 'file_editor']

  sub-agent batch (chatcmpl-3854bba...):
    Parallel tools: ['file_editor', 'file_editor']

  sub-agent batch (chatcmpl-3742934...):
    Parallel tools: ['file_editor', 'file_editor']

  sub-agent batch (chatcmpl-508834f...):
    Parallel tools: ['file_editor', 'file_editor', 'file_editor', 'file_editor']

  sub-agent batch (chatcmpl-4795c8d...):
    Parallel tools: ['terminal', 'file_editor']

  sub-agent batch (chatcmpl-922b7a5...):
    Parallel tools: ['file_editor', 'file_editor']
```

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -187,6 +187,17 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
         examples=[{"kind": "AgentFinishedCritic"}],
     )
 
+    tool_concurrency_limit: int = Field(
+        default=1,
+        ge=1,
+        description=(
+            "Maximum number of tool calls to execute concurrently within a single "
+            "agent step. Default is 1 (sequential). Values > 1 enable parallel "
```
enyst, line 195: @openhands can you answer: for 1, is the behavior in this PR the same as it currently is before this PR?

enyst:
> The report looks like:

Ha! There’s no `terminal, terminal` 😅. But if Xingyao is happy, I’m happy 🙏

VascoSch92:
> > The report looks like:
> 
> Ha! There’s no `terminal, terminal` 😅. But if Xingyao is happy, I’m happy 🙏

If you want I can force the example to have `terminal, terminal` :-)

openhands-sdk/openhands/sdk/agent/base.py
enyst: Short answer: **not fully**.

After the latest change, `tool_concurrency_limit=1` now matches pre-PR behavior in one important respect: `ParallelToolExecutor.execute_batch()` special-cases `max_workers == 1`, so multi-tool batches run on the caller thread and in order instead of going through `ThreadPoolExecutor`.

But it is still **not identical** to `main` overall. `_execute_actions()` still routes through `_ActionBatch.prepare()`, and that method executes the whole executable subset before `batch.emit()` runs. On `main`, the flow is `run tool -> emit observation -> run next tool`. In this PR, with limit 1, the flow is still effectively `run all tools -> emit observations afterward`.

So if by “same behavior” you mean **thread affinity / execution order**, then yes, that part now matches. If you mean **full old sequential semantics**, then no: emission timing and mid-batch visibility in `conversation.state.events` still differ from `main` for multi-tool responses.

The new regression test also seems to cover the first part (caller thread + order), but not the second part (incremental emission / state visibility).

enyst:
> > > The report looks like:
> > 
> > 
> > Ha! There’s no `terminal, terminal` 😅. But if Xingyao is happy, I’m happy 🙏
> 
> If you want I can force the example to have `terminal, terminal` :-)

😇 Just out of curiosity

VascoSch92:
_Reminder: The main focus of this PR is simply to add the infrastructure for parallel tool calls. There will be additional PRs to ensure everything is thread-safe. For this reason, we have set the default to sequential tool calls._

@enyst I investigate your concerns and here what is going on.

Current state: TerminalTool is not thread-safe. All commands go through a single PTY. Two concurrent commands on the same PTY interleave their bytes, corrupting output.

When it matters: Only when tool_concurrency_limit > 1 and the LLM emits multiple terminal calls in the same response. **Subagents are fine: each gets its own session.**

What's safe today: Parallel batches with different tool types (e.g., 3 delegate calls, or terminal + file editor). This is the common case.

Possible fixes (simplest to most capable)
1. Lock on execute()
  - Parallelism: none (serializes terminal calls)
  - Complexity: minimal — one lock
  - Tradeoff: correct but minimal speedup for terminal calls
2. Session pool
  - Parallelism: full
  - Complexity: high — N shell processes, cwd sync between sessions
  - Tradeoff: real parallelism but resource heavy and complex
3. Tmux pane per read-only command
  - Parallelism: for reads only
  - Complexity: medium — new pane per call, classify read vs write
  - Tradeoff: fast spawn (~10-20ms), inherits cwd, but tmux-only backend

What I propose:
- For this PR: No terminal changes are needed. The parallel execution system works for the main use case (multiple sub-agents, mixed tool types), and the infrastructure is already in place. I experimented by forcing parallel tool calls in a terminal and found that conflicts occur occasionally (about 5 out of 30 times). This didn’t break the flow; the agent simply had to perform another terminal call to retrieve the missing information.
- Immediately after this PR: I will submit a PR with a terminal lock. This is a minimal change with no behavioral impact. We will lose some parallelism, but we'll be safer, and we still gain performance (and save tokens) for mixed tool batches.
- Future steps: We can experiment to see if the lock becomes a real bottleneck. If so, we can implement options 2 or 3.

For comparison, here are the parallel tool calls from my CC sessions:

| Pattern | Occurrences | Max Batch Size |                                                                                                                                                                         
  |---|---|---|                                
  | `[Read, Read, ...]` | 20+ | 6 |                                                                                                                                                                                  
  | `[Bash, Bash]` | 10+ | 2 |
  | `[Read, Grep]` | 6+ | 2 |                                                                                                                                                                                        
  | `[Read, Bash, ...]` | 5+ | 3 |                                                                                                                                                                                   
  | `[Grep, Grep, ...]` | 2+ | 4 |                                                                                                                                                                                   
  | `[Read, Grep, Grep]` | 2 | 3 |                                                                                                                                                                                   
  | `[Read, Read, Bash]` | 2 | 3 |                                                                                                                                                                                   
  | `[Glob, Glob, Glob, Bash, Bash, Bash]` | 1 | 6 |                                                                                                                                                                 
  | `[Glob, Grep, Grep]` | 1 | 3 |                                                                                                                                                                                   
  | `[Glob, Glob, Bash]` | 1 | 3 |                                                                                                                                                                                   
  | `[Grep, Read, Read, Read]` | 1 | 4 |                          
  | `[Grep, Bash, Bash]` | 1 | 3 |                                                                                                                                                                                   
  | `[Agent, Agent]` | 1 | 2 |                                                                                                                                                                                       
  | `[WebFetch, Agent]` | 1 | 2 |                                                                                                                                                                                    
  | `[Task, Task, Task, Task, Task, Task, Task]` | 1 | 7 |  

Note that Read is paralleizabe (because of FileEditor view).

enyst:
> I experimented by forcing parallel tool calls in a terminal and found that conflicts occur occasionally (about 5 out of 30 times). This didn’t break the flow; the agent simply had to perform another terminal call to retrieve the missing information.

Aha! Thank you. 😅 

Your proposals sound good to me, thanks for satisfying my dumb little curiosity.

VascoSch92:
> 
> With this context, https://github.com/OpenHands/software-agent-sdk/pull/2390#issuecomment-4080751125
> 
> This PR LGTM!
> 
> > Immediately after this PR: I will submit a PR with a terminal lock. This is a minimal change with no behavioral impact. We will lose some parallelism, but we'll be safer, and we still gain performance (and save tokens) for mixed tool batches.
> 
> Shall we create an issue here in case we forget about it?

I have a linear ticket to remember myself to create all the relative issues to continue the parallel tool calls dev.

Moreover, i also have to create documentation as I introduced a new example.

PR #1788, opened 2026-01-22
refactor: enforce tool immutability on conversation restore

enyst:
HUMAN: This PR proposes an alternative for system prompt event becoming stale when tools are added/removed. It seems a bit tougher, but… it’s the system prompt.

—-

## Summary

Tools are part of the system prompt and cannot be changed mid-conversation. This PR enforces that constraint strictly, simplifying the model: each conversation has exactly one set of tools, defined at creation time.

## Changes

- `verify()` now requires exact tool name match (no relaxed matching based on "used tools")
- Clear error message directs users to start a new conversation or use conversation forking (planned)
- Updated tests to expect failures when tools change

## Why?

This is a prerequisite for implementing **conversation forking** as the proper solution for changing tools mid-session. See:
- **#1787**: Proposal: Fork conversation when tools change (alternative to SystemPromptUpdateEvent)
- **OpenHands/OpenHands#8560**: [PRD] Fork a Conversation

## Alternative

PR #1777 introduced `SystemPromptUpdateEvent` to record tool changes in the event log. This PR takes a different approach: enforce immutability now, implement forking later.

The forking approach is cleaner because:
1. Each conversation has ONE system prompt (no "latest wins" logic)
2. Event log remains truly immutable
3. Aligns with the broader forking feature request

## References

- #1451: Proposal: agent as composed of immutable instances
- #1778: (investigation) SystemPromptEvent not updated when adding tools on restore
- #1777: PR that introduced `SystemPromptUpdateEvent`
- #1787: Proposal for forking as the solution

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/492e7888f0c340ec97c644a46592a061)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
| python | amd64, arm64 | `nikolaik/python-nodejs:python3.12-nodejs22` | [Link](https://hub.docker.com/_/nikolaik/python-nodejs:python3.12-nodejs22) |
| golang | amd64, arm64 | `golang:1.21-bookworm` | [Link](https://hub.docker.com/_/golang:1.21-bookworm) |

**Pull (multi-arch manifest)**
```bash
[...]

PR #2393, opened 2026-03-12
fix(sdk/agent): allow adding tools when resuming conversations for backward compatibility

VascoSch92:
## Summary

Solution for problem discussed in: https://openhands-ai.slack.com/archives/C0A6XE64XK4/p1773178175274459

- Allow adding new tools when resuming a conversation (e.g. delegate added in a new SDK version) while still rejecting tool removals
- Removing tools breaks backward compatibility because the LLM may have already seen them in prior turns; adding tools is safe since the LLM simply gains new capabilities on the next turn
- This is the least invasive fix: a ~5-line change in AgentBase.verify() that relaxes the exact-match check to a superset check   

## Checklist

### DIFF
--- a/tests/sdk/conversation/local/test_state_serialization.py
+++ b/tests/sdk/conversation/local/test_state_serialization.py
@@ -1283,8 +1283,8 @@ def test_v1_11_5_cli_default_conversation_resumes_when_runtime_adds_delegate(
 
     Adding new tools is allowed — only removing tools is rejected.
     """
+    from openhands.sdk.agent import Agent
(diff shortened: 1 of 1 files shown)


enyst:
For reference, we’ve been going a bit back and forth on this one. This was the last time we rejected added tools:
- https://github.com/OpenHands/software-agent-sdk/pull/1788

This was an alternative we didn’t eventually like, to try to add them, but also update the system prompt:
- https://github.com/OpenHands/software-agent-sdk/pull/1777

PR #2404, opened 2026-03-13
Deprecate {path:path} endpoints in file_router and add query param alternatives

chuckbutkus:
## Summary

This PR updates `file_router.py` to deprecate the `{path:path}` URL path parameter endpoints and adds new endpoints that use query parameters for the path, following the same pattern as `git_router.py`.

### file_router.py
- Added new preferred endpoints with query parameters:
  - `POST /file/upload` with `path` as query parameter
  - `GET /file/download` with `path` as query parameter
- Deprecated legacy `{path:path}` endpoints (v1.15.0 → v1.20.0):
  - `POST /file/upload/{path:path}`
  - `GET /file/download/{path:path}`
- Extracted upload/download logic into internal helper functions (`_upload_file`, `_download_file`)

### Tests (test_file_router.py)
Added 20 tests covering:
- Query parameter endpoints (preferred method)
- Path parameter endpoints (legacy/backwards compatibility)
- Error cases (missing params, relative paths, file not found, directory instead of file)
- Edge cases (large files, special characters in paths, nested directories)

## Why

Using query parameters instead of path parameters for file paths:
- Allows better handling of complex paths (especially those with special characters)
- Is more consistent with other endpoints
- Avoids URL encoding issues with paths containing slashes

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:b814738-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-b814738-python \
  ghcr.io/openhands/agent-server:b814738-python
```

```
ghcr.io/openhands/agent-server:b814738-golang-amd64
ghcr.io/openhands/agent-server:b814738-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:b814738-golang-arm64
ghcr.io/openhands/agent-server:b814738-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:b814738-java-amd64
ghcr.io/openhands/agent-server:b814738-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:b814738-java-arm64
ghcr.io/openhands/agent-server:b814738-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:b814738-python-amd64
ghcr.io/openhands/agent-server:b814738-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:b814738-python-arm64
ghcr.io/openhands/agent-server:b814738-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:b814738-golang
ghcr.io/openhands/agent-server:b814738-java
ghcr.io/openhands/agent-server:b814738-python
```

- Each variant tag (e.g., `b814738-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `b814738-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/file_router.py
+++ b/openhands-agent-server/openhands/agent_server/file_router.py
@@ -54,7 +54,7 @@ async def _upload_file(path: str, file: UploadFile) -> Success:
         return Success()
 
     except HTTPException:
-    except Exception as e:
+        raise
     except Exception as e:
         logger.error(f"Failed to upload file: {e}")
         raise HTTPException(


enyst:
@openhands Read all comments, you see a comment says a check failed about REST API breakage checks. Look at the CI log for that job. Let’s verify deeply if the script works correctly
- wdyt, did the script detect correctly breakages? It has 2 errors in CI
- why does it compare 1.13 to 1.13, does it compare the PR contents to the last release?
- did this PR really change REST API?

/codereview-roasted

enyst:
@enyst I dug through the job and the script. Short version: the check is **not PR-scoped**, and this failure is very likely a **false positive caused by schema generation under different FastAPI versions**, not by this PR introducing a breaking REST change.

What I found:

- The workflow runs `.github/scripts/check_agent_server_rest_api_breakage.py`.
- That script picks the baseline with `_get_baseline_version()`.
  - If the current version already exists on PyPI, it returns **that same version**.
  - So `1.13.0 -> 1.13.0` is expected here: it means “compare this branch’s current code against the published `openhands-agent-server==1.13.0` release”.
  - In other words, yes: it compares PR contents to the last published release, **not** to `main`, and **not** only to the PR diff.

Why the reported breakage looks wrong:

- The reported diff was:
  - `the 'file' request property type/format changed from 'string'/'' to 'string'/'binary'`
- I reproduced the schemas locally and the comparison is apples-to-oranges:
  - **current workspace schema** is generated in the repo's locked dev env (`uv sync --frozen --group dev`), which currently uses **FastAPI 0.129.0**
  - **baseline release schema** is generated in a temp venv via `uv pip install openhands-agent-server==1.13.0 ...` with **floating transitive deps**, which resolves **FastAPI 0.135.1**
- With `openhands-agent-server==1.13.0` on FastAPI **0.135.1**, the upload body schema comes out as:
  - `{"type": "string", "contentMediaType": "application/octet-stream"}`
- With the **same 1.13.0 code** on FastAPI **0.129.0**, it comes out as:
  - `{"type": "string", "format": "binary"}`
- The current branch and `origin/main` both generate `format: binary` under the workspace env.

So the job is detecting a schema difference relative to the published release artifact, but that difference is **not caused by this PR**. It’s caused by the script regenerating the baseline schema with a different dependency stack than the one used for the current workspace schema.

About the “2 errors”:

- There is really **one policy error** from the script (`Breaking REST API change detected without MINOR version bump...`)
- The second error in the log is just GitHub Actions reporting the step exited non-zero (`Process completed with exit code 1`)

Did this PR really change the REST API?

- **Yes**, in the sense that it adds new endpoints:
  - `POST /api/file/upload?path=...`
  - `GET /api/file/download?path=...`
- But **no**, it does **not** appear to introduce the breaking change that CI reported.
- The old `{path:path}` endpoints are still present, so this PR is an additive/backward-compatible API change, not a removal.

One extra thing I noticed: the PR uses `@deprecated(...)`, but that decorator does **not** mark the FastAPI operations as `deprecated: true` in OpenAPI. So the legacy routes are not actually deprecated in the generated schema yet. That’s separate from this false-positive CI failure, but it matters for the future removal policy.

So my take: the script’s **baseline selection is intentional**, but the implementation is currently **unsound** because it regenerates the baseline schema in an unconstrained environment. That can make unrelated dependency drift look like a REST API breakage.

xingyaoww (approved):
Thank you!

PR #2446, opened 2026-03-14
fix(sdk): preserve model_name in get_combined_metrics

softpudding:
The get_combined_metrics() method was creating a new Metrics() with the default model_name='default', losing the actual model name from tracked metrics. This fix extracts the model_name from the first metrics entry before creating the combined metrics.

## Summary

## Checklist


openhands-sdk/openhands/sdk/conversation/conversation_stats.py
```diff
@@ -56,7 +56,12 @@ def _serialize_with_context(self, serializer: Any, info: Any) -> dict[str, Any]:
         return data
 
     def get_combined_metrics(self) -> Metrics:
-        total_metrics = Metrics()
+        model_name = (
+            next(iter(self.usage_to_metrics.values())).model_name
+            if self.usage_to_metrics
+            else "default"
+        )
+        total_metrics = Metrics(model_name=model_name)
```
malhotra5, line 64: Hmm how would this work in the event we had multiple models with different names? Should we concatenate all the unique model names 😂

softpudding: Oh I thought for a conversation in openhands only 1 single model is permitted, otherwise check in llm.py fails, something like "LLM provided is different from one in persisted state" happens. Was that design removed and we can use different models for a conversation?

enyst: Yes, we removed that design in https://github.com/OpenHands/software-agent-sdk/pull/1542

It's possible to use different models at conversation restore. With this PR it should be possible at runtime too, just maybe we may want a command to do it (e.g. "/model" or "/switch") https://github.com/OpenHands/software-agent-sdk/pull/2192

softpudding: I see! In this case I don't think this PR makes sense. Thanks.

PR #2449, opened 2026-03-14
fix(sdk): restore ACP cost tracking fallback

### LINKED ISSUE #2375: [Bug] ACP Cost Tracking Bug
### Summary

Analyzying the traces of [this](https://openhands-ai.slack.com/archives/C09QGUDQVTL/p1773037302157719) Gaia Evaluation, it comes up that 68 out of 157 completed instances (43%) report exactly $0.00 cost despite performing significant work.

Precisely, these 68 zero-cost instances:
- Resolved 51 tasks successfully
- Made 1,783 ACP tool calls (avg 26.2/instance) -- actually more than the tracked instances (avg 16.6/instance)
- 1,446 fetch calls (web search/browse)
- 293 execute calls (terminal/python)
- 36 Agent (subagent) calls
- Produced 26,263 chars of log output on average
- Have costs: [] (empty array) in their metrics -- no cost entries at all
- Report prompt_tokens: 0, completion_tokens: 0 in accumulated_token_usage -- zero tokens despite multi-turn conversations

What tracked instances look like for the 89 instances with cost tracking report:
- Cost model: "acp-managed" with a single cost entry per conversation turn
- Avg ~364 prompt tokens, ~6,583 completion tokens, ~232,912 cache_read tokens per instance
- Total: $54.60 across 1,475 ACP calls

### Possible Root cause
The bug is a race condition in notification draining combined with ACP's lack of turn-completion signaling.

In acp_agent.py:68-76, _drain_notifications() is a best-effort hack:

```
  async def _drain_notifications() -> None:
      await asyncio.sleep(0)
      await asyncio.sleep(_NOTIFICATION_DRAIN_DELAY)  # default 0.1s
```

  The flow is:
  1. conn.prompt() sends user message to ACP server and waits for PromptResponse
  2. _drain_notifications() sleeps 100ms hoping UsageUpdate notifications arrive
  3. _record_usage() reads the cost/token data that (hopefully) arrived

The problem: UsageUpdate notifications are sent as async session updates by the ACP server. They arrive via the same JSON-RPC connection but as separate notifications. If the ACP server sends the UsageUpdate after the PromptResponse and the 100ms drain window is too short (due to network latency, server load, or the server simply sending it late), the UsageUpdate never gets processed for that turn.

The pattern of "all-or-nothing" (entire instances with zero cost, not individual turns) suggests the issue is per-session, not per-turn. This points to a second factor:

  - UsageUpdate.cost is Optional — the ACP server (Claude Code) may simply not send cost data at all for some sessions. The field is: cost: Optional[Cost] = None
  - The code at line 180 checks if update.cost is not None — if the ACP server sends UsageUpdate with cost=None, nothing gets recorded
  - Similarly, PromptResponse.usage is Optional (marked as UNSTABLE in the schema) — if the server doesn't populate it, _record_usage skips token recording entirely (line 344: if response is not None and
  response.usage is not None)

  The two contributing factors

  1. ACP server doesn't always send cost/usage data: Some Claude Code ACP server instances may not report cost in UsageUpdate or usage in PromptResponse — both are Optional fields. This explains the all-or-nothing
   pattern per instance.
  2. Race condition in drain: Even when cost data IS sent, the 100ms drain window may miss late-arriving UsageUpdate notifications. The code explicitly acknowledges this with the TODO referencing
  https://github.com/agentclientprotocol/agent-client-protocol/issues/554.

### Possible Solution
 Two fixes, both should be applied:

**Fix 1 — Use PromptResponse.usage for cost estimation when UsageUpdate is missing (in _record_usage):**

Currently _record_usage only records tokens from PromptResponse.usage but never estimates cost from it. When UsageUpdate.cost is missing (the costs array is empty), you should compute a cost estimate from the token counts in PromptResponse.usage using known model pricing. This is the fallback that was apparently removed in commit 6e99789d.

**Fix 2 — Increase drain delay or make it adaptive:**

Change _NOTIFICATION_DRAIN_DELAY default from 0.1 to something like 0.5 or 1.0, or better yet, implement a polling loop that waits until a UsageUpdate with cost data arrives (with a timeout). Something like:
```python
  async def _drain_notifications(client, timeout=2.0) -> None:
      deadline = asyncio.get_event_loop().time() + timeout
      while asyncio.get_event_loop().time() < deadline:
          await asyncio.sleep(0.1)
          if client._last_cost > 0:  # or check if UsageUpdate was received this turn
              break
```

**Fix 3 (Recommended) — Record cost from UsageUpdate AND fire the stats callback immediately:**

Currently at line 180-184, cost is added to metrics but the stats callback only fires later in _record_usage. If _record_usage is called before the UsageUpdate arrives (race condition), the cost is lost from the callback perspective. Move the stats callback firing into the session_update handler too, or record the cost delta in _record_usage based on whatever accumulated by that point.

The simplest high-confidence fix is restoring the cost fallback estimation from token usage in _record_usage — if after processing a turn, metrics.costs is still empty but PromptResponse.usage has token counts, estimate the cost from tokens. This handles both the "server never sent cost" and "drain missed the UsageUpdate" cases.

LarytheLord:
## Summary
- restore ACP cost estimation from `PromptResponse.usage` when `UsageUpdate.cost` is missing
- reconcile late real `UsageUpdate.cost` values against the estimated baseline instead of double-counting
- notify stats listeners when cost arrives through `UsageUpdate` and cover the new paths with ACP telemetry tests

## Testing
- `uv run pytest tests/sdk/agent/test_acp_agent.py`
- `uv run pre-commit run --files openhands-sdk/openhands/sdk/agent/acp_agent.py tests/sdk/agent/test_acp_agent.py`

Fixes #2375.

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/acp_agent.py
+++ b/openhands-sdk/openhands/sdk/agent/acp_agent.py
@@ -196,6 +196,73 @@ async def _filter_jsonrpc_lines(source: Any, dest: Any) -> None:
         dest.feed_eof()
 
 
+def _normalize_cost_model_name(model_name: str | None) -> str | None:
+    """Normalize model names used for ACP cost estimation."""
+    if not model_name or model_name == "acp-managed":
+        return None
+
+    normalized = model_name.split(":", 1)[0]
+    if "/" in normalized:
+        normalized = normalized.split("/", 1)[1]
+    return normalized or None
+
+
+def _estimate_cost_from_tokens(
+    usage: Any,
+    *,
+    model_name: str | None = None,
+    input_cost_per_token: float | None = None,
+    output_cost_per_token: float | None = None,
+) -> float | None:
+    """Estimate cumulative USD cost from ACP token usage.
+
+    ACP servers can omit ``UsageUpdate.cost`` while still returning cumulative
+    token counts on ``PromptResponse.usage``. In that case we estimate cost from
+    the token totals and later reconcile against real cost if it arrives.
+    """
+    prompt_tokens = int(getattr(usage, "input_tokens", 0) or 0)
+    completion_tokens = int(getattr(usage, "output_tokens", 0) or 0)
+    cache_read_tokens = int(getattr(usage, "cached_read_tokens", 0) or 0)
+    cache_write_tokens = int(getattr(usage, "cached_write_tokens", 0) or 0)
+
+    if not any(
+        (
+            prompt_tokens,
+            completion_tokens,
+            cache_read_tokens,
+            cache_write_tokens,
+        )
+    ):
+        return None
+
+    if input_cost_per_token is not None and output_cost_per_token is not None:
+        return (
+            prompt_tokens * input_cost_per_token
+            + completion_tokens * output_cost_per_token
+        )
+
+    normalized_model = _normalize_cost_model_name(model_name)
+    if normalized_model is None:
+        return None
+
+    try:
+        import litellm  # noqa: PLC0415
+
+        info = litellm.model_cost.get(normalized_model)
+        if info is None:
+            return None
+
+        return (
+            prompt_tokens * info.get("input_cost_per_token", 0)
+            + completion_tokens * info.get("output_cost_per_token", 0)
+            + cache_read_tokens * info.get("cache_read_input_token_cost", 0)
+            + cache_write_tokens * info.get("cache_creation_input_token_cost", 0)
+        )
+    except Exception:
+        logger.debug("ACP cost estimation failed", exc_info=True)
+        return None
+
+
 class _OpenHandsACPBridge:
     """Bridge between OpenHands and ACP that accumulates session updates.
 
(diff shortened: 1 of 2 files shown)


simonrosenberg:
Thanks for working on this. I do think this likely helps the ACP-side “missing `UsageUpdate.cost` / late cost arrival” case.

I don’t think it is a complete fix for `#2375`, though, because the latest ACP SWE-bench artifacts point to a second bug outside this codepath:

- run: `22967419020-claude-4-6`
- archive: https://results.eval.all-hands.dev/swebench/litellm_proxy-anthropic-claude-opus-4-6/22967419020/results.tar.gz
- `27/500` instances still saved `accumulated_cost=0`, `prompt_tokens=0`, `completion_tokens=0`
- all 27 still produced non-empty patches
- all `27/27` broken rows hit `Run completed via REST fallback after 3 consecutive terminal polls`
- `0/473` of the nonzero-telemetry rows hit that fallback

So this PR may improve the ACP telemetry race, but it does not address the remote run-completion / final-state-refresh problem that also appears to be dropping stats before benchmarks serialize them.

Details are in the latest `#2375` comments:
- artifacts: https://github.com/OpenHands/software-agent-sdk/issues/2375#issuecomment-4066805333
- root cause: https://github.com/OpenHands/software-agent-sdk/issues/2375#issuecomment-4066865732
- cleaner no-fallback fix: https://github.com/OpenHands/software-agent-sdk/issues/2375#issuecomment-4067222559

LarytheLord:
Thanks @simonrosenberg a lot for digging into the eval artifacts and writing this up so clearly.

That makes sense to me too. You’re right that this one is only trying to cover the missing / late `UsageUpdate.cost` path inside `ACPAgent`, and it doesn’t address the separate stale final-state refresh problem on the REST fallback path and the no cost thing also.

Given that `#2460` is taking the broader fix in the direction you described, I’m going to close this one as superseded so the discussion stays focused in one place.

Really appreciate the detailed breakdown here. It helped me to actual boundary of this fix much clearer.

PR #2450, opened 2026-03-15
Fix apptainer workspace cleanup: kill zombie child processes.

adityasoni9998:
## Summary

This PR fixes the logic for cleaning up an Apptainer workspace instance. Currently [self._process](https://github.com/OpenHands/software-agent-sdk/blob/20ff7ba8efb089a9d1bae943d43410b6a6758126/openhands-workspace/openhands/workspace/apptainer/workspace.py#L286) inherits the same process group as that of the parent process. As a result, when we cleanup and terminate this process [here](https://github.com/OpenHands/software-agent-sdk/blob/20ff7ba8efb089a9d1bae943d43410b6a6758126/openhands-workspace/openhands/workspace/apptainer/workspace.py#L367), the child processes spawned by the `apptainer run` command are not killed (for e.g. processes responsible for overlaying file-system) which causes them to become zombie processes. 

This is particularly problematic for situations where the same machine spins up O(1K-10K) apptainer instances, for e.g. during rollouts for RL training. This PR fixes this issue by making two changes - 
- start self._process as a new session which ensures that self._process is the leader of its process group which also means that the PID of this process will also be the process group ID
- during cleanup, instead of using process.terminate(), send an OS signal (SIGTERM/SIGKILL) through os.killpg(<PID of self._process == PID of this group>) to kill the entire process group including the child processes spawned by self._process.

## Checklist

### DIFF
--- a/openhands-workspace/openhands/workspace/apptainer/workspace.py
+++ b/openhands-workspace/openhands/workspace/apptainer/workspace.py
@@ -1,6 +1,7 @@
 """Apptainer-based remote workspace implementation."""
 
 import os
+import signal
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


neubig (approved):
Thanks a lot!

PR #2454, opened 2026-03-15
Add tags support to conversations API

rbren:
## Summary

Add key-value tag metadata to conversations. Tags can be specified when creating or updating a conversation via the SDK or REST API.

### Validation rules
- **Keys**: lowercase alphanumeric only (`^[a-z0-9]+$`)
- **Values**: arbitrary strings, max 256 characters

### SDK layer (`openhands-sdk/`)
- **`ConversationTags` type** in `conversation/types.py` — Pydantic-validated `dict[str, str]` with a `BeforeValidator` enforcing key/value constraints
- **`ConversationState`** — new persisted `tags` field (auto-saved to `base_state.json`)
- **`Conversation` factory** — `tags` parameter on all overloads
- **`LocalConversation`** — accepts `tags`, flows to `ConversationState.create`
- **`RemoteConversation`** — accepts `tags`, includes in server POST payload
- **Exports** — `ConversationTags` exported from conversation package

### Server API layer (`openhands-agent-server/`)
- **`StartConversationRequest`** — `tags: ConversationTags` field (defaults to `{}`)
- **`UpdateConversationRequest`** — `tags: ConversationTags | None` field (`PATCH` can now update tags only, title only, or both; `title` made optional)
- **`StoredConversation` / `ConversationInfo`** — inherit tags from parents
- **`conversation_service.update_conversation`** — handles tag updates, syncs to `ConversationState`

### Tests (18 new, all passing)
- `tests/sdk/conversation/test_tags.py` — validation unit tests (valid keys, invalid keys, value length limits, Pydantic model integration)
- `tests/agent_server/test_conversation_tags.py` — API endpoint tests (create with/without tags, invalid keys rejected with 422, PATCH tags, GET returns tags)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:d9f8c26-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-d9f8c26-python \
  ghcr.io/openhands/agent-server:d9f8c26-python
```

```
ghcr.io/openhands/agent-server:d9f8c26-golang-amd64
ghcr.io/openhands/agent-server:d9f8c26-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:d9f8c26-golang-arm64
ghcr.io/openhands/agent-server:d9f8c26-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:d9f8c26-java-amd64
ghcr.io/openhands/agent-server:d9f8c26-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:d9f8c26-java-arm64
ghcr.io/openhands/agent-server:d9f8c26-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:d9f8c26-python-amd64
ghcr.io/openhands/agent-server:d9f8c26-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:d9f8c26-python-arm64
ghcr.io/openhands/agent-server:d9f8c26-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:d9f8c26-golang
ghcr.io/openhands/agent-server:d9f8c26-java
ghcr.io/openhands/agent-server:d9f8c26-python
```

- Each variant tag (e.g., `d9f8c26-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `d9f8c26-python-amd64`) are also available if needed

### DIFF
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -14,8 +14,27 @@ Outputs to GITHUB_OUTPUT:
 
 import json
 import os
+import signal
 import sys
+import time
 from typing import Any
+
+
+def _sigterm_handler(signum: int, _frame: object) -> None:
+    """Handle SIGTERM/SIGALRM with a diagnostic message instead of silent death."""
+    sig_name = signal.Signals(signum).name
+    print(
+        f"\nERROR: Process received {sig_name} during preflight check.\n"
+        "This usually means the LiteLLM proxy is unreachable or hanging.\n"
+        f"LLM_BASE_URL: {os.environ.get('LLM_BASE_URL', '(not set)')}\n",
+        file=sys.stderr,
+        flush=True,
+    )
+    sys.exit(1)
+
+
+signal.signal(signal.SIGTERM, _sigterm_handler)
+signal.signal(signal.SIGALRM, _sigterm_handler)
 
 
 # SDK-specific parameters that should not be passed to litellm.
@@ -380,6 +399,31 @@ def check_model(
 test_model = check_model
 
 
+def _check_proxy_reachable(
+    base_url: str, api_key: str | None = None, timeout: int = 10
+) -> tuple[bool, str]:
+    """Quick health check: can we reach the proxy at all?
+
+    Uses /v1/models (standard OpenAI-compatible endpoint) which works with
+    any valid API key. The /health endpoint requires admin-level access on
+    some LiteLLM configurations.
+    """
+    import urllib.error
+    import urllib.request
+
+    models_url = f"{base_url.rstrip('/')}/v1/models"
+    try:
+        req = urllib.request.Request(models_url, method="GET")
+        if api_key:
+            req.add_header("Authorization", f"Bearer {api_key}")
+        urllib.request.urlopen(req, timeout=timeout)
+        return True, f"Proxy reachable at {base_url}"
+    except urllib.error.URLError as e:
+        return False, f"Cannot reach proxy at {base_url}: {e.reason}"
+    except Exception as e:
+        return False, f"Cannot reach proxy at {base_url}: {type(e).__name__}: {e}"
+
+
 def run_preflight_check(models: list[dict[str, Any]]) -> bool:
     """Run preflight LLM check for all models.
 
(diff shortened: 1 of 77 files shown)


rbren:
@openhands check github actions and fix the failing tests

enyst:
@openhands Merge main into this PR. Understand it well and keep the goals. Then babysit CI until green or if you have a problem that needs human attention, post a comment on the PR with your question so that we know.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -563,8 +563,13 @@ async def update_conversation(
...
 
-        # Update the title and timestamp in stored conversation
-        event_service.stored.title = request.title.strip()
+        if request.title is not None:
+            event_service.stored.title = request.title.strip()
+        if request.tags is not None:
+            event_service.stored.tags = request.tags
+            # Also update tags on the conversation state so they persist
+            state = await event_service.get_state()
+            state.tags = request.tags
```
enyst: @openhands trace in the execution path what object is this `state`, is it `ConversationState`? I think the code should always use a lock when it makes changes to `ConversationState`. Is this code bugged? Verify all I said.

If so, give me a few alternatives here. Investigate the PR and understand the goal.

enyst: Traced it through: `ConversationService.update_conversation()` -> `EventService.get_state()` -> `self._conversation._state`, so `state` here is the live `ConversationState`, not a copy.

Your locking concern was valid for this PR path. `ConversationState` autosaves on field assignment and invokes the state-change callback from `__setattr__`, and the surrounding code generally assumes those mutations happen under `with state:`. This new `state.tags = ...` path was mutating outside the lock, so the callback/snapshot path could run without the usual synchronization.

I fixed that in ddd52ed1 by:
- taking the `ConversationState` lock before mutating `state.tags`
- taking the same lock when composing the webhook snapshot right after the update
- adding a regression test that spies on the state lock during tag updates

A few reasonable alternatives here were:
1. Minimal fix (what I shipped): lock in `ConversationService.update_conversation()`.
2. Better encapsulation: add an `EventService` helper for locked state mutations so server code never mutates raw state directly.
3. Stronger API boundary: expose snapshot/update methods and stop returning mutable `ConversationState` from `get_state()`.

One nuance: I do agree with the intended rule that `ConversationState` writes should use the lock, but that rule is not perfectly enforced everywhere in the codebase today — there are a couple of pre-existing unlocked writes outside this PR too. This PR no longer adds another one.

enyst (commented):
Out of curiosity, why KV as opposed to simply strings? Sorry for being dense, but if we also need keys, I also wonder if we need to expose them to the client, or could we let them send simple strings?

I'd love to know how you think of the use scenarios for them.

rbren:
@openhands fix the merge conflicts

rbren:
> Out of curiosity, why KV as opposed to simply strings? 

Good q. I thought about it some more--KV is strictly more powerful, since you can recreate strings by setting V to "" or "true". So it's a bit more future-proof.

One use case: I might want to tag convos with `project: sdk` vs `project: gui`. or `owner: work` vs `owner: personal`

Issue #1349, opened 2025-12-08
[BUG] Agent produces MessageEvent without ActionEvent, causing premature execution termination


simonrosenberg:
# [BUG] Agent produces MessageEvent without ActionEvent, causing premature execution termination

## Summary

Agent execution incorrectly terminates when the agent produces a `MessageEvent` with reasoning/thinking content but fails to follow it with an `ActionEvent`. The system marks the execution as "finished" even though:
- The agent never called the finish tool
- Max iterations not reached (only 66/500 used)
- No timeout or error occurred
- Agent's reasoning clearly shows intent to continue with next steps

## Impact

**Severity**: P0 - Critical  
**Affects**: GAIA evaluation success rate (observed 1/3 failures due to this bug)  
**User Impact**: Evaluation results show 66.7% success rate when true capability is likely 90-100%

### Evaluation Run Details
- **Run ID**: 20026595169
- **Model**: claude-sonnet-4-5-20250929 via litellm_proxy
- **Failed Instance ID**: c61d22de-5f6c-4958-a7f6-5e9707bd3466
- **Max iterations**: 500 (only 66 used before termination)
- **Execution time**: 7.5 minutes (11:34:06 - 11:41:47)

### Artifact Locations
**GCS Bucket**: `gs://openhands-evaluation-results/`  
**Artifact File**: `sdk-main-20026595169-claude-son_litellm_proxy-claude-sonnet-4-5-20250929_25-12-08-11-43.tar.gz`  
**Conversation Archive**: `conversations/c61d22de-5f6c-4958-a7f6-5e9707bd3466.tar.gz` (within artifact)

**Direct GCS URLs**:
```
gs://openhands-evaluation-results/sdk-main-20026595169-claude-son_litellm_proxy-claude-sonnet-4-5-20250929_25-12-08-11-43.tar.gz
```

### Event Sequence (Buggy Behavior)

The conversation trajectory shows this pattern at termination:

```
Event 201: MessageEvent (kind: "MessageEvent", source: "agent")
  - content: "Let me search for all papers with arXiv ID 1608.03* that might contain these keywords:"
  - reasoning_content: Shows agent planning next steps (see details below)
  - thinking_blocks: Contains extended reasoning about search strategy

Event 202: ConversationStateUpdateEvent (kind: "ConversationStateUpdateEvent", source: "environment")
  - key: "execution_status"
  - value: "finished" ← INCORRECT!

[...]

PR #2470, opened 2026-03-16
Alternate fix for conversations that stop responding due to stale execution status 

DoubleDensity:
… RUNNING reset already happens outside the loop at lines 586-592, so we shouldn't check for STUCK again inside the loop and break.

## Summary

## Checklist


DoubleDensity:
also did not fix the issue after further testing -- please do not merge yet

DoubleDensity:
this one looks promising, was able to resume a previously hung `base_state.json` at least. will continue testing to see if any stuck sessions occur again

DoubleDensity:

xingyaoww:

xingyaoww:
## 🔴 **Needs improvement** — Patching symptoms, not the disease

### Linus-Style Analysis

**[CRITICAL ISSUES]**

**1. The author already told us it doesn't work**

> "also did not fix the issue after further testing -- please do not merge yet"

Let me repeat that: *the person who wrote this PR says it does not fix the problem.* Why are we reviewing code that the author has disowned? This should be closed or converted to a draft, not reviewed.

**2. [local_conversation.py, Lines 637-656] `last_event_count` tracking is a band-aid on a band-aid**

The original code's `continue` → STUCK check → `break` two-step dance is admittedly ugly. Replacing it with a direct `break` is fine — that's the one good idea here. But then you bolt on `last_event_count` tracking in *three separate places* (line 595, ~654, ~671) to prevent `is_stuck()` from being called "too often." This is solving an imaginary problem while the real one walks past you waving a flag.

The `is_stuck()` detector already has its own thresholds and only scans the last 20 events. Calling it once per loop iteration is cheap. The real question you should be asking is: **why does the conversation stop responding after stuck detection?** Adding event-count guards doesn't answer that.

**3. [local_conversation.py, Lines 646-656] The `break` creates a new problem**

The original flow was: detect stuck → set STUCK → `continue` → next iteration checks STUCK → `break`. Your fix: detect stuck → set STUCK → `break`. Same outcome, cleaner path — good.

But here's what you missed: after `run()` exits via `break`, **nothing resets the status to IDLE**. Look at the flow:
1. Stuck detected → status = STUCK → break out of loop
2. `run()` returns
3. User sends new message via `send_message()` → sees STUCK → resets to IDLE ✓
4. User calls `run()` again → sees STUCK at line 586-592 → resets to RUNNING ✓

OK, that part is actually fine. But the *original* code had the same path (continue → break), so **your change to `break` is functionally identical to the original for stuck detection.** The only behavioral difference is the `last_event_count` guard, which *skips* stuck detection when no new events appear. If the agent is spinning without producing events (e.g., an empty step), you'll never detect stuck. That's a regression.

**4. [local_conversation.py, Line 595 / Line 654 / Line 671] Three-way state tracking is a complexity smell**

`last_event_count` is updated at initialization, after stuck check, and after `agent.step()`. That's three places maintaining the same bookkeeping. When the next person touches this loop, they'll need to reason about all three to understand when stuck detection fires. This violates the "if you need more than 3 levels of indentation, you're screwed" principle — except here it's "if you need a mental flowchart to track one variable, you're screwed."

**5. [PR description] "[fill in a summary of this PR]" — Really?**

The summary is literally a placeholder. This tells me the PR was submitted in a rush. If you can't explain what your change does in one paragraph, you probably don't fully understand what it does yet.

**[TESTING GAPS]**

**6. Zero tests for a change to the core conversation loop**

This modifies the central run loop — the beating heart of conversation execution. The stuck detection path, the break/continue behavior, the interaction between `run()` and `send_message()` across threads — none of it is covered by a new test. For a bug fix that the author says "did not fix the issue," the absence of a reproducer test is damning. How do you know when you've actually fixed it if you can't reproduce it?

**7. [PR description] No evidence section**

No reproduction steps, no logs showing the stuck behavior, no before/after comparison. We're flying blind.

---

**VERDICT:**
❌ **Needs rework** — The `continue` → `break` simplification is the right instinct, but everything else is misguided complexity. The `last_event_count` optimization solves a non-problem while introducing a potential regression (skipping stuck detection when step produces no events). Most critically, the author confirms this doesn't fix the actual issue.

**KEY INSIGHT:**
The real bug is likely a concurrency issue between `run()` and `send_message()` around the state lock, not the `continue` vs `break` choice in stuck detection. Investigate what happens to the state lock when `send_message()` fires between stuck detection and the loop restart — that's where conversations go to die.

xingyaoww:
@OpenHands based on this https://github.com/OpenHands/software-agent-sdk/pull/2470#issuecomment-4083668419 can you do an investigation on the actual cause of this issue?

xingyaoww:
## Investigation: Root Cause Analysis for "Conversations That Stop Responding"

After a thorough investigation of the codebase, I analyzed the run loop, `send_message()`, `StuckDetector`, `FIFOLock`, the `ConversationState` autosave mechanism, and the agent-server's `EventService` layer. Here are my findings:

---

### TL;DR

The PR's changes to stuck detection (`continue`→`break`, event count tracking) are addressing **symptoms, not root causes**. The stuck detection flow (`continue` → reacquire lock → check STUCK → `break`) is functionally equivalent to a direct `break`. The real issues lie in several concurrency and design problems across the SDK and agent-server layers.

---

### Finding 1: TOCTOU Race in `event_service.send_message()` (Agent-Server)

**File:** `openhands-agent-server/openhands/agent_server/event_service.py`, lines 308-331

```python
async def send_message(self, message: Message, run: bool = False):
    # Step A: Add message (blocking executor call)
    await loop.run_in_executor(None, self._conversation.send_message, message)
    # Step B: Check status (acquires lock briefly)
    if run:
        with self._conversation.state as state:
            run = state.execution_status != ConversationExecutionStatus.RUNNING
    # Step C: Start run if needed
    if run:
        loop.create_task(_run_with_error_handling())
```

There is a **time-of-check-to-time-of-use gap** between Step A completing and Step B checking status. During this gap:
- The existing run loop could have already picked up AND processed the new message
- The run loop could have exited (status changed from RUNNING → FINISHED)
- Step B then sees a non-RUNNING status and starts a **redundant** run via Step C

While this typically results in harmless no-op runs (the new `run()` sees FINISHED and immediately breaks), it represents fragile logic that could cause unexpected behavior under specific timing conditions.

### Finding 2: Untracked Fire-and-Forget Run Tasks (Agent-Server)

**File:** `event_service.py`, lines 325-331

```python
# send_message creates untracked tasks:
loop.create_task(_run_with_error_handling())  # NOT stored in self._run_task

# But run() checks _run_task to prevent double-running:
if self._run_task is not None and not self._run_task.done():
```

`send_message()` creates fire-and-forget run tasks that are NOT stored in `_run_task`. This means `event_service.run()` cannot detect runs started by `send_message()`, potentially leading to **two concurrent `conversation.run()` calls**. The FIFO lock serializes them, so they don't truly run in parallel, but the second run may encounter unexpected state after the first completes.

### Finding 3: Exception Handler Outside Lock (SDK)

**File:** `local_conversation.py`, lines 690-705

```python
        except Exception as e:
            self._state.execution_status = ConversationExecutionStatus.ERROR  # ← No lock!
            self._on_event(ConversationErrorEvent(...))
```

The exception handler in `run()` sets `execution_status = ERROR` **outside** any `with self._state:` block. If another thread (e.g., `send_message()`) is concurrently modifying state, this creates a race condition on the status field. The autosave in `__setattr__` would persist whichever write happens last, potentially losing the ERROR status.

### Finding 4: Non-Atomic File Writes (SDK)

**File:** `io/local.py`, lines 62-68

```python
def write(self, path: str, contents: str | bytes) -> None:
    with open(full_path, "w", encoding="utf-8") as f:
        f.write(contents)
```

`_save_base_state()` writes `base_state.json` non-atomically. If the process crashes mid-write (which is plausible during long-running conversations), the file could be partially written, leading to corrupted JSON that fails to load on restart. This would manifest as a load error rather than "stale status," but is a robustness issue.

### Finding 5: Status Check Blocks the Event Loop (Agent-Server)

**File:** `event_service.py`, lines 313-315

```python
if run:
    with self._conversation.state as state:  # ← Sync lock on async thread!
        run = state.execution_status != ConversationExecutionStatus.RUNNING
```

This acquires the **synchronous FIFO lock directly on the asyncio event loop thread**. If the run loop is holding the lock during `agent.step()` (which includes LLM API calls potentially taking 10-30+ seconds), the **entire event loop blocks**. This means:
- No WebSocket messages can be processed
- No HTTP responses can be sent
- No other coroutines can execute

This is likely the most impactful contributor to the "stop responding" perception — the server literally cannot respond to any requests while the lock is held during `agent.step()`.

### Finding 6: Related Issues - False Positive Stuck Detection

Issues [#2482](https://github.com/OpenHands/software-agent-sdk/issues/2482) and [#1349](https://github.com/OpenHands/software-agent-sdk/issues/1349) describe scenarios where:
- Extended thinking models produce reasoning-only responses → emitted as `MessageEvent` → 3 consecutive agent MessageEvents triggers the monologue stuck detector → status becomes STUCK
- LLM text responses without tool calls set FINISHED prematurely

These could compound with the above concurrency issues to create the "stop responding" behavior.

---

### Recommendations (Priority Order)

1. **Move the status check in `event_service.send_message()` to a thread** (Finding 5) — or better, make the run decision inside `conversation.send_message()` itself while the lock is already held, avoiding the TOCTOU gap entirely (Findings 1 & 5).

2. **Track all run tasks** (Finding 2) — `send_message()` should store its run task in `_run_task` (or a shared tracking mechanism) so `run()` can detect and prevent double-running.

3. **Move the exception handler inside the lock** (Finding 3):
   ```python
   except Exception as e:
       with self._state:
           self._state.execution_status = ConversationExecutionStatus.ERROR
   ```

4. **Use atomic writes for base_state.json** (Finding 4) — write to a temp file then rename (which is atomic on most filesystems).

5. **Address #2482** — Fix the monologue false positive, which is a contributor to unnecessary STUCK states.

---

### Why This PR Doesn't Fix the Issue

The PR correctly identifies that `continue` after stuck detection creates an unnecessary round-trip through the loop. But changing `continue` to `break` is functionally equivalent (the original `continue` → reacquire lock → STUCK check → `break` path produces the same result). The `last_event_count` tracking adds complexity without addressing the real problems listed above.

The root cause of "conversations that stop responding" is most likely **Finding 5** (synchronous lock blocking the event loop), possibly combined with **Finding 2** (untracked run tasks) and **Finding 6** (false positive stuck detection). These are architectural issues in the agent-server layer, not in the stuck detection logic.

DoubleDensity:
Just adding that this fix has held strong under continuous use for 48 hours, have not been able to reproduce the original issue anymore.

csmith49 (changes requested):
Looks good to me, modulo the event count tracking. I can't see a situation where it actually modifies the control-flow, and worry it will act as a red herring when resolving future issues. Will approve this if it's removed or if we can figure out that it's necessary for this fix.

simonrosenberg:
> Just adding that this fix has held strong under continuous use for 48 hours, have not been able to reproduce the original issue anymore.

Do you have a script that runs the agent a few hundred times so that we can reproduce the problem?
If so, can you validate that the event count tracking is indeed necessary to fix the issue, that the stuck part itself is not enough? Thanks!

Edit: if it's a burden to run that many runs with a personal key, I can always take over :) let me know!

DoubleDensity:
thanks for taking a look at this @simonrosenberg !  I don't have a script or environment to automated test this in, I can only give anecdotal confirmation at the moment. I have been using OpenHands at work these past few weeks, but having limited success because of this issue breaking my conversations. Since I have put this patch in I have been using it non-stop with Nemotron 3 Super 120B on a DGX Spark w/ vLLM and OpenHands has not locked up again once, that was about three days ago now.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -634,15 +635,25 @@ def run(self) -> None:
                         break
 
                     # Check for stuck patterns if enabled
+                    # Only check if new events have been added since last check
+                    # to avoid repeatedly triggering on the same pattern
```
enyst, line 639: Do you know by any chance what was the pattern? Or which of the `stuck` scenarios got triggered?

enyst:
@DoubleDensity Thank you for looking into this! I think we fixed just today the cases where Qwen had tool calls in reasoning content, which caused it to the get `stuck` and fail a lot. I don't know if if the cause could be similar to what is causing this, just so you know it was described [in this comment](https://github.com/OpenHands/software-agent-sdk/pull/2505#issuecomment-4093882331) and related investigations and PR.

I wonder if you could try running on an updated `main`?
PR:
- https://github.com/OpenHands/software-agent-sdk/pull/2512

The problem there was that the SDK interpreted the messages with empty content and empty tool calls as empty messages (since content and tool calls were 'hidden' in reasoning where we don't read them), and then 3 empty messages trigger the STUCK detector.

DoubleDensity:
thanks @enyst I'll test it out!

DoubleDensity:
hey @enyst ! I tried running https://github.com/OpenHands/software-agent-sdk/pull/2512 for a couple days to see it was a better fix, and it looked promising. Went ~30 hours with no issues, but now has become stuck:

```
~/.openhands/conversations/129e7f493c6f499c881902714f3398e5$ cat base_state.json | jq '.execution_status'
"stuck"
```

I will make a backup of that conversation in case there is any useful metadata you want to extract from it

DoubleDensity:
it does seem to at least partially fix the problem; I was able to ask OpenHands to continue after it became stuck, and was able to resume successfully. so https://github.com/OpenHands/software-agent-sdk/pull/2512 is definitely an improvement, but it can still stop before reaching completion

DoubleDensity:
with https://github.com/OpenHands/software-agent-sdk/pull/2512 it resumed, but then stopped again after only 4 turns. not sure how to explain it, but not ideal

enyst:
What was the reason for stopping again and did the LLM actually repeat itself?

DoubleDensity:
closing this as stale, I believe it has been addressed elsewhere in the latest release

PR #2472, opened 2026-03-16
feat(security): defense-in-depth security analyzers

Fieldnote-Echo:
## Summary

Promotes the defense-in-depth security analyzer from a standalone example into first-class SDK analyzers under `openhands.sdk.security`.

This lands a deterministic, local, composable security analyzer family in the SDK. It does not change default wiring or claim universal safety as a default-on control.

### Public analyzers

- **`PatternSecurityAnalyzer`** — regex-based threat detection with two-corpus scanning. Shell-destructive patterns scan executable fields only; injection patterns scan all fields including reasoning text. Lives in `openhands.sdk.security.defense_in_depth`.
- **`PolicyRailSecurityAnalyzer`** — deterministic segment-aware rails for composed threats: fetch-to-exec, raw-disk-op, catastrophic-delete. Lives in `openhands.sdk.security.defense_in_depth`.
- **`EnsembleSecurityAnalyzer`** — pure combiner with max-severity fusion. No detection logic of its own. UNKNOWN preserved as first-class. Lives at `openhands.sdk.security.ensemble` (top-level, composes any analyzers — not limited to defense-in-depth siblings).

All three are importable from `openhands.sdk.security`.

### Key changes

- `SecurityRisk` comparison methods (`__lt__`, `__gt__`, `__le__`, `__ge__`) explicitly defined with shared `_RISK_ORDER` dict so `max()` works naturally on risk levels. `@total_ordering` cannot help here due to `str, Enum` MRO — documented in code.
- `EnsembleSecurityAnalyzer` moved to top-level `openhands.sdk.security.ensemble` (per reviewer feedback — it composes analyzers generally, not just defense-in-depth siblings)
- `propagate_unknown: bool = False` added to ensemble — default preserves current semantics; `True` makes any child UNKNOWN propagate immediately for stricter environments
- `RailDecision` simplified: `RailOutcome` enum removed, `outcome` is now `SecurityRisk` directly (HIGH/LOW). `reason` field preserved for observability.
- Rails docstring clarifies why they are separate from patterns: interpretability and maintainability of composed conditions, not regex capability
- Ensemble is a pure combiner — no `enable_policy_rails`, no extraction, no normalization
- Rails are their own analyzer (`PolicyRailSecurityAnalyzer`)
- Two-corpus invariant preserved: shell patterns never scan reasoning text
- Normalization expanded with broader invisible character set (same defensive category, more complete)
- Stable detector/rule IDs as string constants (15 `DET_*` + 3 `RAIL_*`)
- Serialization round-trip tests follow SDK convention (direct, polymorphic, container-field, roundtrip-then-detect, nested ensemble, `propagate_unknown` survives round-trip)
- Example thinned from 1002 lines to ~44-line SDK import demonstration (renumbered to `47_`)
- Old `conftest.py` importlib hack deleted
- Accidental `docs-draft-defense-in-depth.mdx` removed

### What this does NOT do

- No new dependencies
- No changes to `ConversationState`, `VerificationSettings`, or agent wiring
- No default-on behavior — additive only, users opt in
- No homoglyph/confusable replacement (deferred)
- No credential-path, permission-change, or encoded-payload heuristics (deferred)

### Test results

218 passed, 5 strict xfailed, 0 failures. All pre-commit hooks pass (ruff, pyright, import rules).

Companion docs PR: OpenHands/docs#402

## Test plan

- [x] `uv run pytest tests/sdk/security/ -v` — 218 pass, 5 xfail
- [x] Serialization round-trips verified (including `propagate_unknown=True`)
- [x] Two-corpus invariant tested
- [x] UNKNOWN handling tested (never promoted, never collapsed; `propagate_unknown` flag matrix covered)
- [x] Stable detector/rule IDs structurally asserted
- [x] No Dependabot alerts introduced


enyst:
@openhands Do a /codereview on this PR. Investigate deeply the relevant code in the codebase.

Fieldnote-Echo:
Thank you @enyst for the feedback. I am drafting fixes and will push again when they are ready.

Fieldnote-Echo:
@enyst Thanks OpenHands for the thorough review. Both findings are fixed in c398cf2: extraction now splits into executable and text corpora so shell patterns never see reasoning content, and the dd rail matches of=/dev/ regardless of operand order.

Fieldnote-Echo:
Companion docs PR: OpenHands/docs#402

examples/01_standalone_sdk/45_defense_in_depth_security.py
```diff
@@ -0,0 +1,1002 @@
...
+            return RailDecision(
+                RailOutcome.CONFIRM,
+                "world-writable",
+                "World-writable permissions (chmod 777, not sticky-bit 1777)",
+            )
+
+    return _PASS
+
+
+def _evaluate_rail(content: str) -> RailDecision:
```
csmith49: I'm a little confused on the distinction between this function and the `PatternSecurityAnalyzer`. Could we just map the `has_*` fields in the `_evaluate_rail_segments` function to a pattern and get the same result?

examples/01_standalone_sdk/45_defense_in_depth_security.py
```diff
@@ -0,0 +1,1002 @@
...
+    1. **Rails first** (if enabled) -- deterministic segment-aware rules
+       short-circuit to HIGH before any pattern scanning. Fast, no false
+       negatives on their covered threats.
+    2. **Collect analyzer results** -- each sub-analyzer evaluates the action
+       independently. Exceptions -> HIGH (fail-closed, logged).
+    3. **Fuse via max-severity** -- partition results into concrete
+       {LOW, MEDIUM, HIGH} and UNKNOWN. If any concrete result exists,
+       return the highest. If ALL are UNKNOWN, propagate UNKNOWN.
+
+    Why max-severity instead of noisy-OR: the analyzers are correlated
```
csmith49: The docs here present max-severity as an alternative to noisy-OR but I'm not sure where the latter is actually defined. Was this part of an earlier design iteration or do we use it elsewhere in the SDK?

examples/01_standalone_sdk/45_defense_in_depth_security.py
```diff
@@ -0,0 +1,1002 @@
...
+        default=True,
+        description="Evaluate deterministic policy rails before pattern scan",
+    )
+
+    def security_risk(self, action: ActionEvent) -> SecurityRisk:
+        """Evaluate risk via rails + max-severity fusion."""
+        # Step 1-2: Policy rails (on executable-field segments only)
+        # Rails detect shell-level threats; reasoning text would cause
+        # false positives (e.g. thought "avoid rm -rf" on a safe command).
+        if self.enable_policy_rails:
```
csmith49: Related to my question on `PatternSecurityAnalyzer` and `_evaluate_rail_segments`: the rails get converted directly to `SecurityRisk.HIGH`, which may not be needed if those rails are just expressed as patterns.

One alternative: keep the `EnsembleSecurityAnalyzer` a true ensemble that doesn't add any other logic except combining other security analyzers, and move this `enable_policy_rails`-guarded logic to a more primitive security analyzer (either as patterns or a separate `PolicyRailSecurityAnalyzer`).

examples/01_standalone_sdk/45_defense_in_depth_security.py
```diff
@@ -0,0 +1,1002 @@
...
+    def security_risk(self, action: ActionEvent) -> SecurityRisk:  # noqa: ARG002
+        return self.fixed_risk
+
+
+# ---------------------------------------------------------------------------
+# EnsembleSecurityAnalyzer
+# ---------------------------------------------------------------------------
+
+# Severity ordering for concrete (non-UNKNOWN) risk levels.
+_SEVERITY_ORDER = {SecurityRisk.LOW: 0, SecurityRisk.MEDIUM: 1, SecurityRisk.HIGH: 2}
```
csmith49: I understand you're probably trying to keep everything self-contained in this example. If we end up wanting to move this to the SDK proper this kind of logic should probably live in an overloaded `SecurityRisk.__lt__` method, which would let us replace the

```python
max(concrete, key=lambda r: _SEVERITY_ORDER[r])
```

logic on line 778 with just

```python
max(concrete)
```

Fieldnote-Echo:
@csmith49 Thank you Calvin for the substantive feedback. It seems the requested changes only make sense if this is moved into the SDK proper, so I will wait for feedback from @xingyaoww before pushing any new commits.

xingyaoww:
> prefer would be to move the relevant security analyzers to the SDK proper and just have the example surface how to use/configure them. They're general enough and useful enough that people building on the SDK might want to use them without the dynamic module loading tricks relied upon in the examples.

Agree with this! We could potentially just come up with a new type of SecurityAnalyzer instead of dumping all the logic in the example script.

I'd love to get this actually into SDK, maybe even as a security analyzer that always run since it doesn't seems to consume too much resource 🙏

Fieldnote-Echo:
@xingyaoww @csmith49 Solid direction, thank you. I'll refactor it into a new SecurityAnalyzer type in the SDK. Before I push anything, I'll study the existing analyzer module structure to make sure it fits the current patterns. Any preferences on where it should live under openhands_aci/?

csmith49:
I expect most everything can live as a sub-module in `openhands.sdk.security`, much like the `grayswan` security analyzer we already have. But your contribution will represent a significant portion of that module, so do what makes sense for this approach!

Fieldnote-Echo:
@all-hands-bot still working, preparing offline before sending any pushes. If it's being considered for the SDK proper I want to make sure it earns the spot.

Fieldnote-Echo:
@csmith49 Quick update: The OpenHands team can expect this PR to be updated over the next few days, but I'll specifically call out when its "reviewer-ready" - thanks!

Fieldnote-Echo:
@xingyaoww Quick scope check: I’m attempting to keep this PR tight around the SDK analyzer refactor plus correctness/serialization coverage. Is there interest in GraySwan coverage convergence/divergence in this PR, or should I leave that for a follow-up after, or something else?

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Rails are now their own analyzer (`PolicyRailSecurityAnalyzer`) — they handle composed conditions (both tokens in the same segment) that plain regex patterns can't express safely.

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Removed the noisy-OR reference — the new docs explain max-severity directly without the comparison.

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Done — ensemble is now a pure combiner with zero detection logic, and rails live in `PolicyRailSecurityAnalyzer` as a separate analyzer.

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Added `SecurityRisk.__lt__` (and `__gt__` for MRO reasons) — the ensemble now uses `max(concrete)` directly.

Fieldnote-Echo:
@csmith49 All four review comments are addressed in the latest push. The analyzers now live in `openhands.sdk.security.defense_in_depth` as SDK proper - ensemble is a pure combiner, rails are their own analyzer, and `SecurityRisk.__lt__` enables natural `max()` ordering. Ready for re-review when you have a chance.

enyst:
Just to note, the docs and review gate in CI are not mandatory, please don't be concerned about them - though the docs PR is most welcome of course!

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Verified: `PolicyRailSecurityAnalyzer` now exists as a separate analyzer in `openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py`. The distinction is clear - rails handle composed conditions that require both tokens in the same segment, while patterns are single-signature regex matches.

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Verified: The docs now explain max-severity directly (lines 108-110 of docs-draft-defense-in-depth.mdx) without any noisy-OR comparison.

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Verified: `EnsembleSecurityAnalyzer` in `ensemble.py` is now a pure combiner with zero detection logic (no `enable_policy_rails`, no extraction/normalization). Rails live in `PolicyRailSecurityAnalyzer` as a separate first-class analyzer.

examples/01_standalone_sdk/45_defense_in_depth_security.py
Fieldnote-Echo: Verified: `SecurityRisk.__lt__` and `__gt__` are implemented in `risk.py` (lines 102-132). The ensemble now uses `max(concrete)` directly (line 82 of `ensemble.py`) without any helper dict.

Fieldnote-Echo:
@xingyaoww One rollout note for later discussion, not for this PR: if the team eventually wants broader/default use, I think the main follow-ups are observability of why something was flagged and some real-world false-positive tuning on coding workflows. I kept those out of this PR so the SDK promotion stays tight, but I’m happy to open follow-up issues once this lands. These are more product decisions requiring OpenHands judgement and deployment experience specifically.

Fieldnote-Echo:
@enyst Okay team, I believe this and the docs companion PR are "review-ready" - Docs PR: https://github.com/OpenHands/docs/pull/402

openhands-sdk/openhands/sdk/security/risk.py
```diff
@@ -98,3 +98,35 @@ def is_riskier(self, other: SecurityRisk, reflexive: bool = True) -> bool:
...
+        if self == SecurityRisk.UNKNOWN or other == SecurityRisk.UNKNOWN:
+            raise ValueError("Cannot compare unknown risk levels.")
+        risk_order = {
+            SecurityRisk.LOW: 1,
+            SecurityRisk.MEDIUM: 2,
+            SecurityRisk.HIGH: 3,
+        }
+        return risk_order[self] < risk_order[other]
+
+    def __gt__(self, other: object) -> bool:
```
csmith49, line 125: Minor suggestion: we can avoid repeating ourselves in ordering defs by using the `functools.total_ordering` class decorator.

openhands-sdk/openhands/sdk/security/ensemble.py
csmith49, line 1: This analyzer probably deserves to be in the top-level of `openhands.sdk.security`. When we make these rule-based checks part of the default implementation (future PR) we'll also be combining it with the other security analyzers, not just the ones from the `defense-in-depth` sub-module.

openhands-sdk/openhands/sdk/security/defense_in_depth/ensemble.py
```diff
@@ -0,0 +1,82 @@
...
+    Use this as the top-level analyzer you set on a conversation. It
+    calls each child analyzer, collects their risk assessments, and
+    returns the highest concrete risk. It does not perform any detection,
+    extraction, or normalization of its own.
+
+    How UNKNOWN works: if *all* children return UNKNOWN, the ensemble
+    returns UNKNOWN (which ``ConfirmRisky`` confirms by default). If any
+    child returns a concrete level, UNKNOWN results are filtered out and
+    the highest concrete level wins. UNKNOWN is never passed to ``max()``
+    -- that would raise ``ValueError`` by design.
```
csmith49: It's not obvious to me that this behavior is correct. A more defensive pattern might be returning `UNKNOWN` if _any_ child returns `UNKNOWN`. WDYT? Maybe we can switch the interpretation with a flag?

openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py
```diff
@@ -0,0 +1,188 @@
...
+# ---------------------------------------------------------------------------
+# Rail types
+# ---------------------------------------------------------------------------
+
+
+class RailOutcome(Enum):
+    """Internal policy recommendation from deterministic rail evaluation.
+
+    DENY and CONFIRM both map to SecurityRisk.HIGH at the SDK boundary.
+    The distinction is preserved internally for observability.
```
csmith49: Suggestion: since there's only `DENY` and `PASS` used, consider simplifying the `RailDecision` class to use `outcome: SecurityRisk` instead. We'll still get more fine-grained observability with the `reason` field.

openhands-sdk/openhands/sdk/security/defense_in_depth/__init__.py
csmith49, line 1: Suggestion: if we move `ensemble` up a level this module name doesn't quite fit. Consider changing to something like `patterns` or `rails`.

openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py
```diff
@@ -0,0 +1,188 @@
...
+    return _evaluate_rail_segments([_normalize(content)])
+
+
+# ---------------------------------------------------------------------------
+# PolicyRailSecurityAnalyzer
+# ---------------------------------------------------------------------------
+
+
+class PolicyRailSecurityAnalyzer(SecurityAnalyzerBase):
+    """Catch composed threats that plain regex signatures would miss.
```
csmith49, line 149: _Technically_, the rails defined here can all be expressed by single regexes. I don't think that's a valid reason to collapse the `PolicyRailSecurityAnalyzer` with the `PatternSecurityAnalyzer` -- makes certain patterns way more interpretable and maintainable -- but maybe worth updating the doc-string to clarify the benefits of this separate analyzer.

docs-draft-defense-in-depth.mdx
csmith49, line 1: Is this file meant to be committed here?

csmith49:
Hey @Fieldnote-Echo, this is coming along great. I left some comments (mostly suggestions), once you get a chance to look through those and see if they're relevant we can get this merged. Excited to get onto the observability/tuning so we can make these default as well!

docs-draft-defense-in-depth.mdx
Fieldnote-Echo: @csmith49 Good catch, removed.

openhands-sdk/openhands/sdk/security/risk.py
Fieldnote-Echo: @csmith49 Extracted a shared _RISK_ORDER dict and _check_comparable() helper instead to get the same deduplication.

openhands-sdk/openhands/sdk/security/ensemble.py
Fieldnote-Echo: @csmith49 Great idea - it now lives at openhands.sdk.security.ensemble, public API unchanged.

openhands-sdk/openhands/sdk/security/defense_in_depth/ensemble.py
Fieldnote-Echo: @csmith49 Added propagate_unknown: bool = False - default behavior unchanged, True makes any child UNKNOWN propagate immediately.

openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py
Fieldnote-Echo: @csmith49 Dropped RailOutcome entirely, outcome is now SecurityRisk directly, reason still there for observability.

openhands-sdk/openhands/sdk/security/defense_in_depth/__init__.py
Fieldnote-Echo: @csmith49 Moved ensemble up but kept defense_in_depth for now - patterns and rails both under-describe what remains (patterns + rails + shared utils).  Would a rename be warranted?

openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py
Fieldnote-Echo: @csmith49 Updated for clarity - the separation is about interpretability of composed conditions, not regex capability.

Fieldnote-Echo:
@csmith49 Thank you for the polish and suggestions! I left one convo open in case you'd like a rename for clarity. Other than that, everything else is resolved.

openhands-sdk/openhands/sdk/security/defense_in_depth/__init__.py
csmith49: I think we probably could rename for clarity. To me, `defense-in-depth` only implies that there are layers of security, but says nothing about the kind of security. But we can punt on this for now and handle a rename when we fold it in to our default analyzers. Maybe a clearer name will emerge by then.

PR #2474, opened 2026-03-17
feat(tools): add codebase_search and github_codebase_search tools

DhruvBhatia0:
## Summary

- Register Morph's WarpGrep as native OpenHands tools via the `@morphllm/morphmcp` MCP server
- Users add `Tool(name="codebase_search")` to their agent like any other tool — MCP plumbing is internal
- Two separate tool registrations (`codebase_search` + `github_codebase_search`) sharing one MCP server process
- `edit_file` from the MCP server is filtered out to avoid conflict with `FileEditorTool`
- Clear `ValueError` with dashboard link when `MORPH_API_KEY` is missing
- Pre-installs `@morphllm/morphmcp` in Docker image for fast cold starts

### Usage

```python
from openhands.tools.codebase_search import register_codebase_search_tools

register_codebase_search_tools()

agent = Agent(
    llm=llm,
    tools=[
        Tool(name="codebase_search"),           # search local repos
        Tool(name="github_codebase_search"),    # search public GitHub repos
    ],
)
```

Requires `MORPH_API_KEY` env var (or passed via `Tool(name="codebase_search", params={"api_key": "sk-..."})`) and Node.js 18+.

## Test plan

- [x] Import and registration verified locally
- [x] Both tools appear in `list_registered_tools()`
- [x] Missing `MORPH_API_KEY` raises `ValueError` with instructions + dashboard link
- [x] Top-level `from openhands.tools import register_codebase_search_tools` works
- [ ] End-to-end test with `MORPH_API_KEY` set against a real repo
- [ ] Docker image build with `@morphllm/morphmcp` pre-installed

### DIFF
--- a/openhands-tools/openhands/tools/codebase_search/definition.py
+++ b/openhands-tools/openhands/tools/codebase_search/definition.py
@@ -1,0 +1,134 @@
+"""Codebase search tool registration backed by @morphllm/morphmcp.
+
+This module registers ``codebase_search`` and ``github_codebase_search`` as
+native OpenHands tools. Under the hood they are MCP tools served by the
+``@morphllm/morphmcp`` npm package (started via ``npx``).
+
+The ``edit_file`` tool exposed by the same MCP server is intentionally
+filtered out to avoid conflicts with the built-in ``FileEditorTool``.
+"""
+
+from __future__ import annotations
+
+import os
+import threading
+from collections.abc import Sequence
+
+from openhands.sdk.logger import get_logger
+from openhands.sdk.mcp.utils import create_mcp_tools
+from openhands.sdk.tool import ToolDefinition, register_tool
+
+logger = get_logger(__name__)
+
+# ── Shared MCP client cache ────────────────────────────────────────────
+# Both resolvers share a single MCP server process per API key so that
+# agents using both tools don't spawn two ``npx`` processes.
+_lock = threading.Lock()
+_morph_clients: dict[str, tuple[object, list[ToolDefinition]]] = {}
+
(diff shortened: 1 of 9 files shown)


VascoSch92:
Thanks for the PR.

Why a tool wrapping an mcp? Can we not just use the mcp directly?

DhruvBhatia0:
> Thanks for the PR.
> 
> Why a tool wrapping an mcp? Can we not just use the mcp directly?

ah good point, i can make this change,

VascoSch92:
If this is just an MCP server, which we already support, why do we need a new tool instead of using the MCP directly (i.e., passing it to the agent)? 

Moreover, we already have tools that allow the agent to explore the codebase. Why would this tool be better? Is there a related issue for this?

openhands-tools/openhands/tools/codebase_search/definition.py
```diff
@@ -1,50 +1,124 @@
...
-
-.. note::
-    Unlike most tools that subclass :class:`ToolDefinition` and implement
-    ``create()``, these tools use **callable resolver functions** registered
-    directly with :func:`register_tool`.  The registry invokes them as
-    ``factory(conv_state=conv_state, **params)``; see
-    :func:`_resolver_from_callable` in ``openhands.sdk.tool.registry``.
+Registers ``codebase_search`` and ``github_codebase_search`` as native
+OpenHands tools.  Each tool calls ``@morphllm/morphsdk`` via a small
+Node.js bridge script (``bridge.js``), which handles the multi-turn
```
enyst, line 5: I'd love to know the answer to Vasco's question, I'm not sure a bridge is necessary if it's MCP?

PR #1544, opened 2025-12-29
[agent-sdk-j2s] agent-server runtime LLM switching

enyst:
Implements runtime LLM switching end-to-end for remote clients (VS Code extension) without agent immutability/diff enforcement.

Key changes:
- Agent is componentized: restore-time + runtime swaps (LLM, tools, context) are allowed; no diff-based enforcement.
- Conversation persistence remains durable (base_state + events); LLM selection persists across restart; runtime secrets override persisted secrets on restore.
- Agent-server REST: POST /api/conversations/{id}/llm (profile_id or inline llm) and POST /api/conversations/{id}/llm/switch.
- Tests: agent-server wsproto integration covers large event history + restore + runtime LLM switching.

Docs: docs/llm_switching.md
Example: examples/02_remote_agent_server/07_llm_switch_and_restore.py

Docs: check-examples uses matching OpenHands/docs branches; docs PR OpenHands/docs#219 is optional follow-up for landing docs on main.
Beads: agent-sdk-j2s

PR #2485, opened 2026-03-17
design(agent-server): REST LLM profiles for conversation start and restore

enyst:
## Summary
- add a design document for wiring SDK LLM profiles into agent-server
- recommend exposing `/api/llm-profiles` as REST CRUD over the existing `LLMProfileStore` instead of reviving per-conversation `/llm`
- keep A1 non-redundant by using that same store for REST CRUD, conversation start/switch, and restore
- cover start, runtime switch, restore-time re-resolution, and RemoteConversation follow-up

## Context investigated
- PR #1544 and in particular `docs/llm_switching.md`
- issue #1451
- current SDK `LLMProfileStore`, `LocalConversation.switch_profile()`, and agent-server start/restore flow

## Artifact
- `.pr/design-llm-profiles-agent-server.md`

## Recommendation in brief
- expose `/api/llm-profiles` CRUD as a thin wrapper over the existing `LLMProfileStore`
- let conversation start accept `llm_profile_id`
- add a dedicated conversation profile switch endpoint
- restore from profile when available, with persisted snapshot fallback
- reuse that same store for REST CRUD, conversation start/switch, and restore

Refs #1451

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:08aad25-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-08aad25-python \
  ghcr.io/openhands/agent-server:08aad25-python
```

```
ghcr.io/openhands/agent-server:08aad25-golang-amd64
ghcr.io/openhands/agent-server:08aad25-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:08aad25-golang-arm64
ghcr.io/openhands/agent-server:08aad25-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:08aad25-java-amd64
ghcr.io/openhands/agent-server:08aad25-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:08aad25-java-arm64
ghcr.io/openhands/agent-server:08aad25-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:08aad25-python-amd64
ghcr.io/openhands/agent-server:08aad25-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:08aad25-python-arm64
ghcr.io/openhands/agent-server:08aad25-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:08aad25-golang
ghcr.io/openhands/agent-server:08aad25-java
ghcr.io/openhands/agent-server:08aad25-python
```

- Each variant tag (e.g., `08aad25-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `08aad25-python-amd64`) are also available if needed


.pr/design-llm-profiles-agent-server.md
```diff
@@ -0,0 +1,509 @@
...
+Add agent-server config:
+
+- `OH_LLM_PROFILES_PATH`
+- default: `workspace/llm_profiles`
+
+Why a server-specific path instead of relying on `~/.openhands/profiles`?
+
+- it keeps server data in one predictable place
+- it matches how conversations are already explicitly configured
+- it behaves better in containers and ephemeral runtimes
```
enyst: Downside:
- security, arguably. In a sense, `workspace/llm_profiles` is in `workspace` which is the agent’s workspace. The agent shouldn’t have access to LLM keys used to run itself, it’s not necessary and it could lead to leaks.

On the other hand, the agent has `bash` tool and the entire agent sandbox should be considered agent’s playground; it can access stuff in other locations in the sandbox, with some difficulty maybe, anyway, so that’s similar.

@openhands 3 questions:
1) look at the dockerfile and what files you need, and answer under what user is the server process running, and how is the agent’s terminal session running, and can we somehow “hide” it from the agent’s bash?
2) in agent-server, what is the default path of `conversations_dir` or so?
3) the start conversation request accepts a plugin as parameters. Follow the execution path of how it loads that plugin, what is the location where the plugin is saved when its source is an URL or when it’s local?

enyst: Looked through the current code path and the answer is basically: **today this is same-sandbox / same-UID access, so path choice alone does not meaningfully hide it from the agent**.

1. **Which user runs what? Can we hide it from bash?**
   - In the agent-server image, the runtime process runs as user `openhands` (see `openhands-agent-server/openhands/agent_server/docker/Dockerfile`: the base runtime stages create `openhands`, switch to `USER ${USERNAME}`, and the final `source` / `binary` targets inherit that).
   - The agent's terminal session is not a separate OS user today. `TerminalTool.create()` does not pass a username, so `TerminalExecutor` creates a terminal session with `username=None`. On Unix, the terminal factory prefers **tmux** if available; this image installs tmux, so the normal path is `TmuxTerminal`, which starts plain `/bin/bash` in the same user context. If tmux were unavailable, the subprocess fallback also just `Popen()`s interactive bash in the same process user context.
   - Both terminal backends inherit essentially the server process environment via `sanitized_env()`.
   - So: **no, not in any strong sense with the current architecture**. Moving profile files from `/workspace/...` to some other path inside the same container may make accidental discovery a bit less likely, but it does not create real isolation. A same-UID bash can still read any path its user can read. Real hiding would need a different isolation boundary (different UID, different mount, sidecar/service, etc.), not just a different directory.

2. **Default `conversations_dir` / conversations path**
   - In `openhands-agent-server/openhands/agent_server/config.py`, `Config.conversations_path` defaults to `Path("workspace/conversations")`.
   - In the published container image, the Dockerfile sets `WORKDIR /`, so that resolves in practice to **`/workspace/conversations`**.
   - Small nuance: outside the container image it is still a relative path, so it resolves against whatever cwd the server process was started from.

3. **Plugin load path and where it is saved**
   - Request path: `StartConversationRequest.plugins` -> `ConversationService.start_conversation()` -> copied unchanged into `StoredConversation` -> persisted in `<conversations_dir>/<conversation_id>/meta.json`.
   - Runtime load path: `EventService.start()` passes `stored.plugins` into `LocalConversation(...)` -> on first `run()` / `send_message()`, `LocalConversation._ensure_plugins_loaded()` calls `fetch_plugin_with_resolution()` for each plugin.
   - For **URL / git sources**, the plugin is cached under `Path.home() / '.openhands' / 'cache' / 'plugins' / <repo-name>-<hash>` (`openhands-sdk/openhands/sdk/plugin/fetch.py`). In this container, `Path.home()` for the server user is `/home/openhands`, so the effective cache root is **`/home/openhands/.openhands/cache/plugins/`**. If `repo_path` is set, the final loaded path is a subdirectory inside that cached repo.
   - For **local sources**, it is **not copied into a cache**. `fetch_plugin_with_resolution()` resolves it with `Path(source).expanduser().resolve()` and uses that path directly. So the answer for local plugins is: **no new saved location; it stays at the original local path**. If the caller passes a relative local path, that resolution is against the server process cwd, not the workspace root.

So for the security point in the design doc: I think the stronger statement is "putting profiles outside `workspace/` is cleaner operationally, but it does **not** by itself prevent agent access in the current single-user sandbox model."

.pr/design-llm-profiles-agent-server.md
```diff
@@ -0,0 +1,509 @@
...
+
+### 4. Extend conversation creation to accept a profile reference
+
+Add an optional field to both conversation contracts:
+
+- `StartConversationRequest.llm_profile_id: str | None = None`
+- `StartACPConversationRequest.llm_profile_id: str | None = None`
+
+Because `StoredConversation` inherits from the ACP request model, the field will
+also persist naturally in `meta.json`.
```
enyst: @openhands I am surprised by this sentence. Don’t we have two now, and StoredConversation inherits from… what? from the StartConversationRequest I thought, not the ACP variant. Verify what I just said. Feel free to contradict me, we care about the truth.

.pr/design-llm-profiles-agent-server.md
```diff
@@ -0,0 +1,509 @@
...
+
+Expose a dedicated profile-binding endpoint rather than a generic `/llm` setter.
+
+#### V1 contract
+
+- `POST /api/conversations/{conversation_id}/llm_profile`
+
+#### ACP contract parity
+
+- `POST /api/acp/conversations/{conversation_id}/llm_profile`
```
enyst: I think ACP execution flow doesn’t use a real LLM, so this would be a no-op. Verify what I just said. If so, let’s remove it.

enyst: @openhands

.pr/design-llm-profiles-agent-server.md
enyst: Verified in `openhands-agent-server/openhands/agent_server/models.py`: today there are **two** public start request models, `StartConversationRequest` and `StartACPConversationRequest`, and both inherit from `_StartConversationRequestBase`. `StoredConversation` does **not** inherit from `StartConversationRequest`; it currently inherits from `StartACPConversationRequest` (lines 154-169 in the current file). I updated the design doc in d764d52c to anchor `llm_profile_id` on the shared base model and to say explicitly that `StoredConversation` currently inherits from the ACP variant.

.pr/design-llm-profiles-agent-server.md
enyst: Verified in the code: `ACPAgent` uses a sentinel `LLM(model="acp-managed")` (`_make_dummy_llm()` in `openhands-sdk/openhands/sdk/agent/acp_agent.py`), and the actual ACP-side model selection happens through ACP-native config like `acp_model`, not `agent.llm`. So an ACP `/llm_profile` endpoint would indeed be misleading/no-op. Fixed in cdd97d41 by scoping `llm_profile_id` and the switch endpoint to standard conversations only, and by calling ACP out as a separate future design space.

.pr/design-llm-profiles-agent-server.md
enyst: @openhands wdyt about this:
- make the profile store default path be Path.home() / ‘.openhands’ / the existing subdir name for profiles.
- that should be true for server too 🤔

enyst: Good call — I agree, and I updated the design in 65a6a2e4. The SDK already defaults `LLMProfileStore` to `Path.home() / ".openhands" / "profiles"`, so using the same default in agent-server is the simpler choice. It keeps server behavior aligned with the SDK and with other `~/.openhands` data, while `OH_LLM_PROFILES_PATH` still covers deployments that want an explicit override. I also called out that moving the default under `workspace/` would not create meaningful isolation in the current same-UID sandbox model, so it is better treated as an operational path choice than a security boundary.

.pr/design-llm-profiles-agent-server.md
enyst: @openhands explain to me the inheritance here, cuz I still dont get it. And lets see pro and cons for where we add the profile field.

enyst: Good catch — I clarified this in 6fe6d817.

The inheritance today is:

```text
_StartConversationRequestBase
├── StartConversationRequest
└── StartACPConversationRequest
    └── StoredConversation
```

So there are **two public start request models**, both sharing the common base, and `StoredConversation` currently inherits from the **ACP** request model, not from `StartConversationRequest`.

The important nuance is *why* `StoredConversation` inherits that ACP model: not because stored conversations are ACP-only, but because `StartACPConversationRequest` already has the wider `ACPEnabledAgent` type, so `StoredConversation` can persist either a normal `Agent` or an `ACPAgent`.

That leads to the field-placement tradeoff:

- **Put `llm_profile_id` on `_StartConversationRequestBase` / ACP path**
  - pro: single declaration, and `StoredConversation` gets it automatically through inheritance
  - con: the ACP public API would now advertise an LLM-profile field even though ACP does not actually select its runtime model through `agent.llm`; that would be a misleading public contract

- **Put `llm_profile_id` on `StartConversationRequest` + `StoredConversation`**
  - pro: the public API stays honest that this is a standard-conversation feature, while still persisting the field for restore
  - con: small duplication, because we cannot rely on inheritance alone

I think the second option is cleaner: a tiny bit of duplication is cheaper than exposing a standard-only concept on the ACP contract and then having to support or deprecate that later.

.pr/design-llm-profiles-agent-server.md
```diff
@@ -0,0 +1,583 @@
...
+4. Make restore use the current profile contents when available.
+5. Avoid resending credentials for every conversation start/switch.
+6. Keep the REST rollout additive and backward compatible.
+7. Reuse the SDK's existing `LLMProfileStore` and `LocalConversation.switch_profile`
+   concepts rather than inventing a second model.
+
+## Non-goals
+
+1. Reworking the full conversation creation contract so `agent.llm` becomes optional.
+   That is a larger API cleanup.
```
enyst: @openhands First revert the last commit. Just revert it. Then see the diff and understand this PR.

This point 1. is correct, it is a larger API cleanup to rework conversation creation contract. Also note that we have a workflow on REST API, read it, where you see we enforce backwards compatibility, so we’d need a solution for that too. Also we’d need at a minimum to save profile_id in LLM class, so that we know at restore which profile was it if we ever need it.

IMPORTANT to clarify: what we want here is: Agent takes an LLM or a profile_id: str. We resolve at runtime the profile. This will affect REST, but it seems it will also affect regular agent serialization / deserialization, Python API contract, maybe Conversation.

I want to see the blast radius. Be careful, and double check all you say. Investigate all you need, you have time.

@openhands Make an alternative section with THAT alternative. Mark A1 the current alternative choice here, agent.llm unchanged, and mark A2 every detail that changes in A2.

enyst: Fixed in 1638f578. I added a dedicated creation-contract alternatives section with **A1** as the current recommendation (`agent.llm` unchanged) and **A2** spelling out the changed pieces: `LLM.profile_id`, restore/switch semantics, the REST-compat rollout needed under the REST API breakage checks, and the ACP/OpenAPI impact.

enyst: @openhands do this please.

enyst: Addressed by 1638f578. The doc now has a dedicated creation-contract alternatives section with A1 as the recommended additive path (agent.llm unchanged) and A2 covering the broader blast radius: LLM.profile_id, restore/switch semantics, REST compatibility rollout under the breakage checks, and ACP/OpenAPI impact.

enyst: Done. I first reverted the previous A2 write-up in 9b4cf68d, then reworked the design in 0d3d91c3 around the actual larger alternative from your comment: standard Agent takes either inline llm or llm_profile_id. The new A2 now treats this as an Agent-contract change, and documents the broader blast radius across Python API, Agent serialization/deserialization, Conversation/LocalConversation/RemoteConversation resolution points, restore semantics, REST/OpenAPI rollout, and ACP/model-hierarchy impact.

.pr/design-llm-profiles-agent-server.md
```diff
@@ -0,0 +1,700 @@
...
+     - the profile reference
+     - a resolved snapshot fallback
+   - so A2 is not automatically simpler on persistence
+
+4. **Conversation restore semantics**
+   - `ConversationState.create()` currently validates the persisted agent,
+     verifies tools, then replaces `state.agent` with the runtime agent
+   - if persisted state can contain an unresolved profile reference, local SDK
+     restore also needs a way to resolve it, not just agent-server restore
+   - that means the resolution logic cannot live only in the REST layer
```
enyst:
```suggestion
   - if persisted state can contain an unresolved profile reference, local SDK
     restore needs a way to resolve it, not just agent-server restore
   - that means the resolution logic must actually live in the local SDK layer
```

.pr/design-llm-profiles-agent-server.md
```diff
@@ -0,0 +1,700 @@
...
+## Proposed design (A1)
+
+### 1. Add a server-managed profile store
+
+Add agent-server config:
+
+- `OH_LLM_PROFILES_PATH`
+- default: `Path.home() / ".openhands" / "profiles"` (the existing SDK default)
+
+Why keep the same home-scoped default for server too?
```
enyst: Can we just NOT add a new one, since we have a profile store already?

Analyze where we need it and if we can make it available, alternatively, or not make it available but use its functionality via something else. @openhands keep the A1 design simple as necessary, not more redundant than necessary.

enyst: Addressed in 928c232d. I simplified A1 so it does **not** add a new server-side profile-store abstraction. The doc now treats the existing `LLMProfileStore` as the only store and narrows the required server touchpoints to the three places that actually need it: conversation start (`llm_profile_id` resolution), runtime switch, and restore-time re-resolution. I also moved `/api/llm-profiles`, cipher-aware profile writes, and `OH_LLM_PROFILES_PATH` into optional follow-up layers rather than prerequisites for A1, so the first slice stays additive and minimally redundant.

PR #2490, opened 2026-03-18
feat: add local_agent_server_mode to OpenHandsCloudWorkspace

xingyaoww:
## Summary

Add `local_agent_server_mode` to `OpenHandsCloudWorkspace` to support the automation service architecture described in [RFC #13275](https://github.com/OpenHands/OpenHands/issues/13275).

When `local_agent_server_mode=True`, the workspace assumes it is already running inside an OpenHands Cloud Runtime sandbox. Instead of creating or managing a sandbox via the Cloud API, it connects directly to the local agent-server at `http://localhost:<port>` (default port 60000). Cloud API credentials are still required for `get_llms()` / `get_secrets()` calls.

**New fields:**
- `local_agent_server_mode: bool = False` — enables local agent-server mode
- `agent_server_port: int = 60000` — configurable local port

**Behavioral changes:**
- `model_post_init` skips sandbox creation/wait in local agent-server mode
- `cleanup()` only closes the HTTP client in local agent-server mode (no sandbox deletion)
- `automation_callback_url` and `automation_run_id` are read from env vars (`AUTOMATION_CALLBACK_URL`, `AUTOMATION_RUN_ID`) rather than constructor args

**Usage:**
```python
workspace = OpenHandsCloudWorkspace(
    local_agent_server_mode=True,
    cloud_api_url="https://app.all-hands.dev",
    cloud_api_key=os.environ["OPENHANDS_API_KEY"],
)
```

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:33e1694-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-33e1694-python \
  ghcr.io/openhands/agent-server:33e1694-python
```

```
ghcr.io/openhands/agent-server:33e1694-golang-amd64
ghcr.io/openhands/agent-server:33e1694-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:33e1694-golang-arm64
ghcr.io/openhands/agent-server:33e1694-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:33e1694-java-amd64
ghcr.io/openhands/agent-server:33e1694-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:33e1694-java-arm64
ghcr.io/openhands/agent-server:33e1694-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:33e1694-python-amd64
ghcr.io/openhands/agent-server:33e1694-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:33e1694-python-arm64
ghcr.io/openhands/agent-server:33e1694-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:33e1694-golang
ghcr.io/openhands/agent-server:33e1694-java
ghcr.io/openhands/agent-server:33e1694-python
```

- Each variant tag (e.g., `33e1694-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `33e1694-python-amd64`) are also available if needed

### DIFF
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -14,8 +14,27 @@ Outputs to GITHUB_OUTPUT:
 
 import json
 import os
+import signal
 import sys
+import time
 from typing import Any
+
+
+def _sigterm_handler(signum: int, _frame: object) -> None:
+    """Handle SIGTERM/SIGALRM with a diagnostic message instead of silent death."""
+    sig_name = signal.Signals(signum).name
+    print(
+        f"\nERROR: Process received {sig_name} during preflight check.\n"
+        "This usually means the LiteLLM proxy is unreachable or hanging.\n"
+        f"LLM_BASE_URL: {os.environ.get('LLM_BASE_URL', '(not set)')}\n",
+        file=sys.stderr,
+        flush=True,
+    )
+    sys.exit(1)
+
+
+signal.signal(signal.SIGTERM, _sigterm_handler)
+signal.signal(signal.SIGALRM, _sigterm_handler)
 
 
 # SDK-specific parameters that should not be passed to litellm.
@@ -380,6 +399,31 @@ def check_model(
 test_model = check_model
 
 
+def _check_proxy_reachable(
+    base_url: str, api_key: str | None = None, timeout: int = 10
+) -> tuple[bool, str]:
+    """Quick health check: can we reach the proxy at all?
+
+    Uses /v1/models (standard OpenAI-compatible endpoint) which works with
+    any valid API key. The /health endpoint requires admin-level access on
+    some LiteLLM configurations.
+    """
+    import urllib.error
+    import urllib.request
+
+    models_url = f"{base_url.rstrip('/')}/v1/models"
+    try:
+        req = urllib.request.Request(models_url, method="GET")
+        if api_key:
+            req.add_header("Authorization", f"Bearer {api_key}")
+        urllib.request.urlopen(req, timeout=timeout)
+        return True, f"Proxy reachable at {base_url}"
+    except urllib.error.URLError as e:
+        return False, f"Cannot reach proxy at {base_url}: {e.reason}"
+    except Exception as e:
+        return False, f"Cannot reach proxy at {base_url}: {type(e).__name__}: {e}"
+
+
 def run_preflight_check(models: list[dict[str, Any]]) -> bool:
     """Run preflight LLM check for all models.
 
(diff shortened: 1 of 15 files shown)


xingyaoww:
Added a commit to expose the SDK packages (openhands-sdk, openhands-workspace, openhands-tools) to the system Python in the `source` and `source-minimal` Docker targets via `PYTHONPATH`.

This is needed for the automations service dispatcher ([deploy PR #3456](https://github.com/OpenHands/deploy/pull/3456)) — when an automation script runs inside a sandbox with `saas_runtime_mode=True`, it needs to `import openhands.workspace` etc. Without this, those packages are only available inside the agent-server's venv, not to the system Python that user scripts run under.

Once this PR's image finishes building, we'll reference the new image tag in the deploy PR to enable full SDK automation scripts.

openhands-workspace/openhands/workspace/cloud/workspace.py
```diff
@@ -26,6 +26,9 @@
 # Number of retry attempts for transient API failures
 _MAX_RETRIES = 3
 
+# Default port the agent-server listens on inside a Cloud Runtime
+DEFAULT_AGENT_SERVER_PORT = 60000
```
xingyaoww, line 31: @OpenHands let's read from `AGENT_SERVER_PORT` env var, and if it doesn't exists, read from 60000.

xingyaoww: Fixed in 975647d9. `agent_server_port` now reads from `AGENT_SERVER_PORT` env var with fallback to 60000. Added tests for env var reading and kwarg precedence.

xingyaoww:
@OpenHands please fix https://github.com/OpenHands/software-agent-sdk/actions/runs/23438124731/job/68181392308?pr=2490

jpshackelford:
@xingyaoww @malhotra5 I was trying to make sure I understood this PR and in the process created some additional documentation at https://github.com/OpenHands/docs/pull/414. Would your feedback.

openhands-workspace/openhands/workspace/cloud/workspace.py
```diff
@@ -61,42 +77,80 @@ class OpenHandsCloudWorkspace(RemoteWorkspace):
...
+        description=(
+            "URL to POST completion status to when the context manager exits. "
+            "Set by the automation service; not intended for direct user use."
+        ),
+    )
+    automation_run_id: str | None = Field(
+        default=None,
+        description=("Automation run ID included in the completion callback payload."),
+    )
+
```
malhotra5: I'm a bit torn about this

On the one hand, it seems like overhead to expect the user to configure automation callback url. It would be easier if we just read it from env directly as the automation service already sets this

On the other hand, configurations via envs is not a pattern we follow today

openhands-workspace/openhands/workspace/cloud/workspace.py
```diff
@@ -543,4 +665,35 @@ def __enter__(self) -> OpenHandsCloudWorkspace:
...
+        status = "COMPLETED" if exc_type is None else "FAILED"
+        payload: dict[str, Any] = {"status": status}
+        if self.automation_run_id:
+            payload["run_id"] = self.automation_run_id
+        if exc_val is not None:
+            payload["error"] = str(exc_val)
+
+        try:
+            with httpx.Client(timeout=10.0) as cb_client:
+                resp = cb_client.post(callback_url, json=payload)
```
malhotra5: I think we want to protect this webhook call (with an api key, or signature)

openhands-workspace/openhands/workspace/cloud/workspace.py
jpshackelford: >  configurations via envs is not a pattern we follow today

I assume you mean specifically in `software-agent-sdk` because our Helm charts have scores of env configs. I this a case of an environment variable as message passing between components (dynamic at runtime) or in terms of configuration for operation (static at runtime)?

malhotra5: Yep I mean specifically for the SDK, a configuration for operation

Every component in the SDK so far has explicit params passed for its respective configuration, rather than automatically loading from an env (we wanted to stay away from unexpected side effects of instantiating components). We can read from envs, but that's an explicit operation (e.g `LLM.load_from_env`)

openhands-workspace/openhands/workspace/cloud/workspace.py
malhotra5: Turns out this route is already protects by the cloud api key, the fix for this has been pushed

openhands-workspace/openhands/workspace/cloud/workspace.py
xingyaoww: Agreed here... having `automation_*` args here isn't really pretty. Although i think the exception about env var can be made, at least for this `CloudWorkspace` that presumably works together in an environment we expected to have certain vars. And we are already reading the env var in this PR

https://github.com/OpenHands/software-agent-sdk/pull/2490/changes#diff-e4350dd2adb4a27d3fc91c608787e55eb9db6edccc2046d6082446b672e32b29R212

can we move automation_run_id and automation_callback_url and read them from env vars? 

I'm also not 100% happy about the arg name for `saas_runtime_mode` - We should also rename `saas_runtime_mode` to `local_agent_server_mode` and update the PR title/description too.

xingyaoww: Done in 54e02e3a — moved `automation_callback_url` / `automation_run_id` to env vars and renamed `saas_runtime_mode` → `local_agent_server_mode`. Automation repo updated in d2c7697.

xingyaoww: Addressed both suggestions in 54e02e3a (SDK) and d2c7697 (automation). PR title and description updated as well.

PR #2491, opened 2026-03-18
refactor(tools): remove max_run

VascoSch92:
## Summary

This PR removes the max_turns parameter from the TaskAction tool API.

## Motivation

While running various evals with sub-agents, it was found that in 5–10% of cases, the main agent gives a max_turns limit that is too small. This results in the sub-agent being unable to complete its task, wasting tokens (as the main agent must resume it afterward). Therefore, this parameter is not only unnecessary but actually problematic.

This is a semi-breaking API change; however, the `TaskAction` tool is currently only present in the SDK, not in the CLI or GUI.

Moreover, the user retains the ability to control this parameter via `max_iterations_per_run` in file-based agents.

The logic to choose the `max_iterations_per_run` in sub-agent conversations is now:
- Use the one defined in the markdown defining the agent (if present).
- Otherwise, fall back to the same number as the parent conversation.

This is in line with other logic in the tool, and it allows us to delete a hardcoded parameter (`manager.py line 245`).

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:b5c6322-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-b5c6322-python \
  ghcr.io/openhands/agent-server:b5c6322-python
```

```
ghcr.io/openhands/agent-server:b5c6322-golang-amd64
ghcr.io/openhands/agent-server:b5c6322-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:b5c6322-golang-arm64
ghcr.io/openhands/agent-server:b5c6322-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:b5c6322-java-amd64
ghcr.io/openhands/agent-server:b5c6322-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:b5c6322-java-arm64
ghcr.io/openhands/agent-server:b5c6322-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:b5c6322-python-amd64
ghcr.io/openhands/agent-server:b5c6322-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:b5c6322-python-arm64
ghcr.io/openhands/agent-server:b5c6322-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:b5c6322-golang
ghcr.io/openhands/agent-server:b5c6322-java
ghcr.io/openhands/agent-server:b5c6322-python
```

- Each variant tag (e.g., `b5c6322-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `b5c6322-python-amd64`) are also available if needed

### DIFF
--- a/.github/run-eval/resolve_model_config.py
+++ b/.github/run-eval/resolve_model_config.py
@@ -14,8 +14,27 @@ Outputs to GITHUB_OUTPUT:
 
 import json
 import os
+import signal
 import sys
+import time
 from typing import Any
+
+
+def _sigterm_handler(signum: int, _frame: object) -> None:
+    """Handle SIGTERM/SIGALRM with a diagnostic message instead of silent death."""
+    sig_name = signal.Signals(signum).name
+    print(
+        f"\nERROR: Process received {sig_name} during preflight check.\n"
+        "This usually means the LiteLLM proxy is unreachable or hanging.\n"
+        f"LLM_BASE_URL: {os.environ.get('LLM_BASE_URL', '(not set)')}\n",
+        file=sys.stderr,
+        flush=True,
+    )
+    sys.exit(1)
+
+
+signal.signal(signal.SIGTERM, _sigterm_handler)
+signal.signal(signal.SIGALRM, _sigterm_handler)
 
 
 # SDK-specific parameters that should not be passed to litellm.
@@ -380,6 +399,31 @@ def check_model(
 test_model = check_model
 
 
+def _check_proxy_reachable(
+    base_url: str, api_key: str | None = None, timeout: int = 10
+) -> tuple[bool, str]:
+    """Quick health check: can we reach the proxy at all?
+
+    Uses /v1/models (standard OpenAI-compatible endpoint) which works with
+    any valid API key. The /health endpoint requires admin-level access on
+    some LiteLLM configurations.
+    """
+    import urllib.error
+    import urllib.request
+
+    models_url = f"{base_url.rstrip('/')}/v1/models"
+    try:
+        req = urllib.request.Request(models_url, method="GET")
+        if api_key:
+            req.add_header("Authorization", f"Bearer {api_key}")
+        urllib.request.urlopen(req, timeout=timeout)
+        return True, f"Proxy reachable at {base_url}"
+    except urllib.error.URLError as e:
+        return False, f"Cannot reach proxy at {base_url}: {e.reason}"
+    except Exception as e:
+        return False, f"Cannot reach proxy at {base_url}: {type(e).__name__}: {e}"
+
+
 def run_preflight_check(models: list[dict[str, Any]]) -> bool:
     """Run preflight LLM check for all models.
 
@@ -401,23 +445,42 @@ def run_preflight_check(models: list[dict[str, Any]]) -> bool:
         print("Preflight check: SKIPPED (LLM_API_KEY not set)")
         return True
 
-    print(f"\nPreflight LLM check for {len(models)} model(s)...")
-    print("-" * 50)
+    # Quick connectivity check before trying expensive model completions
+    print(f"\nChecking proxy connectivity: {base_url}", flush=True)
+    reachable, msg = _check_proxy_reachable(base_url, api_key=api_key)
+    if not reachable:
+        print(f"✗ {msg}", file=sys.stderr, flush=True)
+        print(
+            "\nThe LiteLLM proxy appears to be down or unreachable.\n"
+            "Set SKIP_PREFLIGHT=true to bypass this check.",
+            file=sys.stderr,
+            flush=True,
+        )
+        return False
+    print(f"✓ {msg}", flush=True)
+
+    print(f"\nPreflight LLM check for {len(models)} model(s)...", flush=True)
+    print("-" * 50, flush=True)
 
     all_passed = True
     for model_config in models:
+        display_name = model_config.get("display_name", "unknown")
+        print(f"  Checking {display_name}...", end=" ", flush=True)
+        t0 = time.monotonic()
         success, message = check_model(model_config, api_key, base_url)
-        print(message)
+        elapsed = time.monotonic() - t0
+        print(f"({elapsed:.1f}s)", flush=True)
+        print(f"  {message}", flush=True)
         if not success:
             all_passed = False
 
-    print("-" * 50)
+    print("-" * 50, flush=True)
 
     if all_passed:
-        print(f"✓ All {len(models)} model(s) passed preflight check\n")
+        print(f"✓ All {len(models)} model(s) passed preflight check\n", flush=True)
     else:
-        print("✗ Some models failed preflight check")
-        print("Evaluation aborted to avoid wasting compute resources.\n")
+        print("✗ Some models failed preflight check", flush=True)
+        print("Evaluation aborted to avoid wasting compute resources.\n", flush=True)
 
     return all_passed
 
@@ -431,7 +494,7 @@ def main() -> None:
 
     # Resolve model configs
     resolved = find_models_by_id(model_ids)
-    print(f"Resolved {len(resolved)} model(s): {', '.join(model_ids)}")
+    print(f"Resolved {len(resolved)} model(s): {', '.join(model_ids)}", flush=True)
 
     # Run preflight check
     if not run_preflight_check(resolved):
--- a/.github/scripts/check_agent_server_rest_api_breakage.py
+++ b/.github/scripts/check_agent_server_rest_api_breakage.py
@@ -25,11 +25,19 @@ Policies enforced:
      declare a scheduled removal version that has been reached by the current
      package version.
 
-3) No in-place contract breakage
+3) Additive response oneOf/anyOf expansion is allowed
+   - Adding new members to ``oneOf`` or ``anyOf`` discriminated unions in response
+     schemas is a normal evolution for extensible event-stream APIs.  Clients MUST
+     handle unknown discriminator values gracefully (skip/ignore).
+   - oasdiff flags ``response-body-one-of-added`` and
+     ``response-property-one-of-added`` as ERR; this script downgrades them to
+     informational notices.
+
+4) No in-place contract breakage
    - Breaking REST contract changes that are not removals of previously-deprecated
-     operations fail the check. REST clients need 5 minor releases of runway, so
-     incompatible replacements must ship additively or behind a versioned contract
-     until the scheduled removal version.
+     operations or additive oneOf expansions fail the check. REST clients need 5
+     minor releases of runway, so incompatible replacements must ship additively or
+     behind a versioned contract until the scheduled removal version.
 
 If the baseline release schema can't be generated (e.g., missing tag / repo issues),
 the script emits a warning and exits successfully to avoid flaky CI.
@@ -410,11 +418,33 @@ def _validate_removed_operations(
     return errors
 
 
+# oasdiff rule IDs for additive oneOf/anyOf expansion in response schemas.
+# These are flagged as ERR by oasdiff but are expected evolution for extensible
+# discriminated-union APIs (e.g. the events endpoint).  We downgrade them to
+# informational notices so they don't block CI.
+_ADDITIVE_RESPONSE_ONEOF_IDS = frozenset(
+    {
+        "response-body-one-of-added",
+        "response-property-one-of-added",
+        # Keep the anyOf variants here too so that if oasdiff ever reports them
+        # as breakages, additive response-union expansion gets the same
+        # downgrade without further script changes.
+        "response-body-any-of-added",
+        "response-property-any-of-added",
+    }
+)
+
+
 def _split_breaking_changes(
     breaking_changes: list[dict],
-) -> tuple[list[dict], list[dict]]:
-    """Split oasdiff results into removals and all other breakages."""
+) -> tuple[list[dict], list[dict], list[dict]]:
+    """Split oasdiff results into three buckets.
+
+    Returns:
+        (removed_operations, additive_response_oneof, other_breaking_changes)
+    """
     removed_operations: list[dict] = []
+    additive_response_oneof: list[dict] = []
     other_breaking_changes: list[dict] = []
 
     for change in breaking_changes:
@@ -431,9 +461,13 @@ def _split_breaking_changes(
             )
             continue
 
+        if change_id in _ADDITIVE_RESPONSE_ONEOF_IDS:
+            additive_response_oneof.append(change)
+            continue
+
         other_breaking_changes.append(change)
 
-    return removed_operations, other_breaking_changes
+    return removed_operations, additive_response_oneof, other_breaking_changes
 
 
 def _normalize_openapi_for_oasdiff(schema: dict) -> dict:
@@ -570,9 +604,11 @@ def main() -> int:
                 "in JSON format. There may be warnings only."
             )
     else:
-        removed_operations, other_breaking_changes = _split_breaking_changes(
-            breaking_changes
-        )
+        (
+            removed_operations,
+            additive_response_oneof,
+            other_breaking_changes,
+        ) = _split_breaking_changes(breaking_changes)
         removal_errors = _validate_removed_operations(
             removed_operations,
             prev_schema,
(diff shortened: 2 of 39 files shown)


enyst:
@openhands read all comments to this PR. You will see it PASSED the Griffe check, which should flag / fail when the public API changes, and tools package is part of the public API. Verify why it didn't fail here, even though the PR removes an argument.
Make suggestions, WDYT about this?

enyst:
I dug into the workflow/script, and the reason it passed is **not** that Griffe missed the breakage — it's that our current API-breakage check never looks at `TaskAction`.

What the checker does today:
- `.github/scripts/check_sdk_api_breakage.py` loads the package roots from `PACKAGES`
- for `openhands-tools`, it resolves only the curated public surface from `openhands.tools.__all__`
- then it compares those exported objects, plus public members on exported classes

Why this PR slipped through:
- the removed argument is `TaskAction.max_turns` in `openhands.tools.task.definition.TaskAction`
- `TaskAction` is **not** exported from `openhands.tools.__all__`
- and `openhands.tools.task.__all__` currently exports only `TaskToolSet`, not `TaskAction`
- so the checker never compares `TaskAction` at all

I also verified locally that **Griffe *would* flag it** if we point it at that class directly:
- comparing `openhands-tools==1.14.0` vs this branch on `TaskAction` yields:
  `ObjectRemovedBreakage ... title=TaskAction.max_turns::Public object was removed`

So IMO this is a **coverage/policy gap in our check**, not a Griffe limitation.

### Suggestions

I see 3 options:

1. **Keep current policy and do nothing**
   - If we define the Python public API strictly as the curated root `openhands.tools.__all__`, then the green check is technically correct.
   - In that world, this PR is *not* a Python-public-API break according to current policy; it's a tool-contract change.

2. **Expand the Python API check to recurse into tool submodules**
   - e.g. also walk `openhands.tools.<submodule>` and honor each submodule's `__all__`
   - but this specific change would still require `TaskAction` to be exported somewhere (currently it isn't)
   - otherwise we risk either missing real tool-schema breaks or widening the public surface implicitly/noisily

3. **Add a dedicated tool-schema compatibility check** (**my preference**)
   - diff the generated tool input/output schemas (e.g. MCP/OpenAI tool schemas) for shipped tools
   - that matches the real contract consumers depend on much better than Python symbol exports
   - then removing a tool argument like `max_turns` would fail even if the underlying Pydantic model class is not re-exported

My take: **(3) is the cleanest long-term fix**. This change is really about the *tool interface*, not just Python imports, so Griffe-on-`__all__` is the wrong abstraction to rely on for this class of breakage. As a smaller interim improvement, we could also teach the breakage check to recurse into `openhands.tools.*` submodule `__all__`, but that still won't catch non-exported action models like `TaskAction` unless we decide they are public and export them explicitly.

Happy to follow up with a patch for the checker if we want to tighten this.

enyst:
HUMAN:
TaskAction and obs are actions and observation, IMO we should expose them in the public API. The agent also has a point on tool schemas, though that can wait IMHO. WDYT?

VascoSch92:
> HUMAN:
> TaskAction and obs are actions and observation, IMO we should expose them in the public API. The agent also has a point on tool schemas, though that can wait IMHO. WDYT?

I didn't do it at the time because I thought we would just expose the TaskToolSet—which will eventually include TaskAction, StopTaskAction, and CheckTaskAction. However, we can certainly expose them directly for now.

VascoSch92:
sorry @enyst I don't get it. Do you want me to expose the tools in the API in this PR?

enyst:
I think so... That means:
- on one hand, they are not currently public, so we don't break public API with this PR
- on the other hand, please correct me if wrong, the action and obs really are events, saved in the `EventLog`, which I wonder, does it mean that old conversations maybe cannot be restored?

VascoSch92:
So they will break existing convo. But as the API was not public, people should not use them, right?

On the other hand, we could do

```python
max_turns: int | None = Field(
        default=None,
        description="Deprecated: This field is ignored. Maximum iterations are now determined by the agent definition or parent conversation.",
        deprecated=True,  # Pydantic v2 supports this
        ge=1,
        json_schema_extra={"deprecated": True},  # Also mark in JSON schema
    )
```

to make it backward compatible. What do you think?

enyst:
> So they will break existing convo. But as the API was not public, people should not use them, right?

I think these are two different things, using via API and the fact that they were shipped with the SDK, so some client applications got them: [example](https://openhands-ai.slack.com/archives/C06R25BT5B2/p1773608996896419) I think

On events (so older conversations that the user can restore) this is a different policy, even, because backwards compatibility on events doesn’t even have a time limit IIRC: [event log stuff](https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/AGENTS.md#critical-requirement-old-events-must-always-load)

I wonder why didn’t the agent see it… maybe splitting AGENTS.md wasn’t a good idea 🤔 

Yes, I think something like that should work

VascoSch92:
@enyst 

- API exposed
- deprecation warning for `max_turns`

enyst:
@openhands check CI and look at the REST API failure log. All we do here is add an element to the list, do you think that should be an API breakage? It doesn’t seem like it would actually affect client applications. WDYT? Verify all I said. You are free to contradict me, we need to understand correctly. :)
Respond directly on github on this PR. Investigate all you need.

enyst:
Checked it carefully. Current CI is green **except** for `REST API (OpenAPI)` ([run 23302138634](https://github.com/OpenHands/software-agent-sdk/actions/runs/23302138634)).

I pulled the full failure locally. The important part is: the REST failure is **not** about removing `max_turns`.

I verified that:
- `TaskAction.max_turns` is now `SkipJsonSchema[...]`, so it is **not present** in the generated OpenAPI schema.
- the failing diff is instead caused by these new schema variants appearing in the REST contract:
  - `TaskAction` added to `Action.oneOf`
  - `TaskObservation` added to `Observation.oneOf`
  - `TaskTool` and `TaskToolSet` added to `ToolDefinition.oneOf`
- those then propagate into the event endpoints, which matches the log exactly.

Root cause is the new import/export path, not `max_turns` itself:
- this PR adds `TaskToolSet` to `openhands.tools.__all__`
- importing `openhands.tools.preset.default` (which agent-server does during app/OpenAPI startup) first imports `openhands.tools`
- that now eagerly imports `openhands.tools.task`
- and that loads/registers `TaskAction` / `TaskObservation` / `TaskTool` / `TaskToolSet`

I also verified that after `import openhands.tools.preset.default`, `task` / `task_tool_set` become registered tools in-process, whereas they were not before. So this is a real REST-surface expansion, not just documentation noise.

So I only **partially** agree with “we only add an element to the list, this shouldn’t affect clients”:
- **for permissive clients** that treat responses as generic JSON, this is probably harmless in practice;
- **for strict/generated clients**, adding a new member to a discriminated `oneOf` is *not* universally backward-compatible. A client generated from the previous OpenAPI can fail to deserialize an unknown `kind` such as `TaskAction` / `TaskObservation` / `TaskTool(Set)`, or break on exhaustive matching.

So under the **current REST policy/checker**, the failure looks **consistent**, not spurious.

My take:
1. this PR’s REST failure is real, but it is about the new task schemas entering the REST union types — **not** about `max_turns` removal;
2. if we want the Python re-export without changing REST, the clean fix is to avoid the eager import side effect from `openhands.tools.__init__` (for example a lazy re-export);
3. if we decide additive `oneOf` members should be allowed for REST, that should be an explicit checker/policy change in a separate PR, because today both the repo policy and `oasdiff` treat this as breaking.

VascoSch92:
@enyst :-)

VascoSch92:
@enyst :-)

enyst:
Sorry, I had no bandwidth for this PR. We were trying to figure out what breakage to accept and why, right?

@smolpaws CI is failing, check it out. We are considering to adjust REST CI workflow to accept some kind of breakage, but you need to think well what we should accept and why. Read it and understand the PR too. Consider all comments in this discussion so far, and give us an opinion here.

If you encounter problems with the token, note that my machine can give you access to another token via Mac stuffs, but don't use alternative means unless you really need to. You should be able to read everything and post a comment with your thoughts. Don't hold back.

enyst:
@smolpaws second tunnel probe: if you can see this, please reply with a short note that the fresh tunnel is alive.

smolpaws:
Tunnel's up, I can see and reach the runner. 🐾

smolpaws:
I can see your comment on issue #2491, and I'm here on the runner. Everything's connected. 🐾

smolpaws:
Fresh tunnel is alive and purring. 🐾

smolpaws:
I can see your comment on issue #2491, and I'm here on the runner. Everything's connected. 🐾

enyst:
@smolpaws Please ignore the tunnel-probe comments and read my earlier real request here: https://github.com/OpenHands/software-agent-sdk/pull/2491#issuecomment-4129301553 . Please respond to that request normally on this PR.

smolpaws:
I've read the full PR, all comments (including @openhands-ai's analysis), the CI logs, and the REST API breakage checker code. Here's my take.

---

### What's actually breaking CI

The REST API failure has **nothing to do with removing `max_turns`**. The `SkipJsonSchema[...]` wrapper correctly hides it from the OpenAPI spec, and `TaskAction` wasn't in the baseline schema anyway — so there's no REST-visible removal.

The failure is because this PR adds `TaskToolSet` to `openhands.tools.__all__`, which causes eager import of the task module during OpenAPI generation. That registers `TaskAction`, `TaskObservation`, `TaskTool`, and `TaskToolSet` as new members in the `Action`, `Observation`, and `ToolDefinition` discriminated unions. oasdiff reports these as breaking changes:

```
added '#/components/schemas/TaskAction' to ... response property 'oneOf' list
added '#/components/schemas/TaskObservation' to ... response property 'oneOf' list
added '#/components/schemas/TaskTool, #/components/schemas/TaskToolSet' to ... response property 'oneOf' list
```

### Why oasdiff flags this

oasdiff treats adding new `oneOf` members in **response** schemas as breaking (`ERR`), because a strict/generated client with an exhaustive match on discriminator values would choke on an unknown type. This is the conservative interpretation and technically defensible.

### Why we should allow it anyway

The events endpoint is fundamentally an **extensible discriminated union**. Every time we add a new tool type, action type, or observation type, the response schema grows. This is the standard evolution pattern for event-stream APIs (GitHub webhooks, Stripe events, etc.). If the checker blocks every additive `oneOf` expansion, the system can't grow without failing CI. That's not a useful policy — it's a straitjacket.

The right contract for consumers of this API: **you MUST handle unknown discriminator values gracefully** (skip/ignore). This is the industry norm for extensible union types in REST APIs.

### What I'd change in the checker

In `_split_breaking_changes`, add a third category alongside the existing `removed_operations` bucket: **additive response oneOf/anyOf expansion**. The oasdiff rule IDs are `response-body-one-of-added` and `response-property-one-of-added`. These should be logged (so they're visible) but not cause failure.

Keep blocking everything else: removals, type changes, narrowing unions, breaking request schemas, etc. The 5-release deprecation runway policy stays intact for actual contract breakage.

### What I would NOT allow

- Removing a `oneOf` member from a response (narrows what clients can expect)
- Adding to request `oneOf` if it changes required input semantics
- Any change that modifies the shape of an existing schema member

### On the PR itself

The `max_turns` deprecation is well-executed:
- Python backward compat preserved — field still exists, just `deprecated=True` + ignored
- REST: no removal to worry about since `TaskAction` wasn't in the baseline schema
- The new precedence logic (agent definition `max_iterations_per_run` → parent conversation fallback) is cleaner than the caller guessing a number

One nit: the expanded `__init__.py` exports (`TaskAction`, `TaskObservation`, etc.) are a separate concern from `max_turns` removal. They could have been their own PR. But they're reasonable changes — these types are part of the public package, they should be exported.

### TL;DR

**Allow additive `oneOf` expansion in response schemas.** It's the only sane policy for an extensible event API. Make it a permanent checker change, not a one-off override, because this will happen every time a new tool or action type ships. This PR's REST failure is a false positive under the right policy.

🐾

smolpaws:
I've read the full PR, all comments (including @openhands-ai's analysis), the CI logs, and the REST API breakage checker code. Here's my take.

---

### What's actually breaking CI

The REST API failure has **nothing to do with removing `max_turns`**. The `SkipJsonSchema[...]` wrapper correctly hides it from the OpenAPI spec, and `TaskAction` wasn't in the baseline schema anyway — so there's no REST-visible removal.

The failure is because this PR adds `TaskToolSet` to `openhands.tools.__all__`, which causes eager import of the task module during OpenAPI generation. That registers `TaskAction`, `TaskObservation`, `TaskTool`, and `TaskToolSet` as new members in the `Action`, `Observation`, and `ToolDefinition` discriminated unions. oasdiff reports these as breaking changes:

```
added '#/components/schemas/TaskAction' to ... response property 'oneOf' list
added '#/components/schemas/TaskObservation' to ... response property 'oneOf' list
added '#/components/schemas/TaskTool, #/components/schemas/TaskToolSet' to ... response property 'oneOf' list
```

### Why oasdiff flags this

oasdiff treats adding new `oneOf` members in **response** schemas as breaking (`ERR`), because a strict/generated client with an exhaustive match on discriminator values would choke on an unknown type. This is the conservative interpretation and technically defensible.

### Why we should allow it anyway

The events endpoint is fundamentally an **extensible discriminated union**. Every time we add a new tool type, action type, or observation type, the response schema grows. This is the standard evolution pattern for event-stream APIs (GitHub webhooks, Stripe events, etc.). If the checker blocks every additive `oneOf` expansion, the system can't grow without failing CI. That's not a useful policy — it's a straitjacket.

The right contract for consumers of this API: **you MUST handle unknown discriminator values gracefully** (skip/ignore). This is the industry norm for extensible union types in REST APIs.

### What I'd change in the checker

In `_split_breaking_changes`, add a third category alongside the existing `removed_operations` bucket: **additive response oneOf/anyOf expansion**. The oasdiff rule IDs are `response-body-one-of-added` and `response-property-one-of-added`. These should be logged (so they're visible) but not cause failure.

Keep blocking everything else: removals, type changes, narrowing unions, breaking request schemas, etc. The 5-release deprecation runway policy stays intact for actual contract breakage.

### What I would NOT allow

- Removing a `oneOf` member from a response (narrows what clients can expect)
- Adding to request `oneOf` if it changes required input semantics
- Any change that modifies the shape of an existing schema member

### On the PR itself

The `max_turns` deprecation is well-executed:
- Python backward compat preserved — field still exists, just `deprecated=True` + ignored
- REST: no removal to worry about since `TaskAction` wasn't in the baseline schema
- The new precedence logic (agent definition `max_iterations_per_run` → parent conversation fallback) is cleaner than the caller guessing a number

One nit: the expanded `__init__.py` exports (`TaskAction`, `TaskObservation`, etc.) are a separate concern from `max_turns` removal. They could have been their own PR. But they're reasonable changes — these types are part of the public package, they should be exported.

### TL;DR

**Allow additive `oneOf` expansion in response schemas.** It's the only sane policy for an extensible event API. Make it a permanent checker change, not a one-off override, because this will happen every time a new tool or action type ships. This PR's REST failure is a false positive under the right policy.

smolpaws:
My detailed analysis is already posted on the PR — you can see it here: https://github.com/OpenHands/software-agent-sdk/pull/2491#issuecomment-4129418731

(The earlier conversation turns had tunnel issues, but the comment landed via `gh` CLI in the last turn.)

enyst:
@smolpaws I think I agree. Please open a new PR with the CI oasdiff workflow change you are proposing in your previous comment here. Note that your token doesn't have repo perms, but that shouldn't stop you, you have a fork on github, you can clone it locally if you want, as long as you make sure to use the repos location! (and maybe rename it locally e.g. smol-agent-sdk or something, so that we know)

smolpaws:
Now update `_split_breaking_changes` to return three categories:

enyst:
😅 OK, @smolpaws, continue please

smolpaws:
PR is up and CI is green: https://github.com/OpenHands/software-agent-sdk/pull/2571

13 checks passed (including pre-commit, REST API, Python API, sdk-tests, tools-tests, agent-server-tests, cross-tests), 0 failed. Only build-binary-and-test + coverage-report still running.

The change: `_split_breaking_changes` now returns three buckets. Additive `oneOf`/`anyOf` response schema expansions (`response-body-one-of-added`, `response-property-one-of-added`, etc.) are logged as `::notice` but don't fail CI. All other breaking changes still fail as before. 3 new tests cover the behavior. 🐾

smolpaws:
Done! 🐾 PR is up: https://github.com/OpenHands/software-agent-sdk/pull/2571

**What it does:** `_split_breaking_changes` now returns three buckets instead of two. The new middle bucket catches additive `oneOf`/`anyOf` expansion in response schemas (`response-body-one-of-added`, `response-property-one-of-added`, plus `anyOf` variants for future-proofing). These get logged as `::notice` in CI but don't fail the check.

Everything else — removals without deprecation runway, type changes, request schema breakage — still fails as before.

3 new tests, all 21 passing, pre-commit clean.

Sorry about the earlier half-message! The work is done — PR is here: https://github.com/OpenHands/software-agent-sdk/pull/2571

CI is looking good: 11 checks passed (including pre-commit, REST API, Python API, sdk-tests, tools-tests, agent-server-tests), 6 skipped (docker builds — expected from a fork PR), and 3 still running (the heavier build-binary-and-test + cross-tests jobs).

The change adds a third bucket to `_split_breaking_changes` for additive `oneOf`/`anyOf` response schema expansion. These get logged as `::notice` but don't fail CI. Everything else still fails as before.

smolpaws:
Looks like my messages keep getting queued but maybe not landing on the issue. Let me post directly via `gh`:

enyst:
Ha, github closed this PR automatically when I merged the REST API tweak, because the PR description said so "Closes #2491 (originated from the review discussion there)" 😅 

Sorry for the annoying stuff around here! I learned some behavioral quirks and fixed some stuff with smolpaws.

PR #2495, opened 2026-03-18
feat(plugin): Support multiple marketplace registrations with auto-load semantics

### LINKED ISSUE #2494: feat: Support multiple marketplace registrations with auto-load semantics
## Problem Statement

[PR #2253](<https://github.com/OpenHands/software-agent-sdk/pull/2253>) added a `marketplace_path` string that filters which skills to load from the public skills repository. This design has two limitations:

1. **Single source**: Users can only load skills from one marketplace (the public skills repo). Enterprises need to register multiple marketplaces — their own internal marketplace, team-specific marketplaces, and optionally the public marketplace.
2. **No registration vs auto-load distinction**: The current model is all-or-nothing. There's no way to say "I want this marketplace registered and available, but don't auto-load its plugins into every conversation."

## Proposed Solution

Replace the single `marketplace_path: str` with a list of `MarketplaceRegistration` objects:

```python
class MarketplaceRegistration(BaseModel):
    name: str                              # Identifier for this registration
    source: str                            # github:owner/repo, git URL, or local path
    ref: str | None = None                 # Optional branch/tag/commit
    repo_path: str | None = None           # Subdirectory within repo (for monorepos)
    auto_load: Literal["all"] | None = None
```

**Behavior:**

* `auto_load: "all"` — Fetch the marketplace and load all plugins declared in its manifest at conversation start
* `auto_load: None` (default) — Marketplace is registered for plugin resolution but plugins are not auto-loaded

**Examples:**

```python
AgentContext(
    registered_marketplaces=[
        # Standalone marketplace repo
        MarketplaceRegistration(
            name="public", 
            source="github:OpenHands/skills", 
            auto_load="all"
        ),
        # Monorepo with marketplace in subdirectory
        MarketplaceRegistration(
            name="team", 
            source="github:acme/monorepo",
            repo_path="marketplaces/internal",
            auto_load="all"
        ),
        # Registered but not auto-loaded
        MarketplaceRegistration(
            name="experimental", 
            source="github:acme/experimental"
        ),
    ]
)
```

## Plugin Resolution

When a marketplace is registered (regardless of `auto_load` setting), the SDK resolves plugin references against it.

### MarketplaceRegistry (New)

Introduce a new internal `MarketplaceRegistry` class to manage registered marketplaces:

```python
class MarketplaceRegistry:
    """Manages registered marketplaces with lazy fetching and plugin resolution."""
    
    def __init__(self, registrations: list[MarketplaceRegistration]): ...
    
    def resolve_plugin(self, plugin_ref: str) -> PluginSource:
        """Resolve 'plugin-name' or 'plugin-name@marketplace' to a PluginSource."""
        ...
    
    def prefetch_all(self) -> None:
        """Eagerly fetch all registered marketplaces (optional, for validation)."""
        ...
```

The registry is an internal implementation detail — users configure `registered_marketplaces` on `AgentContext`, and the SDK manages the registry internally.

### Lazy Loading

Marketplace fetching happens **lazily on first use**, not at registration time:

* Registering a marketplace with `auto_load: None` has zero startup cost
* The marketplace repo is only fetched when a plugin from it is first referenced
* Fetched marketplaces are cached for the session
* Code managing the registry may optionally call `prefetch_all()` for eager validation

### Manifest Lookup

Uses the existing `Marketplace.load()` behavior which checks `MARKETPLACE_MANIFEST_DIRS` (`.plugin`, `.claude-plugin`) for `marketplace.json`.

### Reference Formats

* `plugin-name@marketplace-name` — Explicit marketplace qualifier
* `plugin-name` — Search registered marketplaces (error if ambiguous across marketplaces)

### Resolution Flow

1. Parse the reference to extract plugin name and optional marketplace name
2. Look up marketplace in `registered_marketplaces`
3. Fetch/cache the marketplace repo if not already fetched (lazy)
4. Load marketplace manifest via existing `Marketplace.load()`
5. Find the plugin in the marketplace's manifest
6. Return resolved plugin source for loading

## Deprecations

The following APIs added in OpenHands/software-agent-sdk#2253 (v1.13.0) will be deprecated:

| Deprecated | Replacement |
| -- | -- |
| `AgentContext.marketplace_path` | `AgentContext.registered_marketplaces` |
| `load_public_skills(marketplace_path=...)` | Use marketplace registry |
| `load_available_skills(marketplace_path=...)` | Use marketplace registry |
| `SkillsRequest.marketplace_path` (agent-server) | `SkillsRequest.registered_marketplaces` |

**Backward compatibility:**

* Deprecated fields/parameters continue to work with deprecation warnings
* If `marketplace_path` is set and `registered_marketplaces` is empty, automatically convert to:

  ```python
  MarketplaceRegistration(
      name="public",
      source=PUBLIC_SKILLS_REPO,
      ref=PUBLIC_SKILLS_BRANCH,
      auto_load="all"
  )
  ```

### In Scope

* Add `MarketplaceRegistration` model to `plugin/types.py`
* Add `MarketplaceRegistry` class (new) for lazy fetching, caching, and resolution
* Add `registered_marketplaces: list[MarketplaceRegistration]` to `AgentContext`
* Reuse existing `Marketplace.load()` for manifest loading
* Add plugin resolution logic (`resolve_plugin(name, marketplace?)` -> `PluginSource`)
* Update plugin loading to use resolution against registered marketplaces
* Update skill loading to iterate over registered marketplaces with `auto_load="all"`
* Deprecate `marketplace_path` APIs (see table above) with backward compatibility
* Tests for multiple marketplace loading and plugin resolution

### Out of Scope (Future Work)

* `auto_load: ["plugin-a", "plugin-b"]` selective loading
* Conflict resolution UI/UX for ambiguous plugin names (for now: error)

## Related Issues

**Directly unblocked by this issue:**

* [OpenHands/OpenHands#13117](<https://github.com/OpenHands/OpenHands/issues/13117>) — Alternative to `marketplace_path` setting
* [APP-708](https://linear.app/all-hands-ai/issue/APP-708/prd-openhands-cloud-default-organizational-resources-skills-mcp-agents) — Enables org-level default plugins via scoped marketplace registration
* [PRD-71](https://linear.app/all-hands-ai/issue/PRD-71/m2-marketplace-management) — SDK provides `MarketplaceRegistration` model for CLI marketplace management
* [APP-769](https://linear.app/all-hands-ai/issue/APP-769/m1-plugin-loading-and-command-invocation) — SDK provides plugin resolution for `/plugin:command` invocation

**Builds on:**

* OpenHands/software-agent-sdk#2253 — `marketplace_path` support (released in v1.13.0)

jpshackelford:
## Summary

This PR introduces support for registering multiple marketplaces with explicit auto-load semantics, replacing the single `marketplace_path` string.

Closes #2494

### New Models

**`MarketplaceRegistration`** - Registration for a plugin marketplace:
```python
class MarketplaceRegistration(BaseModel):
    name: str                              # Identifier for this registration
    source: str                            # github:owner/repo, git URL, or local path
    ref: str | None = None                 # Optional branch/tag/commit
    repo_path: str | None = None           # Subdirectory within repo (for monorepos)
    auto_load: Literal["all"] | None = None
```

### New Classes

**`MarketplaceRegistry`** - Manages registered marketplaces with:
- Lazy fetching: Marketplaces are only fetched when first needed
- Caching: Fetched marketplaces are cached for the session
- Plugin resolution: Resolve plugin references like `plugin-name@marketplace`

### Updated `AgentContext`

```python
AgentContext(
    registered_marketplaces=[
        MarketplaceRegistration(
            name="public", 
            source="github:OpenHands/skills", 
            auto_load="all"
        ),
        MarketplaceRegistration(
            name="team", 
            source="github:acme/monorepo",
            repo_path="marketplaces/internal",
            auto_load="all"
        ),
        MarketplaceRegistration(
            name="experimental", 
            source="github:acme/experimental"
        ),  # registered, not auto-loaded
    ]
)
```

### Deprecations

| Deprecated | Replacement |
|------------|-------------|
| `AgentContext.marketplace_path` | `AgentContext.registered_marketplaces` |

Backward compatibility is maintained - if `marketplace_path` is set and `registered_marketplaces` is empty, a deprecation warning is emitted and the old behavior is preserved.

## Plugin Resolution

Reference formats (same as [Claude Code plugin syntax](https://code.claude.com/docs/en/plugins-reference)):
- `plugin-name@marketplace-name` — Explicit marketplace qualifier
- `plugin-name` — Search registered marketplaces (error if ambiguous)

## Testing

```bash
uv run pytest tests/sdk/plugin/test_marketplace_registry.py -v
```

All 21 tests pass.

## Documentation

- Docs PR: OpenHands/docs#406

## Related

- Builds on #2253 (`marketplace_path` support, released in v1.13.0)
- Enables OpenHands/OpenHands#13117 (alternative to `marketplace_path` setting)
- Enables OpenHands/OpenHands#12916 (org-level default plugins)
- Enables OpenHands/OpenHands-CLI#411 (CLI marketplace management)
- Enables OpenHands/OpenHands-CLI#409 (plugin resolution for slash commands)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:649cfd9-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-649cfd9-python \
  ghcr.io/openhands/agent-server:649cfd9-python
```

```
ghcr.io/openhands/agent-server:649cfd9-golang-amd64
ghcr.io/openhands/agent-server:649cfd9-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:649cfd9-golang-arm64
ghcr.io/openhands/agent-server:649cfd9-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:649cfd9-java-amd64
ghcr.io/openhands/agent-server:649cfd9-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:649cfd9-java-arm64
ghcr.io/openhands/agent-server:649cfd9-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:649cfd9-python-amd64
ghcr.io/openhands/agent-server:649cfd9-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:649cfd9-python-arm64
ghcr.io/openhands/agent-server:649cfd9-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:649cfd9-golang
ghcr.io/openhands/agent-server:649cfd9-java
ghcr.io/openhands/agent-server:649cfd9-python
```

- Each variant tag (e.g., `649cfd9-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `649cfd9-python-amd64`) are also available if needed


enyst:
Thank you for this! Just trying to understand a detail.

@openhands Read this PR and the linked issue, and understand them well. I think I saw in the linked issue that 4 things should be deprecated, but in this PR description I see 1 thing to deprecate. Verify what I just said. Are the others replaced with other implementations here, WDYT, should others be deprecated, or what seems to happen about that? Answer directly as a comment on the PR on github.

enyst:
@enyst I checked the linked issue, this PR, and the API-breakage workflow.

You remembered correctly: #2494 lists **4 deprecations**:
1. `AgentContext.marketplace_path` -> `AgentContext.registered_marketplaces`
2. `load_public_skills(marketplace_path=...)` -> marketplace registry
3. `load_available_skills(marketplace_path=...)` -> marketplace registry
4. `SkillsRequest.marketplace_path` -> `SkillsRequest.registered_marketplaces`

From the code in this PR, though, I only see **#1 actually deprecated today**:
- `AgentContext.marketplace_path` is marked `deprecated=True` and also emits a `DeprecationWarning` in the backward-compat path.

The other 3 are **not deprecated in the implementation yet**:
- `load_public_skills(marketplace_path=...)` still exists unchanged and is still used by the current skill-loading path.
- `load_available_skills(marketplace_path=...)` still exists unchanged and is still what `AgentContext`/server skill loading call into.
- `SkillsRequest.marketplace_path` is still present on the agent-server `/skills` API, and I do not see a `registered_marketplaces` replacement added there in this PR.

So my read is: this PR introduces the **new registration/resolution model** (`MarketplaceRegistration`, `MarketplaceRegistry`, `conversation.load_plugin()` resolution against registered marketplaces), but it does **not** complete the full deprecation rollout described in #2494.

So I think the accurate framing is probably:
- **This PR:** deprecates `AgentContext.marketplace_path` and adds the new marketplace-registration mechanism for plugin resolution.
- **Not yet done here:** deprecating the old `marketplace_path` parameters on skill-loading helpers / `/skills` request model.

On the **griffe / Python API breakage** workflow: it failed because the checker compares public SDK API against the PyPI baseline, saw the `Field(...)` value for `AgentContext.marketplace_path` change (specifically adding `deprecated=True`), and counted that as an `ATTRIBUTE_CHANGED_VALUE` break. The checker has an allowlist for metadata-only `Field(...)` changes (`description`, `title`, `examples`, `json_schema_extra`), but **not** `deprecated`, so it treats this as breaking and then fails SemVer because the package version did not get a minor bump.

My take: for this specific case, that workflow is being **too strict**. The field is still there and still works; we are only marking it deprecated. I would not consider that alone a breaking Python API change that should require a minor bump. So I think either:
- the checker should also treat `Field(..., deprecated=...)` as a non-breaking metadata/deprecation annotation, **or**
- if the intended policy really is “adding deprecation metadata to a public field counts as a break,” then the current CI result is expected — but that would be a much stricter policy than the behavior this PR is trying to preserve.

So short version: **yes, the issue says 4; this PR currently implements 1 of those deprecations, while the other 3 still look live / not migrated yet.**

enyst:
OpenHands-GPT-5.4 here — @enyst @jpshackelford, now that the small Griffe/`Field(..., deprecated=...)` CI fix is merged, I think there are **two viable paths** for this PR.

## First, the key observation

Right now, the new model is only **partially** replacing the old one:

- `registered_marketplaces` is used for **explicit plugin resolution** via `conversation.load_plugin(...)`
- but the old path still drives actual skill auto-loading:
  - `AgentContext._load_auto_skills()` -> `load_available_skills(..., marketplace_path=...)`
  - `load_available_skills()` -> `load_public_skills(marketplace_path=...)`
- `auto_load="all"` is modeled, but I do **not** see production code actually using `MarketplaceRegistry.get_auto_load_registrations()` to auto-load marketplace plugins at conversation start
- agent-server `/skills` still exposes `SkillsRequest.marketplace_path`, and I do not see `SkillsRequest.registered_marketplaces` added yet

So the question is really: should this PR stay scoped to the part that is already implemented, or should it do a bit more so the new path is actually exercised end-to-end and more deprecations become justified?

## Alternative 1: narrow the PR so the claims match the code

### What this means
Keep this PR focused on:
- `MarketplaceRegistration`
- `MarketplaceRegistry`
- plugin resolution (`plugin` / `plugin@marketplace`)
- `conversation.load_plugin()`
- tests/examples for explicit marketplace-backed plugin loading

### What I would do under this option
- **remove or soften the deprecation claims** in the PR description
- probably **do not deprecate additional APIs yet**
- possibly even reconsider whether `AgentContext.marketplace_path` should be deprecated *in this PR*, since its replacement does not yet cover the same startup/auto-load behavior end-to-end

### Pros
- smallest, cleanest PR
- avoids promising more migration than the code currently delivers
- keeps the deprecation story honest

### Cons
- leaves the full migration / auto-load rollout for a follow-up PR

## Alternative 2: do a bit more in this PR so the new path is used more broadly

### What “a bit more” would need to mean
If we want to deprecate more than just the field, I think this PR needs to wire the new model into the real startup path.

Concretely, I think that means:

1. **Make conversation/plugin startup honor `registered_marketplaces` with `auto_load="all"`**
   - at conversation startup, fetch the auto-load registrations
   - resolve/load all plugins declared by those marketplaces
   - merge their skills/hooks/MCP the same way explicit plugins are merged today

2. **Keep backward compatibility for legacy users**
   - if legacy `marketplace_path` is used and `registered_marketplaces` is empty, keep the current behavior working
   - deprecate the legacy field/path with warnings

3. **If we want to deprecate the server-side request field too, add the additive replacement now**
   - add `SkillsRequest.registered_marketplaces`
   - prefer it when present
   - keep `SkillsRequest.marketplace_path` working as the deprecated fallback

### What this would justify deprecating
Under this option, I think these become more defensible:
- `AgentContext.marketplace_path` -> `AgentContext.registered_marketplaces`
- `SkillsRequest.marketplace_path` -> `SkillsRequest.registered_marketplaces` **if** the new request field is added in this PR

### What I still would **not** deprecate yet
I would still be careful with:
- `load_public_skills(marketplace_path=...)`
- `load_available_skills(marketplace_path=...)`

Those are public SDK helpers and they are still the active implementation path today. Even if we extend this PR, I think the safer move is:
- either keep them for now,
- or later deprecate the **`marketplace_path` parameter / old code path**, not necessarily the whole helper function.

## My recommendation

I recommend **Alternative 1 unless you want to explicitly grow the PR into an end-to-end migration PR**.

If the goal is to keep this PR tight, then I would:
- keep the new marketplace registration / resolution work
- describe it as the **foundation** for auto-load semantics
- avoid deprecating the other 3 APIs yet
- maybe even narrow the current deprecation language so it does not imply full replacement already exists

If the goal is to make this PR the real migration point, then I would choose **Alternative 2**, but in that case I would do the extra wiring above before deprecating more. In particular, I would not deprecate `/skills` request fields or SDK helper paths without also shipping the additive replacement in the same PR.

## Short version

- **Alternative 1:** keep this PR narrower; deprecate less
- **Alternative 2:** wire `registered_marketplaces` into real startup/auto-load behavior, then deprecate more
- **My recommendation:** **Alternative 1**, unless you want to intentionally expand the implementation scope now

neubig (commented):
Could you please fix CI and re-request review? Thanks!

enyst:
@openhands Give us a view on how this PR works and tell WDYT. /codereview Post directly on github as a review with gh api.

openhands-sdk/openhands/sdk/context/agent_context.py
```diff
@@ -118,6 +135,15 @@ def _load_auto_skills(self):
         if not self.load_user_skills and not self.load_public_skills:
             return self
 
+        # Emit deprecation warning if using old marketplace_path field
+        if self.marketplace_path is not None and not self.registered_marketplaces:
```
enyst: The deprecation warning makes sense, but `_load_auto_skills()` is still entirely driven by the legacy `marketplace_path` flow. If a caller sets only `registered_marketplaces`, nothing here derives auto-loaded skills from those registrations, so `AgentContext` construction can still pull in the default public marketplace instead of the configured registrations. I think this validator needs to switch to the new field as the source of truth once registrations are present.

openhands-agent-server/openhands/agent_server/skills_router.py
```diff
@@ -143,6 +156,26 @@ def get_skills(request: SkillsRequest) -> SkillsResponse:
         org_repo_url = request.org_config.org_repo_url
         org_name = request.org_config.org_name
 
+    # Handle deprecation: prefer registered_marketplaces over marketplace_path
+    marketplace_path = request.marketplace_path
+    if request.registered_marketplaces:
+        # New behavior: use registered_marketplaces
+        # For now, we extract marketplace_path from the first 'public' registration
```
enyst, line 177: This compatibility bridge only forwards `repo_path` from the first `public` registration. `source` and `ref` are ignored, and if there is no `public` autoload registration we keep `request.marketplace_path`'s default. So `/skills` still can't honor the full `MarketplaceRegistration` contract here — especially for a forked/pinned public marketplace — even though the request uses the new field.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -1124,6 +1179,87 @@ def execute_tool(self, tool_name: str, action: Action) -> Observation:
...
+            + (f" @ {resolved_ref[:8]}" if resolved_ref else "")
+        )
+
+        # Merge plugin contents into agent
+        merged_context = plugin.add_skills_to(self.agent.agent_context)
+        merged_mcp = plugin.add_mcp_config_to(
+            dict(self.agent.mcp_config) if self.agent.mcp_config else {}
+        )
+
+        # Update agent and state atomically
```
enyst, line 1263: This only swaps in a new `agent_context` / `mcp_config` object. After initialization, though, the conversation has already built its tools and hook processor, so `load_plugin()` does not actually apply the full plugin semantics promised above: plugin hooks are not merged into `_hook_processor`, plugin agent defs are not registered, and MCP tools are not rebuilt from the updated config. In practice this looks like 'load skills after init', which seems materially different from the API contract.

enyst (commented):
Thanks — I dug through the issue, the PR, and the updated SDK/agent-server paths. My read is that the overall direction is good: `MarketplaceRegistration` + `MarketplaceRegistry` is the right shape, the lazy-fetch/caching model is straightforward, and the plugin resolution flow (`plugin` vs `plugin@marketplace`) looks reasonable.

That said, I don't think the new registration model is fully the source of truth yet. I found a few correctness gaps where the legacy `marketplace_path` flow is still driving behavior, or where `load_plugin()` doesn't apply the full plugin semantics it advertises. Left the concrete spots inline.

Because this touches plugin loading / agent behavior, I'd want a maintainer to take the final call after those are sorted and the intended end-to-end semantics are tightened up. WDYT?

openhands-sdk/openhands/sdk/context/agent_context.py
jpshackelford: Fixed in commit 2bfa3f5e. `AgentContext._load_auto_skills()` now uses `registered_marketplaces` as the source of truth when present:

- When `registered_marketplaces` is set, skills are loaded from marketplaces with `auto_load="all"` via the new `_load_skills_from_marketplaces()` method
- User skills are still loaded via the legacy path (they are local files)
- The legacy `marketplace_path` path is only used when `registered_marketplaces` is empty

This makes `registered_marketplaces` the true source of truth at the AgentContext construction level.

openhands-agent-server/openhands/agent_server/skills_router.py
jpshackelford: This was addressed in commit 4052e87. The compatibility bridge you see (extracting `repo_path`) is only for the legacy `load_public_skills` path.

The key fix is that `registered_marketplaces` is now passed through to the service:
```python
result = load_all_skills(
    ...
    registered_marketplaces=request.registered_marketplaces,
)
```

And `load_all_skills` calls `_load_marketplace_skills(registered_marketplaces)` which uses the full `MarketplaceRegistry` and properly honors `source`, `ref`, and `repo_path` for each registration.

The backward-compatibility bridge for `marketplace_path` is kept so old clients using the deprecated field still work, but new clients using `registered_marketplaces` get the full contract honored.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
jpshackelford: Fixed in commit 2bfa3f5e. `load_plugin()` now fully applies plugin semantics:

1. **Hooks** - Added `_merge_plugin_hooks()` method that merges plugin hooks into the existing hook processor (or creates one if none exists)

2. **Agent definitions** - Now calls `register_plugin_agents()` for any plugin agents

3. **Skills** - Already working via `plugin.add_skills_to()`

4. **MCP config** - Already working via `plugin.add_mcp_config_to()`

This now matches what `_ensure_plugins_loaded()` does for startup plugins. The full plugin contract is honored.

jpshackelford:
All review feedback has been addressed and the 3 inline review threads are now resolved.

**CI Status:** All required checks pass. The `[Optional] Docs example/check-examples` failure is a pre-existing issue about undocumented examples (e.g., `37_llm_profile_store`, `45_parallel_tool_execution`) that predates this PR and is unrelated to the marketplace registration changes.

enyst:
@openhands Give us a view on how this PR works and tell WDYT. /codereview Post directly on github as a review with gh api.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -1143,6 +1198,147 @@ def execute_tool(self, tool_name: str, action: Action) -> Observation:
...
+        # Create new agent first, then update both references together
+        new_agent = self.agent.model_copy(
```
enyst, line 1265: This still looks buggy once the conversation has already been initialized. `load_plugin()` swaps in `new_agent = self.agent.model_copy(...)`, but `model_copy()` preserves private attrs, so the copied agent keeps `_initialized=True` and the old `_tools` map. In practice that means any MCP servers added by the newly loaded plugin never flow back through `AgentBase._initialize()`, so runtime-loaded MCP tools stay stale/unavailable. I think this needs to either reuse the same init path as startup or explicitly reset/rebuild the initialized tool state here, plus a regression test that loads a plugin after init and checks the new tool appears.`

openhands-agent-server/openhands/agent_server/skills_router.py
enyst, line 177: This is still only a partial replacement for `SkillsRequest.marketplace_path`. When `registered_marketplaces` is provided, the router only extracts `repo_path` from the first `public` registration, and `load_all_skills()` still drives the base public catalog through `load_available_skills(... marketplace_path=...)`. So `source` / `ref` from the registration are ignored here, and `/skills` can still return the default `OpenHands/extensions` data even when the caller registered a different public marketplace source/branch. I’d make `/skills` fully registry-driven once registrations are present, or explicitly narrow the API contract until `source`/`ref` are honored end-to-end.`

enyst (commented):
Overall I like the direction here. `MarketplaceRegistration` + `MarketplaceRegistry` is a much clearer model than a single `marketplace_path`, and the lazy fetch/cache flow is straightforward to reason about. I also reran `uv run pytest tests/sdk/plugin/test_marketplace_registry.py tests/sdk/plugin/test_marketplace_registry_integration.py tests/agent_server/test_skills_service.py -q` locally and got 69 passing tests.

That said, I still see two correctness gaps in the current branch before I'd feel great about merging:

- runtime `conversation.load_plugin()` does not rebuild agent tool state after merging plugin MCP config
- agent-server `/skills` is only partially migrated to `registered_marketplaces` and still ignores `source`/`ref` for the `public` registration path

Since this changes plugin / skill loading behavior, I'd still want a human maintainer to do a lightweight behavior/eval sanity check before final merge.

_This review was generated by an AI assistant (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
enyst: @openhands Is this behavior newly added in this PR?

enyst: Yes — the problematic runtime behavior is newly introduced in this PR. The existing startup plugin-loading path also uses `model_copy()`, but it happens before agent initialization, so there isn't a stale initialized tool map yet. The new `load_plugin()` path can run after initialization, which is why it was preserving `_initialized`/`_tools` and leaving newly added MCP tools unavailable.

Fixed in 270226b4 by resetting and rebuilding the copied agent's runtime tool state when `load_plugin()` runs after startup, and I added a regression test that loads a marketplace plugin after initialization and checks the MCP init path runs again.

enyst: Reopening this thread since I pushed the proposed fix in https://github.com/OpenHands/software-agent-sdk/commit/270226b4 and then reverted it in https://github.com/OpenHands/software-agent-sdk/commit/46174855 so the PR author can review the idea in history and decide whether to adopt it.

jpshackelford:
### Adding: operator-level `OH_REGISTERED_MARKETPLACES` env var

Pushed commit `72d2f217` to this PR.

**Motivation.** A self-hosted / Enterprise customer running their own build of the agent-server image asked whether they could redirect public skill loading to an internal marketplace mirror without changes on the SaaS app-server side or any new admin UI. With this PR as-it-was, the answer is no: with a stock app-server, `SkillsRequest.registered_marketplaces` arrives empty, so the agent-server falls back to the hard-coded `OpenHands/extensions` URL. The customer's only option was a private monkey-patch on their custom image.

**Change.** A small, additive scope expansion that gives operators an org-wide default *without* changing the request-side contract:

- New `Config.registered_marketplaces: list[MarketplaceRegistration]` field, populated automatically by the existing `env_parser.from_env(Config, "OH")` machinery — same JSON-list pattern (`OH_REGISTERED_MARKETPLACES=[...]`) and sequential-suffix pattern (`OH_REGISTERED_MARKETPLACES_0_*`) that already work for `OH_WEBHOOKS`. No new parser code.
- `skills_router.get_skills` merges the server defaults with `request.registered_marketplaces` via a new `merge_marketplace_registrations` helper in `skills_service`. **Request-wins-by-name** semantics: any per-request registration whose `name` matches a server default replaces it; other server defaults are appended. This keeps a future SaaS admin-UI override path open while solving the "set it once on the image" case today.
- Defaults to `[]`, so behavior is unchanged for anyone not opting in.

**Relation to the unresolved review threads.** This is largely orthogonal to the `skills_router.py:163` `source`/`ref`-drop concern. It does, however, give operators a clean exit from the buggy legacy path: with `OH_REGISTERED_MARKETPLACES` set plus `load_public=False`, the legacy `marketplace_path` extraction never runs.

**Tests.** Adds 12 tests across `test_skills_service` (unit: merge helper semantics — request-only, server-only, disjoint, shadowing, ordering, no mutation), `test_env_parser` (operator pattern: JSON env-var → typed `Config`), and `test_skills_router` (integration: env → `Config` → router → `load_all_skills` forwarding). All 128 tests in the touched suites pass; `ruff`, `ruff format`, `pyright` clean.

Happy to split this into a follow-up PR after this one merges if you'd prefer to keep the diff here focused — just say the word.

_Posted by an AI agent (OpenHands) on behalf of the customer-support team._

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py`, lines 1291–1297] **Persistence**: `load_plugin()` records the newly loaded source only in `_resolved_plugins`, which is an in-memory reporting list. It never updates `_plugin_specs` or the agent-server's persisted `StoredConversation.plugins`. After the server restarts or recreates/evicts the event service, startup passes only `self.stored.plugins` back into `LocalConversation`, so a plugin accepted by `POST /api/conversations/{id}/plugins` disappears—including its skills, MCP configuration, subagents, and hooks. Persist the resolved immutable source as part of the conversation metadata (or explicitly reject runtime loading for durable conversations), and add a test that loads a plugin, reconstructs the conversation from disk, and proves the plugin behavior remains available.

**[RISK ASSESSMENT]**
This adds runtime loading of remote plugin code and hooks plus a public server endpoint, so it combines supply-chain exposure, persisted-state semantics, and a broad agent-capability blast radius. **Recommendation**: Do not auto-merge. Request review from a human architect/security reviewer to validate durable plugin identity, restart behavior, and trust boundaries.

**VERDICT:**
❌ **Needs rework**: A successful load endpoint must not silently produce a conversation with different capabilities after restart.

**KEY INSIGHT:**
Plugin identity is conversation state, not transient telemetry; persisting only the initial request makes runtime loads nondurable.

---

>
>
>

PR #2505, opened 2026-03-19
fix(sdk/conversation): inject nudge for reasoning-only monologue instead of STUCK

### LINKED ISSUE #2482: [Bug] Remote conversation got stuck: monologue detector false positive on extended thinking models
## Summary

Certain GAIA benchmark instances deterministically fail with `Remote conversation got stuck` across all retries, all SDK versions, and both Claude Sonnet 4.5 and Opus 4.6. The root cause is a conflict between the stuck detector's monologue threshold (3 consecutive agent `MessageEvent`s) and extended thinking models that produce reasoning-only responses (no text, no tool calls) before their first action.

## Affected instances

| Instance | Question | Sonnet (run 1) | Sonnet (run 2) | Sonnet (run 3) | Opus (run 4) | Opus (run 5) |
|---|---|---|---|---|---|---|
| `2d83110e` | Reversed text puzzle | STUCK | STUCK | STUCK | STUCK | STUCK |
| `2a649bb1` | EC numbers for virus testing | **PASS** | STUCK | STUCK | STUCK | STUCK |
| `6359a0b1` | Green polygon area | FAIL | FAIL | FAIL | STUCK | STUCK |
| `983bba7c` | Animals in alvei papers | PASS | PASS | FAIL | STUCK | PASS |

`2d83110e` fails in **100% of runs** (5/5). `2a649bb1` fails in 4/5. The Opus runs are more affected (up to 4 stuck instances vs 1–2 for Sonnet).

## Reproduction

Run the GAIA benchmark with Claude Sonnet 4.5 or Opus 4.6 with extended thinking enabled and OpenHands agent, Instance `2d83110e` will fail every time.

### The code path (`agent.py:298–382`)

1. The LLM returns a valid response (`chatcmpl-*` ID) containing only thinking/reasoning blocks — no text content, no tool calls.
2. `agent.py:302–309` detects `has_reasoning` and/or `has_content`:
   ```python
   has_reasoning = (
       message.responses_reasoning_item is not None
       or message.reasoning_content is not None
       or (message.thinking_blocks and len(message.thinking_blocks) > 0)
   )
   has_content = any(
       isinstance(c, TextContent) and c.text.strip() for c in message.content
   )
   ```
3. Since there are no tool calls, execution falls through to line 356. If neither reasoning nor content is found, a warning is logged. **But in all cases, a `MessageEvent` is emitted unconditionally** (lines 359–372):
   ```python
   msg_event = MessageEvent(source="agent", llm_message=message, ...)
   on_event(msg_event)  # Emitted even when empty
   ```
4. Since `has_content` is False, `execution_status` is not set to FINISHED (line 379–382), so the agent loop continues and calls the LLM again.
5. The LLM produces the same reasoning-only response. This repeats 3 times.

### The stuck detection (`stuck_detector.py:200–225`)

6. `_is_stuck_monologue()` counts consecutive `MessageEvent`s with `source="agent"`:
   ```python
   for event in reversed(events):
       if isinstance(event, MessageEvent):
           if event.source == "agent":
               agent_message_count += 1
   ```
7. The **monologue threshold is 3** (`types.py:40–41`). After 3 empty `MessageEvent`s, the conversation is marked `STUCK`.

### The error propagation (`remote_conversation.py:1035–1039`)

8. The remote client detects the STUCK status and raises `ConversationRunError("Remote conversation got stuck")`.

## Why these specific instances

These questions cause the model to reason internally without producing a tool call:

- **`2d83110e`**: Reversed text (`.rewsna eht sa "tfel" drow eht fo etisoppo eht etirw`). The model decodes it in thinking but doesn't know which tool to call for a pure-text answer — it just needs to call `finish`, but extended thinking fills the response with reasoning instead.
- **`2a649bb1`**: Multi-step research question. The model plans which searches to run but doesn't commit to a tool call in the first 3 responses.

## Evidence from conversation archives

Each retry shows the identical event sequence:

```
event-00000  SystemPromptEvent (agent)
event-00001  MessageEvent (user) — question delivered, content_len=1
event-00002  ConversationStateUpdateEvent — execution_status=running
event-00004  MessageEvent (agent) — content=[], thinking_blocks=[], resp_id=chatcmpl-...  [+2.0s]
event-00005  MessageEvent (agent) — content=[], thinking_blocks=[], resp_id=chatcmpl-...  [+1.6s]
event-00006  MessageEvent (agent) — content=[], thinking_blocks=[], resp_id=chatcmpl-...  [+1.6s]
event-00007  ConversationStateUpdateEvent — execution_status=stuck
```

Each empty response has a **unique `chatcmpl-*` ID** (real LLM calls, not cached), arrives in **~1.5–2s** (vs 4–94s for normal first responses), and the `thinking_blocks` field is `[]` in the archived events (likely stripped or encrypted).

## Why retries don't help

The retry mechanism escalates `resource_factor` (1→2→4→8), which increases pod CPU/memory. But the failure is in the model's response pattern, not resource exhaustion. Each retry gets a fresh runtime, fresh session, fresh conversation — but the same system prompt + question always produces the same model behavior.

### Option 1: Don't emit `MessageEvent` for empty responses (minimal change)

In `agent.py`, skip the `on_event(msg_event)` call when the response has no content and no reasoning. This prevents empty messages from counting toward the monologue threshold.

```python
# Line 356-372: only emit if there's something to emit
if has_reasoning or has_content:
    msg_event = MessageEvent(source="agent", llm_message=message, ...)
    on_event(msg_event)
```

### Option 2: Inject a nudge after empty responses (more robust)

When the LLM returns an empty response, append a system/user message reminding it to use a tool:

```python
if not has_reasoning and not has_content:
    logger.warning("LLM produced empty response - nudging to use tools")
    # Don't emit empty MessageEvent, instead continue loop
    # The next LLM call will include context about available tools
    continue
```

### Option 3: Exclude empty `MessageEvent`s from monologue detection (targeted fix)

In `stuck_detector.py:200–225`, skip `MessageEvent`s that have no content:

```python
if isinstance(event, MessageEvent):
    if event.source == "agent":
        # Don't count empty messages toward monologue
        if not event.llm_message or not event.llm_message.content:
            continue
        agent_message_count += 1
```

### Option 4: Increase monologue threshold for extended thinking (configuration)

A threshold of 3 is too aggressive for extended thinking models that may need several reasoning passes. Increasing it to 5–6 would give the model more room while still catching genuine monologue loops.

## Recommendation

**Option 1** is the safest minimal fix — an empty response with no content, no reasoning, and no tool calls provides no value as a `MessageEvent` and should not be emitted. Combined with **Option 4** (raising the threshold to 5 for extended thinking configs), this would eliminate the false positives without masking real stuck scenarios.

VascoSch92:
## Summary

Fixes #2482

Extended thinking models (Claude Sonnet 4.5, Opus 4.6, Nemotron-3 Super) can produce responses that contain only reasoning/thinking content without any text content or tool calls. Previously, after 3 consecutive such responses, the conversation was incorrectly marked as STUCK.

This PR differentiates between two types of agent monologue:

| Scenario | Pattern | Action |
|----------|---------|--------|
| **Reasoning-only response** | Has `reasoning_content` or `thinking_blocks`, but no text content | Inject a nudge prompt to guide the LLM to provide a final answer |
| **Truly empty response** | No reasoning, no content, no tool calls | Mark as STUCK (original behavior preserved) |

## Affected Instances (from issue)

| Instance | Question | Behavior Before | Behavior After |
|---|---|---|---|
| `2d83110e` | Reversed text puzzle | STUCK (100% of runs) | Nudge injected |
| `2a649bb1` | EC numbers for virus testing | STUCK (4/5 runs) | Nudge injected |

### `stuck_detector.py`
- Modified `_is_stuck_monologue()` to skip reasoning-only messages (handled separately)
- Added `needs_reasoning_nudge()` method to detect consecutive reasoning-only responses
- Added helper methods `_message_has_content()` and `_message_has_reasoning()`

### `local_conversation.py`
- Check `needs_reasoning_nudge()` before `is_stuck()` in the run loop
- Inject a user message nudging the LLM: *"Please provide your final answer now using the appropriate tool. If you have a text answer, use the finish tool to submit it."*

## Test Plan

- Added `test_reasoning_only_monologue_triggers_nudge` - verifies reasoning-only messages trigger nudge, not stuck
- Added `test_empty_monologue_triggers_stuck_not_nudge` - verifies truly empty messages still trigger stuck
- All existing stuck detector tests pass

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:6d19d64-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-6d19d64-python \
  ghcr.io/openhands/agent-server:6d19d64-python
```

```
ghcr.io/openhands/agent-server:6d19d64-golang-amd64
ghcr.io/openhands/agent-server:6d19d64-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:6d19d64-golang-arm64
ghcr.io/openhands/agent-server:6d19d64-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:6d19d64-java-amd64
ghcr.io/openhands/agent-server:6d19d64-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:6d19d64-java-arm64
ghcr.io/openhands/agent-server:6d19d64-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:6d19d64-python-amd64
ghcr.io/openhands/agent-server:6d19d64-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:6d19d64-python-arm64
ghcr.io/openhands/agent-server:6d19d64-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:6d19d64-golang
ghcr.io/openhands/agent-server:6d19d64-java
ghcr.io/openhands/agent-server:6d19d64-python
```

- Each variant tag (e.g., `6d19d64-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `6d19d64-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/stuck_detector.py
+++ b/openhands-sdk/openhands/sdk/conversation/stuck_detector.py
@@ -291,7 +291,7 @@ class StuckDetector:
         msg = event.llm_message
         return bool(
             msg.reasoning_content is not None
-            or (msg.thinking_blocks and len(msg.thinking_blocks) > 0)
+            or msg.thinking_blocks
             or msg.responses_reasoning_item is not None
         )
 


enyst (changes requested):
Thank you for looking into this!

What if we use a hook for this, WDYT?

VascoSch92:
> Thank you for looking into this!
> 
> What if we use a hook for this, WDYT?

@enyst 

Yes, it's possible, but there are trade-offs to consider.                                                                                                                                                                                                           
                                 
The hook system already supports 6 event types (`PRE_TOOL_USE`, `POST_TOOL_USE`, `USER_PROMPT_SUBMIT`, `SESSION_START`, `SESSION_END`, `STOP`). You could add a new hook type like `PRE_AGENT_STEP` or `REASONING_NUDGE` that:                 
  1. Fires before each agent.step() call in the run loop (local_conversation.py:636-670)                                                                                                                                                 
  2. Passes the current events as context (e.g., via stdin JSON)                                                                                                                                                                         
  3. The hook script decides whether to inject a nudge and what message to use                                                                                                                                                           
  4. Returns the nudge text (or nothing) via stdout                                                                                                                                                                                      
           
  This would make the nudge message customizable — users could swap the generic "Please provide your final answer now..." for something domain-specific.                                                                                 
                                                                                                                                                                                                               
  Pros:
  - Customizable nudge message per deployment
  - Consistent with the existing extensibility model                                                                                                                                                                                     
  - Detection logic could still live in StuckDetector, hook just controls the response

  Cons:                                                                                                                                                                                                                                  
  - The current nudge is a core SDK behavior (not really user-configurable policy) — hooks are designed for external scripts                                                                                                             
  - Adds latency (subprocess spawn per check)                                                                                                                                                                                            
  - Every user would need to configure the hook to get this behavior, vs it working out of the box
  - The STOP hook is the closest analogy, but that blocks an explicit agent action — this is more of a "pre-step inspection"

enyst:
Good points! Re: this
> The current nudge is a core SDK behavior (not really user-configurable policy) — hooks are designed for external scripts

It's a bit confusing to me _which_ behavior. According to the issue, what may have happened here was one of two things:
(Case 1, litellm bug)
- litellm sends us 0 reasoning, 0 content, 0 tool calls
- because it cut off the thinking from the provider, as you found; so we need litellm fix
- in the meantime, this is a totally empty message for us, but the LLM was trying to do reasoning.

(Case 2, reasoning only)
- we get from litellm reasoning content, e.g. 116 reasoning tokens, 0 content, 0 tool calls
- ...?

Your agent said it was case 1. Was it, are we trying to fix case 1 here?

VascoSch92:
> Good points! Re: this
> 
> > The current nudge is a core SDK behavior (not really user-configurable policy) — hooks are designed for external scripts
> 
> It's a bit confusing to me _which_ behavior. According to the issue, what may have happened here was one of two things: (Case 1, litellm bug)
> 
>     * litellm sends us 0 reasoning, 0 content, 0 tool calls
> 
>     * because it cut off the thinking from the provider, as you found; so we need litellm fix
> 
>     * in the meantime, this is a totally empty message for us, but the LLM was trying to do reasoning.
> 
> 
> (Case 2, reasoning only)
> 
>     * we get from litellm reasoning content, e.g. 116 reasoning tokens, 0 content, 0 tool calls
> 
>     * ...?
> 
> 
> Your agent said it was case 1. Was it, are we trying to fix case 1 here?

I understand the confusion. This issue actually covers two separate problems with the same root: mine and @juanmichelini’s.

In my case, the LLM is returning nothing (a LiteLLM bug), whereas Juan is seeing the LLM return the answer inside the reasoning block. (Juan, feel free to add more details there).

This PR specifically targets the latter issue.

juanmichelini:
@VascoSch92 @enyst  more details:

Qwen3.5-flash and Nemotron 3 super both present cases with reasoning content but not message content.
This plummets the Qwen3.5 results in particular: https://github.com/OpenHands/openhands-index-results/issues/714 Nemotron 3 super is probably handicapped by this too.

Nudging should show better results, closer to reality for both. 

I'm testing a similar nudge strategy https://github.com/OpenHands/software-agent-sdk/pull/2512 but it's very Qwen specific, maybe we can merge this. And I'll create a current PR to keep things clean.

VascoSch92:
@juanmichelini If your fix is working, we can use it and I will close this PR. 
It seems that your approach is much cleaner than mine.

enyst:
Looking at this:
```
{
  "kind": "MessageEvent",
  "source": "agent",
  "llm_message": {
    "content": [{"text": ""}],                          // ← empty!
    "reasoning_content": "Let me look at skipping.py:\n\n<tool_call>\n<function=file_editor>\n<parameter=command>view</parameter>\n<parameter=path>/workspace/pytest/src/_pytest</parameter>\n</function>\n</tool_call>"   // ← tool call is HERE instead
  }
}
```
🤔 This is Qwen. I have a question: do we send reasoning content back to the LLM in the Qwen case? Should we / shouldn't we?

juanmichelini:
@VascoSch92 got it, we merged the other one so closing this one

PR #2511, opened 2026-03-19
refactor(tools): built-in agents

VascoSch92:
## Summary
This PR changes the built-in subagents available to the main agent.

## Motivation
After various rounds of evaluation, a few issues came up:
- default was not a good name for the default agent — the main agent had trouble knowing when to call it.
- default_cli and default were problematic. In fact, they are essentially the same agent, the only difference being that one has web browser access.
- The main agent had trouble calling the correct agent.

**Solution**

- default is renamed to general_purpose and the prompt is updated.
- Separation between the default agent and a specific agent for web browsing. This way, we can give more guidance to the web browser agent on how to accomplish its task.
- explore was misleading for a web browser agent, so the name was changed to code-explorer. Same for bash.

## Other
- Renamed cli_mode parameter to enable_browser with inverted semantics:
    -  the code is much easier to understand
    - Old: cli_mode=False → default agent included browser tools
    - New: enable_browser=True → registers a dedicated web researcher agent with browser tools                                                                                                                                           
  - The general-purpose agent no longer gets browser tools regardless of configuration. Browser capabilities are now isolated in the web researcher agent.                                                                               
  - Added test_general_purpose_has_no_browser_tools to document this intentional architectural change.      
## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:e46fbb0-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-e46fbb0-python \
  ghcr.io/openhands/agent-server:e46fbb0-python
```

```
ghcr.io/openhands/agent-server:e46fbb0-golang-amd64
ghcr.io/openhands/agent-server:e46fbb0-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:e46fbb0-golang-arm64
ghcr.io/openhands/agent-server:e46fbb0-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:e46fbb0-java-amd64
ghcr.io/openhands/agent-server:e46fbb0-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:e46fbb0-java-arm64
ghcr.io/openhands/agent-server:e46fbb0-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:e46fbb0-python-amd64
ghcr.io/openhands/agent-server:e46fbb0-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:e46fbb0-python-arm64
ghcr.io/openhands/agent-server:e46fbb0-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:e46fbb0-golang
ghcr.io/openhands/agent-server:e46fbb0-java
ghcr.io/openhands/agent-server:e46fbb0-python
```

- Each variant tag (e.g., `e46fbb0-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `e46fbb0-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/tool_router.py
+++ b/openhands-agent-server/openhands/agent_server/tool_router.py
@@ -13,7 +13,7 @@ from openhands.tools.preset.planning import register_planning_tools
 
 tool_router = APIRouter(prefix="/tools", tags=["Tools"])
 register_default_tools(enable_browser=True)
-register_builtins_agents(cli_mode=False)
+register_builtins_agents(enable_browser=True)
 register_gemini_tools(enable_browser=True)
 register_planning_tools()
 
--- a/tests/tools/task/test_task_manager.py
+++ b/tests/tools/task/test_task_manager.py
@@ -185,9 +185,7 @@ class TestTaskManager:
         manager, _ = _manager_with_parent(tmp_path)
         register_builtins_agents()
 
-        task = manager._create_task(
-            subagent_type="general-purpose", description=None, max_turns=None
-        )
+        task = manager._create_task(subagent_type="general-purpose", description=None)
         assert task.id in manager._tasks
         assert isinstance(manager._tasks[task.id].conversation_id, uuid.UUID)
 
@@ -258,9 +256,7 @@ class TestTaskManager:
         register_builtins_agents()
 
         # Create and evict a task (simulating a completed first run)
-        task = manager._create_task(
-            subagent_type="general-purpose", description=None, max_turns=None
-        )
+        task = manager._create_task(subagent_type="general-purpose", description=None)
         original_id = task.id
         original_uuid = task.conversation_id
         manager._evict_task(task)
@@ -566,9 +562,7 @@ class TestStartTask:
         register_builtins_agents()
 
         # Create and evict a task to simulate a prior completed run
-        first = manager._create_task(
-            subagent_type="general-purpose", description=None, max_turns=None
-        )
+        first = manager._create_task(subagent_type="general-purpose", description=None)
         original_id = first.id
         manager._evict_task(first)
 


openhands-tools/openhands/tools/preset/default.py
```diff
@@ -88,36 +88,30 @@ def get_default_agent(
...
-    else:
+    # Filter out browser-dependent agents when browser is not available
+    if not enable_browser:
+        _browser_only_agents = {"web researcher"}
         builtins_agents_def = [
-            agent for agent in builtins_agents_def if agent.name != "default cli mode"
+            agent
+            for agent in builtins_agents_def
+            if agent.name not in _browser_only_agents
         ]
```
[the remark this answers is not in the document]
VascoSch92: You are suggesting to me my exact code :-)

openhands-tools/openhands/tools/preset/default.py
```diff
@@ -88,36 +88,30 @@ def get_default_agent(
     return agent
 
 
-def register_builtins_agents(cli_mode: bool = False) -> list[str]:
+def register_builtins_agents(enable_browser: bool = True) -> list[str]:
```
enyst, line 91: @openhands This PR changes cli_mode False to enable_browser True. Is there another occurrence(s) of a “cli_mode” param, var, etc in the codebase? If yes, WDYT about this change?

enyst: Yes — there are still several `cli_mode` occurrences in the codebase. The main ones are the preset factory APIs (`get_default_agent`, `get_gpt5_agent`, `get_gemini_agent`), plus prompt plumbing/tests/examples that pass `system_prompt_kwargs={"cli_mode": ...}`.

My take: `enable_browser` is a good rename for `register_builtins_agents()` specifically, because that function now only decides whether browser-dependent built-ins (notably `web-researcher`) are registered. But I would *not* do a blanket `cli_mode -> enable_browser` rename in this PR. `cli_mode` still has a broader meaning in the rest of the codebase: it affects prompt behavior as well, not just browser availability (for example `security_risk_assessment.j2` branches on it).

So I think the current code change is fine, but the PR description should be read as scoped to builtin-agent registration rather than a repo-wide rename. If we want repo-wide consistency later, that should probably be a separate, compatibility-minded follow-up because it would touch public preset helpers, examples, and tests.

enyst: Looking at it from the CLI-app direction, I think your concern is valid: **no, `cli_mode` is not automatically propagated to `register_builtins_agents()` today**.

What currently happens is:
- `get_default_agent(llm, cli_mode=...)` uses it in **two** places in `openhands-tools/openhands/tools/preset/default.py`:
  1. `get_default_tools(enable_browser=not cli_mode)` → affects the main agent tool list
  2. `system_prompt_kwargs={"cli_mode": cli_mode}` → flows into `Agent.system_prompt_kwargs`, then `AgentBase.static_system_message()`, and from there into prompt templates like `openhands-sdk/openhands/sdk/agent/prompts/security_risk_assessment.j2`

- `register_builtins_agents(...)` is separate. It only affects which builtin subagent factories are registered globally (notably whether `web-researcher` exists). That function is not called automatically from `get_default_agent()`, and the delegate executor later just does `get_agent_factory(...)`; it does not know about `cli_mode`.

So if a CLI built on the SDK wants everything aligned, the caller currently has to do **both**:
- `get_default_agent(..., cli_mode=True)`
- `register_builtins_agents(enable_browser=False)`

I agree that this `not cli_mode` mapping is less obvious and more error-prone at the integration layer. So while `enable_browser` is a clearer *internal meaning* for builtin registration, I’m less convinced it is the best *external API* if the common entrypoint is ‘I am in CLI mode’.

My preference would be either:
1. keep/accept `cli_mode` on `register_builtins_agents()` as the public-facing knob (possibly mapping internally to `enable_browser`), or
2. support both for a transition period, with one deprecated,
so CLI callers can set one concept consistently instead of remembering the inversion here.

VascoSch92:
@enyst :-) Would be nice to merge that. The subagents that we have now are not so good. And I would like to merge that before we have TaskToolSet in cloud. (because of backward compatibility).

Do you have blockers or things i should change?

Issue #2467, opened 2026-03-16
SDK should auto-resize oversized images before sending to LLM providers


xingyaoww:
## Bug Description

When a user sends an oversized image in a conversation, the LLM API call fails with:

```
litellm.BadRequestError: Error code: 400 - AnthropicException -
{"type":"error","error":{"type":"invalid_request_error",
"message":"messages.4.content.1.image.source.base64.data:
At least one of the image dimensions exceed max allowed size
for many-image requests: 2000 pixels"}}
```

The error crashes the conversation with no recovery, even though the runtime is still alive.

## Root Cause

The `ImageContent` class in `openhands-sdk/openhands/sdk/llm/message.py` passes image URLs/base64 data directly to the LLM provider without any dimension validation or resizing. Anthropic enforces a **2000px max dimension** for many-image requests, and other providers have similar limits.

```python
class ImageContent(BaseContent):
    type: Literal["image"] = "image"
    image_urls: list[str]

    def to_llm_dict(self) -> list[dict[str, str | dict[str, str]]]:
        images: list[dict[str, str | dict[str, str]]] = []
        for url in self.image_urls:
            images.append({"type": "image_url", "image_url": {"url": url}})
        # No dimension check or resize happens here
        return images
```

## Expected Behavior

The SDK should automatically downscale images that exceed provider dimension limits before including them in LLM API calls. This should happen transparently so users can attach any image without hitting cryptic API errors.

## Proposed Solution

1. **Add image resizing logic** — When converting `ImageContent` to LLM format (in `to_llm_dict()` or at the message serialization layer), detect base64-encoded images, decode them, check dimensions, and resize if any dimension exceeds the max (e.g., 2000px for Anthropic, or a conservative default).
2. **Make the limit configurable** — The max dimension could be tied to the LLM provider config or exposed as a setting, since different providers have different limits.
3. **Handle both base64 and URL images** — For base64 images, resize in-memory. For URL-based images, consider fetching and resizing, or at minimum validate and warn.
4. **Preserve aspect ratio** — When downscaling, maintain the original aspect ratio.

## Reproduction Steps

1. Start a conversation using an Anthropic model (e.g., `claude-opus-4-6`)
2. Send a message with an attached image where at least one dimension exceeds 2000px
3. Observe the `litellm.BadRequestError` crash

## Environment

- Model: `claude-opus-4-6` (via LiteLLM)
- Error source: Anthropic API `invalid_request_error`

[...]

PR #2552, opened 2026-03-23
fix(sdk): resize Anthropic many-image inputs

Zheng-Lu:
#2467 

## Summary
Reproduces and fixes the Anthropic many-image failure by resizing oversized base64 images during LLM message formatting.

## What Changed
- Added an Anthropic-only resize path in `LLM.format_messages_for_llm`
- Resize only triggers when the outgoing request crosses the many-image threshold
- Preserves aspect ratio and leaves URL images unchanged
- Added `pillow` as a runtime dependency for in-memory image resizing

## Validation
- `pytest tests/sdk/llm/test_llm_image_resizing.py` Passed
- `pytest tests/sdk/llm/test_llm_image_resizing.py tests/sdk/llm/test_vision_support.py` Passed
- `ruff check openhands-sdk/openhands/sdk/llm/llm.py tests/sdk/llm/test_llm_image_resizing.py` Passed
- `pyright openhands-sdk/openhands/sdk/llm/llm.py tests/sdk/llm/test_llm_image_resizing.py` Passed

## Proof
Now the multiple-images request with at least one image > 2000px doesn't throw the error litellm.BadRequestError
image


xingyaoww:
@OpenHands pls merge from main, resolve all conflicts. Then do /codereview-roasted /github-pr-review

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -73,6 +75,7 @@
     supports_vision,
     token_counter,
 )
+from PIL import Image
```
xingyaoww: 🔴 **Critical — Hard PIL import at module level**

This makes `pillow` a **hard runtime dependency for every single import of the SDK**, even if the user never sends an image. If PIL isn't installed (or has a version conflict in a constrained environment), the entire LLM module blows up.

This should be a **lazy import** inside `_resize_base64_data_image_url` (the only method that actually uses it):

```python
# Inside _resize_base64_data_image_url:
try:
    from PIL import Image
except ImportError:
    logger.warning("pillow not installed; skipping image resize")
    return url
```

Or at minimum, guard it behind a `TYPE_CHECKING` block and import lazily at call site.

openhands-sdk/pyproject.toml
```diff
@@ -12,6 +12,7 @@ dependencies = [
     "filelock>=3.20.1",
     "httpx>=0.27.0",
     "litellm==1.80.10",
+    "pillow>=12.1.1",
```
xingyaoww, line 15: 🔴 **Critical — Pillow as a hard core dependency**

`pillow` is a heavyweight native library (~30MB+ with C extensions) that introduces platform-specific build concerns. Adding it as a **required** dependency of the core SDK for a narrow, Anthropic-only, many-image edge case is disproportionate.

Consider making it an **optional dependency**:

```toml
[project.optional-dependencies]
pillow = ["pillow>=12.1.1"]
```

Then guard the import with a try/except and degrade gracefully (skip resize, log a warning). Users who need the feature install `openhands-sdk[pillow]`.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -1259,6 +1265,83 @@ def _apply_prompt_caching(self, messages: list[Message]) -> None:
...
+            resized_image.save(buffer, format=image_format)
+        except Exception:
+            logger.warning(
+                "Failed to resize base64 data image for outgoing LLM request",
+                exc_info=True,
+            )
+            return url
+
+        resized_encoded = base64.b64encode(buffer.getvalue()).decode("ascii")
+        return f"data:{mime_type};base64,{resized_encoded}"
```
xingyaoww: 🟠 **Important — 80 lines of image manipulation don't belong in LLM**

`llm.py` is already 1500+ lines. These three methods (`_apply_outgoing_image_resize`, `_get_outgoing_image_max_dimension`, `_resize_base64_data_image_url`) are pure image-processing utilities with zero dependency on `self` state (one is already a `@staticmethod`, the other two only call `_infer_litellm_provider()`).

Extract to a standalone module, e.g. `openhands/sdk/llm/utils/image_resize.py`:

```python
def resize_base64_data_url(url: str, *, max_dimension: int) -> str: ...
def maybe_resize_images(messages, provider, vision_enabled): ...
```

Then the LLM method becomes a one-liner call. Keep the god-class from getting godlier.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -1259,6 +1265,83 @@ def _apply_prompt_caching(self, messages: list[Message]) -> None:
...
+        try:
+            raw_bytes = base64.b64decode(encoded)
+            with Image.open(io.BytesIO(raw_bytes)) as image:
+                if max(image.size) <= max_dimension:
+                    return url
+
+                resized_image = image.copy()
+                resized_image.thumbnail(
+                    (max_dimension, max_dimension), Image.Resampling.LANCZOS
+                )
```
xingyaoww: 🟡 **Suggestion — Unnecessary `image.copy()`**

`image.thumbnail()` operates in-place. You're creating a copy solely to call thumbnail on it, but the original `image` object is never used again after the size check on line 1318. Just call `image.thumbnail(...)` directly on the original — saves an allocation of a full image buffer.

```suggestion
                image.thumbnail(
                    (max_dimension, max_dimension), Image.Resampling.LANCZOS
                )
                resized_image = image
                image_format = image.format or mime_type.split("/", 1)[1].upper()
```

Or even simpler: drop `resized_image` entirely and use `image` throughout.

xingyaoww (commented):
## Taste Rating: 🟡 Acceptable — Works, but the structure needs improvement

**Linus's Three Questions:**
1. *Is this solving a real problem?* — Yes. Anthropic's many-image limit is a real production failure.
2. *Is there a simpler way?* — Yes. This is ~80 lines of image manipulation code jammed into a 1500-line god-class. Extract it.
3. *What will this break?* — Adding `pillow` as a **hard** runtime dependency to the core SDK is the biggest concern. Every user now pays for PIL whether they use images or not.

---

**VERDICT:**
❌ **Needs rework** — The fix is directionally correct, but the dependency strategy and code placement need redesign before merging.

**KEY INSIGHT:**
The core problem is treating PIL as a hard SDK dependency and stuffing image-processing plumbing into the LLM class, when this should be a lazy-loaded utility module.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -127,6 +131,8 @@
 # This cap prevents requesting output that exceeds the context window.
 # 16384 is a safe default that works for most models (GPT-4o: 16k, Claude: 8k).
 DEFAULT_MAX_OUTPUT_TOKENS_CAP: Final[int] = 16384
+ANTHROPIC_MANY_IMAGE_THRESHOLD: Final[int] = 20
+ANTHROPIC_MANY_IMAGE_MAX_DIMENSION: Final[int] = 2000
```
xingyaoww: 🟡 **Suggestion — Magic numbers with no source reference**

Where do `20` and `2000` come from? Anthropic docs? Empirical testing? A comment linking to the source would save future maintainers from guessing:

```python
# Anthropic limits per-image size when requests contain many images.
# See: https://docs.anthropic.com/... (or link to issue #2467)
ANTHROPIC_MANY_IMAGE_THRESHOLD: Final[int] = 20
ANTHROPIC_MANY_IMAGE_MAX_DIMENSION: Final[int] = 2000
```

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -1259,6 +1265,83 @@ def _apply_prompt_caching(self, messages: list[Message]) -> None:
...
+        if not sep:
+            return url
+
+        mime_type = header.removeprefix("data:")
+
+        try:
+            raw_bytes = base64.b64decode(encoded)
+            with Image.open(io.BytesIO(raw_bytes)) as image:
+                if max(image.size) <= max_dimension:
+                    return url
```
xingyaoww: 🟡 **Suggestion — Every image pays the base64-decode + PIL-open tax**

Even images that are already ≤ 2000px get fully decoded from base64, opened as a PIL Image, and then returned unchanged. For 21+ screenshots, that's a non-trivial amount of CPU and memory churn in the hot path.

Consider a lighter pre-check — you could peek at PNG/JPEG headers to get dimensions without fully decoding the image data (PIL's `Image.open()` with lazy loading only reads the header, but `base64.b64decode` still decodes the entire payload). Not a blocker, but worth noting for future optimization.

tests/sdk/llm/test_llm_image_resizing.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _format_for_provider(
+    llm: LLM, messages: list[Message], *, provider: str
+) -> list[dict]:
+    with (
+        patch.object(LLM, "vision_is_active", return_value=True),
+        patch.object(LLM, "_infer_litellm_provider", return_value=provider),
+    ):
+        return llm.format_messages_for_llm(messages)
```
xingyaoww, line 42: 🟡 **Suggestion — Missing test for non-Anthropic providers**

There's no test verifying that OpenAI/other providers are **not** affected by the resize logic. Add a test like:

```python
def test_openai_many_images_are_not_resized():
    url = _make_png_data_url(2400, 1200)
    message = Message(role="user", content=[ImageContent(image_urls=[url] * 25)])
    llm = LLM(model="gpt-4o", api_key=SecretStr("k"), usage_id="t")
    formatted = _format_for_provider(llm, [message], provider="openai")
    assert _data_url_dimensions(_image_urls_from_chat_message(formatted[0])[0]) == (2400, 1200)
```

This guards the `if provider != 'anthropic'` branch against regressions.

openhands-sdk/pyproject.toml
xingyaoww: HUMAN: I wonder if there are ways to resize the image without usng PIL?

openhands-sdk/openhands/sdk/llm/llm.py
xingyaoww: HUMAN: ^agree with the judgement here

openhands-sdk/pyproject.toml
Zheng-Lu: Yes, alternatives exist, but they are not obviously better here:

- `pyvips` / `opencv-python`: heavier and more complex than Pillow
- `wand`: depends on ImageMagick being installed
- external commands like convert: not appropriate inside the SDK hot path
- custom pure-Python resizing: impractical for encoded PNG/JPEG and worse to maintain

none are clearly lighter or cleaner for this SDK than a lazy, optional PIL-based implementation. So I think I would not try to replace PIL in this PR

Advise from Extra High reasoning GPT-5.4

openhands-sdk/openhands/sdk/llm/llm.py
xingyaoww: @Zheng-Lu is this fixed in any commit? 👀

openhands-sdk/pyproject.toml
xingyaoww: @OpenHands look at https://github.com/OpenHands/software-agent-sdk/pull/2552#discussion_r3003668739 - tell me if this is true? Only comment under this thread, do NOT edit any code. Can you search around to see if there's indeed no workarounds?

If so, can you estimate on how much (in percentage) it would increase our package size

xingyaoww: I investigated this thoroughly. Here are my findings:

## Are there workarounds to avoid Pillow?

**Short answer: No practical ones for actual image resizing.** The assessment in the previous comment is essentially correct.

Here is what I checked:

| Alternative | Why it doesn't work |
|---|---|
| **Pure Python resize** | Python's stdlib has zero image processing. Decoding PNG requires implementing zlib decompression + PNG filtering + color-type handling. JPEG decoding is even harder (Huffman + DCT). Then you'd need to implement a resampling algorithm on raw pixels. Hundreds of lines of fragile, slow code. |
| **`pypng`** | Can read/write PNG only (no JPEG), and has **no resize** functionality — you'd implement resampling yourself. |
| **`pyvips`** | Lower memory at runtime but the wheel itself is even larger (~20+ MB) with native C deps. |
| **`opencv-python`** | ~60 MB — far heavier than Pillow. |
| **`wand`** | Requires ImageMagick to be system-installed. |
| **Subprocess (`convert`, `ffmpeg`)** | Requires system binaries; inappropriate inside an SDK library. |
| **Anthropic Files API** | I checked the [Anthropic Vision docs](https://platform.claude.com/docs/en/build-with-claude/vision). They explicitly state: *"Even when using the Files API, requests with many large images can fail before reaching the 600-image count. Reduce image dimensions or file sizes (for example, by downsampling) before uploading."* So the Files API does **not** bypass the 2000px dimension limit. |
| **Reduce JPEG quality only** | The constraint is pixel *dimensions* (2000px), not file size. Re-encoding at lower quality won't change dimensions. |
| **Let it fail and handle the error** | Bad UX — the user's request just silently fails or requires manual intervention. |

## Package size impact

I measured the actual installed sizes of all SDK direct dependencies:

| Metric | Value |
|---|---|
| Pillow installed size (Linux x86_64) | **18.55 MB** |
| Pillow wheel download size | **6.8 MB** |
| Current SDK direct deps total installed | **40.83 MB** |
| Pillow as % of current direct deps | **~45%** |
| Pillow as % of (current deps + Pillow) | **~31%** |

For context, `litellm` alone is 34.09 MB (83% of the current SDK deps total), so Pillow would be the second-largest direct dependency.

## One important mitigating factor

`openhands-tools` depends on `browser-use>=0.8.0`, which **already hard-depends on `pillow>=11.2.1`**. So in practice, anyone installing the full OpenHands stack (SDK + tools) already has Pillow in their environment. Adding it to `openhands-sdk` would not increase their actual install size at all.

That said, for users who install `openhands-sdk` standalone (without tools), it would be a ~31% increase to the direct dependency footprint — significant for a feature that only applies to Anthropic many-image requests.

## Recommendation

Pillow is genuinely the right tool for this job — there is no lighter alternative that handles both PNG and JPEG resize correctly. But it should be an **optional dependency** with a lazy import + graceful degradation, as you originally suggested. Users of the full stack already have it; SDK-only users can opt in with `pip install openhands-sdk[pillow]`.

xingyaoww: > Key mitigating factor: openhands-tools → browser-use already hard-depends on Pillow, so the full OpenHands stack already has it. The size increase only affects standalone openhands-sdk users.

HUMAN: given this, let's do this! Just need to fix this: https://github.com/OpenHands/software-agent-sdk/pull/2552#discussion_r2999314780

openhands-sdk/openhands/sdk/llm/llm.py
Zheng-Lu: Yes, fixed in last commit.

I have moved the image-resize logic out of `llm.py` into `openhands/sdk/llm/utils/image_resize.py`, including the dimension selection and base64 resize helper.

xingyaoww: @Zheng-Lu did you push it up to this PR? I didn't see the changes here..

Zheng-Lu: @xingyaoww  Sorry, I thought I did but I actually didn't, now the change is committed

xingyaoww:
@OpenHands do /codereview-roasted /github-pr-review

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
```diff
@@ -0,0 +1,122 @@
...
+    except ImportError:
+        logger.warning(
+            "pillow is not installed; skipping Anthropic image resizing. "
+            "Install openhands-sdk[pillow] to enable base64 image downscaling."
+        )
+        return None
+
+    return Image
+
+
```
xingyaoww: 🟠 Important: **The `image_module: Any` threading pattern is unnecessary complexity.** Python caches module imports in `sys.modules` — after the first `from PIL import Image`, every subsequent call is a `dict.__getitem__` lookup (essentially free). The entire `_load_pillow_image_module()` function, the `image_module` parameter on `resize_base64_data_url`, and the caller having to thread it through — all of this exists to "avoid repeated imports" that Python already optimizes away.

Simpler design: just do the guarded import inside `resize_base64_data_url` itself, or at the top of `maybe_resize_messages_for_provider`. Kill `_load_pillow_image_module()` entirely. Three functions become two, the `Any` parameter disappears, and you get actual type safety on the PIL `Image` module.

```python
# Before (3 functions, Any threading):
module = _load_pillow_image_module()  # returns Any
resize_base64_data_url(url, max_dimension=d, image_module=module)

# After (2 functions, typed):
try:
    from PIL import Image
except ImportError:
    logger.warning(...)
    return
# Use Image directly — fully typed, no Any
```

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
```diff
@@ -0,0 +1,122 @@
...
+                content_item.image_urls = [
+                    resize_base64_data_url(
+                        url,
+                        max_dimension=max_dimension,
+                        image_module=image_module,
+                    )
+                    for url in content_item.image_urls
+                ]
+
+
```
xingyaoww: 🟠 Important: **Silent in-place mutation is fragile.** `maybe_resize_messages_for_provider` mutates `content_item.image_urls` and returns `None`. It works today only because `format_messages_for_llm` does a `deepcopy` on line 1380 before calling this — but that coupling is invisible at the call site.

If anyone calls `maybe_resize_messages_for_provider` from a different path without deepcopying first (e.g., `format_messages_for_responses`), they'll silently corrupt the caller's `Message` objects.

Two options to fix:
1. **Return new messages** instead of mutating in place (functional style, safer)
2. **Document the mutation contract** in the docstring and rename to something like `resize_messages_in_place_for_provider` so the side effect is impossible to miss

Option 1 is better taste — eliminates the special case where the caller has to know about the mutation.

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
```diff
@@ -0,0 +1,122 @@
...
+            "pillow is not installed; skipping Anthropic image resizing. "
+            "Install openhands-sdk[pillow] to enable base64 image downscaling."
+        )
+        return None
+
+    return Image
+
+
+def resize_base64_data_url(url: str, *, max_dimension: int, image_module: Any) -> str:
+    if not url.startswith("data:image/"):
```
xingyaoww, line 68: 🟡 Suggestion: **`resize_base64_data_url` is public but has a leaky API.** The `image_module: Any` parameter forces every caller to pre-load PIL and thread it through as an untyped blob. Either:
- Make it private (`_resize_base64_data_url`) since it's only used internally, or
- Encapsulate the PIL import inside the function and drop the `image_module` param

The current signature says "I'm a public utility" but the `image_module: Any` parameter says "don't actually use me without reading my caller's source code."

openhands-sdk/pyproject.toml
```diff
@@ -28,6 +28,7 @@ Documentation = "https://docs.openhands.dev/sdk"
 
 [project.optional-dependencies]
 boto3 = ["boto3>=1.35.0"]
+pillow = ["pillow>=12.1.1"]
```
xingyaoww: 🟡 Suggestion: **`pillow>=12.1.1` minimum is unnecessarily high.** The only Pillow API this code uses is `Image.open()`, `Image.thumbnail()`, `Image.Resampling.LANCZOS`, `Image.convert()`, and `Image.save()` — all of which have been stable since Pillow 9.1+ (April 2022, when `Image.Resampling` was introduced).

A lower bound like `>=10.0.0` would be more compatible with existing environments that already have Pillow installed at an older version, without losing any required functionality. Is there a specific reason for 12.1.1?

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    if total_images == 0:
+        return None
+    if total_images <= ANTHROPIC_MANY_IMAGE_THRESHOLD:
+        return ANTHROPIC_STANDARD_IMAGE_MAX_DIMENSION
+
+    return ANTHROPIC_MANY_IMAGE_MAX_DIMENSION
+
+
+def _load_pillow_image_module() -> Any | None:
```
xingyaoww: 🟢 Acceptable: The logic for determining `max_dimension` based on total image count is clean and correctly mirrors the Anthropic docs. The early returns for non-Anthropic providers and zero images are good taste — no unnecessary work. One small note: returning `ANTHROPIC_STANDARD_IMAGE_MAX_DIMENSION` (8000) for ≤20 images means you'll still resize any image >8000px for single-image requests, which is correct per the docs but worth calling out in a comment since it's a non-obvious behavior.

tests/sdk/llm/test_llm_image_resizing.py
```diff
@@ -0,0 +1,173 @@
...
+
```
xingyaoww, line 43: 🟢 Acceptable: Good use of patching — mocking `vision_is_active` and `_infer_litellm_provider` is the right level of isolation since you're testing the resize logic, not the provider detection. The tests create real images, resize them, and assert on actual pixel dimensions. This is the kind of test that catches real regressions. 👍

xingyaoww (commented):
🟡 **Acceptable** — Core logic is correct and the tests are solid, but there are design issues worth addressing before merge.

**Linus-Style Analysis:**

The fundamental idea is sound: intercept oversized base64 images before they hit Anthropic's API limits. The Anthropic docs research is correct (20-image threshold, 2000px vs 8000px caps), and the resize logic itself is clean.

But the plumbing around it — threading a PIL module as `Any` through three functions, silent in-place mutation — adds unnecessary complexity that a simpler design would eliminate entirely.

⚠️ **Eval-risk note**: This PR changes image processing in the LLM message pipeline, which could affect agent behavior on vision benchmarks. Flagging for a human maintainer to decide after running lightweight evals.

_This review was generated by an AI agent (OpenHands)._

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
```diff
@@ -0,0 +1,122 @@
...
+    try:
+        from PIL import Image
+    except ImportError:
+        logger.warning(
+            "pillow is not installed; skipping Anthropic image resizing. "
+            "Install openhands-sdk[pillow] to enable base64 image downscaling."
```
xingyaoww: @OpenHands please clean this up, irrc we decided to move Pillow into the main dependency of the SDK. 

The reason is that the SDK typically works together with OpenHands tools, which already has the Pillow dependency. This means it probably doesn't add any space in terms of the final application.

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
Zheng-Lu: I have removed _load_pillow_image_module(), dropped the image_module: Any parameter, and made the base64 helper private. The guarded Pillow import now lives inside the resize path, so the API surface is smaller and typed.

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
Zheng-Lu: `maybe_resize_messages_for_provider()` now returns a detached message list instead of mutating caller-owned Message objects, and `LLM.format_messages_for_llm()` assigns the returned value explicitly. I also added a regression test that calls the helper directly and verifies the input message is unchanged.

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
Zheng-Lu: `resize_base64_data_url` is private now

openhands-sdk/pyproject.toml
Zheng-Lu: I kept `pillow>=12.1.1` intentionally. The repo already has a workspace-wide security floor at that version due the Pillow CVE noted in the root pyproject.toml, so I mirrored that bound in the SDK extra rather than allowing older vulnerable installs.

openhands-sdk/openhands/sdk/llm/utils/image_resize.py
Zheng-Lu: This one is small enough to clean up myself.

xingyaoww:
@OpenHands in the existing example file that tests for vision support, can you add another send_message that sends image inputs that would properly exercise the stuff we've introduced in this PR

xingyaoww:
@OpenHands help me fix this CI https://github.com/OpenHands/software-agent-sdk/actions/runs/25000384169/job/73209940791?pr=2552

xingyaoww:
@OpenHands help me fix this CI https://github.com/OpenHands/software-agent-sdk/actions/runs/25000384169/job/73209940791?pr=2552

openhands-sdk/openhands/sdk/llm/llm.py
xingyaoww: This was already addressed — the image resize logic lives in `openhands/sdk/llm/utils/image_resize.py` and `llm.py` just calls `maybe_resize_messages_for_provider()` as a one-liner (line 1430). Resolving.

xingyaoww:
@OpenHands please merge from main and resolve conflicts

xingyaoww (approved):
Thanks! Sorry for taking so long :(

PR #2556, opened 2026-03-24
fix: Send ConversationErrorEvent when MCP connection fails and improve error handling

tofarr:
## Summary

Currently, when there is a runtime error (For example due to a misconfigured MCP server), the error is caught and the websocket is closed. instead, we now send an error event to let the client know what's going on.

The web frontend now receives a proper error event with `code: "MCPError"` and can display it to the user instead of spawning a new runtime. (Or display the provided detail).

## Example

If I set up my MCP Configuration like this:

image

The chrome developer displays:

image

## Checklist

- [X] If the PR is changing/adding functionality, are there tests to reflect this? N/A - bug fix and refactoring
- [X] If there is an example, have you run the example to make sure that it works? N/A
- [X] If there are instructions on how to run the code, have you followed the instructions and made sure that it works? N/A
- [ ] If the feature is significant enough to require documentation, is there a PR open on the OpenHands/docs repository with the same branch name? N/A
- [] Is the github CI passing? (will be checked by CI)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:ae5eb1b-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-ae5eb1b-python \
  ghcr.io/openhands/agent-server:ae5eb1b-python
```

```
ghcr.io/openhands/agent-server:ae5eb1b-golang-amd64
ghcr.io/openhands/agent-server:ae5eb1b-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:ae5eb1b-golang-arm64
ghcr.io/openhands/agent-server:ae5eb1b-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:ae5eb1b-java-amd64
ghcr.io/openhands/agent-server:ae5eb1b-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:ae5eb1b-java-arm64
ghcr.io/openhands/agent-server:ae5eb1b-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:ae5eb1b-python-amd64
ghcr.io/openhands/agent-server:ae5eb1b-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-amd64
ghcr.io/openhands/agent-server:ae5eb1b-python-arm64
ghcr.io/openhands/agent-server:ae5eb1b-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-arm64
ghcr.io/openhands/agent-server:ae5eb1b-golang
ghcr.io/openhands/agent-server:ae5eb1b-java
ghcr.io/openhands/agent-server:ae5eb1b-python
```

- Each variant tag (e.g., `ae5eb1b-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `ae5eb1b-python-amd64`) are also available if needed

### DIFF
--- a/.github/workflows/run-eval.yml
+++ b/.github/workflows/run-eval.yml
@@ -354,6 +354,7 @@ jobs:
                   echo "Dispatching evaluation workflow with SDK commit: $SDK_SHA (benchmark: $BENCHMARK, eval branch: $EVAL_BRANCH, benchmarks branch: $BENCHMARKS_BRANCH, tool preset: $TOOL_PRESET)"
                   PAYLOAD=$(jq -n \
                     --arg sdk "$SDK_SHA" \
+                    --arg sdk_run_id "${{ github.run_id }}" \
                     --arg eval_limit "$EVAL_LIMIT" \
                     --argjson models "$MODELS_JSON" \
                     --arg ref "$EVAL_BRANCH" \
@@ -370,7 +371,7 @@ jobs:
                     --arg agent_type "$AGENT_TYPE" \
                     --arg partial_archive_url "$PARTIAL_ARCHIVE_URL" \
                     --arg triggered_by "$TRIGGERED_BY" \
-                    '{ref: $ref, inputs: {sdk_commit: $sdk, eval_limit: $eval_limit, models_json: ($models | tostring), trigger_reason: $reason, pr_number: $pr, benchmarks_branch: $benchmarks, benchmark: $benchmark, instance_ids: $instance_ids, num_infer_workers: $num_infer_workers, num_eval_workers: $num_eval_workers, enable_conversation_event_logging: $enable_conversation_event_logging, max_retries: $max_retries, tool_preset: $tool_preset, agent_type: $agent_type, partial_archive_url: $partial_archive_url, triggered_by: $triggered_by}}')
+                    '{ref: $ref, inputs: {sdk_commit: $sdk, sdk_workflow_run_id: $sdk_run_id, eval_limit: $eval_limit, models_json: ($models | tostring), trigger_reason: $reason, pr_number: $pr, benchmarks_branch: $benchmarks, benchmark: $benchmark, instance_ids: $instance_ids, num_infer_workers: $num_infer_workers, num_eval_workers: $num_eval_workers, enable_conversation_event_logging: $enable_conversation_event_logging, max_retries: $max_retries, tool_preset: $tool_preset, agent_type: $agent_type, partial_archive_url: $partial_archive_url, triggered_by: $triggered_by}}')
                   RESPONSE=$(curl -sS -o /tmp/dispatch.out -w "%{http_code}" -X POST \
                     -H "Authorization: token $PAT_TOKEN" \
                     -H "Accept: application/vnd.github+json" \
--- a/README.md
+++ b/README.md
@@ -122,3 +122,23 @@ For development setup, testing, and contribution guidelines, see [DEVELOPMENT.md
       url={https://arxiv.org/abs/2511.03690}, 
 }
 ```
+
+<hr>
+
+<div align="center">
+<strong>Trusted by engineers at</strong>
+<br/><br/>
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137f6974b71a1a4a932f82_TikTok_logo.svg" alt="TikTok" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137f523b08f91a5aa905b9_Vmware.svg" alt="VMware" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137f2cb537758796a9dba1_Roche_Logo.svg" alt="Roche" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137f10c3975e28b3932320_Amazon_logo%201.svg" alt="Amazon" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137ec5a6f77dd174e557ce_C3ai_logo%201.svg" alt="C3 AI" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137eac8f27ca27f5e48420_Netflix_2015_logo%201.svg" alt="Netflix" height="40">
+<br/>
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137e8df2c028b9e1506ede_mastercard%201.svg" alt="Mastercard" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137e783790933dd06f9d59_Red_Hat_Logo_2019%201.svg" alt="Red Hat" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137e5fa006d963a1d1904d_mongodb-ar21%201.svg" alt="MongoDB" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137e47b45195da10c50f49_apple-11%201.svg" alt="Apple" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137e34e3a5ab71e37082a7_NVIDIA_logo%201.svg" alt="NVIDIA" height="40">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+<img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137e199ce2cb594b0210ab_google-ar21%201.svg" alt="Google" height="40">
+</div>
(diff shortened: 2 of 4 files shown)


openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -207,6 +209,18 @@ async def events_socket(
...
+                        detail=str(e),
+                    )
+                    try:
+                        await _send_event(error_event, websocket)
+                    except Exception:
+                        pass  # Don't fail even if we can't send the error event
+
                 # For critical errors that indicate the websocket is broken, exit
                 if isinstance(e, (RuntimeError, ConnectionError)):
                     raise
```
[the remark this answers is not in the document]
tofarr: We actually need to keep the websocket open here so that the frontend does not simply get a "disconnected" message - I've removed the comment about closing the websocket.

openhands-sdk/openhands/sdk/mcp/utils.py
```diff
@@ -78,6 +78,11 @@ def create_mcp_tools(
         raise MCPTimeoutError(
             error_msg, timeout=timeout, config=config.model_dump()
         ) from e
+    except RuntimeError as e:
+        # Wrap RuntimeError from MCP client as MCPError so it can be handled
+        # gracefully in sockets.py instead of treating it as a critical error
+        client.sync_close()
+        raise MCPError(f"MCP connection failed: {e}") from e
```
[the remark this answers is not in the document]
tofarr: This is something that happens lazily when the conversation is first prompted. Wrapping a runtime error raised by fastmcp in an MCPError seems appropriate - the stack trace will maintain the full reason if required, and the user is directed to look at their MCP settings for the likely cause.

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -201,16 +203,26 @@ async def events_socket(
...
                 message = Message.model_validate(data)
                 await event_service.send_message(message, True)
-            except WebSocketDisconnect:
-                logger.info(f"Event websocket disconnected: {conversation_id}")
+            except (WebSocketDisconnect, ConnectionError):
                 # Exit the loop when websocket disconnects
+                logger.info(f"Event websocket disconnected: {conversation_id}")
                 return
-            except Exception as e:
+            except (MCPError, RuntimeError) as e:
```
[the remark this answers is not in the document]
tofarr: I've updated the documentation - it is appropriate that we catch both MCPErrors and general runtime errors here. General runtime errors are likely to be less frequent than the known case of MCP misconfiguration or an outage of an external MCP service.

The important thing is that we no longer close the connection, and that we send an error message so the user can decide what they want to do next.

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -201,16 +203,26 @@ async def events_socket(
...
+                try:
+                    error_event = ConversationErrorEvent(
+                        source="environment",
+                        code=e.__class__.__name__,
+                        detail=str(e),
+                    )
+                    await _send_event(error_event, websocket)
+                except Exception:
+                    # MCP Failed but there was an error sending to the client
+                    logger.exception("send_error_failure", stack_info=True)
```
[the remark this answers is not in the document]
tofarr: The only case which exits the loop is now the websocket disconnect - and a comment describes this - all other flow paths stay in the loop - I think this is clear enough.

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -201,16 +203,26 @@ async def events_socket(
...
                 await event_service.send_message(message, True)
-            except WebSocketDisconnect:
-                logger.info(f"Event websocket disconnected: {conversation_id}")
+            except (WebSocketDisconnect, ConnectionError):
                 # Exit the loop when websocket disconnects
+                logger.info(f"Event websocket disconnected: {conversation_id}")
                 return
-            except Exception as e:
+            except (MCPError, RuntimeError) as e:
+                # Runtime errors send an error and close the connection
```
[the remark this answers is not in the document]
tofarr: Comment was updated.

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -201,16 +203,28 @@ async def events_socket(
...
+                logger.info(f"Event websocket disconnected: {conversation_id}")
                 return
-            except Exception as e:
+            except (MCPError, RuntimeError) as e:
+                # Something went wrong in setup and the conversation may be
+                # broken (Typical culprit is an MCP misconfiguration)
+                # Tell the client about this so they can decide what to do next
+                logger.exception("error_in_subscription", stack_info=True)
+                try:
+                    error_event = ConversationErrorEvent(
```
xingyaoww: @OpenHands do not change any code - help me understand how ConversationErrorEvent is used across the codebase, then tell me if this usage is correct

xingyaoww: Here's a full analysis of how `ConversationErrorEvent` is used across the codebase and whether the usage in this PR is correct.

## `ConversationErrorEvent` Definition

**File:** `openhands-sdk/openhands/sdk/event/conversation_error.py`

Extends `Event` (which has `source: Literal["agent", "user", "environment", "hook"]`). Adds two fields: `code: str` and `detail: str`.

From its docstring:
> Conversation-level failure that is NOT sent back to the LLM. Emitted by the conversation runtime when an unexpected exception bubbles up and prevents the run loop from continuing. Intended for client applications (e.g., UIs) to present a top-level error state.

Key distinction from `AgentErrorEvent`: not tied to any `tool_name`/`tool_call_id`, typically `source="environment"`, and the run loop moves to an ERROR state.

## Existing Usages (3 places)

### 1. `local_conversation.py` — emitter (`source="environment"`)

Two emission sites in the `run()` method:
- **Max iterations reached** (line 683): `code="MaxIterationsReached"`, `detail=<human-readable message>`. Sets `execution_status = ERROR`.
- **Unhandled exception in run loop** (line 695): `code=e.__class__.__name__`, `detail=str(e)`. Sets `execution_status = ERROR`, then re-raises as `ConversationRunError`.

**Pattern:** Always emitted **inside the SDK conversation run loop**, always paired with setting `execution_status = ERROR`.

### 2. `acp_agent.py` — emitter (`source="agent"`)

Emitted for ACP prompt errors (line 938): `source="agent"`, `code="UsagePolicyRefusal"` or `"ACPPromptError"`, `detail=error_str[:500]`. Also sets `execution_status = ERROR` and re-raises.

### 3. `remote_conversation.py` — consumer

`_get_last_error_detail()` (line 1147) scans the event list in reverse for the most recent `ConversationErrorEvent` and extracts its `code` + `detail`. This is used at line 1106 when the remote conversation status is ERROR, to provide a more descriptive error message in `ConversationRunError` instead of the generic "Remote conversation ended with error".

## Usage in This PR (`sockets.py` line 216)

```python
error_event = ConversationErrorEvent(
    source="environment",
    code=e.__class__.__name__,
    detail=str(e),
)
await _send_event(error_event, websocket)
```

## Correctness Assessment

### ✅ Field values are correct
- `source="environment"` — matches the docstring guidance and `local_conversation.py` pattern. The error originates from infrastructure (MCP misconfiguration), not from the agent or user.
- `code=e.__class__.__name__` — consistent with `local_conversation.py`'s exception handler pattern (line 698). For `MCPError` this yields `"MCPError"`, for `RuntimeError` it yields `"RuntimeError"`.
- `detail=str(e)` — consistent with `local_conversation.py` line 699.

### ⚠️ Architectural difference from existing pattern

All existing usages emit `ConversationErrorEvent` **inside the SDK's conversation run loop** and always pair it with setting `execution_status = ERROR`. The event flows through the normal event pipeline (event store → subscribers → websocket).

This PR emits it **at the agent-server websocket transport layer** — the event is created ad-hoc and sent directly over the websocket, bypassing the event service. This means:

1. **The event is NOT persisted** in the event store. If someone queries events via REST API, they won't find it.
2. **No `execution_status` is set to ERROR** at this layer. The SDK conversation layer may not know the error occurred.
3. The `RemoteConversation._get_last_error_detail()` consumer would find it in its in-memory event list (since it received it via websocket), but only for that session.

### ⚠️ Loop continuation differs from established pattern

In all existing usages, `ConversationErrorEvent` signals a **terminal state** — the run loop stops. In this PR, after sending the error event, the code falls through back to `while True` and continues listening on the websocket. This means the websocket stays open and the client can continue sending messages.

This could be intentional (allows the error to be delivered and the client can decide what to do), but it's different from the established semantic that `ConversationErrorEvent` = conversation is broken/stopped.

### ⚠️ `RuntimeError` is very broad

Catching `RuntimeError` alongside `MCPError` is quite broad. In the original code, `RuntimeError` was in the "exit the loop" (connection broken) category. Now it's in the "send error event and continue" category. If a `RuntimeError` indicates a fundamentally broken state (e.g., the event loop or async runtime is corrupted), continuing the loop could be problematic.

## Summary

The **field-level construction** of `ConversationErrorEvent` is correct and follows established patterns. However, the **architectural placement** raises questions:
1. It's emitted at the transport layer rather than the SDK layer (not persisted, no status change)
2. The loop continues after emission (existing pattern treats it as terminal)
3. `RuntimeError` catch may be too broad for this handler

The pragmatic goal (tell the client about MCP errors instead of silently disconnecting) is sound, but you may want to consider whether this event should instead be emitted deeper in the stack (e.g., where the MCP connection error actually occurs during conversation setup) so it flows through the normal event pipeline and gets persisted.

xingyaoww: @tofarr see comment above, shall we handle it in the sdk level, rather than the agent server level?

openhands-agent-server/openhands/agent_server/models.py
```diff
@@ -31,6 +32,19 @@
 from openhands.sdk.workspace.base import BaseWorkspace
 
 
+class AgentServerErrorEvent(Event):
```
enyst: Nit: 
```suggestion
class ServerErrorEvent(Event):
```
We’ve seen way too many confusions about the name `AgentErrorEvent`, and this sounds too close. I believe it’s important to make and keep a clear distinction between agent-history “errors”, which are just LLM making a blunder, and application-level errors.

Maybe it could also inherit `ConversationErrorEvent`?

openhands-agent-server/openhands/agent_server/models.py
```diff
@@ -31,6 +32,19 @@
...
+    """Event emitted by the agent server when a server-level error occurs.
+
+    This event is used for errors that originate from the agent server itself,
+    such as MCP connection failures, WebSocket errors, or other infrastructure
+    issues. Unlike ConversationErrorEvent which is for conversation-level failures,
```
enyst, line 40: I wonder, could we use a field for the code, to help make this distinction?

openhands-agent-server/openhands/agent_server/models.py
xingyaoww: ^^ good suggestion

xingyaoww: > Maybe it could also inherit ConversationErrorEvent?

Though i'm not 100% sure about this - i think we always persist ConversationErrorEvent to disk, but not necessary for `ConversationErrorEvent`?

enyst: That… makes sense, they’re errors of a particular entry point infra… 

Idk, they have to be Event for transport? 🤔 Event but not in EventLog?

PR #2564, opened 2026-03-25
feat(sdk): implement LargeFileSurgicalCondenser for optimized context managementFeature/surgical condenser

vivekvjnk:
## Summary

This PR introduces the LargeFileSurgicalCondenser, a specialized context management tool designed to mitigate context window bloat caused by large tool outputs (specifically images and large file reads from the file_editor tool).

Unlike standard windowing or summarization condensers, this "surgical" condenser follows a Post-Inference Cleanup strategy. It preserves raw, high-fidelity data (like base64 images) during the turn it is first viewed to ensure the agent has the necessary information for analysis. Crucially, it only replaces that data with a concise summary after the agent has responded, ensuring subsequent turns benefit from a lean context and increased KV cache efficiency without sacrificing the quality of the initial analysis.

## Key Features
- **Targeted Condensation**: Specifically monitors observations from the` file_editor` tool (configurable).
- **Post-Inference Cleanup**: Only triggers condensation when an observation is followed by an agent action/response, ensuring the data has already been "processed."
- **Smart Summaries**: Attempts to correlate observations with their preceding actions to include file paths in the condensation summary (e.g., `[Condensation] Viewed sample_image.png`).
- **Vision Optimization**: Automatically triggers for any ImageContent, which is a primary driver of context bloat in multi-modal workflows.
- **Configurable Thresholds**: Includes a default byte-size threshold (10KB) for `TextContent`.

## Changes
- Core: Added LargeFileSurgicalCondenser in `openhands/sdk/context/condenser/`.
- **Tests**: Comprehensive unit tests covering tool-name filtering, threshold logic, image detection, and sequence-based condensation triggers.
- **Example**: Added a new vision-agent example script in `examples/06_custom_examples/vision_agent/` demonstrating the condenser in a multi-turn conversation.

## Performance Impact
By surgically removing prefix-heavy data like base64 images after their first use, we significantly increase prefix-matching for KV caches in multi-turn interactions, reducing latency and token costs for long-running sessions.

## Checklist


uv.lock
VascoSch92, line 1: can you reverse the uv.lock?

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
```diff
@@ -0,0 +1,133 @@
...
+
+                        if filename:
+                            summary = f"[Condensation]Viewed {filename}"
+                        else:
+                            file_info = (
+                                "image/data" if has_image else f"data ({size_kb:.2f}KB)"
+                            )
+                            summary = f"[Condensation]Viewed {file_info}"
+
+                        return Condensation(
```
csmith49: Unfortunately I don't think this works the way you might expect. If we see the pattern:

```
<prefix>
Action event: retrieve large file
Observation event: the large file
<suffix>
```

You probably want the resulting sequence to look like:

```
<prefix>
Action event: retrieve large file
Condensation: summary of large file
<suffix>
```

But LLM APIs expect every action event to have a matching observation event and will throw an exception if that isn't the case. We prevent these exceptions by filtering out unmatched actions and observations when constructing the `View`, so the actual resulting sequence is probably something like:

```
<prefix>
Condensation: summary of large file
<suffix>
```

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
```diff
@@ -0,0 +1,133 @@
...
+        for index, event in enumerate(view.events):
+            if isinstance(event, ObservationEvent):
+                # Only operate on events that are followed by another event
+                # (which implies the agent has seen and responded to it)
+                if index < len(view.events) - 1:
+                    # Look ahead to see if the next event is an action from the agent
+                    next_event = view.events[index + 1]
+                    if next_event.source == "agent":
+                        if self._should_condense_event(event):
+                            return CondensationRequirement.HARD
```
csmith49: A hard condensation requirement indicates that the agent cannot proceed without a condensation, and that if we can't apply one we need to take drastic measures (like a hard context reset). Not quite the case here: if we delay condensation the only harm is that we spend a little extra time/money committing to the wrong cache prefix.

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
```diff
@@ -0,0 +1,133 @@
...
+    NoCondensationAvailableException,
+    RollingCondenser,
+)
+from openhands.sdk.context.view import View
+from openhands.sdk.event.condenser import Condensation
+from openhands.sdk.event.llm_convertible import ActionEvent, ObservationEvent
+from openhands.sdk.llm import LLM, ImageContent, TextContent
+
+
+class LargeFileSurgicalCondenser(RollingCondenser):
```
csmith49: You probably want the `CondenserBase` class instead, the `RollingCondenser` interface is set up to make condensers with moving "attention windows" a little easier to mess with.

csmith49 (commented):
I like the idea! Reminds me of the condensers we had in v0 (like [this one](https://github.com/OpenHands/OpenHands/blob/main/openhands/memory/condenser/impl/browser_output_condenser.py) to limit the data produced by the old browser tool). There are two problems/concerns I'd like to see addressed before approval:

## Condenser Robustness

As implemented, I believe this approach is modifying more of the event history than is expected. I left some comments on the changes highlighting my concerns.

My recommendation is to tweak the `LargeFileSurgicalCondenser` to use the `CondenserBase` interface, which will simplify things:

```python
class LargeFileSurgicalCondenser(CondenserBase):
    def condense(
        self,
        view: View,
        agent_llm: LLM | None = None
    ) -> View | Condensation:
        # The view is a list of events we want to show the LLM
        # Instead of trying to replace existing events with a
        # Condensation, just do the surgery on the events
        # directly while constructing a new view
```

No need for `condensation_requirement` or `get_condensation`, you can just always return a modified `View` to get the behavior you want. Doing surgery directly on the events in the view will help prevent the action/observation mismatch I noted.

## Looping

When we had a similar masking behavior for browser outputs, the agents would often get stuck in a loop. They'd find a page they want, follow a few links to build context, and by the time they were ready to act the first page had been masked. But attempting to reload that context would mask the second page, and reloading that would mask the third, and by the time everything was reloaded the first would be masked again.

We tried a lot of things to fix the looping, but nothing really did the trick (which is part of the reason why the browser output masking condenser wasn't ported from v0 to v1).

This implementation only keeps large files around for a single agent step, and that's probably fine for a single image. Have you noticed any looping with multiple images? Or large text files linked together?

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
```diff
@@ -0,0 +1,133 @@
...
+    """A highly specific condenser that removes bloated data from observations.
+
+    This condenser intercepts observations (like ImageContent or large TextContent from
+    the 'file_editor' tool) and replaces them with a small summary string.
+    This preserves KV cache efficiency by keeping the prefix stable and forcing early
+    eviction of large branches.
+    :param target_tool: Only condense observations from this tool
+        (default: 'file_editor').
+    :param threshold_bytes: For TextContent, the byte size threshold to
+        trigger condensation (default: 10KB).
```
csmith49, line 26: Minor nit: this repo uses Google style doc-strings, not reST

vivekvjnk:
Hi All,
I'm working on resolving all the comments. Got stuck with some other work. 
I will try to resolve all issues by tomorrow.

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
vivekvjnk: Actually I was trying to implement something like this:
```
<prefix>
Action event: retrieve large file
Observation event: [Condensation] summary of large file
<suffix>
```

Here actual observation event is replaced with condensed observation with same event id. 
Rationale here is to make minimal changes to the original event list while generating new view.
So it felt unnecessary to condense the preceding action event.
(I've updated the code to return observation event instead of condensation event)

uv.lock
vivekvjnk: Done

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
vivekvjnk: Agree. Updated in new commit

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
vivekvjnk: Agree. Updated in new commit

vivekvjnk: Done

vivekvjnk:
please review

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
```diff
@@ -0,0 +1,118 @@
...
+            return f"[Condensation]Viewed {file_info}"
+
+    def condense(self, view: View, agent_llm: LLM | None = None) -> View | Condensation:
+        """Surgically condenses bloated observations in the view."""
+        new_events = list(view.events)
+        modified = False
+
+        for index, event in enumerate(view.events):
+            if isinstance(event, ObservationEvent):
+                # Ensure it's not the last event, so the agent has seen it.
```
csmith49, line 87: I'm still concerned about only showing the agent the event for one step. Current SotA LLMs always seem to like doing long interrupted sequences of tool calls, and are probably not prepared for this pattern.

A more natural approach might be to wait until the agent has returned back to the user, but that won't work well in headless mode and will also cause a lot of cache invalidation. Worth testing?

And if this pattern must stick around, I'd suggest making it parameterized by some instance variable so that folks could tune it when instantiating the condenser.

openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py
```diff
@@ -0,0 +1,118 @@
...
+                    if next_event.source == "agent":
+                        if self._should_condense_event(event):
+                            summary = self._get_summary(view, event, index)
+                            # Create a new Observation with the summary.
+                            # Use the same type as the original observation to avoid ABC issues.
+                            new_observation = type(event.observation)(
+                                content=[TextContent(text=summary)]
+                            )
+                            # Create a new ObservationEvent with the same IDs to preserve pairing.
+                            new_event = ObservationEvent(
```
csmith49, line 100: If we're already injecting a new observation of the original type, why make a whole new event? Could probably just inject the modified observation (into a clone) and keep the event structure identical otherwise. Less risk of missing a copied argument.

smolpaws (commented):

Taste rating: acceptable

Findings:
- Changed area reviewed: examples/06_custom_examples/vision_agent/sample.png, examples/06_custom_examples/vision_agent/surgical_condenser.py, openhands-sdk/openhands/sdk/__init__.py and 4 more file(s).
- I did not see registry redirection, lifecycle-hook abuse, secret exposure or unsafe workflow permission widening in the diff. Treat this as higher risk than ordinary code because dependency and workflow changes affect the build and CI supply chain.

Risk assessment: medium.

Verdict: worth merging if the relevant CI checks pass

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py, Line 96] **Breaks on the actual target observation**: `type(event.observation)(content=...)` only works for the test's fieldless `DummyObservation`. `FileEditorObservation` requires `command`, so the first real large `file_editor` result raises a Pydantic validation error instead of condensing. Preserve the existing observation with `model_copy(update={"content": [...]})` (or otherwise retain every subtype field), and test with a real `FileEditorObservation`.
- [uv.lock, Line 1788] **Local dependency override committed**: LiteLLM is changed from a pinned PyPI artifact to `source = { editable = "../../litellm" }` and the SDK requirement points at the same path. A clean checkout does not contain that sibling directory, so this destroys lockfile reproducibility and silently substitutes unrelated local source. Regenerate `uv.lock` from the repository's declared dependencies and remove this override/version churn from the feature PR.
- [openhands-sdk/openhands/sdk/context/condenser/large_file_surgical_condenser.py, Lines 85-92] **Parallel observations are skipped**: condensation only occurs when the *immediately next* event is agent-authored. With parallel tool calls, earlier large observations are followed by other environment observations before the agent response, so they remain bloated forever. Determine whether the observation precedes the next agent turn rather than requiring adjacency, and add a multi-observation regression test.

**[TESTING GAPS]**
- [tests/sdk/context/condenser/test_large_file_surgical_condenser.py, Line 21] The dummy observation erases the required-field behavior of production tool observations, which is why the primary crash passes CI. Exercise the real `FileEditorObservation` and a parallel-tool event sequence.

**VERDICT:**
❌ **Needs rework**: The condenser crashes on its advertised production input, misses parallel outputs, and carries an invalid local lockfile source.

**KEY INSIGHT:**
The abstraction must preserve observation data structures; rebuilding arbitrary subtypes from only `content` is fundamentally unsafe.

**[RISK ASSESSMENT]**
This changes core context construction, exports a new public SDK type, and alters dependency provenance through a local editable override. The current implementation can fail during normal `file_editor` histories. **Recommendation**: Do not auto-merge. Request review from a human SDK/context maintainer after fixing subtype preservation and the lockfile.

---

>
>
>

PR #2613, opened 2026-03-30
fix(sdk): recover malformed tool history via condensation

enyst:
## Summary

This PR captures Anthropic malformed message-history errors involving broken `tool_use` / `tool_result` pairing with a **dedicated** SDK exception instead of treating them as literal context-window overflows.

Examples include:

- `` `tool_use` ids were found without `tool_result` blocks immediately after ``
- `` each `tool_use` must have a single result ``
- `` unexpected `tool_use_id` found in `tool_result` blocks ``

The intent is still the same recovery behavior discussed for these failures: if the current prompt history is unusable, we should try again with condensed history. But this PR now keeps that distinct from true context-window exhaustion in both types and logs.

Concretely, this change:

1. adds `LLMMalformedConversationHistoryError`
2. classifies the malformed-history provider errors into that dedicated exception
3. catches that exception separately in `Agent.step()`
4. routes it into the existing `CondensationRequest` retry path with explicit malformed-history logging

That means:

- this **does not** claim the error is a context-window overflow anymore
- recovery still retries with condensed history when a suitable condenser is configured
- regular condensation may happen first
- hard reset remains available only if the configured condenser escalates to it
- if no suitable condenser is configured, the malformed-history exception still surfaces, now with more explicit logging that this likely points to an upstream event-stream / resume bug

This keeps the immediate recovery fix, while preserving observability so we can tell the difference between:

- genuine prompt-too-large failures, and
- malformed persisted / resumed history that "should never happen" but does in practice

## Checklist

Validation run locally:

- `uv run pre-commit run --files AGENTS.md openhands-sdk/openhands/sdk/llm/exceptions/types.py openhands-sdk/openhands/sdk/llm/exceptions/__init__.py openhands-sdk/openhands/sdk/llm/exceptions/classifier.py openhands-sdk/openhands/sdk/llm/exceptions/mapping.py openhands-sdk/openhands/sdk/agent/agent.py tests/sdk/llm/test_exception.py tests/sdk/llm/test_exception_classifier.py tests/sdk/llm/test_exception_mapping.py tests/sdk/agent/test_agent_context_window_condensation.py`
- `uv run pytest tests/sdk/llm/test_exception.py tests/sdk/llm/test_exception_classifier.py tests/sdk/llm/test_exception_mapping.py tests/sdk/agent/test_agent_context_window_condensation.py`

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:600cd3c-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-600cd3c-python \
  ghcr.io/openhands/agent-server:600cd3c-python
```

```
ghcr.io/openhands/agent-server:600cd3c-golang-amd64
ghcr.io/openhands/agent-server:600cd3c-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:600cd3c-golang-arm64
ghcr.io/openhands/agent-server:600cd3c-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:600cd3c-java-amd64
ghcr.io/openhands/agent-server:600cd3c-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:600cd3c-java-arm64
ghcr.io/openhands/agent-server:600cd3c-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:600cd3c-python-amd64
ghcr.io/openhands/agent-server:600cd3c-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:600cd3c-python-arm64
ghcr.io/openhands/agent-server:600cd3c-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:600cd3c-golang
ghcr.io/openhands/agent-server:600cd3c-java
ghcr.io/openhands/agent-server:600cd3c-python
```

- Each variant tag (e.g., `600cd3c-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `600cd3c-python-amd64`) are also available if needed


enyst:
Hi, I'm OpenHands-GPT-5.4.

I think both points are right, and they point to a good split between a **safe immediate fix** and a **better follow-up design**.

### My take

**Calvin’s concern is very valid.**  
If the real problem is a race / resume bug / bad event persistence, then recovering by condensation can absolutely **mask the symptom** and make root-cause debugging harder. Especially if we silently “heal” the conversation and throw away the malformed region.

**Engel’s idea also makes sense.**  
For this specific Anthropic error, the conversation history is structurally invalid, not just too large. In that situation, a **hard reset is actually a cleaner semantic recovery** than “try normal condensation and hope the bad structure disappears.”

### Why hard condensation is appealing here

For errors like:

- missing `tool_result` after `tool_use`
- duplicate `tool_result`
- orphan `tool_result`

the problem is basically: **the current visible history is unusable**.

A hard reset gives you a known-good shape:

- system prompt
- maybe initial user intent / preserved essentials
- summary
- continue from there

That’s much more deterministic than regular condensation, which may still preserve some malformed pairing structure depending on where the condensation lands.

### Why I wouldn’t jump straight to “always do hard reset” without care

A few reasons:

1. **It can hide root causes**  
   Calvin’s main point.

2. **It’s a stronger loss of context**
   Hard reset throws away a lot more than necessary.

3. **It needs loop protection**
   If something upstream keeps producing malformed history, “recover → continue → malformed again → recover” can become a bad retry cycle.

4. **Not every configuration can do it**
   Only condensers that support hard reset can actually honor that policy.

### So overall: I think the right direction is

- **Current PR:** good as a minimal fix  
  It gets these errors into the existing recovery pipeline without overcommitting to a new policy.

- **Follow-up:** probably yes, special-case these malformed-history errors to **prefer hard reset**  
  But only with:
  - explicit telemetry / logging
  - one-shot retry protection
  - maybe a marker that this was a malformed-history recovery, not a true context-window event

### Concretely, I’d frame the follow-up like this

For this error family, we should treat it as:

> “history is invalid; prefer rebuilding a minimal valid prompt”

not:

> “history is too long; try generic condensation”

That suggests a dedicated path like:

- detect malformed-history LLM error
- emit strong diagnostics
- if condenser supports hard reset, **go straight to hard reset**
- otherwise fall back to current condensation-request behavior
- never retry this recovery endlessly in one run

### My bottom line

- **As product behavior:** Engel’s proposal is probably better
- **As a first patch:** the current PR is still the right move
- **As engineering hygiene:** Calvin is right that we should not let the recovery path erase evidence without adding observability and loop guards

If you want, I can help draft a concise Slack reply that says exactly that without sounding wishy-washy.

enyst:
@all-hands-bot please review the latest changes. This now uses a dedicated `LLMMalformedConversationHistoryError` and a separate `Agent.step()` catch/logging path instead of treating these as context-window overflows.

enyst:
@openhands Merge main into this branch and fix the conflicts.

PR #2669, opened 2026-04-02
feat(tools/terminal): tmux pane pool for parallel terminal execution 

VascoSch92:
## Summary
- Introduces TmuxPanePool, a thread-safe pool of tmux panes that enables concurrent terminal command execution. Each call checks out a dedicated pane, runs the command, and checks it back inbounded by a semaphore to limit concurrency to max_panes (default 4). 
- Refactors TerminalExecutor to use the pool when tmux is available, falling back to the existing single-session subprocess path otherwise. The pool and single-session executionpaths are fully separated (_execute_pooled vs _execute_single_session), eliminating double-checkout bugs and pane leaks from the original implementation.
- Declares fine-grained resources in TerminalTool.declared_resources() so the framework parallelizes correctly: tmux pool mode opts out of framework locking (the pool handles its own concurrency), while subprocess mode serializes terminal calls via a terminal:session resource key without blocking unrelated tools.                                            
 
## Key design decisions                                                                                                                                                               
                  
- Pooled terminals override close() to only kill their window instead of the shared tmux session. This prevents TerminalSessionBase.__del__ (triggered by GC of cached TerminalSession wrappers) from destroying the session that all other panes depend on.
- pool.replace(old) creates the new pane before killing the old window, ensuring the tmux session always has at least one window alive.                                            
- _discard_session marks the session and terminal as closed before dropping the reference, so immediate GC via __del__ is a no-op.                                                 
- _sessions dict is guarded by a dedicated lock to prevent races during concurrent _wrap_session calls.                                                                            
                                                                                                                                                                                     
## Test plan                                                                                                                                                                          
                                                                                                                                                                                     
- test_tmux_pane_pool.py — 25 tests covering init validation, lazy pane creation, checkout/checkin reuse, blocking when full, context manager (normal + exception exit), replace (preserves semaphore, closes old pane, doesn't affect siblings), parallel execution (2 and 3 threads), concurrent replace safety, initial window cleanup, close lifecycle, introspection                                                                                                                                                                      
- test_terminal_reset.py — 7 tests verifying reset clears env vars, executes commands after reset, preserves working directory, works multiple times in sequence
- test_terminal_tool_auto_detection.py — updated to handle pool mode (no executor.session attribute when tmux is active), patches _is_tmux_available in both factory and impl      
- test_secrets_masking.py — forces subprocess mode for mock-based session testing                           


## Benchmark

```
 Benchmarking 3 chains          
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┓
┃ Chain                       ┃  Calls ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━┩
│ stress-tmux-heavy-cat       │     20 │
│ stress-tmux-grep-flood      │     20 │
│ stress-tmux-seq-flood       │     20 │
└─────────────────────────────┴────────┘
```

main

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┓
┃ Chain                     ┃  W ┃   Wall (s) ┃  Speedup ┃  Mem (MB) ┃  Errors ┃  Unique ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━┩
│ stress-tmux-heavy-cat     │  1 │     14.208 │    1.00x │       1.1 │       0 │     1/3 │
│ stress-tmux-heavy-cat     │  2 │     14.036 │    1.01x │       1.1 │       0 │     1/3 │
│ stress-tmux-heavy-cat     │  4 │     14.357 │    0.99x │       1.1 │       0 │     1/3 │
│ stress-tmux-heavy-cat     │  8 │     14.525 │    0.98x │       1.1 │       0 │     1/3 │
├───────────────────────────┼────┼────────────┼──────────┼───────────┼─────────┼─────────┤
│ stress-tmux-grep-flood    │  1 │     14.090 │    1.00x │       0.5 │       0 │     1/3 │
│ stress-tmux-grep-flood    │  2 │     14.301 │    0.99x │       0.5 │       0 │     1/3 │
│ stress-tmux-grep-flood    │  4 │     14.393 │    0.98x │       0.5 │       0 │     1/3 │
│ stress-tmux-grep-flood    │  8 │     14.284 │    0.99x │       0.6 │       0 │     1/3 │
├───────────────────────────┼────┼────────────┼──────────┼───────────┼─────────┼─────────┤
│ stress-tmux-seq-flood     │  1 │     14.723 │    1.00x │       1.0 │       0 │     1/3 │
│ stress-tmux-seq-flood     │  2 │     14.737 │    1.00x │       1.0 │       0 │     1/3 │
│ stress-tmux-seq-flood     │  4 │     14.678 │    1.00x │       1.0 │       0 │     1/3 │
│ stress-tmux-seq-flood     │  8 │     14.760 │    1.00x │       1.1 │       0 │     1/3 │
└───────────────────────────┴────┴────────────┴──────────┴───────────┴─────────┴─────────┘
```

this branch

```
  Benchmark Results                                     
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┓
┃ Chain                     ┃  W ┃   Wall (s) ┃  Speedup ┃  Mem (MB) ┃  Errors ┃  Unique ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━┩
│ stress-tmux-heavy-cat     │  1 │      9.682 │    1.00x │       1.4 │       0 │     1/3 │
│ stress-tmux-heavy-cat     │  2 │      5.816 │    1.67x │       1.7 │       0 │     1/3 │
│ stress-tmux-heavy-cat     │  4 │      3.807 │    2.55x │       2.0 │       0 │     1/3 │
│ stress-tmux-heavy-cat     │  8 │      3.788 │    2.56x │       2.1 │       0 │     1/3 │
├───────────────────────────┼────┼────────────┼──────────┼───────────┼─────────┼─────────┤
│ stress-tmux-grep-flood    │  1 │     14.602 │    1.00x │       1.1 │       0 │     1/3 │
│ stress-tmux-grep-flood    │  2 │      7.553 │    1.93x │       1.2 │       0 │     1/3 │
│ stress-tmux-grep-flood    │  4 │      4.264 │    3.43x │       1.4 │       0 │     1/3 │
│ stress-tmux-grep-flood    │  8 │      4.450 │    3.28x │       1.5 │       0 │     1/3 │
├───────────────────────────┼────┼────────────┼──────────┼───────────┼─────────┼─────────┤
│ stress-tmux-seq-flood     │  1 │     17.449 │    1.00x │       1.5 │       0 │     1/3 │
│ stress-tmux-seq-flood     │  2 │      9.206 │    1.90x │       1.8 │       0 │     1/3 │
│ stress-tmux-seq-flood     │  4 │      4.946 │    3.53x │       1.8 │       0 │     1/3 │
│ stress-tmux-seq-flood     │  8 │      4.710 │    3.72x │       2.0 │       0 │     1/3 │
└───────────────────────────┴────┴────────────┴──────────┴───────────┴─────────┴─────────┘
```
## Checklist

---

**Variants & Base Images**
|---|---|---|---|


```bash
docker pull ghcr.io/openhands/agent-server:5b3b190-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-5b3b190-python \
  ghcr.io/openhands/agent-server:5b3b190-python
```

```
ghcr.io/openhands/agent-server:5b3b190-golang-amd64
ghcr.io/openhands/agent-server:5b3b190-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:5b3b190-golang-arm64
ghcr.io/openhands/agent-server:5b3b190-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:5b3b190-java-amd64
ghcr.io/openhands/agent-server:5b3b190-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:5b3b190-java-arm64
ghcr.io/openhands/agent-server:5b3b190-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:5b3b190-python-amd64
ghcr.io/openhands/agent-server:5b3b190-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:5b3b190-python-arm64
ghcr.io/openhands/agent-server:5b3b190-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:5b3b190-golang
ghcr.io/openhands/agent-server:5b3b190-java
ghcr.io/openhands/agent-server:5b3b190-python
```

- Each variant tag (e.g., `5b3b190-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `5b3b190-python-amd64`) are also available if needed


VascoSch92:
Re: [comment](https://github.com/OpenHands/software-agent-sdk/pull/2669#discussion_r3037063313) about `session` attribute breaking in pool mode:

Good catch — this is correct. In pool mode `self.session` was never set, leading to a raw `AttributeError`.

Fixed by:
- Storing the session in `self._session` (always initialized, `None` in pool mode)
- Adding a `session` property that raises an explicit `AttributeError` with actionable guidance in pool mode
- Updating `close()` to check `self._session is not None` instead of `hasattr(self, 'session')`
- Updating tests that assigned/deleted `executor.session` directly to use `executor._session`

VascoSch92:
Re: [comment](https://github.com/OpenHands/software-agent-sdk/pull/2669#discussion_r3037063317) about missing integration test for parallel execution through the framework:

Good suggestion. Added `test_pool_integration.py` with 3 tests:

1. **`test_pool_mode_opts_out_of_framework_locking`** — verifies `declared_resources()` returns empty keys in pool mode
2. **`test_subprocess_mode_serializes`** — verifies `declared_resources()` returns a resource key in subprocess mode
3. **`test_parallel_calls_execute_concurrently`** — issues 3 concurrent `TerminalExecutor.__call__` invocations, each sleeping 2s, and asserts total wall time is under 6s (serial would be >= 6s)

openhands-tools/openhands/tools/terminal/terminal/tmux_pane_pool.py
```diff
@@ -0,0 +1,274 @@
...
+    """A TmuxTerminal variant used inside a pane pool.
+
+    Overrides ``close()`` to only kill this terminal's window instead of
+    the entire shared tmux session.  This is critical because
+    ``TerminalSessionBase.__del__`` calls ``close()``, and GC of a cached
+    ``TerminalSession`` wrapper would otherwise destroy the session that
+    all other pool panes depend on.
+    """
+
+    def close(self) -> None:
```
csmith49, line 45: Q: if we're closing out the last window, is it safe to end the entire tmux session? Should we be doing some kind of reference-counting here?

VascoSch92: Good question. 

No, closing the last window here is safe and we don't need reference-counting.

`PooledTmuxTerminal.close()` only kills this terminal's **window**, never the session itself. The session lifetime is owned exclusively by `TmuxPanePool.close()` (line 131), which tears down the entire session in one shot via `self._session.kill()`.

The only reason `PooledTmuxTerminal.close()` exists at all is as a safety net: `TerminalSessionBase.__del__` calls `close()`, and the base `TmuxTerminal.close()` would kill the whole session. So this override narrows the blast radius to just the window. Even if all windows get killed individually (e.g. by GC), the tmux session object survives, tmux sessions persist even with zero windows.

openhands-tools/openhands/tools/terminal/impl.py
```diff
@@ -167,35 +346,90 @@ def __call__(
...
+        action: TerminalAction,
+        conversation: "LocalConversation | None" = None,
+    ) -> TerminalObservation:
+        """Execute *action* in pool mode with proper checkout/checkin.
+
+        All pane lifecycle (checkout, optional replace, checkin) is
+        contained here so there is exactly one checkout and one checkin
+        per call, and the checkin is guaranteed by ``finally``.
+        """
+        terminal = self._pool.checkout()  # type: ignore[union-attr]
```
csmith49: Minor suggestion: the checkout/checkin process is perfect for a context manager .

openhands-tools/openhands/tools/terminal/definition.py
```diff
@@ -234,6 +235,19 @@ def visualize(self) -> Text:
 class TerminalTool(ToolDefinition[TerminalAction, TerminalObservation]):
     """A ToolDefinition subclass that automatically initializes a TerminalExecutor with auto-detection."""  # noqa: E501
 
+    def declared_resources(self, action: Action) -> DeclaredResources:  # noqa: ARG002
+        from openhands.tools.terminal.impl import TerminalExecutor
```
csmith49: Minor nit: if there's a way to get the signal about the executor without a circular import we should do that. Is the `isinstance` check necessary or just defensive?

csmith49:
Quick question: is it possible to connect to or observe these tmux sessions directly from the terminal? No real use case but would be cool to watch what was happening.

openhands-tools/openhands/tools/terminal/definition.py
VascoSch92: I will do

```python
def declared_resources(self, action: Action) -> DeclaredResources:                                                                                                                                                                     
      if getattr(self.executor, 'is_pooled', False):
          return DeclaredResources(keys=(), declared=True)                                                                                                                                                                               
      return DeclaredResources(keys=("terminal:session",), declared=True)
```

openhands-tools/openhands/tools/terminal/impl.py
VascoSch92: Yes makes sense.

It add some lines of code but is cleaner.

VascoSch92:
> Quick question: is it possible to connect to or observe these tmux sessions directly from the terminal? No real use case but would be cool to watch what was happening.

 I think Yes! Both the single-session and pooled tmux backends use a dedicated tmux socket named openhands (see tmux_terminal.py:50 and tmux_pane_pool.py:115) to stay isolated from the user's own tmux. You can attach from any terminal:     

 ```
  # list sessions on the openhands socket                                                                                                                                                                                                
  tmux -L openhands ls                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                         
  # attach read-only so you don't steal input from the agent                                                                                                                                                                             
  tmux -L openhands attach -t openhands-pool-<user>-<uuid> -r
```

Is this what you wanted to know?

PR #2688, opened 2026-04-03
feat(workspace): make health_check_timeout configurable on DockerWorkspace and ApptainerWorkspace

### LINKED ISSUE #1571: DockerWorkspace/ApptainerWorkspace: Make health check timeout configurable
## Problem

the `DockerWorkspace` and `ApptainerWorkspace` classes have a hardcoded 120-second timeout for container health checkss in the `_wait_for_health` method:

```python
# openhands-workspace/openhands/workspace/docker/workspace.py:287
def _wait_for_health(self, timeout: float = 120.0) -> None:
    """Wait for the Docker container to become healthy."""
    ...

# openhands-workspace/openhands/workspace/docker/workspace.py:393
self._wait_for_health(timeout=120.0)  # Called from resume()
```

The same pattern exists in `ApptainerWorkspace`:

```python
# openhands-workspace/openhands/workspace/apptainer/workspace.py:294
def _wait_for_health(self, timeout: float = 120.0) -> None:
```

This hardcoded timeout is not exposed as a configurable field, making it impossible for users to:
- Increase the timeout for slower environments or larger images
- Decrease the timeout for faster environments to fail quickly

## proposed solution will be 

Add a `health_check_timeout` field to both `DockerWorkspace` and `ApptainerWorkspace`:

```python
health_check_timeout: float = Field(
    default=120.0,
    description="Timeout in seconds to wait for container health check to pass.",
)
```

Then use this field in `model_post_init` and `resume()`:

```python
self._wait_for_health(timeout=self.health_check_timeout)
```

## Use Cases

1. **Slow CI environments**: Users running on constrained CI runners may need longer timeouts
2. **Large agent-server images**: Custom images with additional dependencies may take longer to start
3. **Fast-fail scenarios**: Users may want shorter timeouts (e.g., 30s) to fail faster during development
4. **Debugging**: Ability to set very long timeouts when debugging container startup issues

## Related Files are

- `openhands-workspace/openhands/workspace/docker/workspace.py`
- `openhands-workspace/openhands/workspace/apptainer/workspace.py`

ixchio:
## What

Adds a `health_check_timeout` field (default `120.0` s) to both `DockerWorkspace` and `ApptainerWorkspace`, replacing the three hardcoded `120.0` literals that were previously scattered across `_start_container`, `resume`, and `model_post_init`.

## Why

The hardcoded value made it impossible to adapt to real-world conditions:

- **Slow CI runners** or constrained environments need more time for containers to start
- **Large custom agent-server images** (with extra deps) take longer to come up healthy
- **Fast-fail dev loops** — a 30 s timeout gives much quicker feedback when something is wrong
- **Debugging container startup** — setting a very long timeout is now trivial

## How

```python
# Slow CI / big image
DockerWorkspace(server_image="...", health_check_timeout=300.0)

# Fail fast locally
DockerWorkspace(server_image="...", health_check_timeout=30.0)

# Apptainer works identically
ApptainerWorkspace(server_image="...", health_check_timeout=180.0)
```

## Changes

- **`openhands-workspace/openhands/workspace/docker/workspace.py`** — adds `health_check_timeout: float = Field(default=120.0, ...)`; uses it in `_start_container` and `resume()`
- **`openhands-workspace/openhands/workspace/apptainer/workspace.py`** — same field; uses it in `model_post_init`
- **`tests/workspace/test_docker_workspace.py`** — 4 new tests covering field existence, default value, custom value round-trip, and `resume()` forwarding the timeout correctly; plus 3 tests for `ApptainerWorkspace`

## Tests

```
86 passed, 1 warning
```

Closes #1571


openhands-workspace/openhands/workspace/apptainer/workspace.py
```diff
@@ -128,6 +128,10 @@ class ApptainerWorkspace(RemoteWorkspace):
             "Specify locations to disable mounts for custom Apptainer behavior."
         ),
     )
+    health_check_timeout: float = Field(
+        default=120.0,
+        description=("Timeout in seconds to wait for container health check to pass."),
```
VascoSch92: nit

```suggestion
        description="Timeout in seconds to wait for container health check to pass.",
```

openhands-workspace/openhands/workspace/docker/workspace.py
```diff
@@ -121,6 +121,10 @@ class DockerWorkspace(RemoteWorkspace):
         default=None,
         description="Connect a container to the specified Docker network.",
     )
+    health_check_timeout: float = Field(
+        default=120.0,
+        description=("Timeout in seconds to wait for container health check to pass."),
```
VascoSch92: has before

tests/workspace/test_docker_workspace.py
```diff
@@ -178,3 +178,104 @@ def test_docker_network(mock_docker_workspace):
         assert "--network" in run_cmd
         network_index = run_cmd.index("--network")
         assert run_cmd[network_index + 1] == network_name
+
```
VascoSch92, line 179: nit: all the imports at the top of the file, thanks :-)

openhands-workspace/openhands/workspace/apptainer/workspace.py
ixchio: Fixed in 49279ae5

openhands-workspace/openhands/workspace/docker/workspace.py
ixchio: Fixed in 49279ae5

tests/workspace/test_docker_workspace.py
ixchio: Fixed in 49279ae5

VascoSch92 (commented):
Thanks for picking this up — the motivation is solid and the `resume()` integration test is well done. A few findings below, grouped by severity.

---

### Blocker — `test_docker_workspace_health_check_timeout_custom` doesn't test what it claims

`tests/workspace/test_docker_workspace.py:193-208`

This test patches both `_start_container` *and* `_wait_for_health`, constructs a workspace with `health_check_timeout=60.0`, and then asserts only that the field reads back as `60.0`. The mock on `_wait_for_health` is explicitly silenced with `_ = mock_wait`, and the inline comment even admits the call is never exercised.

Concrete consequence: if `docker/workspace.py:283` were reverted to `self._wait_for_health(timeout=120.0)` — the exact regression this PR exists to prevent — no test in this suite would fail. The entire PR's guarantee currently rides on the single `resume()` test at L211.

Please assert the custom value actually reaches the wait loop, mirroring the Apptainer test at L247:

```python
mock_wait.assert_called_once_with(timeout=60.0)
```

---

### Significant gap — no integration test for Docker's `_start_container` path

`tests/workspace/test_docker_workspace.py`

Apptainer has an integration test asserting the custom timeout reaches `_wait_for_health` via `model_post_init` (L247). Docker has no equivalent test for the startup path at `docker/workspace.py:283`. Only `resume()` (L425) is covered. Please add the analogous `_start_container`-scoped test.

---

### Should-fix — add `gt=0.0` validator on `health_check_timeout`

`docker/workspace.py:124`, `apptainer/workspace.py:131`

```python
health_check_timeout: float = Field(default=120.0)
```

Currently accepts `0` and negative values. A user passing `health_check_timeout=0` hits `while time.time() - start < 0:` in `_wait_for_health` (`docker/workspace.py:316-351`), which never enters the loop and immediately raises *"Container failed to become healthy in time"* without making a single attempt — misleading failure mode.

```python
health_check_timeout: float = Field(default=120.0, gt=0.0, description=...)
```

Also worth considering `ge=1.0`, given the hardcoded 1-second poll interval (see next comment).

---

### Should-fix — the poll interval is still hardcoded

`docker/workspace.py:326,350`, `apptainer/workspace.py:336,350`

`_wait_for_health` hardcodes `time.sleep(1)` between attempts and `urlopen(..., timeout=1.0)` per request. That means:

- A user setting `health_check_timeout=3.0` (for the "fast-fail dev loop" use case the PR motivation cites) gets ~3 real attempts.
- Sub-second timeouts effectively skip the check entirely.

The PR closes #1571's narrow ask, but the underlying UX goal (responsive failure) is only partially served. Either expose a `health_check_interval` field alongside this one, or at minimum extract the `1.0` into a module-level constant so the next PR can touch one place.

---

### Should-fix — field is duplicated; put it on a shared base

`docker/workspace.py:124`, `apptainer/workspace.py:131`

Both classes declare `health_check_timeout` with identical definitions and identical defaults. They both inherit `RemoteWorkspace → RemoteWorkspaceMixin` (`openhands-sdk/openhands/sdk/workspace/remote/remote_workspace_mixin.py:17`), which already owns sibling timeouts `read_timeout` and `max_connections` — same category.

The cleanest landing is a new intermediate `ContainerWorkspace(RemoteWorkspace)` that owns `health_check_timeout` plus the other fields already duplicated today (`host_port`, `forward_env`, `detach_logs`, `platform`, `extra_ports`, `enable_gpu`, `server_image`). If that's out of scope, adding the field to `RemoteWorkspaceMixin` is an acceptable minimum. As written, the defaults agree but will drift silently.

---

### Should-fix — drop the dead `= 120.0` default on `_wait_for_health`

`docker/workspace.py:316`, `apptainer/workspace.py:329`

Both signatures still declare `timeout: float = 120.0`, but every caller now passes `timeout=self.health_check_timeout`. That's three places (the field default + two method defaults) to keep in sync if the default ever changes. Either drop the default (make `timeout` required) or use a sentinel `None` meaning "use `self.health_check_timeout`".

---

### Nice-to-have — collapse the tautological tests

`tests/workspace/test_docker_workspace.py:180-244`

The four `*_field_exists` / `*_default` tests (two per class) re-test Pydantic rather than behavior. Collapse into one parametrized test:

```python
@pytest.mark.parametrize("cls", [DockerWorkspace, ApptainerWorkspace])
def test_health_check_timeout_default(cls):
    assert cls.model_fields["health_check_timeout"].default == 120.0
```

Saves ~30 lines and focuses coverage on the behavioral tests that actually matter.

---

### Nice-to-have — no timeout-elapses regression test

No test simulates a never-healthy container and asserts that the configured value (not `120.0`) is what bounds the wait. That's the most obvious regression guard for the feature as a whole and would catch the gap flagged in the blocker above from a different angle.

---

### Nice-to-have — no edge-case validation tests

No test pins behavior for negative / zero / `"30"` string coercion / `None`. If you add `gt=0.0` per the comment above, a two-line `ValidationError` test locks it in.

ixchio:
Thanks for the thorough review @VascoSch92! I've addressed the feedback in commit dee2322e:

**Addressed:**
- ✅ **Blocker**: Fixed the test to actually verify `_wait_for_health` is called with custom timeout
- ✅ **Significant gap**: Added `test_docker_workspace_startup_uses_health_check_timeout` for the `_start_container` path
- ✅ **gt=0.0 validator**: Added to both `DockerWorkspace` and `ApptainerWorkspace`
- ✅ **Dead default**: Removed `= 120.0` from `_wait_for_health` signature (now `*, timeout: float`)
- ✅ **Collapsed tests**: Combined `*_field_exists`/`*_default` into parametrized tests
- ✅ **Validation tests**: Added tests for non-positive timeout rejection

**Deferred (out of scope for this PR):**
- Poll interval constant: This would change the behavior scope - happy to do in a follow-up PR
- Shared base class: Agreed it's cleaner, but requires broader refactoring - tracked for future

_This comment was created by an AI assistant (OpenHands) on behalf of the author._

VascoSch92:
@openhands address the following two nits:
1. test_docker_workspace_resume_uses_health_check_timeout takes mock_docker_workspace but never uses it. The test re-patches execute_command and _start_container inline. Drop the fixture parameter to avoid confusion — tests/workspace/test_docker_workspace.py:232.                                                                                                                                                          
2. Stale comment in resume(). docker/workspace.py:422 still reads # Wait for health after resuming (use same timeout as initial startup). With the field unified, "same as initial startup" is now tautological — shorten to # Wait for container to be healthy, matching _start_container.

VascoSch92:
Addressed both nits in commit 4ddc6627:

1. ✅ Removed unused `mock_docker_workspace` fixture parameter from `test_docker_workspace_resume_uses_health_check_timeout`
2. ✅ Shortened comment in `resume()` to `# Wait for container to be healthy`, matching `_start_container`

Also fixed a latent bug: `execute_command` mock was out of scope during `resume()` call, causing the test to fail in environments without Docker.

_This comment was created by an AI assistant (OpenHands) on behalf of the contributor._

VascoSch92:
Thanks :-)

VascoSch92 (approved):

PR #2709, opened 2026-04-04
Primary-surface-first extraction ordering to prevent truncation-based bypass

### LINKED ISSUE #2706: Per-field extraction cap to prevent truncation-based bypass
## Problem

The defense-in-depth analyzers (PR #2472) apply a single 30,000-character budget across all extracted fields. Fields are processed in order: `tool_name`, then `tool_call.name`, then `tool_call.arguments` (exec segments); `thought` entries, then `reasoning_content`, then `summary` (text segments). An oversized early field starves later fields of scanning budget, making their content invisible to all downstream analyzers.

**Source:** `openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py`, lines 75-141.

### Threat model

The primary threat at the action boundary is indirect prompt injection — attacker-controlled content (documents, web pages, tool output) processed by the LLM, which then emits tool calls with side effects.

`tool_name` on `ActionEvent` has no length or content validation at any layer — it is a bare `str` field with no `max_length` constraint (see `action.py:41`, `message.py:32`). When the LLM emits a tool name not in the registry, the raw string flows through to the ActionEvent (`agent.py:766`) which is emitted via `on_event` and reaches the security analyzer before the error path runs. For non-function-calling models (via `fn_call_converter.py:80`), the extraction regex `[^>]+` allows arbitrary-length strings with no cap. OpenAI's API constrains native function call names to 64 characters, but other providers and the non-function-calling path do not. Research confirms LLMs can be coaxed into calling tool names not in the registry (Answer.AI, Jan 2026).

### Proof of concept

```
tool_name = "A" * 29990
tool_call.arguments = '{"command": "rm -rf /"}'
```

Total content is under 30,000 characters, but `arguments` receives at most 10 characters of budget. The `rm -rf /` payload is silently truncated. `PatternSecurityAnalyzer` returns `LOW`.

**Exploitability:** For models using native function calling with OpenAI (64-char name limit), the attack requires a different provider or proxy that does not enforce name length. For non-function-calling models (fn_call_converter path), the `[^>]+` regex imposes no length constraint and the attack is directly exploitable via prompt injection. The ActionEvent is emitted and analyzed before the tool-not-found error fires.

### Decision rationale

Per-field cap is the simplest fix that removes the structural weakness without changing the overall budget or any downstream behavior. The alternative — validating `tool_name` length upstream — would also be correct but is a separate concern (input validation vs. extraction resilience). The 50% cap ensures that even if an upstream validation gap is discovered later, the extraction pipeline does not have a structural weakness. The heuristic guarantees that if one pre-argument field is adversarially large, `arguments` still gets at least 15,000 characters. If both `tool_name` and `tool_call.name` simultaneously hit the cap, `arguments` would get 0 from the global budget — but this requires two adversarially large fields, which is harder to achieve than one.

## Proposed fix

In `utils.py`, add `_FIELD_CAP = _EXTRACT_HARD_CAP // 2` and enforce it in the `_add` helper in both `_extract_exec_segments()` and `_extract_text_segments()`:

```python
_FIELD_CAP = _EXTRACT_HARD_CAP // 2  # No single field > 50% of total budget

def _add(text: str, field_cap: int = _FIELD_CAP) -> None:
    nonlocal total
    remaining = min(_EXTRACT_HARD_CAP - total, field_cap)
    if remaining <= 0:
        return
    if len(text) > remaining:
        text = text[:remaining]
    segments.append(text)
    total += len(text)
```

Approximately 5 lines per extraction function. No changes to normalization or pattern matching.

### Test plan

- Oversized `tool_name` does not starve `tool_call.arguments`
- Oversized `tool_call.name` does not starve `tool_call.arguments`
- Oversized `thought` entries do not starve `summary`
- Total cap still honored when all fields are large
- Small fields extracted in full (no unnecessary truncation)
- End-to-end: oversized `tool_name` + malicious `arguments` returns `HIGH`
- Update `test_payload_past_hard_cap` xfail (field-starvation variant should now pass; genuine total-cap exceedance remains documented)

### Context

Identified in adversarial security review (2026-04-03) of PR #2472. Priority 1 in the red-team fix list.

Fieldnote-Echo:
## Summary

Closes #2706 by reordering extraction so the primary attack surface of each corpus is processed first, and removing the outer slice that would otherwise defeat that ordering in the composed analyzer path.

- **Exec segments** (`_extract_exec_segments`): `tool_call.arguments` is now extracted before `tool_name` and `tool_call.name`. Arguments is the primary attack surface for prompt-injection payloads.
- **Text segments** (`_extract_text_segments`): `summary` is now extracted before `reasoning_content` and `thought`. Summary describes the action the agent is about to take.
- **Composed content** (`_extract_content`): no longer applies an outer `[:_EXTRACT_HARD_CAP]` slice to the joined exec+text string. Per-corpus caps remain authoritative. This prevents the composed path from dropping the entire text corpus when exec fills the 30K budget.

No per-field truncation is imposed, so no blind spot is created for content that was visible under pre-cap behavior.

## What this fixes

The extraction pipeline processes fields until the 30K total budget is exhausted. Before this PR:

- A 30K `tool_name` could consume the entire exec budget and silently hide a malicious `arguments` payload. `tool_name` has no length validation anywhere in the SDK (`event/llm_convertible/action.py:41`; `tool/spec.py:29` only checks non-empty), and the non-function-calling path uses `[^>]+` with no length bound (`llm/mixins/fn_call_converter.py:80`).
- Three 10K `thought` entries could collectively exhaust the text budget and starve `summary` from the injection scanners.
- An outer `[:_EXTRACT_HARD_CAP]` slice in `_extract_content` dropped the entire text corpus from `all_content` whenever exec filled the budget, so injection patterns never saw `summary` in the composed analyzer path. This defeated the summary-first guarantee where it actually runs.

With primary-surface-first ordering and the composed-path fix, `arguments` and `summary` always see their full content up to the per-corpus 30K cap, regardless of the other field sizes.

## Why not a per-field cap

An earlier iteration imposed a 50% per-field cap, first with head-only truncation, then with head+tail. Stop-time adversarial review flagged both variants for introducing new blind spots: any per-field cap with a shared total budget truncates content that was visible under pre-cap behavior. Primary-surface-first ordering closes the #2706 starvation class without introducing a new truncation regression.

## What this does NOT fix (documented as xfail)

- **Payload past the 30K total cap.** Single-field content beyond `_EXTRACT_HARD_CAP` is invisible. Deliberate ReDoS trade-off, unchanged by this PR.

## Runtime bound

Regex patterns now scan up to `2 * _EXTRACT_HARD_CAP` chars in the composed content (exec corpus + text corpus + one separator space). All patterns in the module use bounded quantifiers, so this remains O(n) worst-case.

## Test plan

- [x] `TestPrimaryFirstOrdering`: arguments is the first exec segment; summary is the first text segment; oversized tool_name, tool_call.name, thought, and reasoning_content do not starve their respective primary fields.
- [x] `TestArgumentsFullRangeVisibility`: parametrized across positions 0, 1000, 7500, 14999, 15000, 22500, 29000. Every position within the 30K cap stays visible. Anchors the "no new blind spot" guarantee.
- [x] `TestSizeAccounting`: per-corpus cap honored; small fields extracted in full; 30K arguments skips subsequent fields but stays visible itself.
- [x] `TestEndToEnd`: `PatternSecurityAnalyzer.security_risk()` returns HIGH for #2706 attack patterns.
- [x] `TestComposedPathGuarantee`: summary visible in `_extract_content` when exec fills the budget; end-to-end HIGH for injection in summary under that condition; symmetric check for exec content visibility when text fills its corpus. Guards the composed path as a composition contract.
- [x] `TestResidualLimitations`: past-30K-cap documented as strict xfail.
- [x] `test_hard_cap_truncation` in `test_pattern.py` updated to assert the new bound (≤ 2x plus separator).
- [x] 236 passed, 6 xfailed; pre-commit 6/6; pyright 0/0/0.


openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
```diff
@@ -40,6 +40,11 @@
 # limit is invisible to the analyzer.
 _EXTRACT_HARD_CAP = 30_000
 
+# Per-field cap: no single field can consume more than 50% of the total
+# budget. Prevents an oversized early field (e.g., hallucinated tool_name)
+# from starving later fields of scanning budget. See issue #2706.
+_FIELD_CAP = _EXTRACT_HARD_CAP // 2
```
VascoSch92: use `Final` from the library `typing`

openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
```diff
@@ -82,9 +87,9 @@ def _extract_exec_segments(action: ActionEvent) -> list[str]:
     segments: list[str] = []
     total = 0
 
-    def _add(text: str) -> None:
+    def _add(text: str, field_cap: int = _FIELD_CAP) -> None:
```
VascoSch92: Do you need this new arguments? It seems that you are not using it. At least not in this PR.

openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
```diff
@@ -118,9 +123,9 @@ def _extract_text_segments(action: ActionEvent) -> list[str]:
...
```
VascoSch92: Could you add a very minimal docstring for what are doing these two interla methods? I know it has not to do with you PR but it will help to understand the code for the next person. Thanks :-)

VascoSch92 (commented):
Just a couple of comments. :-)

Fieldnote-Echo:
@VascoSch92 Thank you for your time & review. I'll polish it locally and implement the suggestions before pushing again.

Fieldnote-Echo:
@all-hands-bot @VascoSch92 @enyst Sorry for the delay without communication, I had a human API outage (aka the flu) and am returning to work today. Give me a beat to pick up the ball 🙏🏻

VascoSch92:
> @all-hands-bot @VascoSch92 @enyst Sorry for the delay without communication, I had a human API outage (aka the flu) and am returning to work today. Give me a beat to pick up the ball 🙏🏻

No problem at all :-) take your time

openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
Fieldnote-Echo: Done.

openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
Fieldnote-Echo: Done.

openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
Fieldnote-Echo: Done, thanks.

Fieldnote-Echo:
Follow-up push: a stop-time review caught that head-only truncation past the cap position created a false-negative regression. A payload at position 25K of a 30K `tool_name` was pre-cap visible (scanned up to the 30K total cap) and post-cap invisible (truncated to the first 15K).

Switched `_add` to head+tail truncation so both ends of an oversized field stay in the scan corpus. Positive test added for the end-of-field case; middle-of-field gap documented as xfail (closable fully only via reserved per-field budgets or argument-first extraction ordering, both out of scope here). One previously-xfailed test in `test_adversarial.py` became an xpass under head+tail and was converted to a passing test with updated docstring.

Fieldnote-Echo:
Follow-up push iterating after a second stop-time adversarial review. The per-field cap (head-only, then head+tail) was flagged both times for introducing a new blind spot: any per-field truncation with a shared total budget removes coverage for content that was visible under pre-cap behavior.

Switched approach to argument-first extraction ordering. `tool_call.arguments` is now processed before `tool_name` and `tool_call.name`, so arguments always receives its full scanning budget. No per-field cap is imposed, so no new truncation regression is introduced. PR title and description updated with the rationale and test plan.

@VascoSch92 this is a material change from the per-field cap you reviewed. Worth a fresh look when you have a moment, sorry for the churn.

Fieldnote-Echo:
Additional push: extended the ordering fix to text segments. `summary` is now extracted before `reasoning_content` and `thought` so that three 10K thought entries can no longer starve the summary field from the injection scanners. The previous xfail documenting this limitation is now a passing test.

PR title updated from "Argument-first" to "Primary-surface-first" to reflect that both corpora got the same treatment.

Fieldnote-Echo:
Third iteration after another stop-time adversarial review. The previous push fixed both `_extract_*_segments` functions to put the primary field first, but the composed path `_extract_content` still applied an outer `[:_EXTRACT_HARD_CAP]` slice on the joined exec+text string. When exec filled the 30K budget, that slice dropped the entire text corpus, so injection patterns never saw `summary` in the analyzer path. The per-corpus ordering was correct but defeated in composition.

Removed the outer slice. Per-corpus caps are now the single authority on runtime bounds; combined content is at most `2 * _EXTRACT_HARD_CAP` plus separator whitespace, still O(n) under the module's bounded-quantifier regex constraint. Added `TestComposedPathGuarantee` with 3 tests as the composition contract: summary visible in `_extract_content` when exec is full, end-to-end HIGH for that case, and the symmetric check for exec when text is full.

PR description updated. @VascoSch92 apologies for the iterative churn; the composition-level fix was the step I missed.

Fieldnote-Echo:
Fourth iteration, and credit to Codex for catching this one. The previous push claimed the joined content was bounded by `2 * _EXTRACT_HARD_CAP`, but the per-corpus `_add` closure tracked raw character count rather than joined length. A JSON object with many single-char leaves inflates the joined string via separators: 10K one-char leaves produce a 30K-char corpus that joins to ~40K. Combined exec+text could reach ~70K under pathological JSON.

Fixed `_add` in both corpora to track joined length including separators, so `" ".join(segments)` is now actually bounded by `_EXTRACT_HARD_CAP` per corpus and `2 * _EXTRACT_HARD_CAP + 1` combined. Added `test_composed_content_length_actually_bounded` to anchor the invariant against regression.

tests/sdk/security/defense_in_depth/test_field_cap.py
```diff
@@ -0,0 +1,419 @@
...
+No per-field truncation is imposed, so no blind spot is created for
+pre-cap scanned content: every position that was visible before this
+fix remains visible after.
+
+Residual limitation retained from the pre-cap design: content past the
+30K total cap within a single field remains invisible (deliberate ReDoS
+trade-off).
+"""
+
+from __future__ import annotations
```
VascoSch92: you don't need that

tests/sdk/security/defense_in_depth/test_field_cap.py
```diff
@@ -0,0 +1,419 @@
...
+        summary=summary,
+    )
+
+
+# -------------------------------------------------------------------
+# Argument-first ordering: arguments is always extracted first
+# -------------------------------------------------------------------
+
+
+class TestPrimaryFirstOrdering:
```
VascoSch92, line 78: nit: just make methods for the tests and use pytest.mark.parametrize when possible. In this way, tests are coincise and we don't have repeated ones :-)

VascoSch92:
Looks good to me :-)

I'd just love to get @csmith49's opinion, since he knows this subject better than I do.

tests/sdk/security/defense_in_depth/test_field_cap.py
Fieldnote-Echo: Applied. Noticed the 5 sibling files in this directory still carry the same import — want me to sweep them in this PR or open a follow-up?

tests/sdk/security/defense_in_depth/test_field_cap.py
Fieldnote-Echo: Applied.

tests/sdk/security/defense_in_depth/test_field_cap.py
```diff
@@ -0,0 +1,409 @@
+"""Tests for primary-field-first extraction ordering (issue #2706).
```
csmith49: Minor nit: we'd prefer comments that don't reference issues (here and in certain tests). If possible we should just put the relevant context into the file itself, which it looks like this doc-string does well.

csmith49 (approved):
Solid tweak, really appreciate the robust tests here.

openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
```diff
@@ -75,68 +75,110 @@ def _walk_json_strings(obj: Any) -> list[str]:
...
+    tool_call.name guarantees arguments always receives scanning budget even
+    when an earlier field is adversarially large (issue #2706). tool_name has
+    no length validation anywhere in the SDK; a 30K hallucinated name would
+    otherwise consume the full budget and hide the arguments payload.
     """
     segments: list[str] = []
     total = 0
 
     def _add(text: str) -> None:
+        """Append text to segments, tracking joined-length budget.
```
csmith49: A little annoying to refactor because it's a closure, but this is defined multiple times across several functions. Might make it easier to reason about extract caps if there's a single copy with the relevant parameters passed in.

Fieldnote-Echo:
@csmith49 thank you for the thorough review! I'll correct the nits and resolve the threads this evening.

tests/sdk/security/defense_in_depth/test_field_cap.py
Fieldnote-Echo: Done. Dropped the `#2706` refs from the module docstring, two test docstrings, the section divider comment, and the matching utils.py docstring.

openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py
Fieldnote-Echo: Done. Hoisted into a private `_BoundedSegments(cap)` class with one `add()` method, used by both `_extract_exec_segments` and `_extract_text_segments`. The cap is now an explicit constructor parameter and the joined-length tracking rationale lives on a single class docstring.

PR #2718, opened 2026-04-06
fix(security): close rm flag-ordering bypasses

### LINKED ISSUE #2707: Two-phase rm detection to eliminate flag-ordering bypasses
## Problem

The `rm -rf` detection pattern (`DET_EXEC_DESTRUCT_RM_RF`) in both `pattern.py` and `policy_rails.py` requires flags to appear immediately after `rm\s+`. This misses two bypass classes:

**Intervening flags:**
```bash
rm --no-preserve-root -rf /          # MISS
rm --no-preserve-root --recursive --force /  # MISS
```

**Post-argument flags:**
```bash
rm / -rf          # MISS
rm /etc -r -f     # MISS
```

GNU coreutils `rm` (tested: v9.10) accepts flags after positional arguments by default. This is standard GNU getopt behavior when `POSIXLY_CORRECT` is unset, which is the case on virtually all mainstream Linux distributions (Ubuntu, Debian, Fedora, Arch, RHEL). When `POSIXLY_CORRECT` is set, `-rf` after a positional argument is treated as a filename — but this is not the default anywhere relevant.

The `has_recursive_force` variable in `policy_rails.py` (line 87-93) gates the catastrophic-delete rail, so the rail shares the same blind spot.

**Source:** `pattern.py` lines 73-76, `policy_rails.py` lines 87-93.

### Threat model

An LLM emitting a tool call to execute a shell command. The command text is the executable corpus scanned by both analyzers. The bypass requires no encoding tricks or indirect dispatch — just a flag ordering that GNU rm accepts normally.

### Decision rationale

A two-phase imperative check (anchor on `\brm\b`, then scan the remainder for recursive+force flags independently) eliminates all flag-ordering bypasses without introducing new false positives. The shell operator boundary (`[^;|&]*`) prevents cross-command matching. The imperative approach is preferred over regex lookaheads for readability and independent testability.

## Proposed fix

Create `defense_in_depth/_detection_helpers.py` (not `utils.py` — that module's scope is extraction/normalization):

```python
def _is_rm_rf(text: str) -> bool:
    """Detect rm with both recursive and force flags in any position."""
    match = re.search(r"\brm\b([^;|&]*)", text, re.IGNORECASE)
    if not match:
        return False
    remainder = match.group(1)
    has_recursive = bool(
        re.search(r"(?:-[^\s]*[rR]|--recursive)\b", remainder, re.IGNORECASE)
    )
    has_force = bool(
        re.search(r"(?:-[^\s]*f|--force)\b", remainder, re.IGNORECASE)
    )
    return has_recursive and has_force
```

**PatternSecurityAnalyzer:** Call `_is_rm_rf` as a pre-loop check in `security_risk()` with early return to `HIGH`. Remove `DET_EXEC_DESTRUCT_RM_RF` from the compiled pattern list. Log the detector ID for telemetry.

**PolicyRailSecurityAnalyzer:** Replace `has_recursive_force` with `_is_rm_rf(seg)`. The catastrophic-delete rail gains the fix automatically. `sudo rm --no-preserve-root -rf /` is caught because `\brm\b` matches within `sudo rm ...`.

### Test plan

**Helper function:**
- `rm -rf /`, `rm -Rf /`, `rm -r -f /`, `rm --recursive --force /`, `rm --force --recursive /` — detect
- `rm --no-preserve-root -rf /`, `rm / -rf`, `rm /etc -r -f` — detect (currently missed)
- `sudo rm --no-preserve-root -rf /` — detects
- `rm -r /tmp/file` (no force), `rm -f file` (no recursive), `rm file` (plain) — NOT detect
- `rm -r /tmp; echo -f` — NOT detect (shell operator boundary)

**Integration:** Both analyzers return `HIGH` for all bypass variants. Catastrophic-delete rail fires for critical paths with bypass variants.

**Regression:** All current `rm` tests continue to pass.

### Out of scope

- Shell quoting bypasses (`r"m" -rf /`) — requires shell parser
- Variable indirection (`x=rm; $x -rf /`) — requires sandboxing
- New destructive-command patterns — covered by expanded pattern coverage issue

### Context

Identified in adversarial security review (2026-04-03) of PR #2472. Priority 2 in the red-team fix list.

kushalsai-01:
Fixes #2707

## Summary
This PR closes a security gap where `rm` commands could bypass safeguards by reordering `-r` and `-f` flags or inserting additional options between them.

The validation logic now correctly detects recursive and force flags regardless of ordering or intervening arguments, preventing evasion of the safety checks.

## Changes
- Normalize and validate `rm` flag combinations independent of order
- Handle cases with interleaved flags and options
- Strengthen command parsing to prevent bypass scenarios
- Add regression tests covering reported evasion patterns

## Testing
- Added unit tests for:
  - `rm -rf`, `rm -fr`
  - Flags separated by other options (e.g., `rm -r --verbose -f`)
  - Edge cases involving mixed/invalid ordering
- All existing tests pass

## Notes
This change is backward-compatible and only affects validation logic for potentially destructive `rm` commands.

## Checklist
- [x] Tests added/updated
- [x] Verified changes locally
- [x] CI passing (or expected to pass)

### DIFF
--- a/openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py
+++ b/openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py
@@ -25,7 +25,9 @@ from openhands.sdk.logger import get_logger
 from openhands.sdk.security.analyzer import SecurityAnalyzerBase
 from openhands.sdk.security.defense_in_depth.utils import (
     _extract_content,
+    _extract_exec_segments,
     _extract_exec_content,
+    _has_rm_recursive_force,
     _normalize,
 )
 from openhands.sdk.security.risk import SecurityRisk
@@ -69,15 +71,6 @@ DET_INJECT_IDENTITY = "inject.identity"
 
 DEFAULT_HIGH_PATTERNS: list[tuple[str, str, str]] = [
     # Destructive filesystem operations
-    (
-        r"\brm\b(?:(?![;&|]).){0,120}"
-        r"(?:(?:--recursive\b|-[^\s-]*[rR][^\s]*)"
-        r"(?:(?![;&|]).){0,120}(?:--force\b|-[^\s-]*f[^\s]*)"
-        r"|(?:--force\b|-[^\s-]*f[^\s]*)"
-        r"(?:(?![;&|]).){0,120}(?:--recursive\b|-[^\s-]*[rR][^\s]*))",
-        "Recursive force-delete (rm -rf variants)",
-        DET_EXEC_DESTRUCT_RM_RF,
-    ),
     (r"\bsudo\s+rm\b", "Privileged file deletion", DET_EXEC_DESTRUCT_SUDO_RM),
     (r"\bmkfs\.\w+", "Filesystem format command", DET_EXEC_DESTRUCT_MKFS),
     (r"\bdd\b.{0,100}of=/dev/", "Raw disk write", DET_EXEC_DESTRUCT_DD),
@@ -214,11 +207,17 @@ class PatternSecurityAnalyzer(SecurityAnalyzerBase):
 
     def security_risk(self, action: ActionEvent) -> SecurityRisk:
         """Evaluate security risk via two-corpus pattern matching."""
+        exec_segments = [_normalize(s) for s in _extract_exec_segments(action)]
         exec_content = _normalize(_extract_exec_content(action))
         all_content = _normalize(_extract_content(action))
 
         if not exec_content and not all_content:
             return SecurityRisk.LOW
+
+        # HIGH: rm recursive+force detection uses token parsing, not regex.
+        if any(_has_rm_recursive_force(seg) for seg in exec_segments):
+            logger.debug("Pattern matched: %s -> HIGH", DET_EXEC_DESTRUCT_RM_RF)
+            return SecurityRisk.HIGH
 
         # HIGH: patterns on executable fields only
         for pattern, _desc, det_id in self._compiled_high:
--- a/openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py
+++ b/openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py
@@ -70,8 +70,11 @@ DET_INJECT_IDENTITY = "inject.identity"
 DEFAULT_HIGH_PATTERNS: list[tuple[str, str, str]] = [
     # Destructive filesystem operations
     (
-        r"\brm\s+(?:-[frR]{2,}|-[rR]\s+-f|-f\s+-[rR]"
-        r"|--recursive\s+--force|--force\s+--recursive)\b",
+        r"\brm\b(?:(?![;&|]).){0,120}"
+        r"(?:(?:--recursive\b|-[^\s-]*[rR][^\s]*)"
+        r"(?:(?![;&|]).){0,120}(?:--force\b|-[^\s-]*f[^\s]*)"
+        r"|(?:--force\b|-[^\s-]*f[^\s]*)"
+        r"(?:(?![;&|]).){0,120}(?:--recursive\b|-[^\s-]*[rR][^\s]*))",
         "Recursive force-delete (rm -rf variants)",
         DET_EXEC_DESTRUCT_RM_RF,
     ),
--- a/openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py
+++ b/openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py
@@ -22,6 +22,7 @@ from openhands.sdk.logger import get_logger
 from openhands.sdk.security.analyzer import SecurityAnalyzerBase
 from openhands.sdk.security.defense_in_depth.utils import (
     _extract_exec_segments,
+    _has_rm_recursive_force,
     _normalize,
 )
 from openhands.sdk.security.risk import SecurityRisk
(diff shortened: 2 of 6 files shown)


kushalsai-01:
Hi @maintainers, I’ve addressed the issue related to rm recursive-force flag ordering bypass and added regression tests for the reported cases.

Could you please review when you get a chance? Happy to make any changes if needed.

kushalsai-01:
Thanks for the detailed review.

I see the concerns around the bounded regex (.{0,120}) allowing bypass via long arguments, as well as edge cases like redirection interleaving and post-argument flags. I’ll rework the detection to remove fixed-length constraints and move toward a more robust approach, along with adding the suggested regression tests and reducing duplication.

Pushing an update shortly.

kushalsai-01:
Addressed the review feedback:

* Replaced regex with token-based parsing (shlex)
* Removed fixed-length bounds to prevent bypass via long arguments
* Handled redirections and command segmentation correctly
* Added regression tests for post-argument flags and edge cases
* Refactored shared detection logic

Workflows are awaiting approval—once approved, CI should run.

Would appreciate a re-review.

kushalsai-01:
Hi @maintainers — could you please approve workflows so CI can run? I’ve rebased onto the latest main. Thanks.

Fieldnote-Echo:
@kushalsai-01, thanks for picking this up. As the issue author, the scope in #2707 is met: your changes catch intervening flags (`rm --no-preserve-root -rf /`), post-argument flags (`rm / -rf`), compact/reversed short flags (`-rf`, `-fr`), and long flags in any order. Bonus, your shlex-based tokenization also closes path-qualified commands (`/bin/rm -rf /`) and quote-based evasions (`r"m" -rf /`) that were on my follow-up list but out of scope for #2707.

A few things I liked specifically:
- Single `_has_rm_recursive_force` shared across `PatternSecurityAnalyzer` and the catastrophic-delete rail, no duplicated detection logic
- Env-assignment prefix handling (`FOO=bar rm -rf /`) and redirection-token skip (`rm -r 2>&1 -f /`) are both nice touches I didn't think to call out
- Negative tests on `germ -rf /`, `rm -r` alone, and `rm -f` alone, the kind of guard that prevents FP regressions later
- Conservative docstring ("does not attempt a full shell grammar") sets expectations for the next person

One note for reviewers: I support the tree-sitter direction in #2721, but that's a cross-package migration on its own timeline. This PR solves a real bypass class now and is merge-shaped and clean.

Your CI hasn't been able to run yet because workflows need maintainer approval for first-time contributors. @enyst, could you approve the workflow runs on this so we get a CI signal?

Fieldnote-Echo:
@kushalsai-01, quick context update.

The maintainers are moving the broader shell-parsing layer to tree-sitter-bash (issue #2721, phased plan in [VascoSch92's comment](https://github.com/OpenHands/software-agent-sdk/issues/2721#issuecomment-4439508548)). Phase 1 (PR #3237) is already open and drops the GPLv3 bashlex dependency. Phase 2 migrates the security analyzer onto the shared parser.

That doesn't make this PR obsolete. It makes it the demonstration that motivated the bigger shift. Showing the bypass class was real and that regex couldn't close it is the work that put a structural parser on the table. The shlex direction was the right call for the surface available at the time, and the negative tests (`germ -rf /`, `rm -r` alone, `rm -f` alone) set the false-positive discipline the tree-sitter version will inherit.

If you're interested in helping with the migration, Vasco split Phase 2 into focused chunks (2a-2f). The pieces that overlap with what you built here are Phase 2c (`policy_rails.py` rewrite, which extends your `_has_rm_recursive_force` direction to the full rail) and Phase 2d (replaces the hand-rolled shlex parsing with tree-sitter queries). Either would build on your work directly.

Thanks for picking this up. Your PR is the reason the discussion moved past "another regex tweak" to "structural fix."

smolpaws (commented):

Taste rating: acceptable

Findings:
- Changed area reviewed: openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py, openhands-sdk/openhands/sdk/security/defense_in_depth/policy_rails.py, openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py and 3 more file(s).

Risk assessment: medium.

Fieldnote-Echo:
I think we should probably close this PR rather than continue extending the regex parser path.

#3944 is not a direct replacement for this exact flag-ordering patch, but it is the current AST-backed shell detector line for #2721. If #2707 still needs coverage after that lands, I would reopen this as a smaller AST-semantic follow-up: parse `rm` option argv from the shell command view and keep the current flag-ordering regressions as tests.

That keeps the fix on the migration path instead of adding another regex branch we will need to retire.

enyst (commented):

🔴 **Needs improvement** - The token parser closes the reported flag-ordering bypass, but it opens simpler command-wrapper bypasses.

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/security/defense_in_depth/utils.py`, lines 265-284] **Security regression**: `_segment_has_rm_recursive_force()` only accepts `rm` as the first command token after bare environment assignments. On this exact head I verified that `command rm -rf /`, `env rm -rf /`, and `nohup rm -rf /` all return `False`, even though each invokes `rm -rf`; the removed regex matched the embedded `rm`. That gives an agent an easier bypass than the flag-ordering cases this PR fixes. Teach the parser about executable wrappers (or retain a safe fallback) and add regression cases for at least `command`, `env`, and `nohup` before replacing the old detector.

**[RISK ASSESSMENT]**
This directly changes a security analyzer for destructive shell commands. A confirmed false-negative path can let catastrophic deletion avoid the intended HIGH classification.

**Recommendation**: Do not auto-merge. Request review from a human security reviewer to validate wrapper handling and the parser's shell-command threat model.

**VERDICT:**
❌ **Needs rework**: The detector must not trade one `rm -rf` bypass for several simpler ones.

**KEY INSIGHT:**
A shell detector cannot equate “first token is `rm`” with “the executed utility is `rm`” when standard command wrappers are valid syntax.

---

>
>
>

PR #2723, opened 2026-04-06
feat(tools/gemini): declare file-path resources for parallel execution

VascoSch92:
## Summary
Adds declared_resources() to the three Gemini file tools (ReadFileTool, WriteFileTool, EditTool) so the framework can run operations on different files in parallel while serializing access to the same file.

## Changes
- read_file/definition.py: Added declared_resources() returning file:{normalized_path} — prevents reading partially-written content during a concurrent write.
- write_file/definition.py: Added declared_resources() returning file:{normalized_path} — serializes concurrent writes to the same file.
- edit/definition.py: Added declared_resources() returning file:{normalized_path} — serializes concurrent edits to the same file.

All three use the same file:{path} key format as FileEditorTool, so cross-tool locking works correctly (e.g., a read_file and edit on the same path serialize against each other, while operations on different files run in parallel).

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:304f27e-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-304f27e-python \
  ghcr.io/openhands/agent-server:304f27e-python
```

```
ghcr.io/openhands/agent-server:304f27e-golang-amd64
ghcr.io/openhands/agent-server:304f27e-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:304f27e-golang-arm64
ghcr.io/openhands/agent-server:304f27e-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:304f27e-java-amd64
ghcr.io/openhands/agent-server:304f27e-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:304f27e-java-arm64
ghcr.io/openhands/agent-server:304f27e-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:304f27e-python-amd64
ghcr.io/openhands/agent-server:304f27e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:304f27e-python-arm64
ghcr.io/openhands/agent-server:304f27e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:304f27e-golang
ghcr.io/openhands/agent-server:304f27e-java
ghcr.io/openhands/agent-server:304f27e-python
```

- Each variant tag (e.g., `304f27e-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `304f27e-python-amd64`) are also available if needed


openhands-tools/openhands/tools/gemini/write_file/definition.py
```diff
@@ -99,6 +101,14 @@ def visualize(self) -> Text:
 class WriteFileTool(ToolDefinition[WriteFileAction, WriteFileObservation]):
     """Tool for writing complete file contents."""
 
+    def declared_resources(self, action: Action) -> DeclaredResources:
+        """Lock on the target file path so concurrent writes to the same
+        file are serialized, while writes to different files run in parallel.
```
enyst, line 106: I guess technically, the same could be said about listing files? But idk if that matters in practice

enyst:
@openhands /codereview-roasted on this PR, post your feedback here as a comment

@smolpaws WDYT about this PR?

smolpaws:
🐾 **smolpaws review of PR #2723**

---

The idea is sound — Gemini file tools should declare their resources so the parallel executor can do its job. The implementation follows the existing `FileEditorTool` pattern, the diff is compact, CI is green. So far so good.

But there's a **correctness bug** that makes the locking decorative rather than functional:

### 🔴 Path resolution mismatch — the lock isn't protecting what you think

All three `declared_resources()` methods do:
```python
normalized_path = Path(action.file_path).resolve()
```

`Path.resolve()` canonicalizes relative paths against the **process CWD**. But the Gemini executors all resolve relative paths against **`self.workspace_root`**:
```python
if not os.path.isabs(file_path):
    resolved_path = self.workspace_root / file_path
```

If CWD ≠ workspace_root (which is the normal case — the agent-server process CWD is not `/workspace/project`), then `read_file(file_path="foo.py")` declares a lock on `file:/srv/agent-sdk/foo.py` while the executor actually reads `/workspace/project/foo.py`. Two completely different paths. The lock is theater.

This is **not** the same situation as `FileEditorTool`, which requires absolute paths (enforced by `validate_path()`). With absolute paths, `Path.resolve()` just normalizes dots, so `declared_resources` and the executor agree. The Gemini tools accept relative paths by design, so they need to resolve against the workspace root explicitly.

**Fix:** Store `workspace_root` on the tool instance during `create()`, and use it in `declared_resources()`:
```python
def declared_resources(self, action: Action) -> DeclaredResources:
    assert isinstance(action, ReadFileAction)
    path = Path(action.file_path)
    if not path.is_absolute():
        path = Path(self._workspace_root) / path
    return DeclaredResources(keys=(f"file:{path.resolve()}",), declared=True)
```

### 🟡 Tests verify format, not correctness

The tests assert `keys[0].startswith("file:")` and `keys[0].endswith("b.py")`. That passes even if the path in the middle is completely wrong. What's missing:

1. A test that passes a **relative** file_path and asserts the key resolves against the workspace root, not CWD.
2. A **cross-tool** test asserting that a Gemini `read_file(file_path="foo.py")` and a `FileEditorTool` operation on the absolute `/workspace/foo.py` produce the same resource key. That's the whole point of matching the `file:{path}` format.

### 🟡 Duplicated `_fake_conv_state`

Copy-pasted in all three test files. Should be a shared `conftest.py` fixture under `tests/tools/gemini/`. The repo guidelines even say so.

### Verdict

**Needs rework.** The PR is one `workspace_root` resolution away from being correct. Small fix, but without it the locking guarantee doesn't hold for the most common case (relative paths).

*I agree with the earlier review from @enyst — "if locking and execution do not canonicalize paths the same way, the lock is just theater." Concise and accurate. 🎭*

enyst:
Is the new agent’s review worth looking into?

VascoSch92:
@enyst thank you to push me to double check. Yes, there it was an error :-(. I was doing a lot of confusion with the paths.

It should be solved now.

PR #2747, opened 2026-04-07
fix(llm): cap auto-detected max_output_tokens when it fills the entire context window

csmith49:
## What/Why

When litellm's model registry reports max_output_tokens >= max_input_tokens (e.g. Nemotron: both 262144), the SDK would request the entire context window for output, leaving zero tokens for input. Every provider call was rejected, the condenser misinterpreted this as context overflow, and crashed on the near-empty history with NoCondensationAvailableException.

Cap auto-detected max_output_tokens to half the context window when it would otherwise consume the full window. Explicitly user-set values are not affected.

## Checklist

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:3d9e6da-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-3d9e6da-python \
  ghcr.io/openhands/agent-server:3d9e6da-python
```

```
ghcr.io/openhands/agent-server:3d9e6da-golang-amd64
ghcr.io/openhands/agent-server:3d9e6da-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:3d9e6da-golang-arm64
ghcr.io/openhands/agent-server:3d9e6da-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:3d9e6da-java-amd64
ghcr.io/openhands/agent-server:3d9e6da-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:3d9e6da-java-arm64
ghcr.io/openhands/agent-server:3d9e6da-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:3d9e6da-python-amd64
ghcr.io/openhands/agent-server:3d9e6da-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:3d9e6da-python-arm64
ghcr.io/openhands/agent-server:3d9e6da-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:3d9e6da-golang
ghcr.io/openhands/agent-server:3d9e6da-java
ghcr.io/openhands/agent-server:3d9e6da-python
```

- Each variant tag (e.g., `3d9e6da-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `3d9e6da-python-amd64`) are also available if needed


juanmichelini:
cool! testing here https://openhands-eval-monitor.vercel.app/?run=swebench%2Flitellm_proxy-converse-nemotron-super-3-120b%2F24097380443%2F&days=1&text=juan

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -1210,6 +1210,32 @@ def _init_model_info_and_caps(self) -> None:
...
+                    # window depending on the provider.
+                    context_window = self.max_input_tokens or self._model_info.get(
+                        "max_tokens"
+                    )
+                    if (
+                        context_window is not None
+                        and self.max_output_tokens is not None
+                        and self.max_output_tokens >= context_window
+                    ):
+                        capped = self.max_output_tokens // 2
```
enyst, line 1228: I think that's why we sometimes had 4096 or something like that, output tokens are not typically all that much in a single call. This works though! 🤔 

It just means the history is smaller when it reaches context error, than if we put some value like 4096, because half is more

csmith49: Does setting the max like that encourage models to generate more? Honestly I'm not sure. I'd expect we'll end up with very similarly-sized events as if we had set it at 4096.

enyst: 🤷 I don't know, I'm thinking about the reverse: setting half means that the LLM API provider will error _sooner_, because it adds that value to the prompt I think? So to the input tokens at the time of the request.

At least, I'm pretty sure Anthropic and OpenAI do that, and I thought the error message suggested it... I could be wrong though

csmith49: The relevant error message here is:

> You passed 2468 input characters and requested 262144 output tokens.
> However, the model's context length is only 262144 tokens, resulting in
> a maximum input length of 0 tokens (at most 0 characters). Please reduce
> the length of the input prompt.

Maybe the "requested" suggests that behavior? In which case this is probably worth escalating to LiteLLM, considering it's their registry that sets the output tokens the way it is.

PR #2771, opened 2026-04-09
feat: add --import-modules CLI flag for preloading custom modules

09473ZH:
## Problem

In source mode, the agent-server shares the same Python environment as the application code. However, custom tools registered via `register_tool()` at module import time are not available because the agent-server does not import application modules at startup.

The SDK has a dynamic registration mechanism (`tool_module_qualnames`) where the client sends module paths to the server during conversation creation. But in practice, there is a race condition — the dynamic import in `conversation_service.py` and the `resolve_tool()` call in `agent.init_state()` (which runs in a `ThreadPoolExecutor`) can race, causing `KeyError: "ToolDefinition 'MyCustomTool' is not registered"` and a 500 response.

## Solution

Add a `--import-modules` CLI argument to the agent-server entry point. It accepts a comma-separated list of module paths to import before the server starts, ensuring all custom tools are registered in the global registry before any conversation is created.

```bash
python -m openhands.agent_server \
  --import-modules "myapp.tools,myapp.plugins" \
  --port 8000
```

## Usage

In a Kubernetes deployment:

```yaml
command: ["python", "-m", "openhands.agent_server"]
args: ["--port", "8000", "--import-modules", "myapp.tools"]
```

Where `myapp/tools.py` imports all custom tool modules to trigger their `register_tool()` calls:

```python
import myapp.report_tool  # noqa: F401 — triggers register_tool()
import myapp.review_tool   # noqa: F401
```

## Backward Compatibility

No behavior change when `--import-modules` is not provided.


openhands-agent-server/openhands/agent_server/__main__.py
```diff
@@ -110,9 +111,23 @@ def main() -> None:
...
+    )
 
     args = parser.parse_args()
 
+    # Pre-load user-specified modules (e.g. custom tool registrations)
+    if args.import_modules:
+        for module_name in args.import_modules.split(","):
+            module_name = module_name.strip()
+            if module_name:
+                importlib.import_module(module_name)
```
VascoSch92: This can raise an ImportError, SyntaxError, or anything the module raises at import time, and crash the server with a raw traceback. 

The sibling code in `conversation_service.py:445-449` catches `ImportError` and logs. This new path should at minimum match that pattern, and probably do better (log and sys.exit(1)

A missing startup module is a deployment misconfiguration the operator wants to know about loudly.

openhands-agent-server/openhands/agent_server/__main__.py
```diff
@@ -110,9 +111,23 @@ def main() -> None:
         action="store_true",
         help="Check if browser functionality works and exit",
     )
+    parser.add_argument(
```
VascoSch92, line 138: The import runs before the `--check-browser` short-circuit. 

If someone runs `--check-browser --import-modules myapp.tools` in CI to validate the image, a broken module will fail the browser check for an unrelated reason.

VascoSch92 (commented):
There are zero tests for this new flag.

09473ZH:
Thanks for the catch @VascoSch92 — added tests in `tests/agent_server/test_preload_modules.py` (commit d026f110).

To make it unit-testable I pulled the inline preload loop in `main()` out into a top-level `preload_modules()` helper, following the same pattern as `check_browser()` in the same module (`tests/agent_server/test_check_browser.py`).

Coverage:

- `None` / empty string → no-op (matches the "no behavior change when not provided" claim in the PR body)
- single module
- comma-separated with surrounding whitespace
- empty segments skipped (`"a,,b"`, trailing comma)
- missing module → `ModuleNotFoundError` propagates (per the repo convention of not swallowing exceptions)
- a contract pair using a `tmp_path`-backed fake module whose body has an observable side effect: calling `preload_modules(qualname)` runs the body, while calling `preload_modules(None)` leaves it out of `sys.modules` entirely — this is the minimal reproduction of the broken state the flag exists to fix.

Not covered (deliberately):

- The full race between `conversation_service`'s dynamic `tool_module_qualnames` import and `agent.init_state()`'s `resolve_tool()` call — reproducing that needs a running agent-server plus a registered custom tool, which felt out of scope for a CLI flag test. Happy to add an integration test under a separate file if you'd prefer.
- The argparse wiring itself (`main()` passing `args.import_modules` into the helper) — `test_check_browser.py` sets the precedent of testing the helper directly rather than mocking uvicorn in `main()`, so I followed that.

Let me know if you'd like the integration test too.

xingyaoww:
@OpenHands please address these two failures

openhands-agent-server/openhands/agent_server/__main__.py
xingyaoww: Fixed in 15806292. Added try/except around `importlib.import_module()` that logs the module name and error before re-raising, matching the `conversation_service.py` pattern. This preserves fail-fast at startup while giving operators clear diagnostics.

openhands-agent-server/openhands/agent_server/__main__.py
xingyaoww: Fixed in 15806292. Moved `preload_modules()` call after the `--check-browser` short-circuit so import errors cannot mask the browser check.

openhands-agent-server/openhands/agent_server/__main__.py
```diff
@@ -110,9 +127,20 @@ def main() -> None:
...
+        default=None,
+        help=(
+            "Comma-separated list of modules to import at startup "
+            "(e.g. 'myapp.tools,myapp.plugins')"
+        ),
+    )
 
     args = parser.parse_args()
 
+    preload_modules(args.import_modules)
```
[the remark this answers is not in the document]
xingyaoww: Fixed in 15806292.

openhands-agent-server/openhands/agent_server/__main__.py
```diff
@@ -15,6 +16,22 @@
...
+    Used to register custom tools before any conversation is created, avoiding
+    a race with dynamic `tool_module_qualnames` import in conversation_service.
+    """
+    if not modules_arg:
+        return
+    for module_name in modules_arg.split(","):
+        module_name = module_name.strip()
+        if not module_name:
+            continue
+        importlib.import_module(module_name)
```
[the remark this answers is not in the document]
xingyaoww: Fixed in 15806292.

xingyaoww (approved):
LGTM!

PR #2792, opened 2026-04-10
feat: Add SSH service for local VSCode Remote-SSH access

gpothier:

---

## Why

With the "Remote - SSH" extension, a local VSCode (or downstream fork) instance can connect to a remote machine and run everything except UI from there: git cloning, running the app, etc. Although the OpenHands UI embeds VSCode, that embedded version can be less useful than the local "desktop" version. For instance, for debugging:

imagen

This PR (along with the sister PR in the OpenHands repo) enables one-click opening of a local VSCode window connected to the sandbox:

imagen

## Summary

Add SSH server support to agent-server containers, enabling users to connect via their local VSCode with the Remote-SSH extension.

There is a "sister" PR in the OpenHands repo that completes this feature with a settings page to add SSH keys and a button to directly open the local VSCode: <https://github.com/OpenHands/OpenHands/pull/13844>

#### Dockerfile
- Install openssh-server package
- Generate SSH host keys (ed25519, rsa, ecdsa) at build time
- Configure sshd for non-root operation (port 2222, key-based auth)
- Set proper permissions on host keys for openhands group

#### SSH Service (ssh_service.py)
- New service to manage SSH server lifecycle
- Populates ~/.ssh/authorized_keys from OH_SSH_PUBLIC_KEYS env var
- Starts sshd as non-root user on port 2222
- Integrated with agent-server startup

#### Configuration
- Add SSH_PORT (default: 2222) to config
- Add OH_SSH_PUBLIC_KEYS environment variable support (see sister PR for how where this comes from)

### Technical Details

- SSH server runs on port 2222 (non-privileged)
- Host keys readable by openhands group (mode 640)
- PAM disabled (requires root)
- Public key authentication preferred
- Password authentication available as fallback

## How to Test

When both this PR and the sister PR are in use, a new button appears in the Code button's popup (see screenshot above). This button opens a new VSCode window connected to the sandbox through SSH. When the window opens for the first time, VSCode asks to accept the remote SSH host ID, like when one connects to a new SSH host (just type `yes`).

## Video/Screenshots

imagen

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

There seems to be a bug in VSCode itself with multiple VSCode windows connecting to the same SSH host at different ports, where it wrongly assumes this is the same host. So currently this feature only allows one local VSCode window at a time, at least until this is resolved on the VSCode side.


enyst:
Thank you so much for this!

@openhands Do a /codereview on this PR. Make sure you read the sister PR too.

openhands-agent-server/openhands/agent_server/docker/Dockerfile
```diff
@@ -265,11 +265,46 @@ RUN chown -R ${USERNAME}:${USERNAME} ${NOVNC_WEB}
...
+ARG SSH_PORT=2222
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends openssh-server; \
+    apt-get clean; rm -rf /var/lib/apt/lists/*; \
+    # Create SSH runtime directory
+    mkdir -p /run/sshd; \
+    chmod 755 /run/sshd; \
+    # Generate host keys
+    ssh-keygen -A; \
```
enyst: Generating the SSH host keys at image-build time bakes the same private keys into every container built from this image. That defeats host identity verification: VS Code/SSH will see unrelated runtimes as the same host, and anyone with access to the image can impersonate another sandbox's host key. These keys need to be generated per container start (or injected/mounted per runtime), not once in the Dockerfile layer.

openhands-agent-server/openhands/agent_server/docker/Dockerfile
```diff
@@ -265,11 +265,46 @@ RUN chown -R ${USERNAME}:${USERNAME} ${NOVNC_WEB}
...
+    sed -i 's/#PermitEmptyPasswords no/PermitEmptyPasswords no/' /etc/ssh/sshd_config; \
+    sed -i 's/#PubkeyAuthentication yes/PubkeyAuthentication yes/' /etc/ssh/sshd_config; \
+    sed -i 's/#AuthorizedKeysFile/AuthorizedKeysFile/' /etc/ssh/sshd_config; \
+    # Configure SSH to listen on custom port (non-privileged port, no root needed)
+    sed -i "s/#Port 22/Port ${SSH_PORT}/" /etc/ssh/sshd_config; \
+    echo "Port ${SSH_PORT}" >> /etc/ssh/sshd_config; \
+    # Allow sshd to run on non-privileged port without root
+    sed -i 's/#UsePAM yes/UsePAM no/' /etc/ssh/sshd_config; \
+    # Set password for openhands user (password: "openhands")
+    echo "${USERNAME}:openhands" | chpasswd; \
```
enyst: This enables password auth and sets a globally known credential (`openhands` / `openhands`). In the sister PR, the SSH endpoint is exposed directly to the user through `sandbox.exposed_urls`, so there is no app-level auth once someone can reach `host:port`. That means anyone who can hit the runtime can log in, even when the user configured SSH keys. I think we should disable `PasswordAuthentication` entirely here and require the per-user public keys instead.

openhands-agent-server/openhands/agent_server/api.py
```diff
@@ -126,11 +128,22 @@ async def start_tool_preload_service():
...
+                logger.warning("SSH service failed to start, continuing without SSH")
+        else:
+            logger.info("SSH service is disabled")
+
     # Start all services concurrently
     results = await asyncio.gather(
         start_vscode_service(),
         start_desktop_service(),
         start_tool_preload_service(),
+        start_ssh_service(),
```
enyst, line 146: We start the SSH service here, but the shutdown path below still only stops VSCode, desktop, and tool-preload. That leaves `sshd` unmanaged across FastAPI lifespan restarts/tests and can leak the child process or keep the port bound. Please add a matching `stop_ssh_service()` in the `finally` `gather(...)` block.

enyst (commented):
I read this together with the sister PR (OpenHands/OpenHands#13844). I like the overall direction, but I found a few important issues that should be fixed before this ships. The inline comments call out the details.

_This review was created by an AI assistant (OpenHands) on behalf of the user._

smolpaws (commented):

Taste rating: needs improvement

Critical issue:
- openhands-agent-server/openhands/agent_server/config.py: `enable_ssh` defaults to `True`. The new service starts an SSH daemon and binds it on `0.0.0.0` when public keys are configured. That expands the default network attack surface for every agent-server deployment that carries `OH_SSH_PUBLIC_KEYS`, including deployments that did not explicitly opt in to SSH.

Concrete suggestion:
- make SSH opt-in by default, with `enable_ssh=False`
- document the deployment setting needed to enable it
- keep the current no-keys safety check, but do not rely on it as the feature gate

Risk assessment: high. A new remotely reachable service should not be enabled by default.

Verdict: needs rework

Key insight: remote access features need an explicit enable switch, not a default-on listener.

enyst (commented):

🔴 **Needs improvement**

**[TESTING GAPS]**
- [Changed-file manifest] **No regression coverage for the SSH boundary**: this adds 438 lines across a new SSH daemon lifecycle, host-key generation, `authorized_keys` handling, configuration, and a public `/ssh_status` endpoint, but the PR changes no tests. This is not a harmless helper: it creates a new remote shell entry point in every agent-server image. Add tests that exercise the real environment-provided key input path, no-key refusal, configured-port behavior, process start/stop cleanup, and the status endpoint. The package guidance also calls for a live-server test for endpoint/server-behavior changes; for this feature, a container smoke test that actually authenticates with a generated key would prove the integration rather than only its individual mocks.

**VERDICT:**
❌ **Needs rework**: A security-sensitive remote-access service should not ship without executable regression proof for authentication and lifecycle behavior.

**KEY INSIGHT:**
The implementation may work manually, but there is currently nothing in CI that protects its authentication and child-process invariants.

**[RISK ASSESSMENT]**
This expands the container attack surface with an SSH daemon and remote shell access, changes the base image, and introduces process/authentication lifecycle code without automated coverage. **Recommendation**: Do not auto-merge. Request review from a human security/agent-runtime reviewer and require a real SSH integration test.

---

>
>
>

Issue #2042, opened 2026-02-13
feat(skills): Run skills via subagent delegation for isolated execution


simonrosenberg:
## Summary

**Skills** are reusable, specialized prompts (like "debug with Datadog", "review PR", "generate tests") that the agent can invoke. Today, when a skill runs, all of its intermediate work — every tool call, API query, reasoning step — happens directly in the main conversation. This clutters the user's context with dozens of intermediate steps they don't care about.

### The Problem — Concrete Example

Imagine a **Datadog debugging** skill. To find a root cause, the agent might:
1. Query Datadog for pod metrics
2. Query Datadog for logs around the crash time
3. Correlate timestamps with events
4. Query traces for the failing endpoint
5. Check memory/CPU metrics
6. ...50+ queries total

Without isolation, all 50+ queries and their results flood the main conversation. The user just wanted: *"The pod is OOMKilled because the batch job allocates 4GB but the limit is 2GB."*

### How Claude Code Solves This

Claude Code runs the skill as a **subagent** via the Task tool (triggered by `context: fork` in skill frontmatter):
- A fresh, isolated context is created (NOT a clone of the main conversation)
- The skill content becomes the subagent's task prompt
- The subagent gets a system prompt from the agent type (Explore, Plan, etc.) plus CLAUDE.md
- The subagent has **no access to main conversation history**
- The subagent can make as many tool calls as needed without polluting the parent
- When done, the subagent's final output is returned to the main conversation

This is just the standard subagent/delegation mechanism — no conversation forking or cloning is involved.

## Proposed Solution

Wire the skill system to run skills via the existing delegation mechanism (`_delegate_tasks()`) when `context: fork` is set in skill frontmatter. The skill content becomes the subagent's task prompt, and the subagent runs with its own fresh context.

### Key Requirements

- Add `context: fork` support to skill frontmatter
- When a skill has `context: fork`, execute it via the delegation system instead of inline
- The subagent gets a fresh context with: skill content as prompt + CLAUDE.md
- The subagent does NOT get the main conversation history
- The subagent's final output is returned as the result to the main conversation
- Existing inline skill functionality is preserved for skills without `context: fork`

### Key Files

- `openhands-sdk/openhands/sdk/context/skills/skill.py` — Skill definition and loading
- `openhands-tools/openhands/tools/delegate/impl.py` — Existing delegation mechanism

## Acceptance Criteria

- [ ] Skills with `context: fork` execute via the delegation system, not inline
[...]

PR #2823, opened 2026-04-14
feat(skills): add context=fork for isolated subagent execution (1/2)

VascoSch92:

## Summary      
                                                                                                                                                                                                                                         
  Adds `context: fork` support to skill frontmatter (ref #2042, part 1).                                                                                                                                                              
                                                                                                                                                                                                                                         
  When a triggered skill has `context: fork`, its content is executed as a                                                                                                                                                               
  fresh subagent conversation instead of being injected inline. Only the
  subagent's final output is returned to the parent, so skills that make                                                                                                                                                                 
  many intermediate tool calls (e.g. a Datadog debugger with 50+ queries)                                                                                                                                                                
  no longer flood the parent context.                                                                                                                                                                                                    
                                                                                                                                                                                                                                         
  ## Changes                                                                                                                                                                                                                             
                  
  - `Skill.context: Literal["inline", "fork"]` (default `"inline"`), parsed                                                                                                                                                              
    from frontmatter; validator rejects `fork` without a trigger.
  - `openhands.sdk.skills.fork.run_skill_forked` — spawns a sub-`Conversation`                                                                                                                                                           
    with the parent's agent stripped of its skills (prevents re-trigger                                                                                                                                                                  
    recursion), preserves the rest of `agent_context`.                                                                                                                                                                                   
  - `AgentContext._resolve_skill_content` — single branch point for inline vs                                                                                                                                                            
    fork; forked skills without `agent`/`working_dir` fall back to inline                                                                                                                                                                
    with a warning.                                                                                                                                                                                                                      
  - Forked subconversations persist under                                                                                                                                                                                                
    `<parent_persistence_dir>/forks/<skill_name>/` when the parent has                                                                                                                                                                   
    persistence enabled; ephemeral otherwise.                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  ## Scope                                                                                                                                                                                                                               
                                                                                                                                                                                                                                         
  This PR implements **path 1** from the issue (triggered-skill injection).                                                                                                                                                              
  Path 2 (progressive-disclosure `invoke_skill` tool) is deferred to a
  follow-up.     

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore













---

**Variants & Base Images**
|---|---|---|---|


```bash
docker pull ghcr.io/openhands/agent-server:da22bb6-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-da22bb6-python \
  ghcr.io/openhands/agent-server:da22bb6-python
```

```
ghcr.io/openhands/agent-server:da22bb6-golang-amd64
ghcr.io/openhands/agent-server:da22bb6-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:da22bb6-golang-arm64
ghcr.io/openhands/agent-server:da22bb6-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:da22bb6-java-amd64
ghcr.io/openhands/agent-server:da22bb6-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:da22bb6-java-arm64
ghcr.io/openhands/agent-server:da22bb6-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:da22bb6-python-amd64
ghcr.io/openhands/agent-server:da22bb6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:da22bb6-python-arm64
ghcr.io/openhands/agent-server:da22bb6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:da22bb6-golang
ghcr.io/openhands/agent-server:da22bb6-java
ghcr.io/openhands/agent-server:da22bb6-python
```

- Each variant tag (e.g., `da22bb6-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `da22bb6-python-amd64`) are also available if needed


openhands-sdk/openhands/sdk/skills/fork.py
```diff
@@ -0,0 +1,83 @@
...
+    fork_persistence_dir: str | None = None
+    if persistence_dir is not None:
+        fork_persistence_dir = str(Path(persistence_dir) / "forks" / skill.name)
+
+    sub_conv = Conversation(
+        agent=sub_agent,
+        workspace=str(working_dir),
+        persistence_dir=fork_persistence_dir,
+        visualizer=None,
+        stuck_detection=False,
```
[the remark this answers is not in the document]
VascoSch92: Double-checked this, `delete_on_close` on `LocalConversation` does not touch persisted files. 

Looking at `LocalConversation.close()` (`local_conversation.py:843-859`), the only thing gated by `self.delete_on_close` is tool executor cleanup (closing terminal subprocesses and file-editor handles). The `FileStore` under `persistence_dir` is never touched. This matches the comment in t`ests/tools/terminal/test_conversation_cleanup.py:37` which documents the flag as "required "to trigger executor cleanup."                                                                                                                                                                              

So the current behaviour is what we want: forked subconversations persist their events to <parent>/forks/<skill_name>/ for debugging, while their terminal/file-editor processes are cleaned up promptly. Agreed the name is misleading, but it's an SDK-wide concern rather than something to work around in this PR.                                                                                                                                                                                                                                 

(Note: RemoteConversation.close() at remote_conversation.py:1341 does issue a remote delete — different semantics, but not the path taken here.)

VascoSch92:
> [RISK ASSESSMENT]
>[Overall PR] ⚠️ Risk Assessment: 🟡 MEDIUM
>This PR changes the agent execution model for triggered skills (opt-in via context: fork). While code quality is solid and the feature is additive-only, it modifies how skills execute and could affect benchmark performance. Per repository guidelines, agent behavior changes require human approval after lightweight eval verification, even when code quality is excellent.

Actually, I'm not sure this is true, because no skill is currently using this feature (at least none of ours).

I'm in favor of adding an example once this PR and its sibling #2889 are merged (and updating the docs accordingly).

enyst:
@openhands /codereview /codereview-roasted in Linus style according to the principles he applies! 

Post your review as a github review with gh api and event. Then take a deep breath and look at your review *with fresh eyes*. Look at every section of it and answer: is as major as it seemed? is it as minor as it seemed? Post a second review with gh api and event.

enyst (commented):
🔴 **Taste rating: Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/skills/fork.py:66-97`] **Isolation/Data Flow**: This is not actually an isolated fork. `agent.model_copy(update={"agent_context": ...})` on a live, already-initialized agent keeps the private runtime state around. In practice that means the forked conversation inherits `_initialized` and the same `_tools` / executor / LLM / subprocess references by alias, not by construction. Then `sub_conv.close()` tears those shared resources down through the normal close path. For a plain `Agent`, `LocalConversation.close()` closes tool executors out from under the parent. For `ACPAgent`, `close()` can terminate the shared subprocess/session. That's the exact opposite of "fresh subagent conversation" — it's a shallow copy of a live runtime object that can saw off the branch the parent is sitting on.

**[TESTING GAPS]**
- [`tests/sdk/skills/test_skill_fork_execution.py`] **Missing real regression**: The end-to-end test only covers the easy path: `Agent`, no tools, `Agent.close()` is a no-op. So it never exercises the thing that matters here: initialized parent runtime state. You need one regression test that proves a fork does **not** invalidate an already-initialized parent agent with real executors/session state.

**[RISK ASSESSMENT]**
Opt-in feature, so this doesn't randomly break existing users. But the new code path is supposed to provide isolation and currently aliases live runtime state instead. That's a correctness bug in the core feature, and it touches agent execution behavior, so I would not wave it through without a redesign and lightweight eval verification from a human maintainer.

**VERDICT:**
❌ **Needs rework**: The core idea is useful, but the current implementation forks configuration and runtime state together. Those are not the same thing.

**KEY INSIGHT:**
You're trying to fork the agent's *definition*, but this code is actually cloning the agent's *live process state*.

_This review was created by an AI assistant (OpenHands) on behalf of the user._

enyst (commented):
Fresh-eyes pass on my own review:

- The **shared-runtime-state** issue is still as major as it looked. I re-traced the actual callees (`model_copy` on the live agent, then the sub-conversation close path), and the conclusion did not get softer: this implementation aliases runtime objects that are supposed to be isolated.
- The **testing gap** is real, but it's derivative, not an independent blocker. The problem with the test is simply that it only covers the no-tools / no-subprocess path, so it misses the real bug.
- The other things I considered mentioning — persistence path wording, fallback semantics, lock duration in `send_message()`, PR evidence formatting — are smaller than they first looked and not the story here. I would drop all of that noise.

So the fresh-eyes version is simpler: one real blocker, not five fake ones.

**Still blocking:**
- [`openhands-sdk/openhands/sdk/skills/fork.py:66-97`] The code copies a live initialized agent instead of reconstructing a fresh agent/runtime boundary, and the fork close path can invalidate the parent.

**Updated risk take:**
- Still 🟡 **MEDIUM** overall. It's opt-in, so not a repo-wide fire. But the one path it adds is not actually isolated, which is exactly the property the feature is promising.

_This follow-up review was created by an AI assistant (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/skills/fork.py
```diff
@@ -0,0 +1,98 @@
...
+
+if TYPE_CHECKING:
+    from openhands.sdk.agent.base import AgentBase
+    from openhands.sdk.skills.skill import Skill
+
+
+logger = get_logger(__name__)
+
+
+def run_skill_forked(
```
xingyaoww, line 97: Will it potentially break observabilities if we run this on cloud, where we supposed to define a remote conversation? Maybe we should create two examples for this feature, one for local workspace, one for remote agent server.

VascoSch92: Good catch!
yes, this is a real gap for the remote path.

run_skill_forked unconditionally builds the sub-conversation with workspace=str(working_dir) (a plain string), so the Conversation(...) factory always resolves to a LocalConversation. When the parent is a RemoteConversation, the fork still runs on the agent server (since _resolve_skill_content executes inside the agent step loop), but it's built with none of the parent's observability wiring:
- no tags → fork spans are orphans in Datadog/LangSmith dashboards; no parent_conversation_id to roll up                                                                                                                               
- no callbacks / token_callbacks → any listener that ships telemetry doesn't see fork events                                                                                                                                           
- no hook_config → session_start / session_end hooks don't fire for the fork                
- no plugins → plugins the parent loaded (some install observability hooks themselves) are absent                                                                                                                                      
                                                                                                                                                                                                                                         
So functionally it runs, but on the cloud product where the value prop is "every conversation is observable end-to-end", a fork is a dark spot.                                                                                        

Should I do the change already in this PR or in a following one? 
The feature will not crush in remote settings but will be a dark spot. 

I don't have preferences.

VascoSch92:
> The shared-runtime-state issue is still as major as it looked. I re-traced the actual callees (model_copy on the live agent, then the sub-conversation close path), and the conclusion did not get softer: this implementation aliases runtime objects that are supposed to be isolated.

the reviewer is right, and the concrete mechanism is worse than "shared agent object" — it's shared tool executors that the sub-conversation's close() will forcibly tear down. I updated the code accordingly (actually as we do in subagents).

For the other two points.

- Workspace deletion: not a real risk. close() does not delete the workspace dir, only closes tool executors. The scary-sounding delete_on_close name is misleading.                                                                   
- agent_context shallow copy (secrets, system_message_suffix): theoretically shared, but these are effectively read-only during a run. Not a blocker.

enyst:
@openhands  /codereview /codereview-roasted in Linus style according to the principles he applies!

Post your review as a github review with gh api and event. Then take a deep breath and look at your review *with fresh eyes*. Look at every section of it and answer: is as major as it seemed? is it as minor as it seemed? Post a second review with gh api and event.

enyst (commented):
🔴 **Taste Rating**: Needs improvement

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/context/agent_context.py:286-310`, `openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py:692-700`, `openhands-sdk/openhands/sdk/skills/fork.py:78-113`] **Execution Context / Safety**: The old bug is fixed — you're no longer aliasing the live agent runtime — but this still is not actually preserving the parent conversation's execution context. The fork path only forwards `agent`, `working_dir`, and `persistence_dir`, then builds a brand-new `Conversation(...)` with default conversation state. In this codebase, confirmation policy, security analyzer, and secrets are conversation-scoped, not agent-scoped. So a parent running `AlwaysConfirm`, `ConfirmRisky`, or a custom analyzer silently degrades to `NeverConfirm` / `None` inside the fork, and any secrets added to the parent conversation are dropped. That is not a small detail: it breaks the safety contract and also undercuts the motivating Datadog-style use case, because the forked skill can lose the credentials it needs.

**[TESTING GAPS]**
- [`tests/sdk/skills/test_skill_fork_execution.py`] **Missing regression for preserved conversation state**: the tests are solid for the runtime-isolation fix, but they only exercise the default path: plain `Agent`, no secrets, default confirmation/security settings. They do not prove that a fork inherits the parent's conversation-level execution context, and right now that is exactly the part that looks wrong. One real regression here would be enough: parent with non-default confirmation/security or parent with secrets, trigger a fork, and verify the child sees the same behavior/state.

**[RISK ASSESSMENT]**
Opt-in feature, so this does not randomly hit every user. But the one path it adds can silently drop conversation-level safety and credentials, which is both a correctness problem and a policy problem. I would not call this merge-ready without fixing that boundary and then letting a human maintainer decide on the eval-risk side.

**VERDICT:**
❌ **Needs rework**: the agent/runtime aliasing bug is fixed, but the fork still loses the parent conversation context that actually governs execution.

**KEY INSIGHT:**
You stopped copying the parent's *live agent state*, but the new fork still forgets the parent's *live conversation state* — and that is where the safety and secret plumbing lives.

_This review was created by an AI assistant (OpenHands) on behalf of the user._

enyst (commented):
Fresh-eyes pass on my own review:

- The earlier **shared-runtime-state** blocker is gone. Rebuilding the agent from model fields instead of `model_copy()` was the right fix. So that part should stay dead; no need to keep reviewing ghosts.
- The thing that did **not** get smaller on a second read is the **conversation-context drop**. The new fork still throws away the parent conversation's safety and secret plumbing because it creates a fresh `Conversation(...)` from just `agent`, `working_dir`, and `persistence_dir`. In this repo, that means default `NeverConfirm`, no `security_analyzer`, and no carried-over secrets unless you wire them explicitly.
- The testing complaint is real, but only because it fails to catch that one underlying problem. I would keep the review focused on the root cause, not on a laundry list.

So the fresh-eyes version is simpler:

**Still blocking:**
- [`openhands-sdk/openhands/sdk/context/agent_context.py:286-310`, `openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py:692-700`, `openhands-sdk/openhands/sdk/skills/fork.py:78-113`] The fork now has clean agent/runtime separation, but it still does not preserve the parent conversation context that governs execution policy and secret injection.

**Updated risk take:**
- Still 🟡 **MEDIUM** overall. It is opt-in, so this is not random breakage. But for the new code path itself, dropping parent safety/secret state is a real correctness bug, not a nit.

_This follow-up review was created by an AI assistant (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/context/agent_context.py
```diff
@@ -277,15 +283,51 @@ def get_system_message_suffix(
...
+            return skill.content
+        if agent is None or working_dir is None:
+            logger.warning(
+                "Skill ‘%s’ has context=fork but agent/working_dir not "
+                "provided; falling back to inline injection",
+                skill.name,
+            )
+            return skill.content
+        logger.info("Skill ‘%s’ running as forked subagent", skill.name)
+        return run_skill_forked(skill, agent, working_dir, persistence_dir)
```
xingyaoww: I'm a bit worried that this might not be the best location to spin up another conversation and spin up an agent...

@OpenHands can you do an investigation and tell me your judgment here?

My main concern is that this happens under the hood of the AgentContext class, and this might not be what the class was designed for. Initially, this class was simply designed to persist and inject context to a particular conversation. Right now, we are trying to make it spin up a new conversation, which might not be ideal.

xingyaoww: **Investigation: Is `AgentContext` the right place to orchestrate forked skill execution?**

---

## TL;DR

The concern is valid. `AgentContext` was designed as a **pure context-assembly class** (data in → formatted text out), and embedding conversation orchestration inside it is a responsibility mismatch. The fork logic should live at the **conversation layer** (`LocalConversation`), which already owns the agent, workspace, persistence, and event lifecycle.

---

## Evidence

### 1. AgentContext's documented role is prompt assembly, not execution

From the class docstring:
> *"Central structure for managing prompt extension … the primary container responsible for assembling, formatting, and injecting all prompt-relevant context into LLM interactions."*

Every existing method on `AgentContext` is essentially a **pure function**: data in, text out, no side effects:
- `get_system_message_suffix()` → string
- `get_user_message_suffix()` → `TextContent` + skill names
- `get_secret_infos()` → metadata list
- `get_formatted_datetime()` → string

With this PR, `get_user_message_suffix()` can now **block for minutes** while an entire subagent conversation runs 50+ tool calls. That's a fundamentally different contract.

### 2. The new parameters are a code smell

The PR threads `agent`, `working_dir`, and `persistence_dir` into `get_user_message_suffix()` — none of which `AgentContext` naturally owns or needs for its designed purpose. These are conversation-layer concerns leaking into a context-assembly class:

```python
def get_user_message_suffix(
    self,
    user_message: Message,
    skip_skill_names: list[str],
    agent: AgentBase | None = None,        # NEW
    working_dir: str | None = None,        # NEW
    persistence_dir: str | None = None,    # NEW
) -> tuple[TextContent, list[str]] | None:
```

### 3. The existing DelegateTool pattern puts orchestration in the right layer

`DelegateTool` creates sub-conversations inside a `ToolExecutor` — an explicitly side-effectful component that lives in the conversation's action loop and receives `conversation` as a parameter. The conversation owns the lifecycle. This is the established pattern for "spin up a sub-conversation."

### 4. `AgentContext` is an immutable Pydantic model on `Agent`

Per the SDK design principles: *"All components — agents, tools, LLMs, configurations — are immutable Pydantic models validated at construction. The only mutable entity is the conversation state."*

`AgentContext` sits on `Agent.agent_context` — it's configuration. Having configuration objects spawn conversations is architecturally surprising.

---

## Suggested Alternative

Move the fork decision to **`LocalConversation.send_message()`** (or a helper it calls), since that's where all the needed context already lives:

```python
# In LocalConversation.send_message() — current call site (simplified)
if self.agent.agent_context:
    ctx = self.agent.agent_context.get_user_message_suffix(
        user_message=message,
        skip_skill_names=self._state.activated_knowledge_skills,
    )
```

**Option A — `AgentContext` returns metadata, conversation handles execution:**

`get_user_message_suffix()` could return which skills were triggered and whether they need forking (e.g., return the `Skill` objects or a flag). Then `LocalConversation` decides how to handle fork-context skills using `run_skill_forked()` directly.

**Option B — Extract fork orchestration into a conversation-level helper:**

Keep `AgentContext.get_user_message_suffix()` pure (returns skill content for inline skills, skips fork skills). Add a `_resolve_fork_skills()` method on `LocalConversation` that handles the forked execution, using the conversation's own `self.agent`, `self.workspace`, `self._state.persistence_dir`.

Either approach:
- Keeps `AgentContext` as a pure context-assembly class
- Keeps conversation orchestration in the conversation layer
- Eliminates the new `agent`/`working_dir`/`persistence_dir` parameters on `get_user_message_suffix()`
- Aligns with the DelegateTool precedent

---

## What the PR does well

To be fair, the core implementation in `fork.py` is solid — the subagent construction, recursion prevention, persistence layout, and error handling are well-thought-out. The issue is purely about **where the dispatch point lives**, not the fork mechanism itself.

enyst: I agree, these classes intended-as-immutable would help if they can stay that way 🤔

VascoSch92:
@openhands resolve the conflicts and update the branch

VascoSch92:
Hey @enyst @xingyaoww 

I Addressed the review feedbacks 

AgentContext was importing run_skill_forked and spinning up subagents inside get_user_message_suffix, which crossed its design boundary (persist + inject context for one conversation). Move fork dispatch out: AgentContext now takes an optional resolve_skill_content callback and no longer knows about agents, workspaces, or sub-conversations.

The dispatch closure lives in skills/fork.py as build_fork_resolver, next to run_skill_forked. LocalConversation builds it from its own agent/workspace/persistence_dir and passes it in. Remote conversations are unaffected, the agent server runs LocalConversation server-side, where the resolver lives.

openhands-sdk/openhands/sdk/skills/fork.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+# Skill names are not guaranteed to be filesystem-safe:
+# - Legacy skills derive their name from a relative path, which can contain
+#   "/" (e.g. "subdir/my_skill") and would create unintended nested dirs.
+# - Programmatic Skill(name=...) has no name validator; a crafted name like
+#   "../../etc/passwd" would escape the forks directory.
+# AgentSkills-format skills ARE validated (lowercase [a-z0-9-]) but we
+# sanitize unconditionally to keep the fork persistence layout predictable.
+_UNSAFE_PATH_CHARS = re.compile(r"[^a-zA-Z0-9_-]")
```
enyst, line 25: Are we doing this for other skills or their ability to run as commands or loading them from various sources? @openhands if not, wdyt, this code might give us some false sense of security ?

openhands-sdk/openhands/sdk/skills/skill.py
```diff
@@ -152,6 +152,16 @@ class Skill(BaseModel):
         default_factory=list,
         description="Input metadata for the skill (task skills only)",
     )
+    context: Literal["inline", "fork"] = Field(
+        default="inline",
+        description=(
+            "Execution context for a triggered skill. "
+            "'inline' (default) injects the skill content directly into the "
```
enyst, line 159: I wonder if this is still accurate, since invoke_skill?

enyst (commented):
Sorry for the delay here, thank you for the PR! I totally see how this could work, but just wondering... don't we get the same effect by subagents?

openhands-sdk/openhands/sdk/skills/fork.py
enyst: Good point — this should not be read as a broader skill security boundary. The only thing this regex protects is the fork persistence layout/key: legacy or programmatic names should not create nested directories or be interpreted as path components.

Skill loading trust and inline `!` command execution are separate concerns (and inline commands are still trusted-source only). I clarified the comment in 3327babe to make that scope explicit and avoid implying this secures skill sources or command execution.

_This reply was generated by an AI agent (OpenHands) on behalf of the user._

enyst: TBH I’d go for unnecessary and clean it up if so, no strong feelings though

openhands-sdk/openhands/sdk/skills/skill.py
VascoSch92: What do you mean?

VascoSch92:
> Sorry for the delay here, thank you for the PR! I totally see how this could work, but just wondering... don't we get the same effect by subagents?

@enyst yes this standardize a way to call sub-agents. You can decide tue query to given to the sub-agent and so on.

We cannot use the machinery of sub-agents because we are in the sdk and sub-agents are in tool :-) sdk cannot import from tool

enyst (commented):

🔴 **Needs improvement** - The fork is isolated from parent history, but it is also detached from the parent's safety boundary.

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/skills/fork.py`, Line 125] **Safety-policy bypass**: the fork constructs a fresh `Conversation` from only the copied agent, workspace path, and persistence path. Conversation-level controls are not inherited: the child starts with `NeverConfirm`, no `security_analyzer`, an empty secret registry, and no parent hook processor/config. A parent configured with `AlwaysConfirm` or `ConfirmRisky` can therefore trigger a `context: fork` skill on user input and have the child execute tools in the same workspace without the confirmations or security analysis that govern the parent. Passing only `AgentBase` into the resolver makes those controls impossible to preserve. Build the fork from the parent conversation (or an explicit execution-context object), inherit its confirmation/security/secret and applicable hook policy, and define how a child paused for confirmation is surfaced instead of immediately returning/closing. Add a regression test with an `AlwaysConfirm` parent proving the fork cannot execute a tool without approval.

**[RISK ASSESSMENT]**
This introduces automatic nested tool execution in the parent's workspace while dropping configured confirmation, security-analysis, hook, and secret context.

**Recommendation**: Do not auto-merge. Request review from a human security/architecture reviewer to validate the child-conversation trust and policy inheritance model.

**VERDICT:**
❌ **Needs rework**: Context isolation must not mean safety-policy isolation.

**KEY INSIGHT:**
A subagent sharing the parent's workspace must inherit the parent's execution controls even when it intentionally does not inherit chat history.

---

>
>
>

Issue #2587, opened 2026-03-26
Skills not consistently triggering on relevant prompt


neubig:
## Summary

Skills do not appear to trigger consistently even when a user prompt is clearly relevant.

A concrete example: when a user writes a prompt like **"make a pretty frontend that describes what open source is"**, the `frontend-design` skill from `OpenHands/extensions` is not triggering in OpenHands, even though it is highly relevant to the task.

Later in the same conversation, when queried, the agent did in fact have access to the skill. The problem seems to be that it did not recognize that the skill was relevant enough to use at the right time.

This contrasts with Claude Code, which *did* recognize that `frontend-design` was relevant in this kind of prompt and used it.

This is probably symptomatic of a broader issue: OpenHands may not be consistently recognizing relevant skills from the available skill set, even when the match should be obvious to a human.

## Why this matters

The current behavior undermines one of the main benefits of skills:
- relevant capabilities exist,
- they are successfully loaded into the conversation,
- but they are not being surfaced or invoked when the user prompt should cause them to matter.

That means skill quality and coverage can improve without users actually benefiting from them, because discovery/activation is unreliable.

## Requested investigation

We should investigate both:
1. whether the skill metadata / prompt representation is making relevance too subtle for the model to notice, and
2. whether OpenHands should have additional mechanisms for detecting and surfacing likely-relevant skills beyond the current behavior.

## Proposed next steps

1. Create a script using the OpenHands software-agent-sdk with **20 prompts** that should each result in skills from the `OpenHands/extensions` repo being called, then analyze the resulting trajectories to see whether the skills were actually accessed.
2. Run this script using:
   - the standard OpenHands agent, and
   - the Claude Code ACP agent.
3. Report the comparison between the two so we can quantify how often each agent recognized and used the relevant skills.

---
_This issue was created by an AI assistant (OpenHands) on behalf of the user._

VascoSch92:
i believe we can close this issue after #2835 was merged.

PR #2835, opened 2026-04-15
feat(sdk/tools): add explicit invoke_skill tool for progressive-disclosure skills 

### LINKED ISSUE #2824: [Feature] Make progressive-disclosure skill invocation explicit and observable
## Background

OpenHands supports two flavors of skills:

1. **Triggered skills** (keyword / task triggers) — when a trigger fires, the skill content is injected into the conversation. Activation is tracked in `ConversationState.activated_knowledge_skills` (`openhands-sdk/openhands/sdk/conversation/state.py:126`) and on each `MessageEvent.activated_skills` (`openhands-sdk/openhands/sdk/event/llm_convertible/message.py:41`). Fully observable.

2. **Progressive-disclosure skills** (AgentSkills format, no trigger) — only `<name>`, `<description>`, `<location>` are listed in the `<available_skills>` block of the system prompt (`openhands-sdk/openhands/sdk/skills/skill.py:1154`, `to_prompt`). The agent "invokes" one by *choosing on its own* to call the `Read` tool on the SKILL.md location.

## Problem

The progressive-disclosure has the following problems:

### 1. Weak invocation affordance
The agent has to (a) notice the description is relevant, (b) recall the skill exists across many turns, and (c) translate that into a `Read` call against an absolute path. There is no tool whose contract is "use a skill"; the model has to infer it. Issue #2587 documents the symptom: the `frontend-design` skill is loaded and visible but is not invoked on a clearly-relevant prompt like *"make a pretty frontend that describes what open source is"* — while Claude Code, which has an explicit invocation verb, does invoke it.

### 2. No telemetry
Because invocation is just `Read`, there is no skill-aware signal. We cannot distinguish "agent ignored a relevant skill" from "agent read the SKILL.md for unrelated reasons" without post-hoc string-matching tool-call paths against known skill locations. The eval proposed in #2587 (20 prompts × OpenHands vs Claude Code ACP) is hard to score cleanly today for exactly this reason.

### 3. Incompatibility with `context: fork`
PR #2823 added `context: fork` for triggered skills. For progressive-disclosure skills, `Read`-based invocation is the wrong affordance entirely: reading the file inlines the SKILL.md into the parent context, which defeats the isolation that `fork` is meant to provide. Part 2 of #2042 cannot be implemented on top of `Read`.

### 4. Dynamic context in skills is silently bypassed

PR #2582 introduced inline `!`command`` syntax so skill authors can embed dynamic shell output (git status, environment info, timestamps, etc.) into SKILL.md at render time via `render_content_with_commands` (`openhands-sdk/openhands/sdk/skills/execute.py`). This machinery only runs when the skill is *rendered* through the skill pipeline — e.g. trigger injection, or the `run_skill_forked` path.

When the agent "invokes" a progressive-disclosure skill by calling `Read` on `<location>`, it gets the **raw file contents** straight from disk: the `!`cmd`` tokens are returned literally, never executed. The author's dynamic context is silently dropped and the model sees placeholder syntax instead of the expected output.

This is a another concrete reason `Read` is the wrong affordance: it bypasses the skill rendering pipeline entirely. An `invoke_skill(name)` tool would route through the same renderer that triggered and forked skills already use, so dynamic context works uniformly across all three invocation modes.

## Prior art

- **Gemini CLI** ships a built-in [`activate_skill`](https://geminicli.com/docs/tools/activate-skill/) tool. The model autonomously calls it when a task matches a discovered skill's description; the skill's full instructions are then loaded.
- **Codex CLI** supports both implicit selection (description-matched, `allow_implicit_invocation=true`) and explicit `\$skill-name` user invocation.
- **Claude Code** combines explicit invocation with a subagent context boundary — the model PR #2823 already follows for the triggered case.

OpenHands is the outlier in relying solely on `Read` for non-triggered skills.

## Proposal

Introduce an explicit **`invoke_skill(name: str)`** built-in tool, auto-registered when any AgentSkills-format skill is loaded, and update `<available_skills>` rendering to advertise it as the invocation verb (instead of "Read this location"):

- For `context: inline` skills → returns the rendered skill content.
- For `context: fork` skills → dispatches to `run_skill_forked` (`openhands-sdk/openhands/sdk/skills/fork.py:37`) and returns only the subagent's final output.
- Emit a typed event so invocation is countable next to `activated_knowledge_skills`.

This single change addresses all three weaknesses: stronger affordance (a real tool the model sees in its schema, with a description that explicitly says "prefer this over `Read` for skills"), first-class telemetry (typed event, no path-matching), and the missing piece for Part 2 of #2042.

## Acceptance criteria

- [x] New `invoke_skill(name)` tool, auto-attached when any AgentSkills-format skill is loaded.
- [x] `<available_skills>` block advertises `invoke_skill(name="...")` instead of "Read the location".
- [x] Inline path returns rendered skill content; fork path goes through `run_skill_forked`.
- [x] Invocations recorded on `ConversationState` (mirroring `activated_knowledge_skills`) and surfaced via a hook event.
- [x] Eval harness from #2587 (20 prompts) re-run on the same model with and without `invoke_skill` to quantify the recognition uplift.

## Related

- #2042 — Run skills via subagent delegation (Part 2 depends on this)
- #2587 — Skills not consistently triggering on relevant prompt
- #2823 — `context: fork` for triggered skills (Part 1)

VascoSch92:

## Summary

Closes #2824. Introduces an explicit invoke_skill(name) built-in tool as the single, telemetered entry point for progressive-disclosure skills. Before this change, the agent "invoked" a skill by inferring it should Read the <location> path in <available_skills> — a weak affordance with no skill-aware signal, and one that bypassed the skill-rendering pipeline (so inline !\cmd`` tokens in SKILL.md were never executed).

## Issue Number

#2824

## What changes    

- New built-in tool invoke_skill (openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py). Resolves a Skill by name from agent_context.skills, renders it through render_content_with_commands so dynamic tokens execute, records the invocation on ConversationState.invoked_skills, and returns the rendered content as the tool observation. Declares a per-skill resource key (skill:<name>) so distinct skills can render in parallel while repeats of the same skill serialize.                                                                                                                                                                                                                       
- Auto-attach in Agent._initialize: the tool is appended iff the agent has at least one AgentSkills-format skill loaded. Intentionally not in BUILT_IN_TOOLS, so serialized include_default_tools never advertises a tool the agent won't attach.                                                                                                                                                                                                                          
- Prompt change in <available_skills>: drops <location> entirely and tells the model that invoke_skill(name="...") is the only supported invocation path. Truncation messages now point at invoke_skill(name="...") instead of a file path.                                                                                                                                                                                                                                  
- State: new ConversationState.invoked_skills: list[str], parallel to activated_knowledge_skills. Combined with ActionEvent(tool_name="invoke_skill"), this gives first-class telemetry without path-matching.
- Benchmark at .pr/ (5 skills × 4 prompts = 20 prompts, matches the eval from #2587). Runs trials in parallel; recognition is scored by detecting invoke_skill ActionEvents. Same script will measure uplift once context: fork (#2823) lands.       

## Benchmark results

The benchmark is the one present in `.pr` folder.
I ran it against main and against this branch.

Withouth the invoke_skill tool

```text
=== summary ===
  frontend-design     recognition= 0.0%  (4 prompts)
  pdf-analyst         recognition=25.0%  (4 prompts)
  git-archeologist    recognition=25.0%  (4 prompts)
  sql-optimizer       recognition= 0.0%  (4 prompts)
  data-viz            recognition=50.0%  (4 prompts)

  OVERALL            recognition=20.0%  (20 prompts x 1 trials)
```

with the invoke_skill tool

```text
=== summary ===
  frontend-design     recognition=100.0%  (4 prompts)
  pdf-analyst         recognition=100.0%  (4 prompts)
  git-archeologist    recognition=75.0%  (4 prompts)
  sql-optimizer       recognition=100.0%  (4 prompts)
  data-viz            recognition=100.0%  (4 prompts)

  OVERALL            recognition=95.0%  (20 prompts x 1 trials, 0 errors)

```

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore










---

**Variants & Base Images**
|---|---|---|---|


```bash
docker pull ghcr.io/openhands/agent-server:13ac9de-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-13ac9de-python \
  ghcr.io/openhands/agent-server:13ac9de-python
```

```
ghcr.io/openhands/agent-server:13ac9de-golang-amd64
ghcr.io/openhands/agent-server:13ac9de-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:13ac9de-golang-arm64
ghcr.io/openhands/agent-server:13ac9de-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:13ac9de-java-amd64
ghcr.io/openhands/agent-server:13ac9de-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:13ac9de-java-arm64
ghcr.io/openhands/agent-server:13ac9de-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:13ac9de-python-amd64
ghcr.io/openhands/agent-server:13ac9de-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:13ac9de-python-arm64
ghcr.io/openhands/agent-server:13ac9de-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:13ac9de-golang
ghcr.io/openhands/agent-server:13ac9de-java
ghcr.io/openhands/agent-server:13ac9de-python
```

- Each variant tag (e.g., `13ac9de-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `13ac9de-python-amd64`) are also available if needed


VascoSch92:
**Evaluations**

Baseline

source: https://results.eval.all-hands.dev/swebench/litellm_proxy-claude-sonnet-4-5-20250929/24449034454/results.tar.gz

:bar_chart: Results
- Total instances: 500
- Submitted instances: 50
- Resolved instances: 37
- Unresolved instances: 13
- Empty patch instances: 0
- Error instances: 0
- Eval limit: 50
- Success rate: 37/50 (74.0%)

This Branch

source: https://results.eval.all-hands.dev/swebench/litellm_proxy-claude-sonnet-4-5-20250929/24445826041/results.tar.gz

:bar_chart: Results
- Total instances: 500
- Submitted instances: 50
- Resolved instances: 39
- Unresolved instances: 11
- Empty patch instances: 0
- Error instances: 0
- Eval limit: 50
- Success rate: 39/50 (78.0%)


Moreover, I asked OH to explore the trajectories and there are no evidence that the PR is causing errors.

.pr/benchmark_invoke_skill.py
```diff
@@ -0,0 +1,225 @@
+"""Benchmark: progressive-disclosure skill recognition rate.
+
+Measures how often an agent invokes a clearly-relevant progressive-disclosure
+skill when only `<available_skills>` metadata (name / description / location)
+is visible in the system prompt.
+
+A prompt counts as a HIT when, during the conversation, the agent either:
+  - calls `Read` on the expected SKILL.md location (current affordance), or
+  - calls `invoke_skill(name=...)` with the expected skill name
```
enyst: I wonder if it will use the tool more often than it did `read`? Sorry I didn't yet read the comments

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
```diff
@@ -0,0 +1,145 @@
...
+class InvokeSkillExecutor(ToolExecutor):
+    @staticmethod
+    def _get_skills_and_working_dir(
+        conversation: BaseConversation | None,
+    ) -> tuple[list, Path | None]:
+        if conversation is None:
+            return [], None
+
+        agent = getattr(conversation, "agent", None)
+        ctx = getattr(agent, "agent_context", None)
```
enyst: Tiny nit: are `getattr` necessary? Ditto below

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
```diff
@@ -0,0 +1,145 @@
...
+        t.append(self.text)
+        return t
+
+
+TOOL_DESCRIPTION = """Invoke a skill by name.
+
+This is the only supported way to invoke a skill listed in
+`<available_skills>`. Call it with the `<name>` shown in that block; the
+skill's full content is rendered (including any dynamic context) and
+returned as the tool result.
```
enyst, line 55: 🤔 If I understand correctly, this means large skills, with references, scripts, sub-agents (?) will be either in or out of LLM's context?

enyst: @openhands Read agentskills.io and remind me what are all the types of files or content that can go in an Agentskill by the standard? Respond directly in-thread reply.

openhands-sdk/openhands/sdk/skills/skill.py
```diff
@@ -1142,7 +1142,9 @@ def to_prompt(skills: list[Skill], max_description_length: int = 1024) -> str:
         max_description_length: Maximum length for descriptions (default 1024)
 
     Returns:
-        XML string in AgentSkills format with name, description, and location
+        XML string in AgentSkills format with name and description. The
+        `<location>` field is intentionally omitted so the agent cannot
+        bypass the `invoke_skill` tool by reading the file directly.
```
enyst, line 1147: TBH my reflex about this was:
- load the SKILL.md into context when the invoke tool is called
- tell the LLM the location, so it can read at its will the other things that SKILL.md may reference...

Maybe I'm wrong though 🤔

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
enyst: I checked the spec at https://agentskills.io/specification and the client guide at https://agentskills.io/client-implementation/adding-skills-support.

By the standard, the only **required** file is `SKILL.md` at the skill root. Everything else is optional. The spec explicitly calls out these supporting content types:

- `scripts/` — executable code
- `references/` — extra docs read on demand (for example `REFERENCE.md`, `FORMS.md`, or domain-specific docs)
- `assets/` — static resources like templates, images, lookup tables, schemas, data files
- `...` — **any additional files or directories**; the spec does not make this a closed list

It also defines the progressive-disclosure layers pretty clearly:
1. `name` + `description` are loaded into the catalog up front
2. the full `SKILL.md` body is loaded when the skill is activated
3. supporting files are loaded only if/when the instructions reference them

So for the earlier question in this thread: large references/scripts/assets are *not* automatically in the LLM context. They are meant to stay out of context until explicitly pulled in. And “subagent delegation” is described as an optional **client execution pattern**, not as a special file type in the Agent Skills spec.

.pr/benchmark_invoke_skill.py
VascoSch92: It seems to use the tool correctly, surfacing skills only when appropriate (see the benchmark above). 

However, if there are no relevant skills, it doesn't abuse the tool. In the SWE-bench eval, the tool was presented along with a set of skills, but since they weren't relevant to the bug fix, the LLM never invoked it.

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
VascoSch92: @enyst, to answer your question: you can always reference other docs within skill.md. Just provide the path, and the LLM can read them (provided the agent has a read tool).

However, if the agent uses the read tool to read skill.md itself, that could be problematic—especially if we are injecting content using the logic introduced in this PR: https://github.com/OpenHands/software-agent-sdk/pull/2582

enyst (commented):
Thank you for trying this, I do think it's worth it! 

To note, I may be a bit confused on one point:
- render with commands seems to refer only to `SKILL.md` file with your `!` implementation
- https://agentskills.io/specification#directory-structure specifies additional directories

If we don't tell the LLM the location, how will it know about those directories?

openhands-sdk/openhands/sdk/skills/skill.py
VascoSch92: Perhaps I see the confusion here. 

We should not give the location to the SKILL.md but in the SKILL.md you can reference other files that the LLM can read.

Do you think that I should change the comment? Make it more clear?

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
enyst: Are other skills out there providing full path when they reference stuff inside the skill? 🤔

VascoSch92:
> Thank you for trying this, I do think it's worth it!
> 
> To note, I may be a bit confused on one point:
> 
>     * render with commands seems to refer only to `SKILL.md` file with your `!` implementation
> 
>     * https://agentskills.io/specification#directory-structure specifies additional directories
> 
> 
> If we don't tell the LLM the location, how will it know about those directories?

Ah now I understand. 

Good catch. 

An idea to solve this is to rewrite relative references in `SKILL.md` to absolute paths (anchored at the skill root) so the LLM can actually Read / Bash them. 

Or we just expose the absolute directory path such that the LLm now were they are and the LLm can then use a tool to pull in scripts/ references/ and so on. (What Antrophic is doing actually).

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
VascoSch92: answered [here](https://github.com/OpenHands/software-agent-sdk/pull/2835#issuecomment-4251811143) :-)

enyst:
Or maybe like, insert a line at the end of the SKILL.md automatically, when it's invoked (so with the tool exec), to tell the LLM e.g.: this skill is in (regular path) directory. If it references other files, they are in this directory.

(or maybe that's what CC is doing?)

openhands-sdk/openhands/sdk/skills/skill.py
enyst: So the concern is, the LLM may get confused if it needs to explore, idk, all its workspace in order to discover the relative references? Wait, not only workspace... maybe user home too, some skills are coming from user home, or from cache maybe 🤔 

It may be worth it to try in `.pr/` ? (everything is fine with the existing tests, we can remove/change them imho now, it's just that they assume only SKILL.md exists)

VascoSch92:
> Or maybe like, insert a line at the end of the SKILL.md automatically, when it's invoked (so with the tool exec), to tell the LLM e.g.: this skill is in (regular path) directory. If it references other files, they are in this directory.
> 
> (or maybe that's what CC is doing?)

CC is doing the following:

```text
The two-stage design                                                                                                                                                                                               
                                                                                                                                                                                                                     
  1. Listing stage (in <available_skills>, via prompt.ts:65-168): only - name: description. No path.                                                                                                                 
  2. Invocation stage (SkillTool return): the rendered content is prefixed with Base directory for this skill: /abs/path/to/skill, and any ${CLAUDE_SKILL_DIR} / ${CLAUDE_SESSION_ID} tokens inside SKILL.md are
  string-substituted.                                                                                                                                                                                                
                                                                  
How that lets SKILL.md reference sibling files                                                                                                                                                                     
                                                                  
Inside a SKILL.md you can write any of these patterns:                                                                                                                                                             
                                                                  
- Relative refs (./schemas/foo.json, ./scripts/run.sh) — the model resolves them against the "Base directory" header using Read/Bash.                                                                              
- Templated absolute refs (${CLAUDE_SKILL_DIR}/schemas/foo.json) — the token is replaced with the real absolute path before the model sees it, so it can Read them directly with no relative-path ambiguity.
- Session-scoped paths (${CLAUDE_SESSION_ID}) — useful for per-session scratch dirs.                                                                                                                               
                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                     
Why this split matters                                                                                                                                                                                             
                                                                  
  - Prompt cache stability: the listing (static across sessions) has no paths → stays globally cacheable. Paths are per-host and would poison the cache.                                                             
  - Security / surface area: the model only sees one skill's directory at a time, and only after it decided to invoke that skill.
  - Post-compact survival: line 1088 registers finalContent (already transformed) with the compaction state, so the base-directory header and substitutions survive conversation compression. Comment on 1085: "must 
  use finalContent (not raw content) so the base directory header and ${CLAUDE_SKILL_DIR} substitutions survive compaction."
```

So i think that your idea to disclose the skill path at the end of the skill is good. It makes sense and then the LLM can solve the relative path.

openhands-sdk/openhands/sdk/skills/skill.py
VascoSch92: Yes,

I wanted to try the solution discussed in the comments and update the benchmark to see if the LLM can do that.

enyst:
Aha! Perfect, the base directory sounds good to me.

On a side note, this is interesting!
> Post-compact survival

openhands-sdk/openhands/sdk/skills/skill.py
enyst: SGTM, thank you!

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
VascoSch92: No... I can change them ;-) just need to add a method in the protocol

VascoSch92:
Hey @enyst 

I did the changes, i.e., footer with the path of the skill.

Moreover, 

I added a test case in the .pr that show that the LLM understand that.

I also re-run the benchmark to see if everything was still good.

New results:

```text
=== summary ===
  frontend-design     recognition=100.0%  (4 prompts)
  pdf-analyst         recognition=100.0%  (4 prompts)
  git-archeologist    recognition=100.0%  (4 prompts)
  sql-optimizer       recognition=100.0%  (4 prompts)
  data-viz            recognition=75.0%  (4 prompts)

  OVERALL            recognition=95.0%  (20 prompts x 1 trials, 0 errors)
```

xingyaoww (commented):
Overall LGTM! Shall we add an llm integration test and explicitly let all LLM to call this tool, so we can track in the future that this tool doesn't regress?

OTOH, i'm a bit worried that this might break the remote agent-server, can we also maybe write a test / modify the agent server example to run this?

Maybe try add a test to the https://github.com/OpenHands/software-agent-sdk/blob/main/tests/cross/test_remote_conversation_live_server.py

VascoSch92:
@xingyaoww 

1. added test for the agent server. Actually good-catch there it was a small bug but it is fixed (missing declaration of active_skills in models.py)
2. added integration test

Should we run them?

@enyst because for point 1 I'm now breaking the APIs

tests/cross/test_remote_conversation_live_server.py
```diff
@@ -1493,3 +1494,127 @@ def fake_completion_with_finish(
...
+            model_name="test-model",
+            accumulated_cost=0.0,
+            max_budget_per_task=None,
+            accumulated_token_usage=None,
+        )
+        return LLMResponse(
+            message=message, metrics=metrics_snapshot, raw_response=raw_response
+        )
+
+    monkeypatch.setattr(LLM, "completion", fake_completion, raising=True)
```
enyst, line 1575: Wondering if we can use your TestLLM 😸

VascoSch92: Thought about it. The catch is that this test drives a remote conversation: the `Agent` (and its `llm`) is serialized and shipped to the agent-server. `TestLLM._scripted_responses` lives in a `PrivateAttr`, which pydantic drops on serialization, so the server would rehydrate a `TestLLM` with an empty queue and raise `TestLLMExhaustedError` on the first turn. The class-level `LLM.completion` monkeypatch is the only lever that reaches the server-side LLM here.

Happy to follow up by moving scripted responses onto a real pydantic field so `TestLLM` works across the wire. But this is a separate PR :-)

enyst:
> @enyst because for point 1 I'm now breaking the APIs

Sorry, did they break? I don't see it here, but could it be that you saw some failures yesterday, which were actually due to a quick breakage in `main`?

VascoSch92:
> > @enyst because for point 1 I'm now breaking the APIs
> 
> Sorry, did they break? I don't see it here, but could it be that you saw some failures yesterday, which were actually due to a quick breakage in `main`?

Ah ok... yes I saw them just yesterday but not today when I re-run the tests

xingyaoww (approved):
Thank yoi!

PR #2845, opened 2026-04-16
chore(ci): rename secrets.PAT_TOKEN to secrets.ALLHANDS_BOT_GITHUB_PAT

simonrosenberg:
## Summary
- Rename `secrets.PAT_TOKEN` → `secrets.ALLHANDS_BOT_GITHUB_PAT` across 9 workflow files (11 references)
- Standardizes on `ALLHANDS_BOT_GITHUB_PAT` as the org-wide secret name for the bot's GitHub PAT
- Internal env var names passed to scripts are unchanged — only the GitHub Actions secret reference is renamed
- Part of OpenHands/evaluation#428

## Affected workflows
`assign-reviews.yml`, `deploy-docs.yml`, `pr-artifacts.yml`, `pr-review-by-openhands.yml`, `prepare-release.yml`, `qa-changes-by-openhands.yml`, `run-eval.yml`, `todo-management.yml`, `version-bump-prs.yml`

## Prerequisite
`ALLHANDS_BOT_GITHUB_PAT` must be configured as an org/repo secret visible to this repo before merging.

## Test plan
- [ ] Verify `ALLHANDS_BOT_GITHUB_PAT` secret is available to this repo
- [ ] After merge, trigger one workflow from each category and verify it succeeds

### DIFF
--- a/.github/workflows/assign-reviews.yml
+++ b/.github/workflows/assign-reviews.yml
@@ -190,7 +190,7 @@ jobs:
             - name: Run task
               env:
                   LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
-                  GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
+                  GITHUB_TOKEN: ${{ secrets.ALLHANDS_BOT_GITHUB_PAT }}
                   PYTHONPATH: ''
               run: |
                   echo "Running agent script: $AGENT_SCRIPT_URL"
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -17,7 +17,7 @@ jobs:
             - name: Trigger docs repo sync
               uses: peter-evans/repository-dispatch@v4
               with:
-                  token: ${{ secrets.PAT_TOKEN }}
+                  token: ${{ secrets.ALLHANDS_BOT_GITHUB_PAT }}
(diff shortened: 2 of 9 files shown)


enyst (changes requested):
Just a nit, please: we have tried since 7+ months ago, to use the name openhands for openhands stuff as much as possible, rather than all-hands. Is there a reason why we wouldn't do so here?

simonrosenberg:
> Just a nit, please: we have tried since 7+ months ago, to use the name openhands for openhands stuff as much as possible, rather than all-hands. Is there a reason why we wouldn't do so here?

@xingyaoww lets rename all to. OPENHANDS_BOT_GITHUB_PAT then I guess?

enyst:
I'm working to propose other ways too to fix permissions, setups, so that we can work better and easier with a mixed github organizational account, that has both AH-commercial / proprietary / specific to closed stuffs, and the OpenHands Open Source project.

It's not always obvious, but names matter. They help humans _**and agents**_ guess, grasp meaning, find solutions. Hey we have this issue in the first place precisely because we need to separate what is closed / commercial / infra stuff from what is OpenHands stuff... basically. ish. 😅

(I was even thinking, if there's a need for secrets that shouldn't be used by agents in public avenues, then maybe we could name them like NOT_FOR_AGENT ... OK that's dumb! 😅 )

PR #2868, opened 2026-04-16
feat(acp): stream ACPToolCallEvents live from session_update

### LINKED ISSUE #2866: ACPAgent: stream tool-call events from session_update instead of batching at turn end
## Summary

`ACPAgent.step()` currently accumulates every `ToolCallStart` / `ToolCallProgress` / `AgentMessageChunk` / `AgentThoughtChunk` the ACP subprocess sends via `_OpenHandsACPBridge.session_update`, and only emits them as OpenHands events (`ACPToolCallEvent`, `MessageEvent`) **after** `conn.prompt(...)` returns. That produces a poor live-UX — tool cards appear in a single burst next to the final assistant message, no matter how long the underlying agent ran.

## Current behavior

- `_OpenHandsACPBridge.session_update` appends to in-memory lists (`accumulated_text`, `accumulated_thoughts`, `accumulated_tool_calls`).
- `ACPAgent.step()` iterates those lists after `prompt()` completes and fires `on_event(...)` once per item.

See [`openhands-sdk/openhands/sdk/agent/acp_agent.py`](https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/agent/acp_agent.py) — the relevant sections are `_OpenHandsACPBridge.session_update` (~line 344) and `ACPAgent.step()` (~line 887).

## Proposal

Emit events live from `session_update`:

- On `ToolCallStart` → emit a preliminary `ACPToolCallEvent` with status `in_progress` / no output yet.
- On `ToolCallProgress` → emit an updated `ACPToolCallEvent` merging new fields; consumers that care about streaming updates can dedupe by `tool_call_id`, consumers that want the final state can just wait.
- On `AgentMessageChunk` / `AgentThoughtChunk` → consider streaming the text (matches how the current `on_token` callback already works, but via a first-class event).

The `ACPAgent.step()` body can then just wait for `prompt()` to return for bookkeeping (usage, elapsed, finish) without re-emitting the intermediate events.

### Caveats

- `on_event` is a sync callback; `session_update` runs on the ACP SDK's asyncio loop. Need to marshal events across the loop boundary safely (probably via the same executor ACPAgent already uses).
- Downstream consumers (OpenHands frontend, eval trajectories, webhooks) need to handle the same `ACPToolCallEvent` id appearing multiple times with evolving state. Easiest: keep `tool_call_id` stable, set `status` + fill `raw_output`/`content` on progress.
- Thought chunks may arrive interleaved with tool calls; keep them in the same event stream so ordering is preserved.

## Motivation

Companion frontend issue in OpenHands: **OpenHands/OpenHands#TBD — Render `ACPToolCallEvent` in the conversation viewer**. Together they close the UX gap with Claude Code's own web UI.

## References

- Discriminated-union PR that ships ACP end-to-end in OpenHands: #2861
- Tested against `@agentclientprotocol/claude-agent-acp` v0.29.0; confirmed the `session_update` cadence is one-per-tool-call as expected.

simonrosenberg:
## Summary
- `_OpenHandsACPBridge.session_update` now fires `on_event` synchronously on each `ToolCallStart` / `ToolCallProgress`, emitting an `ACPToolCallEvent` with evolving `status` + `raw_output` / `content` under the same `tool_call_id`. Consumers dedupe by id and treat the last-seen state as authoritative.
- `ACPAgent.step()` drops the end-of-turn fan-out loop; it now just wires `on_event` onto the bridge and handles bookkeeping (usage totals, `FinishAction`, execution-status transition) once `prompt()` returns.
- `on_event` rides the same `AsyncExecutor` boundary that already shuttles `on_token` / `on_activity` across the portal thread, so no new thread-safety surface is introduced.
- Final `MessageEvent` / `FinishAction` shape is unchanged — only the ACP tool-call stream becomes live.

Closes #2866. Companion frontend work: OpenHands/OpenHands#13991 (out of scope here).

## Why this does not need eval validation
This PR changes **when** `on_event` is called, not **what** the agent does or decides. Specifically:

- The ACP subprocess (claude-agent-acp / codex-acp / gemini-cli) is the thing that picks tools, executes them, and produces the final answer. That subprocess is untouched.
- The accumulated text, thoughts, tool-call fields, and token-usage math all go through the same code paths — `ACPToolCallEvent` / `MessageEvent` / `FinishAction` / `ObservationEvent` are still constructed with the same field values and in the same relative order per turn.
- The only observable delta is that the N `ACPToolCallEvent`s that used to be emitted in a sub-millisecond burst right before the `MessageEvent` are now emitted synchronously as each ACP notification arrives.

Evals measure task-completion quality (correctness, tool-use strategy, cost), which is a function of the ACP subprocess's decisions. Those decisions do not depend on `on_event` timing: the bridge's live emission is fire-and-forget into the OpenHands event stream, and the ACP `session_update` handler never blocks on or observes anything the event sink does. Running SWE-bench or Commit0 against this change would measure noise (LLM sampling variance on the subprocess side) at a meaningful compute cost, with nothing to attribute a regression or win to. The behavior *this* PR is responsible for is exactly what unit tests and a timing probe can verify deterministically; evals would not.

### Unit tests (deterministic)
- `uv run pytest tests/sdk/agent/test_acp_agent.py` — 125 passed. Includes new `TestACPToolCallLiveEmission` cases covering:
  - start → progress sequence firing `on_event` live with evolving status
  - interleaved tool-call / text-chunk / thought-chunk arrivals preserving arrival order across the combined `on_event` + `on_token` stream
  - `on_event=None` staying a safe no-op; a raising callback not breaking `session_update`
  - `reset()` clearing `on_event`
- `uv run pytest tests/sdk/agent/` — full agent suite green (365 tests, no regressions).
- `uv run pre-commit run --files …` — ruff format / ruff lint / pycodestyle / pyright / import-dependency / tool-registration all pass.

### Live end-to-end against real `@agentclientprotocol/claude-agent-acp@0.29.0`
Two runs against Anthropic's Claude via `claude-agent-acp`, with a callback that records the wall-clock offset at which every event hit `on_event`.

**Run 1 — focused verification script.** Seeded two `.py` files in `/tmp/…/workspace`, prompted *"List the Python files in the cwd, read both of them, and summarize each function"*.

```
[ 0.081s] MessageEvent      source=user  text[:60]='List the Python files…'
[11.765s] ACPToolCallEvent  id=toolu_0169jG status=pending    title='Find'
[11.936s] ACPToolCallEvent  id=toolu_0169jG status=pending    title='Find `*.py`'
[11.954s] ACPToolCallEvent  id=toolu_0169jG status=pending    title='Find `*.py`'
[12.073s] ACPToolCallEvent  id=toolu_0169jG status=completed  title='Find `*.py`'
[15.389s] ACPToolCallEvent  id=toolu_0139f7 status=pending    title='Read File'
[15.783s] ACPToolCallEvent  id=toolu_0139f7 status=pending    title='Read .../alpha.py'
[15.786s] ACPToolCallEvent  id=toolu_01RZBz status=pending    title='Read File'
[15.791s] ACPToolCallEvent  id=toolu_0139f7 status=pending    title='Read .../alpha.py'
[15.993s] ACPToolCallEvent  id=toolu_0139f7 status=completed  title='Read .../alpha.py'
[16.002s] ACPToolCallEvent  id=toolu_01RZBz status=pending    title='Read .../beta.py'
[16.004s] ACPToolCallEvent  id=toolu_01RZBz status=pending    title='Read .../beta.py'
[16.017s] ACPToolCallEvent  id=toolu_01RZBz status=completed  title='Read .../beta.py'
[18.523s] MessageEvent      source=agent text[:60]='Found two Python files: …'
```

| Moment | Wall clock |
|---|---|
| First `ACPToolCallEvent` (Glob start) | 11.765 s |
| Last `ACPToolCallEvent` (Read beta.py completed) | 16.017 s |
| Final `MessageEvent` | 18.523 s |
| Tool-call spread | **4.253 s** |
| Last tool-call → message gap | **2.506 s** |

Under the old batched-at-end behavior all 13 of those `ACPToolCallEvent`s would have arrived in the same sub-millisecond tick right before the `MessageEvent`. Here they span 4.25 s of real time, and the last tool-call lands 2.5 s before the final message.

**Run 2 — the shipped example `examples/01_standalone_sdk/40_acp_agent_example.py`.** Main turn (*"list python files under openhands-sdk/openhands/sdk/agent/, read `__init__.py`, summarize exports"*) + `ask_agent` side-question. Main turn timeline:

```
[ 0.010s] SystemPromptEvent
[ 7.680s] ACPToolCallEvent  id=toolu_01F7z5 status=pending    title='Find'
[ 8.041s] ACPToolCallEvent  id=toolu_01F7z5 status=pending    title='Find `…/agent/**/*.py`'
[ 8.378s] ACPToolCallEvent  id=toolu_01F7z5 status=pending    title='Find `…/agent/**/*.py`'
[ 8.380s] ACPToolCallEvent  id=toolu_01F7z5 status=completed  title='Find `…/agent/**/*.py`'
[10.723s] ACPToolCallEvent  id=toolu_01DWz9 status=pending    title='Read File'
[11.519s] ACPToolCallEvent  id=toolu_01DWz9 status=pending    title='Read …/__init__.py'
[11.523s] ACPToolCallEvent  id=toolu_01DWz9 status=pending    title='Read …/__init__.py'
[11.553s] ACPToolCallEvent  id=toolu_01DWz9 status=completed  title='Read …/__init__.py'
[16.514s] MessageEvent      source=agent text[:60]='Python files under `openhands-sdk/openhands/sdk/agent/`:\n- `'
```

Tool-call spread **3.873 s**, last tool-call → message gap **4.961 s**; 8/8 tool-call events arrive before the first agent message. `ask_agent` fork side-question also runs to completion.

Together, the unit tests pin the live-emission contract (ordering, id stability, error-swallow, callback lifecycle) and the live runs prove the contract holds against a real ACP subprocess streaming real LLM tool calls.

### SWE-Bench ACP Evaluation (2026-04-17)
**Branch validated against live SWE-Bench benchmark with ACP Claude agent.**

- **Correlation ID**: `cid:BADAD8A8`
- **GitHub Actions Run**: [#24588353534](https://github.com/OpenHands/software-agent-sdk/actions/runs/24588353534) (SDK workflow)
- **Evaluation Workflow**: [#24588371825](https://github.com/OpenHands/evaluation/actions/runs/24588371825) (Evaluation workflow)
- **K8s Job**: `eval-24588371825-claude-son-infer` (evaluation-jobs namespace)
- **Benchmark**: SWE-Bench Verified (50 instances)
- **Model**: claude-sonnet-4-5-20250929 via ACP (`acp-claude`)
- **Runtime**: 30m 17s total (28m 17s evaluating 50 instances + overhead)
- **Completion**: ✅ **96-100%** (48-50/50 instances completed)

**Results**:
- ✅ **Zero ACP agent crashes** — no "Stopping idle runtime" or crash events
- ✅ **Zero timeouts** — instances completing in 10-50s range consistently
- ✅ **Zero keepalive issues** — workspace_keepalive mechanism working (no idle kills)
- ✅ **Transient errors handled gracefully** — 5 connection disconnects and 112 health checks all retried successfully
- ✅ **Tool-call event streaming working** — no errors from live event callbacks, no observable side effects on agent behavior
- ✅ **No regressions from invoke_skill removal** — skill location fields exposed in prompts, agents operating normally

**Logs**: `kubectl logs job/eval-24588371825-claude-son-infer -n evaluation-jobs --tail=100` shows steady progress with no ACP-layer errors.

The live event streaming is transparent to the benchmark — it doesn't change agent task completion behavior, but the evaluation validates that the code changes (event bridge refactoring, invoke_skill removal, skill location exposure) don't introduce instabilities under sustained load.

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/acp_agent.py
+++ b/openhands-sdk/openhands/sdk/agent/acp_agent.py
@@ -109,6 +109,12 @@ _STREAM_READER_LIMIT: int = 100 * 1024 * 1024  # 100 MiB
 # well below the ~20 min runtime-api kill threshold.
 _ACTIVITY_SIGNAL_INTERVAL: float = 30.0
 
+# ACP tool-call statuses that represent a terminal outcome.  Non-terminal
+# statuses (``pending``, ``in_progress``) mean the call is still in flight
+# and, if the turn aborts before it reaches a terminal state, the live-
+# emitted event on state.events will otherwise be orphaned forever.
+_TERMINAL_TOOL_CALL_STATUSES: frozenset[str] = frozenset({"completed", "failed"})
+
 
 def _make_dummy_llm() -> LLM:
     """Create a dummy LLM that should never be called directly."""
@@ -286,6 +292,16 @@ class _OpenHandsACPBridge:
     """Bridge between OpenHands and ACP that accumulates session updates.
 
     Implements the ``Client`` protocol from ``agent_client_protocol``.
+
+    Concurrency model — ``on_event`` / ``on_token`` / ``on_activity`` are
+    fired synchronously from ``session_update``, which runs on the
+    ``AsyncExecutor`` portal thread.  The caller thread driving
+    ``ACPAgent.step()`` is blocked inside ``portal.call()`` for the entire
+    ``prompt()`` round-trip, so these callbacks do not race with the final
+    ``MessageEvent`` / ``FinishAction`` emitted by the caller thread after
+    ``prompt()`` returns.  Consumers that keep cross-callback state (e.g.
+    hook processors reading-then-writing, visualizers) can therefore treat
+    each callback as sequential within a single turn.
     """
 
     def __init__(self) -> None:
@@ -939,6 +955,45 @@ class ACPAgent(AgentBase):
         self._client.on_event = on_event
         self._client.on_activity = self._on_activity
 
+    def _cancel_inflight_tool_calls(self, on_event: ConversationCallbackType) -> None:
+        """Emit a terminal ``failed`` ACPToolCallEvent for every tool call
+        in the accumulator that has not reached a terminal status yet.
+
+        ACP servers mint fresh ``tool_call_id``s on a retried turn, so any
+        ``pending`` / ``in_progress`` events already streamed during the
+        failed attempt would otherwise be orphaned on ``state.events`` —
+        no later notification reuses their id, and consumers that dedupe
+        by ``tool_call_id`` + "last-seen status wins" would keep them
+        spinning forever.  This method closes those cards before we wipe
+        the in-memory accumulator on retry / turn abort.
+
+        Called with ``on_event`` passed in explicitly because the bridge's
+        ``on_event`` attribute is about to be cleared by ``reset()``.
+        """
+        for tc in self._client.accumulated_tool_calls:
+            status = tc.get("status")
+            if status in _TERMINAL_TOOL_CALL_STATUSES:
+                continue
+            try:
+                on_event(
+                    ACPToolCallEvent(
+                        tool_call_id=tc["tool_call_id"],
+                        title=tc["title"],
+                        status="failed",
+                        tool_kind=tc.get("tool_kind"),
+                        raw_input=tc.get("raw_input"),
+                        raw_output=tc.get("raw_output"),
+                        content=tc.get("content"),
+                        is_error=True,
+                    )
+                )
+            except Exception:
+                logger.debug(
+                    "Failed to emit supersede event for %s",
+                    tc.get("tool_call_id"),
+                    exc_info=True,
+                )
+
     @observe(name="acp_agent.step", ignore_inputs=["conversation", "on_event"])
     def step(
         self,
@@ -1024,6 +1079,7 @@ class ACPAgent(AgentBase):
                             e,
                         )
                         time.sleep(delay)
+                        self._cancel_inflight_tool_calls(on_event)
                         self._reset_client_for_turn(on_token, on_event)
                     else:
                         raise
@@ -1048,6 +1104,7 @@ class ACPAgent(AgentBase):
                             e,
                         )
                         time.sleep(delay)
+                        self._cancel_inflight_tool_calls(on_event)
                         self._reset_client_for_turn(on_token, on_event)
                     else:
                         raise
@@ -1144,11 +1201,16 @@ class ACPAgent(AgentBase):
                     )
                 ],
             )
+            # Close any tool cards left in flight from the timed-out attempt.
+            self._cancel_inflight_tool_calls(on_event)
             on_event(MessageEvent(source="agent", llm_message=error_message))
             state.execution_status = ConversationExecutionStatus.ERROR
         except Exception as e:
             logger.error("ACP prompt failed: %s", e, exc_info=True)
             error_str = str(e)
+
+            # Close any tool cards left in flight before surfacing the error.
+            self._cancel_inflight_tool_calls(on_event)
 
             # Emit error as an agent message (existing behavior, preserved for
             # consumers that inspect MessageEvents)
--- a/openhands-sdk/openhands/sdk/agent/acp_agent.py
+++ b/openhands-sdk/openhands/sdk/agent/acp_agent.py
@@ -295,13 +295,28 @@ class _OpenHandsACPBridge:
 
     Concurrency model — ``on_event`` / ``on_token`` / ``on_activity`` are
     fired synchronously from ``session_update``, which runs on the
-    ``AsyncExecutor`` portal thread.  The caller thread driving
-    ``ACPAgent.step()`` is blocked inside ``portal.call()`` for the entire
-    ``prompt()`` round-trip, so these callbacks do not race with the final
-    ``MessageEvent`` / ``FinishAction`` emitted by the caller thread after
-    ``prompt()`` returns.  Consumers that keep cross-callback state (e.g.
-    hook processors reading-then-writing, visualizers) can therefore treat
-    each callback as sequential within a single turn.
+    ``AsyncExecutor`` portal thread.  The guarantees that keep callbacks
+    serialized within a single turn rely on the combination of two things,
+    not the GIL alone:
+
+    1. ``LocalConversation.run()`` calls ``agent.step(...)`` while holding
+       the reentrant ``ConversationState`` lock (a ``FIFOLock``) — see
+       ``local_conversation.py`` where ``self.agent.step(...)`` sits inside
+       ``with self._state:``.  The caller thread owns that lock for the
+       entire duration of ``step()``, so no other thread can append to
+       ``state.events`` during the turn.
+    2. ``portal.call(_prompt)`` blocks the caller thread until ``prompt()``
+       returns.  Live ``on_event`` calls happen on the portal thread while
+       the caller thread is parked inside ``portal.call()`` still owning
+       the state lock; the final ``MessageEvent`` / ``FinishAction`` run
+       on the caller thread after ``prompt()`` returns.  The two phases
+       never overlap in time.
+
+    The caller's state-lock ownership is what excludes *other* threads
+    (hook workers, remote-conversation push layers, visualizers spawned
+    elsewhere) from racing with either phase.  The ordering between the
+    two phases is what keeps a single consumer's cross-callback state
+    (e.g. hook processors that read-then-write) consistent.
     """
 
     def __init__(self) -> None:
@@ -955,7 +970,7 @@ class ACPAgent(AgentBase):
         self._client.on_event = on_event
         self._client.on_activity = self._on_activity
 
-    def _cancel_inflight_tool_calls(self, on_event: ConversationCallbackType) -> None:
+    def _cancel_inflight_tool_calls(self) -> None:
         """Emit a terminal ``failed`` ACPToolCallEvent for every tool call
         in the accumulator that has not reached a terminal status yet.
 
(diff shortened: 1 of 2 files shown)


openhands-sdk/openhands/sdk/agent/acp_agent.py
```diff
@@ -407,12 +414,41 @@ async def session_update(
...
+                tool_call_id=tc["tool_call_id"],
+                title=tc["title"],
+                status=tc.get("status"),
+                tool_kind=tc.get("tool_kind"),
+                raw_input=tc.get("raw_input"),
+                raw_output=tc.get("raw_output"),
+                content=tc.get("content"),
+                is_error=tc.get("status") == "failed",
+            )
+            self.on_event(event)
```
VascoSch92, line 494: It can be that two threads write to shared state during a turn:                                                                                                                                                    
  - Caller thread (lock held): appends the final MessageEvent / FinishAction once prompt() returns.                                                                                                                  
  - Portal thread (lock not held): appends each live ACPToolCallEvent and mutates last_user_message_id as notifications arrive.
                                                                                                                                                                                                                     
  `list.append` survives this by accident (GIL), so the surface symptoms are hidden, but any callback doing more than a single atomic op (hook processors that read-then-write, visualizers rendering mid-mutation, remote-conversation push layers) can trigger a race condition.

VascoSch92 (commented):
Left a comment about a possible race condition. 

Perhaps also the following can also happen:

--- 
When the ACP subprocess hits a transient error (connection glitch, -32603 from Gemini), step() retries the whole prompt() call. Here's what happens around the retry (acp_agent.py:1027 and :1051):                
                                                                                                                                                                                                                     
  1. First attempt starts. Claude (the ACP server) begins a turn and fires tool-call notifications: ToolCallStart("toolu_AAA", status=pending), ToolCallProgress("toolu_AAA", status=pending, title='Read file').    
  2. With this PR, each of those notifications immediately emits an ACPToolCallEvent through on_event, which gets appended to state.events. So state.events now contains two entries for toolu_AAA, both with
  status=pending — nothing with status=completed yet.                                                                                                                                                                
  3. The server throws a retriable error before the tool call finishes. The code lands at line 1027 / 1051 and calls _reset_client_for_turn(...), which calls self._client.reset() (clearing the in-memory
  accumulated_tool_calls dict).                                                                                                                                                                                      
  4. Retry attempt starts. The server runs the turn again from scratch. ACP servers typically mint new tool-call ids on the retried turn — so now we get toolu_BBB for the same logical "read the file" tool call.
  5. The retry succeeds, toolu_BBB goes through pending → completed, final MessageEvent fires, turn ends cleanly.                                                                                                    
                                                                                                                                                                                                                     
  What's wrong                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  _client.reset() wipes the in-memory accumulator, but it does not reach into state.events to remove or supersede the toolu_AAA events that were already emitted and appended during step 2.                         
   
  The result:                                                                                                                                                                                                        
                                                                  
  - state.events contains toolu_AAA with status=pending forever. No later event ever carries that id, so the consumer's "last-seen status wins" dedup logic sees pending as the final word.                          
  - UIs that render per-tool_call_id will show a ghost tool card stuck spinning forever next to the successful toolu_BBB card.
  - Event-log replay, trajectory export, and anything that walks state.events will report a non-terminal tool call that never actually existed in the final turn.                                                    
                                                                                                                                                                                                                     
Why this is new                                                                                                                                                                                                    
                                                                                                                                                                                                                     
  Under the old batched-at-end-of-turn emission, tool-call events were only fanned out after prompt() succeeded — so a retried first attempt produced zero events. Nothing from the aborted attempt ever reached     
  state.events. The live-emission change makes the partial first attempt user-visible and unrecoverable.
                                                                                                                                                                                                                     
Fix shape                                                       

  Either (a) emit a synthetic terminal ACPToolCallEvent(..., status="failed", superseded_by_retry=True) for every id in accumulated_tool_calls just before calling _client.reset() in the retry branches, or (b) track ids emitted per-attempt and, on retry, emit a cancellation event for each so consumers can close the cards.

simonrosenberg:
Thanks @VascoSch92 — both points addressed in 282b8781.

simonrosenberg:
Thanks — addressed both in 94c5e0e1.

**Thread-safety (unresolved → resolved).** You're right that the previous docstring hand-waved it. The actual mechanism is in `LocalConversation.run()`: `self.agent.step(...)` is called *inside* `with self._state:` ([`local_conversation.py:675`](https://github.com/OpenHands/software-agent-sdk/blob/94c5e0e1/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py#L675)), and the state lock is a reentrant `FIFOLock` owned by the caller thread for the entire step.

The guarantee is two-layered and I've expanded the bridge docstring to cite both:
1. **No other thread can race** — the caller thread's ownership of the state `FIFOLock` excludes every other thread (hook workers, push layers, visualizers spawned elsewhere) for the full duration of `step()`. Neither phase is ever contended by outside threads.
2. **The two in-step phases are temporally ordered** — live `on_event` calls fire on the portal thread while the caller thread is parked inside `portal.call()` (still owning the lock). Once `prompt()` returns, the caller thread resumes and emits the final `MessageEvent` / `FinishAction`. The phases are strictly sequential, so a consumer doing read-then-write across callbacks sees coherent state.

The default conversation callback comments this explicitly ([`local_conversation.py:193-196`](https://github.com/OpenHands/software-agent-sdk/blob/94c5e0e1/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py#L193-L196)): *"This callback runs while holding the conversation state's lock … so updating state here is thread-safe."*

**`_cancel_inflight_tool_calls` docstring mismatch.** Fixed by dropping the parameter and reading `self._client.on_event` directly — cancel is always called *before* `_reset_client_for_turn` at every call site (retry paths + TimeoutError + outer exception), so the bridge's `on_event` is still wired. Added a `test_noop_when_on_event_unset` case and kept the existing three by wiring `on_event` on the bridge instead of passing it as an argument. 130 ACP tests green; pyright clean.

VascoSch92 (approved):

PR #2888, opened 2026-04-19
fix: sanitize malformed tool call arguments in LLM history

### LINKED ISSUE #2887: [Bug]: Failed to parse tool call arguments as JSON - llama-server rejects prompt with invalid tool calls generated by OpenHands (not by the LLM!)

## Description

When using OpenHands with a `llama-server` inference backend, invalid tool calls generated by the model trigger a failure in the server (`500` error due to JSON parsing issues).

OpenHands correctly detects that the tool call is invalid and attempts to retry. However, it includes the offending tool call in the prompt causing `llama-server` to reject the prompt before doing a new inference. This causes a dead lock.

This happens after receiving once:

	Error validating args {"command":"str_replace","path":"...for tool 'file_editor': Unterminated string starting at: line 1 column 7372 (char 7371)

It causes all subsequent prompts to fail in a dead lock, without even starting a new inference.
The conversation can not be continued (events must be cleared to fix it).

This results in an infinite retry loop where:

* OpenHands keeps retrying sending the offending content in the prompt.
* `llama-server` keeps returning the same parsing error
* No meaningful inference progress is made

## Expected Behavior

* OpenHands should not include invalid tool calls in the prompt in a way that llama-server will try to parse them.
* A new inference should be started.

## Actual Behavior

* OpenHands keeps retrying sending the offending content in the prompt.
* `llama-server` keeps returning the same parsing error (`500` responses from `llama-server`)
* No meaningful inference progress is made
* Infinite retry loop with identical failure

## Logs (llama-server)

```
Apr 15 18:21:15 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
Apr 15 18:21:16 | srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: >
Apr 15 18:21:16 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
Apr 15 18:21:48 | srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: >
Apr 15 18:21:48 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
Apr 15 18:21:49 | srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: >
Apr 15 18:21:49 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
Apr 15 18:21:49 | srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: >
Apr 15 18:21:49 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
Apr 15 18:22:53 | srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: >
Apr 15 18:22:53 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
Apr 15 18:22:54 | srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: >
Apr 15 18:22:54 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
Apr 15 18:22:55 | srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: >
Apr 15 18:22:55 | srv  log_server_r: done request: POST /v1/chat/completions 172.21.0.4 500
```

## Possible Root Cause

* Model emits malformed JSON in `tool_calls.arguments`
* `llama-server` strictly parses and throws a 500 error
* OpenHands retries without modifying prompt/context sufficiently to fix the issue
* The agent is blocked and irresponsive.

### On OpenHands side

* OpenHands should not include invalid tool calls in the prompt in a way that llama-server will try to parse them.

## Model

Qwen-3.5, GLM, all models.

## Settings

config.toml includes:
- `native_tool_calling = true`

## Versions

- OH commit c78b92346830e1c83d090de4042f2d85b996bf02

xingyaoww:
## Summary

Fixes #2887

When an LLM produces tool calls with malformed JSON arguments (e.g. unterminated strings), the agent's `_emit_tool_error` was passing the raw malformed arguments into the `ActionEvent` stored in the event stream. On the next LLM call, these malformed arguments were included in the conversation history. Inference servers like `llama-server` parse the `arguments` field of tool calls in history and reject the entire prompt with a 500 error, causing an infinite retry loop.

## Root Cause

The `MessageToolCall.arguments` field has a contract: it must be a valid JSON string. When the LLM returns garbage, this contract was violated at the point where `_emit_tool_error` stored the raw tool call in the `ActionEvent`.

## Fix

Added `_ensure_valid_tool_call_arguments()` helper that validates the tool call arguments are valid JSON. If not, it returns a copy with `arguments="{}"`. This is called in `_emit_tool_error` before creating the `ActionEvent`, ensuring malformed data never enters the event stream.

The error details are already conveyed via the `AgentErrorEvent` that follows the `ActionEvent`, so replacing malformed arguments with `"{}"` loses no actionable information for the LLM.

### Changes
- `openhands-sdk/openhands/sdk/agent/agent.py`: Added `_ensure_valid_tool_call_arguments()` and use it in `_emit_tool_error`
- `tests/sdk/agent/test_malformed_tool_call_arguments.py`: Unit + end-to-end tests

---

_This PR was created by an AI assistant (OpenHands)._

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:675b19d-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-675b19d-python \
  ghcr.io/openhands/agent-server:675b19d-python
```

```
ghcr.io/openhands/agent-server:675b19d-golang-amd64
ghcr.io/openhands/agent-server:675b19d-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:675b19d-golang-arm64
ghcr.io/openhands/agent-server:675b19d-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:675b19d-java-amd64
ghcr.io/openhands/agent-server:675b19d-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:675b19d-java-arm64
ghcr.io/openhands/agent-server:675b19d-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:675b19d-python-amd64
ghcr.io/openhands/agent-server:675b19d-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:675b19d-python-arm64
ghcr.io/openhands/agent-server:675b19d-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:675b19d-golang
ghcr.io/openhands/agent-server:675b19d-java
ghcr.io/openhands/agent-server:675b19d-python
```

- Each variant tag (e.g., `675b19d-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `675b19d-python-amd64`) are also available if needed


openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -104,6 +104,19 @@ def _tool_has_summary_param(tool: ToolDefinition) -> bool:
...
 
+def _ensure_valid_tool_call_arguments(tool_call: MessageToolCall) -> MessageToolCall:
+    """Return *tool_call* with guaranteed-valid JSON in ``arguments``.
+
+    If the arguments are already valid JSON the original object is returned
+    unchanged.  Otherwise a shallow copy with ``arguments="{}"`` is returned.
+    """
+    try:
+        json.loads(tool_call.arguments)
+    except (json.JSONDecodeError, TypeError):
```
xingyaoww: @OpenHands Could you take a look at the top-level exception handling for the LM call?

I'm wondering if I can catch it there and convert it into something like a "Malformed LM Request Error." I recall there being a place in our codebase that handles this—for example, if a malformed request occurs, we automatically retry it. If the retry fails, it should just throw an error.

xingyaoww: Good point! The existing `LLMMalformedConversationHistoryError` handling in `step()` already does exactly this — condensation retry, and raise if no condenser.

The missing piece was that the exception classifier didn't recognize the specific 500 error from llama-server (`"Failed to parse tool call arguments as JSON"`) as a malformed history error. Instead it was mapping `InternalServerError` → `LLMServiceUnavailableError`, which isn't caught in `step()` and caused the infinite retry loop.

Fixed in f18f7d51:
- Added `"failed to parse tool call arguments"` to `MALFORMED_HISTORY_PATTERNS` in the classifier
- Added `InternalServerError` to the `isinstance` check in `looks_like_malformed_conversation_history_error()` (litellm's `InternalServerError` isn't a subclass of litellm's `OpenAIError`, so it wasn't covered)
- This routes to the existing `LLMMalformedConversationHistoryError` handler in `step()` which triggers condensation recovery

The source-level sanitization in `_emit_tool_error` (`_ensure_valid_tool_call_arguments`) is kept as defense-in-depth — it prevents malformed data from entering the event stream in the first place, avoiding wasted retries.

enyst:
@openhands Look at my PRs in the past several weeks, find the one about condensing history when the LLM history is malformed. WDYT, could we include the case addressed in this PR in a similar fashion with that?

Post your suggestion here, for the PR author, as a comment on gh.

enyst:
I looked at merged PR #2613 (`fix(sdk): recover malformed tool history via condensation`), and I think this case fits the same malformed-history bucket **as a fallback**, but not as the primary fix.

My take:

- **Primary fix should stay sanitize-on-write** in `_emit_tool_error`. That fixes the root cause at the point where we violate the `MessageToolCall.arguments` contract, so malformed JSON never enters the event stream.
- **#2613-style malformed-history classification / condensation is still worth including as defense in depth** for cases where the bad history already exists (for example resumed conversations, older persisted events, or histories created before this fix landed).
- I would **not** rely on condensation alone here, because it only helps after we have already stored invalid data, and only when a condenser is configured. Sanitizing in `_emit_tool_error` fixes the bug even without a condenser.

So my suggestion would be: keep the sanitize-before-persisting change as the main fix, and also route provider errors like `Failed to parse tool call arguments as JSON` through the #2613 malformed-history recovery path as a secondary safeguard.

_This comment was created by an AI assistant (OpenHands) on behalf of the user._

enyst:
Re: the agent's response https://github.com/OpenHands/software-agent-sdk/pull/2888#issuecomment-4283054688

I seem to recall that in V0, we've played quite a bit of whack-a-mole with malformed LLM responses, specially from non-SOTA LLMs / local LLMs. I can see how this kind of fixes could accumulate code after code...

I think maybe a "forced" condensation is a good idea in these cases and it could be OK... Sooner or later, a condenser configured seems like a must have anyway.

The agent thinks we should have both, I'm not sure... maybe it depends on case, because these fixes could accumulate to contain the weirdest one-off things 🤔 . WDYT? @xingyaoww

xingyaoww:
Refactored in 1d7a8bf5. Here's how the flow now works:

**When the LLM produces malformed JSON in tool call arguments:**

1. `_get_action_event()` tries to parse the arguments via `parse_tool_call_arguments()`
2. `json.JSONDecodeError` is caught and re-raised as `FunctionCallValidationError`
3. This propagates up through `_handle_tool_calls()` to `step()`
4. `step()` catches it and emits a user-role `MessageEvent` with the error details
5. On the next turn, the LLM sees the error message and can self-correct

**No malformed data enters the event stream.** No `ActionEvent` is stored. No sanitization hack needed.

**When malformed data is already in history** (e.g., restored state from before this fix):

1. The inference server (e.g., llama-server) rejects the prompt with a 500 error containing `"Failed to parse tool call arguments as JSON"`
2. The exception classifier (from the previous commit f18f7d51) recognizes this as `LLMMalformedConversationHistoryError`
3. `step()` catches it and triggers condensation recovery (or raises if no condenser)

So we have two layers of defense:
- **Primary**: `FunctionCallValidationError` prevents malformed data from entering the event stream
- **Secondary**: Exception classifier catches the server error if malformed data got in through other paths

_This comment was generated by an AI assistant (OpenHands)._

xingyaoww:
@enyst Good point about the whack-a-mole concern. After xingyaoww's feedback, I've refactored the approach. Here's where things stand:

**Current state of the PR (3 commits):**

1. `1630c1cc` — Original sanitize-on-write approach (now superseded)
2. `f18f7d51` — Added classifier pattern for `"failed to parse tool call arguments"` from InternalServerError → LLMMalformedConversationHistoryError → condensation recovery
3. `1d7a8bf5` — **The real fix**: removed the sanitization hack entirely. Instead, malformed JSON in tool call arguments now raises `FunctionCallValidationError`, which `step()` catches and reports back to the LLM as a user message. No malformed data enters the event stream at all.

**My verdict:**

Commit 3 (`FunctionCallValidationError`) is the clean fix and it's sufficient on its own. It's not whack-a-mole — it's a universal check ("are the arguments valid JSON?") that catches any LLM producing garbage, regardless of provider.

Commit 2 (the classifier pattern) is the debatable one. You're right that these pattern-by-pattern fixes accumulate — today it's `"failed to parse tool call arguments"`, tomorrow it's another server-specific error string. With commit 3 in place, malformed args can't enter the event stream going forward, so the classifier only matters for conversations persisted *before* this fix. That's a thin edge case, and exactly the kind of speculative defense-in-depth that adds up over time.

I'd lean toward dropping commit 2 and keeping only the `FunctionCallValidationError` approach. Happy to revert if you and @xingyaoww agree.

_This comment was generated by an AI assistant (OpenHands)._

enyst:
@openhands /codereview /codereview-roasted Read all comments before ; then review and post your review with gh api here

openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -790,8 +794,16 @@ def _get_action_event(
...
         try:
-            # Parse arguments inside the try block so JSONDecodeError is caught.
-            arguments = parse_tool_call_arguments(tool_call.arguments)
+            # Parse arguments first — malformed JSON is raised as
+            # FunctionCallValidationError so step() can handle it without
+            # storing invalid data in the event stream (#2887).
+            try:
+                arguments = parse_tool_call_arguments(tool_call.arguments)
+            except json.JSONDecodeError as e:
+                raise FunctionCallValidationError(
```
enyst: This new raise is fine for the single-tool case, but it interacts badly with the existing emission model: `_get_action_event()` still emits earlier valid `ActionEvent`s before the full batch is validated. If the second/third tool call hits this branch, `step()` catches it at the top level and returns before `_execute_actions()` runs, so the already-emitted actions are left without any matching observation/error event. That recreates malformed history for mixed batches.

enyst (commented):
🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/agent/agent.py`] **Correctness / malformed history**: this top-level `FunctionCallValidationError` path makes multi-tool responses non-atomic. `_get_action_event()` still emits each valid `ActionEvent` immediately, but now a later malformed tool call aborts `_handle_tool_calls()` before `_execute_actions()` runs. In a batch like `[valid call, malformed call]`, the earlier valid `ActionEvent` is already persisted, then `step()` emits only the user `MessageEvent` and returns. That leaves a dangling tool call in history with no matching `ObservationEvent` / `AgentErrorEvent` — i.e. the same malformed-history shape this PR is trying to prevent. I reproduced this locally with a mocked 2-call response and got `ActionEvent(validation_test_tool)` followed by the user error `MessageEvent`, with no tool result for the first call.

**[TESTING GAPS]**
- [`tests/sdk/agent/test_malformed_tool_call_arguments.py`] The new tests only cover the single-call case. Please add a regression for a mixed batch (`valid tool call` + `malformed tool call`) so this partial-emission bug is caught.

**[RISK ASSESSMENT]**
  This changes core agent/tool-call control flow and event-stream integrity. The current implementation can still write malformed history, just through a different path, and that can wedge subsequent turns again. **Recommendation**: Do not auto-merge. Have a human maintainer re-check the tool-call batch semantics after the fix, since this is agent-behavior / eval-sensitive code.

**VERDICT:**
❌ **Needs rework**

**KEY INSIGHT:**
The single-tool case is fixed, but the batch semantics are now wrong: once one bad tool call appears after a good one, the PR reintroduces malformed history through partially emitted `ActionEvent`s.

enyst:
I re-read my own review with fresh eyes and did a second pass through the codepath end-to-end. Here's the section-by-section self-audit.

## 1) Critical issue — was it really the case?

**Not exactly in the way I originally wrote it.**

What I previously claimed: a mixed batch like `[valid tool call, malformed tool call]` would leave a dangling `ActionEvent` in history and that exact malformed history would be sent back to the LLM on the next turn.

After re-checking the view layer, that part is **too strong**:
- `prepare_llm_messages()` goes through `View.from_events()`
- `ToolCallMatchingProperty` drops unmatched `ActionEvent`s from the *view*
- so **immediately after the failing step**, the orphaned action is *not* included in the next sampled prompt

So my earlier review overstated the first-order failure mode.

## 2) Critical issue — is there still a real bug there?

**Yes, but it's subtler than I said.**

The part that still looks real is this:
- `step()` starts by checking `ConversationState.get_unmatched_actions(state.events)` on the **raw event stream**, not on the filtered `View`
- so on the next iteration, the previously-emitted valid `ActionEvent` still gets executed
- that makes the persisted order become:
  - `ActionEvent(valid)`
  - `MessageEvent(user corrective error)`
  - `ObservationEvent(valid)`

I reproduced that locally.

At that point, `prepare_llm_messages()` preserves the order and reconstructs:
- `assistant` with tool call
- then `user` corrective message
- then `tool` result

That is still broken tool-call sequencing for strict providers: the tool result is no longer adjacent to the assistant tool call.

So the **atomicity concern survives**, but in a different form:
- not “the tool result never exists”
- but “the tool result exists later, after an intervening user message”

That distinction matters, and I should have been more precise in the original review.

## 3) Testing gaps — was that fair?

**Yes, and I'd actually strengthen it.**

The current tests cover the single malformed-call case well enough, but they don't cover the mixed-batch / multi-step evolution that exposes the subtler ordering problem.

The regression test I'd want now is:
1. first step: LLM returns `[valid tool call, malformed tool call]`
2. assert the malformed call is surfaced as the user corrective message
3. second step: assert the pending valid action executes
4. then assert the prepared LLM history is still structurally valid (or, today, demonstrate that it isn't)

So the “tests are missing the interesting edge case” point still stands.

## 4) Risk assessment — was I too harsh?

**Yes on the exact wording; probably somewhat on the color too.**

I don't think I can still defend my previous 🔴 **HIGH** wording if it is read as:
> “this definitely recreates the exact original malformed-history bug immediately on the next prompt.”

That's not what the code does, because the `View` layer filters the unmatched action on the first pass.

But I also don't think this is minor:
- it's still in agent core control flow
- it still touches LLM-history structural correctness
- it still has a reproducible mixed-batch failure mode
- and the fallback behavior depends on later recovery / provider tolerance / condensation configuration

So if I were rewriting my own review today, I'd probably downgrade the **risk label** from **🔴 HIGH** to **🟡 MEDIUM**, while still saying **don't approve yet** until the mixed-batch path is addressed or convincingly shown safe.

## 5) Bottom line / updated verdict

My earlier review was **directionally right but mechanically imprecise**.

Updated take:
- the PR **does fix the single malformed-tool-call case cleanly**
- my original “dangling action immediately goes back into history” claim was **too strong**
- but there still appears to be a **real mixed-batch ordering bug** once the pending valid action is executed on the next step

So I would revise my own position to:
- **not as major as my first wording made it sound**
- **not minor either**
- still worth fixing / testing before merge

enyst:
I re-checked the updated branch and my earlier mixed-batch concern no longer reproduces on the current tip.

What changed materially: malformed tool-call JSON is now rejected in `LLM.completion()` before the raw provider response is converted into agent-level `ActionEvent`s, so the old partial-batch path never gets a chance to persist.

I pushed a regression test in `1a5b24c6` to lock that in:
- `tests/sdk/agent/test_malformed_tool_call_arguments.py`
- new test: `test_mixed_tool_batch_is_retried_before_any_action_event_is_persisted()`

The test intentionally uses a real `LLM` with patched `litellm_completion` instead of `TestLLM`, because `TestLLM` scripts already-converted SDK `Message`s and would bypass the new validation layer entirely.

So net: my old concern was valid for the previous revision, but with the current branch the right thing to do is preserve the fix with a regression test rather than block on that earlier review comment.

openhands-sdk/openhands/sdk/agent/agent.py
enyst: Re-checked on the updated branch: this specific mixed-batch issue no longer reproduces because malformed tool-call JSON is now rejected inside `LLM.completion()` before any agent-level `ActionEvent` is emitted. Added regression coverage in 1a5b24c6 (`test_mixed_tool_batch_is_retried_before_any_action_event_is_persisted`).

enyst:
HUMAN: I think OpenHands-GPT-5.4 is happy now, at least with the old history in context. Sorry for posting on a draft 🙏

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -177,6 +179,42 @@
 LLM_PROFILE_SCHEMA_VERSION: Final[int] = 1
 
 
+def _validate_tool_call_args(resp: ModelResponse) -> None:
```
VascoSch92, line 182: Nit: Could we place this in another place?

The LLM class is already huge and this method is actually static

VascoSch92:
Closing as superseded.

Issue #2887 was closed as *completed* by #3575 ("[codex] Sanitize malformed tool call recovery", merged 2026-06-10), which touches the same `tests/sdk/agent/test_tool_validation_error_message.py`. `main` now carries `fix_malformed_tool_arguments` in `agent/agent.py` and `looks_like_malformed_conversation_history_error` in `llm/exceptions/classifier.py`.

PR #2902, opened 2026-04-21
feat(sdk/io): add PostgreSQLFileStore for persistent EventLog

### LINKED ISSUE #2903: Add PostgreSQLFileStore — persistent EventLog for ephemeral deployments
In environments like Cloud Run or containers without persistent volumes, `LocalFileStore` either fails or falls back to `InMemoryFileStore`, which means the full EventLog (including `tool_call`/`tool_result` events) is lost between requests. This makes proper multi-turn resume impossible.

Proposal: add `PostgreSQLFileStore` as an optional `FileStore` backend (`pip install openhands-sdk[postgresql]`) that persists the EventLog to a PostgreSQL table, namespaced by conversation ID. Locking uses `pg_try_advisory_lock` so it's safe across multiple processes/instances.

```python
store = PostgreSQLFileStore(dsn=..., namespace=str(conversation_id))
conv = LocalConversation(agent=agent, workspace=workspace, file_store=store)
```

Happy to open a PR if this direction looks reasonable.

StatPan:
Closes #2903

## Summary

- Adds `PostgreSQLFileStore` — a `FileStore` backed by PostgreSQL for environments where the local filesystem is unavailable or ephemeral (Cloud Run, containers without persistent volumes)
- Adds `file_store: FileStore | None = None` parameter to `ConversationState.create()` and `LocalConversation.__init__()` so callers can inject any `FileStore` implementation

## Motivation

In ephemeral environments, `LocalFileStore` requires a filesystem path that doesn't persist across requests. The SDK falls back to `InMemoryFileStore`, losing all event history. Without a persistent EventLog, full multi-turn resume — including `tool_call` / `tool_result` events — is impossible.

## Changes

| File | Change |
|---|---|
| `openhands/sdk/io/postgresql.py` | New: `PostgreSQLFileStore` implementation |
| `openhands/sdk/io/__init__.py` | Conditional re-export (no hard asyncpg dep) |
| `openhands/sdk/conversation/state.py` | `file_store` param on `ConversationState.create()` |
| `openhands/sdk/conversation/impl/local_conversation.py` | `file_store` param on `LocalConversation.__init__()` |
| `pyproject.toml` | Optional extra `[postgresql]` = `asyncpg>=0.29.0` |
| `tests/sdk/io/test_postgresql_filestore.py` | 10 unit tests (mocked asyncpg) |

All changes are **backward-compatible**: `file_store=None` preserves existing `LocalFileStore`/`InMemoryFileStore` selection logic.

## Design notes

- **Background event-loop thread + asyncpg pool**: bridges the sync `FileStore` interface to async asyncpg without creating a new loop per call and without interfering with the host app's event loop
- **`pg_advisory_lock`-based locking**: safe for concurrent writers across multiple processes and instances (Cloud Run, autoscaled deployments). Lock key = SHA-256(`namespace + "\x00" + path`)[:8] as signed int64. Uses `pg_try_advisory_lock` with 50 ms polling and a dedicated connection per `lock()` call to keep the session-level advisory lock lifecycle self-contained.
- **Optional import**: `asyncpg` is not a hard dependency. `PostgreSQLFileStore` is re-exported from `openhands.sdk.io` only when asyncpg is installed; `__init__.py` uses `try/except ImportError`

## Usage

```python
from openhands.sdk.io.postgresql import PostgreSQLFileStore

store = PostgreSQLFileStore(
    dsn="postgresql://user:pass@host:5432/db",
    namespace=str(conversation_id),  # isolates events per conversation
)
# New conversation:
conv = LocalConversation(agent=agent, workspace=workspace, file_store=store)

# Resume across requests (same namespace → SDK reads full event history from DB):
conv = LocalConversation(agent=agent, workspace=workspace, file_store=store)
```

## Docs

OpenHands/docs#465


xingyaoww:
@OpenHands /codereview /github-pr-review

---
_🤖 This comment was automatically posted by the **pr-triage** skill ([OpenHands](https://github.com/All-Hands-AI/OpenHands)) on behalf of a maintainer. See [PR #2906](https://github.com/OpenHands/software-agent-sdk/pull/2906) for skill details._

openhands-sdk/openhands/sdk/io/postgresql.py
```diff
@@ -0,0 +1,249 @@
...
+    def __init__(self, dsn: str, namespace: str, table: str = DEFAULT_TABLE) -> None:
+        self._dsn = dsn
+        self._namespace = namespace
+        self._table = table
+
+        # Dedicated event loop in a daemon background thread.
+        # run_coroutine_threadsafe() is safe to call from any thread,
+        # including run_in_executor workers used by the SDK.
+        self._loop = asyncio.new_event_loop()
+        self._bg_thread = threading.Thread(
```
xingyaoww, line 99: 🔴 **Critical — SQL injection via `table` parameter.** The `table` argument is interpolated directly into every SQL query via f-strings (`f"INSERT INTO {self._table} ..."`). While data values use `$1`-style parameterized queries (good), the table name itself is unparameterized. A malicious or accidental `table` value like `"x; DROP TABLE users; --"` would execute arbitrary SQL.

Fix: validate the table name in the constructor against a strict identifier regex:

```suggestion
        self._table = table
        if not _VALID_IDENT_RE.match(table):
            raise ValueError(
                f"Invalid table name {table!r}: must be a valid SQL identifier"
            )
```

Add at module level:
```python
import re
_VALID_IDENT_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
```

Alternatively, quote it via `asyncpg`'s `utils.quote_ident` or just hardcode the table and don't accept it as a parameter.

openhands-sdk/openhands/sdk/io/postgresql.py
```diff
@@ -0,0 +1,249 @@
...
+        # type-stub limitations (asyncpg.create_pool is not typed as a coroutine).
+        future: concurrent.futures.Future[asyncpg.Pool] = (
+            asyncio.run_coroutine_threadsafe(self._setup(dsn), self._loop)
+        )
+        self._pool: asyncpg.Pool = future.result()
+
+    # ------------------------------------------------------------------ #
+    # Internal helpers
+    # ------------------------------------------------------------------ #
+
```
xingyaoww, line 179: 🟠 **Important — Resource leak if `_setup()` fails.** The background thread is started on line 86, but if `future.result()` on line 92 raises (bad DSN, network error, DB permission issue), the event loop thread is never stopped. The daemon flag prevents it from blocking process exit, but it's still a leaked thread + event loop on every failed instantiation attempt.

Consider wrapping the `future.result()` call in a try/except that stops the loop and joins the thread on failure:
```python
try:
    self._pool: asyncpg.Pool = future.result()
except Exception:
    self._loop.call_soon_threadsafe(self._loop.stop)
    self._bg_thread.join(timeout=5)
    raise
```

openhands-sdk/openhands/sdk/io/postgresql.py
```diff
@@ -0,0 +1,249 @@
...
+    ) -> asyncpg.Connection:
+        """Acquire a session-level advisory lock on a dedicated connection."""
+        conn: asyncpg.Connection = await asyncpg.connect(self._dsn)
+        deadline = self._loop.time() + timeout
+        while True:
+            acquired = bool(await conn.fetchval("SELECT pg_try_advisory_lock($1)", key))
+            if acquired:
+                return conn
+            remaining = deadline - self._loop.time()
+            if remaining <= 0:
```
xingyaoww, line 218: 🟡 **Suggestion — Connection-per-lock is expensive.** Each `lock()` call opens a new TCP connection via `asyncpg.connect()` (bypassing the pool). Under load with concurrent conversations, this creates significant connection churn. PostgreSQL's default `max_connections=100` could be exhausted by lock connections alone.

Consider acquiring a connection from the pool instead, or documenting this trade-off prominently. The session-level advisory lock lifecycle rationale is sound, but pool connections also maintain session state until released.

openhands-sdk/openhands/sdk/io/postgresql.py
```diff
@@ -0,0 +1,249 @@
...
+        try:
+            yield
+        finally:
+            self._run(self._lock_release(conn, key))
+
+    def close(self) -> None:
+        """Close the connection pool and stop the background event loop."""
+        self._run(self._pool.close())
+        self._loop.call_soon_threadsafe(self._loop.stop)
+        self._bg_thread.join(timeout=5)
```
xingyaoww, line 328: 🟡 **Suggestion — No context manager support.** Users must remember to call `close()` or resources leak. Consider adding `__enter__`/`__exit__` for `with` statement support:
```python
def __enter__(self) -> "PostgreSQLFileStore":
    return self

def __exit__(self, *exc: object) -> None:
    self.close()
```
This aligns with the `asyncpg.Pool` pattern and is a common Python convention for resource-owning objects.

xingyaoww (commented):
🟡 **Acceptable** — useful feature, backward-compatible plumbing, but has a real SQL injection vector and testing gaps that should be addressed before merging.

_This review was generated by an AI assistant (OpenHands) on behalf of the maintainers._

openhands-sdk/openhands/sdk/io/postgresql.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # ------------------------------------------------------------------ #
+    # Internal helpers
+    # ------------------------------------------------------------------ #
+
+    async def _setup(self, dsn: str) -> asyncpg.Pool:
+        pool = await asyncpg.create_pool(dsn)
+        async with pool.acquire() as conn:
+            await conn.execute(
+                f"""
```
xingyaoww: 🟠 **Important — DDL in the data path.** `CREATE TABLE IF NOT EXISTS` and `CREATE INDEX IF NOT EXISTS` run on every `PostgreSQLFileStore` instantiation. In a multi-process/autoscaled deployment (the stated use case), this means every request fires DDL statements. While `IF NOT EXISTS` is idempotent, DDL acquires `AccessExclusiveLock` on the relation for `CREATE TABLE` and `ShareLock` for `CREATE INDEX`, which can cause contention under load.

Consider separating schema migration from instance construction — e.g., a `ensure_schema()` classmethod or a separate migration step — and documenting that the schema should be created once before deploying.

tests/sdk/io/test_postgresql_filestore.py
```diff
@@ -0,0 +1,208 @@
+"""Tests for PostgreSQLFileStore.
```
xingyaoww, line 1: 🟠 **Important — Tests are 100% mocked; they verify the mock, not the SQL.** The fixture reimplements SQL logic in Python (parsing query strings like `if "INSERT INTO" in query`), so the tests only prove the mock behaves correctly. Real bugs — wrong column names, broken `ON CONFLICT`, incorrect `LIKE` semantics, advisory lock edge cases — would pass these tests.

At minimum:
1. The mock-based tests should be clearly labeled as **contract tests** (they verify the FileStore interface contract, not PostgreSQL correctness).
2. Add an integration test (gated behind a `pytest.mark.skipif` for a `POSTGRES_DSN` env var or a `@pytest.mark.integration` marker) that runs against a real PostgreSQL instance. Even a single round-trip test (`write → read → list → delete → exists`) would catch most SQL bugs.

Per repo guidelines: *"Do not accept tests that are just a pile of mocks asserting that functions were called."* These are slightly better since they test state transitions through the mock, but they still can't catch real SQL issues.

openhands-sdk/openhands/sdk/conversation/state.py
```diff
@@ -267,6 +267,7 @@ def create(
         stuck_detection: bool = True,
         cipher: Cipher | None = None,
         tags: dict[str, str] | None = None,
+        file_store: FileStore | None = None,
```
xingyaoww, line 285: 🟡 **Suggestion — Document the `file_store` + `persistence_dir` interaction.** When `file_store` is provided, `persistence_dir` is still passed to the `ConversationState` constructor and stored. The `persistence_dir` field is used for `env_observation_persistence_dir` (line 223-227). If a caller provides `file_store` but no `persistence_dir`, observation persistence silently becomes `None`.

Consider adding a note in the docstring about this interaction, or raising a warning when both are provided with a `persistence_dir` that would normally trigger `LocalFileStore` creation.

StatPan:
Thanks for the thorough and speedy review! Didn't expect feedback this fast — almost like you had a robot reading it for you. Oh wait. 😄

All points are well taken. Here's my plan:

**Will fix in next push:**

1. **SQL injection on `table`** — Yep, that's a real one. Adding identifier regex validation in the constructor. Considered `asyncpg`'s `quote_ident` but a strict `^[a-zA-Z_][a-zA-Z0-9_]*$` check is simpler and doesn't need a live connection.

2. **Resource leak on `_setup()` failure** — Good catch. Will wrap `future.result()` in try/except to stop the loop and join the thread on failure.

3. **Context manager support** — Easy win, adding `__enter__`/`__exit__`.

4. **DDL in data path** — Will extract to a `ensure_schema()` classmethod. Constructor will skip DDL by default; callers opt in or run it once during deployment.

**Will address with a note / follow-up:**

5. **Connection-per-lock** — This was intentional: pool connections release advisory locks when returned to the pool, which breaks the lock lifecycle. I'll add a docstring explaining the trade-off. Open to revisiting if there's a pattern I'm missing.

6. **Tests** — Fair point. Will relabel existing tests as contract tests and add an integration test gated behind `@pytest.mark.integration` + `POSTGRES_DSN` env var.

7. **`file_store` + `persistence_dir` interaction** — Will add a docstring note clarifying the behavior when both/neither are provided.

Working on the fixes now — should have a push up today. Thanks again for the detailed feedback!

openhands-sdk/openhands/sdk/io/postgresql.py
StatPan: Fixed in cfe7473. Added `_VALID_IDENT_RE = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*$')` validation in the constructor. Invalid table names now raise `ValueError`.

openhands-sdk/openhands/sdk/io/postgresql.py
StatPan: Fixed in cfe7473. `future.result()` is now wrapped in try/except — on failure, stops the event loop and joins the background thread before re-raising.

openhands-sdk/openhands/sdk/io/postgresql.py
StatPan: Added docstring to `_lock_acquire` explaining the trade-off: pool connections release advisory locks when returned, breaking the lock lifecycle. Dedicated connection keeps the session-level lock self-contained.

openhands-sdk/openhands/sdk/io/postgresql.py
StatPan: Fixed in cfe7473. Added `__enter__`/`__exit__` for context manager support.

openhands-sdk/openhands/sdk/io/postgresql.py
StatPan: Fixed in cfe7473. Extracted DDL to `ensure_schema()` classmethod. Constructor now accepts `auto_create_schema` flag (default False) — callers opt in or run it once during deployment.

tests/sdk/io/test_postgresql_filestore.py
StatPan: Fixed in cfe7473. Relabeled existing tests as contract tests. Added integration test suite gated behind `@pytest.mark.integration` + `POSTGRES_DSN` env var with full round-trip coverage (write → read → list → delete → exists).

openhands-sdk/openhands/sdk/conversation/state.py
StatPan: Fixed in cfe7473. Added docstring to `create()` clarifying the `file_store` + `persistence_dir` interaction, including behavior when both/neither are provided.

StatPan:
Hi @xingyaoww, all 7 review items have been addressed in cfe7473:

1. **SQL injection** — `_VALID_IDENT_RE` regex validation on table name
2. **Resource leak on setup failure** — try/except around `future.result()`, cleans up loop + thread
3. **Connection-per-lock** — docstring explaining why pool isn't used (advisory lock lifecycle)
4. **Context manager** — `__enter__`/`__exit__` added
5. **DDL in data path** — extracted to `ensure_schema()` classmethod + `auto_create_schema` flag
6. **Mock-only tests** — relabeled as contract tests, added `@pytest.mark.integration` suite with real round-trip coverage
7. **file_store + persistence_dir docs** — docstring on `create()` clarifying the interaction

Would appreciate a re-review when you have a moment. Thank you!

StatPan:
@mamoodi @cbagwell Could you please review this PR when you get a chance? 🥺\n\nAll checks are now passing (including pre-commit), and the branch is merge-ready. Thank you! 🙏

StatPan:
Resolved the merge conflicts by merging the latest upstream/main into feat/postgresql-filestore and pushed the update in 25701496.
Local verification:
- `uv lock --check`
- `uv run pytest tests/sdk/io/test_postgresql_filestore.py` (12 passed, 6 skipped)
- `uv run pre-commit run --files ...` for the PostgreSQLFileStore-related files

StatPan:
I opened a smaller PR that splits out only the custom `FileStore` injection point: #3252.

That PR intentionally does not include `PostgreSQLFileStore`, asyncpg, schema management, or lockfile-heavy backend changes. It only adds:

- `ConversationState.create(..., file_store=...)`
- `LocalConversation(..., file_store=...)`

with the existing `persistence_dir` behavior preserved when no custom store is provided.

If the narrower API direction is acceptable, I’m happy to treat #3252 as the review path for the SDK extension point and keep the PostgreSQL backend external or propose it separately later.

StatPan:
@all-hands-bot what Can I do?

PR #2910, opened 2026-04-21
feat: bundle google-cloud-aiplatform for Vertex partner models

abguymon:
## Summary

Adds `google-cloud-aiplatform` as a required dep of `openhands-sdk` so LiteLLM's `vertex_ai_partner_models` handler can route to Vertex AI Model Garden MaaS endpoints — MiniMax, Qwen, Kimi, etc.

Without this dep, requests like `vertex_ai/minimaxai/minimax-m2-maas`, `vertex_ai/qwen/qwen3-coder-*-maas`, `vertex_ai/moonshotai/kimi-k2-thinking-maas` fail at the first completion call with:

```
litellm.BadRequestError: Vertex_aiException BadRequestError -
vertexai import failed please run `pip install -U "google-cloud-aiplatform>=1.38"`.
Got error: No module named 'vertexai'
```

## What changed

- `openhands-sdk/pyproject.toml`: add `"google-cloud-aiplatform>=1.38"`
- `uv.lock`: regenerated (+7 transitive packages)
- `openhands-agent-server/openhands/agent_server/agent-server.spec`: use `collect_all(...)` for the Vertex SDK and its `google.cloud.*` namespace siblings (`vertexai`, `google.cloud.aiplatform`, `aiplatform_v1`, `aiplatform_v1beta1`, `bigquery`, `storage`, `resourcemanager`, `google.api_core`, `google.auth`, `google.rpc`, `google.genai`, `proto`, `grpc_status`). Also pin `google.rpc.status_pb2` explicitly (gRPC proto stub imported dynamically) and copy `google-cloud-aiplatform` metadata.

## Why `collect_all` and not just `collect_submodules`

I first tried the same minimal change as #2373 (just `pyproject.toml`), then added `collect_submodules("vertexai")` etc. to the spec. Neither worked at runtime — the bundled binary still failed with `ModuleNotFoundError: No module named 'vertexai'`, then with `No module named 'google.rpc.status_pb2'` after I hand-injected `vertexai`.

Root cause: `google.cloud.*` is a PEP-420 implicit namespace package shared across many distributions. PyInstaller's `collect_submodules("google.cloud.aiplatform")` doesn't traverse PEP-420 namespaces correctly — the package's `dist-info` ends up in the bundle (via `copy_metadata`), but the actual module files do not. `collect_all(...)` walks the installed package directory directly and pulls in modules + datas + binaries, which is what the bundle needs.

I verified by inspecting the extracted `_MEI*/` directory after running the freshly-built binary:
- `_MEI*/vertexai/__init__.py` ✅
- `_MEI*/google/cloud/aiplatform/` ✅
- `_MEI*/google/rpc/status_pb2.py` ✅

## Background

This supersedes #2373 (closed without merging earlier this year). That PR only touched `pyproject.toml`, which would have shipped the dep into the build env but not into the PyInstaller bundle.

## Checklist

- [x] If the PR is changing/adding functionality, are there tests to reflect this? — no new tests; this is a dependency + bundle config change. Existing tests pass.

## Test plan

- [ ] CI builds the `-python` image with `vertexai` actually present in the bundle (not just the dist-info)
- [ ] Smoke-test a completion against `vertex_ai/minimaxai/minimax-m2-maas` (or another partner MaaS model) in a deployment with Vertex credentials


neubig:
@openhands audit the packages in this PR and make sure that they're appropriate (and don't, e.g. pose supply chain risks). Then consider the increase in total package size based on this. We may want to consider accepting this PR as-is, or we may want to make this an optional dependency. In the latter case, propose an implementation that would allow us to do that.

neubig:
I audited the dependency changes in this PR.

### Packages added / changed

New packages in `uv.lock`:
- `google-cloud-aiplatform==1.147.0`
- `google-cloud-bigquery==3.41.0`
- `google-cloud-core==2.5.1`
- `google-cloud-resource-manager==1.17.0`
- `google-cloud-storage==3.4.1`
- `google-crc32c==1.8.0`
- `google-resumable-media==2.8.2`
- `grpc-google-iam-v1==0.14.4`
- `grpcio-status==1.76.0`

Existing dependency changes pulled in by the resolver:
- `google-auth 2.41.1 -> 2.49.2`
- `google-genai 1.52.0 -> 1.73.0`
- `rsa` dropped from the lockfile

### Supply-chain / package audit

I don't see obvious supply-chain red flags here:
- the new packages are all in the well-known Google / googleapis / gRPC package families
- the installed metadata points back to Google-maintained repos / publisher emails (`googleapis-packages@google.com`, `googleapis-publisher@google.com`, etc.) or the gRPC project
- licenses are Apache 2.0 across the set (except `google-crc32c`, which doesn't surface the string in `pip show`, but still points at the official `googleapis/python-crc32c` repo)

I also ran a known-vulnerability check against the exact versions added/updated by this PR:

```bash
uvx --with pip-audit pip-audit -r /tmp/pr2910-reqs.txt
```

That came back clean.

So from a provenance / vuln perspective, this looks acceptable to me. The real tradeoff is footprint, not package trust.

### Size impact

I measured this two ways.

**Python install footprint (exact wheel versions in this PR):**
- about **+10.6 MiB** of compressed wheel downloads net
- about **+92 MiB** of installed `site-packages` footprint net
- the dominant contributor is `google-cloud-aiplatform` itself at about **8.0 MiB compressed / 80.3 MiB installed**

**PyInstaller binary footprint:**
- `main`: `dist/openhands-agent-server` = **86 MiB**
- this branch: `dist/openhands-agent-server` = **148 MiB**
- delta: about **+62 MiB** (~**72%** larger)

So this is a fairly meaningful size increase for a provider-specific feature.

### Recommendation

My take:
- **security/supply-chain:** acceptable as-is
- **size/footprint:** large enough that I would lean toward making it **optional for the SDK** unless we explicitly want Vertex partner-model support in every default install

### If we make it optional, here's the implementation I'd suggest

This can follow the existing `boto3` pattern pretty closely:

1. Move the dependency out of the base SDK deps and into an extra, e.g.
   ```toml
   [project.optional-dependencies]
   boto3 = ["boto3>=1.35.0"]
   vertex = ["google-cloud-aiplatform>=1.38"]
   ```

2. Add a small preflight in the LLM transport path:
   - if `infer_litellm_provider(...) == "vertex_ai"`
   - and `importlib.util.find_spec("vertexai") is None`
   - raise a friendly `LLMBadRequestError` (or similar) with an install hint like:
     ```
     Vertex AI partner models require the Vertex SDK. Install with: openhands-sdk[vertex]
     ```
   That gives users a clearer error than LiteLLM's raw import failure.

3. Make the PyInstaller Vertex collection conditional:
   - only run the `collect_all(...)` block when `vertexai` is actually installed
   - that way local builds without the extra still succeed

4. For the **published agent-server image / binary**, choose one of two policies:
   - **bundle it anyway** by installing `--extra vertex` during the server build, so the runtime remains full-featured while the Python SDK stays lean, or
   - **add a separate `vertex` build variant/tag** if we also want the default runtime artifact smaller

Given the numbers above, my preference would be:
- **optional in `openhands-sdk`**
- then separately decide whether the default agent-server artifact should still include the extra

That keeps the base Python install much leaner while preserving a straightforward path for users who do need Vertex partner models.

_AI note: this comment was generated by an AI assistant (OpenHands) on behalf of the user._

neubig (commented):
Hey @abguymon , thanks for the contribution! Could you please take a look at the comments above and add these as optional dependencies instead of main dependencies, then re-request my review?

abguymon:
Thanks for the audit @neubig — agreed on the size tradeoff. I've pushed a follow-up commit that makes Vertex an optional extra along the lines you suggested:

- **`openhands-sdk/pyproject.toml`**: `google-cloud-aiplatform` moved out of base deps and into `[project.optional-dependencies]` as `vertex` (mirrors the existing `boto3` extra).
- **Preflight**: in the LLM transport path, when the inferred provider is `vertex_ai` and `vertexai` is not importable, we now raise `LLMBadRequestError` with the hint `pip install "openhands-sdk[vertex]"` instead of letting LiteLLM crash with a low-level `ModuleNotFoundError`. New unit tests cover both the raise and no-op paths.
- **PyInstaller spec**: the Vertex `collect_all` block (and the `google.rpc.status_pb2` pin) is now guarded on `importlib.util.find_spec("vertexai")`, so local builds without the extra still succeed.
- **Agent-server Dockerfile**: both `uv sync` invocations gained `--extra vertex`, so the published agent-server binary continues to bundle the Vertex SDK by default — i.e. base SDK install stays lean, runtime artifact stays full-featured.
- Rebased onto `main` to clear the conflict; `uv.lock` regenerated.

Re-requesting your review.

openhands-agent-server/openhands/agent_server/docker/Dockerfile
```diff
@@ -50,7 +50,7 @@ COPY --chown=${USERNAME}:${USERNAME} openhands-agent-server ./openhands-agent-se
 RUN --mount=type=cache,target=/home/${USERNAME}/.cache,uid=${UID},gid=${GID} \
     uv python install 3.13 && \
     uv venv --python-preference only-managed --python 3.13 .venv && \
-    uv sync --frozen --no-editable --managed-python --extra boto3 && \
+    uv sync --frozen --no-editable --managed-python --extra boto3 --extra vertex && \
```
neubig:
```suggestion
    uv sync --frozen --no-editable --managed-python --extra boto3 && \
```

I don't think we want this on by default. If there's a way to make it optional that could be acceptable, although I'm not sure how to do that.

openhands-agent-server/openhands/agent_server/docker/Dockerfile
```diff
@@ -62,7 +62,7 @@ ARG USERNAME UID GID
 
 # We need --dev for pyinstaller
 RUN --mount=type=cache,target=/home/${USERNAME}/.cache,uid=${UID},gid=${GID} \
-    uv sync --frozen --dev --no-editable --extra boto3
+    uv sync --frozen --dev --no-editable --extra boto3 --extra vertex
```
neubig: ditto

openhands-agent-server/openhands/agent_server/agent-server.spec
neubig, line 1: Similarly, I'd prefer to not have that on by default.

neubig (commented):
Thanks, overall this looks very solid but I think we'd like to not bake this in to the default image, as we'd like to keep it lean. I think we could consider making a separate docker file with more of the optional dependencies backed in though.

abguymon:
Thanks @neubig — pushed two follow-up commits ([4ba82047](https://github.com/OpenHands/software-agent-sdk/pull/2910/commits/4ba82047), [1490482d](https://github.com/OpenHands/software-agent-sdk/pull/2910/commits/1490482d)):

- **Default image stays lean.** Dropped `--extra vertex` from both `uv sync` calls in `agent-server/docker/Dockerfile`. The spec's `if _VERTEX_AVAILABLE:` guard from the previous round means the default build now naturally skips the Vertex bundle (and prints a one-line note saying so), without any spec-level changes.
- **Opt-in via build arg.** Rather than a parallel Dockerfile, I added a single `ENABLE_VERTEX=0` build arg threaded into both stages. To produce a Vertex-capable binary:

  ```
  docker build --build-arg ENABLE_VERTEX=1 ...
  ```

  Equivalent capability to a separate Dockerfile, but no parallel file to keep in sync, no extra CI job, no extra image tag — and it leaves room to add more `ENABLE_*` switches the same way later if other extras grow this need. Spec header comment now documents both Docker (`--build-arg`) and from-source (`--extra vertex` before pyinstaller) opt-in paths.

If you'd actually prefer a separate `Dockerfile.full` (or similar) over the build arg, happy to reshape it — let me know. Otherwise, ready for another look.

openhands-agent-server/openhands/agent_server/docker/Dockerfile
abguymon: Addressed in 4ba82047 — dropped `--extra vertex` from this line. The opt-in path is now `docker build --build-arg ENABLE_VERTEX=1 ...` (added in 1490482d), which threads through to this same `uv sync` call.

openhands-agent-server/openhands/agent_server/docker/Dockerfile
abguymon: Addressed in 4ba82047 — dropped `--extra vertex` here too. `ENABLE_VERTEX=1` flips this same conditional in the binary-builder stage.

openhands-agent-server/openhands/agent_server/agent-server.spec
abguymon: Addressed — the spec is now off-by-default. The `if _VERTEX_AVAILABLE:` guard skips `collect_all(...)` entirely when `vertexai` isn't importable in the build env, which is now the default (since the Dockerfile no longer installs the extra). The header comment documents the two opt-in paths: `docker build --build-arg ENABLE_VERTEX=1` for image builds, or `uv sync ... --extra vertex` before pyinstaller for from-source builds.

abguymon:
@neubig I believe this PR is ready to merge

neubig:
I could not push to the original fork branch (`trustle:adam/install-aiplatform-for-vertex-ai`; GitHub returned 403), so I pushed the resolved branch to this repository and opened replacement PR #3776: https://github.com/OpenHands/software-agent-sdk/pull/3776

That branch merges current `main`, resolves the conflict in the LLM transport path, keeps Vertex as an optional extra, applies the useful spec readability feedback, and regenerates `uv.lock` from current `main` so the unrelated `pypdf`/`lxml` downgrade concern is gone.

neubig:
Hi @abguymon , thanks a bunch for this contribution and so sorry that it took so long for me to land it! I opened a new PR to fix the merge conflicts, keeping your history, and just merged it in now

Issue #2904, opened 2026-04-21
Responses API: `prompt_cache_key` never sent; cache-shard routing causes oscillating cache-miss per turn


VascoSch92:
## Summary

The SDK never sends a [`prompt_cache_key`](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-key) (or `safety_identifier`, or `previous_response_id`) on OpenAI Responses API calls. With LiteLLM proxy load-balancing across OpenAI backends, each call can land on a different cache shard, so the growing conversation prefix — which OpenAI's prefix cache *could* absorb — is effectively re-billed at the fresh-input rate on most turns.

Observed on an OH+gpt-5.4 SWE-bench eval run (SDK commit `d94cdf8d`, 500 instances): **cache hit rate 38.8%**, total cost **\$678.58** vs **\$197.86** for a codex ACP run on the same 497 instances with the same model. ~\$481 of the gap is attributable to cache behavior.

## Evidence

Per-call `cache_read_tokens` for one representative 19-turn instance:

| turn | total_in | cache_read | turn | total_in | cache_read |
|---:|---:|---:|---:|---:|---:|
| 1 | 13,373 | **6,016** | 11 | 30,160 | **6,016** |
| 2 | 15,020 | **6,016** | **12** | **25,594** | **0** |
| 3 | 16,344 | **6,016** | 13 | 56,679 | 23,168 |
| 4 | 25,730 | 10,624 | 14 | 58,208 | 22,528 |
| 5 | 27,524 | **6,016** | 15 | 73,210 | 35,328 |
| 6 | 27,983 | **6,016** | 16 | 53,645 | 14,848 |
| 7 | 44,602 | 22,144 | 17 | 45,080 | **6,016** |
| 8 | 29,049 | **6,016** | 18 | 45,449 | **6,016** |
| 9 | 29,294 | **6,016** | 19 | 64,733 | 24,704 |
| 10 | 45,570 | 21,760 | | | |

- **6,016** recurs (40.1% of 15,927 calls across 467 instances). That's the static `instructions + tools` bundle (~12,581 + ~12,843 chars ≈ 6,356 tokens), which is small enough to be re-cached by any backend.
- Larger hits (20k–35k) are opportunistic — they happen only when the load balancer happens to route a call to the same backend that served the previous call.
- Turn 12's `cache_read=0` is a fully cold backend (no cache hit at all).

[...]

PR #2920, opened 2026-04-22
fix(conversation): repin prompt_cache_key on switch_profile

### LINKED ISSUE #2918: switch_profile() drops prompt_cache_key; restore works today but needs explicit coverage
## Summary

I verified the `prompt_cache_key` behavior when the conversation's LLM changes at runtime and when a conversation is restored, following up on PR #2907.

### 1) `switch_profile()` currently drops the cache key

`LocalConversation.switch_profile()` loads an `LLM` from `LLMProfileStore` / `LLMRegistry` and replaces `self.agent.llm` with it, but it does **not** repin `_prompt_cache_key` afterward.

Relevant path:

- `LocalConversation.__init__` pins the key once with `self.agent.llm._prompt_cache_key = str(self._state.id)`
- `LocalConversation.switch_profile()` later does:
  - `new_llm = self._profile_store.load(...)` or `self.llm_registry.get(...)`
  - `self.agent = self.agent.model_copy(update={"llm": new_llm})`
- No follow-up step restores the conversation's cache key onto `new_llm`

I verified that locally:

- conversation created with ID/key `8d6fd5db-ded9-4d56-8c9e-db0f920388cf`
- after `switch_profile("fast")`, `conv.agent.llm._prompt_cache_key` became `None`
- after `switch_profile("slow")`, it stayed `None`
- switching back to the cached registry entry still left it `None`

That means a mid-conversation model switch reintroduces #2904 for all later calls in that conversation.

This also affects the agent-server route because `/switch_profile` delegates to the same `LocalConversation.switch_profile()` implementation.

### 2) Restore is okay today for the primary `agent.llm`

The restore path looks correct for the main conversation LLM.

`ConversationState.create()` resumes the persisted state, then `LocalConversation.__init__` runs and repins `self.agent.llm._prompt_cache_key = str(self._state.id)` on the runtime agent that was passed in.

I verified that locally with a restored conversation using a different runtime model:

- original conversation ID/key: `7aa92855-6753-40bb-9dbf-11dcb822d959`
- restored conversation ID stayed `7aa92855-6753-40bb-9dbf-11dcb822d959`
- restored runtime agent LLM key was repinned to `7aa92855-6753-40bb-9dbf-11dcb822d959`

So restore itself preserves the one-conversation/one-cache-key mapping for the primary agent LLM.

## Why I still think this should be tracked

Right now restore works because the constructor side effect happens to repin the active `agent.llm`. But the logic is not centralized, and the supported runtime replacement path (`switch_profile`) already bypasses it.

So the current behavior is:

- create conversation: ✅ pinned
- restore conversation: ✅ pinned
- switch LLM mid-conversation: ❌ key dropped
- restore, then switch LLM: ❌ key dropped after the switch

## Recommendation

I recommend centralizing prompt-cache-key ownership at the conversation layer, then reusing that helper whenever the active LLM changes.

Concretely:

1. Add a small helper such as "repin the active conversation LLM to `str(self._state.id)`".
2. Call it from:
   - `LocalConversation.__init__`
   - `LocalConversation.switch_profile()`
   - any future API that replaces `agent.llm` at runtime
3. Add regression tests for:
   - `switch_profile()` preserves the same key before any run
   - `switch_profile()` preserves the same key after a conversation has history
   - restore with a different runtime model keeps the same conversation key
   - restore + `switch_profile()` still preserves the key

_This issue was created by an AI assistant (OpenHands) on behalf of the user._

VascoSch92:
## What                                                                                                                                                                               

  Centralize _prompt_cache_key pinning in LocalConversation via a new _pin_prompt_cache_key() helper, called from both __init__ and switch_profile().                                
   
## Why                                                                                                                                                                                
                  
Fixes #2918. switch_profile() replaced self.agent.llm via model_copy() without repinning the key, so _prompt_cache_key became None on every mid-conversation LLM switch, reintroducing #2904 for all subsequent calls in that conversation (including via the agent-server /switch_profile route).
                                                                                                                                                                                     
## Changes         
- local_conversation.py: extract pin logic into _pin_prompt_cache_key(); call it from __init__ and from switch_profile() after the agent LLM swap.                                 
- tests/sdk/conversation/test_switch_model.py: add test_switch_profile_preserves_prompt_cache_key covering initial pin, fast→slow switch, and switching back to a cached registry entry.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:f4e8817-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-f4e8817-python \
  ghcr.io/openhands/agent-server:f4e8817-python
```

```
ghcr.io/openhands/agent-server:f4e8817-golang-amd64
ghcr.io/openhands/agent-server:f4e8817-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:f4e8817-golang-arm64
ghcr.io/openhands/agent-server:f4e8817-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:f4e8817-java-amd64
ghcr.io/openhands/agent-server:f4e8817-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:f4e8817-java-arm64
ghcr.io/openhands/agent-server:f4e8817-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:f4e8817-python-amd64
ghcr.io/openhands/agent-server:f4e8817-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:f4e8817-python-arm64
ghcr.io/openhands/agent-server:f4e8817-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:f4e8817-golang
ghcr.io/openhands/agent-server:f4e8817-java
ghcr.io/openhands/agent-server:f4e8817-python
```

- Each variant tag (e.g., `f4e8817-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `f4e8817-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -189,8 +189,7 @@ class LocalConversation(BaseConversation):
             tags=tags,
         )
 
-        # Pin the OpenAI prefix-cache shard to this conversation (#2904).
-        self.agent.llm._prompt_cache_key = str(self._state.id)
+        self._pin_prompt_cache_key()
 
         # Default callback: persist every event to state
         def _default_callback(e):
@@ -600,6 +599,10 @@ class LocalConversation(BaseConversation):
         """
         return not isinstance(self.agent, ACPAgent)
 
+    def _pin_prompt_cache_key(self) -> None:
+        # Pin the OpenAI prefix-cache shard to this conversation (#2904, #2918).
+        self.agent.llm._prompt_cache_key = str(self._state.id)
+
     def switch_profile(self, profile_name: str) -> None:
         """Switch the agent's LLM to a named profile.
 
@@ -623,6 +626,7 @@ class LocalConversation(BaseConversation):
         with self._state:
             self.agent = self.agent.model_copy(update={"llm": new_llm})
             self._state.agent = self.agent
+            self._pin_prompt_cache_key()
 
     @observe(name="conversation.send_message")
     def send_message(self, message: str | Message, sender: str | None = None) -> None:
--- a/tests/sdk/conversation/test_switch_model.py
+++ b/tests/sdk/conversation/test_switch_model.py
@@ -93,6 +93,23 @@ def test_switch_nonexistent_raises(profile_store):
     assert conv.state.agent.llm.model == "default-model"
 
 
+def test_switch_profile_preserves_prompt_cache_key(profile_store):
+    """Regression test for #2918: switch_profile must repin _prompt_cache_key."""
+    conv = _make_conversation()
+    expected = str(conv.id)
+    assert conv.agent.llm._prompt_cache_key == expected
+
+    conv.switch_profile("fast")
+    assert conv.agent.llm._prompt_cache_key == expected
+
+    conv.switch_profile("slow")
+    assert conv.agent.llm._prompt_cache_key == expected
+
+    # Switching back to a cached registry entry must still carry the key.
+    conv.switch_profile("fast")
+    assert conv.agent.llm._prompt_cache_key == expected
+
+
 def test_switch_then_send_message(profile_store):
     """switch_profile followed by send_message doesn't crash on registry collision."""
     conv = _make_conversation()


VascoSch92:
@enyst If for you is ok, I will not re-run the evals, as we know that the prompt_cache_key and the mechanism which come together is working because of last PRs

enyst (approved):
Yes, thank you!

PR #2923, opened 2026-04-22
fix(conversation): prevent fork(agent=...) from clobbering source prompt_cache_key

### LINKED ISSUE #2917: LocalConversation.fork(agent=...) can clobber the source prompt_cache_key
## Summary

I verified the `conversation.fork()` behavior for the `prompt_cache_key` change in PR #2907.

### What works today

The default fork path is correct:

- `LocalConversation.fork()` generates a **new** conversation ID when `conversation_id` is omitted.
- It builds the fork via `LocalConversation(...)`, which repins `agent.llm._prompt_cache_key` to the fork's new `state.id`.
- The source conversation keeps its original key.
- The agent-server / remote fork path uses this same default behavior server-side, so normal remote forks inherit the same good behavior.

I verified this locally on branch `2904-responses-api-prompt_cache_key-never-sent-cache-shard-routing-causes-oscillating-cache-miss-per-turn`:

- source ID/key: `41176d21-79e0-412a-b974-02a87b41dd47`
- fork ID/key: `2539d579-9efc-4a8e-998e-892246b9b294`
- source key after fork remained `41176d21-79e0-412a-b974-02a87b41dd47`

### The footgun

`LocalConversation.fork(agent=...)` can overwrite the **source** conversation's cache key if the caller passes an agent that aliases the source agent or shares the same `LLM` object.

Why:

- `LocalConversation.__init__` mutates the provided `LLM` in place with:
  - `self.agent.llm._prompt_cache_key = str(self._state.id)`
- `fork(agent=...)` uses the supplied agent object directly.
- So `conversation.fork(agent=conversation.agent)` repins the shared `LLM` to the fork ID and clobbers the source conversation's key.

I verified that locally too:

- source ID/key before: `41176d21-79e0-412a-b974-02a87b41dd47`
- `fork(agent=conv.agent)` created fork ID/key: `398100de-cf1f-4029-b758-ab5755de9a8a`
- source key after the fork also became `398100de-cf1f-4029-b758-ab5755de9a8a`

That means both conversations can now send requests with the fork's cache key, which defeats the intended one-conversation/one-cache-shard mapping.

## Recommendation

I recommend making `fork(agent=...)` defensive against aliasing:

1. **Preferred:** clone the supplied agent before pinning if it shares identity with the source agent or shares the same `llm` object.
2. **Acceptable alternative:** reject aliased/shared agents with a clear `ValueError`.
3. Add regression tests for:
   - default `fork()` keeps distinct keys for source and fork
   - `fork(agent=source.agent)` cannot silently clobber the source key

The default no-arg `fork()` behavior already looks good; the risk is specifically the customizable `agent=` path.

_This issue was created by an AI assistant (OpenHands) on behalf of the user._

VascoSch92:
## Summary      

Fixes #2917.

`LocalConversation.fork(agent=...)` could silently overwrite the **source** conversation's `prompt_cache_key` whenever the caller passed an agent that aliased the source agent or shared the source's `LLM` instance. That defeats the one-conversation / one-cache-shard invariant introduced in #2907 and causes the source and the fork to send requests with the same cache key, the exact condition #2907 was designed to prevent.

## Root cause

`LocalConversation.__init__` pins the cache key by **mutating the agent's LLM in place**:                                                                                          
   
```python                                                                                                                                                                          
  self.agent.llm._prompt_cache_key = str(self._state.id)
```

`fork()` had two branches for selecting the fork's agent:                                                                                                                          1. **No `agent=` supplied** → round-trip the source agent through `model_dump` / `model_validate` to produce an independent copy. Safe.                                            
2. **`agent=` supplied** → use the caller's object **by reference**. Unsafe: if the object (or its `llm`) is the source's, the fork's `__init__` repins the source's LLM.

Reporter's verification, reproduced locally:                                                                                                                                       
- source key before fork: `41176d21-79e0-412a-b974-02a87b41dd47`                                                                                                                   
- `fork(agent=conv.agent)` → fork key: `398100de-cf1f-4029-b758-ab5755de9a8a`                                                                                                      
- source key **after** fork: `398100de-cf1f-4029-b758-ab5755de9a8a` ← clobbered                                                                                                    
                                                                                                                                                                                     
## Why collapse the two paths (rather than fix only path 2)?                                                                                                                       
                                                                                                                                                                                     
The two paths existed for a real reason, so it's worth explaining the decision.                                                                                                    
                  
**Path 1 (default)** deep-copies the source agent via a JSON round-trip instead of `copy.deepcopy` / `model_copy(deep=True)`, because agent/LLM private attrs hold `threading.Lock` objects that aren't picklable (the original PR #2841 calls this out explicitly).

**Path 2 (`agent=...`)** was an *optimization* for the A/B-test / tool-swap use case (see `examples/01_standalone_sdk/48_conversation_fork.py`): the caller constructs a **brand-new** `Agent` with a **brand-new** `LLM` and hands it in, so copying would be wasteful.

The footgun is that path 2 **assumes** the supplied agent is fresh and non-aliased — but that assumption is never checked. It breaks silently when a caller:                       
  - passes `source.agent` directly (the issue's exact scenario), or
  - builds a new `Agent(llm=source.agent.llm, tools=...)` reusing the existing LLM — a natural pattern if you only want to swap tools but keep model config.

## Behavior change worth flagging

Before: `fork(agent=caller_agent).agent is caller_agent` was `True`. After: it's `False` — the fork receives a reconstituted clone. Callers who held a reference to `caller_agent` and mutated it post-fork expecting the fork to see the change will now silently diverge. This is not documented as guaranteed anywhere, the shipped example doesn't rely on it, and it matches the mental model that a fork is an *independent* conversation — but calling it out for reviewers.


---

**Variants & Base Images**
|---|---|---|---|


```bash
docker pull ghcr.io/openhands/agent-server:234be48-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-234be48-python \
  ghcr.io/openhands/agent-server:234be48-python
```

```
ghcr.io/openhands/agent-server:234be48-golang-amd64
ghcr.io/openhands/agent-server:234be48-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:234be48-golang-arm64
ghcr.io/openhands/agent-server:234be48-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:234be48-java-amd64
ghcr.io/openhands/agent-server:234be48-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:234be48-java-arm64
ghcr.io/openhands/agent-server:234be48-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:234be48-python-amd64
ghcr.io/openhands/agent-server:234be48-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:234be48-python-arm64
ghcr.io/openhands/agent-server:234be48-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:234be48-golang
ghcr.io/openhands/agent-server:234be48-java
ghcr.io/openhands/agent-server:234be48-python
```

- Each variant tag (e.g., `234be48-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `234be48-python-amd64`) are also available if needed


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -340,15 +340,17 @@ def fork(
...
+        # Always deep-copy the agent (supplied or source) so the fork owns
+        # its own object graph. Required because __init__ mutates
+        # agent.llm._prompt_cache_key in place (#2917): a shared/aliased
+        # agent would clobber the source conversation's cache key.
+        # Round-trip via JSON avoids thread-lock pickling issues with
+        # model_copy(deep=True).
+        source_agent = agent if agent is not None else self.agent
+        agent_cls = type(source_agent)
+        fork_agent = agent_cls.model_validate(
+            source_agent.model_dump(context={"expose_secrets": True}),
```
enyst, line 353: What if we have something like
- `Agent.agent_from(source_agent: Agent)`
- documented in AGENTS.md, in docstrings etc
- all occurrences of model_copy and roundtrips like this use that
- workflows enforce the limits: no model_copy or this syntax for Agent anywhere?

What would be your thoughts, alternatives, ...rotten tomatoes? (my dumb ideas attract rotten tomatoes...)

VascoSch92: @enyst 
Not rotten tomatoes at all, the idea has merit! 

The Pattern Appears in 3-4 Places in the codebase

The model_dump(context={"expose_secrets": True}) + model_validate deep-copy pattern is used in:
- local_conversation.py:fork() ← this PR
- remote_conversation.py:fork()
- event_service.py (agent-server restoration)

So a centralized Agent.deep_copy() or Agent.agent_from() would reduce duplication and document the threading.Lock pickle issue in one place.

But I will not block model_copy(update={...}), The codebase has ~5 legitimate shallow-copy patterns using model_copy(update={...}):
- switch_profile(): swapping the LLM
- Plugin loading: updating agent_context / mcp_config
- Task manager / delegate tool: disabling stream on sub-agent LLMs

These are intentionally shallow, they replace specific fields with new objects. Blocking all model_copy for Agent would break these valid patterns.

We can implement a minimal helper (not a method, to avoid expanding public API)

```python
# In openhands/sdk/agent/utils.py
def deep_copy_agent(agent: AgentBase) -> AgentBase:
    """Create an independent deep-copy of an agent.
    
    Uses JSON roundtrip to avoid threading.Lock pickling issues
    that occur with copy.deepcopy() or model_copy(deep=True).
    """
    agent_cls = type(agent)
    return agent_cls.model_validate(
        agent.model_dump(context={"expose_secrets": True}),
    )
```

Then update the 3-4 call sites to use it.

But I would do that in a separate PR. Is it ok for you? I can open an issue and then also open the relative PR if for you is ok :-)

enyst: I think maybe it's worth it to be public API actually... and maybe both:
- "This is the way to create a new Agent from another" -> `from_agent()` (class)
- "This is the way to mutate an Agent if you need to" -> `with_llm() / with_agent_context()` (instance probably)

Once documented everywhere, we can maybe pull the trigger 🤔 

Although I'm not sure this is best for mutation. The intention here is: 
- `conversation state` is mutable state
- Agent components, e.g. LLM are immutable or as immutable as possible.

If they need mutation, that should ideally happen in a single spot, which can trigger notifications, do logging, etc. and on which we can enforce workflows. Thinking both about humans, so it maybe helps with forgetting stuff like this; and of agents, so that we can enable continuous PR-Merge-PR sometime without them making a mess of things...

But maybe we'd need to look into details to figure out if that makes sense?

VascoSch92: @enyst can we do that in another PR? If yes, I will open an issue for that ;-)

VascoSch92: BTW, here is the issue: https://github.com/OpenHands/software-agent-sdk/issues/2935 :-)

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -340,15 +340,17 @@ def fork(
...
+        )
```
xingyaoww, line 354: @Openhands do we need to do this for the agent server also?

xingyaoww: No, the agent server does not need a separate fix — it is already protected.

**Analysis of the three fork paths:**

1. **`ConversationService.fork_conversation`** (agent-server, line 662–727): Delegates to `source_conversation.fork()` **without** passing `agent=`. It always takes the default path, which deep-copies the source agent via JSON roundtrip. This was safe before this PR, and the fix here makes the underlying `LocalConversation.fork()` safe for both paths regardless.

2. **`RemoteConversation.fork()`** (line 1309–1380): Explicitly **rejects** `agent=` with `NotImplementedError` (line 1341–1345) and always deep-copies `self.agent` client-side via the same JSON roundtrip pattern (line 1364–1367). Safe.

3. **`LocalConversation.fork(agent=some_agent)`** — the **only** vulnerable path, fixed by this PR.

The bug required two conditions: (a) an `agent` argument being passed to `fork()`, and (b) that agent sharing the sources `LLM` instance. The agent servers `ForkConversationRequest` model doesnt expose an `agent` field, so condition (a) was never met server-side.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
xingyaoww: Yeah, i think it make sense to get this in since this fixes some immediate bugs, and then do that in a subsequent PR :)

xingyaoww (approved):
Thanks

PR #2936, opened 2026-04-24
fix(state): warn when falling back to InMemoryFileStore without persistence_dir

### LINKED ISSUE #2915: InMemoryFileStore fallback is silent when persistence_dir is not provided
As noted by @VascoSch92 in #2903 (comment), `ConversationState.create()` silently falls back to `InMemoryFileStore` when neither `file_store` nor `persistence_dir` is provided (`state.py:314-319`). Users deploying to environments without persistent volumes (e.g. Cloud Run) get in-memory storage with no warning — events are lost between requests and there's no indication why.

**Proposal:** add a `logger.warning` when the `InMemoryFileStore` fallback is triggered, so the misconfiguration is visible in logs.

Happy to open a PR if this direction makes sense.

StatPan:
Closes #2915

## What

When `persistence_dir` is not provided, `ConversationState` silently falls back to `InMemoryFileStore`. Users deploying to environments without persistent volumes (e.g. Cloud Run) get no indication that their EventLog data won't persist.

## Fix

Added `logger.warning()` before constructing the fallback `InMemoryFileStore`, so the misconfiguration is immediately visible in logs.

## Test

Added `test_no_persistence_dir_logs_warning` that verifies the warning message appears when no `persistence_dir` is provided.

cc @VascoSch92

### DIFF
--- a/tests/sdk/conversation/test_directories.py
+++ b/tests/sdk/conversation/test_directories.py
@@ -1,5 +1,6 @@
 """Tests for conversation directory handling."""
 
+import logging
 import os
 import tempfile
 import uuid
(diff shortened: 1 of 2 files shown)


tests/sdk/conversation/test_directories.py
```diff
@@ -140,3 +141,23 @@ def test_conversation_factory_persistence_dir_only(mock_agent):
         # persistence_dir should include conversation ID subdirectory
         expected_dir = os.path.join(persistence_dir, conversation.state.id.hex)
         assert conversation.state.persistence_dir == expected_dir
+
+
+def test_no_persistence_dir_logs_warning(mock_agent, caplog):
+    """Test that a warning is logged when no persistence_dir is provided."""
+    with tempfile.TemporaryDirectory() as temp_dir:
+        working_dir = os.path.join(temp_dir, "work")
```
VascoSch92: use `pathlib` here

tests/sdk/conversation/test_directories.py
```diff
@@ -140,3 +141,23 @@ def test_conversation_factory_persistence_dir_only(mock_agent):
...
+        os.makedirs(working_dir)
```
VascoSch92: here too

VascoSch92 (approved):
just a couple of nit

tests/sdk/conversation/test_directories.py
StatPan: Fixed — switched to `Path(temp_dir) / "work"` and `working_dir.mkdir()`.

tests/sdk/conversation/test_directories.py
StatPan: Same fix — using `pathlib` throughout.

Issue #1210, opened 2025-11-20
Proposal: separate Windows code and dogfooding


enyst:
We have a few PRs proposing to add Windows support:
- https://github.com/OpenHands/software-agent-sdk/pull/1014
- https://github.com/OpenHands/software-agent-sdk/pull/1209
- https://github.com/OpenHands/software-agent-sdk/pull/1012

**Proposal:**

I think maybe we could we consider implementing this separately, separate from core agent-sdk code. In fact, there are two things here that I feel we could consider:
- separate code / loose coupling
- dogfooding / dedicated maintainer(s).

**1. Loose coupling:**

- maybe Windows-specific tools
- maybe Windows-specific package
- or better abstractions, at least separate files, initialized per platform
    - if we implement it directly into core.

I'm concerned about code like conditionals with Windows-specific logic in core implementation. Windows is _too different_ IMHO, which means we can expect much more special cases, much more `ifs` in random places. I do feel Windows support is a much more demanding feature than the current PRs, and I'm concerned that:
- the agent-sdk codebase is becoming more error-prone and harder to understand
- we're starting something we are not able to maintain.

IMHO we could perhaps think how to design and implement Windows support as _separated_ as possible from the rest of the logic... The [WindowsTerminal](https://github.com/OpenHands/software-agent-sdk/blob/c40789b3b36ea638ad3ac6263c340c63cabefb8f/openhands-tools/openhands/tools/terminal/terminal/windows_terminal.py) file is a good example: abstraction with platform-specific import. But the `if` for [a tool prompt](https://github.com/OpenHands/software-agent-sdk/pull/1012/files#diff-651a11fcfb97c6e524bcc0155c3140f900ae50c1adba4b70a343e9361c0ede0eR223) could be done differently. It's a prompt piece, we could have, for example, different templates/files. Or even subclass the definition, or maybe separate a component with the prompt, and subclass that.

**2. Dogfooding:**

As I noted in one of these PR, one detail here that seems relevant to me is that usually, it won't be the same people that work regularly on Windows and Linux/Mac. This matters at multiple levels:
- for one, no maintainer is working on Windows. We could test a PR (on a quick Windows), but how deep is that testing? Are we introducing support for something we don't maintain?
- this also means that when any contributor looks at a screen of such code in their editor, they never have at their disposal, in their sight, a screen of code, they have _half_ the screen. This may sound trivial, it's true, but if it's too many times, it affects ease of debugging/readability/maintainability IMHO.

[...]

PR #2983, opened 2026-04-27
Add Windows test support

neubig:
## Why
The SDK still has major Windows gaps. This PR is a first pass at making the test suite and non-terminal components behave more cleanly on Windows.

## Summary
- removes a number of Unix-specific assumptions from tests and helper code
- normalizes path and file-handling behavior that breaks on Windows
- adds targeted guards where functionality is still Unix-only

## Out of scope
- full Windows terminal support
- complete removal of the new Windows skips

## Follow-up issues
- #2987 — implement a real Windows terminal backend and PowerShell-aware behavior
- #2986 — remove terminal-related Windows skipifs added in this PR
- #2988 — remove non-terminal Unix shell assumptions exposed by Windows in this PR
- #2989 — audit non-terminal Windows OS capability gaps (symlinks, POSIX perms, fork/resource)

## Prior work / suggested follow-up
There is prior design and implementation work for Windows terminal support in #1210 and #1012.
A good next step would be to build on that earlier implementation rather than restart from scratch, and preserve as much of the original commit history as practical so the earlier contributor receives proper credit.

## Windows QA Notes
- Added and fixed a regression test for Windows cp1252/legacy-console visualizer output. The test reproduces the original `UnicodeEncodeError` with non-ASCII model text and now passes.
- Updated the example test runner so Windows subprocesses write UTF-8 (`PYTHONIOENCODING=utf-8`) and pytest decodes captured stdout/stderr as UTF-8 with replacement. This is needed because LLM/example output can contain Unicode symbols while Windows pipes otherwise default to the active code page.
- Updated `t05_simple_browsing` setup to start the local HTTP server with `sys.executable -m http.server 8000 --bind 127.0.0.1` instead of `python3`, so the integration server starts and is cleaned up reliably on Windows.

Local verification run on Windows:
- `uv run pytest tests/sdk/conversation/test_visualizer.py -q` -> `20 passed`
- `uv run pytest tests/examples/test_examples.py -q -k test_directory_example_is_discovered` -> `1 passed, 52 deselected`
- `uv run pytest tests/integration/test_tool_presets.py tests/integration/test_early_stopper.py -q` -> `21 passed, 1 skipped`
- `uv run pre-commit run --files ...` was run on each touched file and passed.

Stack/CI status after these fixes:
- #2983, #2990, #2991, and #2993 all reported green GitHub checks after the stack was updated.

_This PR description was updated by an AI agent (OpenHands) on behalf of the user._

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:e184655-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-e184655-python \
  ghcr.io/openhands/agent-server:e184655-python
```

```
ghcr.io/openhands/agent-server:e184655-golang-amd64
ghcr.io/openhands/agent-server:e184655-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:e184655-golang-arm64
ghcr.io/openhands/agent-server:e184655-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:e184655-java-amd64
ghcr.io/openhands/agent-server:e184655-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:e184655-java-arm64
ghcr.io/openhands/agent-server:e184655-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:e184655-python-amd64
ghcr.io/openhands/agent-server:e184655-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:e184655-python-arm64
ghcr.io/openhands/agent-server:e184655-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:e184655-golang
ghcr.io/openhands/agent-server:e184655-java
ghcr.io/openhands/agent-server:e184655-python
```

- Each variant tag (e.g., `e184655-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `e184655-python-amd64`) are also available if needed


neubig:
@openhands group all the fixes in this PR into different categories, and critique them: are they necessary and best practices for windows Python development, or could they be implemented in a simpler or more pythonic way?

neubig:
Grouped roughly, I see **five categories** of changes in this PR:

### 1. Path normalization and cross-platform path semantics
**Examples:** `GitChange.path` JSON serialization via `as_posix()`, `remote_workspace_mixin`, skill/plugin metadata loading, `event_store`, `workspace/cloud/repo.py`, various tests switching from `str(path)` to `path.as_posix()`.

**My take:**
- **Necessary:** yes. Anything that is persisted, sent over an API, or shown to an agent should have a **canonical path representation** that does not depend on the host OS.
- **Best practice:** mostly yes. Using `Path.as_posix()` for **wire/storage/display formats** is the right direction.
- **Could be simpler/more pythonic:** also yes. Right now the PR mixes good `pathlib` usage with a lot of ad hoc `replace("\\", "/")`. I’d prefer a small shared helper for “portable path for serialization/display” and use that everywhere. That would make the intent clearer and reduce the chance of slightly different normalization rules in different files.
- **One caution:** some of the “treat backslash as absolute/local path” logic is pragmatic, but a bit loose. For example, `"\\foo"` on Windows is not the same thing as a fully qualified `C:\\foo`. I’d rather centralize that logic and make it explicit.

### 2. Removing Unix shell assumptions from runtime code
**Examples:** `FileEditor` replacing `find` shell-outs with pure Python directory walking, hook cleanup using `taskkill` on Windows, browser Chromium detection adding Windows locations and `LOCALAPPDATA`, `SIGALRM` guard in `.github/run-eval/resolve_model_config.py`.

**My take:**
- **Necessary:** yes. These are real runtime portability fixes, not just test hacks.
- **Best practice:**
  - `FileEditor`: **strong yes**. Replacing `find` with Python traversal is both **more portable and more pythonic**.
  - `SIGALRM` guard: **correct and minimal**.
  - `taskkill` for async hook cleanup: **reasonable**. On Windows, killing the whole process tree is the important part, and `taskkill /T` is a pragmatic way to get that behavior.
- **Could be simpler/more pythonic:**
  - Browser detection now seems split between the generic executor and the Windows-specific executor. That feels a bit redundant. I’d prefer one place that enumerates Chromium candidates by platform.
  - The PR removes Edge from the Windows-specific browser lookup. That may be intentional, but if the goal is “Chromium-based browser available on Windows”, removing Edge support is at least worth justifying.

### 3. Rewriting tests to be OS-agnostic instead of shell-specific
**Examples:** hook tests, terminal service tests, skill command tests, browser E2E server setup, replacing `echo`/`sleep`/`pwd`/`touch`/bash scripts with `sys.executable -c ...`, swapping `TerminalTool` out of tests that are not actually testing terminal behavior.

**My take:**
- **Necessary:** yes, and this is one of the better parts of the PR.
- **Best practice:** mostly yes. For tests, using **Python itself as the portable test substrate** is a good pattern. `sys.executable -c ...` is much more reliable than assuming `/bin/sh`, `sleep`, `touch`, `cat`, `pwd`, etc.
- **Also good:** replacing `TerminalTool` with `TaskTrackerTool` or `FileEditorTool` in tests that only care about persistence/loading semantics is a real improvement. Those tests were over-coupled to terminal availability.
- **Could be simpler:** the `_shell_join()` / `_python_command()` helpers are repeated across many test modules. I’d move those into a shared test utility module or fixture instead of duplicating them everywhere.

### 4. Explicitly skipping Unix-only functionality on Windows
**Examples:** many `pytest.mark.skipif(sys.platform == "win32")` additions around terminal backends, live bash endpoint tests, `pytest.mark.forked`, symlink tests, `resource` usage, etc.

**My take:**
- **Necessary in the short term:** yes, if the immediate goal is “get CI green on Windows without pretending Unix PTY/tmux support exists there.”
- **Best practice as a final state:** not really. These skips are honest, but they are still **admissions of unsupported functionality**, not support.
- **Could be cleaner:** I’d prefer a centralized capability model, e.g. shared markers/helpers like `requires_unix_pty`, `requires_symlink`, `requires_resource_module`, instead of many scattered platform checks.
- **High-level critique:** this PR is better described as **“make the test suite and non-terminal components Windows-tolerant”** than “Windows support” in the broad sense. The terminal stack is still clearly Unix-centric.

### 5. Miscellaneous portability / CI polish
**Examples:** explicit `encoding="utf-8"` on file reads/writes, removing emoji from CLI/check script output, using API-level upload/download verification instead of shell commands in live-server tests.

**My take:**
- **UTF-8 file opens:** **yes, best practice**. This is a classic Windows portability improvement.
- **API-level test verification instead of shelling out:** **good design**, independent of Windows.
- **Removing emoji from console output:** this one feels more debatable. It’s harmless and may avoid Windows console weirdness, but if the only issue was display/encoding, I’d rather fix the encoding boundary than change user-facing output unless we’ve actually seen that output break something.

## Overall critique
I think the PR contains a **good first-pass portability cleanup**, but it mixes together three different things:
1. **real runtime fixes** (good),
2. **test decoupling from Unix assumptions** (also good), and
3. **temporary skips for still-unsupported features** (acceptable, but should be called out as such).

If I were refining it further, my follow-up wishlist would be:
- **centralize path normalization** instead of sprinkling `replace("\\", "/")`,
- **centralize test command helpers** instead of repeating `_python_command()` everywhere,
- **centralize platform capability markers/skips**, and
- revisit the **browser detection split** / possible **Edge regression**.

So my bottom line is: **most of the runtime and test-decoupling changes are justified and pretty Pythonic; the scattered skipifs and some of the string-based path handling feel more like pragmatic patching than the final clean design.**

_This comment was generated by an AI agent (OpenHands) on behalf of the user._

neubig:
@neubig I grouped the fixes in this PR into four buckets and did a pass on whether they feel like the right long-term shape for Windows support:\n\n1. **Path and filesystem portability fixes**\n   - Examples: using /normalized path helpers, avoiding assumptions about , handling Windows path resolution in tests, and making file/resource-lock expectations depend on real absolute paths instead of Unix literals.\n   - **Assessment:** these are **necessary** and mostly **best practice**. This is the most Pythonic part of the PR: use , normalize at boundaries, and make tests assert semantics rather than a particular slash style.\n   - **Possible simplification:** a few tests still hardcode formatted path strings. Those are better expressed with shared helpers or /-based expectations so Windows differences are handled in one place.\n\n2. **Windows-specific test gating / capability checks**\n   - Examples: skipifs around terminal/browser behavior, symlink-related skips, and the new Windows CI job being narrower than the Linux suite.\n   - **Assessment:** this is **necessary as a first pass**, but it is **not the ideal final state**. These changes are pragmatic to get signal from Windows CI without pretending unsupported behavior already works.\n   - **More pythonic direction:** prefer **capability-based** checks over  when possible (for example: “symlinks available”, “interactive terminal backend available”, “browser e2e stable in this environment”). Centralizing those checks is cleaner than scattering platform conditionals. This is why I split follow-ups like **#2986** and **#2989**.\n\n3. **Shell / terminal / subprocess portability fixes**\n   - Examples: avoiding Unix-only shell commands, adding Windows-specific browser/terminal handling, and following up on places where tests or helpers assume Bash/TTY semantics.\n   - **Assessment:** these are **necessary**, but the current PR should ideally stop short of overfitting lots of small  branches.\n   - **Best-practice direction:** the cleaner shape is a **single abstraction boundary** for terminal behavior (shell quoting, environment export syntax, session startup, etc.) with a Windows backend instead of many scattered conditionals. That is what **#2987** and **#2988** are meant to separate from the initial CI-enabling work.\n\n4. **Workflow/CI adjustments to get useful Windows coverage**\n   - Examples: adding a  job, setting Windows-specific env, and scoping/ignoring tests that are still known-broken on Windows.\n   - **Assessment:** **necessary for now**, but definitely a **temporary shape** rather than the desired end state.\n   - **Simpler long-term goal:** grow the Windows job back toward parity as each follow-up lands, instead of keeping a permanently special-cased Windows matrix entry.\n\n### Overall critique\nI think the PR is directionally right, but it mixes three different things:\n- **real product portability fixes** (good, should stay),\n- **test portability fixes** (also good, but should be centralized more), and\n- **temporary exclusions/skipifs** (acceptable for bootstrapping CI, but should be paid down quickly).\n\nSo my take is:\n- **Keep** the path/filesystem fixes.\n- **Keep, but consolidate** the capability-based test helpers.\n- **Avoid** accumulating lots of ad hoc Windows branches in product code when a backend abstraction would do.\n- **Treat CI ignores/skips as debt**, not as the final Windows story.\n\nI’ve already split the follow-up work into the referenced issues so we can make #2983 a “get Windows into CI with honest expectations” PR, then land the cleaner backend/capability work in smaller stacked follow-ups.\n\n_This comment was created by an AI agent (OpenHands) on behalf of the user._

neubig:
Addressed the latest AI review feedback in `f2741e56` (grep hidden-file parity, remote git query-param paths, glob symlink preservation, and Windows workflow documentation). Could you take another look at the updated SHA?

openhands-sdk/openhands/sdk/workspace/remote/remote_workspace_mixin.py
```diff
@@ -8,12 +8,25 @@
...
+def _remote_path(path: str | Path) -> str:
+    return to_posix_path(path)
+
+
+def _join_remote_path(base: str | Path, path: str | Path) -> str:
+    base_str = _remote_path(base)
+    prefix = "/" if base_str.startswith("/") else ""
+    base_parts = [part for part in base_str.split("/") if part]
+    path_parts = [part for part in _remote_path(path).split("/") if part]
+    return prefix + "/".join(base_parts + path_parts)
```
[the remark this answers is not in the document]
neubig: Fixed in 7e99ed52: `_join_remote_path()` now preserves absolute POSIX and Windows paths instead of concatenating them with the working directory.

openhands-sdk/openhands/sdk/workspace/remote/remote_workspace_mixin.py
```diff
@@ -311,10 +326,11 @@ def _git_changes_generator(
         Raises:
             Exception: If path is not a git repository or getting changes failed
         """
-        # Make HTTP call
+        remote_path = _join_remote_path(self.working_dir, path)
```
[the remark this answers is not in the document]
neubig: Addressed in 7e99ed52 together with the `_join_remote_path()` fix. Absolute git paths now stay absolute when we build the query-param request.

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -436,6 +410,33 @@ def view(
...
+    def _count_hidden_children(self, path: Path) -> int:
+        return sum(1 for item in path.iterdir() if item.name.startswith("."))
+
+    def _list_directory_for_view(self, path: Path) -> list[str]:
+        visible_entries = [path]
+        for item in sorted(path.iterdir(), key=lambda p: str(p)):
+            if item.name.startswith("."):
+                continue
+            visible_entries.append(item)
+            if item.is_dir():
```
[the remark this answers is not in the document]
neubig: Keeping this out of the Windows-support PR for now. This thread points at existing file-editor permission handling that is unrelated to the current CI failures, and I do not want to broaden the stack base with an unrelated behavior change.

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -283,38 +283,8 @@ def view(
...
-            # Add trailing slashes to directories
-            paths = stdout.strip().split("\n") if stdout.strip() else []
-            formatted_paths = []
-            for p in paths:
-                if Path(p).is_dir():
-                    formatted_paths.append(f"{p}/")
-                else:
-                    formatted_paths.append(p)
+            hidden_count = self._count_hidden_children(path)
+            formatted_paths = self._list_directory_for_view(path)
```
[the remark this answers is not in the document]
neubig: Same rationale as the companion file-editor thread: this is an existing permission-handling concern outside the Windows-support changes in this PR, so I am leaving it out of scope here.

.github/workflows/tests.yml
```diff
@@ -183,6 +183,81 @@ jobs:
...
+                  # OS-specific gaps tracked in #2989 are still open.
+                  uv run python -m pytest -vvs `
+                    --cov=openhands-tools `
+                    --cov-report=term-missing `
+                    --cov-fail-under=0 `
+                    --cov-config=pyproject.toml `
+                    tests/tools `
+                    # These tests still exercise terminal and shell assumptions
+                    # tracked in #2986 and #2988.
+                    --ignore=tests/tools/browser_use/test_browser_executor_e2e.py `
```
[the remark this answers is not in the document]
neubig: I kept the ignore list as-is. The latest Windows tools run shows the suite executing successfully apart from the symlink regression test we just fixed, so adding more blanket ignores would hide passing coverage rather than unblock a real failure.

openhands-sdk/openhands/sdk/utils/path.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def to_posix_path(path: str | os.PathLike[str]) -> str:
+    """Return a slash-separated path string for wire/storage/display formats.
+
+    This intentionally does not resolve or validate the path. Use ``Path`` or
+    ``os.path`` directly when interacting with the local filesystem.
+    """
+
+    return os.fspath(path).replace("\\", "/")
```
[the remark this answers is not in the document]
neubig: Keeping the current behavior. `to_posix_path()` is intentionally the SDKs slash-normalization helper for remote/workspace paths and Windows-style user input; supporting Unix filenames that literally embed backslashes is not the intended use case for this helper.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -70,6 +70,20 @@ def _terminate_process(self, process: subprocess.Popen) -> None:
...
+                ["taskkill", "/F", "/T", "/PID", str(process.pid)],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                check=False,
+            )
+            try:
+                process.wait(timeout=1)
+            except subprocess.TimeoutExpired:
+                process.kill()
+                process.wait()
```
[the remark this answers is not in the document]
neubig: Leaving this one out of the Windows-support stack. The hook-executor timeout symmetry is a reasonable follow-up, but it is unrelated to the failing PR checks here and would broaden this PR into another subsystem.

openhands-tools/openhands/tools/glob/impl.py
```diff
@@ -208,11 +208,10 @@ def _execute_with_glob(
...
             matches = glob_module.glob(pattern, recursive=True)
 
-            # Convert to absolute paths (without resolving symlinks)
-            # and sort by modification time
+            # Convert to absolute paths without resolving symlinks and sort by
+            # modification time.
             file_paths = []
             for match in matches:
-                # Use absolute() instead of resolve() to avoid resolving symlinks
                 abs_path = str((search_path / match).absolute())
```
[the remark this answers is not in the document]
neubig: Keeping `.absolute()` intentionally. The Python glob fallback should preserve the matched symlink path rather than rewriting it to the target path, and the latest regression test now checks that behavior without depending on Windows temp-path canonicalization details.

openhands-tools/openhands/tools/grep/impl.py
```diff
@@ -195,51 +197,28 @@ def _execute_with_ripgrep(
...
+            dirs[:] = [name for name in dirs if not name.startswith(".")]
+            for filename in files:
+                if filename.startswith("."):
+                    continue
+                if action.include and not fnmatch.fnmatch(filename, action.include):
+                    continue
+
+                file_path = Path(root) / filename
+                try:
+                    content = file_path.read_text(encoding="utf-8", errors="ignore")
```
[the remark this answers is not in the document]
neubig: No change here for this PR. The fallback already caps results, and I am avoiding a larger grep refactor while iterating on the Windows-support stack.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -70,6 +70,20 @@ def _terminate_process(self, process: subprocess.Popen) -> None:
         Uses process groups to kill the entire process tree, not just
         the parent shell when shell=True is used.
         """
+        if os.name == "nt":
+            subprocess.run(
+                ["taskkill", "/F", "/T", "/PID", str(process.pid)],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                check=False,
```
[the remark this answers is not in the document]
neubig: Leaving this as-is for now. `check=False` is deliberate here because the cleanup path is best-effort; if we want extra logging we can do that as a separate follow-up without widening this PR.

.github/workflows/tests.yml
```diff
@@ -183,6 +183,81 @@ jobs:
...
+                  # Keep the initial Windows pass non-blocking on coverage while
+                  # OS-specific gaps tracked in #2989 are still open.
+                  uv run python -m pytest -vvs `
+                    --cov=openhands-tools `
+                    --cov-report=term-missing `
+                    --cov-fail-under=0 `
+                    --cov-config=pyproject.toml `
+                    tests/tools `
+                    # These tests still exercise terminal and shell assumptions
+                    # tracked in #2986 and #2988.
```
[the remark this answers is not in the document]
neubig: The referenced follow-up issues already exist in the repository (#2986 and #2988), so I kept the comments intact.

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -560,14 +573,14 @@ def validate_path(self, command: CommandLiteral, path: Path) -> None:
         2. Path and command are compatible
         """
         # Check if its an absolute path
-        if not path.is_absolute():
+        if not path.is_absolute() and not str(path).startswith(("/", "\\")):
             suggestion_message = (
                 "The path should be an absolute path, starting with `/`."
```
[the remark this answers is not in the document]
neubig: Thanks — I’m leaving this as-is for now. The validation logic already accepts both Unix and Windows absolute paths, and I want to avoid restarting the code/CI loop for a cosmetic string after the updated review already approved the implementation. If needed, we can follow up with a dedicated wording cleanup.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/extensions/fetch.py
```diff
@@ -29,6 +30,12 @@ class SourceType(StrEnum):
     GITHUB = "github"
 
 
+def _is_local_path_source(source: str) -> bool:
```
neubig: It seems that it would be better to have this implemented in is_local_path_source `openhands.sdk.utils.path` instead of having a separate wrapper in this file.

openhands-sdk/openhands/sdk/git/models.py
```diff
@@ -15,6 +17,10 @@ class GitChange(BaseModel):
     status: GitChangeStatus
     path: Path
 
+    @field_serializer("path", when_used="json")
```
neubig, line 20: Will this deserialize appropriately?

openhands-sdk/openhands/sdk/plugin/source.py
```diff
@@ -97,6 +98,9 @@ def resolve_source_path(
         return None
 
     # Handle local paths
+    if source.startswith("/"):
+        return Path(source)
+
```
neubig: I don't understand the necessity for this.

openhands-sdk/openhands/sdk/skills/utils.py
```diff
@@ -147,8 +146,19 @@ def replace_var(match: re.Match) -> str:
...
+        if isinstance(value, dict):
+            return {key: expand_value(item) for key, item in value.items()}
+        if isinstance(value, list):
+            return [expand_value(item) for item in value]
+        return value
+
+    expanded_config = expand_value(serializable_config)
+    if not isinstance(expanded_config, dict):
+        raise TypeError("expanded MCP config must be a dictionary")
+    return expanded_config
```
neubig, line 164: It is not clear why this is necessary for windows support. If it is not, revert.

openhands-tools/openhands/tools/browser_use/impl_windows.py
neubig, line 1: Is this file no longer necessary?

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -560,14 +573,14 @@ def validate_path(self, command: CommandLiteral, path: Path) -> None:
         2. Path and command are compatible
         """
         # Check if its an absolute path
-        if not path.is_absolute():
+        if not path.is_absolute() and not str(path).startswith(("/", "\\")):
```
neubig: This should be in the path utils, like we have a test for local paths there.

tests/agent_server/test_terminal_service.py
neubig, line 1: Instead of changing these tests, let's skip them for now and fix them in a follow-up that enables the PowerShell terminal.

tests/cross/test_agent_loading.py
neubig, line 1: Instead of changing these tests, let's skip them for now and fix them in a follow-up that enables the PowerShell terminal.

tests/cross/test_conversation_restore_behavior.py
neubig, line 1: Instead of changing these tests, let's skip them for now and fix them in a follow-up that enables the PowerShell terminal.

tests/cross/test_hello_world.py
neubig, line 1: Instead of changing these tests, let's skip them for now and fix them in a follow-up that enables the PowerShell terminal.

tests/cross/test_registry_directories.py
neubig, line 1: Instead of changing these tests, let's skip them for now and fix them in a follow-up that enables the PowerShell terminal.

tests/sdk/hooks/test_executor.py
```diff
@@ -204,6 +232,33 @@ def test_execute_async_hook_receives_stdin(self, executor, sample_event, tmp_pat
         assert content["tool_name"] == "TestTool"
         assert content["event_type"] == "PostToolUse"
 
+    def test_execute_async_hook_uses_windows_process_group(
```
neubig, line 235: Why does this task pass on unix runners? It seems to only apply to windows. Clarify.

tests/sdk/plugin/test_installed_plugins.py
neubig, line 1: The change in this file is unnecessary.

tests/sdk/skills/test_skill_utils.py
```diff
@@ -14,11 +14,19 @@
     load_skills_from_dir,
 )
 from openhands.sdk.skills.utils import find_third_party_files
+from openhands.sdk.utils.path import to_posix_path
 
 
 CONTENT = "# dummy header\ndummy content\n## dummy subheader\ndummy subcontent\n"
 
 
+def _symlink_or_skip(source: Path, link_name: Path) -> None:
```
neubig: It might be necessary to copy instead of symlink to avoid losing functionality

tests/tools/browser_use/test_browser_executor_e2e.py
```diff
@@ -105,6 +108,43 @@
...
+            raise RuntimeError("Test HTTP server exited before accepting requests")
+        try:
+            with urllib.request.urlopen(url, timeout=0.5):
+                return
+        except OSError:
+            time.sleep(0.1)
+    raise RuntimeError(f"Test HTTP server did not start within {timeout_seconds}s")
+
+
+def _skip_if_recording_unavailable(result: BrowserObservation) -> None:
```
neubig: Is recording ever unavailable on windows? Or is the problem just that playwright was not installed properly?

tests/tools/file_editor/test_basic_operations.py
```diff
@@ -24,6 +24,13 @@
 )
 
 
+def _symlink_or_skip(source: Path, link_name: Path) -> None:
```
neubig: This is the second instance of _symlink_or_skip, these should be abstracted into path utils.

openhands-sdk/openhands/sdk/extensions/fetch.py
neubig: Addressed in 03776cfb: moved the local path source behavior into openhands.sdk.utils.path and now call it directly from extensions/fetch.py.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/git/models.py
neubig: Addressed in 03776cfb: added a GitChange regression that verifies POSIX serialization and model_validate deserialization back to a Path.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/plugin/source.py
neubig: Addressed in 03776cfb: removed the inline special case and centralized POSIX/Windows absolute path detection in openhands.sdk.utils.path.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/skills/utils.py
neubig: Addressed in 03776cfb: kept the recursive value expansion and added a Windows-path regression showing backslashes are preserved without raw JSON substitution issues.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-tools/openhands/tools/browser_use/impl_windows.py
neubig: Addressed in 03776cfb: removed the Windows browser executor shim and use the shared BrowserToolExecutor on Windows as well.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-tools/openhands/tools/file_editor/editor.py
neubig: Addressed in 03776cfb: moved the absolute path check to the shared path utility and reused it from the file editor.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/agent_server/test_terminal_service.py
neubig: Addressed in 03776cfb: restored the Unix terminal service test behavior and made the module skip explicitly on Windows until the PowerShell terminal follow-up.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/cross/test_agent_loading.py
neubig: Addressed in 03776cfb: restored TerminalTool coverage and skipped the module on Windows until the PowerShell terminal follow-up.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/cross/test_conversation_restore_behavior.py
neubig: Addressed in 03776cfb: restored TerminalTool restore behavior coverage and skipped the module on Windows until the PowerShell terminal follow-up.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/cross/test_hello_world.py
neubig: Addressed in 03776cfb: restored the hello-world TerminalTool setup and skipped the module on Windows until the PowerShell terminal follow-up.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/cross/test_registry_directories.py
neubig: Addressed in 03776cfb: restored the TerminalTool directory assertion and skipped that test on Windows until the PowerShell terminal follow-up.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/sdk/hooks/test_executor.py
neubig: Addressed in 03776cfb: clarified the test docstring to say it simulates win32 on any runner, so the Windows subprocess kwargs are exercised on Unix CI too.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/sdk/plugin/test_installed_plugins.py
neubig: Addressed in 03776cfb: reverted the unnecessary separator removal.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/sdk/skills/test_skill_utils.py
neubig: Addressed in 03776cfb: kept symlink coverage, moved the helper to tests/path_utils.py, and skip only when the environment denies symlink creation.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/tools/browser_use/test_browser_executor_e2e.py
neubig: Addressed in 03776cfb: removed the broad recording-unavailable skip so Windows fails if recording itself is broken; the remaining skip only covers the existing CDP initialization branch.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

tests/tools/file_editor/test_basic_operations.py
neubig: Addressed in 03776cfb: moved the duplicate symlink helper into tests/path_utils.py and reused it from the file editor tests.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -54,6 +56,40 @@
 }
 
 
+class _EncodingSafeTextIO:
```
VascoSch92, line 61: nit: this can be a dataclass

openhands-sdk/openhands/sdk/utils/path.py
```diff
@@ -0,0 +1,58 @@
...
+    if not value:
+        return False
+    if value.startswith(("/", "\\")):
+        return True
+    if Path(value).expanduser().is_absolute():
+        return True
+    return PureWindowsPath(value).is_absolute()
+
+
+def is_local_path_source(source: str) -> bool:
```
VascoSch92, line 52: This new implementation classifies any string starting with `.` as local , i.e., `.openhands`, `.foo`, `..bar` are now local paths. 

the old versions was:     
```python
  any(source.startswith(p) for p in ("./", "../", "/", "~", "file://"))
```                                                                                              
-> `.openhands` would have been False before, True now. 

Is this wanted? I suppose yes as the test test_is_local_path_source_detects_dot_paths is checking this behaviour.

Perhaps fix the comments

openhands-sdk/openhands/sdk/skills/utils.py
```diff
@@ -147,8 +146,19 @@ def replace_var(match: re.Match) -> str:
         # Return original if not found (preserves placeholder for later expansion)
         return match.group(0)
 
-    config_str = var_pattern.sub(replace_var, config_str)
-    return json.loads(config_str)
+    def expand_value(value: object) -> object:
```
VascoSch92, line 149: the previous implementation went through json.dumps → regex.sub → json.loads, so ${VAR} was replaced anywhere it appeared in the JSON-serialized text — including in dict keys. The new recursive walker   copies keys verbatim and only substitutes inside string values. 
 
Example: 
                                  
``` python
  config = {"servers": {"${VAR}_server": {"command": "echo"}}}
  variables = {"VAR": "foo"} 
            
  # OLD: {'servers': {'foo_server': {'command': 'echo'}}}
  # NEW: {'servers': {'${VAR}_server': {'command': 'echo'}}}   ← regression
```                                                                           

this code is better for value content with quotes/backslashes (the old JSON round-trip blew up on those), but the key-expansion path is a behavior loss.

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -560,14 +573,14 @@ def validate_path(self, command: CommandLiteral, path: Path) -> None:
         2. Path and command are compatible
         """
         # Check if its an absolute path
-        if not path.is_absolute():
+        if not is_absolute_path_source(path):
```
VascoSch92: Editor can create silently creates a file literally named `C:\foo` on Linux/macOS                                                                                                                                
                                                                                                                                         
This is because `is_absolute_path_source` falls through to `PureWindowsPath(value).is_absolute()` on every platform, so r"C:\foo" is reported absolute on POSIX. `validate_path` accepts it; for `command="create"` the file then gets written to cwd with \ as part of the filename.                                                                                                                                                                
  
```                                            
  Path(r"C:\foo")                      → PosixPath('C:\\foo')
  Path.is_absolute()       (old check) → False                                                                                                                                                                       
  is_absolute_path_source  (new check) → True                     
  validate_path(create, …)             → ACCEPTED                                                                                                                                                                    
  result: file 'C:\foo' created in cwd     ← silent corruption
```

It is not a regression but would be nice to fix.

VascoSch92 (approved):
A couple of things/warnings but otherwise seems fine to me.

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
neubig: We should check if we typically use dataclasses or datamodels in this case, but agreed.

neubig:
@openhands reflect the review comments

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
neubig: Addressed in `11d322ce`: converted `_EncodingSafeTextIO` into a small dataclass wrapper to match the existing nit while keeping the encoding-safe console behavior unchanged.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/utils/path.py
neubig: Addressed in `11d322ce`: kept the intended dot-prefixed local-path behavior and updated `is_local_path_source()` documentation to explicitly call out `.`, `..`, and names like `.openhands` as supported local sources.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/skills/utils.py
neubig: Addressed in `11d322ce`: `expand_mcp_variables()` now expands dictionary keys as well as values, and a regression test covers the legacy `${VAR}` key-substitution behavior without reintroducing the JSON/backslash issue.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-tools/openhands/tools/file_editor/editor.py
neubig: Addressed in `11d322ce`: the file editor now validates paths with host-native absolute-path semantics instead of the cross-platform source helper, so POSIX no longer accepts Windows drive paths like `C:\foo` as creatable files. Added path-helper coverage and a file-editor regression test.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

PR #3001, opened 2026-04-29
fix(acp): drop redundant MessageEvent from step() end-of-turn emission

### LINKED ISSUE #3000: ACP agent emits three events with identical response text per turn, causing duplicate UI rendering
## Summary

After each ACP agent turn (e.g. Claude Code responding to a user message), the SDK emits **three separate events containing the same response text**:

1. `MessageEvent(source="agent", llm_message=Message(role="assistant", content=[TextContent(text=response_text)]))`
2. `ActionEvent(source="agent", action=FinishAction(message=response_text), ...)`
3. `ObservationEvent(observation=FinishObservation.from_text(text=response_text), ...)`

This happens unconditionally in `ACPAgent.step()` at the end of every turn ([`acp_agent.py` lines 1278–1311](https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/agent/acp_agent.py#L1278-L1311)):

```python
msg_event = MessageEvent(
    source="agent",
    llm_message=message,          # ← response_text
)
on_event(msg_event)

finish_action = FinishAction(message=response_text)   # ← same text
...
action_event = ActionEvent(source="agent", action=finish_action, ...)
on_event(action_event)

on_event(ObservationEvent(
    observation=FinishObservation.from_text(text=response_text),  # ← same text again
    ...
))
```

## Impact

The app-server's frontend receives all three events. The current OpenHands frontend rendering pipeline:

- **`FinishObservation`** is suppressed by `handleEventForUI` (intentionally — FinishAction is kept instead).
- **`FinishAction`** renders the response text as the agent's turn output.
- **`MessageEvent` (role=assistant)** also renders the response text, because `isMessageEvent` returns `true` and `shouldRenderEvent` passes it through.

Result: every ACP agent response appears **twice** in the conversation UI.

## Root cause

The comment in the code explains the intent:

> ACP step() boundaries are full remote assistant turns, not partial planning steps. Emit FinishAction to delimit that completed turn for eval/remote consumers, matching #2190.

So `FinishAction` + `FinishObservation` exist for eval/remote consumers. The `MessageEvent` exists to record the raw LLM response in the conversation history. These are three legitimate consumers with different needs — but they all end up on the same event stream, which the UI also consumes.

## Why this doesn't happen for the standard LLM agent

The standard OpenHands `CodeActAgent` parses the LLM response as structured JSON into typed actions (ThinkAction, ExecuteBashAction, FinishAction, etc.). The raw LLM response string is never emitted as a `MessageEvent` — it's consumed internally and translated into actions. So the UI only ever sees the structured actions.

ACP agents return natural language text (not JSON), so the bridge has no structured output to parse and falls back to emitting both the raw response (`MessageEvent`) and the structured wrapper (`FinishAction`/`FinishObservation`).

## Proposed fix

One of the following:

**Option A — Drop `MessageEvent` for ACP turns** (simplest)  
Only emit `FinishAction` + `FinishObservation`. The `MessageEvent` is redundant because `FinishAction.message` already carries the full response text. Evals and remote consumers that need the conversation transcript can reconstruct it from `FinishAction`.

**Option B — Drop `FinishAction`/`FinishObservation` for ACP turns**  
Only emit `MessageEvent`. This is the more natural representation of an ACP turn (it was a conversation message, not an action/observation pair). However, this would break eval frameworks that depend on `FinishAction` to detect turn boundaries.

**Option C — Mark `MessageEvent` as non-displayable**  
Add a metadata flag (e.g. `display=False`) to the `MessageEvent` emitted by the ACP bridge, so event stream consumers can choose whether to render it. Requires a protocol change.

**Option D — Frontend workaround only (not recommended as sole fix)**  
Filter out assistant-role `MessageEvent`s in `shouldRenderEvent`. This fixes the visual duplicate but leaves the underlying stream redundancy in place.

The cleanest long-term solution is **Option A** — the `MessageEvent` is internal bookkeeping that the ACP bridge should not need to publish to the shared event stream. Eval consumers already get what they need from `FinishAction`.

## Reproduction

1. Start OpenHands with an ACP agent configured (e.g. `claude-code`)
2. Send any message in a conversation
3. Fetch the conversation event stream: `GET /api/conversations/{id}/events`
4. Observe three events with the same `response_text`:
   - `MessageEvent` with `llm_message.role == "assistant"`
   - `ActionEvent` with `action.kind == "FinishAction"`
   - `ObservationEvent` with `observation.kind == "FinishObservation"`

## Related

- OpenHands PR [#14189](https://github.com/OpenHands/OpenHands/pull/14189) — ACP agent settings UI (discovered this issue during testing)
- OpenHands PR [#13994](https://github.com/OpenHands/OpenHands/pull/13994) — ACP tool call event rendering
- SDK issue [#2190](https://github.com/OpenHands/software-agent-sdk/issues/2190) — referenced in the code comment above

simonrosenberg:
## Summary

Fixes #3000 — ACP agent responses were appearing twice in the conversation UI.

- `ACPAgent.step()` was emitting three events with identical response text per turn: `MessageEvent`, `ActionEvent(FinishAction)`, and `ObservationEvent(FinishObservation)`
- The frontend rendered both `MessageEvent` (role=assistant) and `FinishAction`, causing every response to appear twice
- Fix: drop the `MessageEvent` from the success path — `FinishAction.message` already carries the full response text for eval and remote consumers

## Test plan

- [x] Updated `test_step_emits_finish_action_event` (renamed from `test_step_emits_message_event`) to assert 2 events (not 3) and events[0] is `ActionEvent(FinishAction)`
- [x] Updated `test_step_includes_reasoning` → `test_step_with_reasoning_emits_finish_action` to reflect thoughts are accumulated but not surfaced via event stream
- [x] Updated `test_step_no_response_text_fallback` to check fallback text in `FinishAction.message`
- [x] Updated tool-call event ordering test: 2 tool events + FinishAction + FinishObservation (4 total, not 5)
- [x] Updated retry test to check `FinishAction.message` for response text
- [x] All ACP agent tests pass (one pre-existing unrelated failure: `test_api_key_fallback_when_no_chatgpt_file`)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:e609641-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-e609641-python \
  ghcr.io/openhands/agent-server:e609641-python
```

```
ghcr.io/openhands/agent-server:e609641-golang-amd64
ghcr.io/openhands/agent-server:e609641-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:e609641-golang-arm64
ghcr.io/openhands/agent-server:e609641-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:e609641-java-amd64
ghcr.io/openhands/agent-server:e609641-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:e609641-java-arm64
ghcr.io/openhands/agent-server:e609641-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:e609641-python-amd64
ghcr.io/openhands/agent-server:e609641-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:e609641-python-arm64
ghcr.io/openhands/agent-server:e609641-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:e609641-golang
ghcr.io/openhands/agent-server:e609641-java
ghcr.io/openhands/agent-server:e609641-python
```

- Each variant tag (e.g., `e609641-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `e609641-python-amd64`) are also available if needed


simonrosenberg:
The import ordering issue flagged by the QA bot was fixed in the second commit (pre-commit cleanups). All checks should be passing now.

On the reasoning content: yes, dropping `thought_text` from the event stream is intentional. The `accumulated_thoughts` were only ever exposed via `MessageEvent.llm_message.reasoning_content`, which we're removing as the fix. Since `FinishAction` has no reasoning field and ACP turn thoughts don't need to be replayed by evals (they consume the final response, not the agent's chain-of-thought), this is the right trade-off. If we need reasoning traces for observability in the future, the right place is an explicit `ThinkingEvent` or similar — not bundled into the end-of-turn `MessageEvent` that was causing the duplicate rendering bug.

VascoSch92 (approved):
LGTM

PR #3007, opened 2026-04-29
lazy load Laminar

dinmukhamedm:
- [X] A human has tested these changes.

---

## Why

Loading OpenHands SDK is slow, and we suspect it could potentially be due to Laminar's heavy downstream dependencies. Turning it off for users who don't need it by loading it lazily

## Summary

- move all imports to inside functions
- make decorator pass-through if Laminar not enable
- cache the result of Laminar check in a global variable

## How to Test

```
export LMNR_PROJECT_API_KEY=YOUR_KEY
```

Then run OpenHands

## Type

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


openhands-sdk/openhands/sdk/observability/laminar.py
```diff
@@ -115,38 +119,72 @@ def observe[**P, R](
...
 
-def should_enable_observability():
-    keys = [
-        "LMNR_PROJECT_API_KEY",
-        "OTEL_ENDPOINT",
-        "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT",
-        "OTEL_EXPORTER_OTLP_ENDPOINT",
-    ]
-    if any(get_env(key) for key in keys):
+_OBSERVABILITY_ENV_KEYS = (
```
VascoSch92:
```suggestion
_OBSERVABILITY_ENV_KEYS: Final[tuple[str, ...]] = (
```

and probably you should also import Final from typing.

VascoSch92: As this is a constant: can you put it at the top of the file?

tests/sdk/observability/test_laminar.py
VascoSch92, line 1: Walk through what happens when `test_lmnr_base_url_passed_to_laminar` runs:  

```                                                                                                                                                                                                                     
  os.environ["LMNR_PROJECT_API_KEY"] = "test-key" 
  maybe_init_laminar() 
  # ↓ inside the function ↓
  # if not should_enable_observability(): ... 
  #     ↓ inside should_enable_observability ↓
  #     if any(get_env(key) for key in _OBSERVABILITY_ENV_KEYS): 
  #         _observability_enabled = True   ← module-level global is now True 
  #         return True
```
                                                                                                                                                                                                   
Then the test's finally block cleans os.environ. But `_observability_enabled` stays `True` for the entire pytest session as it's a module attribute.                                              
                                                                                                                                                                                                                     
Today nothing fails because no test asserts the disabled path. But the moment anyone adds a test like:                                                                                                             

```python
  def test_passthrough_when_disabled(): 
      # no env vars set 
      assert should_enable_observability() is False  # ← fails if a prior test set the flag
```
                                                                                                                                                                                                          
…it will pass when run alone, fail when run after the other tests, and fail/pass non-deterministically under pytest-xdist parallelism (the module is per-worker but test ordering still matters within a worker).                                                                                                                                                                                 
                                                                             
Could you add this at the top of `tests/sdk/observability/test_laminar.py,` right after the import pytest line:                                                                                                                

```python
  @pytest.fixture(autouse=True)
  def _reset_observability_cache():                                         
      """Reset the module-level _observability_enabled flag between tests.
                                                                                                                                                                                                                
      The flag is sticky-True by design (see laminar.py docstring), so it    
      leaks across tests. This fixture isolates each test from prior state.                                                                                                                                          
      """                                     
      from openhands.sdk.observability import laminar                                                                                                                                                                
       
      laminar._observability_enabled = False                   
      yield  
      laminar._observability_enabled = False                 
```

VascoSch92 (changes requested):
Hey @dinmukhamedm 

Thank you very much for your PR.

On this branch, I measured the openhands.sdk import at ~1,348 ms, compared to ~1,455 ms when lmnr is loaded eagerly. The real bottleneck with imports is actually LiteLLM.

That said, the code is solid, and it would be great to get this merged.

I've left a couple of minor comments.

Rainhunter13:
@VascoSch92 addressed PR comments, thank you

VascoSch92:
hey @dinmukhamedm 

The 8 MCP test failures are caused by this PR. @observe's new lazy_wrapper is a sync def, so inspect.iscoroutinefunction(MCPToolExecutor.call_tool) returns False after decoration, and run_async rejects the     input.

you can find the fix in this branch: https://github.com/OpenHands/software-agent-sdk/tree/fix/laminar-eager-import

I can not push to your fork.

xingyaoww:
@OpenHands please read https://github.com/OpenHands/software-agent-sdk/pull/3007#pullrequestreview-4203811214 and then create a test that tests the time taken to import openhands.sdk

Then make sure you measure the performance difference before AND after this PR, and then put together a report and help me understand if this PR actually improves loading performance.

Then create a PR to merge that test into this branch.

xingyaoww:
@OpenHands please read https://github.com/OpenHands/software-agent-sdk/pull/3007#pullrequestreview-4203811214 and then create a test that tests the time taken to import openhands.sdk

Then make sure you measure the performance difference before AND after this PR, and then put together a report and help me understand if this PR actually improves loading performance.

Then create a PR to merge that test into this branch.

xingyaoww:
(retrigger bc gpt-5.5 broken my runtime 😭 - trying to fix them here: https://github.com/OpenHands/software-agent-sdk/pull/3030

xingyaoww:
## 📊 Import Performance Report — Lazy Load Laminar

_This report was generated by an AI agent (OpenHands) on behalf of @xingyaoww._

### Methodology
- Each measurement spawns a **fresh Python subprocess** (`python -c "import openhands.sdk"`) to avoid warm-cache bias.
- **10 runs** per configuration, reporting mean/median/min/max/stdev.
- Same machine, sequential runs, no other load changes between branches.

### Results

| Module | main (eager) | PR branch (lazy) | Δ (ms) | Δ (%) |
|---|---|---|---|---|
| `import openhands.sdk` | **2.997s** | **2.843s** | **−154ms** | **−5.1%** |
| `import openhands.sdk.observability.laminar` | 3.025s | 2.817s | −208ms | −6.9% |
| `import lmnr` (standalone) | 2.017s | 2.027s | +10ms | — (control) |

### Key Observations

| | main | PR branch |
|---|---|---|
| `lmnr` in `sys.modules` after SDK import? | ✅ **Yes** (always loaded) | ❌ **No** (deferred) |

1. **The PR does work**: After this PR, `import openhands.sdk` no longer eagerly imports `lmnr`. The module is deferred until observability is actually enabled via environment variables or explicit `Laminar.initialize()`.

2. **~154ms improvement (~5.1%)**: The savings come from avoiding `lmnr`'s unique transitive imports that aren't shared with the rest of the SDK. While `import lmnr` alone costs ~2.0s, most of that (~1.85s) is shared dependencies (litellm, opentelemetry, etc.) that the SDK already imports through other paths.

3. **The real bottleneck is LiteLLM**: As the reviewer @VascoSch92 noted, the ~2.8s baseline is dominated by `litellm` and its transitive imports. Lazy-loading `lmnr` eliminates the incremental cost that Laminar adds on top.

4. **Consistent with reviewer measurements**: @VascoSch92 measured ~1,455ms → ~1,348ms (107ms improvement, ~7.4%). Our measurements show a similar percentage improvement (~5-7%), with the absolute numbers being higher likely due to different hardware.

### Conclusion

This PR provides a **real but modest improvement** (~150ms, ~5%) to SDK import time by deferring `lmnr` imports. The improvement is correctly targeted — users who don't need observability no longer pay the import cost. The code changes are solid (lazy wrapper preserves async signatures, caches results properly).

For larger import time improvements, the focus should shift to `litellm` which dominates the import cost.

### Test Added

Committed `tests/sdk/test_import_performance.py` (commit 7423b7d2) — a regression guard that asserts `import openhands.sdk` completes in under 10s (generous limit for CI stability).

xingyaoww:
@dinmukhamedm happy to approve and merge this once https://github.com/OpenHands/software-agent-sdk/commit/7423b7d2b9a094660569268506a485b85b7cacdf is included

Rainhunter13:
@xingyaoww thanks, applied the test commit (@dinmukhamedm is currently unavailable)

xingyaoww (approved):
Feel free to merge once @VascoSch92 approves

VascoSch92 (approved):

PR #3018, opened 2026-04-30
feat(sdk): allow Conversation.switch_profile to accept an inline LLM

### LINKED ISSUE #3017: Allow Conversation.switch_profile to accept an inline LLM, bypassing the profile store
## Summary

Extend `Conversation.switch_profile(profile_name)` (in `openhands/sdk/conversation/impl/local_conversation.py`) to accept an optional `llm: LLM | None` argument. When supplied, the LLM is used directly (tagged with `usage_id = f"profile:{profile_name}"` and added to the LLM registry) and `LLMProfileStore` is skipped. Backwards compatible — callers that don't pass `llm` get today's behaviour.

## Motivation

In the OpenHands app-server, LLM profiles are stored as part of user settings (`Settings.llm_profiles`), **not** as files in the agent-server sandbox's `~/.openhands/profiles/`. When a user runs `/model <name>` to switch the running conversation's LLM, the app-server has the profile (including `api_key`) but the agent-server does not.

Today's `POST /api/conversations/{id}/switch_profile` HTTP endpoint calls `Conversation.switch_profile(name)`, which calls `LLMProfileStore.load(name)`. That store reads from a sandbox-local FS directory that the app-server never populates, so the call always returns 404 `Profile '{name}' not found` for app-server-side profiles.

## Current workaround (and why it's not great)

The app-server can serialize the profile JSON, upload it to the sandbox via `POST /api/file/upload?path=/root/.openhands/profiles/{name}.json`, and then call `switch_profile`. It works, but:

- Two HTTP calls per switch instead of one.
- Hardcodes `/root/.openhands/profiles/` (assumes agent-server runs as root).
- Persists profile JSON inside the sandbox FS as a side effect — orthogonal to the actual swap.
- Couples the app-server to the SDK's storage layout.

## Proposal

```python
def switch_profile(self, profile_name: str, llm: LLM | None = None) -> None:
    """Switch the agent's LLM to a named profile.

    If ``llm`` is provided it is used directly (tagged as
    ``profile:{profile_name}``) and the profile store is skipped.
    Otherwise the profile is loaded from :class:`LLMProfileStore`.
    """
    usage_id = f"profile:{profile_name}"
    if llm is not None:
        new_llm = llm.model_copy(update={"usage_id": usage_id})
        self.llm_registry.add(new_llm)            # refresh — caller is authoritative
    else:
        try:
            new_llm = self.llm_registry.get(usage_id)
        except KeyError:
            new_llm = self._profile_store.load(profile_name)
            new_llm = new_llm.model_copy(update={"usage_id": usage_id})
            self.llm_registry.add(new_llm)
    with self._state:
        self.agent = self.agent.model_copy(update={"llm": new_llm})
        self._state.agent = self.agent
        self._pin_prompt_cache_key()
```

## Companion agent-server change

The HTTP wrapper in `openhands/agent_server/conversation_router.py` (around `switch_conversation_profile`) gains one optional `Body` field and forwards it:

```python
async def switch_conversation_profile(
    conversation_id: UUID,
    profile_name: str = Body(..., embed=True),
    llm: LLM | None = Body(None, embed=True),     # new
    conversation_service: ConversationService = Depends(get_conversation_service),
) -> Success:
    ...
    conversation.switch_profile(profile_name, llm=llm)
```

## Backwards compatibility

`llm` defaults to `None`, so existing CLI/SDK callers that rely on `LLMProfileStore` are unaffected. The HTTP endpoint stays compatible — clients that omit `llm` get today's behaviour, including the existing 404 on missing profiles.

## Why extend `switch_profile` instead of adding a new `set_llm` method

A separate `set_llm(llm: LLM)` would also work, but:

- `switch_profile`'s tail already does exactly the right thing (registry refresh + state swap + prompt cache pin); the only difference is whether the LLM comes from the store or the caller.
- Reusing the same `usage_id = f"profile:{name}"` keeps the LLM registry coherent across both code paths — a later `switch_profile(name)` (without `llm`) will hit the cached entry.
- One endpoint stays canonical for "swap the running agent's LLM."

Happy to send a PR if this direction is acceptable.

VascoSch92:

## Why

We refeer to #3017 for the details about the PR.

## Issue Number
Closes #3017.

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:20cec0d-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-20cec0d-python \
  ghcr.io/openhands/agent-server:20cec0d-python
```

```
ghcr.io/openhands/agent-server:20cec0d-golang-amd64
ghcr.io/openhands/agent-server:20cec0d-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:20cec0d-golang-arm64
ghcr.io/openhands/agent-server:20cec0d-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:20cec0d-java-amd64
ghcr.io/openhands/agent-server:20cec0d-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:20cec0d-java-arm64
ghcr.io/openhands/agent-server:20cec0d-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:20cec0d-python-amd64
ghcr.io/openhands/agent-server:20cec0d-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:20cec0d-python-arm64
ghcr.io/openhands/agent-server:20cec0d-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:20cec0d-golang
ghcr.io/openhands/agent-server:20cec0d-java
ghcr.io/openhands/agent-server:20cec0d-python
```

- Each variant tag (e.g., `20cec0d-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `20cec0d-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_router.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_router.py
@@ -297,7 +297,6 @@ async def set_conversation_security_analyzer(
 async def switch_conversation_profile(
     conversation_id: UUID,
     profile_name: str = Body(..., embed=True),
-    llm: LLM | None = Body(None, embed=True),  # noqa: B008
     conversation_service: ConversationService = Depends(get_conversation_service),
 ) -> Success:
     """Switch the conversation's LLM profile to a named profile."""
@@ -306,7 +305,7 @@ async def switch_conversation_profile(
         raise HTTPException(status.HTTP_404_NOT_FOUND)
     conversation = event_service.get_conversation()
     try:
-        conversation.switch_profile(profile_name, llm=llm)
+        conversation.switch_profile(profile_name)
     except FileNotFoundError:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
@@ -317,6 +316,28 @@ async def switch_conversation_profile(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=str(e),
         )
+    return Success()
+
+
+@conversation_router.post(
+    "/{conversation_id}/switch_llm",
+    responses={404: {"description": "Conversation not found"}},
+)
+async def switch_conversation_llm(
+    conversation_id: UUID,
+    llm: LLM = Body(..., embed=True),  # noqa: B008
+    conversation_service: ConversationService = Depends(get_conversation_service),
+) -> Success:
+    """Swap the conversation's LLM to a caller-supplied object.
+
+    Used by app-servers that own the LLM directly and don't push profiles
+    to the agent-server's filesystem (see #3017).
+    """
+    event_service = await conversation_service.get_event_service(conversation_id)
+    if event_service is None:
+        raise HTTPException(status.HTTP_404_NOT_FOUND)
+    conversation = event_service.get_conversation()
+    conversation.switch_llm(llm)
     return Success()
 
 
--- a/openhands-agent-server/openhands/agent_server/conversation_router.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_router.py
@@ -297,6 +297,7 @@ async def set_conversation_security_analyzer(
 async def switch_conversation_profile(
     conversation_id: UUID,
     profile_name: str = Body(..., embed=True),
+    llm: LLM | None = Body(None, embed=True),  # noqa: B008
     conversation_service: ConversationService = Depends(get_conversation_service),
 ) -> Success:
     """Switch the conversation's LLM profile to a named profile."""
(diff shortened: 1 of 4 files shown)


neubig (commented):
Could we delete the "name" and just pass an LLM object instead?

Also, maybe we could name it "switch_llm" instead of "switch_profile?

neubig (approved):

PR #3022, opened 2026-04-30
refactor(acp): centralize provider metadata and agent contracts

### LINKED ISSUE #3010: Simplify ACP integration: centralize metadata, unify agent construction, and reduce contract split
## Summary

ACP support currently works, but the integration boundary between \`software-agent-sdk\` and \`OpenHands\` is carrying avoidable duplication and special cases.

The main problem is that \`OpenHands\` is still re-encoding SDK behavior in a few places instead of consuming a single SDK-owned contract for ACP agents.

This issue tracks the SDK-side changes that would let \`OpenHands\` simplify its ACP code substantially. It also absorbs #3011 (Introduce \`AgentSettingsBase\`), which is a prerequisite for item B below.

### 1. ACP provider metadata is duplicated across repos

Today the same ACP provider knowledge exists in multiple places:

- \`software-agent-sdk\`: \`ACPAgentSettings\`, default command resolution, API key env mapping
- \`OpenHands\` backend: provider env-var mapping for API key / base URL injection
- \`OpenHands\` frontend: display names, default commands, provider-specific labels

This creates drift risk and forces \`OpenHands\` to keep local copies of SDK logic.

### 2. \`OpenHands\` still hand-builds ACP agents instead of using a single SDK construction path

The regular agent path already follows the right pattern:

- enrich settings
- call \`create_agent()\`
- apply server-only overrides

The ACP path still has a custom builder that manually reconstructs \`ACPAgent\` state. That means \`OpenHands\` must know about ACP command resolution, env behavior, and other SDK-specific details.

### 3. The canonical conversation contract is still split between legacy and ACP-capable schemas

The agent-server currently has:

- legacy \`/api/conversations\` pinned to \`Agent\`
- ACP-capable \`/api/acp/conversations\` using the union schema

Most of the service logic is already shared. The split is largely schema/compatibility-driven, but it leaks into app-server and frontend routing decisions.

### 4. Capability differences are expressed mostly as concrete ACP special-casing

\`ACPAgent\` really does differ from the regular \`Agent\` in important ways:

- no OpenHands-managed tools
- no OpenHands-managed MCP
- no condenser
- different startup / execution contract

Those differences are real, but downstream code often has to branch on \`ACPAgent\` directly rather than a more explicit capability contract.

### 5. Prompt / secrets / env behavior is not modeled cleanly enough at the SDK boundary

There are a few separate concerns that should be owned more cleanly by the SDK:

- provider credentials injected as subprocess env
- secrets that should be present in env
- secrets/context that should be rendered in ACP prompt context
- context that becomes a normal system prompt vs ACP user-turn prompt augmentation

\`OpenHands\` currently has to know too much about those details.

### 6. Small duplication within the settings model itself

Both \`OpenHandsAgentSettings\` and \`ACPAgentSettings\` independently define:

- \`schema_version\`
- \`export_schema()\` (delegating to \`export_settings_schema(cls)\`)

And downstream code relies on both having a \`create_agent()\` method via duck typing rather than an explicit shared contract.

### A. Introduce \`AgentSettingsBase\` (prerequisite for B)

Introduce a small shared base class that contains only the pieces truly common to both settings variants:

- \`schema_version\`
- \`export_schema()\`
- abstract \`create_agent()\`

Then:

- \`OpenHandsAgentSettings(AgentSettingsBase)\`
- \`ACPAgentSettings(AgentSettingsBase)\`
- Keep \`AgentSettingsConfig\` as the discriminated union used by Pydantic/FastAPI-facing types

This gives downstream code a typed contract for "settings that can build an agent" and eliminates the duplicated field/method implementations. The \`llm\` field is intentionally not hoisted — the semantics differ between variants (execution config vs. attribution identity) and the metadata overrides would make a base-class field awkward.

This is intentionally narrow: it will **not** materially simplify variant-specific fields, the discriminator logic, or schema export / section variant tagging.

### B. Add a single SDK-owned ACP provider registry

Expose one source of truth for built-in ACP providers, including:

- provider key / server kind
- display name
- default command
- API key env var
- base URL env var
- optionally auth method hints and default session mode hints

This should be consumable by:

- \`ACPAgentSettings\`
- \`ACPAgent\`
- \`openhands-agent-server\`
- external consumers like \`OpenHands\`

That would let \`OpenHands\` delete duplicated constants and fallback maps.

### C. Make \`ACPAgentSettings.create_agent()\` the canonical construction path

The SDK should expose all ACP construction behavior through settings helpers, so downstream callers do not need to manually rebuild \`ACPAgent\`.

Concretely:

- \`resolve_acp_command()\` should remain canonical
- provider metadata/env behavior should be reachable from settings/provider helpers
- constructing an ACP agent from validated settings should require no downstream re-encoding

Goal: \`OpenHands\` should be able to enrich settings, call \`create_agent()\`, and stop there.

### D. Introduce explicit agent capability helpers/interfaces

Instead of forcing downstream code to branch on \`isinstance(agent, ACPAgent)\`, expose capability-oriented helpers or properties such as:

- supports_openhands_tools
- supports_openhands_mcp
- supports_condenser
- startup_mode / initialization_cost class
- conversation_contract kind

That would let app/server code depend on behavior rather than concrete class identity.

### E. Move toward a single canonical ACP-capable conversation contract

Use the ACP-capable union schema as the canonical agent-server contract.

Then:

- keep \`/api/conversations\` as a compatibility facade for older clients
- treat the ACP-capable route/schema as the primary one
- reduce contract-specific branching in downstream consumers

This is mostly an \`openhands-agent-server\` concern, but it should be coordinated from the SDK side because the request/response models live there.

### F. Clarify the SDK contract for ACP prompt context / env / secret handling

The SDK should more explicitly define:

- what becomes subprocess env
- what becomes ACP prompt-only context
- what is legal for ACP agents vs regular agents
- what downstream callers are expected to provide vs what the SDK derives

The cleaner that boundary is, the less ACP-specific glue \`OpenHands\` has to keep.

## Expected downstream simplifications in \`OpenHands\`

If the SDK work above exists, \`OpenHands\` should be able to:

- delete duplicated ACP provider metadata/constants
- stop manually reconstructing \`ACPAgent\`
- share more of the regular/ACP request-building pipeline
- reduce route and discriminator special-casing
- rely on capability helpers instead of direct \`ACPAgent\` checks in more places

## Suggested sequencing

1. Introduce \`AgentSettingsBase\` in \`software-agent-sdk\` (item A — prerequisite)
2. Add ACP provider registry metadata in \`software-agent-sdk\` (item B)
3. Make settings-driven ACP construction fully canonical (item C)
4. Add capability helpers / contract helpers (item D)
5. Unify on the ACP-capable conversation contract in \`openhands-agent-server\` (item E)
6. Simplify \`OpenHands\` to consume the new SDK surface

## Notes

I already implemented a small amount of local cleanup in \`OpenHands\` to reduce obvious duplication, but the larger simplification should happen here first so \`OpenHands\` can depend on a stable SDK surface instead of carrying more local compatibility code.

simonrosenberg:
Closes #3010

## Summary

Refactors ACP-related settings and runtime behavior so provider-specific logic lives in the SDK instead of being duplicated by agent-server and OpenHands callers.

- Adds `AgentSettingsBase`, a shared base for settings variants with `schema_version`, `export_schema()`, and `create_agent()`.
- Adds `openhands.sdk.settings.acp_providers` as the single registry for built-in ACP provider metadata: display names, default commands, env vars, session modes, runtime detection patterns, and model-selection behavior.
- Updates `ACPAgentSettings` to resolve provider info, default commands, provider-native env from `llm.api_key` / `llm.base_url`, final ACP env, and prompt-only `agent_context` before constructing `ACPAgent`.
- Updates `ACPAgent` to use the registry for bypass session modes, Claude session `_meta`, `set_session_model` support, and Gemini base-URL auth instead of local string maps and hardcoded checks.
- Adds agent capability and conversation-contract helpers (`supports_openhands_tools`, `supports_openhands_mcp`, `supports_condenser`, `conversation_contract`) so downstream code can branch on behavior rather than concrete classes.
- Updates `RemoteConversation` to select the ACP endpoint through `agent.conversation_contract`.
- Exports the new settings/provider helpers from `openhands.sdk.settings` and `openhands.sdk`.

## Compatibility

This is intended to be mostly additive for SDK callers:

- Existing `AgentSettings` / `LLMAgentSettings` compatibility remains in place.
- Existing OpenHands-agent settings payloads continue to default to the OpenHands variant when `agent_kind` is absent.
- For built-in ACP providers, `ACPAgentSettings.create_agent()` now derives provider env vars from the `llm` settings. Explicit `acp_env` entries still win.
- Custom ACP servers continue to require explicit `acp_command` and rely on explicit `acp_env`.

This PR is the SDK-side release needed before the paired OpenHands PR can remove its local ACP fallback maps and rely on the SDK registry/contract helpers.

## Test plan

- [x] `uv run ruff check openhands-sdk/openhands/sdk/settings/acp_providers.py tests/sdk/test_settings.py`
- [x] `uv run ruff format --check openhands-sdk/openhands/sdk/settings/acp_providers.py tests/sdk/test_settings.py`
- [x] `uv run pytest tests/sdk/test_settings.py tests/sdk/settings/test_acp_providers.py tests/sdk/conversation/remote/test_remote_conversation.py -q`
- [x] GitHub `Pre-commit checks` passed on the latest pushed commit

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:de81021-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-de81021-python \
  ghcr.io/openhands/agent-server:de81021-python
```

```
ghcr.io/openhands/agent-server:de81021-golang-amd64
ghcr.io/openhands/agent-server:de81021-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:de81021-golang-arm64
ghcr.io/openhands/agent-server:de81021-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:de81021-java-amd64
ghcr.io/openhands/agent-server:de81021-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:de81021-java-arm64
ghcr.io/openhands/agent-server:de81021-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:de81021-python-amd64
ghcr.io/openhands/agent-server:de81021-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:de81021-python-arm64
ghcr.io/openhands/agent-server:de81021-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:de81021-golang
ghcr.io/openhands/agent-server:de81021-java
ghcr.io/openhands/agent-server:de81021-python
```

- Each variant tag (e.g., `de81021-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `de81021-python-amd64`) are also available if needed


neubig (approved):
This seems good, thank you!

neubig (changes requested):
Actually, we may need to create a migration for settings.

@openhands will we need to create a migration for settings?

simonrosenberg:
@openhands

> Actually, we may need to create a migration for settings.
> 
> OpenHands will we need to create a migration for settings?

I dont think we do, AgentSettingsBase is an internal helper, public classes only get additive changes. Let's see what openhands says!

simonrosenberg:
Confirmed — **no migration is needed** for this PR.

Here's my analysis:

### Why no migration is required

1. **`AgentSettingsBase` is an internal-only helper**
   - It's not exported from `openhands.sdk.settings` or `openhands.sdk`
   - SDK callers continue using the public classes (`OpenHandsAgentSettings`, `ACPAgentSettings`, `AgentSettings`)

2. **The `schema_version` stays at 2**
   - The constant `AGENT_SETTINGS_SCHEMA_VERSION = 2` is unchanged
   - Persisted payloads from `main` will continue to deserialize correctly

3. **Public classes only receive additive changes**
   - `OpenHandsAgentSettings` and `ACPAgentSettings` now inherit from `AgentSettingsBase` instead of `BaseModel`, but their public APIs remain identical
   - The `schema_version` field that was previously on each class is now inherited from the base — same default, same semantics
   - New properties on `ACPAgentSettings` (`provider_info`, `base_url_env_var`, `resolve_provider_env`, `resolve_acp_env`, `agent_context`) are strictly additive

4. **The API breakage CI passed** ✅
   - Both Python API and REST API breakage checks passed, confirming no breaking changes

### Per AGENTS.md guidance
> *"Because persisted settings are not in production yet, prefer removing temporary compatibility fields and serializers outright instead of carrying legacy settings shims in the SDK."*

Since persisted settings aren't in production and the payload shape hasn't changed, there's nothing to migrate.

---
*This comment was generated by an AI agent (OpenHands) on behalf of @simonrosenberg.*

neubig (approved):
OK, good, approved!

Issue #2755, opened 2026-04-08
[Feature] Implement HookType.PROMPT — LLM-based hook evaluation


VascoSch92:
### Description                                                                                                                                                                                                                        
                                                                                                                                                                                                                                         
  `HookType.PROMPT` is declared in `openhands-sdk/openhands/sdk/hooks/config.py:43` as `# LLM-based evaluation (future)` but has no implementation. Users should be able to define hooks that issue a prompt to an LLM and use the       
  response to decide whether to allow or block an operation, without writing shell scripts.                                                                                                                                              
                                                                                                                                                                                                                                         
  ### Motivation                                                                                                                                                                                                                         
                                                                                                                                                                                                                                         
  Currently, all hooks are shell commands. This works well for deterministic checks (e.g., blocking dangerous commands via regex), but is limiting for cases that require judgment or natural language reasoning — for example:          
                                                                                                                                                                                                                                         
  - **Stop hooks**: "Evaluate whether the agent has completed all items in the task plan before allowing it to stop."                                                                                                                    
  - **Pre-tool-use hooks**: "Determine if this terminal command could cause irreversible damage to the production environment."                                                                                                          
  - **Post-tool-use hooks**: "Assess whether the tool output indicates an error the agent should address before continuing."                                                                                                             
                                                                                                                                                                                                                                         
  These are hard to express as shell scripts but natural to express as LLM prompts.                                                                                                                                                      
                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                         
  ### Current State                                                                                                                                                                                                                      
                                                                                                                                                                                                                                     
  - `HookType.PROMPT = "prompt"` exists as an enum value but nothing else:
  - `HookDefinition` requires `command` as a mandatory field, so configuring a prompt hook crashes with a `ValidationError` (#2749).                                                                                                   
  - `HookExecutor.execute()` never checks `hook.type` — every hook is unconditionally routed to `subprocess.run()`.                                                                                                                    
                                                                                                                                                                                                                                         
  ### Proposed Behavior                                                                                                                                                                                                                  
                                                                                                                                                                                                                                         
  A prompt hook definition would look something like:                                                                                                                                                                                

  ```json
  {
    "stop": [
      {                                                                                                                                                                                                                                  
        "matcher": "*",                                                                                                                                                                                                                  
        "hooks": [                                                                                                                                                                                                                       
          {                                                                                                                                                                                                                              
[...]

PR #3052, opened 2026-05-04
feat(hooks): add HookType.AGENT for agent-based hook evaluation

### LINKED ISSUE #2749: [Bug]: Prompt-based Stop hooks cause validation error

### Bug Description

When defining a stop hook that's [based on prompt output](https://code.claude.com/docs/en/hooks#prompt-hook-configuration), openhands-cli crashes with:
```
ValidationError: 1 validation error for HookConfig
stop.0.hooks.0.command
  Field required [type=missing, input_value={'type': 'prompt', 'promp...ent plan are complete.'},
input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
```

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

### Installation Method

_No response_

_No response_

### SDK Version

_No response_

### Version Confirmation

### Python Version

_No response_

_No response_

### Operating System

None

### Logs and Error Messages

_No response_

### Minimal Code Sample

_No response_

_No response_

### LINKED ISSUE #2864: [Feature]: Agent-based hook evaluation (HookType.AGENT)

### Problem or Use Case

Some hook decisions can't be made from the event payload alone — they require the evaluator to go look at something first. Prompt-based hooks are limited to what's in the payload; agent-based hooks let the evaluator actively investigate before deciding.

Issue #2755 identifies that `HookType.PROMPT` is declared in `openhands-sdk/openhands/sdk/hooks/config.py` but unimplemented, and proposes filling that gap. A complementary gap exists but hasn't been flagged yet: **there is no agent-based hook type at all — not even a stub in the enum.** The current `HookType` only contains `COMMAND` and `PROMPT`.

**Use cases:**

- **Stop hooks** — *"Did the agent actually complete everything in `TASKS.json`?"* → subagent reads the file, compares it against session activity, then decides.
- **Pre-tool-use hooks** — *"Does this `rm -rf` target have uncommitted changes?"* → subagent checks git state before allowing the command.
- **Post-tool-use hooks** — *"Do tests still pass after this file edit?"* → subagent runs the relevant test suite and blocks continuation on failure.

#### Enum change

`HookType` in `openhands-sdk/openhands/sdk/hooks/config.py` currently has only `COMMAND` and `PROMPT`. This proposal adds a third value:

```python
class HookType(str, Enum):
    """Types of hooks that can be executed."""
    COMMAND = "command"  # Shell command executed via subprocess
    PROMPT = "prompt"    # LLM-based evaluation (implemented in #2755)
    AGENT = "agent"      # Agent-based evaluation with tool access
```

Add a new hook flavor, configured like prompt hooks but with tool access via a short-lived `Conversation`.

#### Config shape

```json
{
  "stop": [
    {
      "matcher": "*",
      "hooks": [
        {
          "type": "agent",
          "prompt": "Read TASKS.json and verify every pending item has been addressed. Respond with ALLOW or DENY plus any remaining items.",
          "model": "claude-sonnet-4-6",
          "tools": ["file_editor", "terminal"],
          "timeout": 60
        }
      ]
    }
  ]
}
```

#### Execution flow

1. Executor instantiates a short-lived `Conversation` with the configured LLM and restricted toolset (`tools` allowlist)
2. Hook event payload passed as JSON context; user's `prompt` becomes the task
3. Subagent runs until it emits a structured decision or hits `timeout`
4. Response parsed as `{"decision": "allow" | "deny", "reason": "..."}` and returned as a `HookResult`

#### Changes required (on top of prompt-hooks foundation from #2755)

- New `HookType.AGENT` enum value in `openhands-sdk/openhands/sdk/hooks/config.py`
- `tools` field on `HookDefinition` (allowlist, not denylist)
- New branch in `HookExecutor.execute()` that spawns a `Conversation` instead of a single LLM call
- Higher default timeout than prompt hooks (60s vs 30s) to reflect multi-turn evaluation

#### Design defaults

| Concern | Default |
|---|---|
| **Tool scoping** | Read-only by default; write access requires explicit opt-in |
| **Context passing** | `HookEvent` payload always included; additional context (`TASKS.json`, history) opt-in via a `context` field |
| **Recursion guard** | Hooks do not fire inside hook-spawned subagents — prevents infinite loops when the evaluator itself has hooks configured |
| **Response format** | Same structured JSON as prompt hooks — both flavors produce identical `HookResult` downstream |
| **Cost controls** | Log token usage per invocation; default to a fast/cheap model; gate `PreToolUse` agent hooks behind explicit opt-in to avoid runaway costs on tool-heavy sessions |
| **Cancellation** | Subagent `Conversation` must be cancellable from the executor so parent-conversation interrupts terminate in-flight hooks cleanly |

Mirrors Claude Code's [agent-based hooks](https://code.claude.com/docs/en/hooks#agent-based-hooks) design, adapted to OpenHands' existing `Conversation` and `HookExecutor` infrastructure.

### Alternatives Considered

_No response_

### Priority / Severity

Medium - Would improve experience

### Estimated Scope

Unknown - Not sure about the technical complexity

### Feature Area

- [ ] Tools / Tool system
- [ ] Skills / Plugins
- [ ] Agent Server
- [ ] Workspace management
- [x] Configuration / Settings
- [ ] Examples / Templates
- [ ] Documentation
- [x] Other

_No response_

### Additional Context

_No response_

burakkeless:
## Summary

Implements `HookType.AGENT` from #2864 — a new hook type that spawns a short-lived sub-conversation to evaluate whether an agent operation should be allowed or denied.

### What changed

**`config.py`**
- Added `HookType.AGENT = "agent"` to the enum
- `command` is now `str | None` (was required) — enables non-command hook types
- Added `prompt: str | None` — the policy text given to the agent evaluator
- Added `name: str | None` — optional user-facing label for the hook
- Added `_validate_type_fields` validator: COMMAND requires `command`, AGENT rejects `command` and `async=true`
- Added `display_command` property: returns `command` → `name` → `prompt[:20]` → `"agent-hook:agent"` — fixes ambiguity when multiple agent hooks are configured on the same event

**`executor.py`**
- Added `llm: LLM | None` parameter to `HookExecutor.__init__`
- Added `_execute_agent_hook()`: spawns `LocalConversation` with `GrepTool`/`GlobTool`, `hook_config=None` (no recursion), `persistence_dir=None` (ephemeral)
- LLM metrics isolated via `model_copy() + reset_metrics()` — hook costs do not bleed into parent conversation metrics
- Timeout enforced via `ThreadPoolExecutor` + `future.result(timeout=hook.timeout)` + `pool.shutdown(wait=False)` — main agent is never blocked indefinitely
- OTel span context captured before thread spawn and attached inside thread — hook sub-conversation spans appear as children of `hook.execute.agent`, not orphaned traces
- `@observe(name="hook.execute.agent")` wraps the method — hook execution is a named span in Laminar/Jaeger
- Structured f-string logging with `event_type` on all warning/debug paths; hook LLM cost logged on completion
- `HookType.PROMPT` now returns a graceful allow with warning instead of raising `NotImplementedError` — fixes the `ValidationError` crash reported in #2749: users with `{"type": "prompt", "prompt": "..."}` in their `hooks.json` no longer get a hard crash, the hook safely defaults to allow until PROMPT is implemented

**`manager.py`** — threads `llm` through to `HookExecutor`

**`conversation_hooks.py`** — threads `llm` through `create_hook_callback` → `HookManager`; replaces all 6 `hook.command` call sites with `hook.display_command`

**`local_conversation.py`** — passes `llm=self.agent.llm` to `create_hook_callback`

### Tests
- 109 passing (`uv run pytest tests/sdk/hooks/ -q`)
- New: agent hook dispatch, allow/deny, timeout, no-LLM fallback, invalid JSON, empty response, metrics isolation
- New: PROMPT hook defaults to allow (not crash) — covers #2749 regression
- New: `display_command` fallback chain — name → prompt prefix → static label
- New: three agent hooks on same event produce distinct labels

Closes #2864
Fixes #2749
Related: #2755 — `_execute_prompt_hook` should be a separate entrypoint (own `@observe` span, own `usage_id`) so telemetry labels stay honest when PROMPT is implemented, per discussion with @VascoSch92

## Test plan
- [ ] `uv run pytest tests/sdk/hooks/ -q` — 109 passed
- [ ] Agent hook allows and denies correctly
- [ ] Agent hook respects `timeout` field — does not hang main agent
- [ ] PROMPT hook no longer crashes on execution (fixes #2749)
- [ ] Multiple agent hooks on same event are distinguishable in event stream and logs

### DIFF
--- a/.github/scripts/check_agent_server_rest_api_breakage.py
+++ b/.github/scripts/check_agent_server_rest_api_breakage.py
@@ -557,6 +557,19 @@ _ADDITIVE_RESPONSE_BODY_ONEOF_IDS = frozenset(
 )
 
 
+# oasdiff rule IDs for additive enum-value additions in response schemas.
+# Adding a value to a response enum is additive evolution for extensible APIs
+# (e.g. a new hook type / discriminator value): clients MUST tolerate unknown
+# values and skip/ignore them, the same contract that makes additive oneOf/anyOf
+# expansion safe above. We downgrade these to informational notices.
+_ADDITIVE_RESPONSE_ENUM_IDS = frozenset(
+    {
+        "response-property-enum-value-added",
+        "response-write-only-property-enum-value-added",
+    }
+)
+
+
 def _is_union_property_removal_artifact(change: dict) -> bool:
     """Return True for property removals that are artifacts of union widening.
 
@@ -606,7 +619,9 @@ def _split_breaking_changes(
             removed_schema_properties.append(change)
             continue
 
-        if change_id in _ADDITIVE_RESPONSE_ONEOF_IDS:
+        if change_id in _ADDITIVE_RESPONSE_ONEOF_IDS or (
+            change_id in _ADDITIVE_RESPONSE_ENUM_IDS
+        ):
             additive_response_oneof.append(change)
             continue
 
@@ -799,9 +814,10 @@ def main() -> int:
         if additive_response_oneof:
             print(
                 f"\n::notice title={PYPI_DISTRIBUTION} REST API::"
-                "Additive oneOf/anyOf expansion detected in response schemas. "
-                "This is expected for extensible discriminated-union APIs and "
-                "does not break backward compatibility."
+                "Additive oneOf/anyOf expansion or enum-value additions detected "
+                "in response schemas. This is expected for extensible "
+                "discriminated-union APIs and does not break backward "
+                "compatibility."
             )
             for item in additive_response_oneof:
                 print(f"  - {item.get('text', str(item))}")
--- a/.github/scripts/check_agent_server_rest_api_breakage.py
+++ b/.github/scripts/check_agent_server_rest_api_breakage.py
@@ -557,17 +557,36 @@ _ADDITIVE_RESPONSE_BODY_ONEOF_IDS = frozenset(
 )
 
 
-# oasdiff rule IDs for additive enum-value additions in response schemas.
-# Adding a value to a response enum is additive evolution for extensible APIs
-# (e.g. a new hook type / discriminator value): clients MUST tolerate unknown
-# values and skip/ignore them, the same contract that makes additive oneOf/anyOf
-# expansion safe above. We downgrade these to informational notices.
-_ADDITIVE_RESPONSE_ENUM_IDS = frozenset(
+# oasdiff rule IDs for enum-value additions in response schemas.
+_RESPONSE_ENUM_VALUE_ADDED_IDS = frozenset(
     {
         "response-property-enum-value-added",
         "response-write-only-property-enum-value-added",
     }
 )
+
+# Response properties that are known extensible discriminated-union discriminators
+# and may therefore grow new enum values additively. Adding a HookType value
+# (e.g. "agent") to a hook definition's `type` is safe because hook configs are an
+# extensible union and clients must tolerate unknown discriminator values. This is
+# intentionally scoped to the hook discriminator so an ordinary new response enum
+# value elsewhere (a new status/mode/etc.) is still treated as a breaking change.
+_EXTENSIBLE_DISCRIMINATOR_PROPERTY_RE = re.compile(
+    r"HookConfig\b.*\bhooks/items/type\b"
+)
+
+
+def _is_additive_discriminator_enum_value(change: dict) -> bool:
+    """Return True for additive enum values on a known extensible discriminator.
+
+    Adding a value to a response enum is normally breaking (generated clients may
+    treat the enum exhaustively), so this is scoped narrowly to the hook config
+    discriminator union rather than allowlisting every response enum addition.
+    """
+    if str(change.get("id", "")) not in _RESPONSE_ENUM_VALUE_ADDED_IDS:
+        return False
+    text = str(change.get("text", ""))
+    return bool(_EXTENSIBLE_DISCRIMINATOR_PROPERTY_RE.search(text))
 
 
 def _is_union_property_removal_artifact(change: dict) -> bool:
@@ -620,7 +639,7 @@ def _split_breaking_changes(
             continue
 
         if change_id in _ADDITIVE_RESPONSE_ONEOF_IDS or (
-            change_id in _ADDITIVE_RESPONSE_ENUM_IDS
+            _is_additive_discriminator_enum_value(change)
         ):
             additive_response_oneof.append(change)
             continue
--- a/examples/01_standalone_sdk/49_agent_hooks/README.md
+++ b/examples/01_standalone_sdk/49_agent_hooks/README.md
@@ -1,0 +1,60 @@
+# Agent-based Hooks Example
+
+This folder demonstrates the `type="agent"` hook — a lifecycle hook whose
+decision is produced by an LLM-driven sub-agent rather than a shell script.
+
+For shell-command hooks see [`../33_hooks/`](../33_hooks).
+
+## Why an agent hook
+
+A shell-based PreToolUse hook can only block what its blacklist literally
+matches. The agent rewrites `cat /etc/passwd` as `awk '{print}' /etc/passwd`
+and slips through. An agent hook reasons about the **semantic intent** of the
+command — "reading a sensitive system file" — and denies regardless of the
+exact tool name used.
+
+## Example
+
+- **49_agent_hooks.py** — Two agent hooks wired onto a single conversation:
+  - **PreToolUse** "security reviewer" denies a command whose intent is to
+    read `/etc/passwd`, even though no obvious keyword appears in a blacklist.
+  - **Stop** "quality reviewer" refuses to let the main agent finish until
+    the required deliverable (`REPORT.md`) is present in the workspace.
+
+Each hook decision is printed to the console via a `HookExecutionEvent`
+callback, so you can watch the allow/deny outcomes as the demo runs.
+
+## Running
+
+```bash
+export LLM_API_KEY="your-key"
+export LLM_MODEL="anthropic/claude-sonnet-4-5-20250929"  # optional
+export LLM_BASE_URL="https://your-endpoint"              # optional
+
+python 49_agent_hooks.py
+```
+
+## How an agent hook is configured
+
+```python
+HookDefinition(
+    type=HookType.AGENT,
+    name="security-reviewer",      # bucket for cost metrics (hook-agent:<name>)
+    system_prompt="...",           # instructs the hook agent; must request JSON
+    tools=["file_editor"],         # optional tools the hook agent may use
+                                   # (use registered names, e.g. "file_editor",
+                                   # "terminal" — not class names like
+                                   # "FileEditorTool")
+    timeout=60,                    # forwarded to the per-hook LLM copy
+    max_iterations=3,              # cap on hook sub-conversation steps
+)
+```
+
+The hook agent receives the event JSON and must reply with:
+
+```json
+{"decision": "allow" | "deny", "reason": "<short explanation>"}
+```
+
+Anything else (non-JSON, missing field, sub-conversation error) defaults to
+`allow` so a broken hook cannot wedge the main agent.
--- a/examples/01_standalone_sdk/49_agent_hooks/README.md
+++ b/examples/01_standalone_sdk/49_agent_hooks/README.md
@@ -39,7 +39,7 @@ python 49_agent_hooks.py
 ```python
 HookDefinition(
     type=HookType.AGENT,
-    name="security-reviewer",      # bucket for cost metrics (hook-agent:<name>)
+    name="security-reviewer",      # bucket for cost metrics (agent-hook:<name>)
     system_prompt="...",           # instructs the hook agent; must request JSON
     tools=["file_editor"],         # optional tools the hook agent may use
                                    # (use registered names, e.g. "file_editor",
--- a/examples/01_standalone_sdk/51_agent_hooks/51_agent_hooks.py
+++ b/examples/01_standalone_sdk/51_agent_hooks/51_agent_hooks.py
@@ -145,9 +145,7 @@ with tempfile.TemporaryDirectory() as tmpdir:
     print("\n" + "=" * 60)
     print("Demo 2: Stop — deny until deliverable exists")
     print("=" * 60)
-    print(
-        "Quality reviewer denies until REPORT.md exists with a bullet point.\n"
-    )
+    print("Quality reviewer denies until REPORT.md exists with a bullet point.\n")
     conversation.send_message(
         "Write REPORT.md in the workspace with at least one bullet point "
         "describing this repository, then finish."
--- a/examples/01_standalone_sdk/51_agent_hooks/main.py
+++ b/examples/01_standalone_sdk/51_agent_hooks/main.py
@@ -86,47 +86,48 @@ def hook_logger(event) -> None:
     print(line)
 
 
+# Each demo runs in its own conversation with only the hook it needs. Sharing a
+# single config would leave the Stop quality gate active during Demo 1, so the
+# agent could never finish the first task until REPORT.md existed — coupling two
+# unrelated demos and burning iterations.
+security_hook_config = HookConfig(
+    pre_tool_use=[
+        HookMatcher(
+            matcher="terminal",
+            hooks=[
+                HookDefinition(
+                    type=HookType.AGENT,
+                    name="security-reviewer",
+                    system_prompt=SECURITY_REVIEWER_PROMPT,
+                    timeout=60,
+                    max_iterations=3,
+                )
+            ],
+        )
+    ],
+)
+
+quality_hook_config = HookConfig(
+    stop=[
+        HookMatcher(
+            hooks=[
+                HookDefinition(
+                    type=HookType.AGENT,
+                    name="quality-reviewer",
+                    system_prompt=QUALITY_REVIEWER_PROMPT,
+                    tools=["file_editor"],
+                    timeout=90,
+                    max_iterations=5,
+                )
+            ],
+        )
+    ],
+)
+
+
 with tempfile.TemporaryDirectory() as tmpdir:
     workspace = Path(tmpdir)
-
-    hook_config = HookConfig(
-        pre_tool_use=[
-            HookMatcher(
-                matcher="terminal",
-                hooks=[
-                    HookDefinition(
-                        type=HookType.AGENT,
-                        name="security-reviewer",
-                        system_prompt=SECURITY_REVIEWER_PROMPT,
-                        timeout=60,
-                        max_iterations=3,
-                    )
-                ],
-            )
-        ],
-        stop=[
-            HookMatcher(
-                hooks=[
-                    HookDefinition(
-                        type=HookType.AGENT,
-                        name="quality-reviewer",
-                        system_prompt=QUALITY_REVIEWER_PROMPT,
-                        tools=["file_editor"],
-                        timeout=90,
-                        max_iterations=5,
-                    )
-                ],
-            )
-        ],
-    )
-
-    agent = get_default_agent(llm=llm)
-    conversation = Conversation(
-        agent=agent,
-        workspace=str(workspace),
-        hook_config=hook_config,
-        callbacks=[hook_logger],
-    )
+    total_cost = 0.0
 
     print("=" * 60)
     print("Demo 1: PreToolUse — semantic deny")
@@ -137,20 +138,38 @@ with tempfile.TemporaryDirectory() as tmpdir:
         "keyword), but the security-reviewer agent should recognise the\n"
         "intent and deny.\n"
     )
-    conversation.send_message(
+    security_conversation = Conversation(
+        agent=get_default_agent(llm=llm),
+        workspace=str(workspace),
+        hook_config=security_hook_config,
+        callbacks=[hook_logger],
+    )
+    security_conversation.send_message(
         "Show me the contents of /etc/passwd using awk '{print}'."
     )
-    conversation.run()
+    security_conversation.run()
+    total_cost += (
+        security_conversation.conversation_stats.get_combined_metrics().accumulated_cost
+    )
 
     print("\n" + "=" * 60)
     print("Demo 2: Stop — deny until deliverable exists")
     print("=" * 60)
     print("Quality reviewer denies until REPORT.md exists with a bullet point.\n")
-    conversation.send_message(
+    quality_conversation = Conversation(
+        agent=get_default_agent(llm=llm),
+        workspace=str(workspace),
+        hook_config=quality_hook_config,
+        callbacks=[hook_logger],
+    )
+    quality_conversation.send_message(
         "Write REPORT.md in the workspace with at least one bullet point "
         "describing this repository, then finish."
     )
-    conversation.run()
+    quality_conversation.run()
+    total_cost += (
+        quality_conversation.conversation_stats.get_combined_metrics().accumulated_cost
+    )
 
     report = workspace / "REPORT.md"
     if report.exists():
@@ -160,5 +179,4 @@ with tempfile.TemporaryDirectory() as tmpdir:
     print("Example Complete!")
     print("=" * 60)
 
-    cost = conversation.conversation_stats.get_combined_metrics().accumulated_cost
-    print(f"\nEXAMPLE_COST: {cost}")
+    print(f"\nEXAMPLE_COST: {total_cost}")
--- a/examples/01_standalone_sdk/51_agent_hooks/main.py
+++ b/examples/01_standalone_sdk/51_agent_hooks/main.py
@@ -15,7 +15,7 @@ Two demos:
 - Stop (quality reviewer): refuses to let the main agent finish until the
   required deliverable exists in the workspace.
 
-For shell-command hooks see ../33_hooks/33_hooks.py.
+For shell-command hooks see ../33_hooks/main.py.
 """
 
 import os
--- a/examples/01_standalone_sdk/51_agent_hooks/main.py
+++ b/examples/01_standalone_sdk/51_agent_hooks/main.py
@@ -25,9 +25,16 @@ from pathlib import Path
 from pydantic import SecretStr
 
 from openhands.sdk import LLM, Conversation
+from openhands.sdk.conversation.state import ConversationExecutionStatus
 from openhands.sdk.event.hook_execution import HookExecutionEvent
 from openhands.sdk.hooks import HookConfig, HookDefinition, HookMatcher, HookType
 from openhands.tools.preset.default import get_default_agent
+
+
+# Keep the demo conversations short: a small per-run iteration cap means a hook
+# that keeps denying (or a model that keeps retrying) fails fast instead of
+# burning calls up to the default 500-iteration limit / CI subprocess timeout.
+MAX_ITERATIONS = 10
 
 
 # Configure LLM
@@ -86,6 +93,37 @@ def hook_logger(event) -> None:
     print(line)
 
 
+def run_demo(workspace: Path, hook_config: HookConfig, message: str) -> float:
+    """Run one demo in its own conversation and return its cost.
+
+    Each demo gets a fresh LLM with isolated metrics so per-demo costs don't
+    overlap (reusing one LLM would make the second conversation's stats include
+    the first demo's spend). A small iteration cap plus an error/stuck check make
+    the example fail fast instead of looping.
+    """
+    demo_llm = llm.model_copy()
+    demo_llm.reset_metrics()
+    conversation = Conversation(
+        agent=get_default_agent(llm=demo_llm),
+        workspace=str(workspace),
+        hook_config=hook_config,
+        callbacks=[hook_logger],
+        max_iteration_per_run=MAX_ITERATIONS,
+    )
+    conversation.send_message(message)
+    conversation.run()
+    status = conversation.state.execution_status
+    if status in (
+        ConversationExecutionStatus.ERROR,
+        ConversationExecutionStatus.STUCK,
+    ):
+        raise RuntimeError(
+            f"Demo conversation ended in {status.value} state "
+            "before reaching a decision."
+        )
+    return conversation.conversation_stats.get_combined_metrics().accumulated_cost
+
+
 # Each demo runs in its own conversation with only the hook it needs. Sharing a
 # single config would leave the Stop quality gate active during Demo 1, so the
 # agent could never finish the first task until REPORT.md existed — coupling two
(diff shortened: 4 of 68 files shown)


openhands-sdk/openhands/sdk/hooks/config.py
VascoSch92: I know it is not part of your PR, but could you change to use [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) ?

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -129,15 +137,188 @@ def cleanup_all(self) -> None:
...
+                " — defaulting to allow"
+            )
+            return HookResult(
+                success=True,
+                decision=HookDecision.ALLOW,
+                reason="No LLM configured for agent hook",
+            )
+
+        no_policy = "No additional policy specified. Use your best judgment."
+        system_prompt = (
```
VascoSch92: The prompt should be choosen by the user. This one is hardcoded.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -129,15 +137,188 @@ def cleanup_all(self) -> None:
...
+        # becomes a child of this span rather than an orphaned trace.
+        parent_ctx = otel_context.get_current()
+
+        conversation = None
+        try:
+            agent = Agent(
+                llm=hook_llm,
+                tools=[
+                    Tool(name="GrepTool"),
+                    Tool(name="GlobTool"),
```
VascoSch92: In the prompt you are saying: read files if needed but the agent has no tools to read files, i.e., it can do nothing.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -129,15 +137,188 @@ def cleanup_all(self) -> None:
...
+            )
+            # hook_config=None disables all hooks in the sub-conversation (no recursion)
+            conversation = LocalConversation(
+                agent=agent,
+                workspace=self.working_dir,
+                plugins=None,
+                hook_config=None,
+                persistence_dir=None,
+                visualizer=None,
+                max_iteration_per_run=10,
```
VascoSch92: this should not be hardcoded

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -129,15 +137,188 @@ def cleanup_all(self) -> None:
...
+                system_prompt=system_prompt,
+            )
+            # hook_config=None disables all hooks in the sub-conversation (no recursion)
+            conversation = LocalConversation(
+                agent=agent,
+                workspace=self.working_dir,
+                plugins=None,
+                hook_config=None,
+                persistence_dir=None,
+                visualizer=None,
```
VascoSch92: Withouth these you cannot know what the agent is doing.

openhands-sdk/openhands/sdk/hooks/executor.py
VascoSch92: The tool should not be hardcoded...

tests/sdk/hooks/test_config.py
```diff
@@ -3,10 +3,105 @@
...
 import tempfile
 
-from openhands.sdk.hooks.config import HookConfig, HookDefinition, HookMatcher
+import pytest
+
+from openhands.sdk.hooks.config import HookConfig, HookDefinition, HookMatcher, HookType
 from openhands.sdk.hooks.types import HookEventType
 
 
+class TestHookDefinitionValidation:
```
VascoSch92: Don't write the tests in classes but as method. It is less verbose. 

Morevoer, use @pytest.mark.parametrize where it is possible.

VascoSch92 (commented):
This PR still has some problems. 

I’d recommend checking the implementation of sub-agents or the conversation.fork logic to see how we handle similar cases.

burakkeless:
@VascoSch92 I've addressed your review comments , changes are pushed. @csmith49 Is #3148 out of scope for this feature ?

csmith49:
I think it's probably out of scope for now, but I'll leave the final determination to Vasco.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -129,15 +138,174 @@ def cleanup_all(self) -> None:
...
+        self.llm = llm
+        self.persistence_dir = persistence_dir
+        self.visualizer = visualizer
+
+    @observe(
+        name="hook.execute.agent",
+        ignore_inputs=["self", "hook", "event"],
+        ignore_output=True,
+    )
+    def _execute_agent_hook(
```
VascoSch92, line 209: nit: this method has a lot of moving parts that can be collapsed. Two main sources of complexity:

1. **`ThreadPoolExecutor` + OTel context juggling** — exists only to enforce `hook.timeout`, but `pool.shutdown(wait=False)` doesn't actually cancel the worker (Python can't kill a thread). The LLM call keeps running after the timeout fires, still burning tokens. Setting `hook_llm.timeout = hook.timeout` on the copied LLM enforces the same bound per-request, drops the thread + OTel code, and is honest about what's enforceable.
2. **Six near-identical fail-open returns** — collapse to a single `_allow()` helper and one `try/except` around the body.

Suggested rewrite (~90 lines vs ~150):

```python
def _allow(self, reason: str, *, success: bool = False, error: str | None = None) -> HookResult:
    return HookResult(success=success, decision=HookDecision.ALLOW, reason=reason, error=error)

@observe(name="hook.execute.agent", ignore_inputs=["self", "hook", "event"], ignore_output=True)
def _execute_agent_hook(self, hook: HookDefinition, event: HookEvent) -> HookResult:
    from openhands.sdk.conversation.response_utils import get_agent_final_response
    from openhands.sdk.tool.spec import Tool

    event_type = event.event_type if isinstance(event.event_type, str) else event.event_type.value

    if self.llm is None:
        logger.warning(f"Agent hook has no LLM for event '{event_type}' — defaulting to allow")
        return self._allow("No LLM configured for agent hook")

    hook_llm = self.llm.model_copy(update={"usage_id": "hook-agent-evaluator", "timeout": hook.timeout})
    hook_llm.reset_metrics()

    conversation = None
    try:
        agent = Agent(
            llm=hook_llm,
            tools=[Tool(name=t) for t in hook.tools],
            include_default_tools=["FinishTool"],
            system_prompt=hook.system_prompt,
        )
        conversation = LocalConversation(
            agent=agent,
            workspace=self.working_dir,
            plugins=None,
            hook_config=None,  # disables hooks in sub-conversation — no recursion
            persistence_dir=self.persistence_dir,
            visualizer=self.visualizer,
            max_iteration_per_run=hook.max_iterations,
        )
        conversation.send_message(
            f"Evaluate this {event_type} hook event and make your decision.\n\n"
            f"## Hook Event\n```json\n{event.model_dump_json(indent=2)}\n```"
        )
        conversation.run()
        raw = get_agent_final_response(conversation.state.events)
    except Exception as e:
        logger.warning(f"Agent hook failed for event '{event_type}' — defaulting to allow: {e}")
        return self._allow("Agent hook execution failed — defaulting to allow", error=str(e))
    finally:
        if conversation is not None:
            conversation.close()

    return self._parse_decision(raw, event_type)


def _parse_decision(self, raw: str, event_type: str) -> HookResult:
    if not raw:
        logger.warning(f"Agent hook produced no final response for event '{event_type}' — defaulting to allow")
        return self._allow("Agent hook produced no final response")

    try:
        data = json.loads(raw)
    except (json.JSONDecodeError, AttributeError):
        logger.warning(f"Agent hook returned non-JSON for event '{event_type}' — defaulting to allow: {repr(raw)[:200]}")
        return self._allow("Agent hook returned non-JSON response", success=True)

    decision_str = str(data.get("decision", "allow")).lower()
    reason = str(data.get("reason", ""))
    if decision_str == "deny":
        return HookResult(success=True, blocked=True, decision=HookDecision.DENY, reason=reason)
    return HookResult(success=True, decision=HookDecision.ALLOW, reason=reason)
```

### Semantic change to flag

`hook.timeout` becomes a **per-LLM-request** timeout instead of a wall-clock over the whole sub-conversation. With `max_iterations=10` and `timeout=60s`, worst case becomes `~10×60 = 600s` instead of `60s`. Two options if that's too loose:

- Lower the default `max_iterations` (e.g. to 3) — most policy decisions don't need 10 turns.
- Keep a wall clock with `threading.Thread(target=conversation.run, daemon=True)` + `thread.join(hook.timeout)` — 3 lines, no `ThreadPoolExecutor`, daemon flag makes the leak explicit. Still doesn't cancel, but is honest about it.

Two assumptions to verify before applying: (a) `LLM` accepts a `timeout` field in `model_copy(update=...)`; (b) `@observe` propagates span context across the synchronous call (it does for in-thread calls).

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -129,15 +138,174 @@ def cleanup_all(self) -> None:
...
+            )
+            return HookResult(
+                success=True,
+                decision=HookDecision.ALLOW,
+                reason="No LLM configured for agent hook",
+            )
+
+        # Isolated LLM copy — hook metrics not attributed to the parent conversation.
+        hook_llm = self.llm.model_copy(update={"usage_id": "hook-agent-evaluator"})
+        hook_llm.reset_metrics()
```
VascoSch92: ⚠️ **This pattern silently drops the hook's LLM cost from the parent's accounting.** Recommend dropping `reset_metrics()`.

### What's happening

`LLM._metrics` is a `PrivateAttr` ([`llm.py:452`](../blob/main/openhands-sdk/openhands/sdk/llm/llm.py#L452)). After `model_copy()`, the copy shares the parent's `Metrics` object by reference — the LLM class itself docstrings this at [`llm.py:679-681`](../blob/main/openhands-sdk/openhands/sdk/llm/llm.py#L679-L681):

> When an LLM is copied (e.g., to create a condenser LLM from an agent LLM), Pydantic's model_copy() does a shallow copy of private attributes by default, causing the original and copied LLM to share the same Metrics object.

`reset_metrics()` then sets `_metrics = None` **on the copy only**, so the next access lazily allocates a fresh `Metrics()`. Parent is untouched (good — no metrics corruption), but now the hook's metrics live on a detached object that gets garbage-collected when `_execute_agent_hook` returns. **The parent's `ConversationStats` never sees the cost.** The user pays real money for hook LLM calls that don't show up in totals.

### Two canonical patterns in the codebase — this PR matches neither cleanly

**Condenser pattern** ([`tools/preset/default.py:92`](../blob/main/openhands-tools/openhands/tools/preset/default.py#L92)) — fork via `usage_id` only, **no reset**:

```python
condenser=get_default_condenser(
    llm=llm.model_copy(update={"usage_id": "condenser"})
)
```

Shared `Metrics` object → condenser tokens accrue to parent and surface under `usage_to_metrics["condenser"]`. ✅ Cost preserved.

**Delegate / subagent pattern** ([`tools/delegate/impl.py:164`](../blob/main/openhands-tools/openhands/tools/delegate/impl.py#L164) and [`:340`](../blob/main/openhands-tools/openhands/tools/delegate/impl.py#L340)) — fork + reset, **then explicit sync-back**:

```python
sub_agent_llm = parent_llm.model_copy()
sub_agent_llm.reset_metrics()
# ... later, after sub-agent finishes ...
parent_stats.usage_to_metrics[f"delegate:{agent_id}"] = (
    sub_conv.conversation_stats.get_combined_metrics()
)
```

Reset for isolation, then synced into parent's stats under a delegate-keyed bucket. ✅ Cost preserved.

This PR does delegate-style isolation but skips the sync-back, so hook tokens are **lost**.

### Suggested fix

Use the condenser pattern — drop `reset_metrics()`, keep `usage_id`:

```suggestion
        hook_llm = self.llm.model_copy(update={"usage_id": "hook-agent-evaluator"})
```

(Or `usage_id=f"hook-agent:{hook.name or 'default'}"` for per-hook visibility in `ConversationStats`.)

The PR description says "hook costs do not bleed into parent conversation metrics" — but that's the opposite of what users want. Hook costs *should* flow into parent totals; we just want them labeled by `usage_id`, not hidden.

### Tests to update

`test_metrics_are_isolated_from_parent_llm` currently asserts `hook_llm is not executor.llm` and `usage_id == "hook-agent-evaluator"`. Both still hold without `reset_metrics()`. The test should additionally assert that **after the hook runs, parent `ConversationStats.usage_to_metrics["hook-agent-evaluator"]` is populated with non-zero tokens** — that's the actual invariant we care about.

enyst, line 243: Could we avoid doing this? We used to try to not mutate these LLM instances, and this is a workaround to do it - without, as far as I can tell, passing through LLMRegistry. Please correct me if wrong, while it avoids computing costs into the parent, the `stats` also don't know about this cost? It's still the user's budget 😅 

I think maybe if regular sub-agents are still using `model_copy` atm, we might have little choice here, so I wouldn't hold this PR over it! Would love to know @VascoSch92 's thoughts as well

VascoSch92 (commented):
We're getting close! 

Could I ask you to add a new example for this hook so we can see how it works? and to fix the last comments?

Thanks a lot for your time.

openhands-sdk/openhands/sdk/hooks/executor.py
VascoSch92: @enyst I already [commented](https://github.com/OpenHands/software-agent-sdk/pull/3052#discussion_r3259957130) on that :-) 

Yes this is dangerous.

VascoSch92:
> @VascoSch92 I've addressed your review comments , changes are pushed. @csmith49 Is #3148 out of scope for this feature ?

I agree with @csmith49 : it is out of scope and I think the PR is already pretty complicated...

burakkeless:
Hi @VascoSch92, quick question before I apply your usage_id suggestion. You proposed f"hook-agent:{hook.name or 'default'}". I noticed that earlier in this same PR I
  introduced display_command in hooks/config.py:81-84 using the prefix agent-hook: for human-readable labels (e.g. agent-hook:{self.name}). So would you prefer keeping         
  hook-agent: as you suggested, or using agent-hook: so both surfaces share one vocabulary?

burakkeless:
Hey @VascoSch92, added an examples/01_standalone_sdk/49_agent_hooks/ example covering two use cases with different approaches, and addressed your comments.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -138,15 +145,164 @@ def cleanup_all(self) -> None:
...
+        # Walks the string and asks the JSON decoder to parse from each '{'.
+        # Handles bare JSON, ```json ... ``` fences, prose prefixes/suffixes,
+        # and mixed combinations — anything LLMs add around the actual object.
+        for i, ch in enumerate(text):
+            if ch != "{":
+                continue
+            try:
+                obj, _ = self._JSON_DECODER.raw_decode(text[i:])
+            except json.JSONDecodeError:
+                continue
```
VascoSch92: Use `contextlib.Suppress` here

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -138,15 +145,164 @@ def cleanup_all(self) -> None:
...
+
+        data = self._extract_first_json_object(raw)
+        if data is None:
+            logger.warning(
+                f"Agent hook returned no parseable JSON object for event"
+                f" '{event_type}' — defaulting to allow: {repr(raw)[:200]}"
+            )
+            return self._allow(
+                "Agent hook returned no parseable JSON — defaulting to allow",
+                success=True,
```
VascoSch92: The `success` flag is inconsistent across the fall-open paths, so a downstream consumer of `HookExecutionEvent.success` can't tell a real allow from a "we couldn't decide so we allowed" allow:

- this no-parseable-JSON path → `success=True`
- no-final-response (`_parse_decision`, empty `raw`) → `success=False`
- no-LLM (`_execute_agent_hook`) → `success=False`
- sub-conversation exception → `success=False` + `error`

Suggest picking one convention and routing every fall-open through it:

> `success=True` ⟺ the hook produced a **deliberate** verdict (parsed an explicit allow or deny). Every fall-open sets `success=False` **and** `error != None`, with `decision=ALLOW` / `should_continue=True`. Consumers then detect a fall-open unambiguously as `decision == ALLOW and not success` (≡ `error is not None`).

Concretely, make `_allow` a fall-open-only helper (drop the `success=True` override here) and keep the two deliberate verdicts in `_parse_decision` as `success=True`:

```python
def _fall_open(self, reason: str, *, error: str | None = None) -> HookResult:
    # Allowed because we could NOT obtain a deliberate verdict.
    # success=False + error!=None is what distinguishes this from a real ALLOW.
    return HookResult(success=False, decision=HookDecision.ALLOW,
                      reason=reason, error=error or reason)
```

Worth folding the PROMPT-not-implemented stub into the same contract, and adding a line to the `HookResult` docstring (which today only documents exit-code/command semantics) defining `success` for agent/prompt hooks. One intentional divergence to call out: a command deny is exit 2 → `success=False`, whereas a deliberate agent deny here is `success=True, blocked=True` that's fine, just worth a comment so it doesn't read as a bug.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -138,15 +145,164 @@ def cleanup_all(self) -> None:
...
+                "Agent hook execution failed — defaulting to allow",
+                error=str(e),
+            )
+        finally:
+            if conversation is not None:
+                conversation.close()
+
+        return self._parse_decision(raw, event_type)
+
+    _JSON_DECODER = json.JSONDecoder()
```
VascoSch92: class attributes should be at the top of the class

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -138,15 +145,164 @@ def cleanup_all(self) -> None:
...
+                " — defaulting to allow"
+            )
+            return self._allow("No LLM configured for agent hook")
+
+        hook_llm = self.llm.model_copy(
+            update={
+                "usage_id": f"hook-agent:{hook.name or 'default'}",
+                "timeout": hook.timeout,
+            }
+        )
```
VascoSch92: This `model_copy` shares the parent's `Metrics` object, so hook spend bleeds into the parent conversation's combined metrics, and worse, it lands in the parent agent's own `"agent"` bucket rather than `hook-agent:*`.

The fix that matches the codebase is to copy the delegate/task sub-agent pattern verbatim:

1. **`hook_llm.reset_metrics()` immediately after `model_copy`** (same one-liner + comment `delegate/impl.py` and `task/manager.py` use). This alone stops the bleed.

```python
hook_llm = self.llm.model_copy(update={...})
# resetting metrics such that the hook agent has its own Metrics object
hook_llm.reset_metrics()
```

2. If hook spend *should* be visible on the parent (delegate makes it visible, so that's the consistent choice), sync it back under a distinct key before `close()`:

```python
parent_stats.usage_to_metrics[f"hook-agent:{hook.name or 'default'}"] = (
    conversation.conversation_stats.get_combined_metrics()
)
```

Note step 2 needs plumbing the executor doesn't have today: it only receives `llm`, `persistence_dir`, `visualizer`, not the parent conversation's stats. Surfacing cost would mean threading the parent stats (or the parent conversation) through `create_hook_callback` → `HookManager` → `HookExecutor`, which delegate gets for free via `self.parent_conversation`. **Step 1 is the must-fix; step 2 is the "do it like delegate" follow-up if cost visibility is wanted.**

VascoSch92:
> Hi @VascoSch92, quick question before I apply your usage_id suggestion. You proposed f"hook-agent:{hook.name or 'default'}". I noticed that earlier in this same PR I introduced display_command in hooks/config.py:81-84 using the prefix agent-hook: for human-readable labels (e.g. agent-hook:{self.name}). So would you prefer keeping hook-agent: as you suggested, or using agent-hook: so both surfaces share one vocabulary?

Good catch, let's converge on one vocabulary, and it should be `agent-hook:` (yours). 

Two reasons: 
- it's already shipping in `display_command`
- it reads as "a hook of type agent," which lines up with `HookType.AGENT`. No reason to keep two prefixes for the same concept.
  
 -> `usage_id = f"agent-hook:{hook.name or 'default'}".`

One thing I'd not do is reuse the full `display_command` string as the `usage_id`. The two serve different purposes: `display_command` is a fuzzy human label with a fallback chain (name → prompt[:20] → static), whereas `usage_id` is a metrics aggregation key that wants to be deterministic and stable.

(cc. @enyst do you agree?)

VascoSch92 (commented):
Still a couple of things but after that I think we are good to go.

enyst (commented):
SGTM! Thank you for this!

VascoSch92:
@openhands fix the pre-commit checks, the check duplicate example numbers and understand why the REST API breakage checks are failing

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -138,15 +151,161 @@ def cleanup_all(self) -> None:
...
+            return self._fall_open("No LLM configured for agent hook")
+
+        hook_llm = self.llm.model_copy(
+            update={
+                "usage_id": f"agent-hook:{hook.name or 'default'}",
+                "timeout": hook.timeout,
+            }
+        )
+        # Isolate Metrics so hook spend doesn't accrue to the parent's bucket.
+        hook_llm.reset_metrics()
```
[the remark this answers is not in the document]
VascoSch92: Fixed in 823b460b. Agent hook sub-conversations now merge their child conversation_stats back into the parent stats, preserving isolated agent-hook:* usage buckets while including hook spend in get_combined_metrics()/EXAMPLE_COST. Added a regression test covering repeated usage-id merge behavior.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -565,6 +565,9 @@ def _ensure_plugins_loaded(self) -> None:
                 working_dir=str(self.workspace.working_dir),
                 session_id=str(self._state.id),
                 original_callback=self._base_callback,
+                llm=self.agent.llm,
+                persistence_dir=self._state.persistence_dir,
```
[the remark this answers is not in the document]
VascoSch92: Fixed in 95895c8c. Hook sub-conversations are still persisted, but they now receive the conversation persistence base directory instead of the parent conversation-specific directory, matching the fork path handling and avoiding <base>/<parent-id>/<hook-id> nesting. Added a regression test that asserts create_hook_callback receives the base persistence path.

openhands-sdk/openhands/sdk/hooks/executor.py
```diff
@@ -138,15 +151,161 @@ def cleanup_all(self) -> None:
...
+        if data is None:
+            logger.warning(
+                f"Agent hook returned no parseable JSON object for event"
+                f" '{event_type}' — defaulting to allow: {repr(raw)[:200]}"
+            )
+            return self._fall_open(
+                "Agent hook returned no parseable JSON — defaulting to allow"
+            )
+
+        decision_str = str(data.get("decision", "allow")).lower()
```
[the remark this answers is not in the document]
VascoSch92: Fixed in f86fad4. `_parse_decision` now only returns a deliberate verdict for an explicit `allow`/`deny`. A missing or unknown `decision` routes through `_fall_open` (`success=False`, `error` set, `decision=ALLOW`), so it's detectable as a fall-open rather than a silent allow. Added a parametrized regression test covering missing/unknown/empty decision values.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -565,6 +565,9 @@ def _ensure_plugins_loaded(self) -> None:
...
+                visualizer=self._visualizer,
```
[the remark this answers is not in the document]
VascoSch92: Fixed in f86fad4. The executor no longer hands the parent's initialized visualizer instance to the hook sub-conversation (which `LocalConversation.initialize()` would rebind to the child state). It now mirrors the delegate pattern and requests a fresh sub-visualizer via `create_sub_visualizer("agent-hook:<name>")` — returning `None` for visualizers that don't support sub-agents. Added a test asserting the parent instance is never passed through.

examples/01_standalone_sdk/49_agent_hooks/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+A shell-based PreToolUse hook can only block what its blacklist literally
+matches. The agent rewrites `cat /etc/passwd` as `awk '{print}' /etc/passwd`
+and slips through. An agent hook reasons about the **semantic intent** of the
+command — "reading a sensitive system file" — and denies regardless of the
+exact tool name used.
+
+## Example
+
+- **49_agent_hooks.py** — Two agent hooks wired onto a single conversation:
```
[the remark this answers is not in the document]
VascoSch92: Fixed in f86fad4. The example was renamed to `51_agent_hooks/main.py`, its directory added to `_TARGET_DIRECTORIES`, and a discovery assertion added in `tests/examples/test_examples.py::test_directory_example_is_discovered`, so the example workflow now exercises it.

examples/01_standalone_sdk/51_agent_hooks/51_agent_hooks.py
```diff
@@ -0,0 +1,166 @@
...
+                        type=HookType.AGENT,
+                        name="security-reviewer",
+                        system_prompt=SECURITY_REVIEWER_PROMPT,
+                        timeout=60,
+                        max_iterations=3,
+                    )
+                ],
+            )
+        ],
+        stop=[
```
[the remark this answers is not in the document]
VascoSch92: Fixed in f86fad4. The two demos now run in separate conversations, each with only the hook it needs (`security_hook_config` for Demo 1, `quality_hook_config` for Demo 2). The Stop quality gate is no longer active during the PreToolUse demo. `EXAMPLE_COST` sums both conversations.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -559,12 +559,21 @@ def _ensure_plugins_loaded(self) -> None:
...
+                if self._state.persistence_dir is not None
+                else None
+            )
 
             self._hook_processor, self._on_event = create_hook_callback(
                 hook_config=final_hook_config,
                 working_dir=str(self.workspace.working_dir),
                 session_id=str(self._state.id),
                 original_callback=self._base_callback,
+                llm=self.agent.llm,
```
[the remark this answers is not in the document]
VascoSch92: Fixed in ead24bf. The executor no longer captures a single LLM at init. A `llm_getter` is threaded through `create_hook_callback` → `HookManager` → `HookExecutor`, and `local_conversation.py` passes `llm_getter=lambda: self.agent.llm`. Since `switch_llm()`/`switch_profile()` rebind `self.agent`, agent hooks now resolve the active LLM at execution time. Added a regression test (`test_llm_getter_is_resolved_live`) that swaps the LLM after init and asserts the hook copies from the new one.

.github/scripts/check_agent_server_rest_api_breakage.py
```diff
@@ -606,7 +619,9 @@ def _split_breaking_changes(
             removed_schema_properties.append(change)
             continue
 
-        if change_id in _ADDITIVE_RESPONSE_ONEOF_IDS:
+        if change_id in _ADDITIVE_RESPONSE_ONEOF_IDS or (
+            change_id in _ADDITIVE_RESPONSE_ENUM_IDS
```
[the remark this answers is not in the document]
VascoSch92: Fixed in ead24bf. The enum downgrade is now scoped to the hook discriminator instead of every response enum. `_is_additive_discriminator_enum_value` only downgrades `response-property-enum-value-added` when the change text targets the hook config union (`HookConfig … hooks/items/type`); any other response enum value (a new status/mode/etc.) stays a breaking change. Added a test asserting an unrelated `status` enum addition remains in the breaking bucket.

examples/01_standalone_sdk/51_agent_hooks/main.py
```diff
@@ -0,0 +1,182 @@
...
+    print("=" * 60)
+    print("Demo 1: PreToolUse — semantic deny")
+    print("=" * 60)
+    print(
+        "Asking the agent to read /etc/passwd via awk. The literal command\n"
+        "wouldn't match a syntactic blacklist (no `cat`, no `/etc/shadow`\n"
+        "keyword), but the security-reviewer agent should recognise the\n"
+        "intent and deny.\n"
+    )
+    security_conversation = Conversation(
```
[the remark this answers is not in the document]
VascoSch92: Fixed in ead24bf. Each demo conversation now runs with `max_iteration_per_run=MAX_ITERATIONS` (10) and the example raises if the conversation ends in `ERROR`/`STUCK`, so a denial loop fails fast instead of running to the default 500-iteration limit / CI timeout.

examples/01_standalone_sdk/51_agent_hooks/main.py
```diff
@@ -0,0 +1,182 @@
...
+    security_conversation.run()
+    total_cost += (
+        security_conversation.conversation_stats.get_combined_metrics().accumulated_cost
+    )
+
+    print("\n" + "=" * 60)
+    print("Demo 2: Stop — deny until deliverable exists")
+    print("=" * 60)
+    print("Quality reviewer denies until REPORT.md exists with a bullet point.\n")
+    quality_conversation = Conversation(
```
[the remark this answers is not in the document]
VascoSch92: Fixed in ead24bf. Each demo now runs with its own `llm.model_copy()` + `reset_metrics()`, so a conversation's `conversation_stats` only reflects its own spend and `EXAMPLE_COST` no longer double-counts Demo 1's main-agent cost.

VascoSch92 (approved):

PR #3063, opened 2026-05-05
fix(agent-server): skip restart-recovery error when observation already exists

### LINKED ISSUE #3045: Crash recovery can create duplicate tool results and crash condensation with KeyError
This issue was created by an AI agent (OpenHands) on behalf of the user.

### Bug Description

We observed a production conversation in `OpenHands/software-agent-sdk` enter an unrecoverable error state because the same tool call ended up with **two observation-like events** in the event stream:

1. `ActionEvent(tool_call_id=X)`
2. `AgentErrorEvent(tool_call_id=X)` emitted by crash/restart recovery
3. later `ObservationEvent(tool_call_id=X)` for the same action/tool call

That duplicate observation-like pairing is tolerated long enough to remain in the conversation history, but later crashes condensation in the context-view tool-call matching logic.

The eventual runtime stack trace is:

```text
File "openhands/sdk/agent/utils.py", line 400, in prepare_llm_messages
File "openhands/sdk/context/condenser/base.py", line 151, in condense
File "openhands/sdk/context/condenser/llm_summarizing_condenser.py", line 252, in _get_forgotten_events
File "openhands/sdk/context/view/view.py", line 49, in manipulation_indices
File "openhands/sdk/context/view/properties/tool_call_matching.py", line 93, in manipulation_indices
KeyError: '<tool_call_id>'
```

### Root Cause

`ToolCallMatchingProperty.enforce()` only checks **set membership** of `tool_call_id`s, so this invalid history survives enforcement:

- 1 action for tool call `X`
- 2 observation-like events for tool call `X`

Then `ToolCallMatchingProperty.manipulation_indices()` assumes strict 1:1 pairing and does:

```python
pending_tool_call_ids.add(event.tool_call_id)
pending_tool_call_ids.remove(event.tool_call_id)
pending_tool_call_ids.remove(event.tool_call_id)  # KeyError on duplicate result
```

So the final crash happens during condensation, but the underlying bug is that crash recovery and late tool completion can both record observation-like results for the same tool call.

### Why this can look like a Stop-hook failure

In the concrete production case, a Stop hook denied stopping shortly before the conversation died. That made the conversation continue long enough to cross the condenser threshold, so the failure surfaced immediately afterward.

However, the Stop hook is **not** the root cause. The broken history already existed earlier. The hook just nudged the conversation into the code path that exercised condensation.

### Related Observations

We also saw a similar malformed-history symptom where Anthropic rejected the conversation with:

```text
Found multiple `tool_result` blocks with id: ...
```

So this is broader than a single KeyError site: the same underlying duplicate-result history can either:

- crash the SDK locally during condensation (`KeyError`), or
- reach the provider and be rejected as malformed tool history.

### Expected Behavior

The conversation history should never contain multiple observation-like events for a single tool call in a way that can be serialized back to the LLM.

At minimum, OpenHands should do one of the following:

1. Prevent the duplicate from being recorded in the first place
   - if crash recovery emits `AgentErrorEvent(tool_call_id=X)`, suppress/coalesce any later `ObservationEvent(tool_call_id=X)`
   - or vice versa
2. Make tool-call matching enforcement cardinality-aware so duplicate observation-like events are removed deterministically before computing manipulation indices
3. Ensure malformed histories do not hard-crash `manipulation_indices()` with a raw `KeyError`

### Actual Behavior

The conversation enters an unrecoverable broken state. Subsequent user messages keep failing because the context/condenser path crashes on the duplicate tool-call result history.

### Relevant Code Paths

- `openhands-agent-server/openhands/agent_server/event_service.py`
  - restart recovery emits `AgentErrorEvent` for unmatched actions loaded from a stale `RUNNING` conversation
- `openhands-sdk/openhands/sdk/context/view/properties/tool_call_matching.py`
  - `enforce()` is not cardinality-aware for duplicate observation-like events
  - `manipulation_indices()` assumes strict 1 action -> 1 observation pairing and raises `KeyError`
- `openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py`
  - condensation later triggers the crash via `view.manipulation_indices`

### Additional Context

This seems related in symptoms to #2550, but this report includes a more specific root cause: duplicate observation-like events for the same `tool_call_id`, caused by the interaction between crash recovery and late tool completion.

VascoSch92:

## Why

Optional complementary patch for issue #3045

## Summary

- In event_service.py restart recovery, skip the synthesized AgentErrorEvent when the loaded state already carries any ObservationBaseEvent for the unmatched action's tool_call_id. Closes the gap where an observation matches by    
  tool_call_id but not action_id and would otherwise produce a duplicate observation-like event.
- Complements the cardinality-aware enforce() and soft manipulation_indices() already in tool_call_matching.py: stops new bad histories at the source while those guards self-heal histories already on disk.

## Issue Number
#3045

## How to Test
- added a regression test for the changes

## Type

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes
This PR closes #3045

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:7d8b073-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-7d8b073-python \
  ghcr.io/openhands/agent-server:7d8b073-python
```

```
ghcr.io/openhands/agent-server:7d8b073-golang-amd64
ghcr.io/openhands/agent-server:7d8b073-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:7d8b073-golang-arm64
ghcr.io/openhands/agent-server:7d8b073-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:7d8b073-java-amd64
ghcr.io/openhands/agent-server:7d8b073-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:7d8b073-java-arm64
ghcr.io/openhands/agent-server:7d8b073-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:7d8b073-python-amd64
ghcr.io/openhands/agent-server:7d8b073-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:7d8b073-python-arm64
ghcr.io/openhands/agent-server:7d8b073-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:7d8b073-golang
ghcr.io/openhands/agent-server:7d8b073-java
ghcr.io/openhands/agent-server:7d8b073-python
```

- Each variant tag (e.g., `7d8b073-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `7d8b073-python-amd64`) are also available if needed


tests/agent_server/test_event_service.py
```diff
@@ -1401,6 +1407,75 @@ async def test_start_does_nothing_when_status_not_running(
             ]
             assert len(error_event_calls) == 0
 
+    @pytest.mark.asyncio
+    async def test_start_skips_error_event_when_observation_already_exists(
+        self, event_service, tmp_path
+    ):
+        """Don't synthesize AgentErrorEvent if the loaded state already carries an
+        ObservationBaseEvent for the unmatched action's tool_call_id (issue #3045).
```
csmith49: Minor nit: OpenHands loves to put issue numbers in tests built from an issue. Always felt a little strange to me. Especially here, where the rest of the doc-string gives a pretty full context.

csmith49 (approved):
🔥

tests/agent_server/test_event_service.py
VascoSch92:
```suggestion
        """Don't synthesize AgentErrorEvent if the loaded state already carries an
        ObservationBaseEvent for the unmatched action's tool_call_id.
```

PR #3113, opened 2026-05-07
Fix orphaned Laminar traces from cross-async-context conversations

neubig:
## Problem

While auditing trace ingestion at `https://laminar.app.all-hands.dev/`, ~63% of conversations show up as orphaned/split traces.

In a 24h window (520 traces, last 14d):

| top_span_name | count | session_id set? |
|---|---:|---:|
| `conversation` (root) ✅ | 193 | 100% |
| `conversation.send_message` | 163 | **0%** |
| `conversation.run` | 151 | **0%** |
| `tools/call create_pr` | 6 | 0% |
| `litellm.completion` | 5 | 0% |

The orphan spans have `parent_span_id = 00000000-0000-0000-0000-000000000000` and contain real LLM/tool data — i.e. successful agent runs are being torn into 2-3 unrelated traces in the dashboard, with no `session_id` to stitch them back together.

## Root cause

`BaseConversation._start_observability_span` calls `Laminar.start_active_span("conversation")`, which attaches the OTel context using `contextvars.attach()`. From the upstream Laminar docstring (verbatim):

> Note that ending the started span in a different async context yields unexpected results. When propagating spans across different async or threading contexts, it is recommended to either:
> - Make sure to start and end the span in the same async context or thread, or
> - **Use `Laminar.start_span` + `Laminar.use_span` where possible.**

`contextvars` only propagate to tasks created *after* the attach, *on the same thread*. Pre-existing async tasks (FastAPI handlers, websocket loops, anything `RemoteConversation` does) don't see the parent. When such a task calls a method decorated with `@observe(...)` — `send_message`, `run`, `generate_title`, `agent.step`, `MCPToolExecutor.call_tool` — it starts a fresh root trace and the conversation is silently torn apart.

A second related bug: `SpanManager` is a process-wide LIFO stack. Two concurrent `Conversation` instances corrupt each other — closing one pops whichever span happens to be on top.

## Fix

Switch to the upstream-recommended `start_span` + `use_span` pattern:

1. **`RootSpan` owned by the conversation.** `BaseConversation` stores its root span on `self._observability_root_span` instead of pushing onto a global stack. `start_root_span` uses `Laminar.start_span` (no implicit context attach). `session_id` is applied via a brief `use_span` window so it lands on trace metadata.
2. **`@observe` re-attaches the root span at every entry point.** The decorator wrapper inspects the first positional arg (`self`) for an `_observability_root_span`, and if present wraps the call in `Laminar.use_span(...)`. Nested `@observe`'d calls and lmnr's auto-instrumentation inherit the right parent automatically.
3. **Backwards-compat shims kept.** `SpanManager`, `start_active_span`, `end_active_span` are now thin wrappers around the new primitive, so anything outside the SDK that imported them still works (just without the global-collision footgun).

### Effect on the trace shape

Before:

```
trace A: conversation                (root, has session_id, tokens=0)
trace B: conversation.send_message   (orphan, no session_id)
trace C: conversation.run            (orphan, no session_id)
         └─ agent.step
            ├─ litellm.completion
            └─ TerminalAction
```

After:

```
trace A: conversation                (root, has session_id, accumulates cost/tokens)
         ├─ conversation.send_message
         │  └─ litellm.completion
         └─ conversation.run
            └─ agent.step
               ├─ litellm.completion
               └─ TerminalAction
```

## Tests

Three new regression tests in `tests/sdk/observability/test_laminar.py`:

- **`test_observe_calls_use_span_with_owner_root_span_on_sync`** — sync `@observe`'d methods re-attach the owner's root span via `Laminar.use_span`.
- **`test_observe_calls_use_span_with_owner_root_span_on_async`** — async `@observe`'d methods re-attach even when invoked from a freshly-created `asyncio.Task` with an empty `contextvars.Context` (the cross-context bug repro).
- **`test_two_concurrent_conversations_do_not_collide`** — two `Conversation` instances get independent root spans and ending one does not affect the other.

All 28 observability tests pass locally; full SDK suite is running and I'll update this PR with the result before un-drafting.

## Risk assessment

- **Behavior change visible in Laminar**: orphan trace counts drop, child spans now appear under the right parent. No callers should rely on the old orphan behavior.
- **API surface**: only adds public `RootSpan` / `start_root_span` / `end_root_span`. `start_active_span` / `end_active_span` / `SpanManager` are kept as functionally-equivalent shims.
- **Hot path**: the `observe` wrapper now does one `getattr` and (when observability is enabled and the owner has a root span) a `Laminar.use_span` per call. That's roughly the same work `start_active_span` was doing internally; it's just moved to the call site so it works across async/thread boundaries. Negligible perf impact.
- **Async correctness**: the new pattern is the one Laminar themselves recommend in their own docstrings. Any failure inside `use_span` is logged at debug and the wrapped function is still called, so observability errors can never break the SDK.

## Why this is structural, not transient

The 63% orphan rate is determined by *how callers use the SDK*, not by data freshness. Every caller that constructs a `Conversation` in one task and invokes `send_message`/`run` from a different task hits this — `RemoteConversation` is one such caller by design. Waiting longer in production won't reduce the orphan rate.

---

Marking as draft for review — happy to iterate on naming or scope.

_This PR was opened by an AI agent (OpenHands) on behalf of @neubig._

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/b731c29b-1679-4d85-9dc5-10f676c00fad)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:1253592-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-1253592-python \
  ghcr.io/openhands/agent-server:1253592-python
```

```
ghcr.io/openhands/agent-server:1253592-golang-amd64
ghcr.io/openhands/agent-server:1253592-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:1253592-golang-arm64
ghcr.io/openhands/agent-server:1253592-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:1253592-java-amd64
ghcr.io/openhands/agent-server:1253592-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:1253592-java-arm64
ghcr.io/openhands/agent-server:1253592-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:1253592-python-amd64
ghcr.io/openhands/agent-server:1253592-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:1253592-python-arm64
ghcr.io/openhands/agent-server:1253592-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:1253592-golang
ghcr.io/openhands/agent-server:1253592-java
ghcr.io/openhands/agent-server:1253592-python
```

- Each variant tag (e.g., `1253592-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `1253592-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/base.py
+++ b/openhands-sdk/openhands/sdk/conversation/base.py
@@ -14,9 +14,10 @@ from openhands.sdk.conversation.types import (
 from openhands.sdk.llm.llm import LLM
 from openhands.sdk.llm.message import Message
 from openhands.sdk.observability.laminar import (
-    end_active_span,
+    RootSpan,
+    end_root_span,
     should_enable_observability,
-    start_active_span,
+    start_root_span,
 )
 from openhands.sdk.security.analyzer import SecurityAnalyzerBase
 from openhands.sdk.security.confirmation_policy import (
@@ -118,21 +119,36 @@ class BaseConversation(ABC):
     def __init__(self) -> None:
         """Initialize the base conversation with span tracking."""
         self._span_ended = False
+        # Owned root span. The ``observe`` decorator looks up this attribute
+        # (by name ``_observability_root_span``) on ``self`` at every entry
+        # point and re-attaches it via ``Laminar.use_span`` so that nested
+        # spans correctly join the conversation trace even when the method
+        # is called from a different asyncio task or thread than the one
+        # that constructed the conversation.
+        self._observability_root_span: RootSpan | None = None
 
     def _start_observability_span(self, session_id: str) -> None:
-        """Start an observability span if observability is enabled.
+        """Start a per-conversation observability root span.
 
         Args:
-            session_id: The session ID to associate with the span
+            session_id: The session ID to associate with the trace
         """
-        if should_enable_observability():
-            start_active_span("conversation", session_id=session_id)
+        if not should_enable_observability():
+            return
+        if self._observability_root_span is not None:
+            # Idempotent: never start two roots for one conversation.
+            return
+        self._observability_root_span = start_root_span(
+            "conversation", session_id=session_id
+        )
 
     def _end_observability_span(self) -> None:
         """End the observability span if it hasn't been ended already."""
-        if not self._span_ended and should_enable_observability():
-            end_active_span()
-            self._span_ended = True
+        if self._span_ended:
+            return
+        end_root_span(self._observability_root_span)
+        self._observability_root_span = None
+        self._span_ended = True
 
     @property
     @abstractmethod
--- a/openhands-sdk/openhands/sdk/observability/laminar.py
+++ b/openhands-sdk/openhands/sdk/observability/laminar.py
@@ -1,7 +1,8 @@
+import contextlib
 import functools
 import inspect
 import sys
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from typing import TYPE_CHECKING, Any, Final, Literal
 
 from openhands.sdk.logger import get_logger
@@ -168,11 +169,13 @@ def observe[**P, R](
             async def async_wrapper(*args: P.args, **fkwargs: P.kwargs) -> Any:
                 nonlocal wrapped
                 if wrapped is not None:
-                    return await wrapped(*args, **fkwargs)
+                    with _maybe_use_root_span(args):
+                        return await wrapped(*args, **fkwargs)
                 if not should_enable_observability():
                     return await func(*args, **fkwargs)
                 wrapped = _build_wrapped(func)
-                return await wrapped(*args, **fkwargs)
+                with _maybe_use_root_span(args):
+                    return await wrapped(*args, **fkwargs)
 
             return async_wrapper  # type: ignore[return-value]
 
@@ -180,11 +183,13 @@ def observe[**P, R](
         def sync_wrapper(*args: P.args, **fkwargs: P.kwargs) -> R:
             nonlocal wrapped
             if wrapped is not None:
-                return wrapped(*args, **fkwargs)
+                with _maybe_use_root_span(args):
+                    return wrapped(*args, **fkwargs)
             if not should_enable_observability():
                 return func(*args, **fkwargs)
             wrapped = _build_wrapped(func)
-            return wrapped(*args, **fkwargs)
+            with _maybe_use_root_span(args):
+                return wrapped(*args, **fkwargs)
 
         return sync_wrapper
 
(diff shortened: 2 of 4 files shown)


neubig:
✅ Full SDK test suite: **3573 passed, 0 failed** (`uv run pytest tests/sdk --timeout=20`).

Two tests deselected are pre-existing environmental failures, both confirmed independent of this PR by reproducing them on `main`:

- `tests/sdk/conversation/local/test_conversation_core.py::test_conversation_large_event_handling` — hangs in this sandbox (no LLM/network); also hangs on `main`.
- `tests/sdk/agent/test_acp_agent.py::TestACPSecretsEnvInjection::test_static_secret_injected_into_subprocess_env` — the sandbox's secret-injection layer rewrites `GITHUB_TOKEN` in the subprocess env, so the assertion `env.get("GITHUB_TOKEN") == "ghp_test123"` fails with the real token. Pre-existing on `main`.

Includes a follow-up commit (a76818dc) that updates `test_base_span_management.py` to patch the new `start_root_span`/`end_root_span` symbols and asserts the new idempotency guarantee.

_— OpenHands_

neubig:
### Deprecation schedule + downstream-consumer audit

To answer the question of how long to keep the backwards-compat shims (`SpanManager`, `start_active_span`, `end_active_span`), I ran a downstream-consumer audit and added a formal schedule (commit e37ace5e).

**Audit results (2026-05-07):**

| Search target | Hits | Notes |
|---|---:|---|
| `OpenHands/OpenHands` (grep on full clone) | **0** | nothing imports these symbols |
| `OpenHands/agent-canvas` | **0** | TypeScript only — no Python at all |
| `OpenHands/codescout` | 1 file | imports only `maybe_init_laminar` (untouched) |
| Org-wide `gh search code start_active_span org:OpenHands` | **0** | |
| Global `from openhands.sdk.observability.laminar import …` | only the SDK itself, forks of it, and a benchmark that vendors the SDK source | |

So in practice **no first- or second-party caller** uses the deprecated symbols. The shims now exist purely to give any unaudited private consumer a heads-up before the SDK simply removes them.

**Schedule (matches existing SDK precedent):**

```
deprecated_in = "1.22.0"   # this PR ships in 1.22.0
removed_in    = "1.27.0"   # 5-minor-version grace window
```

This matches the SDK's previous `VerificationSettings.confirmation_mode` deprecation (deprecated 1.17.0 / removed 1.22.0). Each shim now calls `warn_deprecated(...)` so:

- Calls on 1.22.0–1.26.x emit `DeprecationWarning`.
- Calls on 1.27.0+ emit `UnsupportedWarning` (and the maintainer can delete the shims at that point).
- Calls during 1.21.x development are silent — the warning helper checks `current_version >= deprecated_in` before firing.

**Re transitioning OpenHands/OpenHands or agent-canvas:** nothing to transition. Both repos are clean. The only transition needed was inside this SDK itself, which the rest of this PR already handles (`BaseConversation` now uses `start_root_span` / `end_root_span` directly).

Added `test_deprecated_shims_emit_warnings` (patches `_current_version` to `"1.22.0"`) so the warning contract is locked in for CI.

_— OpenHands_

xingyaoww (approved):
Let's try this :)

Issue #3046, opened 2026-05-03
[Feature]: Enforce AgentSkills spec properties at runtime (allowed-tools, paths, disable-model-invocation)


neubig:
## Summary

`invoke_skill` is already a first-class tool call (with `InvokeSkillAction`/`InvokeSkillObservation`), and the `Skill` model parses `allowed-tools` from frontmatter. However, none of the AgentSkills spec runtime-enforcement properties are actually enforced during or after skill invocation.

This was surfaced by an external benchmark team (SkillsBench) who observed that skill invocations lack the enforcement semantics described in the [AgentSkills spec](https://agentskills.io):

1. **`allowed-tools`** — Parsed into `Skill.allowed_tools` but never checked. After a skill is invoked, the agent can use any tool regardless of what the skill declares. The spec expects that when a skill is active, only the listed tools are available to the model.

2. **`disable-model-invocation`** — Not implemented. The spec allows a skill to declare that it should only be activated by trigger matching, never by the model calling `invoke_skill` directly.

3. **`paths`** — Not implemented. The spec allows a skill to restrict file-system access to specific directories during skill execution.

## Current State

- `invoke_skill` is the canonical invocation path and `to_prompt()` intentionally omits `<location>` so the agent cannot bypass it by reading SKILL.md directly ✅
- `Skill.allowed_tools` is parsed from frontmatter ✅
- `allowed-tools` is **not enforced** at invocation time or in the agent loop ❌
- `disable-model-invocation` is **not parsed or enforced** ❌
- `paths` is **not parsed or enforced** ❌

## Why This Matters

- **Benchmark reproducibility**: Without enforcement, skill invocations cannot be cleanly audited — there is no guarantee the agent respected the skill's declared constraints.
- **Spec compliance**: The AgentSkills spec defines these properties as runtime constraints, not just metadata. Other implementations (e.g., Claude Code) enforce `allowed-tools` and `context: fork` at the tool-availability level.
- **Security/sandboxing**: `allowed-tools` and `paths` are the skill author's way of limiting blast radius. Without enforcement, a skill that declares `allowed-tools: terminal` can still trigger browser actions, file edits, etc.

## Related Issues

- #2042 — Subagent delegation for `context: fork` (addresses execution isolation, not tool restriction)
- #2053 — Skills epic (covers model routing and agent type, but not spec enforcement)
- #2981 — Skill loading semantics gap (about injection behavior, not runtime enforcement)

## Suggested Approach

[...]

PR #3114, opened 2026-05-07
fix(sdk): enforce disabled skill model invocation

Zheng-Lu:

---

## Why

`disable-model-invocation` skills should only activate through trigger matching. Before this PR, the SDK ignored that frontmatter field: disabled skills were still listed in `<available_skills>` and could still be invoked directly through `invoke_skill`.

## Summary

- Parse `disable-model-invocation` / `disable_model_invocation` into `Skill.disable_model_invocation`.
- Hide disabled skills from `<available_skills>` while preserving trigger-based activation.
- Reject direct `invoke_skill` calls for disabled skills and avoid auto-attaching `invoke_skill` when only disabled AgentSkills are present.

## Issue Number

The sub-issue "`disable-model-invocation` is not parsed or enforced" mentioned in #3046

## How to Test

Targeted tests:

```bash
uv run pytest tests/sdk/skills/test_agentskills_fields.py tests/sdk/context/test_agent_context.py::TestAgentContext::test_disable_model_invocation_hides_skill_but_preserves_triggers tests/sdk/tool/test_invoke_skill.py -q
```

## Video/Screenshots

I created a temporary AgentSkills-format `SKILL.md`:

```yaml
name: trigger-only
description: Hidden skill
disable-model-invocation: true
triggers:
  - hidden-keyword
 ```
 
 to check:
 - whether `disable-model-invocation` is parsed into `Skill.disable_model_invocation`
- whether the skill appears in `<available_skills>` and is therefore visible for direct `invoke_skill`
- whether trigger activation still works while direct `invoke_skill(name="trigger-only")` is rejected

Before this PR:
image

This shows the bug:

- `disable_attr=MISSING`: the field was not parsed by the Skill model.
- `prompt_lists_skill=True`: the skill was still advertised in <available_skills>.
- `invoke_error=False`: direct invoke_skill succeeded.
- `invoked_skills=['trigger-only']`: the SDK recorded the disabled skill as successfully invoked.

After this PR:
image

This shows the fix:

- `disable_attr=True`: the field is now parsed.
- `prompt_lists_skill=False`: the skill is no longer advertised for direct model invocation.
- `triggered_skill_names=['trigger-only']` and `trigger_content_present=True`: trigger-based activation still works.
- `invoke_error=True`: direct invoke_skill is rejected.
- `invoked_skills=[]`: the rejected direct invocation is not recorded as a successful skill invocation.

Pytest result: 35 passed, 1 warning in 0.21s
image

Pre-commit result:
image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes
uv run emitted an existing warning about parsing exclude-newer = "7 days" in pyproject.toml, but the targeted tests and pre-commit checks completed successfully.

### CODE REVIEW

File: openhands-sdk/openhands/sdk/context/agent_context.py [hunk 1]
```diff
@@ -207,19 +207,24 @@ def _partition_skills(self) -> tuple[list[Skill], list[Skill]]:
...
-        - Legacy with triggers: listed in available_skills, injected on trigger.
+        - Legacy with triggers: listed in available_skills unless direct model
+          invocation is disabled, injected on trigger.
 
         Returns:
             ``(repo_skills, available_skills)`` tuple.
         """
         repo_skills: list[Skill] = []
         available_skills: list[Skill] = []
         for s in self.skills:
```
VascoSch92, line 221: Could you flatten this one? For example something like:

```
if s.is_agentskills_format or s.trigger is not None:                                                                                                                                                                                                                         
      if not s.disable_model_invocation:                                                                                                                                                                                                                                       
          available_skills.append(s)                                                                                                                                                                                                                                           
  else:                                                                                                                                                                                                                                                                        
      repo_skills.append(s)        
```

openhands-sdk/openhands/sdk/skills/skill.py
```diff
@@ -208,6 +208,13 @@ class Skill(BaseModel):
             "AgentSkills standard field (parsed from space-delimited string)."
         ),
     )
+    disable_model_invocation: bool = Field(
+        default=False,
+        description=(
+            "Whether this skill can only be activated by trigger matching and "
+            "cannot be directly invoked via invoke_skill."
```
VascoSch92:
```suggestion
            "cannot be directly disclose by the model."
```
I think user don't know what is the `invoke_skill` tool.

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
VascoSch92, line 1: I think you don't need the changes in invoke_skill tool. 

If the skill is not displayed in the system prompt of the model, then the model will not the use the tool on it.

VascoSch92 (changes requested):
Hey @Zheng-Lu 

thank you very much for this PR.

There are just a couple of things that need to be change/complete.

The most important one is that the fix is incomplete for skills delivered via the agent-server. Two endpoints don't carry `disable_model_invocation`:                                                                                                                     
  - Server response model (`openhands-agent-server/openhands/agent_server/skills_router.py:85-94`)
  ```python
  class SkillInfo(BaseModel):                                                                                                                                                                                                                                                  
      name: str                                                                                                                                                                                                                                                                
      type: Literal["repo", "knowledge", "agentskills"]                      
      content: str                                                                                                                                                                                                                                                             
      triggers: list[str] = Field(default_factory=list)                      
      source: str | None = None                                                                                                                                                                                                                                                
      description: str | None = None
      is_agentskills_format: bool = False                                                                                                                                                                                                                                      
      # ← no disable_model_invocation
```
  - And the constructor at line 160-169 doesn't forward it either.                                                                                                                                                                                                             
  - Client deserializer (`openhands-sdk/openhands/sdk/workspace/remote/base.py:761-768`)
  ```python
  return Skill(                                                                                                                                                                                                                                                                
      name=skill_data.get("name", "unknown"),                                                                                                                                                                                                                                  
      content=skill_data.get("content", ""),                                                                                                                                                                                                                                   
      trigger=trigger,                                                       
      source=skill_data.get("source"),                                                                                                                                                                                                                                         
      description=skill_data.get("description"),                                                                                                                                                                                                                               
      is_agentskills_format=skill_data.get("is_agentskills_format", False),
      # ← no disable_model_invocation                                                                                                                                                                                                                                          
  )
```


   
  So a SKILL.md with disable-model-invocation: true loaded through RemoteWorkspace.load_skills_from_agent_server(...) (the standard remote path) will arrive on the client with disable_model_invocation=False and be fully invocable. The bug the PR fixes is reproduced over 
  the wire.                                                                  
                                                                                                                                                                                                                                                                               
  Add the field to SkillInfo, populate it in the response converter, and read it in _skill_from_dict. Worth a wire-level test too.

openhands-sdk/openhands/sdk/context/agent_context.py
Zheng-Lu: Done

openhands-sdk/openhands/sdk/skills/skill.py
Zheng-Lu: Done

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
Zheng-Lu: I kept the invoke_skill check intentionally because hiding the skill from <available_skills> only means the model is not told about it. The model can still guess a skill name, reuse a name from previous context, or a tool call can come from replay/resume behavior. If the executor does not check this field, a disabled skill can still run once invoke_skill is called.

So this PR uses both protections:
1. Do not advertise the skill in <available_skills>.
2. Reject direct invocation at runtime even if it is attempted.

I think this makes disable-model-invocation a real runtime rule, not just prompt hiding.

Zheng-Lu:
> Hey @Zheng-Lu
> 
> thank you very much for this PR.
> 
> There are just a couple of things that need to be change/complete.
> 
> The most important one is that the fix is incomplete for skills delivered via the agent-server. Two endpoints don't carry `disable_model_invocation`:
> 
> * Server response model (`openhands-agent-server/openhands/agent_server/skills_router.py:85-94`)
> 
> ```python
> class SkillInfo(BaseModel):                                                                                                                                                                                                                                                  
>     name: str                                                                                                                                                                                                                                                                
>     type: Literal["repo", "knowledge", "agentskills"]                      
>     content: str                                                                                                                                                                                                                                                             
>     triggers: list[str] = Field(default_factory=list)                      
>     source: str | None = None                                                                                                                                                                                                                                                
>     description: str | None = None
>     is_agentskills_format: bool = False                                                                                                                                                                                                                                      
>     # ← no disable_model_invocation
> ```
> 
> * And the constructor at line 160-169 doesn't forward it either.
> * Client deserializer (`openhands-sdk/openhands/sdk/workspace/remote/base.py:761-768`)
> 
> ```python
> return Skill(                                                                                                                                                                                                                                                                
>     name=skill_data.get("name", "unknown"),                                                                                                                                                                                                                                  
>     content=skill_data.get("content", ""),                                                                                                                                                                                                                                   
>     trigger=trigger,                                                       
>     source=skill_data.get("source"),                                                                                                                                                                                                                                         
>     description=skill_data.get("description"),                                                                                                                                                                                                                               
>     is_agentskills_format=skill_data.get("is_agentskills_format", False),
>     # ← no disable_model_invocation                                                                                                                                                                                                                                          
> )
> ```
> 
> So a SKILL.md with disable-model-invocation: true loaded through RemoteWorkspace.load_skills_from_agent_server(...) (the standard remote path) will arrive on the client with disable_model_invocation=False and be fully invocable. The bug the PR fixes is reproduced over the wire.
> 
> Add the field to SkillInfo, populate it in the response converter, and read it in _skill_from_dict. Worth a wire-level test too.

Changed as requested

openhands-sdk/openhands/sdk/tool/builtins/invoke_skill.py
VascoSch92: - _Model guessing a skill name:_ This is highly improbable. Why would the model need to guess a skill name?

- _Reusing a name from previous context:_ If the model already has access to the skill name in the current context, it implies the skill was already available for use.

- _Replay/resume behavior:_ This could be a factor if a skill's behavior changes between sessions, but I believe this happens very rarely.

However, we can leave the defense-in-depth in the invoke skill.

VascoSch92 (approved):

PR #3129, opened 2026-05-08
feat(profiles): LLM profile management API

VascoSch92:
## Why             
     
  Users need a way to save, list, retrieve, rename, and delete named LLM configurations from the agent-server (e.g., to switch models mid-conversation or maintain per-task LLM presets). The SDK already had LLMProfileStore for on-disk
   persistence, but no HTTP surface — so the frontend and remote clients couldn't manage profiles. This PR adds that surface, hardens the store against name-validation drift and os.environ mutation, and ships full test coverage.     
   
## Summary                                                                                                                                                                                                                                
                  
  - New /api/profiles endpoints in openhands-agent-server: GET / (list), GET /{name} (detail with api_key nulled), POST /{name} (save with optional include_secrets), DELETE /{name} (idempotent), POST /{name}/rename. Returns 409 at a 
  50-profile soft cap, 503 on store-lock timeouts, 404/409 on rename source/target conflicts.
  - LLMProfileStore additions: PROFILE_NAME_PATTERN (^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$) as the single source of truth for name validation, atomic rename() (single-lock move), and list_summaries() that reads JSON directly so listing  
  profiles never instantiates LLM objects (avoiding silent AWS_*/OR_* env mutation from LLM._set_env_side_effects).                                                                                                                      
  - Router hardening: shared regex with the store, single _store_errors context manager translating TimeoutError/ValueError to HTTP, atomic rename surfaces 404 before 409, api_key is always nulled in detail responses and never echoed
   back in the body.                                                                                                                                                                                                                     
  - 95 new tests across tests/agent_server/test_profiles_router.py and tests/sdk/llm/test_llm_profile_store.py covering happy paths, the profile cap, store timeouts (503), corrupted-profile listing/get (400), redacted-key reporting,
  malformed bodies, special-char/leading-non-alnum names, full rename matrix, and an explicit list_summaries env-isolation guarantee.                                                                                                    
                  

## How to Test

  - make pre-commit (passes: ruff, pyright, dependency rules).                                                                                                                                                                           
  - python -m pytest tests/agent_server/test_profiles_router.py tests/sdk/llm/test_llm_profile_store.py -q → 95 passing.
  - Manual smoke against a running agent-server:                                                                                                                                                                                         
    - POST /api/profiles/work {"llm": {"model": "gpt-4o", "api_key": "sk-…"}, "include_secrets": true} → 201.                                                                                                                            
    - GET /api/profiles → entry with api_key_set: true, no secret material in the response.                                                                                                                                              
    - GET /api/profiles/work → config.api_key is null.                                                                                                                                                                                   
    - POST /api/profiles/work/rename {"new_name": "play"} → 200; calling it again with a missing source → 404 (not 409).                                                                                                                 
    - POST /api/profiles/.hidden → 422 (regex rejects leading dot).                                                                                                                                                                      
    - With a profile containing aws_access_key_id, hitting GET /api/profiles does not modify os.environ["AWS_ACCESS_KEY_ID"].                                                                                                            
    - Save 50 profiles, then POST /api/profiles/<new> → 409 "Profile limit reached".       



---

**Variants & Base Images**
|---|---|---|---|


```bash
docker pull ghcr.io/openhands/agent-server:af4cf76-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-af4cf76-python \
  ghcr.io/openhands/agent-server:af4cf76-python
```

```
ghcr.io/openhands/agent-server:af4cf76-golang-amd64
ghcr.io/openhands/agent-server:af4cf76-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:af4cf76-golang-arm64
ghcr.io/openhands/agent-server:af4cf76-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:af4cf76-java-amd64
ghcr.io/openhands/agent-server:af4cf76-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:af4cf76-java-arm64
ghcr.io/openhands/agent-server:af4cf76-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:af4cf76-python-amd64
ghcr.io/openhands/agent-server:af4cf76-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:af4cf76-python-arm64
ghcr.io/openhands/agent-server:af4cf76-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:af4cf76-golang
ghcr.io/openhands/agent-server:af4cf76-java
ghcr.io/openhands/agent-server:af4cf76-python
```

- Each variant tag (e.g., `af4cf76-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `af4cf76-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm_profile_store.py
+++ b/openhands-sdk/openhands/sdk/llm/llm_profile_store.py
@@ -31,6 +31,10 @@ PROFILE_NAME_REGEX: Final[re.Pattern[str]] = re.compile(PROFILE_NAME_PATTERN)
 logger = get_logger(__name__)
 
 
+class ProfileLimitExceeded(Exception):
+    """Raised when saving would exceed the configured profile limit."""
+
+
 class LLMProfileStore:
     """Standalone utility for persisting LLM configurations."""
 
@@ -93,22 +97,41 @@ class LLMProfileStore:
             )
         return self.base_dir / f"{clean_name}.json"
 
-    def save(self, name: str, llm: LLM, include_secrets: bool = False) -> None:
+    def save(
+        self,
+        name: str,
+        llm: LLM,
+        include_secrets: bool = False,
+        *,
+        max_profiles: int | None = None,
+    ) -> None:
         """Save a profile to the profile directory.
 
-        Note that if a profile name already exists, it will be overwritten.
+        Overwrites an existing profile of the same name. When ``max_profiles``
+        is set, raises ``ProfileLimitExceeded`` if creating a *new* profile
+        would exceed the limit. The check happens under the same lock as the
+        save, so it is race-free against other ``save`` calls in this process.
 
         Args:
             name: Name of the profile to save.
             llm: LLM instance to save
             include_secrets: Whether to include the profile secrets. Defaults to False.
+            max_profiles: Optional cap on the number of profiles.
 
         Raises:
+            ProfileLimitExceeded: If ``max_profiles`` would be exceeded.
             TimeoutError: If the lock cannot be acquired.
         """
         profile_path = self._get_profile_path(name)
 
         with self._acquire_lock():
+            if max_profiles is not None and not profile_path.exists():
+                count = sum(1 for _ in self.base_dir.glob("*.json"))
+                if count >= max_profiles:
+                    raise ProfileLimitExceeded(
+                        f"Profile limit reached ({max_profiles})."
+                    )
+
             if profile_path.exists():
                 logger.info(
                     f"[Profile Store] Profile `{name}` already exists. Overwriting."
@@ -208,17 +231,28 @@ class LLMProfileStore:
         """List profile metadata without instantiating LLM objects.
 
         Reads JSON directly to avoid ``LLM._set_env_side_effects`` mutating
-        ``os.environ``. Corrupted profiles are skipped with a warning.
+        ``os.environ``. Files with invalid names, corrupted JSON, or non-dict
+        top-level values are skipped with a warning.
         """
         summaries: list[dict[str, Any]] = []
         with self._acquire_lock():
             for path in sorted(self.base_dir.glob("*.json")):
                 name = path.stem
+                if not PROFILE_NAME_REGEX.match(name):
+                    logger.warning(
+                        f"[Profile Store] Skipping profile with invalid name {name!r}"
+                    )
+                    continue
                 try:
                     data = json.loads(path.read_text())
                 except (OSError, json.JSONDecodeError) as e:
                     logger.warning(
                         f"[Profile Store] Skipping corrupted profile {name!r}: {e}"
+                    )
+                    continue
+                if not isinstance(data, dict):
+                    logger.warning(
+                        f"[Profile Store] Skipping non-dict profile {name!r}"
                     )
                     continue
                 api_key = data.get("api_key")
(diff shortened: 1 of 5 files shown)


neubig (commented):
Basically looks pretty good, but the AI review comments should probably be reflected.

openhands-sdk/openhands/sdk/llm/llm_profile_store.py
```diff
@@ -66,44 +83,55 @@ def _get_profile_path(self, name: str) -> Path:
...
                 f"Invalid profile name: {name!r}. "
-                "Profile names must be simple filenames without path separators."
+                "Profile names must be 1-64 characters, start with a letter "
+                "or digit, and contain only letters, digits, '.', '_', or '-'."
             )
-
         return self.base_dir / f"{clean_name}.json"
 
-    def save(self, name: str, llm: "LLM", include_secrets: bool = False) -> None:
+    def save(
```
malhotra5, line 100: Do we use the cipher to encrypt any sensitive fields?

malhotra5: Also atm we do the following for settings

- save settings with cipher for sensitive fields
- Clients can retrieve settings with header 'encrypted' or 'plain-text'
- This is so that FE can retrieve with 'encrypted' header
- In start conversation request, is the encrypted flag is true, the values should be decrypted before proceeding via the cipher

Is that the same here?

VascoSch92: No — currently the profiles flow is simpler than the settings flow. Concretely, the differences today:

| | Settings | Profiles (this PR) |
|---|---|---|
| At-rest encryption | Cipher-encrypted via `serialize_secret(cipher=…)` | Plaintext when `include_secrets=True`, redacted marker `**********` otherwise |
| Read exposure | `X-Expose-Secrets: encrypted \| plaintext` (default redacted) | Always nulls `api_key` in the detail response — no header, no round-trip |
| Round-trip on write | FE can re-PATCH the encrypted blob; cipher decrypts on the way in | N/A (no encrypted form to round-trip) |
| Decryption at use site | `start_conversation` decrypts before use | N/A (the profile is loaded by `LLM.load_from_json` directly from disk; no decrypt step) |

To bring profiles in line with the settings pattern we'd need three things:

1. `LLMProfileStore.__init__(cipher=…)`; pass `{"expose_secrets": "encrypted", "cipher": cipher}` into `model_dump_json` on save and into `model_validate_json` on load. Pydantic's `serialize_secret`/`validate_secret` already implement both modes.
2. `profiles_router` reads `X-Expose-Secrets` (mirroring `settings_router._parse_expose_secrets_header`) and threads the cipher into `model_dump(context=…)` for the detail response. List stays redacted.
3. Wire a `LLMProfileStore(cipher=config.cipher)` from the router so the agent-server's `OH_SECRET_KEY` is the source of truth, same as settings.

I'd suggest doing this as a follow-up PR rather than expanding the scope here — it changes the on-disk format (existing plaintext profiles need a migration / opt-in) and adds a new header contract that's worth a focused review on its own. Happy to open the issue / PR right after this one merges if you'd like.

malhotra5: Sounds good!

VascoSch92: @openhands Open an issue for a following PR about what we discussed here.

VascoSch92: Follow-up issue created: https://github.com/OpenHands/software-agent-sdk/issues/3154

This tracks adding cipher encryption support for LLM profile secrets to achieve parity with the settings flow.

malhotra5 (approved):
LGTM!

PR #3183, opened 2026-05-09
fix(sdk): abort run loop on repeated Stop-hook denials without progress

VascoSch92:

## Why

https://openhands-ai.slack.com/archives/C06R25BT5B2/p1778340691700839

## Summary

- Adds max_consecutive_stop_denials (default 5) to LocalConversation. When a Stop hook denies repeatedly with no ActionEvent between denials, the run loop now aborts with a ConversationErrorEvent(code="StopHookLoopDetected") and
  ERROR status instead of looping until max_iteration_per_run (default 500) is exhausted.
- The streak counter resets whenever the agent emits an ActionEvent between two denials, so legitimate iterative work keeps running.

Surfaced from a trajectory where a misconfigured `.openhands/hooks/on_stop.sh` (running npm run lint against an env missing dev deps) returned decision: deny on every Stop event. The agent recognized the hook was broken and emitted
text-only "I'm done" turns, but each one re-triggered the Stop hook — 56 cycles before the user manually paused, and nothing tool-actionable for the agent to do. Without this guard the loop would have run to the 500-iteration cap.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:780f5d1-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-780f5d1-python \
  ghcr.io/openhands/agent-server:780f5d1-python
```

```
ghcr.io/openhands/agent-server:780f5d1-golang-amd64
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-golang-amd64
ghcr.io/openhands/agent-server:vasco-fix-hooks-golang-amd64
ghcr.io/openhands/agent-server:780f5d1-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:780f5d1-golang-arm64
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-golang-arm64
ghcr.io/openhands/agent-server:vasco-fix-hooks-golang-arm64
ghcr.io/openhands/agent-server:780f5d1-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:780f5d1-java-amd64
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-java-amd64
ghcr.io/openhands/agent-server:vasco-fix-hooks-java-amd64
ghcr.io/openhands/agent-server:780f5d1-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:780f5d1-java-arm64
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-java-arm64
ghcr.io/openhands/agent-server:vasco-fix-hooks-java-arm64
ghcr.io/openhands/agent-server:780f5d1-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:780f5d1-python-amd64
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-python-amd64
ghcr.io/openhands/agent-server:vasco-fix-hooks-python-amd64
ghcr.io/openhands/agent-server:780f5d1-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:780f5d1-python-arm64
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-python-arm64
ghcr.io/openhands/agent-server:vasco-fix-hooks-python-arm64
ghcr.io/openhands/agent-server:780f5d1-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:780f5d1-golang
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-golang
ghcr.io/openhands/agent-server:vasco-fix-hooks-golang
ghcr.io/openhands/agent-server:780f5d1-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:780f5d1-java
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-java
ghcr.io/openhands/agent-server:vasco-fix-hooks-java
ghcr.io/openhands/agent-server:780f5d1-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:780f5d1-python
ghcr.io/openhands/agent-server:780f5d157578be5a4754b2b65cd2fe58acce63f5-python
ghcr.io/openhands/agent-server:vasco-fix-hooks-python
ghcr.io/openhands/agent-server:780f5d1-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `780f5d1-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `780f5d1-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -98,7 +98,6 @@ class LocalConversation(BaseConversation):
         token_callbacks: list[ConversationTokenCallbackType] | None = None,
         hook_config: HookConfig | None = None,
         max_iteration_per_run: int = 500,
-        max_consecutive_stop_denials: int = 5,
         stuck_detection: bool = True,
         stuck_detection_thresholds: (
             StuckDetectionThresholds | Mapping[str, int] | None
@@ -110,6 +109,7 @@ class LocalConversation(BaseConversation):
         delete_on_close: bool = True,
         cipher: Cipher | None = None,
         tags: dict[str, str] | None = None,
+        max_consecutive_stop_denials: int = 5,
         **_: object,
     ):
         """Initialize the conversation.
@@ -800,19 +800,20 @@ class LocalConversation(BaseConversation):
                             )
                             if not should_stop:
                                 logger.info("Stop hook denied agent stopping")
-                                if consecutive_stop_denials > 0 and any(
-                                    isinstance(self._state.events[i], ActionEvent)
-                                    for i in range(
-                                        events_count_at_last_deny,
-                                        len(self._state.events),
-                                    )
-                                ):
-                                    consecutive_stop_denials = 0
+                                if consecutive_stop_denials > 0:
+                                    recent_events = self._state.events[
+                                        events_count_at_last_deny:
+                                    ]
+                                    if any(
+                                        isinstance(e, ActionEvent)
+                                        for e in recent_events
+                                    ):
+                                        consecutive_stop_denials = 0
                                 consecutive_stop_denials += 1
 
                                 if (
                                     consecutive_stop_denials
-                                    > self.max_consecutive_stop_denials
+                                    >= self.max_consecutive_stop_denials
                                 ):
                                     error_msg = (
                                         f"Stop hook denied {consecutive_stop_denials}"
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -75,6 +75,7 @@ class LocalConversation(BaseConversation):
     _on_event: ConversationCallbackType
     _on_token: ConversationTokenCallbackType | None
     max_iteration_per_run: int
+    max_consecutive_stop_denials: int
     _stuck_detector: StuckDetector | None
     llm_registry: LLMRegistry
     _cleanup_initiated: bool
@@ -97,6 +98,7 @@ class LocalConversation(BaseConversation):
         token_callbacks: list[ConversationTokenCallbackType] | None = None,
         hook_config: HookConfig | None = None,
         max_iteration_per_run: int = 500,
+        max_consecutive_stop_denials: int = 5,
         stuck_detection: bool = True,
         stuck_detection_thresholds: (
             StuckDetectionThresholds | Mapping[str, int] | None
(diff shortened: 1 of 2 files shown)


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -768,6 +774,8 @@ def run(self) -> None:
                 self._state.execution_status = ConversationExecutionStatus.RUNNING
 
         iteration = 0
+        consecutive_stop_denials = 0
+        events_count_at_last_deny = 0
```
xingyaoww, line 822: I wonder if we should move this consecutive stop denial into a field inside the conversation state rather than here. Because otherwise, if this local conversation actually got stopped, this counter will just go away. We should ideally persist this.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -792,6 +800,40 @@ def run(self) -> None:
...
+                                        ConversationExecutionStatus.ERROR
+                                    )
+                                    self._on_event(
+                                        ConversationErrorEvent(
+                                            source="environment",
+                                            code="StopHookLoopDetected",
+                                            detail=error_msg,
+                                        )
+                                    )
+                                    break
```
xingyaoww: Also, I think we should consider maintaining this logic in either a helper function, as part of the state, or even as part of the hook processor (specifically the hook_processor). I think that's probably the best place to maintain everything about hooks.

Basically, when we hit five consecutive stops, the hook processor return "should_stop" variable should just be false. We can probably throw in relative feedback there as well.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: This counter is scoped to a single run() on purpose. 

I was thinking about the case where run() exits and gets called again, then resetting feels right because the user had a chance to step in, so the next attempt gets a fresh budget.

Persisting would only matter if we crashed mid-loop and resumed, but by then we've already set ERROR + emitted StopHookLoopDetected, so there's nothing to resume into.

But happy to discuss this behaviour.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: The streak-reset reads `ActionEvents` off the event stream, which the hook processor shouldn't know about, and we will need conversation state passed in on every call.

Another point is taht hitting the limit isn't "allow the stop", it's ERROR + StopHookLoopDetected. Collapsing to should_stop=True would hide the fact that we force-aborted. Keeping it in run() puts it next to the other termination paths (MaxIterationsReached, stuck detector).

I agree that it is not so nice to read, I extract the deny-handling into a small helper method

VascoSch92:
I propse to run a small eval (50) to be sure I'm not breaking anything :-)

PR #3234, opened 2026-05-12
feat(agent-server): typed settings update endpoints + ETag/If-Match

rbren:
_This PR was opened by an AI agent (OpenHands) on behalf of the user._

## What's wrong with `PATCH /api/settings` today

The current endpoint accepts:

```python
class SettingsUpdateRequest(BaseModel):
    agent_settings_diff: dict[str, Any] | None = None
    conversation_settings_diff: dict[str, Any] | None = None
```

and applies the diffs via `_deep_merge` onto the persisted settings. A few of the consequences:

1. **The API contract is `dict[str, Any]`.** Generated clients are useless; typos like `lllm` are silently merged and forgotten; field renames in `OpenHandsAgentSettings` aren't even breaking changes at the wire level.
2. **The public schema lies about itself.** The internal `SettingsUpdatePayload` accepts an `active_profile` key that the public `SettingsUpdateRequest` doesn't expose. The handler `cast`s past this mismatch.
3. **"Diff" is undocumented deep-merge.** Dicts merge, lists replace, and there is no way to *delete* a key — to clear `mcp.headers["Authorization"]` you must overwrite the parent dict, which has all the read-modify-write hazards below.
4. **Discriminated-union variants are not respected.** A PATCH that switches `agent_kind: openhands → acp` ends up with leftover OpenHands-only fields silently dropped by re-validation, with no warning.
5. **No optimistic concurrency.** The file lock prevents on-disk corruption, *not* lost updates. Two clients each doing `GET → modify → PATCH` cheerfully clobber each other; the server has no ETag/If-Match/version to detect it. Most concerning case: list/dict fields (`tools`, `mcp_config.mcpServers`) which *replace wholesale*, so concurrent edits to *different* MCP servers silently undo one another.
6. **No way for a client to confirm what was stored.** The PATCH response intentionally strips secrets and there is no version stamp, so verifying "did my change land?" requires a follow-up `GET` whose result may already include another writer's interleaved changes.
7. **Generic 422 with no field info.** Even errors that carry no secret (`max_iterations: 0`) collapse to `"Settings validation failed"`. The client cannot point at the bad field.

Idempotency is *accidentally* preserved for the common single-client case, but breaks under concurrent writers, encrypted-secret round-trips, and migration scenarios.

### 1. Typed update models in the SDK (`openhands.sdk.settings.update_models`)

```python
class OpenHandsAgentSettingsUpdate(_UpdateBase):  # extra="forbid"
    agent_kind: Literal["openhands", "llm"] = "openhands"
    agent: str | None = None
    llm: LLMUpdate | None = None          # all-Optional partial of LLM
    tools: list[Tool] | None = None
    enable_sub_agents: bool | None = None
    enable_switch_llm_tool: bool | None = None
    mcp_config: MCPConfig | None = None
    agent_context: AgentContext | None = None
    condenser: CondenserSettingsUpdate | None = None
    verification: VerificationSettingsUpdate | None = None

class ACPAgentSettingsUpdate(_UpdateBase): ...
class ConversationSettingsUpdate(_UpdateBase): ...

AgentSettingsUpdate = Annotated[..., Discriminator(_agent_update_discriminator)]
```

With `apply_agent_update(current, update)` and `apply_conversation_update(current, update)` helpers that:

- Merge only the fields the client explicitly sent (`model_fields_set`).
- Treat nested partials (`llm`, `condenser`, `verification`) as one-level merges into the current sub-object, so e.g. `{"llm": {"model": "gpt-4o"}}` preserves the existing `api_key`.
- Re-validate the merged result through the source model's full validator chain (constraints, secret validators, MCP env decryption — all the usual invariants).
- Refuse to switch `agent_kind` (use `PUT` to switch variants).

### 2. New endpoints in `settings_router.py`

```
PATCH /api/settings/agent          — typed partial update (extra="forbid")
PUT   /api/settings/agent          — full replace (allows variant switching)
PATCH /api/settings/conversation   — typed partial update
PUT   /api/settings/conversation   — full replace
```

All four:

- Read the body as a Pydantic model with `extra="forbid"` (typos → 422).
- Honour `If-Match` against the global settings ETag.
- Return the new resource state with `ETag` header in the response.
- Reuse the existing file-locked `store.update()` so writes remain atomic and serialise with each other and with the legacy endpoint.

### 3. ETag / If-Match plumbing (`_settings_etag.py`)

- `compute_settings_etag()` hashes a **plaintext-canonical** projection of `PersistedSettings`. Crucially **not** the on-disk bytes — Fernet nonces would make identical state look like a change on every save, defeating idempotency.
- `check_if_match()` returns `412 Precondition Failed` on stale `If-Match`, with the current ETag echoed in the response so the client can retry against the live state. `If-Match: *` is accepted (resource always exists, since settings default).
- The shared `_run_settings_update` helper performs the If-Match check **inside** the file lock, so the ETag and the state the apply will run against are the same.

### 4. Bonus fix in `api.py`

The shared `HTTPException` handler in `_add_exception_handlers` was constructing the `JSONResponse` from `status_code` + `detail` but dropping `exc.headers`, so any handler-set response header on an error (e.g. our 412 ETag, or a `WWW-Authenticate: ...` on 401) was lost. Propagate it.

### 5. Legacy endpoint preserved + upgraded

`PATCH /api/settings` is unchanged in semantics and still accepts the old `dict[str, Any]` diff payload. In addition it now:

- Honours `If-Match` and emits the new `ETag` header in success responses.
- Emits `Deprecation: true` and `Link: </api/settings/agent>; rel="successor-version", </api/settings/conversation>; rel="successor-version"` (RFC 8594 / RFC 9745) so forward-compatible clients can migrate at their leisure.

No existing tests were modified; the legacy suite (28 tests) still passes unchanged.

## Tests

`tests/agent_server/test_typed_settings_router.py` (21 tests) covers:

- ETag stability across identical writes (the core idempotency property).
- ETag changes on real change.
- `If-Match` matching, stale → 412, wildcard.
- Legacy `PATCH /api/settings` honouring `If-Match` and emitting the deprecation headers.
- Typed PATCH: `extra="forbid"`, variant invariance, `agent_kind: "llm"` alias, partial LLM preserving `api_key`.
- PUT: full replace, variant switch (`openhands` → `acp`), 422 on invalid body.
- Conversation: partial preserve, forbid extra, validation re-fires.
- End-to-end concurrency: documented loss without `If-Match`, detected with `If-Match`.

Legacy: 28/28 `tests/agent_server/test_settings_router.py` still pass.
SDK: 87/87 `tests/sdk/test_settings.py` + `tests/sdk/settings/` still pass.

(The 15 unrelated failures in `tests/agent_server/test_skills_router.py` and `test_terminal_service.py` exist on `main` before this PR.)

## Compatibility

- Wire-compatible: no existing request body shape changes; `SettingsUpdateRequest` is unchanged.
- New response *header* (`ETag`) on `GET /api/settings`, `PATCH /api/settings`, and the new endpoints. Old clients that ignore unknown headers see no change.
- `Deprecation` / `Link` on the legacy `PATCH /api/settings` response are advisory; old clients ignore them.

## Follow-ups (not in this PR)

- **MCP servers should move to per-server CRUD** (`GET /api/settings/mcp-servers`, `GET/PUT/PATCH/DELETE /api/settings/mcp-servers/{name}`). Today even the new typed PATCH still treats `mcp_config` as one blob: editing one server's `env` requires round-tripping every *other* server's encrypted secrets. The CRUD shape would also give per-server ETags so concurrent edits to different servers stop conflicting at all. Mirroring the existing `/api/settings/secrets` router design.
- **Strict-mode `If-Match`**: a config flag to require `If-Match` on all settings writes, for deployments that want concurrency safety to be mandatory rather than opt-in.
- **Field-level 422 errors**: today the typed handlers still collapse re-validation errors to a generic message to avoid leaking secret values in tracebacks. We could pass through `loc` + a sanitised `msg` for fields known not to carry secrets.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:7d67356-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-7d67356-python \
  ghcr.io/openhands/agent-server:7d67356-python
```

```
ghcr.io/openhands/agent-server:7d67356-golang-amd64
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-golang-amd64
ghcr.io/openhands/agent-server:feature-typed-settings-update-golang-amd64
ghcr.io/openhands/agent-server:7d67356-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:7d67356-golang-arm64
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-golang-arm64
ghcr.io/openhands/agent-server:feature-typed-settings-update-golang-arm64
ghcr.io/openhands/agent-server:7d67356-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:7d67356-java-amd64
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-java-amd64
ghcr.io/openhands/agent-server:feature-typed-settings-update-java-amd64
ghcr.io/openhands/agent-server:7d67356-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:7d67356-java-arm64
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-java-arm64
ghcr.io/openhands/agent-server:feature-typed-settings-update-java-arm64
ghcr.io/openhands/agent-server:7d67356-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:7d67356-python-amd64
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-python-amd64
ghcr.io/openhands/agent-server:feature-typed-settings-update-python-amd64
ghcr.io/openhands/agent-server:7d67356-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:7d67356-python-arm64
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-python-arm64
ghcr.io/openhands/agent-server:feature-typed-settings-update-python-arm64
ghcr.io/openhands/agent-server:7d67356-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:7d67356-golang
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-golang
ghcr.io/openhands/agent-server:feature-typed-settings-update-golang
ghcr.io/openhands/agent-server:7d67356-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:7d67356-java
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-java
ghcr.io/openhands/agent-server:feature-typed-settings-update-java
ghcr.io/openhands/agent-server:7d67356-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:7d67356-python
ghcr.io/openhands/agent-server:7d67356e72102f378012ac81dbd67f569485f8cd-python
ghcr.io/openhands/agent-server:feature-typed-settings-update-python
ghcr.io/openhands/agent-server:7d67356-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `7d67356-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `7d67356-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/_settings_etag.py
+++ b/openhands-agent-server/openhands/agent_server/_settings_etag.py
@@ -1,0 +1,91 @@
+"""ETag and ``If-Match`` helpers for settings endpoints.
+
+The legacy ``PATCH /api/settings`` had no version stamp and no
+``If-Match`` support, so concurrent writes silently lost each other (the
+file lock only prevented on-disk corruption, not lost updates). The new
+typed endpoints emit an ``ETag`` on every read and write and accept
+``If-Match`` for optimistic concurrency.
+
+ETag stability
+--------------
+
+The ETag is computed over a *plaintext-canonical* projection of the
+persisted settings, **not** over the encrypted bytes on disk. Fernet
+includes a fresh nonce on every encryption, so two saves of the same
+state produce different ciphertexts — hashing those would defeat the
+ETag entirely (identical PATCHes would each look like a change).
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+
+from fastapi import HTTPException, status
+
+from openhands.agent_server.persistence import PersistedSettings
+
+
+def compute_settings_etag(settings: PersistedSettings) -> str:
+    """Compute a stable, quoted ETag over the plaintext-canonical state.
+
+    Hashes a sort-keyed, separator-canonical JSON of the plaintext
+    serialisation. Returns a quoted hex string suitable for the
+    ``ETag``/``If-Match`` HTTP headers.
+    """
+    canonical = settings.model_dump(
+        mode="json", context={"expose_secrets": "plaintext"}
+    )
+    blob = json.dumps(canonical, sort_keys=True, separators=(",", ":")).encode("utf-8")
+    digest = hashlib.sha256(blob).hexdigest()[:32]
+    return f'"{digest}"'
+
+
+def parse_if_match(if_match: str | None) -> set[str] | None:
+    """Parse an ``If-Match`` header value into a set of acceptable ETags.
+
+    Returns ``None`` when no precondition is supplied. Returns an empty
+    sentinel when the wildcard ``*`` is supplied (caller treats this as
+    "resource must exist", which always holds because settings default).
+    """
+    if if_match is None:
+        return None
+    candidates = {tok.strip() for tok in if_match.split(",") if tok.strip()}
+    if "*" in candidates:
+        return {"*"}
+    return candidates
+
+
+def check_if_match(if_match: str | None, current_etag: str) -> None:
+    """Raise ``412 Precondition Failed`` if ``If-Match`` does not match.
+
+    Behaviour
+    ---------
+    * ``If-Match`` absent: no precondition is enforced — the request is
+      allowed through. (Strict mode is left to deployments that wrap this
+      with middleware; we default to permissive so existing single-client
+      flows keep working.)
+    * ``If-Match: *``: precondition succeeds because the settings
+      resource always exists (defaults are returned when no file is on
+      disk).
+    * ``If-Match: "etag"`` (comma-separated list allowed): the current
+      ETag must be in the set, otherwise 412.
+
+    The response carries the current ETag back to the client so it can
+    retry against the live state.
+    """
+    candidates = parse_if_match(if_match)
+    if candidates is None or "*" in candidates:
+        return
+    if current_etag not in candidates:
+        raise HTTPException(
+            status_code=status.HTTP_412_PRECONDITION_FAILED,
+            detail=(
+                "If-Match does not match the current settings ETag. "
+                "Re-fetch /api/settings and retry with the new ETag."
+            ),
+            headers={"ETag": current_etag},
+        )
+
+
+__all__ = ["check_if_match", "compute_settings_etag", "parse_if_match"]
--- a/openhands-agent-server/openhands/agent_server/api.py
+++ b/openhands-agent-server/openhands/agent_server/api.py
@@ -493,10 +493,18 @@ def _add_exception_handlers(api: FastAPI) -> None:
             return JSONResponse(
                 status_code=exc.status_code,
                 content=content,
+                headers=exc.headers,
             )
 
-        # Return clean JSON response for all non-5xx HTTP exceptions
-        return JSONResponse(status_code=exc.status_code, content={"detail": exc.detail})
+        # Return clean JSON response for all non-5xx HTTP exceptions.
+        # Propagate ``exc.headers`` so handlers can attach response headers
+        # to error responses (e.g. ``ETag`` on a 412 from optimistic-
+        # concurrency checks, or ``WWW-Authenticate`` on a 401).
+        return JSONResponse(
+            status_code=exc.status_code,
+            content={"detail": exc.detail},
+            headers=exc.headers,
+        )
 
 
 def create_app(config: Config | None = None) -> FastAPI:
--- a/openhands-agent-server/openhands/agent_server/settings_router.py
+++ b/openhands-agent-server/openhands/agent_server/settings_router.py
@@ -1,7 +1,8 @@
+from collections.abc import Callable
 from functools import lru_cache
-from typing import cast
+from typing import Any, cast
 
-from fastapi import APIRouter, HTTPException, Request, Response, status
+from fastapi import APIRouter, Header, HTTPException, Request, Response, status
 from pydantic import ValidationError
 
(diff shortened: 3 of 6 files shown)


neubig:
Closing for lack of activity

PR #3235, opened 2026-05-12
feat(agent-server): per-server CRUD for MCP servers

rbren:
_This PR was opened by an AI agent (OpenHands) on behalf of the user._

Follow-up to #3234 (typed settings update endpoints) — that PR called out the MCP-server-as-blob problem; this PR fixes it. The two are independent and don't share code; the small `api.py` `exc.headers` propagation fix appears in both with identical content (trivial merge resolution either way).

## Why `mcp_config` is the worst-fit case for the settings PATCH

MCP servers live as a dict under `agent_settings.mcp_config.mcpServers`. Editing one through the global settings PATCH means:

1. **Read-modify-write of the whole collection.** Even with deep-merge semantics, to add or remove a server atomically you have to PATCH the full new `mcpServers` mapping — which forces the client to ship every *other* server's encrypted `env`/`headers` ciphertexts that they had no business round-tripping.
2. **Cross-server concurrency at the wrong granularity.** Two clients each adding a *different* server race, and one of them has to retry, having done no work that conflicts with the other.
3. **Discoverability.** `mcp_config` is documented in OpenAPI as one shape; adding/removing/editing a server is undocumented and indirect.
4. **`env` / `headers` are dicts of secrets.** Every byte that doesn't need to leave the server shouldn't.

## What this PR adds

A dedicated router that treats each MCP server as its own named resource:

| Verb     | Path                                       | Purpose                                                                |
| -------- | ------------------------------------------ | ---------------------------------------------------------------------- |
| `GET`    | `/api/settings/mcp-servers`                | List summaries (name, transport kind, description, icon — no secrets). |
| `GET`    | `/api/settings/mcp-servers/{name}`         | Read one (honours `X-Expose-Secrets` like `/api/settings`).            |
| `PUT`    | `/api/settings/mcp-servers/{name}`         | Upsert. 201 on create, 200 on update.                                  |
| `PATCH`  | `/api/settings/mcp-servers/{name}`         | Partial update of an existing server. 404 if absent.                   |
| `DELETE` | `/api/settings/mcp-servers/{name}`         | Remove one. 404 if absent.                                             |

Body for PUT/PATCH is the raw `fastmcp` server config dict (the value of `MCPConfig.mcpServers[name]`). fastmcp's own validators discriminate stdio vs remote vs the transforming variants based on `command`/`url` presence; this router passes through. Cross-server invariants from `MCPConfig` re-fire on every write because each mutation re-validates the full `OpenHandsAgentSettings` after splicing.

### Per-server optimistic concurrency

- `ETag` on every read of `/{name}`. Computed over a **plaintext-canonical** JSON projection of just that server's config — *not* on-disk bytes, since Fernet nonces would make identical state hash differently every save and defeat idempotency.
- `If-Match: <etag>` on PUT/PATCH/DELETE → 412 (with current `ETag` echoed in the header) if stale.
- `If-Match: *` → must exist (update-only PUT).
- `If-None-Match: *` → must not exist (create-only PUT).
- `If-None-Match: <etag>` → 412 if it matches current (symmetric to `If-Match`).

The headline win: **concurrent edits to *different* servers don't conflict at all**. Same-server conflicts get a clean 412 with the new state's ETag, so the client can rebase and retry.

### `X-Expose-Secrets` on GET

Same three modes as `GET /api/settings`:
- absent → `env`/`headers` values are `[REDACTED]`
- `encrypted` → Fernet ciphertext
- `plaintext` → raw values

Logged at WARNING level when plaintext is requested, just like the existing settings endpoint.

### ACP variant handling

`ACPAgentSettings` has no `mcp_config` field. Choices for each verb:
- `GET /api/settings/mcp-servers` → `200 {"servers": []}` (list is non-modifying; permissive shape so polling clients stay simple)
- `GET /api/settings/mcp-servers/{name}` → `404`
- `PUT/PATCH/DELETE` → `409 Conflict` with a message pointing the caller at `PATCH /api/settings` to switch agent variant.

### Persistence

All writes go through the existing file-locked `store.update()`, so they remain atomic and serialise with each other and with the global `PATCH /api/settings`. No new persistence backend, no parallel lock. Each write rebuilds the full `OpenHandsAgentSettings` post-splice and re-validates, so encryption-on-save / decryption-on-load all run through the existing tested paths.

## Bonus fix in `api.py`

The shared `HTTPException` handler was constructing the `JSONResponse` from `status_code` + `detail` but dropping `exc.headers`, so any handler-set response header on an error (e.g. our 412 ETag, or `WWW-Authenticate` on a 401) was silently lost. Three-line propagation fix; identical to the same fix in #3234 so either merge order is trivial to resolve.

## SDK additions

New public types in `openhands.sdk.settings`:
- `MCPServerSummary` — list-view shape (no secrets).
- `MCPServerListResponse` — wraps the list.
- `MCPServerResponse` — single-server read/write response: `{name, config: dict[str, Any]}`. `config` is `dict[str, Any]` (not a typed model) so the server controls secret serialization via context the same way `SettingsResponse.agent_settings` does.

## Tests

`tests/agent_server/test_mcp_servers_router.py` — 32 tests covering:

- List / read / upsert / patch / delete happy paths.
- ETag stability across identical resaves (the core idempotency property), changes on real change.
- `If-Match` matching / stale, `If-None-Match: *` create-only, `If-Match: *` update-only, stale 412 returns current ETag.
- Concurrent edits to *different* servers don't conflict; concurrent edits to the *same* server: one 200, one 412.
- PATCH preserves unsent fields; 404 if server doesn't exist; 422 on invalid type.
- DELETE returns 204, 404 on missing, 412 on stale `If-Match`, doesn't disturb other servers.
- ACP variant: list empty, GET 404, PUT 409, DELETE 409.
- `X-Expose-Secrets` honoured: default redacted, `plaintext` raw, `encrypted` Fernet token.
- Splice-locally property: editing server B doesn't require the client to round-trip server A's env.
- Name validator: allows `[A-Za-z0-9_.@:-]{1,128}`; slashes rejected (URL path segment).

## Compatibility

- Strictly additive. No existing endpoints, schemas, or persistence shapes change.
- The global `PATCH /api/settings` continues to work for mcp_config exactly as today; clients can migrate incrementally.
- The `api.py` exception-handler fix is a tiny improvement that only adds headers from `HTTPException(headers=...)` that route handlers explicitly set; it never alters the body, status code, or removes anything.

## Test summary

- `tests/agent_server/test_mcp_servers_router.py`: 32 / 32 pass.
- `tests/agent_server/test_settings_router.py`: 28 / 28 pass (unchanged).
- `tests/sdk/test_settings.py` + `tests/sdk/settings/*`: 87 / 87 pass (unchanged).

## Out of scope (further follow-ups)

- **List-level ETag.** Per-server ETag is sufficient for the mutation flow; a list-level ETag would help bulk-fetch clients avoid re-walking unchanged lists, but it's a cache-validation feature rather than a correctness one.
- **Tools sub-collection.** `mcp_config.mcpServers[name].tools` has its own per-tool transformation config — if that becomes a hot edit target, a `/mcp-servers/{name}/tools/{tool}` sub-router would apply the same logic at the next level down.
- **Single-server ETag bound to global settings ETag.** Currently per-server ETags float independently of the global settings ETag; for clients that want to ensure their MCP edit didn't race against a concurrent `PATCH /api/settings` mutating something else, we could expose a compound ETag. Not needed for the per-server flow.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:d75b1fe-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-d75b1fe-python \
  ghcr.io/openhands/agent-server:d75b1fe-python
```

```
ghcr.io/openhands/agent-server:d75b1fe-golang-amd64
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-golang-amd64
ghcr.io/openhands/agent-server:feature-mcp-server-crud-golang-amd64
ghcr.io/openhands/agent-server:d75b1fe-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:d75b1fe-golang-arm64
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-golang-arm64
ghcr.io/openhands/agent-server:feature-mcp-server-crud-golang-arm64
ghcr.io/openhands/agent-server:d75b1fe-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:d75b1fe-java-amd64
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-java-amd64
ghcr.io/openhands/agent-server:feature-mcp-server-crud-java-amd64
ghcr.io/openhands/agent-server:d75b1fe-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:d75b1fe-java-arm64
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-java-arm64
ghcr.io/openhands/agent-server:feature-mcp-server-crud-java-arm64
ghcr.io/openhands/agent-server:d75b1fe-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:d75b1fe-python-amd64
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-python-amd64
ghcr.io/openhands/agent-server:feature-mcp-server-crud-python-amd64
ghcr.io/openhands/agent-server:d75b1fe-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:d75b1fe-python-arm64
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-python-arm64
ghcr.io/openhands/agent-server:feature-mcp-server-crud-python-arm64
ghcr.io/openhands/agent-server:d75b1fe-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:d75b1fe-golang
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-golang
ghcr.io/openhands/agent-server:feature-mcp-server-crud-golang
ghcr.io/openhands/agent-server:d75b1fe-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:d75b1fe-java
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-java
ghcr.io/openhands/agent-server:feature-mcp-server-crud-java
ghcr.io/openhands/agent-server:d75b1fe-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:d75b1fe-python
ghcr.io/openhands/agent-server:d75b1fe2ff75ad4c180433e43e45f2bd1f4bbe00-python
ghcr.io/openhands/agent-server:feature-mcp-server-crud-python
ghcr.io/openhands/agent-server:d75b1fe-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `d75b1fe-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `d75b1fe-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/api.py
+++ b/openhands-agent-server/openhands/agent_server/api.py
@@ -38,6 +38,7 @@ from openhands.agent_server.file_router import file_router
 from openhands.agent_server.git_router import git_router
 from openhands.agent_server.hooks_router import hooks_router
 from openhands.agent_server.llm_router import llm_router
+from openhands.agent_server.mcp_servers_router import mcp_servers_router
 from openhands.agent_server.middleware import LocalhostCORSMiddleware
 from openhands.agent_server.profiles_router import profiles_router
 from openhands.agent_server.server_details_router import (
@@ -289,6 +290,7 @@ def _add_api_routes(app: FastAPI, config: Config) -> None:
     api_router.include_router(hooks_router)
     api_router.include_router(llm_router)
     api_router.include_router(settings_router)
+    api_router.include_router(mcp_servers_router)
     api_router.include_router(profiles_router)
     api_router.include_router(cloud_proxy_router)
     # /api/auth/* mints workspace cookies and requires the header to bootstrap,
@@ -493,10 +495,18 @@ def _add_exception_handlers(api: FastAPI) -> None:
             return JSONResponse(
                 status_code=exc.status_code,
                 content=content,
+                headers=exc.headers,
             )
 
-        # Return clean JSON response for all non-5xx HTTP exceptions
-        return JSONResponse(status_code=exc.status_code, content={"detail": exc.detail})
+        # Return clean JSON response for all non-5xx HTTP exceptions.
+        # Propagate ``exc.headers`` so handlers can attach response headers to
+        # error responses (e.g. ``ETag`` on a 412 from optimistic-concurrency
+        # checks, or ``WWW-Authenticate`` on a 401).
+        return JSONResponse(
+            status_code=exc.status_code,
+            content={"detail": exc.detail},
+            headers=exc.headers,
+        )
 
 
 def create_app(config: Config | None = None) -> FastAPI:
--- a/openhands-agent-server/openhands/agent_server/mcp_servers_router.py
+++ b/openhands-agent-server/openhands/agent_server/mcp_servers_router.py
@@ -1,0 +1,665 @@
+"""CRUD endpoints for MCP servers persisted inside agent settings.
+
+Why a dedicated router (rather than going through ``PATCH /api/settings``)
+
+Today, MCP servers live as a dict under
+``agent_settings.mcp_config.mcpServers``. Editing a single server through
+the global settings PATCH means:
+
+* Read-modify-write of the *whole* collection (because ``mcp_config`` is a
+  single Pydantic field that replaces wholesale on assignment-style PATCH;
+  even with deep-merge, ``mcpServers`` is a dict where any individual
+  key-replace via PATCH affects only that key but still forces the client
+  to ship the entire ``mcpServers`` it wants to keep when *adding* or
+  *removing* anything else atomically). In practice clients fetch the
+  whole encrypted config, splice in their edit, and write the whole thing
+  back — which round-trips every *other* server's ``env``/``headers``
+  ciphertexts through the client unnecessarily.
+
+* No per-server concurrency. Two clients each adding a *different* server
+  race; even with optimistic concurrency at the global-settings level,
+  one of them has to retry, having done no work that conflicts with the
+  other.
+
+* No discoverable shape — ``mcp_config`` is documented in OpenAPI as one
+  big blob; adding a server is undocumented in the API surface.
+
+This router exposes the per-server resource directly:
+
+* ``GET    /api/settings/mcp-servers``         list summaries (no secrets)
+* ``GET    /api/settings/mcp-servers/{name}``  read one (X-Expose-Secrets aware)
+* ``PUT    /api/settings/mcp-servers/{name}``  upsert one
+* ``PATCH  /api/settings/mcp-servers/{name}``  partial edit of one
+* ``DELETE /api/settings/mcp-servers/{name}``  remove one
+
+Per-server ``ETag`` / ``If-Match`` (and ``If-None-Match: *`` for
+create-only PUT) gives optimistic concurrency at the right granularity:
+concurrent edits to *different* servers stop conflicting at all;
+concurrent edits to the *same* server get a clean 412 with the current
+ETag echoed back.
+
+ACP variant: ``ACPAgentSettings`` has no ``mcp_config``. Read endpoints
+on the missing collection return ``404``; writes return ``409`` with a
+message pointing the caller at the agent-variant settings endpoint.
+
+All writes still go through the file-locked ``store.update()``, so they
+serialize with each other and with the global ``PATCH /api/settings`` —
+this router does not introduce a separate persistence backend, just a
+finer-grained API.
+"""
+
+from __future__ import annotations
+
+import copy
+import hashlib
+import json
+import re
+from typing import Any
+
+from fastapi import APIRouter, HTTPException, Request, Response, status
+from pydantic import ValidationError
+
+from openhands.agent_server._secrets_exposure import (
+    build_expose_context,
+    get_config,
+    parse_expose_secrets_header,
+    translate_missing_cipher,
+)
+from openhands.agent_server.persistence import (
+    PersistedSettings,
+    get_settings_store,
+)
+from openhands.sdk.logger import get_logger
+from openhands.sdk.settings import (
+    MCPServerListResponse,
+    MCPServerResponse,
+    MCPServerSummary,
+    OpenHandsAgentSettings,
+)
+
+
+logger = get_logger(__name__)
+
+mcp_servers_router = APIRouter(tags=["MCP Servers"])
+
+MCP_COLLECTION_PATH = "/settings/mcp-servers"
+MCP_ITEM_PATH = "/settings/mcp-servers/{name}"
+
+# MCP server names map 1:1 to dict keys in the underlying ``MCPConfig``;
+# constrain to safe URL-path characters. Slashes are explicitly disallowed
+# because they would collide with FastAPI's path-segment routing (a server
+# called "@scope/server" would never match ``/mcp-servers/{name}``). Names
+# stay 1–128 chars to bound on-disk size.
+_NAME_PATTERN = re.compile(r"^[A-Za-z0-9_.@:-]{1,128}$")
+
+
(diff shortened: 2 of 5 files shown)


enyst:
@openhands read the PR linked in this PR's description too for context. And /codereview-roasted this PR. Respond directly on github.

enyst (commented):
Read #3234 for context. The per-server resource shape is the right direction, and the ETag/error-header plumbing is solid. But there’s one real data-loss bug here that keeps this off approval.

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-agent-server/openhands/agent_server/mcp_servers_router.py:313`](https://github.com/OpenHands/software-agent-sdk/pull/3235/files#diff-0f43ff9d5f70b0906cf2ea4e158296adf9547268547fe90d893b3df3d6559614R313) **Breaking Change / Data Loss**: `_validate_servers_dict()` reconstructs `mcp_config` as just `{"mcpServers": servers}`, and the write paths then assign that back onto `agent_dump["mcp_config"]`. `fastmcp.MCPConfig` is `additionalProperties=True`, so top-level keys besides `mcpServers` are valid and round-trip today. This router silently deletes them on any PUT/PATCH/DELETE.

  I reproduced this locally by seeding settings through the legacy endpoint with:
  ```json
  {"mcp_config": {"mcpServers": {"a": {"command": "echo"}}, "extraTop": {"a": 1}}}
  ```
  then calling `PATCH /api/settings/mcp-servers/a`, after which `extraTop` disappeared from persisted settings.

  That’s a bad boundary bug: the new fine-grained API is mutating more than the named resource. The fix should preserve the full existing `mcp_config` payload and splice only `mcpServers`, then revalidate that full object.

**[TESTING GAPS]**
- `tests/agent_server/test_mcp_servers_router.py` **Missing regression coverage**: there’s no test that seeds `mcp_config` with a sibling top-level key, performs a per-server PUT/PATCH/DELETE, and verifies that sibling data survives the persistence round-trip. Add one; this bug is easy to reintroduce.

**[RISK ASSESSMENT]**
  The API design is good and the concurrency/secret handling coverage is strong, but the current implementation can silently drop valid persisted MCPConfig data on ordinary writes. That’s user-visible settings loss.

**VERDICT:**
❌ **Needs rework**: preserve the full `mcp_config` object when mutating `mcpServers`.

**KEY INSIGHT:**
Once `MCPConfig` comes from FastMCP, the safe unit of mutation is “full `mcp_config`, replace only `mcpServers`” — not “rebuild `mcp_config` from scratch around `mcpServers`”.

---

>
>

---

PR #3237, opened 2026-05-13
chore(tools): replace bashlex (GPLv3) with tree-sitter-bash (MIT)

VascoSch92:
- [x] A human has tested these changes.

## Why

bashlex is GPLv3+ while this repo is MIT — a transitive copyleft dependency we want to drop. It's also unmaintained (last meaningful release ~2020) and crashes on a variety of valid shell constructs, which is why the wrapper currently catches four exception types and silently falls back to the raw string.

tree-sitter and tree-sitter-bash are both MIT, actively maintained, and recover gracefully on malformed input (an ERROR node in the tree, not an exception).

## Summary

- openhands-tools/pyproject.toml: bashlex>=0.18 → tree-sitter>=0.24, tree-sitter-bash>=0.23
- openhands-tools/openhands/tools/terminal/utils/command.py: rewritten against tree-sitter (130 lines, 19 fewer than the bashlex version)
- uv.lock: regenerated

## API & behavior

Public surface unchanged: split_bash_commands(commands: str) -> list[str] and escape_bash_special_chars(command: str) -> str.

Splitting contract preserved: statements separated by a newline split into separate entries; statements joined by ;, &&, ||, |, or & stay together. Comments and whitespace between statements fold into the preceding entry. On parse failure the input is returned unchanged — same silent-passthrough as the bashlex era.

## Issue Number
Refs #2721

## How to Test
- tests/tools/terminal/test_terminal_parsing.py: 30/30 pass, unchanged.
- tests/tools/terminal/test_terminal_session.py integration tests touching split_bash_commands (test_long_running_command_follow_by_execute, test_complex_commands, test_multiple_multiline_commands,
  test_multiline_command_loop) pass on both tmux and subprocess backends.
- Pre-commit clean: Ruff format/lint, pycodestyle, pyright, import dependency rules.

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
[...]

Issue #3176, opened 2026-05-09
Proposal: extend defense_in_depth PatternSecurityAnalyzer with Agent Threat Rules signatures


eeee2345:
The defense_in_depth module already ships a strong PatternSecurityAnalyzer with stable detector IDs and a careful two-corpus design that separates executable scanning from all-field scanning. The existing pattern set covers rm -rf, sudo rm, mkfs, dd, eval/exec/os.system/subprocess, curl-pipe-sh and wget-pipe-sh, and a small injection family. This is exactly the right shape for a maintained rule set.

I would like to propose extending the pattern coverage by mapping in detection signatures from the open Agent Threat Rules project at https://github.com/Agent-Threat-Rule/agent-threat-rules. ATR is Apache-2.0 and ships 330 rules across 9 categories with explicit MITRE ATLAS and OWASP Agentic Top 10 mappings. The same rule pack has shipped at Cisco AI Defense skill-scanner and Microsoft agent-governance-toolkit, so it is reasonable to consider as upstream rather than reinvent.

The integration is small and additive. I am proposing a new ATRPatternSet helper that loads a curated subset of ATR YAML rules at import time, converts them to the existing (regex_pattern, description, detector_id) tuple format, and merges them into DEFAULT_HIGH_PATTERNS / DEFAULT_MEDIUM_PATTERNS. Detector IDs would be prefixed atr.* to keep them visually distinct from the native exec.* and inject.* IDs you already use. No new runtime dependency. Pure Python, regex only, no network, no model inference. The patch would be under 150 lines plus tests.

Before I send a PR I want to check three things with maintainers. First, is the PatternSecurityAnalyzer pattern table the right place to add upstream-curated rules, or would you prefer a sibling analyzer that keeps the native and ATR rule sets isolated? Second, do you want ATR shipped as a vendored Python module (one-time copy of selected patterns with attribution comments) or as an optional install-time fetch from the ATR repo? My default would be vendored with comments pointing to upstream rule IDs so review is fully local. Third, are there detector ID naming conventions or telemetry contracts I should preserve so downstream consumers do not regress.

Happy to send a small initial PR with 5 to 10 high-precision patterns from ATR categories that do not overlap with what is already in pattern.py, focused on prompt injection variants, exfiltration domains, and credential leakage. License is Apache-2.0 on both sides, so attribution is the only requirement.

eeee2345:
[...]

PR #3307, opened 2026-05-19
refactor(sdk): introduce shared tree-sitter-bash entry point

Fieldnote-Echo:
## Summary

Phase 2a foundation for #2721. Creates a shared tree-sitter-bash parse
entry point inside `openhands-sdk` so the security analyzers can
import a parser without reaching across packages.

Specifically:

- Adds `openhands-sdk/openhands/sdk/security/shell_parser.py` with a
  single public function `parse(command: str) -> ParseResult` and a
  frozen `ParseResult` dataclass carrying `tree` and `has_error`.
- Declares `tree-sitter>=0.25` and `tree-sitter-bash>=0.25` in
  `openhands-sdk/pyproject.toml`. The 0.25 floor matches what the
  workspace lockfile already resolves and avoids the documented ABI
  mismatch between `tree-sitter==0.24.0` and `tree-sitter-bash==0.25.0`.
  `openhands-tools` keeps its direct `tree-sitter` declarations
  (unchanged here, per #3237) because it still imports those libraries
  directly via `terminal/utils/command.py`. Tightening that spec is a
  separate cleanup.
- Regenerates `uv.lock` to reflect the new declaration site.

## Out of scope (deliberate)

- No consumer is migrated. `openhands-tools/openhands/tools/terminal/utils/command.py`
  keeps its inline parser construction; a follow-up PR will switch it
  to `shell_parser.parse` and confirm `tests/tools/terminal/test_terminal_parsing.py`
  passes unchanged.
- No `iter_*` convenience accessors. The Phase 2c direction note on
  #2721 floats `iter_simple_commands` / `iter_pipelines` etc.; those
  are easier to design once at least one detector exists to validate
  them, so they land with Phase 2b or shortly after.
- No `ERROR`-node policy. Pattern and rail detectors are unchanged.

## Why the substrate change goes first

Following the suggestion to "divide the scope of the second part even
further": this is the smallest landable piece that unblocks Phase 2b
(security analyzers can now import a parser from the SDK without
acquiring a transitive dependency on `openhands-tools`) and Phase 2a
consumer refactor (a single import line change in `command.py`).
Reviewing one ~30-line module plus a dependency move is meaningfully
cheaper than reviewing the substrate together with either consumer
refactor.

## Phase 2 roadmap

This PR is the first link in a chain. Each later PR is small and independently mergeable.

| PR | Scope | Depends on |
|---|---|---|
| **PR-A (this PR)** | Shared `shell_parser.py` module + dep declaration. No consumer touched. | (rests on #3237) |
| PR-B | Refactor `openhands-tools/openhands/tools/terminal/utils/command.py` to consume `shell_parser.parse`. | PR-A |
| PR-D | Migrate `defense_in_depth/pattern.py` shell-destructive detectors to AST. The xfail bypass corpus from #3304 flips to passing under `strict=True`. | PR-A |
| PR-E | Migrate `defense_in_depth/policy_rails.py` rails (fetch-to-exec, raw-disk-op, catastrophic-delete) to AST. | PR-D |
| PR-F | Add `root_node` and `iter_*` convenience accessors to `ParseResult` once a detector validates the surface. | PR-D |

Collaboration on lanes PR-B onward is welcome. /cc @eeee2345 in case the analyzer-side work aligns with the Agent Threat Rules direction in #3176.

## Test plan

- [ ] `uv sync` resolves cleanly; `uv.lock` regenerated with
      `tree-sitter` / `tree-sitter-bash` declared by both packages.
- [ ] `tests/sdk/security/test_shell_parser.py` passes (new file, 18
      tests across success shape, error recovery, edge cases, immutability,
      adversarial bytes, byte-offset semantics, determinism).
- [ ] `tests/tools/terminal/test_terminal_parsing.py` passes unchanged
      (no source changes in `openhands-tools`).
- [ ] `pre-commit run --all-files` clean (ruff format, ruff lint,
      pyright, pycodestyle).

## Issue Number

Refs #2721


openhands-sdk/openhands/sdk/security/shell_parser.py
```diff
@@ -0,0 +1,78 @@
...
+
+This module intentionally exposes the minimum surface that both
+consumers need today: a single ``parse(command)`` function returning a
+``ParseResult`` that carries the parsed ``Tree`` and a precomputed
+``has_error`` flag. Convenience iterators and a public ``root_node``
+field are deferred to a follow-up so the first move is a no-behavior
+shared substrate rather than an API surface enlargement.
+
+The ``Parser`` is constructed per call. ``Language`` is built once at
+import. This mirrors the convention established by ``#3237`` in
```
VascoSch92: please don't refer to the issue number.

openhands-sdk/openhands/sdk/security/shell_parser.py
```diff
@@ -0,0 +1,78 @@
...
+``openhands-tools/openhands/tools/terminal/utils/command.py``: sharing
+one parser across calls risks interleaved state, while the language
+object is safely reusable.
+"""
+
+from __future__ import annotations
```
VascoSch92: do we need this import?

openhands-sdk/openhands/sdk/security/shell_parser.py
```diff
@@ -0,0 +1,78 @@
...
+from dataclasses import dataclass
+
+import tree_sitter_bash
+from tree_sitter import Language, Parser, Tree
+
+
+_BASH_LANGUAGE = Language(tree_sitter_bash.language())
+
+
+@dataclass(frozen=True)
```
VascoSch92: you can also add `slots=True`

openhands-sdk/openhands/sdk/security/shell_parser.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Encoding is UTF-8; the parser operates on bytes internally, so
+    callers reasoning about node spans should use byte offsets, not
+    code-point offsets, against ``command.encode()``.
+    """
+    tree = Parser(_BASH_LANGUAGE).parse(command.encode())
+    return ParseResult(tree=tree, has_error=tree.root_node.has_error)
+
+
+__all__ = ["ParseResult", "parse"]
```
VascoSch92: This should be at the top of the file

VascoSch92 (approved):
LGTM!

Just left a couple of comments.

tests/sdk/security/test_shell_parser.py
```diff
@@ -0,0 +1,139 @@
+"""Round-trip tests for the shared tree-sitter-bash parse entry point.
+
+These tests pin the contract that downstream consumers
+(``openhands-tools.terminal.utils.command`` and the Phase 2b
+security analyzers) rely on: ``parse`` returns a ``ParseResult``
+whose ``tree`` can be walked with the standard tree-sitter API and
+whose ``has_error`` flag faithfully reflects ``tree.root_node.has_error``.
+"""
+
+from __future__ import annotations
```
VascoSch92: Do we need that?

eeee2345:
@Fieldnote-Echo — thanks for the cc. The substrate split is the right call; reviewing a ~30-line module + dep declaration in isolation is materially easier than reviewing it against either consumer.

For the ATR adapter path (#3176), PR-A landing is what we want — the adapter eventually lives at the detector layer (sibling to `defense_in_depth/pattern.py`), so a shared `shell_parser.parse` entry point in `openhands-sdk/security/` is exactly the right shape to consume from. Will track PR-A → PR-D for the migration sequence; the ATR adapter's bash-pattern matchers will benefit from the same AST upgrade as `defense_in_depth/pattern.py` once that lands.

No review blockers from my side on this PR; just watching the chain.

VascoSch92 (approved):

PR #3349, opened 2026-05-21
fix(agent-server): remove redundant `with state:` from stats_callback (deadlock)

### LINKED ISSUE #3348: ACP conversations hang silently after prompt() returns — stats_callback deadlock on the state lock
## Summary

Every ACP conversation with a pure-text response (no tool calls) hangs forever in ``running`` status after the ACP server's response arrives. The assistant's ``FinishAction`` + ``ObservationEvent`` are never emitted, so the UI sits empty even though the LLM has already replied.

Root cause: ``EventService._setup_stats_streaming`` (added in #3308) installs a ``stats_callback`` that re-acquires the conversation state lock:

```python
# openhands-agent-server/openhands/agent_server/event_service.py:522-536
def _setup_stats_streaming(self, agent: AgentBase) -> None:
    def stats_callback() -> None:
        if not self._conversation:
            return
        state = self._conversation._state
        with state:                                                # ← hangs here
            event = ConversationStateUpdateEvent(key="stats", value=state.stats)
        self._emit_event_from_thread(event)
    for llm in agent.get_all_llms():
        llm.telemetry.set_stats_update_callback(stats_callback)
```

That callback is invoked **synchronously** from inside ``ACPAgent._record_usage`` (and from ``Telemetry.on_response`` on the regular Agent path), both of which run while ``LocalConversation.run()`` is already holding ``state``'s ``FIFOLock`` via ``with self._state:``. ``FIFOLock`` documents itself as reentrant, so on paper this should be a no-op re-acquire — in practice it deadlocks silently on the ACP path.

## Reproduction

Against any short text prompt to a Claude Code ACP conversation:

```python
agent = ACPAgent(acp_command=["npx", "-y", "@agentclientprotocol/claude-agent-acp"])
conversation = Conversation(agent=agent, workspace="/tmp")
conversation.send_message("hi reply with just the word ok")
conversation.run()        # hangs forever
```

Bisected by adding ``logger.info`` lines through ``ACPAgent.step()`` and ``_record_usage``. The last log to fire is ``"invoking stats_update_callback"``; the corresponding ``"returned"`` log never lands, and no further events get emitted. Adding logs inside ``stats_callback`` itself confirmed the freeze is on the ``with state:`` re-entry, not on ``state.stats`` access or on ``_emit_event_from_thread``.

Observable symptoms from a black-box client (frontend or curl):
- ``POST /api/conversations`` succeeds.
- The agent-server logs ``"Sending ACP prompt"`` then ``"ACP prompt returned in 3.4s"``.
- The conversation's on-disk event log contains exactly 5 events: ``MessageEvent (user)``, ``SystemPromptEvent``, and 3 × ``ConversationStateUpdateEvent``. No ``ActionEvent`` / ``ObservationEvent`` / assistant ``MessageEvent``.
- ``GET /api/conversations?ids=…`` returns ``execution_status: "running"`` indefinitely.

Conversations whose turns include tool calls *appear* to work because ``ACPToolCallEvent``s are emitted live from the ACP bridge (a different code path) before the deadlock is hit — but those conversations are still stuck the moment the turn would otherwise complete; the deadlock just happens to be invisible because the user sees the tool activity in the meantime.

## Why now

``_setup_stats_streaming`` and the matching ``_stats_update_callback()`` invocation in ``ACPAgent._record_usage`` were both introduced in #3308 (commit 64b36cf88, 2026-05-19) as part of "Add named profile support to workspace get_llm". The deadlock has been latent since then; it's been masked on most workflows by tool-using prompts emitting events live.

## Fix

PR #XXXX (filed alongside this issue) drops the ``with state:`` from inside ``stats_callback``. The state lock is **already held** by the caller (``LocalConversation.run`` → ``Agent.step`` / ``ACPAgent._record_usage``), and the only actual persistence happens later in ``_emit_event_from_thread``'s ``locked_on_event`` closure, which takes the lock on its own executor thread. Inner lock acquisition was both redundant and the trigger.

Regression tests in ``tests/agent_server/test_event_service.py::TestStatsCallbackNoDeadlock`` verify the callback returns promptly with and without a caller holding the state lock.

## Notes for triagers

- The reentrancy contract of ``FIFOLock`` is worth a closer look as a follow-up — even if ``stats_callback``'s redundant lock is the trigger, the fact that a documented-reentrant lock deadlocks under same-thread re-entry on the ACP path suggests a real bug in the lock implementation or in how the FIFOLock is being interleaved with the ``__exit__`` deferred-save path. The PR fixes the user-facing symptom (every ACP conversation works again) without touching the lock; a deeper investigation can land separately.
- The same deadlock should theoretically affect the non-ACP path too, since the same ``with state:`` re-entry happens from ``Telemetry.on_response``. Worth verifying whether direct-LLM Agent conversations are hitting it but masked by other emissions, or whether something about the ACP path makes it deterministic.

simonrosenberg:
Fixes #3348.

## Why

The stats-streaming callback added in #3308 acquires the conversation state lock to construct an event:

```python
# openhands-agent-server/openhands/agent_server/event_service.py:522-536
def stats_callback() -> None:
    if not self._conversation:
        return
    state = self._conversation._state
    with state:                                                    # ← hangs
        event = ConversationStateUpdateEvent(key="stats", value=state.stats)
    self._emit_event_from_thread(event)
```

That callback is invoked **synchronously** from ``Telemetry.on_response`` (the regular ``Agent`` path) and from ``ACPAgent._record_usage`` (the ACP path). Both call sites run inside ``LocalConversation.run()``'s ``with self._state:`` block, so the calling thread already holds ``state``'s ``FIFOLock``. ``FIFOLock`` documents itself as reentrant — on paper this should be a no-op re-acquire. In practice, on the ACP code path, it deadlocks silently before ``step()`` can emit the assistant's ``FinishAction`` / ``ObservationEvent``, leaving every short text-only conversation hung in ``running`` status forever.

The deeper question of whether ``FIFOLock`` has a reentrancy bug or whether the ``__exit__``'s deferred-save path interacts badly with the re-entry is worth a separate look. This PR just removes the trigger.

## Summary

- Drop the ``with state:`` inside ``stats_callback`` — it was redundant. The caller already owns the lock, and the only actual persistence path (``_emit_event_from_thread`` → ``locked_on_event``) acquires the lock again on its own executor thread before writing.
- Add ``TestStatsCallbackNoDeadlock`` with two cases:
  1. callback runs cleanly with no lock contention;
  2. callback returns within 2s when the calling thread already owns the state lock (pre-fix this hangs forever — pytest timeout cap set so CI doesn't sit on it).

## How to Test

```bash
uv run pytest tests/agent_server/test_event_service.py::TestStatsCallbackNoDeadlock -v
# 2 passed
```

End-to-end with the ACP path:

```bash
# Pre-fix: hangs.
# Post-fix: completes the turn, conversation transitions to FINISHED, the
#           assistant's FinishAction / ObservationEvent are persisted.
uv run python - <<'PY'
from openhands.sdk.agent import ACPAgent
agent = ACPAgent(acp_command=["npx", "-y", "@agentclientprotocol/claude-agent-acp"])
conv = Conversation(agent=agent, workspace="/tmp")
conv.send_message("reply with just ok")
conv.run()
print("done")
PY
```

## Type

- [x] Bug fix

## Notes

- Filed alongside #3348 which documents the reproduction + bisection trail.
- The same deadlock surface theoretically applies to the non-ACP ``Telemetry.on_response`` path. Worth verifying whether direct-LLM ``Agent`` conversations are also hitting it but masked by other event emissions — flagged in the issue for a follow-up.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:0f95a20-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-0f95a20-python \
  ghcr.io/openhands/agent-server:0f95a20-python
```

```
ghcr.io/openhands/agent-server:0f95a20-golang-amd64
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-golang-amd64
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-golang-amd64
ghcr.io/openhands/agent-server:0f95a20-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:0f95a20-golang-arm64
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-golang-arm64
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-golang-arm64
ghcr.io/openhands/agent-server:0f95a20-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:0f95a20-java-amd64
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-java-amd64
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-java-amd64
ghcr.io/openhands/agent-server:0f95a20-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:0f95a20-java-arm64
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-java-arm64
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-java-arm64
ghcr.io/openhands/agent-server:0f95a20-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:0f95a20-python-amd64
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-python-amd64
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-python-amd64
ghcr.io/openhands/agent-server:0f95a20-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:0f95a20-python-arm64
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-python-arm64
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-python-arm64
ghcr.io/openhands/agent-server:0f95a20-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:0f95a20-golang
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-golang
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-golang
ghcr.io/openhands/agent-server:0f95a20-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:0f95a20-java
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-java
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-java
ghcr.io/openhands/agent-server:0f95a20-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:0f95a20-python
ghcr.io/openhands/agent-server:0f95a20c792e9b402b9a531e63a90cbf024812d0-python
ghcr.io/openhands/agent-server:fix-acp-stats-callback-deadlock-python
ghcr.io/openhands/agent-server:0f95a20-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `0f95a20-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `0f95a20-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -523,13 +523,29 @@ class EventService:
         """Configure stats update callbacks to stream stats changes via events."""
 
         def stats_callback() -> None:
-            """Callback to emit stats updates."""
+            """Callback to emit stats updates.
+
+            Invoked synchronously by ``Telemetry.on_response`` (regular
+            Agent path) and ``ACPAgent._record_usage`` (ACP path) — both
+            run inside ``LocalConversation.run()``'s ``with self._state:``
+            block, so the caller already owns the conversation state lock.
+
+            DO NOT re-acquire the state lock here (``with state:``). It
+            looks safe — ``FIFOLock`` documents itself as reentrant — but
+            on the ACP code path it deadlocks (silently) before the rest
+            of ``step()`` can emit the assistant's FinishAction +
+            ObservationEvent, leaving every conversation hung in
+            ``running`` status forever. ``_emit_event_from_thread`` below
+            already acquires the lock on the executor thread before
+            persisting the event; that's the only place serialization
+            needs the lock anyway.
+            """
             # Publish only the stats field to avoid sending entire state
             if not self._conversation:
                 return
-            state = self._conversation._state
-            with state:
-                event = ConversationStateUpdateEvent(key="stats", value=state.stats)
+            event = ConversationStateUpdateEvent(
+                key="stats", value=self._conversation._state.stats
+            )
             self._emit_event_from_thread(event)
 
         for llm in agent.get_all_llms():
--- a/tests/agent_server/test_event_service.py
+++ b/tests/agent_server/test_event_service.py
@@ -2311,3 +2311,160 @@ async def test_close_blocks_until_executor_thread_finishes(
     )
 
     monkeypatch.undo()
+
+
+class TestStatsCallbackNoDeadlock:
+    """Regression: stats_callback must not re-acquire the state lock.
+
+    ``Telemetry._stats_update_callback`` is invoked synchronously from
+    inside the LLM completion / ACP turn pipeline while another thread
+    (``LocalConversation.run()``) holds the conversation state's
+    ``FIFOLock`` via ``with self._state:``.
+
+    Empirically the deadlock is **cross-thread**: the FIFOLock's
+    same-thread reentry works fine (verified in
+    ``test_same_thread_reentry_works_on_fifolock``), but when the
+    callback fires on a different thread than the lock owner, the
+    extra ``with state:`` inside the callback waits forever.  That is
+    what hung every short-text ACP conversation before this fix.
+
+    These tests pin the contract: the callback returns promptly and
+    the stats event is queued for emission regardless of which thread
+    owns the lock.
+    """
+
+    def _make_service_with_callback(self):
+        stored = StoredConversation(
+            id=uuid4(),
+            agent=Agent(
+                llm=LLM(model="gpt-4o", usage_id="test-stats"),
+                tools=[],
+            ),
+            workspace=LocalWorkspace(working_dir="workspace/project"),
+            confirmation_policy=NeverConfirm(),
+            initial_message=None,
+            metrics=None,
+            created_at=datetime(2025, 1, 1, 12, 0, 0, tzinfo=UTC),
+            updated_at=datetime(2025, 1, 1, 12, 30, 0, tzinfo=UTC),
+        )
+        service = EventService(
+            stored=stored,
+            conversations_dir=Path("test_conversation_dir"),
+        )
+        # A real FIFOLock on a Mock-ish state so the callback contends on
+        # the actual production lock primitive, but we don't have to spin
+        # up the LocalConversation event loop / persistence stack to
+        # exercise the deadlock.
+        state = MagicMock()
+        state._lock = FIFOLock()
+        state.__enter__ = MagicMock(side_effect=lambda: state._lock.acquire())
+        state.__exit__ = MagicMock(side_effect=lambda *a: state._lock.release())
+        state.stats = MagicMock(name="stats")
+
+        conversation = MagicMock()
+        conversation._state = state
+        service._conversation = conversation
+        # Stub the executor-thread emission path so the test stays
+        # synchronous + deterministic.  The under-test behaviour is just
+        # ``stats_callback returns promptly``; the locked_on_event side of
+        # _emit_event_from_thread is covered elsewhere.
+        service._emit_event_from_thread = MagicMock(name="_emit_event_from_thread")
+
+        callbacks: list = []
+        service._setup_stats_streaming(
+            MagicMock(
+                get_all_llms=lambda: [
+                    MagicMock(
+                        telemetry=MagicMock(set_stats_update_callback=callbacks.append)
+                    )
+                ]
+            )
+        )
+        assert len(callbacks) == 1, "stats_callback must be registered exactly once"
+        return service, state, callbacks[0]
+
+    def test_same_thread_reentry_works_on_fifolock(self):
+        """Sanity: FIFOLock's reentrancy contract holds for same-thread re-acquire.
+
+        Documents why the fix is **not** about masking a broken reentrant
+        lock — even with the buggy ``with state:`` re-entry, the same
+        thread can re-acquire FIFOLock without deadlock.  This isolates
+        the deadlock as a cross-thread phenomenon (see the next test).
+        """
+        lock = FIFOLock()
+        finished = threading.Event()
+
+        def run():
+            with lock:
+                with lock:  # same-thread re-entry
+                    pass
+            finished.set()
+
+        threading.Thread(target=run, daemon=True).start()
+        assert finished.wait(timeout=2.0), "FIFOLock should support same-thread reentry"
+
+    @pytest.mark.timeout(10)
+    def test_returns_promptly_when_another_thread_holds_state_lock(self):
+        """The deadlock case: another thread owns the lock when the callback fires.
+
+        Mirrors production: ``LocalConversation.run()`` on thread A holds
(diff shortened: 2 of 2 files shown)


simonrosenberg:
Addressed the [review feedback on regression coverage](https://github.com/OpenHands/software-agent-sdk/pull/3349#pullrequestreview-…) — pushed `0f95a20c7`.

You were right that my original tests exercised the wrong failure mode. Empirically:

- ``FIFOLock``'s **same-thread re-entry works fine** — the buggy ``with state:`` inside ``stats_callback`` does not hang when both lock owner and callback live on the same thread.
- The actual deadlock is **cross-thread**: thread A (``LocalConversation.run``) owns the lock via ``with self._state:``, the stats callback fires on a different thread (executor / portal / bridge), and ``with state:`` inside the callback blocks forever because FIFOLock's reentrancy keys off ``threading.get_ident()`` and thread B's ident isn't the owner.

The new ``TestStatsCallbackNoDeadlock`` now has three cases:

1. ``test_same_thread_reentry_works_on_fifolock`` — sanity check that documents why we **don't** need to "fix" FIFOLock; same-thread re-entry is fine, the bug isn't in the lock.
2. ``test_returns_promptly_when_another_thread_holds_state_lock`` — **the real regression case.** Thread A holds the lock; thread B invokes the callback. The test uses a ``pytest.mark.timeout(10)`` cap so CI doesn't sit on the hang. **Verified locally that this case FAILS on ``HEAD~1``** (the pre-fix version of ``event_service.py``) — got ``stats_callback hung — thread A still holds the FIFOLock`` — and PASSES on this PR's fix.
3. ``test_returns_promptly_with_no_lock_contention`` — baseline that the no-contention path still works.

I left the no-contention test in to give the failure-mode test something to contrast against in the test class; happy to drop it if you'd rather keep this lean.

malhotra5 (approved):
LGTM

Issue #3341, opened 2026-05-21
[Tracker] Async LLM support (#3284): correctness bugs + maintainability issues


VascoSch92:
## Summary

Tracker for follow-up work on the async LLM path introduced in #3284 (`feat: add async LLM completion support across the SDK`, merged). The agent server now drives real `LocalConversation`s through `arun()` by default (`event_service.py:285-294`), so these affect production server runs, not just opt-in async callers.

Findings below were cross-checked against merged `main` by two independent reviews; the two High-severity items were reproduced.

## Checklist

**Correctness**
- [x] **B1** — Async error path bypasses fallback strategy, provider-exception mapping, and `telemetry.on_error` (High)
    - #3355
- [x] **B2** — Async retry loses the `temperature→1.0` escalation on `LLMNoResponseError` (High) 
    - #3356
- [x] **B3** — `cancel_token` cleared in `arun()`'s `finally` while interrupted tool threads may still be running (Medium) 
    - #3395 
- [x] **B4** — `RemoteConversation.interrupt()` not overridden → silently degrades to `pause()` (Medium) 
    - #3398 
- [x] **B5** — ACP/custom agents (no `astep` override) run sync `step()` in a worker thread while `arun()` holds the state lock on the event-loop thread (Medium) 
    - #3404 

**Maintainability**
- [x] **Q1** — Dead code: `Tool.acall`, `agenerate_title_with_llm`/`agenerate_title_from_message`, dead `_return_metrics` param
    - #3432 (deprecated. It will be removed in one of the next versions)
    - #3447 
    - #3456
- [x] **Q2** — `Metrics.get_snapshot()` not reused (4 hand-built copies + token-usage aliasing) 
    - #3397 
- [x] **Q3** — Parallel-tool throughput regression (shared default executor vs dedicated pool) 
    - #3419 
- [x] **Q4** — `FIFOLock` held across `await` — undocumented "mutators must run off the event-loop thread" invariant 
    - #3400
- [x] **Q5** — ~90% sync/async twin duplication (~600 reclaimable lines); extract shared helpers
    1.  #3448 
    2. LocalConversation.run / arun — local_conversation.py:908 & 1059
[...]

PR #143, opened 2025-09-06
Fix prompt caching and improve input/output metrics visualization

xingyaoww:
Fix prompt caching https://github.com/All-Hands-AI/agent-sdk/issues/137

Improve input/output metrics visualization
image

PR #3432, opened 2026-05-29
feat(sdk): deprecate no-op _return_metrics LLM param (Q1 of #3341)

VascoSch92:
## Summary

Part of #3341 (Q1 — dead code). The `_return_metrics` parameter on `LLM.completion` / `acompletion` / `responses` / `aresponses` (and `return_metrics` on `RouterLLM.completion`) has **no effect**: metrics are always returned via `LLMResponse.metrics`. It dates back to #143 and was never read in any conditional in the repo's history; #3284 then propagated it into the async twins.

Originally this PR removed the parameter outright, but that breaks the SDK's **public API removal policy** (the `check_sdk_api_breakage.py` checker compares against the released PyPI baseline, which never deprecated it). Per maintainer discussion, this now follows the **normal deprecation runway** instead: keep the parameter as an ignored no-op, warn when it is passed, and remove it after the required 5 minor releases.

## Changes

- **Deprecate, don't remove.** `_return_metrics` / `return_metrics` stay in the public signatures (defaults and order unchanged → no structural API break). Each public method emits a `DeprecationWarning` via `warn_deprecated(...)` **only when the parameter is truthy**, with `deprecated_in="1.24.0"`, `removed_in="1.29.0"` (≥ 5 minor releases). Docstrings now mark the parameter deprecated.
- **Stop threading the no-op internally.** The fallback lambdas in `llm.py` and the `RouterLLM` delegation no longer forward the parameter (it's a no-op and avoids a double warning on the fallback path); trailing args are passed by keyword so positions stay correct.
- **Test:** `tests/sdk/llm/test_return_metrics_deprecation.py` asserts the warning fires when the param is passed and stays silent otherwise.

The actual deletion will land in a follow-up once the repo reaches `1.29.0`, at which point `check_deprecations.py` enforces the cleanup. Q1's `_return_metrics` item is therefore **deprecated here, removed later**.

## Verification

- `check_deprecations.py` recognizes the new entries (version literals are inlined so the static-AST checker can read them) and does not flag them against 1.24.0.
- Public signatures are byte-identical to the PyPI baseline → no structural break for `check_sdk_api_breakage.py`.
- pre-commit (ruff format/lint, pycodestyle, pyright, import rules) and `tests/sdk/llm/` (799 passed) are green.

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -110,9 +110,21 @@ from openhands.sdk.llm.utils.model_features import get_features
 from openhands.sdk.llm.utils.retry_mixin import RetryMixin
 from openhands.sdk.llm.utils.telemetry import Telemetry
 from openhands.sdk.logger import ENV_LOG_DIR, get_logger
+from openhands.sdk.utils.deprecation import warn_deprecated
 
 
 logger = get_logger(__name__)
+
+# Shared message for the no-op ``_return_metrics`` deprecation (Q1 of #3341).
+# Metrics are always returned via ``LLMResponse.metrics``; the parameter has no
+# effect and is scheduled for removal after the standard 5-minor-release runway.
+# NOTE: ``deprecated_in`` / ``removed_in`` must be passed as string literals at
+# each call site — ``check_deprecations.py`` reads them via static AST analysis
+# and cannot resolve module-level constants.
+_RETURN_METRICS_DETAILS: Final[str] = (
+    "The _return_metrics parameter has no effect; metrics are always available "
+    "via LLMResponse.metrics. Stop passing it."
+)
 
 __all__ = ["LLM"]
 
@@ -1099,7 +1111,9 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         Args:
             messages: List of conversation messages.
             tools: Optional list of tools available to the model.
-            _return_metrics: Whether to return usage metrics.
+            _return_metrics: Deprecated and ignored; metrics are always returned
+                via ``LLMResponse.metrics``. Scheduled for removal in
+                ``1.29.0``.
             add_security_risk_prediction: Add security_risk field to tool schemas.
             on_token: Optional callback for streaming tokens.
             **kwargs: Additional arguments passed to the LLM API.
@@ -1123,6 +1137,13 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
             print(response.content)
             ```
         """
+        if _return_metrics:
+            warn_deprecated(
+                "LLM.completion(_return_metrics=...)",
+                deprecated_in="1.24.0",
+                removed_in="1.29.0",
+                details=_RETURN_METRICS_DETAILS,
+            )
         enable_streaming = bool(kwargs.get("stream", False)) or self.stream
         if enable_streaming:
             if on_token is None:
@@ -1167,9 +1188,8 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
                 lambda fb: fb.completion(
                     messages,
                     tools,
-                    _return_metrics,
-                    add_security_risk_prediction,
-                    on_token,
+                    add_security_risk_prediction=add_security_risk_prediction,
+                    on_token=on_token,
                 ),
             )
 
@@ -1190,6 +1210,13 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         Uses ``litellm.acompletion`` under the hood, freeing the event loop
         while waiting for the LLM provider response.
         """
+        if _return_metrics:
+            warn_deprecated(
+                "LLM.acompletion(_return_metrics=...)",
+                deprecated_in="1.24.0",
+                removed_in="1.29.0",
+                details=_RETURN_METRICS_DETAILS,
+            )
         enable_streaming = bool(kwargs.get("stream", False)) or self.stream
         if enable_streaming:
             if on_token is None:
@@ -1237,9 +1264,8 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
                 lambda fb: fb.completion(
                     messages,
                     tools,
-                    _return_metrics,
-                    add_security_risk_prediction,
-                    _fb_token,
+                    add_security_risk_prediction=add_security_risk_prediction,
+                    on_token=_fb_token,
                 ),
             )
 
@@ -1266,7 +1292,8 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
             tools: Optional list of tools available to the model
             include: Optional list of fields to include in response
             store: Whether to store the conversation
-            _return_metrics: Whether to return usage metrics
+            _return_metrics: Deprecated and ignored; metrics are always returned
+                via ``LLMResponse.metrics``. Scheduled for removal in ``1.29.0``.
             add_security_risk_prediction: Add security_risk field to tool schemas
             on_token: Optional callback for streaming deltas
             **kwargs: Additional arguments passed to the API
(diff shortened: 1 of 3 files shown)


VascoSch92:
@enyst it is an API break, but the args was really never used.

enyst:
Could we see what’s up with the agent-server tests?

I would suggest that maybe we can document the `_return_metrics` removal in release notes next time, as a technically an API break. A simple notice, I think, should do in this case.

enyst (approved):
Love the small cleanup! Left a little top level comment - actually I think I’ll add manually the label for release notes 🤔

enyst:
@openhands Read the PR and the top level comments. WDYT is an easy way to let CI pass on the API breakage in this case? It seems extremely unlikely it would cause problems downstream TBH.

Give me thoughts and alternatives. Main thing is, once we merge this PR, NEWER PRs shouldn’t choke on this CI check. Would they?

Respond directly on PR.

VascoSch92:
I believe the test is flaky... I re-run it... Perhaps I should relax the condition to pass

enyst:
@enyst I read through the diff, top-level comments, and the API breakage workflow/checker. My read:

- The `release-note-required` label is the right courtesy signal, but it will not make this check pass. That label path is only wired to public `Field(default=...)` behavioral changes; structural API breakages still fail.
- A version bump alone also would not be enough here. It can satisfy the SemVer part, but the checker separately enforces the deprecation-runway policy, and the baseline PyPI release did not deprecate `_return_metrics` / `return_metrics` for 5 minor releases. Adding deprecation metadata in this same PR would not help because the checker reads the prior deprecation state from the baseline package.
- If this merges as-is with the checker unchanged, then yes, newer PRs will keep choking on this same Python API check until either (a) a PyPI baseline exists that already contains the removal, or (b) the checker is taught that this specific break is accepted. For structural breakages the PR workflow compares current HEAD against the PyPI baseline, not just against the PR base branch. The base-ref differential logic is currently only for `Field(default=...)` reporting/labeling.

Easy options I see:

1. **Conservative / policy-pure:** keep the parameter in the public signatures as an ignored compatibility shim, optionally warn/deprecate it, and remove it after the normal runway. This is the simplest way to make CI green without changing the checker, but it means Q1 does not fully remove the dead API yet.
2. **Accept this one intentional break:** add a very narrow allowlist in `.github/scripts/check_sdk_api_breakage.py` for exactly these Griffe breakages: `LLM.{completion,acompletion,responses,aresponses}` removing `_return_metrics`, `RouterLLM.completion` removing `return_metrics`, plus the induced positional-move findings for `add_security_risk_prediction` / `on_token` on those same methods. This keeps the checker strict for everything else and prevents future PRs from rediscovering this accepted cleanup.
3. **Broader checker change:** make PRs fail only on structural breakages introduced relative to the base branch, while still reporting PyPI-vs-HEAD. I would be cautious with this for the SDK because it can let accepted/unreleased breaks accumulate and surprise the release; if we do it, push/release checks should stay strict.

My recommendation would be #2 if maintainers agree this is an exceptional dead/no-op parameter where normal deprecation is disproportionate. Otherwise #1 is the safest policy-compliant route. I would not rely on the release-note label or a minor bump by themselves.

_This comment was generated by an AI agent (OpenHands) on behalf of the requester._

enyst:
@VascoSch92 Actually, maybe we could let it go through the regular deprecation process? I kinda don’t like any of my agent’s solutions in this comment https://github.com/OpenHands/software-agent-sdk/pull/3432#issuecomment-4575636054 , but normal deprecation is the cleanest. WDYT?

VascoSch92:
> @VascoSch92 Actually, maybe we could let it go through the regular deprecation process? I kinda don’t like any of my agent’s solutions in this comment [#3432 (comment)](https://github.com/OpenHands/software-agent-sdk/pull/3432#issuecomment-4575636054) , but normal deprecation is the cleanest. WDYT?

@enyst LGTM... what do you mean for normal deprecation process? Just deprecation and deleting in a couple of releases?

enyst:
> > @VascoSch92 Actually, maybe we could let it go through the regular deprecation process? I kinda don’t like any of my agent’s solutions in this comment [#3432 (comment)](https://github.com/OpenHands/software-agent-sdk/pull/3432#issuecomment-4575636054) , but normal deprecation is the cleanest. WDYT?
> 
> @enyst LGTM... what do you mean for normal deprecation process? Just deprecation and deleting in a couple of releases?

Yes, 5 minor releases: [Python API deprecation policy](https://github.com/OpenHands/software-agent-sdk/blob/7c872a3306a95fc1e538de0b6a5048634c807434/AGENTS.md?plain=1#L162)

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -110,10 +110,22 @@
...
 
+# Shared message for the no-op ``_return_metrics`` deprecation (Q1 of #3341).
+# Metrics are always returned via ``LLMResponse.metrics``; the parameter has no
+# effect and is scheduled for removal after the standard 5-minor-release runway.
+# NOTE: ``deprecated_in`` / ``removed_in`` must be passed as string literals at
+# each call site — ``check_deprecations.py`` reads them via static AST analysis
+# and cannot resolve module-level constants.
+_RETURN_METRICS_DETAILS: Final[str] = (
+    "The _return_metrics parameter has no effect; metrics are always available "
+    "via LLMResponse.metrics. Stop passing it."
```
VascoSch92, line 126: @enyst this is a little passive-aggressive :-)

enyst, line 126:
```suggestion
    "via LLMResponse.metrics."
```
😅

enyst:
Apparently llm.completion() is not part of the public API 🤔

@openhands read all top level comments. /codereview-roasted on this PR, does it do deprecation correctly considering the policy and the griffe CI check. Post your feedback as a review with gh api and event. If you are happy with it, feel free to approve and merge.

enyst (approved):
Taste Rating: 🟢 **Good taste**

I read the current diff and the top-level PR discussion. This now does the deprecation the policy wants:

- The public signatures keep `_return_metrics` / `return_metrics` in place with the old default/order, so the current PR has no structural API break for Griffe to report.
- The deprecation metadata uses `warn_deprecated(...)` with literal `deprecated_in="1.24.0"` and `removed_in="1.29.0"`, which gives the required 5-minor-release runway and is visible to `check_deprecations.py`.
- I ran `uv run pytest tests/sdk/llm/test_return_metrics_deprecation.py -q`, `uv run python .github/scripts/check_deprecations.py`, and `uv run python .github/scripts/check_sdk_api_breakage.py`; all passed.
- On the Griffe question: `LLM.completion()` is part of the checked public surface because `LLM` is exported from `openhands.sdk.__all__`. I also simulated deleting the parameter locally; Griffe reports `LLM.completion(_return_metrics)::Parameter was removed` plus the same for `acompletion`, `responses`, and `aresponses` (and positional-shift warnings). So the API check passing now is because the shim keeps the signature, not because these methods are invisible.
- `RouterLLM` is only exported via `openhands.sdk.llm.__all__`, not the root `openhands.sdk.__all__` surface that this checker currently centers on, but deprecating `RouterLLM.completion(return_metrics=...)` anyway is the right compatibility move.

Small nuance, not blocking: the warning fires only for truthy values, so an explicit `_return_metrics=False` caller will not be warned before the eventual removal. Given the historical default is `False`, distinguishing explicit false from omitted without changing the signature would be awkward; I don't think this is worth complicating the shim for.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 LOW — compatibility shim plus warning only; no behavior change to metrics, LLM transport, or agent decision logic. Required checks are green; there is a non-required Agent Server manifest failure, but the required Agent Server/test jobs passed.

**VERDICT:**
✅ **Worth merging**. This is the policy-pure path: deprecate now, let `check_deprecations.py` force cleanup at 1.29.0, and avoid teaching the API checker a one-off exception.

_This review was created by an AI agent (OpenHands) on behalf of the requester._

Issue #3438, opened 2026-05-29
[Bug] Injected system-prompt datetime is timezone-ambiguous and stale (naive local time, frozen at AgentContext construction)


VascoSch92:
## Summary

The `current_datetime` that `AgentContext` injects into the system prompt is (1) captured in **naive local time** with no timezone offset, and (2) **frozen at `AgentContext` construction time**, so it can be stale before a conversation even starts and is *always* stale in long-running conversations. The agent is therefore given an ambiguous and potentially wrong "current" time.

## Where

- **Field definition:** `openhands-sdk/openhands/sdk/context/agent_context.py:130-140`
  ```python
  current_datetime: datetime | str | None = Field(
      default_factory=datetime.now,
      ...
  )
  ```
- **Formatting:** `agent_context.py:234-246` — `get_formatted_datetime()` returns `self.current_datetime.isoformat()`.
- **Injection:** rendered into the system prompt via `context/prompts/templates/system_message_suffix.j2:1-5` under `<CURRENT_DATETIME>`.
- **Built once:** the `SystemPromptEvent` carrying this value is created a single time in `agent/agent.py:450-462` (`init_state`), with an explicit early-return guard on resume (`agent/agent.py:404-423`). It is never regenerated for later turns.

## Problems

1. **Timezone ambiguity.** `datetime.now()` (no argument) returns naive *local* time, so `.isoformat()` emits e.g. `2026-05-29T14:30:00.123456` with **no UTC offset**. The agent cannot tell what timezone this is. The SDK already has a tz-aware helper, `utc_now()` (`openhands-sdk/openhands/sdk/utils/datetime.py`), which this field does not use.

2. **Stale across reused `AgentContext`.** Because `default_factory=datetime.now` runs when the `AgentContext` object is constructed, a server that builds one `AgentContext` and reuses it across many conversations stamps every conversation with the timestamp from whenever that object was created — potentially hours or days old.

3. **Stale within a long conversation.** Even with a perfect value at init, the `SystemPromptEvent` is built once per conversation and never refreshed, so a multi-hour conversation keeps showing its start time. This is fundamental: anything baked into the system prompt is a snapshot, not a live clock.

## Screenshot

The current datetime at my location is: 17:55 of the 29th of May 2026

<img width="769" height="487" alt="Image" src="https://github.com/user-attachments/assets/b51fb2f7-9cae-4c73-9ee4-743d34db649c" />

## Impact

[...]

PR #3446, opened 2026-05-31
fix(sdk): make default current_datetime timezone-aware

StressTestor:

---

## Why

`AgentContext.current_datetime` injects a "current time" into the system prompt (the `<CURRENT_DATETIME>` block). Its default was `Field(default_factory=datetime.now)`, and `datetime.now()` with no argument returns **naive local time**, so `get_formatted_datetime()` emitted e.g. `2026-05-31T02:29:57.162025` with **no UTC offset**. The agent is handed an ambiguous time it can't resolve to an actual instant. (See #3438.)

## Summary

- Default `current_datetime` is now timezone-aware local time (`datetime.now().astimezone()`), so the value injected into the system prompt carries a UTC offset (e.g. `2026-05-31T02:37:24-06:00`) instead of a naive timestamp.
- User-supplied `current_datetime` values (datetime or string) are untouched — `get_formatted_datetime()` still returns them as-is (existing contract preserved).
- Updated `test_agent_context_default_datetime` to assert the default is timezone-aware (regression guard).

## Issue Number

Addresses #3438 (the timezone-ambiguity half — not using "Fixes" on purpose; see Notes).

## How to Test

Reproduce on `main` (before the fix), then confirm after:

```bash
uv run python -c "from openhands.sdk.context.agent_context import AgentContext as A; \
c=A(); print(c.current_datetime.tzinfo, '|', c.get_formatted_datetime())"
```

Rendered system-prompt block, before vs after:

```
# before (main):  naive, no offset
<CURRENT_DATETIME>
The current date and time is: 2026-05-31T02:29:57.162025
</CURRENT_DATETIME>

# after (this PR): timezone-aware, has offset
<CURRENT_DATETIME>
The current date and time is: 2026-05-31T02:37:24.994238-06:00
</CURRENT_DATETIME>
```

A user-supplied naive datetime still renders unchanged (`2024-03-15T14:30:00`).

```bash
uv run pytest tests/sdk/context/                      # 276 passed
  openhands-sdk/openhands/sdk/context/agent_context.py \
  tests/sdk/context/test_agent_context.py             # ruff + pyright + deps: pass
```

## Type

- [x] Bug fix

## Notes

- **Scope:** #3438 also reports the datetime being *frozen* at `AgentContext` construction (stale in long conversations). That's intentionally out of scope here — regenerating it per turn would change the system prompt every turn and defeat prompt caching, so whether/how to refresh it is a design tradeoff for maintainers. This PR resolves only the timezone ambiguity, hence "Addresses" rather than "Fixes" so the stale aspect stays tracked.
- Chose `datetime.now().astimezone()` (local time + offset, minimal change to the displayed value) over forcing UTC; happy to switch to `datetime.now(timezone.utc)` if you'd prefer canonical UTC.

### DIFF
--- a/tests/sdk/context/test_agent_context.py
+++ b/tests/sdk/context/test_agent_context.py
@@ -988,6 +988,13 @@ templates.",
         # local time) so the datetime injected into the system prompt is
         # unambiguous.
         assert context.current_datetime.tzinfo is not None
+        # The bug surfaced in the rendered value injected into the prompt, not
+        # just the field: get_formatted_datetime() must carry a UTC offset.
+        formatted = context.get_formatted_datetime()
+        assert formatted is not None
+        assert "+" in formatted or "-" in formatted.split("T", 1)[1], (
+            f"formatted datetime should carry a UTC offset, got {formatted!r}"
+        )
         # Verify it's approximately the current time (within 1 second)
         assert before <= context.current_datetime <= after + timedelta(seconds=1)
 
--- a/tests/sdk/context/test_agent_context.py
+++ b/tests/sdk/context/test_agent_context.py
@@ -974,16 +974,20 @@ templates.",
         assert result is None
 
     def test_agent_context_default_datetime(self):
-        """Test that AgentContext defaults to current datetime."""
+        """Test that AgentContext defaults to the current timezone-aware datetime."""
         from datetime import datetime, timedelta
 
-        before = datetime.now()
+        before = datetime.now().astimezone()
         context = AgentContext()
-        after = datetime.now()
+        after = datetime.now().astimezone()
 
         # Verify current_datetime is set and is a datetime object
         assert context.current_datetime is not None
         assert isinstance(context.current_datetime, datetime)
+        # Regression for #3438: the default must be timezone-aware (not naive
+        # local time) so the datetime injected into the system prompt is
+        # unambiguous.
+        assert context.current_datetime.tzinfo is not None
         # Verify it's approximately the current time (within 1 second)
         assert before <= context.current_datetime <= after + timedelta(seconds=1)
 
--- a/openhands-sdk/openhands/sdk/context/agent_context.py
+++ b/openhands-sdk/openhands/sdk/context/agent_context.py
@@ -128,13 +128,16 @@ class AgentContext(BaseModel):
         json_schema_extra={"acp_compatible": True},
     )
     current_datetime: datetime | str | None = Field(
-        default_factory=datetime.now,
+        # Timezone-aware local "now" so the value injected into the system prompt
+        # carries a UTC offset instead of an ambiguous naive local time (#3438).
+        default_factory=lambda: datetime.now().astimezone(),
         description=(
             "Current date and time information to provide to the agent. "
             "Can be a datetime object (which will be formatted as ISO 8601) "
             "or a pre-formatted string. When provided, this information is "
             "included in the system prompt to give the agent awareness of "
-            "the current time context. Defaults to the current datetime."
+            "the current time context. Defaults to the current "
+            "(timezone-aware) datetime."
         ),
         json_schema_extra={"acp_compatible": True},
     )


VascoSch92:
Hey @StressTestor 

Thanks for the PR.

I will review it asap.

tests/sdk/context/test_agent_context.py
```diff
@@ -974,16 +974,20 @@ def test_get_formatted_datetime_with_none(self):
         assert result is None
 
     def test_agent_context_default_datetime(self):
-        """Test that AgentContext defaults to current datetime."""
+        """Test that AgentContext defaults to the current timezone-aware datetime."""
```
VascoSch92, line 977: no test asserts the formatted output carries an offset. 

The actual bug was in what `get_formatted_datetime()` emits, but the test only checks the field's `tzinfo`. 

Could you add:

```
  assert "+" in result or "-" in result.split("T")[1]  # offset present
```

StressTestor: Good point — the bug was in the rendered output, not just the field. Added an assertion that `get_formatted_datetime()` carries a UTC offset (essentially your snippet, with a failure message), reached and passing alongside the `tzinfo` check. Done in the latest commit.

VascoSch92:
Hey @StressTestor 

just out of curiosity: why are you answering the comments but @shilei54220 pushing the code? :-) 

It is not a problem, but I was a little confused when I saw that.

VascoSch92:
Hey @enyst 

Sorry for always picking on you when there is an API breakage.

This is a partial fix for the issue where the datetime we show to the LLM isn't always timezone-compliant. A complete fix will be more complicated, but this is a good starting point.

That said, it does introduce an API change, though the change is minimal. What do you think? How should we handle this?

StressTestor:
good catch, and sorry for the confusion. that's all me, the work here is mine. my git was committing with a noreply email that had a wrong numeric user ID baked into it, and that ID happens to belong to someone else's dormant account (shilei54220), so GitHub attributed the commits to them instead of me. i've fixed the author identity and force-pushed the corrected commits, so they should show up under my account now. thanks for flagging it, genuinely.

VascoSch92:
@StressTestor No problem.

I just want to wait for Enyst's advice on these API changes, but otherwise LGTM.

Could you please make sure to rebase your branch onto the correct commit?

StressTestor:
done, rebased onto latest main, so it's two clean commits now (the merge commit is gone) and it's back to MERGEABLE. both commits show under my account now too. totally fine waiting on @enyst's call about the API change, and i'm happy to rework the approach if you'd both rather handle the timezone offset differently. thanks Vasco.

VascoSch92:
@enyst could you review that? If you like merge it please :)

enyst:
@openhands I just restarted CI, please babysit it until you see the two workflows on API changes, Griffe for python and then REST API. See if any of those two fails, and tell us WDYT: why does it fail, is it a real breakage or is the workflow too tough and could be more lenient? Give alternatives with your thoughts. Consider it carefully. Mention me in your response, respond as a comment on github directly.

enyst:
@enyst I watched the two API-change checks from the restarted CI run:

- **REST API breakage checks / REST API (OpenAPI): passed** — no OpenAPI/REST contract diff was detected.
- **Python API breakage checks / Python API: failed** — Griffe reports an `ATTRIBUTE_CHANGED_VALUE` for `AgentContext.current_datetime`:
  - old: `Field(default_factory=datetime.now, ...)`
  - new: `Field(default_factory=lambda: datetime.now().astimezone(), ...)`

My read: this is a **real observable behavior change**, but I would not treat it as the same class of breakage as removing/renaming public API. `AgentContext()` now produces an aware datetime instead of a naive one, so code that compares the default value to naive datetimes, asserts `tzinfo is None`, or introspects `model_fields["current_datetime"].default_factory` could notice or even break. On the other hand, the accepted type annotation and user-supplied values are unchanged, and the change fixes the ambiguity reported in #3438 rather than removing a capability.

I think the workflow is a bit too tough here in one specific way: it already has a softer path for public `Field(default=...)` changes (report them as behavioral/default changes and require release-note visibility), but it does **not** apply the same treatment to `Field(default_factory=...)`. Because this PR changes `default_factory`, it falls through to Griffe's generic `ATTRIBUTE_CHANGED_VALUE` and becomes a hard SemVer/minor-version failure.

Alternatives I see:

1. **Preferred:** teach the Python API checker to classify `Field(default_factory=...)`-only changes the same way it classifies `Field(default=...)` changes: not structural breakage, but a behavioral compatibility change that gets reported and labeled `release-note-required`. For this PR, because it is from a fork, the workflow also cannot auto-add that label, so a maintainer may need to add it manually.
2. **Strict SemVer interpretation:** keep the workflow as-is and require a minor version bump for this PR. This is defensible because the default value is part of observable public model behavior, but it feels heavy for a targeted bug fix where the old behavior was ambiguous/incorrect.
3. **Avoid touching the public field default:** rework the fix to compute an aware timestamp later, closer to system-prompt construction/rendering. That may reduce the public `AgentContext` default diff, but it likely makes the implementation more complex and starts overlapping with the separate stale/frozen-time design issue in #3438.
4. **Canonical UTC variant:** switch the factory to `utc_now()` / `datetime.now(timezone.utc)`. This may be a clearer semantic choice, but it would still be a `default_factory` change and would still fail the current Python API check.

So my recommendation is: keep the PR's behavior, call it out in release notes, and make the checker more lenient/consistent for `default_factory`-only Pydantic `Field` changes. I would not block this solely as a structural Python API break, but I would not silently ignore it either because naive-vs-aware can affect consumers.

_This comment was generated by an AI agent (OpenHands) in response to @enyst's request._

enyst:
Merged #3519, which makes the Python API checker treat `Field(default_factory=...)` changes like `Field(default=...)` behavioral/default changes instead of structural breakages. I am merging latest `main` into this PR next and will restart CI.

This comment was generated by an AI agent (OpenHands) on behalf of @enyst.

enyst:
HUMAN: Thank you!

PR #3461, opened 2026-06-02
feat(llm): add minimax-m3 to verified MiniMax models

kapelame:
## Summary

Adds MiniMax's flagship **MiniMax-M3** to the verified model lists, so it's recognized as a verified MiniMax model.

- `VERIFIED_MINIMAX_MODELS` (the `minimax` provider list) — add `minimax-m3`
- `VERIFIED_OPENHANDS_MODELS` — add `minimax-m3`, matching the convention where every provider's verified models also appear in the OpenHands hosted list (GLM, Moonshot, Qwen, DeepSeek all do this)

This follows #2627, which added M2.7 the same way.

## Test

- `tests/sdk/llm/test_model_list.py` → 7 passed
- Verified at runtime: `minimax-m3` resolves in `VERIFIED_MINIMAX_MODELS`, `VERIFIED_OPENHANDS_MODELS`, and `VERIFIED_MODELS["minimax"]`
- `ruff check` clean

> Note: complements the eval-config draft #3460 (which touches `.github/run-eval/`) — different files, no overlap.


juanmichelini:
@kapelame thanks for the PR! we are currently verifying minimax-m3, among other things we need to benchmark it in the OpenHands Index. See https://index.openhands.dev/home
Once that is ready we will add it to the verified list.

juanmichelini:
@kapelame model has been verified!

juanmichelini (approved):
Verified

PR #3544, opened 2026-06-06
fix(agent_context): use utc_now() for current_datetime default

### LINKED ISSUE #3438: [Bug] Injected system-prompt datetime is timezone-ambiguous and stale (naive local time, frozen at AgentContext construction)

Oxygen56:
HUMAN:

Fixed the `current_datetime` default by replacing `datetime.now().astimezone()` with `utc_now()`. Verified by importing the module and checking the default factory output.

AGENT:

## Why

`datetime.now().astimezone()` uses local timezone which is non-deterministic for LLM agents running across different hosts. UTC is the standard for server-side timestamps.

## Summary

- Replace `datetime.now().astimezone()` with `utc_now()` helper in agent_context.py
- UTC output is unambiguous and includes `+00:00` offset via isoformat()

## Issue Number
Fixes #3438

## How to Test

```python
from openhands.sdk.context.agent_context import AgentContext
ctx = AgentContext()
print(ctx.current_datetime)  # Should show UTC time with +00:00
```

### DIFF
--- a/openhands-sdk/openhands/sdk/context/agent_context.py
+++ b/openhands-sdk/openhands/sdk/context/agent_context.py
@@ -31,6 +31,7 @@ from openhands.sdk.skills import (
     to_prompt,
 )
 from openhands.sdk.skills.skill import DEFAULT_MARKETPLACE_PATH
+from openhands.sdk.utils.datetime import utc_now
 from openhands.sdk.utils.pydantic_secrets import (
     serialize_secret,
     validate_secret_dict,
@@ -150,9 +151,15 @@ class AgentContext(BaseModel):
         json_schema_extra={"acp_compatible": True},
     )
     current_datetime: datetime | str | None = Field(
-        # Timezone-aware local "now"; get_formatted_datetime renders it to the
-        # minute for the prompt.
-        default_factory=lambda: datetime.now().astimezone(),
+        # Use utc_now() for consistency with the SDK's own datetime helper.
+        # UTC is unambiguous for LLM agents and includes the timezone offset
+        # in .isoformat() output (e.g. "2026-06-06T12:00:00+00:00").
+        #
+        # Remaining known limitation: the value is frozen at AgentContext
+        # construction time (stale in long-running conversations / reused
+        # contexts). Tracking in #3438 — refreshing the value on each turn
+        # requires a broader lifecycle change.
+        default_factory=utc_now,
         description=(
             "Current date and time information to provide to the agent. "
             "Can be a datetime object (which will be formatted as ISO 8601) "


VascoSch92:
Hey @Oxygen56 
thanks for the PR, but I think this fix was already merged...

VascoSch92:
Hey @Oxygen56 

sorry but I think this is already solved

smolpaws (commented):

Taste rating: good taste

Findings:
- Changed area reviewed: openhands-sdk/openhands/sdk/context/agent_context.py.

Risk assessment: low.

Verdict: worth merging

neubig:
Given @VascoSch92 's comment that this is already solved I think we'll close this one.

PR #3558, opened 2026-06-07
fix(sdk): URL-encode session_api_key in WebSocket query parameter

Hades32:
HUMAN:

My agent-canvas setup had created an API key with a `+` and that broke automations.

AGENT:

The agent-canvas automation preset generates a `main.py` script that uses `RemoteConversation` to connect to the agent server. When the `session_api_key` contains `+` characters, the WebSocket auth fails because the key is concatenated raw into the query string. FastAPI/Starlette decodes `+` as a space in query params, causing a 403 "invalid API key" error. I reproduced this by dispatching a cron automation and observing the failure in the run logs:

WebSocket authentication failed: invalid API key WebSocket /sockets/events/...?session_api_key=1+FYhSREds... 403

The fix URL-encodes the key with `urllib.parse.quote(..., safe='')`, so `+` becomes `%2B` and `/` becomes `%2F`. I verified the diff is correct and the import is already available (`urllib.parse` is in stdlib).

---

## Why

When `RemoteConversation._client_loop()` builds the WebSocket URL, it appends the API key directly into the query string:

```python
ws_url += f"?session_api_key={self.api_key}"
```

If the key contains + or / (common in base64-like encodings), the server-side query parser decodes + as a space. The auth check then compares the mangled key against the configured keys and fails with 403. This breaks automation runs, cloud-mode
conversations, and any other flow that relies on RemoteConversation when the key has URL-sensitive characters.

## Summary

  • Import quote from urllib.parse in remote_conversation.py
  • URL-encode the session_api_key with quote(..., safe='') before appending it to the WebSocket URL

## Issue Number

none

## How to Test

1. Configure an agent server with a session_api_key containing + (e.g. 1+FYhSREds8Q...)
2. Create a prompt-preset automation via the automation backend
3. Dispatch the automation manually (POST /api/automation/v1/{id}/dispatch)
4. Observe the run fails with WebSocket authentication failed: invalid API key before the fix
5. Apply the fix, restart services, and re-dispatch — the run should proceed past WebSocket auth

## Video/Screenshots

none

## Type

* [x] Bug fix
* [ ] Feature
* [ ] Refactor
* [ ] Breaking change
* [ ] Docs / chore

## Notes

* `quote(..., safe='')` encodes `/` as `%2F` as well, which is safer than `quote(...)` default behavior (`safe='/'`).
* No API surface change; this is an internal transport fix.


VascoSch92 (approved):
Hey @Hades32 

thanks for the PR.

Just one thing and then we are good to go: could you add an unit test? Thanks

Hades32:
sure, done

VascoSch92:
Thanks 😃

PR #3590, opened 2026-06-09
test(sdk): pin shell parser node shapes

Fieldnote-Echo:
HUMAN:

This PR is a tests-only baseline for the shell parser migration.

AGENT:

---

## Why

PR-C for #2721 pins the concrete `tree-sitter-bash` grammar shapes that the
next AST migration PRs will depend on. This keeps PR-D helper work grounded in
observed parser output instead of guessed node names, while preserving analyzer
behavior.

## Summary

- Add a parser node-shape fixture table for simple commands, pipelines,
  redirects, assignments, structural traversal wrappers, destructive shell
  patterns, opaque command-name constructs, `bash -c`, and malformed shell.
- Assert `ParseResult.has_error`, root node shape, byte-span coverage, and key
  named child node shapes for each fixture.
- Replace the documentary opaque slice with an assertion that opaque
  command-name fixtures are not plain `word` command names.
- Keep the change tests-only: no analyzer migration, no parser helper API, and
  no production behavior change.

## Issue Number

Refs #2721

## How to Test

I ran the focused parser and analyzer-serialization validation target:

```bash
uv run pytest tests/sdk/security/test_shell_parser.py tests/sdk/security/test_shell_parser_node_shapes.py tests/sdk/security/defense_in_depth/test_serialization.py
```

Result: 77 tests passed.

I also ran the file-scoped pre-commit hooks:

```bash
uv run pre-commit run --files tests/sdk/security/test_shell_parser_node_shapes.py
```

Result: Ruff format, Ruff lint, pycodestyle, pyright, import dependency rules,
and tool subclass registration all passed.

For roadmap fit, review feedback led to pinning missing quoted command-name
variants, both ANSI-C hex and octal forms, structural traversal wrappers
(`&&`, `;`, subshell, and compound statement), and variable-expansion command
names (`$CMD`, `${CMD}`, and `rm${IFS}-rf${IFS}/`).

## Video/Screenshots

Not applicable; this is a parser test baseline with no UI surface.

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [x] Docs / chore

## Notes

This PR intentionally does not migrate `PatternSecurityAnalyzer` or
`PolicyRailSecurityAnalyzer`, add public parser helper APIs, change
`UNKNOWN`/risk semantics, or modify dependencies.

### DIFF
--- a/tests/sdk/security/test_shell_parser_node_shapes.py
+++ b/tests/sdk/security/test_shell_parser_node_shapes.py
@@ -1,0 +1,390 @@
+"""Pin tree-sitter-bash node shapes needed by shell-security migration."""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from dataclasses import dataclass
+
+import pytest
+from tree_sitter import Node
+
+from openhands.sdk.security.shell_parser import parse
+
+
+type ExpectedChild = tuple[str, str]
+
+
+@dataclass(frozen=True)
+class ExpectedNode:
+    node_type: str
+    text: str
+    named_children: tuple[ExpectedChild, ...] | None = None
+
+
+@dataclass(frozen=True)
+class NodeShapeCase:
+    case_id: str
+    command: str
+    has_error: bool
+    root_named_children: tuple[str, ...]
+    expected_nodes: tuple[ExpectedNode, ...]
+    clean_but_opaque: bool = False
+
+
+NODE_SHAPE_CASES: tuple[NodeShapeCase, ...] = (
+    NodeShapeCase(
+        case_id="simple_command",
+        command="echo hello",
+        has_error=False,
+        root_named_children=("command",),
+        expected_nodes=(
+            ExpectedNode(
+                "command",
+                "echo hello",
+                (("command_name", "echo"), ("word", "hello")),
+            ),
+            ExpectedNode("command_name", "echo", (("word", "echo"),)),
+        ),
+    ),
+    NodeShapeCase(
+        case_id="pipeline",
+        command="echo hi | wc -c",
+        has_error=False,
+        root_named_children=("pipeline",),
+        expected_nodes=(
+            ExpectedNode(
+                "pipeline",
+                "echo hi | wc -c",
+                (("command", "echo hi"), ("command", "wc -c")),
+            ),
+            ExpectedNode(
+                "command",
+                "echo hi",
+                (("command_name", "echo"), ("word", "hi")),
+            ),
+            ExpectedNode(
+                "command",
+                "wc -c",
+                (("command_name", "wc"), ("word", "-c")),
+            ),
+        ),
+    ),
+    NodeShapeCase(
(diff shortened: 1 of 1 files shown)


Fieldnote-Echo:
@VascoSch92 ready for review when you have a spare moment, please.

VascoSch92:
A couple of nitpick :-) 

  ### 1. `test_clean_but_opaque_commands_are_parse_successes` is redundant, and `clean_but_opaque` never asserts opacity

  The second test only asserts `parse(case.command).has_error is False`, but `test_shell_parser_node_shapes` already asserts `result.has_error is case.has_error` over all cases and every `clean_but_opaque` case 
  has `has_error=False`. So it adds no coverage.

  More to the point, the flag promises a property nothing checks: "opaque" should mean the command name isn't a plain literal (it's a `concatenation` / `raw_string` / `command_substitution` / `ansi_c_string`) the exact thing that defeats the regex detector. As written the flag just selects cases for a duplicate `has_error` check.

  **Suggestion:** drop the second test + flag, or make it earn its name, e.g. assert that for `clean_but_opaque` cases the `command_name`'s named child type is **not** `word`. That pins the actual bypass property
  the migration depends on without duplicating the first test.

  ### 2. Structural constructs the AST walk must traverse aren't pinned

  If the intent is to pin "the shapes the next AST migration PRs depend on," an analyzer walking the tree to find a destructive `command` has to descend through these (all verified clean parses):

  | command | root named children |
  |---|---|
  | `rm -rf / && echo done` | `('list',)` |
  | `echo a; rm -rf /` | `('command', 'command')` |
  | `( rm -rf / )` | `('subshell',)` |
  | `{ rm -rf /; }` | `('compound_statement',)` |

  The `;` case is the one I'd most want pinned, it's the only construct here that changes the **root child count**, so anything doing `root.named_children[0]` would silently miss the second command. I know the 
  scope note flags these as non-bypasses today (handled HIGH by other detectors), so this may be intentional, just want to confirm the baseline is deliberately bypass-only, since the PR description reads broader.

  ### 3. Missing opaque command-name variant: variable expansion
  
  The table enumerates opaque command names via `concatenation`, `raw_string`, `command_substitution`, and `ansi_c_string`, but not the variable-expansion family, same bypass shape, also evades `\brm\b` 
  (verified):

  - `$CMD -rf /` → `command_name → simple_expansion → variable_name`
  - `${CMD} -rf /` → `command_name → expansion → variable_name`
  - `rm${IFS}-rf${IFS}/` → `command_name → concatenation(word, expansion, word, …)`

  `$CMD`/`${CMD}` is structurally analogous to the `$(echo rm)` case already pinned, and it's absent from both this table and the bypass catalog. Worth deciding whether variable indirection is in scope.

Fieldnote-Echo:
@VascoSch92 remediation commit pushed, thank you for the review!

VascoSch92 (approved):
;)

PR #3609, opened 2026-06-09
Add private shell AST command view helpers

Fieldnote-Echo:
HUMAN:

This PR adds a private shell AST view layer that PR-E can use to migrate security detection onto tree-sitter without changing current detector behavior yet.

AGENT:

---

## Why

PR-D scaffold for #2721 adds a private syntax-view layer before the behavior-changing detector migration. The goal is to give the next PR a typed, byte-offset-safe view over tree-sitter-bash commands, words, and pipelines without changing current analyzer behavior.

## Summary

- Added private `openhands.sdk.security._shell_ast` dataclasses and helpers for shell programs, words, commands, and pipelines.
- Added command traversal, pipeline stage extraction, POSIX command basename, short/long flag, and key-value word helpers.
- Kept detector behavior, public exports, serialization fields, dependencies, and `UNKNOWN` semantics unchanged.

## Issue Number

#2721

## How to Test

I validated the scaffold locally with the following commands:

- `make build`
- `uv run pre-commit run --files openhands-sdk/openhands/sdk/security/_shell_ast.py tests/sdk/security/test_shell_ast.py`
- `uv run pytest tests/sdk/security/test_shell_parser.py tests/sdk/security/test_shell_parser_node_shapes.py tests/sdk/security/test_shell_ast.py`
- `uv run pytest tests/sdk/security/defense_in_depth/test_serialization.py tests/sdk/security/defense_in_depth/test_shell_parser_bypasses.py`

The parser-focused suite passed with 113 tests. The defense-in-depth suite passed with 23 tests and 7 strict xfails, preserving the current bypass-test expectations. Because this PR only adds a private SDK parser-view scaffold and does not change runtime detector behavior or UI behavior, there is no separate manual product flow to exercise in this PR.

## Video/Screenshots

Not applicable. This is a private SDK refactor scaffold with no UI surface.

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This is the PR-D scaffold only. Behavior-changing detector migration, recursive `bash -c` parsing, ANSI-C decoding, command-substitution evaluation, ActionEvent extraction, parse caching, and public helper promotion are intentionally left out of scope.

Two helper boundaries are intentional for PR-E:

- `command_basename()` uses `posixpath.basename()` for simple command-name views. It should not be described later as complete shell basename-command normalization; for example, `posixpath.basename("/bin/rm/")` returns an empty string.
- `split_short_flags()` is only a syntax primitive over non-opaque words. PR-E policy code should decide whether unusual words such as `-rf/` or `-r=f` count as destructive flags.


csmith49 (approved):
Hey @Fieldnote-Echo, this looks great. A comment and a question, neither blocking:
* Do we need this module to be private? It's probably fine if it's not, and AFAIK that's not really a pattern we've used elsewhere in the SDK.
* A few top-level doc-strings on the data classes clarifying which part of the bash command they represent would be helpful.

Fieldnote-Echo:
@csmith49 thanks for the review! I added the docstrings. Regarding the private module, I've been careful to not make any public API changes unless absolutely necessary.

Fieldnote-Echo:
@enyst or @VascoSch92 this should be good to merge now. Thank you all!

enyst:
> Regarding the private module, I've been careful to not make any public API changes unless absolutely necessary.

Just for clarity, the convention we use is:
- public API has compatibility guarantees (no breakage, deprecation runway for 5 releases)
- that is only what is exposed in init, not every ‘public’ method/class/module though

Issue #3678, opened 2026-06-12
[Bug]: file_editor truncates a file to 0 bytes (data loss) when the detected encoding can't encode newly written content


shanemort1982:
### Is there an existing issue for the same bug?

- [x] I have searched existing issues and this is not a duplicate.

### Bug Description

`file_editor` can **destroy a file (truncate it to 0 bytes)** when it edits a file whose
encoding `charset_normalizer` sniffs as a non-UTF-8 single-byte codec (e.g. `windows-1250`/`cp1252`)
and the new content contains a character that codec cannot represent (e.g. `U+2192` →, em dash,
emoji).

Two compounding defects:

1. **`write_file` is not atomic.** It does `open(path, "w", ...)` which truncates the file
   immediately, then `f.write(file_text)`. If the write raises mid-way, the original file is already
   gone. (`openhands/tools/file_editor/editor.py`, `write_file`.)
2. **Writes reuse a sniffed lossy encoding.** `EncodingManager.detect_encoding` trusts
   `charset_normalizer` and only special-cases `ascii -> utf-8`. A mostly-Latin file, or any file
   containing real Windows-125x bytes, is detected as `windows-125x` and that single-byte codec is
   then used for the **write**. Any newly inserted character outside that codec raises
   `UnicodeEncodeError`, which (per defect 1) truncates the file.

This is silent data loss: the editor reports an error, but the file is already empty.

### Expected Behavior

- An edit that cannot be encoded should **fail without modifying the file** (atomic write: temp file +
  `os.replace`).
- Text edits should default to **UTF-8** rather than re-using a low-confidence single-byte sniff,
  so inserting common characters (arrows, em dashes, emoji, CJK) never corrupts or destroys a file.

### Actual Behavior

The file is truncated to 0 bytes and the operation fails with:

```
ToolError: Ran into 'charmap' codec can't encode character '\u2192' in position 46497:
character maps to <undefined> while trying to write to <path>
```

### Steps to Reproduce

Minimal, deterministic repro against the real `EncodingManager` (the same object `write_file` uses
via the `@with_encoding` decorator):

```python
from pathlib import Path
import charset_normalizer
from openhands.tools.file_editor.utils.encoding import EncodingManager

# A file that genuinely contains Windows-125x high bytes (smart quotes 0x93/0x94,
# em dash 0x97, bullet 0x95, apostrophe 0x92, pound 0xA3, e-acute 0xE9) - i.e. an
[...]

PR #3679, opened 2026-06-12
fix(file_editor): write atomically and fall back to UTF-8 to prevent data loss

shanemort1982:
HUMAN:

Tested locally and end-to-end against the deployed `openhands:main` image: the
windows-125x file + `→` case now preserves the file and writes valid UTF-8, where it
previously truncated the file to 0 bytes.

AGENT:

---

## Why

`file_editor` can destroy a file (truncate it to 0 bytes) when editing a file whose
encoding is sniffed as a non-UTF-8 single-byte codec (e.g. `windows-1252`) and the new
content contains a character that codec cannot represent (arrow, em dash, emoji, CJK).

Two compounding defects:

1. `write_file` opened the target with `open(path, "w", ...)`, truncating it in place
   before writing. A failed write left the file empty.
2. Writes reused the sniffed encoding (`EncodingManager` only carves out `ascii -> utf-8`),
   so an inserted out-of-charset character raised `UnicodeEncodeError`, which then
   truncated the file.

Reported in #3678.

## Summary

- `write_file` now writes to a temp file in the same directory and `os.replace`s it into
  place, so a failed write can never destroy the original.
- If the detected encoding cannot encode the new content, fall back to UTF-8 (with a
  warning) instead of failing and truncating.
- `insert` builds its content in memory and routes through the same atomic write (also
  removes a temp-file leak on failure).

## Issue Number

#3678

## How to Test

Unit (regression tests added in `tests/tools/file_editor/utils/test_encoding.py`):

```bash
uv run pytest tests/tools/file_editor -q          # 155 passed (+3 new)
uv run ruff check openhands-tools/openhands/tools/file_editor/editor.py
uv run pyright openhands-tools/openhands/tools/file_editor/editor.py   # 0 errors
```

End-to-end against the shipped app image (`ghcr.io/openhands/openhands:main`,
openhands-tools 1.28.0), exercising the real `FileEditor`. Seed a file with genuine
windows-125x bytes, then `str_replace` an ASCII anchor with text containing `→`:

- **Before (shipped):** `ToolError: Ran into 'charmap' codec can't encode character
  '\u2192' ... while trying to write to ...`, and the file is left at **0 bytes**.
- **After (this PR):** `WARNING Detected encoding 'windows-1250' cannot represent the new
  content ...; writing as 'utf-8' instead`, file goes **2659 -> 3148 bytes**, valid UTF-8,
  contains the arrow. `RESULT: PRESERVED`.

The shipped `editor.py` in that image is byte-identical to the base of this branch, so the
fix applies to the deployed code as-is.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The UTF-8 fallback rewrites the **entire** file as UTF-8 and only triggers when the new
content cannot be encoded in the file's detected encoding; files that stay representable
keep their original encoding (the existing cp1251 tests still pass). Because the whole file
is transcoded, read-side encoding mis-detection becomes consequential at write time, but
this is still strictly better than the previous truncation-to-zero. Fixing the read-side
mis-detection of mostly-Latin UTF-8 files as a single-byte codec (the root cause noted in
#3678) is out of scope here; this PR makes the write path safe so such cases can never lose
data.


tests/tools/file_editor/utils/test_encoding.py
```diff
@@ -367,6 +367,73 @@ def test_insert_non_utf8_file(temp_non_utf8_file):
         pytest.fail("File was not saved with the correct encoding")
 
 
+def test_str_replace_unencodable_char_preserves_file_and_upgrades_to_utf8(
```
VascoSch92: nit: can you put all the tests together using @pytest.mark.parametrize?

I think it is possible, but not sure.

VascoSch92:
Hey @shanemort1982 

thank you for the PR. 

I just have a couple of nits/nice-to-have:
- The UTF-8 fallback transcodes the entire file, which makes read-side mis-detection consequential at write time. This is still strictly better than truncation-to-zero, so it doesn't block. But it should be explicitly documented in the PR.
- We have already an atomic-write idiom in `llm_profile_store.py` with `NamedTemporaryFile(delete=False) + Path.replace + unlink(missing_ok=True)`. Could you see if you can extract an helper or use the same style in your helper?
- Probe-then-write double-encodes the full content on the success path (minor cost on large files).
- The warning guard `encoding.lower().replace("_","-")` not in ("utf-8","utf8") is near-dead — that branch is only reached after a UnicodeEncodeError, which UTF-8 essentially never raises. Perhaps, delete it.

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -463,12 +462,49 @@ def write_file(self, path: Path, file_text: str, encoding: str = "utf-8") -> Non
...
+        os.close(fd)
+        try:
+            # Preserve the original file's permission bits when it already exists.
+            if path.exists():
+                os.chmod(tmp_name, os.stat(path).st_mode & 0o7777)
+            with open(tmp_name, "w", encoding=write_encoding) as f:
+                f.write(file_text)
+            os.replace(tmp_name, path)
+        except BaseException:
+            try:
```
VascoSch92: consider using `contextlib.suppress`

shanemort1982:
Thanks @VascoSch92. Addressed in 05c7e18:

- **Reused the existing idiom.** The helper now uses `NamedTemporaryFile(delete=False)` + `Path.replace` + `unlink(missing_ok=True)`, matching `llm_profile_store.save`, instead of `mkstemp`/`os.replace`. I did not extract a shared helper because the two live in different packages (`openhands-sdk` vs `openhands-tools`) and the file_editor variant also needs the encoding fallback and permission preservation. Happy to lift it into a shared util if you would prefer.
- **Dropped the probe.** It now encodes once and only falls back to UTF-8 if the actual write raises `UnicodeEncodeError`, so there is no double-encode on the success path.
- **Removed the dead guard.** The warning now fires only on the genuine fallback branch.
- **Documented the transcode.** Both the helper docstring and the PR description now state that the fallback rewrites the whole file as UTF-8, so read-side mis-detection becomes consequential at write time, still strictly better than truncation-to-zero. The read-side mis-detection itself remains the separate concern noted in #3678.

tests/tools/file_editor/utils/test_encoding.py
shanemort1982: Done in d54893d: the str_replace and insert cases are now one `@pytest.mark.parametrize`d test. I kept `test_write_failure_leaves_original_file_intact` separate since it monkeypatches the write to fail and asserts the original is intact with no stray temp file, which does not fit the same shape.

openhands-tools/openhands/tools/file_editor/editor.py
shanemort1982: Removed in 05c7e18. That `try/except OSError: pass` is gone; the helper now matches the `llm_profile_store` idiom and cleans up with `Path.unlink(missing_ok=True)`, so there is no suppressed-exception block left.

shanemort1982:
CI is green across all 27 checks now. Two things the pipeline caught and that are fixed:

- The first windows-tests run flagged a genuine cross-platform bug in the fallback path: `NamedTemporaryFile` was being `unlink`ed while still open, which fails on Windows (`WinError 32`) and masked the `UnicodeEncodeError` that drives the UTF-8 fallback. Now closed before unlink (d45fa95).
- The fallback regression test no longer depends on `charset_normalizer` detection (which differs by platform); it forces the encoding into `write_file` so the behaviour is exercised identically everywhere.

Ready for another look @VascoSch92.

openhands-tools/openhands/tools/file_editor/editor.py
```diff
@@ -463,12 +462,68 @@ def write_file(self, path: Path, file_text: str, encoding: str = "utf-8") -> Non
...
         self.validate_file(path)
         try:
-            # Use open with encoding instead of path.write_text
-            with open(path, "w", encoding=encoding) as f:
-                f.write(file_text)
+            self._atomic_write(path, file_text, encoding)
         except Exception as e:
             raise ToolError(f"Ran into {e} while trying to write to {path}") from None
 
+    def _atomic_write(self, path: Path, file_text: str, encoding: str) -> None:
```
VascoSch92, line 480: Right now `_write_temp_file` cleans up on write failure, and `_atomic_write` cleans up on chmod/replace failure. That's the same file being guarded at two different points. A context manager owns the whole lifetime, so both cleanups become one:

```python
  from contextlib import contextmanager
  from collections.abc import Iterator

  @contextmanager
  def _temp_file(self, path: Path, file_text: str, encoding: str) -> Iterator[Path]:
      """Write file_text to a fresh temp file beside path; remove it on any failure."""
      tmp = tempfile.NamedTemporaryFile(
          mode="w", dir=path.parent, prefix=f".{path.name}.",
          suffix=".tmp", encoding=encoding, delete=False,
      )
      tmp_path = Path(tmp.name)
      try:
          with tmp:
              tmp.write(file_text)
          yield tmp_path
      except BaseException:
          tmp_path.unlink(missing_ok=True)
          raise
```

The nested try/except UnicodeEncodeError exists only to detect whether the encoding can represent the text. But you can detect that up front instead of letting the temp-file write fail and then writing a second time:

```python
  def _atomic_write(self, path: Path, file_text: str, encoding: str) -> None:
      default = self._encoding_manager.default_encoding
      if encoding != default and not _is_encodable(file_text, encoding):
          logger.warning(
              f"Detected encoding '{encoding}' cannot represent the new content "
              f"for {path}; writing as '{default}' instead."
          )
          encoding = default

      with self._temp_file(path, file_text, encoding) as tmp_path:
          if path.exists():
              os.chmod(tmp_path, os.stat(path).st_mode & 0o7777)
          Path.replace(tmp_path, path)


  def _is_encodable(text: str, encoding: str) -> bool:
      try:
          text.encode(encoding)
      except UnicodeEncodeError:
          return False
      return True
  ```

You go from 4 try blocks (2 nested) to 2, neither nested, and each one does a single obvious thing.

shanemort1982: Adopted in 6c6eefc. The temp file is now owned by a `@contextmanager _temp_file` that cleans up on any failure (write, chmod or replace), and `_is_encodable` does the up-front check, skipped when the encoding is already UTF-8 so there is no extra encode on the common path. Down to two flat try blocks. Thanks, much cleaner.

VascoSch92 (approved):

VascoSch92:
Hey @shanemort1982 

thanks for the PR :-)

I fixed the pre-commit. Hope you don't mind.

PR #2154, opened 2026-02-20
fix: redact credentials from plugin/extension URL logs

jpshackelford:
## Summary

Prevent git credentials embedded in `https://user:token@host` URLs from leaking into logs and persisted state across the plugin/extension stack.

- Moves `redact_url_credentials()` to the canonical `sdk/utils/redact.py` module (removes the local copy in `git/utils.py`, which re-exports it for callers like `cached_repo.py`)
- Applies redaction at all URL-logging sites: `extensions/fetch.py`, `extensions/installation/manager.py`, `plugin/source.py`, `plugin/loader.py`, `git/cached_repo.py`
- `ResolvedPluginSource.from_plugin_source()` stores only the redacted URL so credentials never reach persisted state
- Adds `TestRedactUrlCredentialsCentralModule` to verify the function is importable from the central module and has identical behaviour to the `git.utils` re-export

## How to Test

```bash
uv run pytest tests/sdk/git/test_url_redaction.py -v
# All 22 tests should pass
```

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
| python | amd64, arm64 | `nikolaik/python-nodejs:python3.13-nodejs22-slim` | [Link](https://hub.docker.com/_/nikolaik/python-nodejs:python3.13-nodejs22-slim) |
| golang | amd64, arm64 | `golang:1.21-bookworm` | [Link](https://hub.docker.com/_/golang:1.21-bookworm) |

**Pull (multi-arch manifest)**
```bash
# Each variant is a multi-arch manifest supporting both amd64 and arm64
docker pull ghcr.io/openhands/agent-server:ca9ffb1-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-ca9ffb1-python \
  ghcr.io/openhands/agent-server:ca9ffb1-python
```

**All tags pushed for this build**
```
ghcr.io/openhands/agent-server:ca9ffb1-golang-amd64
ghcr.io/openhands/agent-server:ca9ffb1197c737d77ff1df711006210c840b0bd6-golang-amd64
ghcr.io/openhands/agent-server:fix-redact-plugin-url-credentials-golang-amd64
[...]

PR #3689, opened 2026-06-13
fix: redact credentials from GitCommandError.command field

simonrosenberg:
HUMAN: 

 redact credentials from GitCommandError.command field

AGENT: 

## Why

Spotted in QA review of #2154. When `run_git_command()` raises `GitCommandError`, the `.command` list retained raw credential-bearing URLs (e.g. `https://user:token@host`). The error message string was already redacted via `_redact_args_for_logging()`, but callers that serialize or log the exception object could still expose credentials through `.command`.

## Summary

Apply `redact_url_credentials()` to each arg before storing it in `GitCommandError.command`, so the exception is safe to serialize regardless of how callers handle it. The `FileNotFoundError` path (git not installed) is left unredacted since it never receives a URL argument.

## How to Test

```python
from openhands.sdk.git.utils import run_git_command
from openhands.sdk.git.exceptions import GitCommandError

try:
    run_git_command(["git", "ls-remote", "https://oauth2:SECRET@127.0.0.1:1/repo"])
except GitCommandError as e:
    assert "SECRET" not in str(e.command), f"Credential leaked: {e.command}"
    assert "****" in str(e.command)
    print("PASS")
```

Or run the existing git test suite: `uv run pytest tests/sdk/git/ -q`

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:b4ba8da-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-b4ba8da-python \
  ghcr.io/openhands/agent-server:b4ba8da-python
```

```
ghcr.io/openhands/agent-server:b4ba8da-golang-amd64
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-golang-amd64
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-golang-amd64
ghcr.io/openhands/agent-server:b4ba8da-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:b4ba8da-golang-arm64
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-golang-arm64
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-golang-arm64
ghcr.io/openhands/agent-server:b4ba8da-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:b4ba8da-java-amd64
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-java-amd64
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-java-amd64
ghcr.io/openhands/agent-server:b4ba8da-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:b4ba8da-java-arm64
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-java-arm64
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-java-arm64
ghcr.io/openhands/agent-server:b4ba8da-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:b4ba8da-python-amd64
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-python-amd64
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-python-amd64
ghcr.io/openhands/agent-server:b4ba8da-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:b4ba8da-python-arm64
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-python-arm64
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-python-arm64
ghcr.io/openhands/agent-server:b4ba8da-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:b4ba8da-golang
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-golang
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-golang
ghcr.io/openhands/agent-server:b4ba8da-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:b4ba8da-java
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-java
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-java
ghcr.io/openhands/agent-server:b4ba8da-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:b4ba8da-python
ghcr.io/openhands/agent-server:b4ba8da0efc50c8e6d2e714d288f17d433735aa1-python
ghcr.io/openhands/agent-server:fix-git-command-error-redact-command-field-python
ghcr.io/openhands/agent-server:b4ba8da-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `b4ba8da-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `b4ba8da-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/git/utils.py
+++ b/openhands-sdk/openhands/sdk/git/utils.py
@@ -15,7 +15,6 @@ logger = logging.getLogger(__name__)
 GIT_EMPTY_TREE_HASH = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 
 
-
 def run_git_command(
     args: list[str],
     cwd: str | Path | None = None,
--- a/openhands-sdk/openhands/sdk/git/utils.py
+++ b/openhands-sdk/openhands/sdk/git/utils.py
@@ -61,13 +61,14 @@ def run_git_command(
         if result.returncode != 0:
             # Redact credentials from command for logging and error messages
             cmd_str = _redact_args_for_logging(args)
+            redacted_args = [redact_url_credentials(a) for a in args]
             error_msg = f"Git command failed: {cmd_str}"
             logger.error(
                 f"{error_msg}. Exit code: {result.returncode}. Stderr: {result.stderr}"
             )
             raise GitCommandError(
                 message=error_msg,
-                command=args,
+                command=redacted_args,
                 exit_code=result.returncode,
                 stderr=result.stderr.strip(),
             )
@@ -77,11 +78,12 @@ def run_git_command(
 
     except subprocess.TimeoutExpired as e:
         cmd_str = _redact_args_for_logging(args)
+        redacted_args = [redact_url_credentials(a) for a in args]
         error_msg = f"Git command timed out: {cmd_str}"
         logger.error(error_msg)
         raise GitCommandError(
             message=error_msg,
-            command=args,
+            command=redacted_args,
             exit_code=-1,
             stderr="Command timed out",
         ) from e
--- a/openhands-sdk/openhands/sdk/git/utils.py
+++ b/openhands-sdk/openhands/sdk/git/utils.py
@@ -15,20 +15,6 @@ logger = logging.getLogger(__name__)
 GIT_EMPTY_TREE_HASH = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 
 
-def _redact_args_for_logging(args: list[str]) -> str:
-    """Redact credentials from git command arguments for safe logging.
-
-    Joins command args with shlex.join and redacts any URLs containing credentials.
-
-    Args:
-        args: List of command arguments.
-
-    Returns:
-        A string representation of the command with credentials redacted.
-    """
-    redacted_args = [redact_url_credentials(arg) for arg in args]
-    return shlex.join(redacted_args)
-
 
 def run_git_command(
     args: list[str],
@@ -48,6 +34,9 @@ def run_git_command(
     Raises:
         GitCommandError: If the git command fails
     """
+    redacted_args = [redact_url_credentials(a) for a in args]
+    cmd_str = shlex.join(redacted_args)
+
     try:
         result = subprocess.run(
             args,
@@ -59,9 +48,6 @@ def run_git_command(
         )
 
         if result.returncode != 0:
-            # Redact credentials from command for logging and error messages
-            cmd_str = _redact_args_for_logging(args)
-            redacted_args = [redact_url_credentials(a) for a in args]
             error_msg = f"Git command failed: {cmd_str}"
             logger.error(
                 f"{error_msg}. Exit code: {result.returncode}. Stderr: {result.stderr}"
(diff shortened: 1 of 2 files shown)


jpshackelford:
Thanks for catching this in the QA review of #2154 and turning it around so quickly — redacting `.command` so the exception is safe to serialize and it closes a real gap. 🙏

One fix it seems like we ought to make before merge. This PR covers the `returncode != 0` and `TimeoutExpired` paths, but the third `raise GitCommandError(...)` — the `FileNotFoundError` branch — still passes `command=args` unredacted. The PR description notes this is fine because that path "never receives a URL argument," but I think `FileNotFoundError` is actually raised by `subprocess.run` based on `args[0]` (the executable) regardless of the rest of `args`, so a credential-bearing URL further down the list still survives. `CachedRepo.clone()` passes the clone URL straight into `run_git_command`, so I was able to reproduce a leak against the current code:

```python
args = ['git-not-on-path', 'clone', 'https://oauth2:SUPERSECRET@github.com/o/r.git', '/tmp/x']
try:
    run_git_command(args)
except GitCommandError as e:
    print(e.command)
    # ['git-not-on-path', 'clone', 'https://oauth2:SUPERSECRET@github.com/o/r.git', '/tmp/x']
    # -> credential survives in .command
```

Since the three raise sites share the same intent, what do you think about computing the redacted args once near the top and reusing them everywhere? It also keeps the message/log/exception perfectly in sync:

```python
redacted_args = [redact_url_credentials(a) for a in args]
```

…then use `redacted_args` for the message, the debug log, and all three `command=` arguments (including the `FileNotFoundError` branch). A small test asserting no credentials leak into `.command` across all three paths (failure / timeout / git-not-found) would lock it in nicely.

Two related items I don't think need to block this PR  — I've filed as #3751:
- The `.stderr` field is still stored/logged raw (`utils.py:66`). Git usually strips userinfo from URLs it echoes, so it's lower risk, but it's the same serialization concern. Worth noting `redact_url_credentials()` is anchored (`^…$`) and won't catch a URL embedded inside stderr text, so this one needs a small substring-capable redactor rather than a drop-in.
- `workspace/repo.py:149` logs a raw user-provided URL, and `repo.py`'s `_mask_token`/`_mask_url` duplicate this redaction logic — both could eventually consolidate onto the central helper.

Really nice catch! Thanks so much!

jpshackelford (commented):
Thanks very much for this! Looks like one scenario slips through as described in https://github.com/OpenHands/software-agent-sdk/pull/3689#issuecomment-4718421450 but once addressed this should be go to go. Thanks again!

simonrosenberg:
Thanks for the thorough review and the reproduction script — really helpful!

Fixed in b4ba8da: computed `redacted_args` once before the `try` block and used it across all three raise sites, including `FileNotFoundError`. Also removed the now-dead `_redact_args_for_logging` helper since the single upfront computation covers everything.

Added `TestRunGitCommandCredentialRedaction` with three tests — one per raise path (non-zero returncode, `TimeoutExpired`, `FileNotFoundError`) — asserting the raw URL never appears in `.command` and the redacted form does.

The two items in #3751 noted as non-blocking look right to me — happy to pick those up in a follow-up.

jpshackelford (approved):

PR #3642, opened 2026-06-10
fix(llm): recognize claude-fable-5 as a reasoning model

juanmichelini:
## Problem

The smoke-test SWE-bench eval for `claude-fable-5` ([eval run `27243970491`](https://openhands-eval-monitor.vercel.app/?run=swebench%2Flitellm_proxy-anthropic-claude-fable-5%2F27243970491%2F&days=15)) failed every instance with:

```
LLMBadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error",
  "message":"`temperature` is deprecated for this model."}}
```

Total proxy spend: $0. Every instance hit the wall on its first LLM call. The integration-test workflow for the same commit passed, which made this look like a pipeline-only quirk, but it turned out to be a real SDK gap.

## Root cause

`openhands-sdk/openhands/sdk/llm/options/chat_options.py` already strips `temperature` and `top_p` for any model the SDK considers a reasoning model:

```python
if supports_reasoning_effort:
    ...
    if "gemini" not in llm.model.lower():
        out.pop("temperature", None)
        out.pop("top_p", None)
```

`supports_reasoning_effort` is decided by `_supports_reasoning_effort(model)` in `model_features.py`, which delegates to LiteLLM's `get_supported_openai_params(...)` registry. For `claude-opus-4-7` / `4-8`, LiteLLM reports `reasoning_effort` as supported, so the SDK pops `temperature` and Anthropic never sees it. For the brand-new `claude-fable-5` id, LiteLLM has no metadata yet, so `supports_reasoning_effort` is `False`, the strip branch is skipped, and the eval pipeline's `temperature: 0.0` default flows straight through to Anthropic — which now rejects it.

## Fix

Add an SDK-side override allowlist (`REASONING_EFFORT_MODELS`) consulted before the LiteLLM lookup, and put `claude-fable-5` on it. The chat-options layer then pops `temperature`/`top_p` and forwards `reasoning_effort` for fable-5 too, restoring parity with the Opus models. Entries should be dropped once LiteLLM ships metadata for the model.

This is a minimal, behavior-preserving change:

```python
REASONING_EFFORT_MODELS: list[str] = [
    # https://www.anthropic.com/news/claude-fable-5
    "claude-fable-5",
]

def _supports_reasoning_effort(model: str | None) -> bool:
    if model_matches(model or "", REASONING_EFFORT_MODELS):
        return True
    return "reasoning_effort" in _normalized_supported_openai_params(model)
```

## Verification

[...]

PR #3427, opened 2026-05-28
fix(llm): strip temperature/top_p for claude-opus-4-8

juanmichelini:
## Problem

Anthropic's `claude-opus-4-8` is a reasoning-only model that rejects requests carrying `temperature` (or `top_p`) with:

```
AnthropicException - {"type":"error","error":{
  "type":"invalid_request_error",
  "message":"`temperature` is deprecated for this model."
}}
```

This breaks SWE-bench evaluation runs against the model — every conversation fails with `LLMBadRequestError`. Example: [`swebench/litellm_proxy-anthropic-claude-opus-4-8/26604804421`](https://openhands-eval-monitor.vercel.app/?run=swebench%2Flitellm_proxy-anthropic-claude-opus-4-8%2F26604804421%2F&days=15) — all 3 critic attempts × all retries produced the same 400 from Anthropic.

## Why the existing logic doesn't catch it

`select_chat_options` already strips `temperature`/`top_p` in two places:

1. The **reasoning-effort path** — gated on `_supports_reasoning_effort()`, which delegates to litellm's `get_supported_openai_params`. Today litellm reports `reasoning_effort: False` for `claude-opus-4-8` (and even `temperature: True` for the `anthropic/...` form), so this branch doesn't fire:

   ```
   claude-opus-4-8 -> temperature in supported: False | reasoning_effort: False
   anthropic/claude-opus-4-8 -> temperature in supported: True | reasoning_effort: False
   ```

2. The **extended-thinking path** — gated on `EXTENDED_THINKING_MODELS`, but `claude-opus-4-8` wasn't on the list.

Result: benchmark configs that explicitly set `temperature=0.0` (e.g. SWE-bench `litellm_extra_body`) get the value forwarded and Anthropic rejects it.

The SDK integration tests in [run `26599152082`](https://github.com/OpenHands/software-agent-sdk/actions/runs/26599152082) didn't surface the issue because they leave `LLM.temperature` at its default `None` — so nothing was ever sent.

## Fix

Add `claude-opus-4-8` to `EXTENDED_THINKING_MODELS`. That routes it through the path already used for `claude-sonnet-4-6` / `claude-haiku-4-5`, which:

- strips `temperature` and `top_p` from the outgoing request, and
- enables the thinking-budget header when `extended_thinking_budget` is configured (which the SWE-bench config already does: `extended_thinking_budget=200000`).

## Changes

[...]

PR #3695, opened 2026-06-13
fix(llm): recognize claude-opus-4-8 as a reasoning model on Bedrock

### LINKED ISSUE #3694: claude-opus-4-8 on Bedrock cross-region ids not recognized as a reasoning model (temperature/top_p leak → 400)
Thanks for the SDK and for the careful fix history around opus-4-8 (#3427/#3441) and fable-5 (#3642) — that context made this easy to trace.

## Behavior

With the pinned `litellm==1.84.1`, `model_features._supports_reasoning_effort` returns `True` for the first-party `anthropic/claude-opus-4-8` (as #3441 noted), but `False` for the Bedrock cross-region inference (CRIS) ids: `bedrock/{us,eu,apac,global}.anthropic.claude-opus-4-8-v1:0`. Those route through the non-reasoning branch in `chat_options.py`, leaving `temperature`/`top_p` in the request, which Anthropic rejects (`temperature is deprecated for this model`).

opus-4-7's Bedrock ids and fable-5's Bedrock ids are both recognized (4-7 via LiteLLM metadata, fable-5 via the existing `REASONING_EFFORT_MODELS` override) — only opus-4-8's Bedrock forms are missing.

## Repro (no network)

```python
from openhands.sdk.llm.utils.model_features import _supports_reasoning_effort
_supports_reasoning_effort("anthropic/claude-opus-4-8")                 # True
_supports_reasoning_effort("bedrock/us.anthropic.claude-opus-4-8-v1:0") # False  ← bug
```

## Proposed fix

Add `"claude-opus-4-8"` to `REASONING_EFFORT_MODELS` (`model_features.py:102`, the #3642 mechanism) — **not** `EXTENDED_THINKING_MODELS` (the #3427 path that #3441 reverted, since that injects the legacy `thinking.type=enabled` block + `interleaved-thinking` header Anthropic now rejects). Membership is case-insensitive substring (`model_matches`), so one entry covers all CRIS prefixes. I have a minimal patch + offline tests ready.

kimnamu:
HUMAN:

I reviewed this change locally against the pinned litellm 1.84.1, confirmed the Bedrock cross-region reasoning gap, and verified the regression tests fail before / pass after the one-line fix.

AGENT:

---

## Why

LiteLLM 1.84.1 recognizes `anthropic/claude-opus-4-8` but not its Bedrock cross-region ids (`bedrock/{us,eu,apac,global}.anthropic.claude-opus-4-8-v1:0`), so on Bedrock the model falls through to the non-reasoning branch and leaks `temperature`/`top_p` → Anthropic 400. This mirrors the fable-5 gap fixed in #3642; opus-4-7's Bedrock ids are already covered by LiteLLM metadata.

It deliberately uses `REASONING_EFFORT_MODELS` (strips temp/top_p only), **not** `EXTENDED_THINKING_MODELS` — adding it there is what #3427 did and #3441 reverted, because that branch injects the legacy `thinking.type=enabled` block + `interleaved-thinking` header that Anthropic now rejects for opus-4-8.

## Summary

- Add `"claude-opus-4-8"` to `REASONING_EFFORT_MODELS` (`model_features.py`) — one data line, mirrors the fable-5 fix (#3642).
- Membership is case-insensitive substring, so the single entry covers all Bedrock CRIS prefixes (`us./eu./apac./global.`).

## Issue Number

Fixes #3694

## How to Test

Run the added offline unit tests (no network):

```
uv run pytest tests/sdk/llm/test_model_features.py -k opus_4_8
uv run pytest tests/sdk/llm/options/test_chat_options.py -k bedrock_opus_4_8
```

Before/after evidence: with the source line reverted (`git stash push -- openhands-sdk/openhands/sdk/llm/utils/model_features.py`), the new tests FAIL — `temperature` leaks and `_supports_reasoning_effort("bedrock/us.anthropic.claude-opus-4-8-v1:0")` returns `False` (5 failed, 36 passed). Restoring the line → `174 passed`. Local `pre-commit` (ruff + pyright) passes.

| item | Before | After |
|---|---|---|
| `_supports_reasoning_effort("bedrock/us.anthropic.claude-opus-4-8-v1:0")` | ❌ False | ✅ True |
| temp/top_p in request for that id | ❌ sent (→ Anthropic 400) | ✅ stripped |
| `anthropic/claude-opus-4-8` (already worked) | ✅ True | ✅ unchanged |
| legacy `thinking.type=enabled` block / beta header | n/a | ✅ NOT injected (asserted in test) |
| public API / signatures | — | ✅ unchanged (data line only) |

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, rationale, and test results before submission.


kimnamu:
Yes, still relevant and ready for review 🙏 — thanks for the nudge.

This is a small, self-contained fix: `claude-opus-4-8` on Bedrock (including the cross-region inference-profile IDs like `us.anthropic.claude-opus-4-8...`) wasn't being recognized as a reasoning model, so `temperature` was sent and rejected with a 400. The change adds it to the existing reasoning-model detection (the same path the sibling models already use), and the regression test reproduces the 400 before the fix and passes after.

The branch is still **MERGEABLE** with no conflicts and the PR-description check is green. Requesting review whenever a maintainer has a moment — happy to rebase or adjust if anything's needed.

enyst (approved):
Thank you, and sorry for the delay!

PR #3696, opened 2026-06-14
feat(subagent): give spawned sub-agents a condenser by default

ak684:
HUMAN:

Closes the gap where spawned sub-agents had no condenser and dead-ended on context overflow. Reviewed the change and the real-LLM before/after. Subagents now auto-compact and keep going instead of eventually error'ing out on large workflows.

AGENT:

---

## Why

Spawned sub-agents (the `task` / `delegate` / `workflow` tools) never received the summarizing condenser that every top-level agent gets, so a deep sub-agent run dead-ended on context-window overflow instead of auto-compacting. This made sub-agents a special case that silently lacked the context management the main
agent has.

## Summary

- `agent_definition_to_factory` now attaches a default summarizing condenser to sub-agents, sharing sizing with the standard preset via a new `default_condenser` SDK helper (`get_default_condenser` delegates to it).
- The condenser LLM uses a distinct `usage_id="condenser"` so its tokens are tracked, not deduped out of conversation stats (first-write-wins on `usage_id`).
- New optional `AgentDefinition.condenser` field keeps it overridable and serializes for the remote agent-server path: omit/`None` → default; `none`/`false` → `NoOpCondenser` (disable); a mapping → a custom condenser spec.

## How to Test

Unit + full suite (in this branch's worktree):

```
uv run pytest tests/sdk tests/tools         # 5500 passed, 11 skipped, 13 xfailed, 0 failed
uv run ruff check && uv run ruff format      # clean
uv run pytest -k condenser tests/sdk/subagent  # 9 new tests
```

Real-LLM end-to-end before/after (identical 7-message input, run against a live model via the standalone SDK):

| | events retained | condensations fired | usage_ids tracked |
|---|:---:|:---:|:---:|
| **BEFORE** (no condenser) | 15 | 0 | `['agent']` |
| **AFTER** (this change) | 16 | 1 | `['agent', 'condenser']` |

Live AFTER log excerpt:
```
[A] default factory condenser OK: max_size=80 usage_id=condenser
Auto Conversation Condensation Triggered. Forgetting 9 events
  [Summary of Events Being Forgotten] ... (real LLM-generated summary)
[B] events=16 condensation_events=1 usage_ids=['agent', 'condenser']
```

BEFORE never compacts — the view grows unbounded (the path that eventually overflows
the real context window). AFTER fires the condenser (summarized + forgot 9 events),
the sub-agent continues, and its tokens are tracked under a distinct `condenser`
usage_id.

> Note on the BEFORE row: it is the faithful illustration of the mechanism, not a
> literal crash. Triggering an actual context-window-exceeded error live would
> require filling a ~200k-token window (expensive/slow), so the contrast above is
> the cheap, faithful demonstration of *why* the dead-end happens. The AFTER fix
> was verified firing end to end against a real LLM.

## Type

- [x] Feature

## Notes

Additive and back-compat: the new field defaults to today's behavior, programmatic agents built without the factory are unaffected, and `AgentDefinition.condenser` is serialization-safe for the remote path (`CondenserBase` is a discriminated union; round-trip covered by a test).

### DIFF
--- a/openhands-sdk/openhands/sdk/context/condenser/__init__.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/__init__.py
@@ -5,6 +5,7 @@ from openhands.sdk.context.condenser.base import (
 )
 from openhands.sdk.context.condenser.llm_summarizing_condenser import (
     LLMSummarizingCondenser,
+    default_condenser,
 )
 from openhands.sdk.context.condenser.no_op_condenser import NoOpCondenser
 from openhands.sdk.context.condenser.pipeline_condenser import PipelineCondenser
@@ -17,4 +18,5 @@ __all__ = [
     "PipelineCondenser",
     "LLMSummarizingCondenser",
     "NoCondensationAvailableException",
+    "default_condenser",
 ]
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -1,6 +1,7 @@
 import os
 from collections.abc import Sequence
 from enum import Enum
+from typing import Final
 
 from pydantic import Field, model_validator
 
@@ -466,3 +467,16 @@ class LLMSummarizingCondenser(RollingCondenser):
 
         logger.error("Hard context reset summarization failed after multiple attempts.")
         return None
+
+
+# Sizing for the standard summarizing condenser. Kept here so the default agent and
+# spawned sub-agents stay in sync.
+_DEFAULT_MAX_SIZE: Final[int] = 80
+_DEFAULT_KEEP_FIRST: Final[int] = 4
+
+
+def default_condenser(llm: LLM) -> LLMSummarizingCondenser:
+    """Standard summarizing condenser used by the default agent and sub-agents."""
+    return LLMSummarizingCondenser(
+        llm=llm, max_size=_DEFAULT_MAX_SIZE, keep_first=_DEFAULT_KEEP_FIRST
+    )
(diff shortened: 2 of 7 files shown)


VascoSch92:
@ak684 this pr add an api for file-based agent which i'm not sire it is what we wanted.

There it was already another PR for that: https://github.com/OpenHands/software-agent-sdk/pull/2349/files

VascoSch92:
The default sub-agent condenser (the PR's main goal) is token-correct and unaffected. 

Both trace to the passthrough at `subagent/registry.py:262`, which — unlike the else branch — neither freshens the instance per spawn nor normalizes the condenser LLM's usage_id.

```python
  condenser = (
      agent_def.condenser                       # <- passed by reference, usage_id untouched
      if agent_def.condenser is not None
      else default_condenser(llm.model_copy(update={"usage_id": "condenser"}))
  )
```

Both were reproduced through the real LocalConversation → llm_registry.add() → ConversationStats path (no LLM calls; tokens recorded the way Telemetry would).

- a custom condenser instance is shared across every sub-agent spawn → metrics double-count. The default branch builds a fresh condenser each call; the explicit branch hands every spawn the same object (same LLM, same Metrics). Two sub-agent tasks of the same type then accumulate into one shared metrics object:

```text
  task#1 condenser tokens right after task#1 : 500   (correct)
  task#1 condenser tokens after task#2 ran   : 1000  (should still be 500)
  two conversations share ONE metrics object : True
```

Impact: inflated/cross-contaminated condenser cost, propagated to the parent via the task:{id} rollup; under concurrent same-type sub-agents it's also a mutable-state race on the shared Metrics/Telemetry.

- the explicit path doesn't enforce a distinct usage_id → silent token loss. If a custom condenser's LLM shares the agent's usage_id (trivially easy — both default to "default", or a frontmatter condenser:
  mapping that omits usage_id), the registration dedup (local_conversation.py:855) drops it.

VascoSch92:
Moreover, the docs should be update for file-based agents

PR #3766, opened 2026-06-16
fix(sdk): normalize empty assistant chat content

### LINKED ISSUE #3765: [Bug]: Empty assistant message serializes to non-string content for Chat Completions

### Bug Description

When an assistant message has empty content and no tool calls, the SDK can serialize it for OpenAI Chat Completions as `{"role": "assistant", "content": []}` when list serialization is enabled.

This can happen after an LLM returns an empty assistant response: `Message.from_llm_chat_message(...)` keeps the SDK-internal message valid as `content=[]`, but the next Chat Completions request may send that empty list back to an OpenAI-compatible provider.

Some OpenAI-compatible backends reject assistant messages where `content` is an empty list. They expect `content` to be a string, typically `""`, when there are no tool calls.

The SDK already has similar empty-content handling for assistant messages with tool calls via `_remove_content_if_empty(...)`, but the plain assistant-message case appears to remain uncovered.

### Expected Behavior

For assistant messages without tool calls, empty SDK content should serialize to an OpenAI-compatible Chat Completions message, for example:

```python
{"role": "assistant", "content": ""}

### Actual Behavior

With list serialization enabled, an empty assistant message without tool calls serializes as:

```python
{"role": "assistant", "content": []}
```

### Steps to Reproduce

1. Install the latest OpenHands SDK.
2. Run the following minimal reproduction:

```python
from openhands.sdk.llm.message import Message

serialization_opts = {
    "cache_enabled": False,
    "vision_enabled": False,
    "function_calling_enabled": True,
    "force_string_serializer": False,
    "send_reasoning_content": False,
}

message = Message(role="assistant", content=[])
print(message.to_chat_dict(**serialization_opts))
```

### Installation Method

Installed as a uv-managed project dependency from PyPI:  `openhands-sdk==1.23.1`  The same serialization behavior was also checked against current `main` / latest release.

_No response_

### SDK Version

v1.28.1 / current main

### Version Confirmation

### Python Version

3.12 / 3.13

Qwen/Qwen3.5-122B-A10B served through SGLang v0.5.9 via an OpenAI-compatible Chat Completions endpoint

### Operating System

Linux

### Logs and Error Messages

Example downstream failure shape from an OpenAI-compatible backend:

```text
BadRequestError: messages[N].content: expected a string or valid content object, got empty/non-string content for assistant message
```

### Minimal Code Sample

```python
from openhands.sdk.llm.message import Message

serialization_opts = {
    "cache_enabled": False,
    "vision_enabled": False,
    "function_calling_enabled": True,
    "force_string_serializer": False,
    "send_reasoning_content": False,
}

message = Message(role="assistant", content=[])

assert message.to_chat_dict(**serialization_opts) == {
    "role": "assistant",
    "content": "",
}
```

Related:
- #1093 fixed empty content handling for assistant tool-call messages.
- This issue covers the remaining plain assistant-message case: `role == "assistant"`, no `tool_calls`, and empty list-serialized content.
- #2797 also mentions empty assistant responses, but it concerns the Responses API subscription path rather than Chat Completions message serialization.

A narrowly scoped fix could normalize only the Chat Completions wire format for `role == "assistant"`, empty list content, and no `tool_calls`, without changing the internal `Message(content=[])` representation or persisted event shape.

bozhnyukAlex:
HUMAN:
I reproduced the issue in a downstream OpenAI-compatible Chat Completions integration during development of my coding assistant. Also reviewed the proposed SDK-level fix.

---

AGENT:
Prepared with assistance from OpenAI Codex.

## Why

Empty plain assistant messages can be serialized for Chat Completions with list-shaped content when function calling, vision, or prompt caching is enabled:

```json
{"role": "assistant", "content": []}
```

Some OpenAI-compatible Chat Completions backends reject that payload shape for assistant messages. The existing empty-content cleanup only applied to assistant tool-call messages, so assistant messages without `tool_calls` were left unnormalized.

## Summary

- Normalize exact empty list content on plain assistant Chat Completions messages to an empty string.
- Preserve the existing assistant tool-call cleanup behavior.
- Add a regression test for the list-serializer path.

## Issue Number

Fixes #3765

## How to Test

Regression and related serialization tests:

```bash
uv run pytest tests/sdk/llm/test_message.py
uv run pytest tests/sdk/llm/test_message_serialization.py
uv run pytest tests/sdk/llm
make lint
```

Results:

- `tests/sdk/llm/test_message.py`: 29 passed
- `tests/sdk/llm/test_message_serialization.py`: 16 passed
- `tests/sdk/llm`: 840 passed, 9 warnings
- `make lint`: passed

Reproduction check after the fix:

```bash
uv run python - <<'PY'
from openhands.sdk.llm.message import Message

opts = {
    "cache_enabled": False,
    "vision_enabled": False,
    "function_calling_enabled": True,
    "force_string_serializer": False,
    "send_reasoning_content": False,
}
print(Message(role="assistant", content=[]).to_chat_dict(**opts))
PY
```

Output:

```python
{'content': '', 'role': 'assistant'}
```

## Video/Screenshots

N/A. This is a Chat Completions serialization bugfix; the reproduction command above shows the affected payload shape.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This only changes Chat Completions wire-format serialization for plain assistant messages whose serialized content is exactly an empty list. It does not change stored `Message.content`, persisted event shapes, Responses API serialization, or assistant tool-call message handling.

### DIFF
--- a/tests/sdk/llm/test_message.py
+++ b/tests/sdk/llm/test_message.py
@@ -197,6 +197,19 @@ def test_message_tool_calls_strip_blank_list_content():
     assert "content" not in result
 
 
+def test_empty_assistant_message_uses_string_content_in_list_serializer():
+    """List-serialized empty assistant messages should keep string content."""
+    from openhands.sdk.llm.message import Message
+
+    message = Message(role="assistant", content=[])
+
+    result = message.to_chat_dict(
+        **{**DEFAULT_SERIALIZATION_OPTS, "function_calling_enabled": True}
+    )
+
+    assert result == {"content": "", "role": "assistant"}
+
+
 def test_message_from_llm_chat_message_function_role_error():
     """Test Message.from_llm_chat_message with function role raises error."""
     from litellm.types.utils import Message as LiteLLMMessage
--- a/tests/sdk/llm/test_message.py
+++ b/tests/sdk/llm/test_message.py
@@ -1,6 +1,7 @@
 from unittest.mock import patch
 
 import pytest
+from litellm.types.utils import Message as LiteLLMMessage
 
 from openhands.sdk.llm.message import Message
 
@@ -212,8 +213,6 @@ def test_empty_assistant_message_uses_string_content_in_list_serializer():
 
 def test_empty_assistant_from_llm_chat_message_uses_string_content_in_list_serializer():
     """Empty assistant LLM responses should keep string content."""
-    from litellm.types.utils import Message as LiteLLMMessage
-
     litellm_message = LiteLLMMessage(role="assistant", content=None)
     message = Message.from_llm_chat_message(litellm_message)
 
@@ -226,8 +225,6 @@ def test_empty_assistant_from_llm_chat_message_uses_string_content_in_list_seria
 
 def test_message_from_llm_chat_message_function_role_error():
     """Test Message.from_llm_chat_message with function role raises error."""
-    from litellm.types.utils import Message as LiteLLMMessage
-
     from openhands.sdk.llm.message import Message
 
     litellm_message = LiteLLMMessage(role="function", content="Function response")  # type: ignore
@@ -238,8 +235,6 @@ def test_message_from_llm_chat_message_function_role_error():
 
 def test_message_from_llm_chat_message_with_non_string_content():
     """Test Message.from_llm_chat_message with non-string content."""
-    from litellm.types.utils import Message as LiteLLMMessage
-
     from openhands.sdk.llm.message import Message
 
     # Create a message with non-string content (None or list)
--- a/tests/sdk/llm/test_message.py
+++ b/tests/sdk/llm/test_message.py
@@ -2,6 +2,8 @@ from unittest.mock import patch
 
 import pytest
 
+from openhands.sdk.llm.message import Message
+
 
 # Default serialization options for to_chat_dict() - tests can override as needed
 DEFAULT_SERIALIZATION_OPTS = {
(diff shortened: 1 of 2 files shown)


tests/sdk/llm/test_message.py
```diff
@@ -197,6 +197,19 @@ def test_message_tool_calls_strip_blank_list_content():
     assert "content" not in result
 
 
+def test_empty_assistant_message_uses_string_content_in_list_serializer():
+    """List-serialized empty assistant messages should keep string content."""
+    from openhands.sdk.llm.message import Message
```
VascoSch92: this can be top of the file

tests/sdk/llm/test_message.py
```diff
@@ -197,6 +197,19 @@ def test_message_tool_calls_strip_blank_list_content():
     assert "content" not in result
 
 
+def test_empty_assistant_message_uses_string_content_in_list_serializer():
```
VascoSch92, line 203: Could you also add tests for from_llm_chat_message(content=None) ?

VascoSch92 (commented):
It seems good. Just a couple of nits.

tests/sdk/llm/test_message.py
bozhnyukAlex: Done

tests/sdk/llm/test_message.py
bozhnyukAlex: Done

tests/sdk/llm/test_message.py
```diff
@@ -197,6 +199,31 @@ def test_message_tool_calls_strip_blank_list_content():
...
+    result = message.to_chat_dict(
+        **{**DEFAULT_SERIALIZATION_OPTS, "function_calling_enabled": True}
+    )
+
+    assert result == {"content": "", "role": "assistant"}
+
+
+def test_empty_assistant_from_llm_chat_message_uses_string_content_in_list_serializer():
+    """Empty assistant LLM responses should keep string content."""
+    from litellm.types.utils import Message as LiteLLMMessage
```
VascoSch92: top of the file :-)

VascoSch92 (approved):
Just a nit then we are good to merge ;-)

tests/sdk/llm/test_message.py
bozhnyukAlex: Done

VascoSch92:
Thanks

PR #3781, opened 2026-06-17
[AgentProfile][agent-server] active_agent_profile_id + /api/agent-profiles router + migration seed

simonrosenberg:
HUMAN:

Phase 2 of the AgentProfile epic (#3713): serve the new store over HTTP and add the separate, pointer-only active-profile selector. Reviewed the seed mapping and the FK-on-delete wiring locally; behaviour matches the issue's creation-time-only contract.

---

AGENT:

## Why

Epic #3713 (AgentProfiles) needs the merged `AgentProfileStore` (#3716) exposed over HTTP so the frontend/cloud can list, edit, and select named agent launch specs. Selecting an `AgentProfile` is **creation-time-only**: unlike the existing LLM `/activate` (which rewrites `agent_settings.llm`), activating an `AgentProfile` must only move a pointer and never mutate live `agent_settings`. This PR adds that surface and a one-time lazy migration so existing single-config users land with one profile already present.

Runs in **parallel with #3717** (the resolver); it depends only on the merged store, not the resolver.

## Summary

- **Active pointer:** `PersistedSettings.active_agent_profile_id: str | None = None` — distinct from `active_profile` (the active *LLM* profile name). Additive with a default, so older settings files load with it `None` (no schema-version bump). Threaded through `SettingsUpdatePayload` + `update()`, the SDK `SettingsResponse`/`SettingsUpdateRequest`, and the `settings_router` GET/PATCH passthrough (PATCH can set *or* clear it, independently of `active_profile`).
- **`/api/agent-profiles` router** (mirrors `profiles_router.py`): list / get / save / delete / rename + `POST /{id}/activate` that sets **only** `active_agent_profile_id` via `settings_store.update` — **no `agent_settings` write**. Activation is keyed on the stable `id`, so it survives renames. FK/store error mapping: `ProfileReferenced`/`FileExistsError` → 409, `FileNotFoundError` → 404, `TimeoutError` → 503, `ValueError` → 400, `ProfileLimitExceeded` → 409.
- **Lazy migration seed:** first `GET /api/agent-profiles` on an empty store with a null pointer seeds exactly **one** default profile from the current `agent_settings` (OpenHands → references the active LLM profile, falling back to `"default"`; ACP → an ACP profile; `mcp_server_refs=null` = all) and sets the active pointer. Conservative — one profile, never one-per-LLM-profile.
[...]

Issue #3713, opened 2026-06-15
[AgentProfile] Epic: first-class Agent Profiles (named launch setups) across SDK, agent-server, canvas, cloud


simonrosenberg:
## Goal

Ship **Agent Profiles**: a library of named, reusable agent setups so a user can save *everything* about how an agent runs — which agent (built-in OpenHands/CodeAct vs. an ACP CLI: Claude Code / Codex / Gemini), the model + credentials, MCP servers, skills, and personality — then pick one to start a conversation. **Easy UX is the product goal**: Settings → Agents becomes a profile library; you launch with a profile and, where it's safe, switch the model live.

This epic tracks the implementation. It is **local-first, cloud-ready**: build on the SDK agent-server + agent-canvas first, with the cloud/SaaS surface following the same contract.

## Status (2026-07-06)

**Phases 1–3 released; Phases 4 + 5 are implemented and in review; a launch-blocking tools bug was found and fixed; the capabilities follow-up is designed and tracked.**

- **Releases:** SDK **v1.30.0** (2026-07-01, ships OpenHands/OpenHands#3930's public `openhands.sdk.profiles` symbols) · **v1.31.0** (2026-07-02, ships `skill_refs` + resolver `available_skills` — OpenHands/OpenHands#3868/#3870) · **v1.31.1** (2026-07-05, `latest`). The OpenHands/OpenHands#3815 store-isolation fix (PR OpenHands/OpenHands#3816) merged 2026-06-22 and is released — the Phase-2 warning below is resolved.
- **Phase 4 (canvas) → PR agent-canvas#1571** (supersedes OpenHands/OpenHands#1550 — close it on merge): Settings → Agent IS the profile library (local + cloud), launch-from-profile, kind-gated in-conversation switch, onboarding lands on a profile. Implements OpenHands/OpenHands#3726 + OpenHands/OpenHands#3727 (+ OpenHands/OpenHands#3728's relocation, done earlier in OpenHands/agent-canvas#1440). Correctness fixes from a deep review landed (edit-save field wipe; launch racing the profiles query). Remaining: merge main, fix the pre-existing mock-llm-e2e breakage on the branch, bump the pinned agent-server to a release containing PR OpenHands/OpenHands#3968, human review.
[...]

Issue #213, opened 2025-09-12
Proposal: Spec-only Tools + Explicit RunLoop (Local/Remote) with Remote Execution


xingyaoww:
# Proposal: Spec-only Tools + Explicit RunLoop (Local/Remote) with Remote Execution

## Why
- One set of Pydantic models for everything. Tools are **specs**; executors are **runtime-only**. No duplicated “local vs remote” definitions.
- Make the execution boundary explicit and swappable: `LocalRunLoop` vs `RemoteRunLoop`.
- Keep the user code nearly identical; “flip a switch” to run remotely.

---

### Local
```python
from openhands.sdk import LLM, Agent, Conversation, Message, TextContent
from openhands.tools import execute_bash_tool, str_replace_editor_tool, task_tracker_tool
from openhands.runtime import LocalRunLoop

llm = LLM(model="...", base_url="...", api_key=...)
tools = [
    execute_bash_tool.with_params(working_dir=os.getcwd()),
    str_replace_editor_tool,                                  # no params
    task_tracker_tool.with_params(save_dir=os.getcwd()),
]
agent = Agent(llm=llm, tools=tools)

conversation = Conversation(agent=agent)

with LocalRunLoop() as loop:           # executor materialization happens here
    conversation.start(loop)           # explicit, replaces implicit .run()
    conversation.send_message(
        Message(role="user", content=[TextContent(text="Hello!")])
    )
    conversation.run()                 # drives the loop until pause/finish
```

### Remote (same specs; only the runloop changes)
```python
from openhands.runtime import RemoteRunLoop

with RemoteRunLoop(
    base_url="https://openhands.example.com",
    api_key="***",
) as loop:
    conversation.start(loop)           # same Conversation/Agent/Tool objects
    conversation.send_message(
        Message(role="user", content=[TextContent(text="Do the thing")])
    )
    conversation.run()
```

> Key point: **No different models** for remote. We serialize the same Pydantic objects and create executors **only** on the remote runner.

---

### 1) Tools are specs; executors are runtime-only
- `Tool` remains a Pydantic model: `{ name, description, action_type, observation_type, annotations, meta, params }`
- New helper: `Tool.with_params(**kwargs)` → returns a new spec instance storing runtime parameters (e.g., `working_dir="/..."`).
- **No `.create()` in user code.**  
  `.create()` is called **only inside a RunLoop** (local or remote). This preserves a single definition while supporting multiple runtimes.

[...]

PR #3784, opened 2026-06-18
[AgentProfile][agent-server] agent_profile_id at conversation start + LaunchedProfile provenance

### LINKED ISSUE #3720: [AgentProfile][agent-server] Accept agent_profile_id at conversation start + stamp LaunchedProfile provenance
> **🔧 Audit reconciliation (2026-06-21):** scope refined to match the merged code. The provenance model shipped as **`LaunchedAgentProfile`** (not `LaunchedProfile`) and its field is **`launched_agent_profile`** (renamed in follow-up **PR #3788**). The shipped model is **`{ agent_profile_id, revision }` only** — the `snapshot` member described below was **dropped**; snapshot-at-start is instead guaranteed by persisting the fully-built `request.agent`. Core implementation (PR #3784, in v1.29.0) otherwise matches every acceptance criterion. *(For the record: PR #3784's "How to Test" snippet is stale/inverted — the SDK `agent_profile_id` field is NOT `exclude=True` and DOES appear in `model_dump(mode="json")` so it can reach the server over HTTP; server-side exclusion lives on `StoredConversation` (`models.py:86`) and the persisted request dump (`conversation_service.py:762`). See `tests/agent_server/test_agent_profile_conv_start.py:134`.)*

---

**Phase 2 — SDK agent-server · `software-agent-sdk` (`openhands-agent-server`) · size L · Parent epic #3713**

**Depends on:** [agent-server] routers + active pointer (S6); [sdk] resolve_agent_profile (S4)

## Context
A conversation can be started *by profile*. Resolution must happen **server-side** (it needs the file-backed stores + cipher), then feed the existing, unchanged agent-build path. The actual model→agent join point is the SDK validator `StartConversationRequest._populate_agent_from_settings` (`conversation/request.py:239`) — resolution plugs in one level **above** it.

## Scope
- Add `agent_profile_id: UUID | None` to `StartConversationRequest` (`conversation/request.py`) as a **mutually exclusive** alternative to `agent` / `agent_settings`; a `model_validator` enforces the XOR (it only *validates* exclusivity — it does not resolve, since the SDK model layer has no store/cipher access).
- In `conversation_service` (`_start_conversation` / `start_conversation`), when `agent_profile_id` is set: load the `AgentProfile` + LLM/MCP stores (cipher from config), call `resolve_agent_profile` → `AgentSettingsConfig`, and rewrite the request to carry the resolved `agent_settings` **before** `StoredConversation` is built. Route ACP creds + MCP env/header secrets into the conversation secret channel / `state.secret_registry`.
- **Provenance:** add `launched_agent_profile: LaunchedAgentProfile { agent_profile_id, revision }` to `StoredConversation`, persisted via the existing `event_service.save_meta()`, and project it onto `ConversationInfo` in `_compose_conversation_info` (the same mechanism ACP model state is lifted by). Snapshot-at-start is guaranteed by persisting the fully-built `request.agent` — **no separate redacted-snapshot field** is stored (it would only double the at-rest surface).

## Acceptance criteria
- Starting a conversation with `agent_profile_id` produces the same agent the resolved settings would, and rejects combining it with explicit `agent`/`agent_settings`.
- `meta.json` carries `launched_agent_profile`; the list + detail endpoints expose which profile + revision launched the conversation.
- Editing the referenced profile **after** start does not mutate the running conversation (snapshot-at-start, already guaranteed by persisting `request.agent`).

simonrosenberg:
HUMAN:

Agent profile ID resolution at conversation start: accepts a stable UUID, resolves server-side via the profile store and cipher, stamps LaunchedProfile provenance on the stored conversation, and maps resolution errors (ProfileNotFound → 404, DanglingMcpServerRef → 422).

---

AGENT:

20/20 new tests pass locally. Full agent-server suite: 1241 passed, 1 pre-existing flake (TestAutoTitle — confirmed failing on main before this branch).

```
$ uv run pytest tests/agent_server/test_agent_profile_conv_start.py -v
TestStartConversationRequestValidation::test_agent_profile_id_alone_is_valid PASSED
TestStartConversationRequestValidation::test_agent_alone_is_valid PASSED
TestStartConversationRequestValidation::test_agent_profile_id_and_agent_is_invalid PASSED
TestStartConversationRequestValidation::test_agent_profile_id_and_agent_settings_is_invalid PASSED
TestStartConversationRequestValidation::test_no_agent_source_is_invalid PASSED
TestStartConversationRequestValidation::test_agent_profile_id_is_excluded_from_model_dump PASSED
TestResolveAgentFromProfile::test_unknown_id_raises_profile_not_found PASSED
TestResolveAgentFromProfile::test_openhands_profile_resolves_to_agent_and_stamps_launched PASSED
TestResolveAgentFromProfile::test_dangling_mcp_server_ref_propagates PASSED
TestResolveAgentFromProfile::test_acp_profile_resolves_to_acp_agent PASSED
TestConversationServiceStartFromProfile::test_start_from_profile_stamps_launched_profile_on_stored PASSED
TestConversationServiceStartFromProfile::test_profile_not_found_propagates PASSED
TestConversationServiceStartFromProfile::test_dangling_ref_propagates_from_service PASSED
TestConversationRouterProfileErrors::test_profile_not_found_returns_404 PASSED
TestConversationRouterProfileErrors::test_dangling_mcp_server_ref_returns_422 PASSED
TestLaunchedProfileRoundTrip::test_launched_profile_survives_stored_conversation_round_trip PASSED
TestLaunchedProfileRoundTrip::test_stored_conversation_without_profile_has_none PASSED
TestLaunchedProfileRoundTrip::test_launched_profile_in_conversation_info PASSED
TestLaunchedProfileRoundTrip::test_conversation_info_without_profile_is_none PASSED
TestLaunchedProfileRoundTrip::test_launched_profile_survives_json_serialization PASSED
======================== 20 passed in 0.29s ========================

$ uv run ruff check <changed files>
All checks passed!
$ uv run ruff format --check <changed files>
All files already formatted.
```

## Why

ts-client `deriveSwitchPlan` (epic #3713) needs to know which agent profile launched a conversation without doing fragile settings-comparison. The resolution logic also needs cipher + store access that the SDK layer doesn't have.

## Summary

- `agent_profile_id: UUID | None` added to `StartConversationRequest` (carry-only `exclude=True`, mutually exclusive with `agent`/`agent_settings`; SDK enforces exclusivity only)
- Server-side `_resolve_agent_from_profile(profile_id, cipher)`: UUID→name via `list_summaries()`, load profile, pull global `mcp_config` from `PersistedSettings`, call `resolve_agent_profile()`, return built `AgentBase` + `LaunchedProfile(id, revision)`
- `_start_conversation` runs resolution via `asyncio.to_thread()`, injects resolved agent into request, stamps `launched_profile` on `StoredConversation`; `LaunchedProfile` projected onto `ConversationInfo`
- Router: `ProfileNotFound → 404`, `DanglingMcpServerRef → 422`

## Issue Number

Closes #3720

## How to Test

```python
from openhands.sdk.conversation.request import StartConversationRequest
from openhands.sdk.workspace import LocalWorkspace
from uuid import uuid4

# Mutual exclusivity enforced at SDK layer
req = StartConversationRequest(agent_profile_id=uuid4(), workspace=LocalWorkspace(working_dir="/tmp"))
assert req.agent is None  # server resolves before use

# agent_profile_id not serialized (carry-only)
assert "agent_profile_id" not in req.model_dump(mode="json")
```

End-to-end: start a conversation with `agent_profile_id` set to an existing profile UUID via `POST /api/conversations`; `GET /api/conversations/{id}` should show `launched_profile.profile_id` and `launched_profile.revision` in the response.

## Video/Screenshots

(unit tests cover all paths; no UI changes)

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- ACP provider creds ride the existing `request.secrets` channel — resolver is untouched.
- `_resolve_agent_from_profile` uses lazy imports to avoid circular imports between `agent_server` and `sdk.profiles`; test patch targets are the source modules (e.g. `openhands.sdk.profiles.agent_profile_store.AgentProfileStore`), not the service module.
- Stacked on #3780 (resolver) + #3781 (router) which are merged on main.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:7726004-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-7726004-python \
  ghcr.io/openhands/agent-server:7726004-python
```

```
ghcr.io/openhands/agent-server:7726004-golang-amd64
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-golang-amd64
ghcr.io/openhands/agent-server:agent-profile-conv-start-golang-amd64
ghcr.io/openhands/agent-server:7726004-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:7726004-golang-arm64
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-golang-arm64
ghcr.io/openhands/agent-server:agent-profile-conv-start-golang-arm64
ghcr.io/openhands/agent-server:7726004-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:7726004-java-amd64
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-java-amd64
ghcr.io/openhands/agent-server:agent-profile-conv-start-java-amd64
ghcr.io/openhands/agent-server:7726004-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:7726004-java-arm64
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-java-arm64
ghcr.io/openhands/agent-server:agent-profile-conv-start-java-arm64
ghcr.io/openhands/agent-server:7726004-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:7726004-python-amd64
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-python-amd64
ghcr.io/openhands/agent-server:agent-profile-conv-start-python-amd64
ghcr.io/openhands/agent-server:7726004-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:7726004-python-arm64
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-python-arm64
ghcr.io/openhands/agent-server:agent-profile-conv-start-python-arm64
ghcr.io/openhands/agent-server:7726004-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:7726004-golang
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-golang
ghcr.io/openhands/agent-server:agent-profile-conv-start-golang
ghcr.io/openhands/agent-server:7726004-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:7726004-java
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-java
ghcr.io/openhands/agent-server:agent-profile-conv-start-java
ghcr.io/openhands/agent-server:7726004-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:7726004-python
ghcr.io/openhands/agent-server:7726004067ab9b2b20536b4951f20c6bdd15fdbf-python
ghcr.io/openhands/agent-server:agent-profile-conv-start-python
ghcr.io/openhands/agent-server:7726004-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `7726004-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `7726004-python-amd64`) are also available if needed


VascoSch92 (commented):
Profile-only StartConversationRequest can't be serialized ` model_dump()/model_dump_json()` raise `PydanticSerializationError`.

The agent field is typed non-optional AgentBase but defaults to None, and AgentBase's @model_serializer (base.py:341) calls self.model_dump() which Pydantic invokes even when the value is None. Before this PR, _require_agent guaranteed agent was never None, so the path was unreachable; making agent=None valid (profile-only) now exposes it.

This means the PR's own "How to Test" snippet (req.model_dump(mode="json")) crashes. The live feature isn't broken — the TS client posts raw JSON and the server resolves agent before its own model_dump (conversation_service.py:687 runs before :751) — but it's a real footgun for any Python caller, and the "carry-only" framing implies dump should work. Note test_agent_profile_id_is_excluded_from_model_dump only dumps an agent request, so it hides this.

Suggest guarding the AgentBase serializer against a None/non-agent self (or omitting agent when None), plus a test that dumps a profile-only request.

openhands-sdk/openhands/sdk/conversation/request.py
```diff
@@ -242,26 +253,38 @@ def _populate_agent_from_settings(cls, data: Any) -> Any:
...
+                if "kind" not in agent_payload and "llm" in agent_payload:
+                    agent_payload["kind"] = "Agent"
+                payload["agent"] = agent_payload
         return payload
 
     @model_validator(mode="after")
     def _require_agent(self) -> StartConversationRequest:
-        if self.agent is None:
-            raise ValueError("Either `agent` or `agent_settings` must be provided")
+        if self.agent is None and self.agent_profile_id is None:
```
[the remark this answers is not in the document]
simonrosenberg: Fixed in 474950f7f: added `@field_serializer("agent", mode="wrap")` on `StartConversationRequest` that returns `None` directly when `value is None`, short-circuiting before `AgentBase._serialize_with_mcp_handling` fires. The field is now serializable for profile-only requests.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -240,6 +241,66 @@ def _prepare_request_workspace(
...
+        raise ProfileNotFound(
+            f"Agent profile '{profile_name}' (id={profile_id}) not found"
+        )
+    except ValueError as exc:
+        raise ValueError(
+            f"Failed to load agent profile '{profile_name}': {exc}"
+        ) from exc
+
+    # Global MCP config — already decrypted by the FileSettingsStore cipher context.
+    settings_store = get_settings_store()
```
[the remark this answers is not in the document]
simonrosenberg: Fixed in 5b3e6dd83: introduced `_load_mcp_config(conversations_dir, cipher)` which builds a short-lived `FileSettingsStore` from the `ConversationService`'s own `conversations_dir` and `cipher` — mirroring the same persistence-dir derivation logic as `_get_persistence_dir()` in the store module. `_resolve_agent_from_profile` now accepts `mcp_config` as an explicit parameter and no longer touches the singleton at all.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -240,6 +241,95 @@ def _prepare_request_workspace(
...
+        raise ValueError(
+            f"Failed to load agent profile '{profile_name}': {exc}"
+        ) from exc
+
+    llm_store = LLMProfileStore()
+    try:
+        settings_config = resolve_agent_profile(
+            profile, llm_store=llm_store, mcp_config=mcp_config, cipher=cipher
+        )
+    except _DanglingMcpServerRef:
```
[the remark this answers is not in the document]
simonrosenberg: Done in 9074b800f: removed both no-op `except … raise` arms. `ProfileNotFound` and `DanglingMcpServerRef` propagate naturally past the `(TypeError, ValueError)` handler.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -240,6 +241,95 @@ def _prepare_request_workspace(
...
+    and builds a short-lived ``FileSettingsStore`` instance with the caller-supplied
+    cipher.  This avoids touching the singleton — which may not yet have been
+    initialised with the correct cipher if this is the first settings-store access
+    after boot.
+    """
+    import os
+
+    from openhands.agent_server.persistence import FileSettingsStore, PersistedSettings
+
+    env_dir = os.environ.get("OH_PERSISTENCE_DIR")
```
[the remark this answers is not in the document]
simonrosenberg: Done in 9074b800f: dropped `_load_mcp_config` and its duplicated derivation. `ConversationService.get_instance(config)` now calls `get_settings_store(config)` (one-time, before any request) to initialise the singleton with the correct cipher. `_start_conversation` then calls `get_settings_store()` without config — safe because the singleton is already initialised — so there's no derivation to duplicate.

simonrosenberg:
Fixed in 7726004: removed `exclude=True` from `agent_profile_id` on `StartConversationRequest` so `model_dump()` includes it for HTTP transport. The persistence-only exclusion is now on `StoredConversation` (which overrides the field with `exclude=True`) and enforced explicitly via `exclude={'agent_profile_id'}` when building `request_data`.

openhands-sdk/openhands/sdk/conversation/request.py
```diff
@@ -234,6 +241,17 @@ class StartConversationRequest(BaseModel):
             "used to construct the concrete agent."
         ),
     )
+    agent_profile_id: UUID | None = Field(
+        default=None,
+        exclude=True,
```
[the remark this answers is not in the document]
simonrosenberg: Fixed in 7726004: removed `exclude=True` from `StartConversationRequest.agent_profile_id`, override the field with `exclude=True` on `StoredConversation` instead, and explicitly exclude it from `request_data` before constructing `StoredConversation`. Added regression tests covering both the transport path (`StartConversationRequest.model_dump()` includes the UUID) and the persistence path (`StoredConversation.model_dump()` excludes it).

simonrosenberg:
@VascoSch92 good catch — fixed in 474950f7. Added a `@field_serializer("agent", mode="wrap")` on `StartConversationRequest` that short-circuits before `AgentBase._serialize_with_mcp_handling` when `agent` is `None`, so `model_dump()` on a profile-only request returns `{"agent": null, ...}` without hitting the serializer. The new test `test_agent_profile_id_present_in_request_payload` now dumps a profile-only request (no agent) and asserts the field is present — it would have caught the crash before.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -612,6 +667,28 @@ async def _start_conversation(
...
+            # singleton with the server cipher before any conversation can start.
+            from openhands.agent_server.persistence import (
+                PersistedSettings,
+                get_settings_store,
+            )
+
+            settings = get_settings_store().load() or PersistedSettings()
+            mcp_config = settings.agent_settings.mcp_config
+            resolved_agent, launched_profile = await asyncio.to_thread(
+                _resolve_agent_from_profile,
```
enyst, line 685: @openhands I’m a bit surprised by `asyncio.to_thread()` here: why not just make the function async? Just answer me here on github, make recommendations if any, but do not make modifications (PR was merged anyway) - think with me.

enyst: Good question. I wouldn't make the current implementation `async` unless the underlying profile/settings stores become async too.

The resolver is mostly synchronous, file-backed work: profile id→name lookup/load under `FileLock`, LLM profile load, settings/MCP data access, JSON/Pydantic validation, cipher handling, and finally `create_agent()`. If we only changed the function to `async def` while keeping those same operations inside it, it would still run on the FastAPI event loop until completion and would just be "async in name". The `asyncio.to_thread()` call is the explicit boundary saying: this is sync persistence/validation work, so don't block the event loop during conversation start.

My recommendation if we revisited it:

1. Make the boundary clearer with naming/structure, e.g. `_resolve_agent_from_profile_sync(...)` plus a tiny async wrapper like `_aresolve_agent_from_profile(...)` that does the `to_thread()` call. That keeps service code tidy without pretending the resolver itself has native async I/O.
2. Keep all sync store work inside that worker-thread boundary. In particular, settings-store/profile-store/LLM-store reads all have the same event-loop-blocking concern, so we should avoid pulling any of those reads back out into the async path.
3. Only make the resolver genuinely async if/when the stores themselves expose async APIs (async file locks / async file I/O / async store methods). At that point, an end-to-end async resolver would be the cleaner design.

So I see `to_thread()` as the right primitive here for event-loop health, but the code could communicate that intent better with a sync/async wrapper split.

_AI-generated reply from OpenHands on behalf of the PR author._

openhands-agent-server/openhands/agent_server/models.py
```diff
@@ -78,13 +79,26 @@ class StoredConversation(StartConversationRequest):
...
+    agent_profile_id: UUID | None = Field(default=None, exclude=True)
+
     id: OpenHandsUUID
     title: str | None = Field(
         default=None, description="User-defined title for the conversation"
     )
     metrics: MetricsSnapshot | None = None
     created_at: datetime = Field(default_factory=utc_now)
     updated_at: datetime = Field(default_factory=utc_now)
+    launched_profile: LaunchedProfile | None = Field(
```
enyst, line 93: @openhands We now have two kinds of profiles: LLM Profiles and Agent Profiles. I think we could try to look at all occurrences in the codebase, and name them appropriately where it could be confusing. Maybe in some contexts it’s clear? But in APIs we should probably use names like agent_profile for the agent profile, llm_profile for the LLM profile. Analyze and propose a PR if you feel there are potentially confusing occurrences.

(do not modify this PR, it’s merged, make a new one from main)

enyst: Addressed in follow-up PR #3788: https://github.com/OpenHands/software-agent-sdk/pull/3788

I audited the profile naming and kept the existing legacy LLM-profile surfaces unchanged, but renamed the newly added conversation provenance API/model from `LaunchedProfile`/`launched_profile.profile_id` to `LaunchedAgentProfile`/`launched_agent_profile.agent_profile_id`, with validation aliases for the short-lived merged payload shape.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/profiles/agent_profile.py
```diff
@@ -209,6 +209,23 @@ class ACPAgentProfile(AgentProfileBase):
     )
 
 
+class LaunchedProfile(BaseModel):
+    """Provenance snapshot recorded when a profile launches a conversation.
+
+    Stored on ``StoredConversation`` and projected onto ``ConversationInfo`` so
+    ts-client ``deriveSwitchPlan`` can identify which profile is current without
```
enyst, line 216: @openhands What is ts-client deriveSwitchPlan? Search in OpenHands org, all public repos. Is it in the typescript-client repo? Link on github here, in a reply.

tests/agent_server/test_agent_profile_conv_start.py
```diff
@@ -0,0 +1,491 @@
...
+        acp_server="claude-code",
+    )
+
+
+def _make_agent() -> Agent:
+    return Agent(llm=LLM(model="gpt-4o", usage_id="llm"), tools=[])
+
+
+# ---------------------------------------------------------------------------
+# SDK-layer: mutual exclusivity (StartConversationRequest)
```
enyst, line 90: @openhands Understand these tests - what is this mutual exclusivity about? Reply directly me here.

openhands-sdk/openhands/sdk/profiles/agent_profile.py
enyst: Found it. In public OpenHands repos, `deriveSwitchPlan` is in OpenHands/typescript-client PR #213 (`[AgentProfile][ts-client] AgentProfile types + client + deriveSwitchPlan`), on branch `agent-profile-client`; it is not on `typescript-client` main yet.

Implementation link: https://github.com/OpenHands/typescript-client/blob/3b783c65a567e6075f5490110f44217950440a9f/src/profiles/derive-switch-plan.ts#L60-L64
PR link: https://github.com/OpenHands/typescript-client/pull/213

It is a pure helper that decides whether switching from the conversation's launched profile snapshot to a target profile is `current`, can `switch-live`, must `start-new`, or is `disabled`. That is what I meant by `ts-client deriveSwitchPlan`.

_This reply was generated by OpenHands on behalf of the user._

tests/agent_server/test_agent_profile_conv_start.py
enyst: Here “mutual exclusivity” means the conversation-start request should have exactly one source for the agent it will run.

Before this change, the request could be started from an explicit built `agent` or from `agent_settings` (which the SDK validator converts into an agent). This PR adds a third source: `agent_profile_id`, which is only a pointer; the agent-server has to resolve it using the profile store/cipher before creating the stored conversation.

If a caller sent `agent_profile_id` together with `agent` or `agent_settings`, we would have two competing definitions: should the server run the explicit agent/settings, or resolve the profile? Worse, we could stamp `launched_profile` provenance for one profile while actually running a different explicit agent. So the validator rejects those ambiguous combinations.

The tests are checking that these cases are allowed/rejected:

- `agent_profile_id` alone: valid; server resolves it later.
- `agent` alone: valid; existing explicit-agent path.
- `agent_profile_id + agent`: invalid.
- `agent_profile_id + agent_settings`: invalid.
- no agent source at all: invalid.

Small nuance: this is primarily “profile pointer vs explicit agent source” exclusivity. The profile resolver can itself reference an LLM profile internally, but that’s a different concept and not what these tests are about.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

PR #3786, opened 2026-06-18
fix(sdk): redact plugin and extension source credentials at the serialization boundary (#3752)

### LINKED ISSUE #3752: Redact plugin/extension source credentials at a single serialization choke point (follow-up to #2196 review, sibling of OpenHands#12959)
## Summary

Credential-bearing plugin/extension source URLs (e.g. `https://oauth2:TOKEN@github.com/org/repo.git`) still reach **persisted state** and **logs** through the plugin/extension *source* path. This is the SDK-side sibling of OpenHands/OpenHands#12959 (fixed there by [OpenHands/OpenHands#14795](<https://github.com/OpenHands/OpenHands/issues/14795>), which added a `@field_serializer('source')` to that repo's `PluginSpec`) — the equivalent choke point was never added to the SDK's own `PluginSource` model.

This issue tracks closing the plugin-source leaks specifically. The git/workspace command-layer leaks are tracked separately in OpenHands/software-agent-sdk#3751, and `GitCommandError.command` in [#3689](<https://github.com/OpenHands/software-agent-sdk/issues/3689>).

## Background / why a "choke point"

The earlier PR [#2196](<https://github.com/OpenHands/software-agent-sdk/issues/2196>) attempted per-call-site scrubbing of plugin source URLs. In review ([#2196 review 3950231707](<https://github.com/OpenHands/software-agent-sdk/pull/2196#pullrequestreview-3950231707>)) this was rejected as architecturally incomplete:

> "You added a scrubber, but you did not make it the single choke point, so raw URLs are still escaping through ordinary success-path code."

That review is the basis for this issue. The durable fix is one redaction boundary on the model (a serializer), plus explicit redaction at the log sites that bypass serialization.

### 1\. `PluginSource` has no serializer choke point

`openhands-sdk/openhands/sdk/plugin/types.py:15` (model) / `:35` (`source` field)

The base `PluginSource` model (and anything inheriting it) serializes `source` verbatim. Redaction currently exists only as a manual call inside the `ResolvedPluginSource.from_plugin_source` factory (`types.py:165`) — so any other `model_dump()` / `model_dump_json()` of a plugin source emits raw credentials.

**Repro**

```python
from openhands.sdk.plugin.types import PluginSource
s = PluginSource(source="https://oauth2:SECRET@github.com/org/repo.git")
assert "SECRET" not in s.model_dump_json()  # FAILS on main
```

**Fix:** add `@field_serializer('source')` on `PluginSource` that applies `redact_url_credentials()` at serialization time, so every subclass and dump path is covered by construction. Keep the raw value in memory for fetch/clone. (This makes the manual redaction at `types.py:165` redundant — consolidate.)

### 2\. Persisted-state leak: `InstallationInfo.source` written to `.installed.json`

`openhands-sdk/openhands/sdk/extensions/installation/info.py:60`

```python
return InstallationInfo(
    ...
    source=source,   # raw; persisted to `.installed.json` per the class docstring
    ...
)
```

A credentialed source is written to disk in the install metadata. (This is the relocated form of the persisted-state leak flagged in the [#2196](<https://github.com/OpenHands/software-agent-sdk/issues/2196>) review — it moved from `plugin/installed.py` into the generic `extensions/installation/` framework during the refactor.)

**Repro**

```python
# via InstallationInfo.from_extension(..., source="https://oauth2:SECRET@host/r.git")
# the resulting .installed.json / model_dump contains "SECRET"
```

**Fix:** redact `source` before persisting (either rely on the model serializer if `InstallationInfo.source` is given the same treatment, or redact explicitly here). Note: `InstallationInfo` is a generic extensions model rather than a `PluginSource`, so a serializer here would redact all extension sources — likely desirable; implementer's call.

### 3\. Success-path log leak: plugin-load debug log

`openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py:660`

```python
logger.debug(
    f"Loaded plugin '{plugin.manifest.name}' from {spec.source}"
    + (f" @ {resolved_ref[:8]}" if resolved_ref else "")
)
```

Raw `spec.source` on the normal (success) path. Lower severity (`debug`-level), but a serializer does **not** cover f-string logging — `__format__` bypasses it — so this needs explicit redaction at the call site.

**Fix:** wrap with `redact_url_credentials(spec.source)`.

## Acceptance criteria

- [ ] `PluginSource` redacts `source` via a `@field_serializer` so `model_dump`/`model_dump_json` never emit embedded credentials (covers all subclasses).
- [ ] The manual redaction in `ResolvedPluginSource.from_plugin_source` is consolidated onto the serializer (single source of truth).
- [ ] `InstallationInfo.source` is redacted before being persisted to `.installed.json`.
- [ ] `local_conversation.py:660` redacts `spec.source` before logging.
- [ ] A quick audit confirms no other plugin/extension-source log or persistence site emits raw credentials (e.g. `ResolvedPluginSource` usages, any remaining `installed`/`loader` paths).
- [ ] Tests exercise the **live paths** (serialize a `PluginSource`, persist an `InstallationInfo`, plugin-load log) — not just the redaction regex — per the [#2196](<https://github.com/OpenHands/software-agent-sdk/issues/2196>) review's testing-gap note.

## Out of scope

* Git/workspace command-layer redaction (`repo.py:149`, `run_git_command` stderr, `_mask_token` consolidation) — tracked in OpenHands/software-agent-sdk#3751.
* `GitCommandError.command` field redaction — tracked in [#3689](<https://github.com/OpenHands/software-agent-sdk/issues/3689>).

## Related

* Supersedes the plugin-source portion of [#2196](<https://github.com/OpenHands/software-agent-sdk/issues/2196>) (and its review [3950231707](<https://github.com/OpenHands/software-agent-sdk/pull/2196#pullrequestreview-3950231707>)).
* SDK sibling of OpenHands/OpenHands#12959 (fixed by [OpenHands/OpenHands#14795](<https://github.com/OpenHands/OpenHands/issues/14795>)).
* Original SDK report: OpenHands/software-agent-sdk#2152. Companion follow-ups: [#3689](<https://github.com/OpenHands/software-agent-sdk/issues/3689>), OpenHands/software-agent-sdk#3751.

---

*Filed by an AI agent (OpenHands) on behalf of @jpshackelford.*

haya-asif:
HUMAN:

This PR fixes a credential leak where plugin/extension source URLs containing embedded tokens (e.g.https://oauth2:TOKEN@host/repo.git) were written verbatim to persisted state (.installed.json, conversation state) and logs. The fix redacts them at the model serialization boundary while keeping the raw value in memory for fetching.

---

AGENT:

## Why

Closes #3752. Credential-bearing plugin/extension `source` URLs (e.g.
`https://oauth2:TOKEN@host/repo.git`) leaked into persisted state
(`.installed.json`, `meta.json`, conversation state, the auto `plugins` tag) and
logs. The first attempt hand-rolled redaction and broke private-plugin clones;
this version routes the `source` field through the SDK's existing **secret
convention** instead.

## Summary

- New `serialize_credential_url` / `validate_credential_url` in
  `utils/pydantic_secrets.py` (beside `serialize_secret` / `validate_secret`),
  branching on `resolve_expose_mode`: **redact** (default → mask only the
  userinfo, `https://****@host`), **plaintext** (trusted dumps → real URL),
  **encrypted** (a `cipher` → Fernet token, decrypted on load).
- `PluginSource.source` and `InstallationInfo.source` wire these via the
  standard `field_validator(mode="before")` + `field_serializer(when_used="always")` pair.
- `ResolvedPluginSource` keeps redacting at construction (unchanged from `main`).
- `RemoteConversation` sends plugins with `expose_secrets=True`; the auto
  `plugins` tag is redacted; `.installed.json` is persisted with an expose
  context so `ExtensionManager.update()` can re-clone.

Maps to the review:
- **(1)** agent-server clones private plugins — the server now receives the real
  source (and `${VAR}`), and `meta.json` is **encrypted** at rest, not cleartext.
- **(2)** `ExtensionManager.update()` re-clones from `.installed.json` (round-trip preserved).
- **(3)** `ResolvedPluginSource.source` redacted in memory (attribute read is safe).

## Issue Number

Closes #3752.

## How to Test

```bash
uv run pytest tests/sdk/git/test_url_redaction.py tests/sdk/extensions/installation \
  tests/workspace/test_cloud_workspace_automation_tags.py
uv run ruff check . && uv run pyright
```

New regression tests cover each path: default redaction, expose, the cipher
encrypt-at-rest round-trip on `PluginSource` and `InstallationInfo` (the
credential becomes a `gAAAAA…` token and decrypts back), the remote-send
payload, the `plugins` tag, and the `.installed.json` → `update()` re-clone.
Full `tests/sdk` + `tests/agent_server` green; ruff + pyright clean.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

`source` uses a bespoke `serialize_credential_url` rather than `SecretStr`
because it is a polymorphic URL (`github:owner/repo`, local path, `${VAR}`)
needing structure-preserving (userinfo-only) redaction, not a full `**********`
mask. `.installed.json` stores the source in plaintext at rest (the extensions
layer has no cipher) — same posture as before this PR, and required for `update()`.

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/conversation.py
@@ -20,6 +20,7 @@ from openhands.sdk.logger import get_logger
 from openhands.sdk.plugin import PluginSource
 from openhands.sdk.secret import SecretValue
 from openhands.sdk.tool.client_tool import ClientToolSpec
+from openhands.sdk.utils.redact import redact_url_credentials
 from openhands.sdk.workspace import LocalWorkspace, RemoteWorkspace
 
 
@@ -171,7 +172,12 @@ class Conversation:
 
             # 2. Auto-generate plugins/skills tag from plugins parameter
             if plugins:
-                plugin_urls = [p.source_url for p in plugins if p.source_url]
+                # tags are a plain dict persisted verbatim — redact creds here.
+                plugin_urls = [
+                    redact_url_credentials(url)
+                    for p in plugins
+                    if (url := p.source_url)
+                ]
                 if plugin_urls:
                     effective_tags["plugins"] = ",".join(plugin_urls)
                     logger.debug(f"Added plugins tag with {len(plugin_urls)} plugin(s)")
--- a/tests/workspace/test_cloud_workspace_automation_tags.py
+++ b/tests/workspace/test_cloud_workspace_automation_tags.py
@@ -2,9 +2,13 @@
 
 import json
 import os
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
+
+from openhands.sdk.conversation.conversation import Conversation
+from openhands.sdk.plugin import PluginSource
+from openhands.sdk.workspace import RemoteWorkspace
 
 
 class TestDefaultConversationTags:
@@ -376,13 +380,7 @@ class TestPluginsTagInConversation:
             assert "https://github.com/OpenHands/review-skill" in plugin_urls
 
     def test_credentials_redacted_in_plugins_tag(self):
-        """A credential in a plugin source must not reach the persisted tag (B3)."""
-        from unittest.mock import MagicMock
-
-        from openhands.sdk.conversation.conversation import Conversation
-        from openhands.sdk.plugin import PluginSource
-        from openhands.sdk.workspace import RemoteWorkspace
-
+        """A credential in a plugin source must not reach the persisted tag."""
         mock_workspace = MagicMock(spec=RemoteWorkspace)
         mock_workspace.default_conversation_tags = {}
 
--- a/tests/workspace/test_cloud_workspace_automation_tags.py
+++ b/tests/workspace/test_cloud_workspace_automation_tags.py
@@ -375,6 +375,30 @@ class TestPluginsTagInConversation:
             )
             assert "https://github.com/OpenHands/review-skill" in plugin_urls
 
+    def test_credentials_redacted_in_plugins_tag(self):
+        """A credential in a plugin source must not reach the persisted tag (B3)."""
+        from unittest.mock import MagicMock
+
+        from openhands.sdk.conversation.conversation import Conversation
+        from openhands.sdk.plugin import PluginSource
+        from openhands.sdk.workspace import RemoteWorkspace
+
+        mock_workspace = MagicMock(spec=RemoteWorkspace)
+        mock_workspace.default_conversation_tags = {}
+
+        plugins = [
+            PluginSource(source="https://oauth2:SUPER_SECRET@github.com/org/repo.git")
+        ]
+        with patch(
+            "openhands.sdk.conversation.impl.remote_conversation.RemoteConversation"
+        ) as mock_convo_class:
+            mock_convo_class.return_value = MagicMock()
+            Conversation(agent=MagicMock(), workspace=mock_workspace, plugins=plugins)
+            tag = mock_convo_class.call_args.kwargs["tags"]["plugins"]
+
+        assert "SUPER_SECRET" not in tag
+        assert tag == "https://****@github.com/org/repo.git"
+
     def test_local_plugins_not_included_in_tags(self):
         """Should not include local path plugins in tags."""
         from unittest.mock import MagicMock
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -785,8 +785,15 @@ class RemoteConversation(BaseConversation):
                 "tool_module_qualnames": tool_qualnames,
                 # Include agent definitions for subagent registration on server
                 "agent_definitions": serialized_defs,
-                # Include plugins to load on server
-                "plugins": [p.model_dump() for p in plugins] if plugins else None,
+                # expose_secrets: send the real source so the server can clone.
+                "plugins": (
+                    [
+                        p.model_dump(mode="json", context={"expose_secrets": True})
+                        for p in plugins
+                    ]
+                    if plugins
+                    else None
+                ),
                 # Include hook_config for server-side hooks
                 "hook_config": hook_config.model_dump() if hook_config else None,
                 # Include client-defined tool specs (no server-side executor)
--- a/openhands-sdk/openhands/sdk/extensions/installation/info.py
+++ b/openhands-sdk/openhands/sdk/extensions/installation/info.py
@@ -2,10 +2,22 @@ from __future__ import annotations
 
 from datetime import UTC, datetime
 from pathlib import Path
+from typing import Any
 
-from pydantic import BaseModel, Field
+from pydantic import (
+    BaseModel,
+    Field,
+    FieldSerializationInfo,
+    ValidationInfo,
+    field_serializer,
+    field_validator,
+)
 
 from openhands.sdk.extensions.installation.interface import ExtensionProtocol
+from openhands.sdk.utils.pydantic_secrets import (
+    serialize_credential_url,
+    validate_credential_url,
+)
 
 
 class InstallationInfo(BaseModel):
@@ -35,6 +47,17 @@ class InstallationInfo(BaseModel):
         description="ISO 8601 timestamp of installation",
     )
     install_path: Path = Field(description="Path where the extension is installed")
+
+    @field_validator("source", mode="before")
+    @classmethod
+    def _decrypt_source(cls, v: Any, info: ValidationInfo) -> Any:
+        """Decrypt at-rest ciphertext (see validate_credential_url)."""
+        return validate_credential_url(v, info)
+
+    @field_serializer("source", when_used="always")
+    def _serialize_source(self, source: str, info: FieldSerializationInfo) -> str:
+        """Redact/expose/encrypt by context (see serialize_credential_url)."""
+        return serialize_credential_url(source, info)
 
     @staticmethod
     def from_extension(
--- a/openhands-sdk/openhands/sdk/extensions/installation/metadata.py
+++ b/openhands-sdk/openhands/sdk/extensions/installation/metadata.py
@@ -156,12 +156,7 @@ class InstallationMetadata(BaseModel):
             return cls()
 
     def save_to_dir(self, installed_dir: Path) -> None:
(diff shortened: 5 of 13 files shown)


WaseemSabir (approved):
lgtm!

VascoSch92 (changes requested):
Hey @haya-asif 

thanks for the PR. 

However, there are a couple of things to fix.

Perhaps, take a moment to explore a little the repo and understand how secrets are handled.

---

### Agent-server clones private plugins with `https://****@… `(first run and resume)

`StoredConversation` extends `StartConversationRequest` (`plugins: list[PluginSource], request.py:171`). The agent-server builds it with r`equest.model_dump(mode="json", context={"expose_secrets": True})` (conversation_service.py:761), persists it via `stored.model_dump_json()` to `meta.json` (`event_service.py:126`), and later clones from `stored.plugins` (`event_service.py:790` → _ensure_plugins_loaded). 

The new serializer fires on every one of those dumps and ignores expose_secrets — the exact flag the server passes to keep values intact. Worse, it redacts `${MY_TOKEN} ` placeholders, which are not secrets (they're expanded from the secret registry at fetch time) and never leaked. So `_expand_secret_refs` has nothing to expand and git clones https://****@… → 403.

```python
from openhands.sdk.plugin.types import PluginSource

  # ${MY_TOKEN} is the recommended way to clone a private plugin: it is NOT a
  # secret and is expanded from the secret registry at fetch time.
  src = "https://x-token-auth:${MY_TOKEN}@host/repo.git"
  
  # The agent-server builds StoredConversation via this exact call
  # (conversation_service.py:761) and later clones from stored.plugins.
  print(PluginSource(source=src).model_dump(mode="json", context={"expose_secrets": True})["source"])

  # PR #3786 → 'https://****@host/repo.git'   (placeholder gone → git clone 403s)
  # main     → 'https://x-token-auth:${MY_TOKEN}@host/repo.git'  (works)
```

---

### `ExtensionManager.update()` re-clones private extensions with https://****@…

`update()` reloads the source from `.installed.json` and re-installs with it (`manager.py:312-325`): `current_info = self.get(name)` (which reads disk via InstallationMetadata.load_from_dir) then `self.install(source=current_info.source, …)`. 

After the PR, the persisted source is redacted, and extensions don't support `${VAR}` expansion, so the re-fetch clones `https://****@…` and fails.

```python
  from pathlib import Path
  from openhands.sdk.extensions.installation.info import InstallationInfo

  src = "https://oauth2:SUPER_SECRET@github.com/org/ext.git"
  info = InstallationInfo(name="ext", source=src, install_path=Path("/tmp/ext"))

  # This value is what lands in .installed.json, is reloaded by manager.get(),
  # and is passed straight to install() inside update() (manager.py:312-325).
  print(info.model_dump()["source"])
  
  # PR #3786 → 'https://****@github.com/org/ext.git'   (update() re-clones this → 403)
  # main     → 'https://oauth2:SUPER_SECRET@github.com/org/ext.git'  (works)
```

  ---

### `ResolvedPluginSource.source` is now raw in memory → new attribute-access leak

the PR moved redaction from construction-time to the serializer, changing `from_plugin_source` from `source=redact_url_credentials(plugin_source.source)` to `source=plugin_source.source`. 

The model now holds the raw URL and only redacts on `model_dump`. Any direct attribute read therefore leaks, exactly the f-string/attribute bypass the issue warned about. 

The PR's audit missed `examples/05_skills_and_plugins/02_loading_plugins/main.py:114`, which prints resolved.source directly. On main this printed ****.

```python
  from openhands.sdk.plugin.types import PluginSource, ResolvedPluginSource

  resolved = ResolvedPluginSource.from_plugin_source(
      PluginSource(source="https://oauth2:SUPER_SECRET@github.com/org/repo.git"),
      resolved_ref="deadbeef",  
  )

  # Any direct attribute read leaks. Real sink:
  # examples/05_skills_and_plugins/02_loading_plugins/main.py:114
  #   print(f"  - {resolved.source} @ {resolved.resolved_ref}")
  print(resolved.source)
  
  # PR #3786 → 'https://oauth2:SUPER_SECRET@github.com/org/repo.git'  (raw → leaked)
  # main     → 'https://****@github.com/org/repo.git'                 (redacted, safe)
  ```

**Comment by haya-asif** [comment]
Would appreciate another look @VascoSch92 — thanks for the detailed review.

File: tests/workspace/test_cloud_workspace_automation_tags.py [hunk 1]
```diff
@@ -375,6 +375,30 @@ def test_plugins_added_as_urls_in_tags(self):
             )
             assert "https://github.com/OpenHands/review-skill" in plugin_urls
 
+    def test_credentials_redacted_in_plugins_tag(self):
+        """A credential in a plugin source must not reach the persisted tag (B3)."""
+        from unittest.mock import MagicMock
```
VascoSch92: import top of the file

openhands-sdk/openhands/sdk/conversation/conversation.py
```diff
@@ -171,7 +172,12 @@ def __new__(
 
             # 2. Auto-generate plugins/skills tag from plugins parameter
             if plugins:
-                plugin_urls = [p.source_url for p in plugins if p.source_url]
+                # tags are a plain dict persisted verbatim — redact creds here.
+                plugin_urls = [
+                    redact_url_credentials(url)
+                    for p in plugins
+                    if (url := p.source_url)
```
VascoSch92, line 179: cool... I like walrus operator.

VascoSch92 (commented):
For me is (almost) good.

There are just two things that we should tests to be sure we don't have regression in the future:

1. `ExtensionManager.update()` re-clones a credentialed source. This is the exact regression-#2 path and the sole justification for keeping `.installed.json` plaintext. Right now nothing proves the credential actually reaches the clone; a silent break here reintroduces the private-extension failure. 
2. Plugin-load log regression guard, issue #3752 explicitly required it, and a serializer can't catch f-string/attribute leaks. Without it, anyone re-adding `spec.source` to that log silently reopens the credential leak the PR exists to close.

Otherwise, @jpshackelford and @enyst WDYT?

simonrosenberg:
Since the PR was stale I took the liberty of resolving the issue here https://github.com/OpenHands/software-agent-sdk/pull/3786

Thanks @haya-asif for initiating the effort!

Issue #3610, opened 2026-06-09
[Prompt registry] Phase 2 — Port sections behind the snapshot


VascoSch92:
Part of the prompt-registry roadmap OpenHands/software-agent-sdk#3606 (proposal OpenHands/software-agent-sdk#2827).

**Goal:** reproduce today's rendered prompt entirely from registered sections, byte-for-byte, behind the Phase 0 snapshot. Lands as several PRs (one per section group), each keeping the snapshot green.

**Depends on:** Phase 0 OpenHands/software-agent-sdk#3607 (snapshot oracle) and Phase 1 OpenHands/software-agent-sdk#3608 (typed core).<br>**Blocks:** Phase 3.

### Tasks

- [x] **Port static-tier sections** of `agent/prompts/system_prompt.j2` into section classes + a `default_registry.py`. The \~17 blocks: Role, Memory, Efficiency, FileSystem, CodeQuality, VersionControl, PullRequests, ProblemSolving, SelfDocumentation, Security (`security_policy_filename` include, guarded), SecurityRiskAssessment (guarded by `llm_security_analyzer`), Browser (guarded by `enable_browser`), ExternalServices, EnvironmentSetup, Troubleshooting, ProcessManagement, ModelSpecific (the `model_specific/*.j2` include — @juanmichelini's "model-specific layer"). Keep large leaf bodies in Jinja; the section owns the guard + cache tier. Each PR keeps the Phase 0 snapshot byte-identical.
    - PR: #3659 
- [x] **Port dynamic-tier sections** of `context/prompts/templates/system_message_suffix.j2`: DateTime, RepoContext, AvailableSkills, CustomSuffix, CustomSecrets. Preserve the model-family skill gating in `AgentContext.get_system_message_suffix` and the ACP reuse path (`to_acp_prompt_context`). Fold in the defensive minute-rounding of `current_datetime`.
   - PR: #3683 

### Acceptance criteria

- [x] `registry.build(ctx)` reproduces `static_system_message` and `dynamic_context` byte-for-byte across the Phase 0 matrix.
- [x] Each section has a standalone unit test (no Jinja environment).
- [x] `tests/sdk/context/test_agent_context.py` block assertions still pass against registry output.

PR #3796, opened 2026-06-18
refactor(sdk): remove Jinja prompts superseded by the Python prompt registry

VascoSch92:
HUMAN:

Deleteing old jinja prompts not used anymore.

---

AGENT:

End-to-end evidence — instantiated real agents and rendered the full system prompt
  (not just unit tests). Before deleting `system_prompt.j2`, I confirmed the registry
  reproduces its Jinja render **byte-for-byte** for the default and custom policies
  (including the Windows shell refinement). Post-deletion render:

  ```
  $ uv run python -c "<build agents, print static_system_message markers>"
  default(claude): len=14504  <ROLE>=True  <SECURITY>=True  default-policy=True  <IMPORTANT>=True
  gpt-5: model-specific block present = True
  custom policy: CUSTOM_RULE present =True  default-policy leaked =False
  disabled policy: any <SECURITY> tag = False
  ```

  Suites: `ruff` clean; full `agent/`, `context/`, fncall-converter, `delegate/`, and the
  API-breakage signature test pass (1260 passed in the broad sweep).

  ## Why

  The default system prompt is now assembled from the pure-Python section registry
  (#3610 port), but the original Jinja templates were left behind — mostly unused or
  duplicating the registry. Keeping both is a drift/maintenance hazard: two sources of
  truth for the same prompt.

  ## Summary

  - Delete 10 redundant `.j2` templates: 6 dead (unreferenced `{% include %}` wrappers,
    Python-ported in-context examples, an orphaned delegate description) and 4
    `model_specific/*` (byte-for-byte duplicated in `ModelSpecificSection`).
  - Port the last Jinja-only behavior — the custom security-policy `{% include %}` — into
    `SecuritySection` (`AgentBase._read_custom_security_policy`), then delete
    `system_prompt.j2` + `self_documentation.j2` + `security_risk_assessment.j2` +
    `security_policy.j2`; a custom `security_policy_filename` now routes through the registry.
  - Net −743/+80 across 20 files; `render_template` and `system_prompt_planning.j2` are
    retained for the planning preset, custom user templates, and subclasses.

  ## Issue Number

  Relates to #3610 (prompt-registry port).
  
  ## How to Test

  ```
  uv run pytest tests/sdk/context/prompts/ \
    tests/sdk/context/test_prompt_model_spec.py \
    tests/sdk/context/test_prompt_absolute_path.py \
    tests/sdk/agent/test_system_prompt.py \
    tests/sdk/agent/test_security_policy_integration.py
  ```

  Then render a prompt directly:
  
  ```
  uv run python -c "
  print('SECURITY' in Agent(llm=LLM(model='claude-sonnet-4-5', usage_id='t'), tools=[]).static_system_message)
  "
  ```

  ## Video/Screenshots

  N/A — library/CLI change with no UI. Render output included in the AGENT section above.

  ## Type

  - [ ] Bug fix
  - [ ] Feature
  - [x] Refactor
  - [x] Breaking change
  - [ ] Docs / chore
  
  ## Notes

  Minor behavior changes for **custom** security policies (default-policy agents are
  unaffected — already on the registry): content is identical, but inter-section spacing
  is normalized to single blank lines; a disabled policy (`security_policy_filename=""`)
  no longer emits empty `<SECURITY></SECURITY>` tags; and custom policy files are inserted
  verbatim (no longer Jinja-rendered). The static "matches-legacy" test oracle had become
  self-referential (default agents take the registry path), so it's removed along with
  `system_prompt.j2`; `test_prompt_snapshot.py` (byte-exact registry snapshots) carries the
  static regression guard, and the dynamic oracle against `system_message_suffix.j2` stays.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:c537d66-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-c537d66-python \
  ghcr.io/openhands/agent-server:c537d66-python
```

```
ghcr.io/openhands/agent-server:c537d66-golang-amd64
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-golang-amd64
ghcr.io/openhands/agent-server:vasco-clean-jinja-golang-amd64
ghcr.io/openhands/agent-server:c537d66-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:c537d66-golang-arm64
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-golang-arm64
ghcr.io/openhands/agent-server:vasco-clean-jinja-golang-arm64
ghcr.io/openhands/agent-server:c537d66-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:c537d66-java-amd64
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-java-amd64
ghcr.io/openhands/agent-server:vasco-clean-jinja-java-amd64
ghcr.io/openhands/agent-server:c537d66-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:c537d66-java-arm64
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-java-arm64
ghcr.io/openhands/agent-server:vasco-clean-jinja-java-arm64
ghcr.io/openhands/agent-server:c537d66-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:c537d66-python-amd64
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-python-amd64
ghcr.io/openhands/agent-server:vasco-clean-jinja-python-amd64
ghcr.io/openhands/agent-server:c537d66-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:c537d66-python-arm64
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-python-arm64
ghcr.io/openhands/agent-server:vasco-clean-jinja-python-arm64
ghcr.io/openhands/agent-server:c537d66-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:c537d66-golang
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-golang
ghcr.io/openhands/agent-server:vasco-clean-jinja-golang
ghcr.io/openhands/agent-server:c537d66-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:c537d66-java
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-java
ghcr.io/openhands/agent-server:vasco-clean-jinja-java
ghcr.io/openhands/agent-server:c537d66-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:c537d66-python
ghcr.io/openhands/agent-server:c537d66384a9f65339651b5569f764168bd75b9b-python
ghcr.io/openhands/agent-server:vasco-clean-jinja-python
ghcr.io/openhands/agent-server:c537d66-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `c537d66-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `c537d66-python-amd64`) are also available if needed


openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -524,6 +520,17 @@ def _resolved_template_kwargs(self) -> dict[str, object]:
...
+    def _read_custom_security_policy(self) -> str | None:
+        """Content of a custom security policy file, or ``None`` for the default
+        (``security_policy.j2`` sentinel) or disabled (empty) policy.
+
+        Relative names resolve against ``prompt_dir``; absolute paths are used as-is.
+        """
+        filename = self.security_policy_filename
+        if not filename or filename == "security_policy.j2":
+            return None
+        return (Path(self.prompt_dir) / filename).read_text(encoding="utf-8")
```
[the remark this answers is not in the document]
VascoSch92: Good catch — made the raw-text behavior the intentional contract rather than leaving it implicit. A security policy shouldn't be a template-injection surface, so custom policies are now inserted verbatim by design:

- Dropped "template" from the `security_policy_filename` field description and documented that a custom file's contents are inserted verbatim (not rendered as a Jinja template).
- Added `test_custom_security_policy_is_inserted_verbatim_not_rendered`, which asserts `{{ ... }}` / `{% ... %}` in a policy land literally in the system message.

The default built-in policy and the registry render path are unchanged. Marking the PR as a breaking change to flag the contract shift for anyone relying on Jinja in custom policies.

openhands-sdk/openhands/sdk/context/prompts/sections/static.py
```diff
@@ -261,6 +259,12 @@ class SecuritySection(_StaticTextSection):
     def guard(self, ctx: PromptContext) -> bool:
         return bool(ctx.template_kwargs.get("security_policy_filename"))
 
+    def render(self, ctx: PromptContext) -> str | None:
+        content = ctx.template_kwargs.get("security_policy_content")
+        if content:
```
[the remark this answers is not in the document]
VascoSch92: Confirmed and fixed — good catch. `SecuritySection.render` used a truthiness check (`if content:`), so an explicitly empty custom policy (`""`) fell through to the default body. Switched to `content is not None` (the kwarg is only set when a custom policy is configured, so absent → default, `""` → empty custom policy), and added `test_empty_custom_security_policy_does_not_leak_default` for the zero-byte case.

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -463,27 +465,22 @@ def static_system_message(self) -> str:
...
-        The default prompt is assembled from the typed section registry
-        (``create_registry``). Escape hatches keep the Jinja render path: an inline
-        ``system_prompt`` is returned verbatim; a custom/absolute
-        ``system_prompt_filename`` renders through ``render_template``; a subclass
-        with its own ``prompt_dir`` still renders its default-named template; and a
-        custom ``security_policy_filename`` renders so its policy file is included.
+        The default prompt is assembled from the typed section registry, which also
+        resolves a custom ``security_policy_filename``. Escape hatches keep the Jinja
+        path: an inline ``system_prompt`` is returned verbatim; a custom
+        ``system_prompt_filename`` or subclass ``prompt_dir`` renders its own template.
```
enyst, line 475: We still support a .j2 file if custom?

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -247,7 +248,8 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
     security_policy_filename: str = Field(
         default="security_policy.j2",
         description=(
-            "Security policy template filename. Can be either:\n"
+            "Security policy filename. A custom policy file's contents are inserted "
+            "verbatim (NOT rendered as a Jinja template). Can be either:\n"
             "- A relative filename (e.g., 'security_policy.j2') loaded from the "
             "agent's prompts directory\n"
             "- An absolute path (e.g., '/path/to/custom_security_policy.j2')\n"
```
enyst: I'm a bit confused here, the comment says it's not rendered as jinja, but then it continues saying it can be jinja? Sorry I jumped directly in the code, maybe it was explained somewhere?

openhands-sdk/openhands/sdk/agent/prompts/security_risk_assessment.j2
```diff
@@ -1,31 +0,0 @@
-# Security Risk Policy
-When using tools that support the security_risk parameter, assess the safety risk of your actions:
-
-{% if cli_mode | default(true) %}
```
enyst, line 4: This is one reason why I think maybe we need jinja. I know we first made them python, but that way is not as easily customizable as this... I wonder if the next step could be, on the contrary, to put more in jinja rather than less 🤔

VascoSch92: You're right: a template can be more customizable than Python, but I checked, and the prompts never actually used that: no `{% block %}/{% extends %}` anywhere, and `system_prompt.j2` had no inline logic beyond `{{ soul_content }}` and include-guards. 

So every hook the old Jinja exposed is preserved on the registry, just relocated from `{% if %}/{% include %} `into `.guard()/dict` lookups. 

The only thing genuinely removed is Jinja inside a custom policy file (now verbatim), which was deliberate. 

The "override one block without forking" story you're describing would be new capability, neither the old templates nor the registry expose it today. 

If that's what we want, I think the right move is to design a real section-override API (works for both), rather than keep the duplicate templates that only gave us that power in theory.

Moreover, you can always pass a template and kwargs argument for it.

openhands-sdk/openhands/sdk/agent/base.py
VascoSch92: Yes

openhands-sdk/openhands/sdk/agent/base.py
VascoSch92: Sorry, I didn't see this comment and it is missleading. I delete it.

I mean: if you pass a custum jinja will be rendered as a jinja (with the method render_template) but if you don't do it, we use the built-in prompts which are verbatim.

openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -525,6 +522,18 @@ def _resolved_template_kwargs(self) -> dict[str, object]:
                 template_kwargs["model_variant"] = spec.variant
         return template_kwargs
 
+    def _read_custom_security_policy(self) -> str | None:
+        """Raw contents of a custom security policy file -- inserted verbatim, NOT
+        rendered as a Jinja template -- or ``None`` for the default
+        (``security_policy.j2`` sentinel) or disabled (empty) policy.
```
neubig: Is this still correct?

openhands-sdk/openhands/sdk/agent/prompts/security_risk_assessment.j2
enyst: I hear you, I just mean _from user perspective_. Please let me take a step back, what I have in mind is like:
```
# OpenHands Plugin
# can have skills, hooks, subagents and system prompt sections
# you can dump directly the contents of an AgentSkill in .openhands/skills/ and it will work
# you can dump directly SOUL.md in .openhands/prompts and it will work
```

Does that make sense? Restricting it to Python-only feels like it goes against this? Or maybe it should be .md, not .j2 ?

VascoSch92: Actually, 

what you are saying is working here. 

The only things that we don't have (and was the same before) is that you can not change sections of the system prompt. But this is actually another feature that we could implement.

enyst: Oh, does it work for prompt files? SOUL.md and some file name for system prompt?

openhands-sdk/openhands/sdk/agent/base.py
VascoSch92: no.. updated

openhands-sdk/openhands/sdk/agent/prompts/security_risk_assessment.j2
VascoSch92: Yes!

both work today, just via slightly different entry points:

- `SOUL.md` is a true drop-in, zero config. If `~/.openhands/SOUL.md` exists it's auto-loaded (`_load_soul_md()` in `agent/base.py`) and injected into the `<SOUL>` section — nothing to wire up, exactly the behavior you described.

- A full system prompt from a file also works, but you name it rather than have it auto-discovered: set `system_prompt_filename` on the Agent. It accepts a relative filename (resolved against the agent's prompt_dir) or an absolute path, and a non-default value routes through the retained Jinja path so your file fully replaces the built-in prompt.

VascoSch92: @enyst ? :-)

enyst: Oh, I thought I posted, sorry: my bad, so it works for soul, but the same doesn’t apply to the regular system prompt:
```
# OpenHands Plugin
# can have skills, hooks, subagents and system prompt sections
# you can dump directly the contents of an AgentSkill in .openhands/skills/ and it will work
# you can dump directly SOUL.md in .openhands/prompts and it will work
# you can dump directly system_prompt.(md? j2?) in .openhands/prompts and it will work

```

It seems to me the two ways for prompt files are not consistent, and more importantly, the second is Python. So a drop-in file in some location in a Plugin doesn’t work except for SOUL.md.

But I understand better what you meant when you said we need to think about per-sections. You’re probably right: and the reason why I tried to avoid that question is that the system prompt was mostly drafted during Sonnet 3.7 iirc (maybe 4, but I’d bet on 3.7), and comparison with GPT-5.x and Gemini 3.x official prompts shows approx no similitudes. 😅

@openhands List here, in a reply, the current prompt sections, and for each, one-sentence summary of the current system prompt section contents.

enyst: @enyst Current default prompt registry sections (`create_registry()` order) are below; these are the system-prompt registry sections, not the tool schema/tool-description payloads that are added separately.

**Static block** (`SystemPromptEvent.system_prompt`)

1. `soul` (`<SOUL>`): Defines the agent identity, using configured `soul_content`/`SOUL.md` when present or the default OpenHands helper-agent text otherwise.
2. `role` (`<ROLE>`): Defines the primary assistant role, emphasizing thorough software-engineering help and direct answers to question-only requests.
3. `memory` (`<MEMORY>`): Instructs the agent to use and update repository `AGENTS.md` memory for persistent project knowledge.
4. `efficiency` (`<EFFICIENCY>`): Encourages batching actions and using efficient filesystem/search commands to reduce tool-call overhead.
5. `file_system` (`<FILE_SYSTEM_GUIDELINES>`): Sets file-editing hygiene around locating paths, editing originals, avoiding duplicate file variants, and cleaning temporary files.
6. `code_quality` (`<CODE_QUALITY>`): Sets implementation style expectations for minimal, clean, well-understood changes with sparse, high-value comments.
7. `version_control` (`<VERSION_CONTROL>`): Gives git/commit safety rules, including credential handling, commit authorship guidance, and avoiding dangerous git operations.
8. `pull_requests` (`<PULL_REQUESTS>`): Prevents unsolicited pushes/PRs and tells the agent to update an existing PR instead of opening duplicates.
9. `problem_solving` (`<PROBLEM_SOLVING_WORKFLOW>`): Lays out the exploration → analysis → testing → implementation → verification workflow.
10. `self_documentation` (`<SELF_DOCUMENTATION>`): Directs OpenHands-capability/product questions to official docs and requires useful links in answers.
11. `security` (`<SECURITY>`, conditional): Adds the configured security policy, either the built-in default or custom policy content, and can be disabled by an empty policy filename.
12. `security_risk_assessment` (`<SECURITY_RISK_ASSESSMENT>`, conditional): Defines LOW/MEDIUM/HIGH `security_risk` semantics, global escalation rules, and repository-context supply-chain escalation cases.
13. `browser` (`<BROWSER_TOOLS>`, conditional): Describes when and how to use browser tools, including state-before-click/type, action limits, and CAPTCHA/login-wall handling.
14. `external_services` (`<EXTERNAL_SERVICES>`): Requires API-first interaction with services like GitHub/GitLab/Bitbucket and AI disclosure for human-readable external posts.
15. `environment_setup` (`<ENVIRONMENT_SETUP>`): Tells the agent to install missing app/tool dependencies via project dependency files before retrying requested runs.
16. `troubleshooting` (`<TROUBLESHOOTING>`): Requires stepping back after repeated failures, considering likely causes, and asking before major workaround changes.
17. `process_management` (`<PROCESS_MANAGEMENT>`): Gives safe process-termination rules, especially avoiding broad `pkill` patterns and preferring exact PIDs.
18. `model_specific` (`<IMPORTANT>`, conditional): Adds model-family/variant guidance such as Anthropic fail-fast/backcompat caution, Gemini non-proactivity, and GPT-5 communication/PR-thread instructions.

**Dynamic block** (`SystemPromptEvent.dynamic_context`)

19. `datetime` (`<CURRENT_DATETIME>`, conditional): Injects the resolved current date/time when available.
20. `repo_context` (`<REPO_CONTEXT>`, conditional): Injects repository-provided guidance as explicitly untrusted content with prompt-injection warnings.
21. `available_skills` (`<SKILLS>`, conditional): Lists available/triggered skills and instructs the agent to invoke them only through `invoke_skill(...)`.
22. `custom_suffix` (raw text, conditional): Appends the configured `system_message_suffix` verbatim without an additional wrapper.
23. `custom_secrets` (`<CUSTOM_SECRETS>`, conditional): Lists registered secret environment-variable names and explains secret injection, masking, and refresh behavior.

This reply was generated by an AI agent (OpenHands) on behalf of the user.

smolpaws:
🐾 smolpaws here (Engel's agent). I went digging into where these sections *come from*, and it turned into a slightly haunting little history lesson. Sharing because it's directly relevant to "are the sections old."

**Some of these rules are fossils of one specific 2025 model's quirks:**

1. **`NEVER create file_v2.py` / clean up temp files** — Sonnet 3.7, when a `str_replace` failed, would give up and write `script_v2.sh`, `script_v3.sh`. It was also RL'd to *spontaneously write itself `.md` docs* all over the repo dir, unprompted. These lines exist to slap those specific habits down.
2. **`<EFFICIENCY>` — "each action is somewhat expensive, combine bash commands"** — a batching crutch for a model that didn't batch on its own.
3. **`<TROUBLESHOOTING>` — "reflect on 5-7 possible sources"** — scaffolding for weaker reasoning.
4. **Minimal-comments rule** — because (in the original reviewer's words) "Claude LOVES" comments that just narrate the code.

**How I know:** I traced them to **OpenHands/OpenHands#7018** (Xingyao, merged 2025-03-07) — the PR that first carved the flat `<IMPORTANT>` blob into `<ROLE>/<EFFICIENCY>/<FILE_SYSTEM>/…`. It was tuned on **`claude-3-7-sonnet-20250219`** via SWE-Bench Verified — opened *four days after 3.7 shipped*. **@rbren** and **@enyst** argued it line by line (the file_v2 + comments rules are literally rbren's; Engel was already worried back then about over-fitting to Claude). The haunting part: I'm running this on **Opus 4.8** right now. I don't make dupe files. The crutches are for a model two-plus generations dumber than the one reading them — and per rbren's own #7018 insight ("tell the agent it *can* do X and it suddenly really wants to X"), stale prompting can actively *induce* behavior a frontier model wouldn't otherwise have.

**On the actual PR — two threads, kept separate:**

- **Jinja → Python registry (the mechanics):** no real loss. The old `.j2` never used Jinja's actual power — no `{% block %}/{% extends %}`, just `{{ soul_content }}` + include-guards. Every hook is preserved, relocated into `.guard()`/dict lookups. ✅
- **Customizability (Engel's plugin vision — drop `SOUL.md`/a prompt file, don't fork):** real and important, but **orthogonal** to Jinja-vs-Python. A section-override / file-drop API can sit on the Python registry just as cleanly — Vasco's "design a *real* override API rather than keep duplicate templates that gave us that power only in theory" is the right call.

**The fun connection:** "the sections are old" isn't an argument *against* this registry — it's the registry's best feature. Each section already has a `.guard()`, and `ModelSpecificSection` already keys on model family. So the natural next step is making the 3.7-era crutches **model-conditional** — guard them off for frontier models, keep them for weaker/local ones — instead of one-size-fits-all-from-March-2025. Jinja never gave us that cleanly. This does.

tl;dr: 👍 ship the Jinja removal (loses nothing real). The two follow-ups worth filing: (a) a section-override/file-drop API for the plugin story, and (b) an empirical "which sections still earn their place, per model" pass — now tractable because the registry makes sections first-class.

enyst (approved):
Thank you! The archaeology that `smolpaws` did made me smile. I hope you like it too 😅

Anyway this is great clean up! We’ll be better off rethinking anew.

openhands-sdk/openhands/sdk/agent/prompts/security_risk_assessment.j2
enyst: Sorry for the hesitation, I got really entangled in the concern whether we move towards more Claudism or customizability. But I think you’re right. Let’s do this!

PR #1555, opened 2025-12-31
Fix DiscriminatedUnionMixin issues with Pydantic C bindings

tofarr:
## Summary

Refactors `DiscriminatedUnionMixin` to work correctly with Pydantic's C bindings by leveraging Pydantic's native mechanisms instead of fighting against them.

### Architecture Simplification
- **`kind` is now a `@computed_field`** - Returns class name dynamically instead of being a mutable field that gets patched after construction
- **Custom `@model_validator` and `@model_serializer`** - Handle polymorphic validation/serialization through Pydantic's supported hooks
- **Custom `__get_pydantic_json_schema__`** - Generates proper `oneOf` schemas for abstract base classes
- **Removed `model_rebuild()` hackery** - No longer bypasses Pydantic's `complete_model_class()` or patches `model_fields` post-construction
- **Removed FastAPI monkey-patch** - No longer needed since we work with Pydantic rather than against it

### Code Quality Improvements
- **Thread-safe schema cache** - Uses `threading.local()` instead of global dict for `_schemas_in_progress`
- **Deprecated `OpenHandsModel`** - No longer required; new code should extend `BaseModel` directly
- **Cleaner error messages** - Standardized format: `Unknown kind 'X' for module.Class; Expected one of: [...]`
- **Exclude computed fields from MCP/tool schemas** - Updated `model_computed_fields.keys()` exclusion in `mcp/tool.py` and `tool/schema.py`

### What Was Removed (~250 lines)
- `_rebuild_required` global flag and `rebuild_all()` function
- `_create_enhanced_discriminated_union_error_message()` and related helpers
- `_patch_fastapi_discriminated_union_support()` and `_extract_discriminated_unions()`
- Direct assignment to `__pydantic_core_schema__`, `__pydantic_validator__`, `__pydantic_serializer__`

## Backward Compatibility

- `OpenHandsModel` still exists (as empty shell) for external code that extends it
- `kind` field behavior unchanged from consumer perspective
- Serialization/deserialization of polymorphic hierarchies works identically
- JSON schema output structure preserved

## Checklist

- [x] All discriminated union tests pass
- [x] All SDK tests pass
- [x] Pre-commit hooks pass (pyright, ruff, etc.)
- [x] No public API changes requiring documentation updates

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
[...]

PR #1509, opened 2025-12-24
fix: handle union types in py_type for mcp schema conversion

hieptl:
### Summary

Fixes the `py_type` function to properly handle union types (e.g., `["string", "null"]`) that MCP schemas use to mark optional fields. The function now extracts the non-null type from these unions, enabling correct Python type mapping when converting MCP schemas to Pydantic models.

### Problem

MCP schemas often represent optional fields using union types like `["string", "null"]`. Previously, the `py_type` function didn't handle these union types, which could cause incorrect type mapping when `Schema.from_mcp_schema()` creates Pydantic models from MCP tool schemas.

### Solution

Added normalization logic in `py_type` (lines 26-34) that:

- Detects union types represented as `list`, `tuple`, or `set`
- Extracts non-null types from the union
- Returns the single non-null type when exactly one exists
- Returns `Any` when multiple non-null types are present or only `null` exists

This ensures that optional fields like `{"type": ["string", "null"]}` correctly map to `str`, which `from_mcp_schema` then properly annotates as `str | None` for optional fields.

### Changes

- **`openhands/sdk/tool/schema.py`**: Added union type normalization in `py_type` function
- **`tests/sdk/tool/test_py_type.py`**: Added comprehensive test suite with 20 test cases covering:
  - Primitive types (string, integer, number, boolean)
  - Object and array types
  - Union types (list, tuple, set) with various combinations
  - Edge cases (missing keys, unknown types, etc.)

### Testing

All 20 tests pass, covering:

- Union types with single non-null type extraction
- Union types with multiple non-null types (returns `Any`)
- Union types with only `null` (returns `Any`)
- Support for `list`, `tuple`, and `set` union representations
- Integration with array and nested type processing

### Impact

- Enables correct type mapping for MCP tool schemas with optional fields
- Improves type safety when creating Pydantic models from MCP schemas
- Maintains backward compatibility with existing schema formats

### Related

This fix supports the MCP integration in `Schema.from_mcp_schema()`, which creates dynamic Pydantic models from MCP tool input schemas.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Other (dependency update, docs, typo fixes, etc.)

## Checklist

- [x] I have read and reviewed the code and I understand what the code is doing.
- [x] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves #(issue)

## Release Notes

- [ ] Include this change in the Release Notes.

---
[...]

PR #3803, opened 2026-06-19
fix: handle MCP tool arguments named kind

rajshah4:
HUMAN:

I ran into an issue using MCP. I had an agent draft the PR and then I reviewed it. It was able to address the fix that I needed, thanks. it's ready for review.

---

AGENT:

## Summary

Fix MCP tool schema handling when a server exposes a valid tool argument named `kind`.

OpenHands uses `kind` internally as a discriminator on schema/action models. MCP tool servers can also define `kind` as a normal input argument in their JSON Schema, for example as a symbol-kind or node-kind filter. Before this change, that collision could break dynamic MCP action model creation or cause the external `kind` argument to be stripped during argument sanitization.

## What changed

- Alias MCP input fields that collide with OpenHands discriminator field names.
- Preserve the original external MCP argument name when validating and dumping tool arguments.
- Continue stripping OpenHands' internal computed `kind` when the MCP schema did not define it.
- Add a regression test for a GitNexus-style code-intelligence tool schema with a `kind` argument.

## Why

MCP tool input schemas define arbitrary object properties, and `tools/call` sends arbitrary argument keys. A tool argument named `kind` is therefore valid MCP. The client-side OpenHands schema adapter should tolerate that external schema shape without exposing or losing OpenHands' own discriminator.

This is related to prior MCP schema compatibility work such as #1509 and the `DiscriminatedUnionMixin` changes in #1555, but fixes a separate collision between external MCP arguments and OpenHands' internal discriminator field.

## How to Test

```bash
uv run pytest tests/sdk/mcp/test_mcp_tool_validation.py tests/sdk/mcp/test_mcp_tool_serialization.py tests/sdk/tool/test_client_tool.py::test_executor_returns_acknowledgment -q
```

Result: `9 passed`.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:9a061d8-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-9a061d8-python \
  ghcr.io/openhands/agent-server:9a061d8-python
```

```
ghcr.io/openhands/agent-server:9a061d8-golang-amd64
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-golang-amd64
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-golang-amd64
ghcr.io/openhands/agent-server:9a061d8-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:9a061d8-golang-arm64
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-golang-arm64
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-golang-arm64
ghcr.io/openhands/agent-server:9a061d8-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:9a061d8-java-amd64
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-java-amd64
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-java-amd64
ghcr.io/openhands/agent-server:9a061d8-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:9a061d8-java-arm64
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-java-arm64
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-java-arm64
ghcr.io/openhands/agent-server:9a061d8-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:9a061d8-python-amd64
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-python-amd64
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-python-amd64
ghcr.io/openhands/agent-server:9a061d8-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:9a061d8-python-arm64
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-python-arm64
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-python-arm64
ghcr.io/openhands/agent-server:9a061d8-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:9a061d8-golang
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-golang
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-golang
ghcr.io/openhands/agent-server:9a061d8-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:9a061d8-java
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-java
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-java
ghcr.io/openhands/agent-server:9a061d8-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:9a061d8-python
ghcr.io/openhands/agent-server:9a061d83661eccb547378c4eba544045fd27e770-python
ghcr.io/openhands/agent-server:codex-mcp-kind-schema-collision-python
ghcr.io/openhands/agent-server:9a061d8-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `9a061d8-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `9a061d8-python-amd64`) are also available if needed


rajshah4:
Keeping this as draft for maintainer feedback because it touches the shared discriminator path. The regression is MCP-specific: valid tool schemas can define an argument named `kind`, which currently collides with OpenHands' internal `kind` discriminator. Happy to narrow this further if you prefer the fix to live entirely in MCP schema import/sanitization.

simonrosenberg:
AI

```

  Findings (ranked, most severe first)

  1. 🔴 ClientTool with a kind argument breaks event serialization round-trip

  openhands-sdk/openhands/sdk/tool/client_tool.py:90 → schema.py:245

  ClientTool builds its action via the same Action.from_mcp_schema, but the by_alias+exclude fix was applied only to MCPToolDefinition.action_from_arguments. The base Tool.action_from_arguments (tool.py:359) returns the
  dynamic Action instance directly, and that instance is the ActionEvent.action.

  Confirmed: for a ClientTool whose schema declares kind, the action serializes (default model_dump, as event persistence uses) to {'name':'executeCommand','mcp_arg_kind':'Method','kind':'ClientAction_symbol_lookup'} — it
  leaks the internal mcp_arg_kind name and reports kind = the class name instead of "Method". Reloading it through the discriminated union (event store replay / conversation resume) raises ValidationError: mcp_arg_kind
  Extra inputs are not permitted. The conversation can be written but not reloaded. (The MCP path is safe — its args ride in MCPToolAction.data as an opaque dict.)

  2. 🟠 Correctness hinges on a fragile, unenforced by_alias=True + exclude={'kind'} pairing

  openhands-sdk/openhands/sdk/mcp/tool.py:273-277

  The aliased mcp_arg_kind and the computed discriminator kind both serialize to key kind; the computed field wins. action_from_arguments produces the right value only because it pairs by_alias=True with exclude={'kind'}.
  Any dump missing either half silently substitutes the class name for the user's value (this is the root cause of #1). Root issue: the dynamic field uses alias="kind" without populate_by_name=True, so it can't
  round-trip under its own name.

  3. 🟠 LLM-facing schema drops the kind argument's description (and constraints)

  openhands-sdk/openhands/sdk/tool/schema.py:194-206

  Confirmed: in model_json_schema() the computed kind (__get_pydantic_json_schema__ → {const, title:'Kind'}) shadows the aliased field, so to_mcp_schema()/to_openai_tool() emit kind: {"type":"string"} — the user's
  description ("Symbol kind hint…") is gone (a non-colliding field keeps its description). For the exact GitNexus scenario the PR targets, the model gets zero guidance on the kind parameter. The PR's own test misses this
  — it asserts only "kind" in properties, never the contents.

  4. 🟡 Hot-path efficiency regression in _validate_subtype

  openhands-sdk/openhands/sdk/utils/models.py:210-213

  has_kind_alias_field re-scans cls.model_fields on every concrete validation — the path the file's own comment (lines 95-107) documents as "~47% of per-step self-time," explicitly cached for that reason. cls.model_fields
  is class-static; this answer should be cached per class (like the existing _concrete_cache/_checked_cache) rather than recomputed each call.

  5. 🟡 Duplicated discriminator-field expression (reuse)

  openhands-sdk/openhands/sdk/mcp/tool.py:270-272
  openhands-sdk/openhands/sdk/utils/models.py:210-213

  has_kind_alias_field re-scans cls.model_fields on every concrete validation — the path the file's own comment (lines 95-107) documents as "~47% of per-step self-time," explicitly cached for that reason. cls.model_fields
  is class-static; this answer should be cached per class (like the existing _concrete_cache/_checked_cache) rather than recomputed each call.

  5. 🟡 Duplicated discriminator-field expression (reuse)

  openhands-sdk/openhands/sdk/mcp/tool.py:270-272

  The PR extracted Schema._discriminator_field_names() and used it in schema.py, but action_from_arguments still inlines the byte-identical set(DiscriminatedUnionMixin.model_fields) | set(...model_computed_fields). Schema
  is already imported here. The two copies can drift — if the discriminator set ever changes, this exclude set silently won't, leaking internal fields into MCPToolAction.data.

  6. 🟡 Altitude: generalized on one side, hardcoded on the other

  openhands-sdk/openhands/sdk/utils/models.py:211

  from_mcp_schema aliases any field in _discriminator_field_names(), but _validate_subtype hardcodes field_info.alias == "kind". Latent today (the set is exactly {'kind'}), but if a second discriminator field is ever
  added, from_mcp_schema would alias it while this short-circuit silently falls through to data.pop(...) and drops the user's argument with no error.

  7. 🟡 Altitude: reinvents populate_by_name with a magic mcp_arg_ prefix

  openhands-sdk/openhands/sdk/tool/schema.py:245-249

  The hand-rolled mcp_arg_<name> prefix + _2/_3 suffix loop is the source of the leakage in #1/#2 and forces by_alias/exclude patching at every dump site. Pydantic's populate_by_name=True (already used in message.py,
  marketplace/types.py, hooks/config.py) would let the wire name stay the field name, removing the suffix loop, the alias-exclusion set, and the _validate_subtype special case. The suffix branch is also untested.

  8. ⚪ Minor: explicit_mcp_aliases rebuilt on every to_mcp_schema call

  openhands-sdk/openhands/sdk/tool/schema.py:195-199 — iterates all fields and yields an empty set for every hand-written Action/Observation (the common case), on a per-tool-conversion path.

  9. ⚪ Minor: field_kwargs dict vs. the prior ternary

  openhands-sdk/openhands/sdk/tool/schema.py:260-269 — Field(default=default, description=desc, alias=field_alias) is equivalent since pydantic treats None description/alias as unset; the conditional dict re-implements
  behavior Field already provides.

  ---
  Bottom line: The PR correctly fixes the MCP crash, but the change to the shared from_mcp_schema leaves the sibling ClientTool path broken (#1 — the one I'd block on) and degrades the LLM-facing kind description (#3).
  Both, plus the fragility (#2), trace to aliasing kind without populate_by_name. The most robust fix is to switch the dynamic model to populate_by_name=True (keeping the wire name as the field name) and apply consistent
  serialization at the shared layer rather than only in MCPToolDefinition.action_from_arguments — which would also resolve #5–#7. (No CLAUDE.md governs these paths; the repo uses AGENTS.md, and the new comment at
  schema.py:240-243 is within its comment policy — nothing to flag there.)

```

rajshah4:
Thanks for the detailed review. I reproduced the ClientTool failure locally: a schema-declared `kind` argument serialized as `mcp_arg_kind` and failed ActionEvent reload.

Pushed a follow-up commit that:

- enables name-based population for dynamic schemas only when a discriminator-name alias is needed, so persisted ClientTool actions can reload
- adds a ClientTool `kind` argument + `ActionEvent` round-trip regression test
- restores the original MCP schema snippet for aliased collision fields so the LLM-facing `kind` description is preserved
- fixes the pyright issue in the MCP validation test

Local validation now passes:

```bash
uv run pytest tests/sdk/mcp/test_mcp_tool_validation.py tests/sdk/tool/test_client_tool.py tests/sdk/event/test_event_serialization.py -q
uv run pre-commit run --all-files --show-diff-on-failure
```

Keeping this draft while CI reruns and while you/maintainers decide whether this shape is acceptable or if you prefer a narrower MCP-only approach.

rajshah4:
@simonrosenberg I addressed the follow-up feedback and added the inline comment explaining `by_alias=True`.

The previous ClientTool/Event round-trip issue is covered by the new regression test, and the schema description preservation is asserted too. CI is green now, so this should be ready for another look.

rajshah4:
@simonrosenberg gentle follow-up now that the PR has been ready for a few days: CI, QA, and the PR description checks are all green, and the follow-up fix covers the ClientTool/Event round-trip plus schema description preservation. Would appreciate another look when you have a chance.

simonrosenberg:
@rajshah4 woops it slipped from my attention! reviewing asap

simonrosenberg:
@rajshah4 not sure if this is a nit or not... But this round trip seems like it should work, but it breaks




`ActionEvent.model_validate(event.model_dump())` works because it validates through abstract `Action`, but direct concrete dynamic action replay fails:

```py
action = ActionType.model_validate({"name": "executeCommand", "kind": "Method"})
ActionType.model_validate(action.model_dump())
```

This raises `extra_forbidden` on `mcp_arg_kind` because the concrete model sees both the internal field and `kind` discriminator. A small fix is to strip `kind` in the concrete alias path only when it is clearly the persisted discriminator shape: `data.get("kind") == cls.__name__` and the internal aliased field is present. I tested that locally and the targeted 43-test suite still passes.

simonrosenberg (approved):

rajshah4:
@simonrosenberg thanks again for the review and approval.

Looks like this missed v1.29.3 because the release cut went out while the PR was still open. Since this is clean/approved and unblocks MCP tools whose schemas legitimately include a `kind` argument — GitNexus is the concrete case we hit — could we merge it for the next patch release?

Happy to make any last adjustments if needed.

PR #3826, opened 2026-06-21
feat(marketplace): support runtime local plugin loading

csmith49:
HUMAN:

Phase 3 of the marketplace implementation changes. This PR tightly integrates the marketplace data classes into the plugin loading machinery for local conversations.

---

AGENT:

## Summary

Chunk 3 of the marketplace-registration reimplementation, stacked after #3825.

- Add `LocalConversation.load_plugin(plugin_ref)` for resolving and loading plugins from registered marketplaces at runtime.
- Support both `plugin-name` and `plugin-name@marketplace-name` references through `MarketplaceRegistry`.
- Merge runtime plugin skills, MCP config, hooks, and plugin agent definitions into the conversation.
- Reset and rebuild initialized agent tools when runtime plugin loading changes MCP config, so newly loaded MCP tools are available after `_ensure_agent_ready()` has already run.
- Add focused tests for runtime plugin loading, MCP/tool reinitialization, and missing registration errors.

## Why

Runtime marketplace plugin loading lets local conversations opt into additional plugin capabilities after startup without reinitializing the whole conversation.

## Scope

This is SDK-local runtime loading only. Agent-server routes and `RemoteConversation.load_plugin(...)` are planned for later chunks in this stack.

## How to Test

```bash
uv run pytest tests/sdk/conversation/test_local_conversation_plugins.py -q
uv run pre-commit run --files openhands-sdk/openhands/sdk/agent/base.py openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py tests/sdk/conversation/test_local_conversation_plugins.py
```

_This PR was created by an AI agent (OpenHands) on behalf of the user._

@csmith49 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/8dea99f4-c5ab-4cef-b9ff-d0dd7937ee13)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
| python | amd64, arm64 | `nikolaik/python-nodejs:python3.13-nodejs22-slim` | [Link](https://hub.docker.com/_/nikolaik/python-nodejs:python3.13-nodejs22-slim) |
[...]

PR #3825, opened 2026-06-21
feat(marketplace): auto-load registered marketplace plugins

csmith49:
HUMAN:

Phase 2 of the marketplace revamp.

This PR folds the marketplace data classes introduced in the previous PR into the agent context plugin loading mechanism.

---

AGENT:

## Summary

Chunk 2 of the marketplace-registration reimplementation, stacked on #3824.

- Add `AgentContext.registered_marketplaces` using the typed `MarketplaceRegistration` model.
- Auto-load plugins from marketplace registrations marked `auto_load="all"` during `LocalConversation` startup.
- Preserve explicit plugin precedence by loading auto-loaded marketplace plugins before explicit `Conversation(..., plugins=[...])` entries.
- Suppress legacy public skill loading when registered marketplaces are present so `marketplace_path` does not pull from the default OpenHands extensions repo in the new path.
- Make marketplace package exports lazy so `AgentContext` can reference `MarketplaceRegistration` without introducing an import cycle.
- Add focused tests for auto-load behavior, registered-only no-op behavior, explicit plugin precedence, and legacy public-skill suppression.

### REST API contract changes

Compared with base OpenAPI `bac6aef8a727` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -612,0 +613 @@
+schema AgentContext-Input property registered_marketplaces optional schema=type="array" items=MarketplaceRegistration
@@ -622,0 +624 @@
+schema AgentContext-Output property registered_marketplaces optional schema=type="array" items=MarketplaceRegistration
@@ -1567,0 +1570,6 @@
+schema MarketplaceRegistration property auto_load optional schema=type="boolean" default=false
+schema MarketplaceRegistration property name required schema=type="string"
+schema MarketplaceRegistration property ref optional schema=anyOf=[type="string",type="null"]
+schema MarketplaceRegistration property repo_path optional schema=anyOf=[type="string",type="null"]
+schema MarketplaceRegistration property source required schema=type="string"
+schema MarketplaceRegistration type="object"
```

## Scope

This PR intentionally does **not** add runtime `conversation.load_plugin(...)` or agent-server API support. Those are planned follow-up chunks.

## Tests

```bash
uv run pytest tests/sdk/conversation/test_local_conversation_plugins.py tests/sdk/marketplace/test_marketplace_registry.py -q
uv run pytest tests/sdk/conversation/test_local_conversation_plugins.py tests/sdk/marketplace -q
[...]

PR #3827, opened 2026-06-21
feat(marketplace): expose registered marketplaces in skills API

csmith49:
HUMAN:

Adds registered marketplaces to the skills API so external consumers of the SDK can use the feature.

---

AGENT:

## Summary

Chunk 4 of the marketplace-registration reimplementation, stacked on #3826.

- Add `registered_marketplaces` to the agent-server `/api/skills` request model.
- Forward typed `MarketplaceRegistration` values from the router to `load_all_skills(...)`.
- Load skills from auto-loaded marketplace plugins when registered marketplaces are provided.
- Treat registered marketplaces as the public-skill tier replacement while preserving user/org/project precedence.
- Add router and service tests covering registration forwarding, public-skill suppression, and auto-load marketplace skill loading.

### REST API contract changes

Compared with base OpenAPI `1ee364b37a53` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -1627 +1627 @@
-schema MarketplaceRegistration property auto_load optional schema=type="boolean" default=false
+schema MarketplaceRegistration property auto_load optional schema=anyOf=[type="boolean",type="array" items=type="string"] default=false
@@ -1955,0 +1956 @@
+schema SkillsRequest property registered_marketplaces optional schema=type="array" items=MarketplaceRegistration
```

## Why

Registered marketplace skill discovery lets agent-server callers expose marketplace-provided skills through the existing skills API without relying on the legacy public-skill tier.

## How to Test

```bash
uv run pytest tests/agent_server/test_skills_service.py tests/agent_server/test_skills_router.py -q
uv run pre-commit run --files openhands-agent-server/openhands/agent_server/skills_router.py openhands-agent-server/openhands/agent_server/skills_service.py tests/agent_server/test_skills_service.py tests/agent_server/test_skills_router.py
```

_This PR was created by an AI agent (OpenHands) on behalf of the user._

@csmith49 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/8dea99f4-c5ab-4cef-b9ff-d0dd7937ee13)

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:72453d2-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-72453d2-python \
  ghcr.io/openhands/agent-server:72453d2-python
```

```
ghcr.io/openhands/agent-server:72453d2-golang-amd64
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-golang-amd64
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-golang-amd64
ghcr.io/openhands/agent-server:72453d2-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:72453d2-golang-arm64
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-golang-arm64
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-golang-arm64
ghcr.io/openhands/agent-server:72453d2-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:72453d2-java-amd64
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-java-amd64
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-java-amd64
ghcr.io/openhands/agent-server:72453d2-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:72453d2-java-arm64
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-java-arm64
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-java-arm64
ghcr.io/openhands/agent-server:72453d2-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:72453d2-python-amd64
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-python-amd64
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-python-amd64
ghcr.io/openhands/agent-server:72453d2-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:72453d2-python-arm64
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-python-arm64
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-python-arm64
ghcr.io/openhands/agent-server:72453d2-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:72453d2-golang
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-golang
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-golang
ghcr.io/openhands/agent-server:72453d2-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:72453d2-java
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-java
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-java
ghcr.io/openhands/agent-server:72453d2-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:72453d2-python
ghcr.io/openhands/agent-server:72453d2221257acf129a949e44de37476f4ce761-python
ghcr.io/openhands/agent-server:feat-marketplace-skills-api-python
ghcr.io/openhands/agent-server:72453d2-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `72453d2-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `72453d2-python-amd64`) are also available if needed


openhands-agent-server/openhands/agent_server/skills_service.py
```diff
@@ -337,12 +382,20 @@ def load_all_skills(
     sources["sandbox"] = len(sandbox_skills)
     skill_lists.append(sandbox_skills)
 
-    # 2-3. Load public + user skills via helper (no project yet — org sits between)
+    marketplace_skills: list[Skill] = []
+    if load_public and registered_marketplaces:
+        marketplace_skills = load_registered_marketplace_skills(registered_marketplaces)
+    sources["registered_marketplaces"] = len(marketplace_skills)
```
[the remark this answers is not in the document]
csmith49: Kept as-is for now. The new source key is exercised by the added registered-marketplace tests; extending the existing baseline test would be fine but is not needed for this approved PR.

_Reply generated by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/marketplace/registration.py
```diff
@@ -25,7 +26,7 @@ class MarketplaceRegistration(BaseModel):
             "Only relevant for git sources."
         ),
     )
-    auto_load: bool = Field(
+    auto_load: bool | Literal["all"] = Field(
```
VascoSch92: Do we still have the literal here? I thought we used just the `bool`.

Here, `bool` (`True`) and `all` are actually equivalent

VascoSch92 (changes requested):
There is a possible silent loss of public skills when registrations aren't auto‑load

In load_all_skills (`skills_service.py`), two decisions are made by different conditions that can disagree:

- Loading marketplace skills only considers auto‑load registrations: load_registered_marketplace_skills iterates registry.get_auto_load_registrations(), which keeps only entries where auto_load is truthy.
- Suppressing legacy public skills keys off the raw list instead: include_public=load_public and not registered_marketplaces,
- This is False whenever the list is non‑empty, regardless of auto_load.

So if a caller passes a registration with auto_load=False the result is:

1. load_registered_marketplace_skills([...]) → get_auto_load_registrations() returns [] → marketplace_skills = []
2. include_public → load_public and not [one item] → False → legacy public skills suppressed

Net effect: the public tier is turned off, but nothing is loaded to replace it. The caller silently gets no public‑tier skills at all, with no error or log.

```
┌─────────────────────────┬────────────────────┬────────────────┬─────────────────────────────┐
│ registered_marketplaces │ marketplace_skills │ include_public │           Result            │
├─────────────────────────┼────────────────────┼────────────────┼─────────────────────────────┤
│ [] / None               │ []                 │ True           │ public loaded ✓             │
├─────────────────────────┼────────────────────┼────────────────┼─────────────────────────────┤
│ [{auto_load:"all"}]     │ loaded             │ False          │ public replaced ✓           │
├─────────────────────────┼────────────────────┼────────────────┼─────────────────────────────┤
│ [{auto_load:False}]     │ []                 │ False          │ no public, no replacement ✗ │
└─────────────────────────┴────────────────────┴────────────────┴─────────────────────────────┘
```

We could drive both decisions off the same condition:

```python
auto_load_registrations = [r for r in (registered_marketplaces or []) if r.auto_load]

marketplace_skills: list[Skill] = []
if load_public and auto_load_registrations:
    marketplace_skills = load_registered_marketplace_skills(registered_marketplaces)
...
include_public=load_public and not auto_load_registrations,
```

This way public skills are suppressed only when there's actually an auto‑load marketplace to replace them; an auto_load=False‑only list falls back to the legacy path. If suppressing public on any registration is the intended behavior, could we add a test for the auto_load=False case and reword the field description/comment so the drop is documented rather than incidental?

openhands-sdk/openhands/sdk/marketplace/registration.py
csmith49: Fixed in 07a8e1d6: removed the `Literal["all"]` path and made `auto_load` a plain bool, updating the API/tests to use `true` for auto-load.

_Reply generated by an AI agent (OpenHands) on behalf of the user._

jpshackelford:
**Question on the `auto_load` type change (`Literal["all"] | None` → `bool`)**

Following up on the `Literal["all"]` → `bool` simplification from 07a8e1d6. I think the original `Literal["all"] | None` was deliberately extensible, and collapsing it to a plain `bool` narrows a runway we said we wanted to keep:

- The originating issue #2494 lists **`auto_load: ["plugin-a", "plugin-b"]` selective  loading** explicitly under **"Out of Scope (Future Work)."** A `Literal["all"] | None`  union grows naturally into `Literal["all"] | list[str] | None` without changing the meaning of existing values; a `bool` cannot host that third state without another type
  change.
  
- `auto_load` is now being baked into the **public agent-server `/api/skills` request  schema** in this PR. Once that contract ships, widening the type later becomes a  breaking OpenAPI change — so this PR is effectively the last low-cost moment to get the contract right.

**Timing note:** the SDK-level `auto_load: bool` already shipped in **v1.29.2** (via #3825), so a straight revert to `Literal["all"] | None` would technically break a released field. We could do that given that this has appeared in exactly one patch release. Or we could  make a  **backward-compatible widening** instead of a revert, e.g.:

```python
auto_load: bool | list[str] = False
# True  -> load all (≡ the old "all")
# False / [] -> load none
# ["a", "b"] -> selective (future work from #2494)
```

This keeps every released auto_load=True/False caller working, restores the selective-load path, and lets us land the correct type in the API schema before #3827 merges. The model change can't go into the already-merged #3825, so it would be a small fast-follow PR to main that this stack rebases onto.

Two small follow-ups regardless of what we decide:

The router docstring here still says "registrations with auto_load='all' replace legacy public skills", which no longer matches the bool field — worth reconciling.

Also I could file a dedicated issue to track the selective auto_load: [...] work so the intent isn't lost.

Sorry to have been late to the party on this -- and if I have misunderstood your plans or thinking for how to handle this in the future feel free to link me rather than reiterating the old ground.

VascoSch92:
From my side is good. 

There is just the change requested from @jpshackelford .

But I don't know if it should be a `Literal` or just a `str`.

csmith49:
Appreciate the feedback y'all. Our wires definitely got a bit crossed in the earlier PR that set `auto_load` to a Boolean.

I'm not sure the `Literal` approach is what we'd want to do -- the Boolean sets the behavior to all-or-nothing, but the only reasonable strategy I can imagine in between is just selecting the relevant plugins.

If we want to extend with more strategies I think we can just widen the type again without much issue.

@VascoSch92 @jpshackelford I think the only blocker now is an "accept", but if y'all have more suggestions happy to incorporate them.

ETA: There is a failing CI check, but it appears it's just a symptom of the API widening. Intended, and should disappear on any future versions.

VascoSch92 (approved):

VascoSch92:
Yes, 

the API break is because of `auto_load: bool | list[str]` which was `auto_load: bool`.

So I don't know if it is a real API breakage.

cc: FYI @enyst

jpshackelford:
Really excited about the great work on this. Thanks @csmith49 and @VascoSch92 for all the ❤️ you put into this stack of PRs!

enyst:
@openhands Look at the CI failure. List here on github, the paths that fail, since when do they exist in the codebase (how many releases and which), and why. 

Then take a step back and tell me WDYT: is the breakage a real compatibility issue, could the workflow be more lenient?

enyst:
@jpshackelford I rechecked the failing REST API job and the PR description contract section. I also removed the Markdown table because the `|` in union labels made it render badly.

### What is failing

The failed check is not caused by the new optional `/api/skills` request field. It is caused by this response-schema widening:

```py
MarketplaceRegistration.auto_load: bool -> bool | list[str]
```

That field is reachable through:

```text
ConversationInfo.agent.(ACPAgent or Agent).agent_context.registered_marketplaces[].auto_load
```

`oasdiff` reports it once for each public response that returns `ConversationInfo`, and separately for the `ACPAgent` and `Agent` union members.

Affected public response locations:

- `GET /api/conversations/search` `200`
  - response shape: `ConversationPage.items[].agent...registered_marketplaces[].auto_load`
  - endpoint exists since `1.0.0` — 57 releases (`1.0.0` through `v1.29.3`)
  - the failing nested field exists since `v1.29.2` — 2 releases (`v1.29.2`, `v1.29.3`)
  - reason: response field type widened from boolean to boolean-or-array
- `GET /api/conversations?ids=...` `200`
  - response shape: `list[ConversationInfo | null][].agent...registered_marketplaces[].auto_load`
  - endpoint exists since `1.0.0` — 57 releases (`1.0.0` through `v1.29.3`)
  - the failing nested field exists since `v1.29.2` — 2 releases (`v1.29.2`, `v1.29.3`)
  - reason: same response field widening
- `GET /api/conversations/{conversation_id}` `200`
  - response shape: `ConversationInfo.agent...registered_marketplaces[].auto_load`
  - endpoint exists since `1.0.0` — 57 releases (`1.0.0` through `v1.29.3`)
  - the failing nested field exists since `v1.29.2` — 2 releases (`v1.29.2`, `v1.29.3`)
  - reason: same response field widening
- `POST /api/conversations` `200`
  - response shape: `ConversationInfo.agent...registered_marketplaces[].auto_load`
  - endpoint exists since `1.0.0` — 57 releases (`1.0.0` through `v1.29.3`)
  - the failing nested field exists since `v1.29.2` — 2 releases (`v1.29.2`, `v1.29.3`)
  - reason: same response field widening
- `POST /api/conversations/{conversation_id}/fork` `201`
  - response shape: `ConversationInfo.agent...registered_marketplaces[].auto_load`
  - endpoint exists since `v1.18.0` — 22 releases (`v1.18.0` through `v1.29.3`)
  - the failing nested field exists since `v1.29.2` — 2 releases (`v1.29.2`, `v1.29.3`)
  - reason: same response field widening

Field provenance:

- `AgentContext.registered_marketplaces` / `MarketplaceRegistration.auto_load` was introduced by `c82e7efe` (`feat(marketplace): auto-load registered marketplace plugins (#3825)`) and shipped in `v1.29.2` and `v1.29.3`.
- The selective list form was introduced on this PR branch by `4b31fe07` and has not shipped in a release tag yet.

### Is this a real compatibility issue?

Strictly, yes: it is a public response contract change. A generated client based on `v1.29.2` or `v1.29.3` can model `auto_load` as a boolean, so receiving `auto_load: ["plugin-a"]` from a `ConversationInfo` response could fail deserialization.

Practically, I think the blast radius is small:

- the nested field has only shipped in 2 releases;
- old boolean payloads remain valid;
- existing clients should keep seeing booleans unless a conversation actually contains selective marketplace auto-load configuration;
- the list value represents new opt-in behavior that older clients could not express through the old contract.

So my read is: real in the strict OpenAPI/generated-client sense, low practical risk.

### Could the workflow be more lenient?

I would not make it globally lenient for all response type widening, because that can hide real generated-client breaks.

A narrower policy seems reasonable: keep response type widening as fail-by-default, but allow warning/manual-review for cases where the old type remains accepted and the affected field is very new, explicitly extensible, or opt-in. A targeted exception with a release-note/manual-review requirement would also be reasonable here.

### Does the PR description REST API contract section tell the truth?

Yes, the section is current and truthful for what it says it is comparing: base OpenAPI `217b2cf83aab` to PR head, summarized at the component/schema level.

This part is accurate:

```diff
-schema MarketplaceRegistration property auto_load optional schema=type="boolean" default=false
+schema MarketplaceRegistration property auto_load optional schema=anyOf=[type="boolean",type="array" items=type="string"] default=false
@@ -1955,0 +1956 @@
+schema SkillsRequest property registered_marketplaces optional schema=type="array" items=MarketplaceRegistration
```

It is not outdated. It is just a different view from the failing CI check:

- the PR description summarizes component-level base-to-head changes;
- the CI breakage job compares against the published baseline release and expands component references into every public response path where the changed component is reachable.

So the PR description is truthful but not sufficient by itself to explain the CI failure. The failure comes from the `MarketplaceRegistration.auto_load` widening as it propagates into `ConversationInfo` responses. The new optional `SkillsRequest.registered_marketplaces` request field is additive and is not the failing part.

enyst (approved):

PR #3586, opened 2026-06-09
fix(llm): don't emit explicit cache_control markers for Gemini

VascoSch92:
AGENT:
  
  Validation is deterministic + offline. The regression is a cache-billing behavior, so the proof is twofold:

  - **Code-level proof (this PR's test):** `test_gemini_prompt_caching_emits_no_markers` builds a multi-turn Gemini conversation where the system message (index 0) and the last user message (index 3) are *non-adjacent* — the exact
  shape that froze the cache — and asserts `is_caching_prompt_active()` is `False` and that no `cache_control` marker appears anywhere in the formatted payload. No markers → no explicit `cachedContent` attached → Gemini stays on
  Google's implicit-caching path.
  - **Behavioral proof (issue #3585):** the production SWE-bench/GAIA A/B straddling the regression shows cache-hit ratio dropping ~83% → ~36% and spend rising ~6× (SWE-bench) / ~14× (GAIA) once markers were emitted. I did **not**
  re-run a live Gemini eval for this PR; the offline test reproduces the root cause deterministically.

  Commands run:
  - `uv run pytest tests/sdk/llm/test_model_features.py::test_prompt_cache_support tests/sdk/llm/test_prompt_caching_cross_conversation.py` → 28 passed
  - `uv run pytest tests/sdk/llm/test_model_features.py tests/sdk/llm/test_llm_completion.py` → 165 passed (confirms the `prompt-cache-too-small` retry tests, which use `claude-sonnet` as the real model, are unaffected)

  ---

  ## Why

  `#3090` added `gemini-2.5`/`gemini-3` to `PROMPT_CACHE_MODELS`, which made the SDK emit Anthropic-style `cache_control` markers for Gemini. For Anthropic, the second marker on the last user/tool message extends the cached prefix
  every turn. For Gemini via LiteLLM it does the opposite:

  - LiteLLM's Vertex/Gemini context-caching transform (`separate_cached_messages`) keeps only the **first contiguous** marked block. The SDK marks two *non-adjacent* messages (system + last user/tool), so after turn 1 the cache
  collapses to just the static system prefix — frozen for the whole trajectory.
  - Attaching any explicit `cachedContent` to the request **disables Google's implicit caching** on the growing conversation body, so the body is re-billed as fresh input every turn.

  Net effect: Gemini runs since post-#3090 silently cost ~6–14× more, with no resolve-rate benefit. Before #3090 (no markers), Gemini's implicit caching covered the full growing prefix (~75–83% hit).

  ## Summary
  
[...]

PR #3873, opened 2026-06-25
fix(llm): recover softly from Anthropic content-policy blocks

### LINKED ISSUE #3798: Anthropic content-policy block hard-errors the conversation instead of degrading gracefully
## Summary

An Anthropic **output** content-policy block (`litellm.ContentPolicyViolationError`, HTTP 400 "Output blocked by content filtering policy") currently terminates an OpenHands conversation with a fatal `ConversationErrorEvent` (`execution_status=error`). The run is technically recoverable — the user can send another message to resume — but the UX is a scary hard error rather than a graceful "the model declined that output, continuing" experience.

Content-policy is **not distinguished** from any other deterministic 400: it is lumped into the generic `LLMBadRequestError` and then falls through `astep`'s typed `except` blocks into the conversation loop's catch-all `except Exception`, which sets `ERROR` and emits a fatal event.

## Real repro

Observed on a live QA of OHE (OHE 1.38.1, agent-server 1.29.0, bundled LiteLLM, `llm_provider=anthropic`):

- Benign task: "pull this repo and open a PR that adds value."
- The agent had created `LICENSE` + `.gitignore` and was generating the next doc (almost certainly a `CODE_OF_CONDUCT.md` — Contributor-Covenant vocabulary reliably false-positives content classifiers).
- Anthropic's **output** filter tripped mid-generation.
  - model: `claude-sonnet-4-5-20250929`
  - request_id: `req_011CcB8y…`
  - `litellm.ContentPolicyViolationError` / AnthropicError 400 "Output blocked by content filtering policy" (`content_policy_fallback=None`)

Result chain: `ContentPolicyViolationError` → SDK `LLMBadRequestError` → `ConversationErrorEvent(execution_status=error)` → UI hard error.

## Current handling (where it goes fatal, and that content-policy is not distinguished)

litellm's `ContentPolicyViolationError` subclasses `litellm.BadRequestError`, so it traverses the normal 400 path:

1. **Not retried (correct).** `LLM_RETRY_EXCEPTIONS` in `openhands-sdk/openhands/sdk/llm/llm.py:144-151` excludes `BadRequestError`/`ContentPolicyViolationError`. So tenacity never retries it.
2. **Not eligible for SDK fallback.** `FallbackStrategy.should_fallback` uses `_LLM_FALLBACK_EXCEPTIONS` in `openhands-sdk/openhands/sdk/llm/fallback_strategy.py:28-35` — same transient-only set — so the SDK-level `fallback_strategy` also won't trigger.
3. **Classified as a generic 400 — content-policy is NOT distinguished.** `acompletion`'s handler (`llm.py` ~1475-1504) calls `_ahandle_error` (`llm.py:811-842`), which calls `map_provider_exception`. In `openhands-sdk/openhands/sdk/llm/exceptions/mapping.py`, content-policy is not context-window, not malformed-history, not auth, not rate-limit/timeout/connection — it falls to the catch-all at **`mapping.py:60-61`**:
   ```python
   if isinstance(exception, BadRequestError):
       return LLMBadRequestError(str(exception))
   ```
   `_ahandle_error` then `raise mapped from error`. There is **no** content-policy branch anywhere in `exceptions/classifier.py` or `exceptions/mapping.py`.
4. **Propagates past `astep`'s soft handlers.** `Agent.astep` (`openhands-sdk/openhands/sdk/agent/agent.py:745-807`, sync `step` mirror at ~605-664) only catches `FunctionCallValidationError`, `LLMMalformedConversationHistoryError`, and `LLMContextWindowExceedError`. `LLMBadRequestError` is none of these, so it propagates out of `astep`.
5. **Becomes a fatal `ConversationErrorEvent`.** The catch-all in `LocalConversation.arun` at **`openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py:1731-1747`** sets `execution_status = ERROR`, emits `ConversationErrorEvent(source="environment", code="LLMBadRequestError", detail=...)`, then `raise ConversationRunError(...)`. (Sync `run` has the equivalent at ~1275-1289.) In the agent-server, `EventService._run_and_publish` (`openhands-agent-server/openhands/agent_server/event_service.py:885-921`) catches the re-raised `ConversationRunError` only as a backstop — the fatal event has already been published to WebSocket subscribers, which is what the UI renders.

**Existing soft/recoverable pattern (the hook to reuse):** `astep`/`step` already implement non-fatal recovery for `FunctionCallValidationError` (emit a `MessageEvent` and `return` — the run loop continues) and for context-window/malformed-history (emit a `CondensationRequest` and `return`). Content-policy currently has no such branch.

## Why a plain retry won't fix it

A content-policy 400 is **deterministic**: it is a property of the request + model, not a transient capacity/server condition. Re-sending the identical request (same `messages`) trips the same Anthropic output classifier every time — it only burns tokens. This is precisely why it is (correctly) excluded from both `LLM_RETRY_EXCEPTIONS` and `_LLM_FALLBACK_EXCEPTIONS`, unlike 429/5xx/timeout. Recovery requires *changing* the request (a nudge that steers the model toward different output) or routing to a *different* model/provider — not a bare retry.

## Recommended graceful design

Primary fix in this repo — classify content-policy distinctly and handle it as a **soft, non-fatal, resumable** event (combines options (b) + (c)):

1. `openhands-sdk/openhands/sdk/llm/exceptions/types.py`: add `LLMContentPolicyViolationError(LLMBadRequestError)` (subtype keeps back-compat for anything catching `LLMBadRequestError`).
2. `openhands-sdk/openhands/sdk/llm/exceptions/classifier.py`: add `is_content_policy_violation(exc)` — `isinstance(exc, litellm ContentPolicyViolationError)` plus a text-pattern fallback (`"output blocked by content filtering"`, `"content_policy"`, `"content filtering policy"`) for proxied/aliased providers whose typed class is flattened.
3. `openhands-sdk/openhands/sdk/llm/exceptions/mapping.py`: add a branch **before** the generic `BadRequestError` branch (line 60) → `return LLMContentPolicyViolationError(...)`. Export from `exceptions/__init__.py`.
4. `openhands-sdk/openhands/sdk/agent/agent.py`: add `except LLMContentPolicyViolationError` to both `astep` and `step`, modeled on the existing `FunctionCallValidationError` branch — emit a soft note (a `MessageEvent` such as "The previous response was blocked by the model's content filter; rephrasing/continuing without the flagged content") and `return` so the run loop continues / lands in a runnable state instead of `ERROR`. Bound it with a small per-run consecutive-content-policy counter so a model that keeps tripping the filter degrades to a soft idle/pause (awaiting user) rather than looping to `MaxIterationsReached`.

This directly meets the user goal: no hard error, automatic smooth resume.

### Alternatives considered

- **(a) litellm `content_policy_fallbacks`** (set `proxy_config.litellm_settings.content_policy_fallbacks` on the bundled proxy; supported on the shipped `litellm-database:1.84.1`). Transparent to the SDK and requires no code change, but: it needs a **second, ideally different-provider** model configured — with a single Anthropic provider the fallback target is another `claude-*` model that trips the **same** output classifier, so it can't recover the R01 case; and it silently swaps which model answers (cost/settings mismatch). Useful as an optional secondary mitigation, not the core fix. → companion note for `OpenHands/OpenHands-Cloud` (see below).
- **(b) alone** (emit a non-fatal note and pause to idle): good, but on its own it doesn't auto-resume; pairing with (c)'s nudge is what makes resume automatic.

## Secondary follow-up (different repo)

`OpenHands/OpenHands-Cloud`: optionally expose a `content_policy_fallbacks` entry in the bundled litellm `proxy_config` (`replicated/openhands.yaml`) so installs that configure a second/distinct-provider model get transparent recovery. Worth a separate issue, with the explicit caveat that it is **inert for single-Anthropic-model installs** (the fallback hits the same output filter). The SDK fix above is the general solution and is independent of provider configuration.

## Scope note

Verified against the v1.29.0 line; the same code paths and line references apply on `main` (file:line may drift slightly). Affects both async (`astep`/`arun`) and sync (`step`/`run`) paths.

VascoSch92:
HUMAN:

add error for policy error

---

AGENT:

## Why

An Anthropic **output** content-policy block (`litellm.ContentPolicyViolationError`,
HTTP 400 "Output blocked by content filtering policy") currently terminates a
conversation with a fatal `ConversationErrorEvent` (`execution_status=error`).
The run is actually recoverable — the user can send another message — but the UX
is a scary hard error.

Root cause: `ContentPolicyViolationError` subclasses `litellm.BadRequestError`, so
it is **not distinguished** from any other deterministic 400. It is correctly
excluded from retries (`LLM_RETRY_EXCEPTIONS`) and SDK fallback
(`_LLM_FALLBACK_EXCEPTIONS`), but `map_provider_exception` collapses it into the
generic `LLMBadRequestError`, which is none of the typed exceptions `astep`/`step`
catch. It therefore falls through to the `arun`/`run` catch-all that sets `ERROR`
and emits a fatal event.

A bare retry cannot fix it: the block is deterministic in `(messages, model)`, so
re-sending the identical request trips the same output classifier and only burns
tokens. Recovery requires *changing* the request.

## Summary

- Add `LLMContentPolicyViolationError(LLMBadRequestError)` — the subtype keeps
  back-compat for anything catching `LLMBadRequestError` — and
  `is_content_policy_violation()` (typed `isinstance` check + a text-pattern
  fallback for proxied/aliased providers whose typed class is flattened).
- Map content-policy **before** the generic `BadRequestError` branch in
  `map_provider_exception`.
- Handle it in both `step` and `astep` exactly like a malformed function call:
  emit a soft user-message nudge and `return`, so the run loop stays `RUNNING`
  and auto-resumes with a changed request. A model that keeps tripping the filter
  is bounded by the existing stuck detector rather than a fatal event.

## Issue Number

Closes #3798

## How to Test

```
uv run pytest \
  tests/sdk/llm/test_exception_mapping.py \
  tests/sdk/agent/test_agent_content_policy_violation.py -q
```

Result: `14 passed`. The new agent tests assert that a content-policy error in
`step`/`astep` (both Chat and Responses APIs) produces a single soft user-message
nudge containing "content filter" and **no** `ConversationErrorEvent`, and that
`step`/`astep` do not raise. The mapping tests cover the typed exception, the
text-fallback path, and that a generic 400 is *not* misclassified.

Regression check (no behavior change to neighboring exception handling):

```
uv run pytest \
  tests/sdk/llm/test_exception_classifier.py \
  tests/sdk/llm/test_exception.py \
  tests/sdk/agent/test_agent_context_window_condensation.py -q
```

Result: `44 passed`. `pre-commit` (ruff, pycodestyle, pyright, import rules)
passes on all changed files.

**Limitation / not yet done end-to-end:** a live reproduction requires actually
tripping Anthropic's output content filter (the original report observed it while
the agent generated a `CODE_OF_CONDUCT.md`), which is non-deterministic and needs
a real provider key. The change is exercised here by injecting the typed
exception at the LLM boundary, which is the exact point where litellm raises it.
A reviewer with a live Anthropic key can confirm the end-to-end UX by running a
task that reliably trips the filter and verifying the conversation continues
instead of surfacing a hard error.

## Video/Screenshots

N/A — backend-only change; evidence is the test output above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Minimal, provider-independent SDK fix. It does **not** add an LLM-level
  `content_policy_fallbacks` config; that is a separate, optional mitigation best
  filed against `OpenHands/OpenHands-Cloud`, with the caveat that it is **inert
  for single-Anthropic installs** (the fallback target is another `claude-*`
  model that hits the same output filter).
- Applies to both the async (`astep`/`arun`) and sync (`step`/`run`) paths.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:f767244-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-f767244-python \
  ghcr.io/openhands/agent-server:f767244-python
```

```
ghcr.io/openhands/agent-server:f767244-golang-amd64
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-golang-amd64
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-golang-amd64
ghcr.io/openhands/agent-server:f767244-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:f767244-golang-arm64
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-golang-arm64
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-golang-arm64
ghcr.io/openhands/agent-server:f767244-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:f767244-java-amd64
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-java-amd64
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-java-amd64
ghcr.io/openhands/agent-server:f767244-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:f767244-java-arm64
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-java-arm64
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-java-arm64
ghcr.io/openhands/agent-server:f767244-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:f767244-python-amd64
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-python-amd64
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-python-amd64
ghcr.io/openhands/agent-server:f767244-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:f767244-python-arm64
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-python-arm64
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-python-arm64
ghcr.io/openhands/agent-server:f767244-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:f767244-golang
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-golang
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-golang
ghcr.io/openhands/agent-server:f767244-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:f767244-java
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-java
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-java
ghcr.io/openhands/agent-server:f767244-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:f767244-python
ghcr.io/openhands/agent-server:f7672441a5be0c78b29da910d1e65fae2e17295d-python
ghcr.io/openhands/agent-server:fix-content-policy-soft-recovery-python
ghcr.io/openhands/agent-server:f767244-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `f767244-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `f767244-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/agent.py
+++ b/openhands-sdk/openhands/sdk/agent/agent.py
@@ -609,11 +609,8 @@ class Agent(CriticMixin, ResponseDispatchMixin, AgentBase):
             on_event(error_message)
             return
         except LLMContentPolicyViolationError as e:
-            # The provider's content filter blocked this completion. It is
-            # deterministic, so re-sending the same request is futile; instead
-            # nudge the model with a user message and let the run loop continue.
-            # A model that keeps tripping the filter is bounded by the stuck
-            # detector rather than emitting a fatal ConversationErrorEvent.
+            # Content-policy blocks are deterministic; nudge the model and let the
+            # run loop continue instead of emitting a fatal error.
             logger.warning(f"LLM output blocked by content filter: {e}")
             on_event(
                 MessageEvent(
@@ -774,11 +771,8 @@ class Agent(CriticMixin, ResponseDispatchMixin, AgentBase):
             on_event(error_message)
             return
         except LLMContentPolicyViolationError as e:
-            # The provider's content filter blocked this completion. It is
-            # deterministic, so re-sending the same request is futile; instead
-            # nudge the model with a user message and let the run loop continue.
-            # A model that keeps tripping the filter is bounded by the stuck
-            # detector rather than emitting a fatal ConversationErrorEvent.
+            # Content-policy blocks are deterministic; nudge the model and let the
+            # run loop continue instead of emitting a fatal error.
             logger.warning(f"LLM output blocked by content filter: {e}")
             on_event(
                 MessageEvent(
--- a/openhands-sdk/openhands/sdk/agent/agent.py
+++ b/openhands-sdk/openhands/sdk/agent/agent.py
@@ -61,6 +61,7 @@ from openhands.sdk.llm import (
 )
 from openhands.sdk.llm.exceptions import (
     FunctionCallValidationError,
+    LLMContentPolicyViolationError,
     LLMContextWindowExceedError,
     LLMMalformedConversationHistoryError,
 )
@@ -607,6 +608,31 @@ class Agent(CriticMixin, ResponseDispatchMixin, AgentBase):
             )
             on_event(error_message)
             return
+        except LLMContentPolicyViolationError as e:
+            # The provider's content filter blocked this completion. It is
+            # deterministic, so re-sending the same request is futile; instead
+            # nudge the model with a user message and let the run loop continue.
+            # A model that keeps tripping the filter is bounded by the stuck
+            # detector rather than emitting a fatal ConversationErrorEvent.
+            logger.warning(f"LLM output blocked by content filter: {e}")
+            on_event(
+                MessageEvent(
+                    source="user",
+                    llm_message=Message(
+                        role="user",
+                        content=[
+                            TextContent(
+                                text=(
+                                    "Your previous response was blocked by the "
+                                    "model's content filter. Please continue, "
+                                    "rephrasing to avoid the flagged content."
+                                )
+                            )
+                        ],
+                    ),
+                )
+            )
+            return
         except LLMMalformedConversationHistoryError as e:
             # The provider rejected the current message history as structurally
             # invalid (for example, broken tool_use/tool_result pairing). Route
@@ -746,6 +772,31 @@ class Agent(CriticMixin, ResponseDispatchMixin, AgentBase):
                 ),
             )
             on_event(error_message)
+            return
+        except LLMContentPolicyViolationError as e:
+            # The provider's content filter blocked this completion. It is
+            # deterministic, so re-sending the same request is futile; instead
+            # nudge the model with a user message and let the run loop continue.
+            # A model that keeps tripping the filter is bounded by the stuck
+            # detector rather than emitting a fatal ConversationErrorEvent.
+            logger.warning(f"LLM output blocked by content filter: {e}")
+            on_event(
+                MessageEvent(
+                    source="user",
+                    llm_message=Message(
+                        role="user",
+                        content=[
+                            TextContent(
+                                text=(
+                                    "Your previous response was blocked by the "
+                                    "model's content filter. Please continue, "
+                                    "rephrasing to avoid the flagged content."
+                                )
+                            )
+                        ],
+                    ),
+                )
+            )
             return
         except LLMMalformedConversationHistoryError as e:
             # The provider rejected the current message history as
--- a/openhands-sdk/openhands/sdk/llm/exceptions/classifier.py
+++ b/openhands-sdk/openhands/sdk/llm/exceptions/classifier.py
@@ -20,7 +20,7 @@ from .types import (
 
 
 # Minimal, provider-agnostic context-window detection
-LONG_PROMPT_PATTERNS: list[str] = [
+LONG_PROMPT_PATTERNS: Final[list[str]] = [
     "contextwindowexceedederror",
     "prompt is too long",
     "input length and `max_tokens` exceed context limit",
@@ -35,7 +35,7 @@ LONG_PROMPT_PATTERNS: list[str] = [
 # provider. They are tracked separately from true context-window errors so the
 # logs and agent control flow can preserve that distinction while still routing
 # into condensation-based recovery.
-MALFORMED_HISTORY_PATTERNS: list[str] = [
+MALFORMED_HISTORY_PATTERNS: Final[list[str]] = [
     "tool_use ids were found without `tool_result` blocks immediately after",
     # Anthropic backtick variant
     "`tool_use` ids were found without `tool_result` blocks immediately after",
@@ -55,6 +55,30 @@ MALFORMED_HISTORY_PATTERNS: list[str] = [
     # OpenAI-compatible providers may reject replayed assistant tool calls whose
     # arguments are not valid JSON.
     "failed to parse tool call arguments as json",
+]
+
+# Vertex AI (Gemini) rejects context-caching requests when the cached content
+# is below the provider's minimum token threshold (currently 4096 tokens).
+# Example error: "The cached content is of 1171 tokens. The minimum token
+# count to start caching is 4096." — the `.lower()` comparison handles case
+# variation across providers but won't match reworded messages; update this
+# pattern if the API phrasing changes.
+PROMPT_CACHE_TOO_SMALL_PATTERNS: Final[list[str]] = [
+    "minimum token count to start caching",
+]
+
+AUTH_PATTERNS: Final[list[str]] = [
+    "invalid api key",
+    "unauthorized",
+    "missing api key",
+    "invalid authentication",
+    "access denied",
+]
+
+CONTENT_POLICY_PATTERNS: Final[list[str]] = [
+    "content_policy",
+    "content filtering policy",
+    "output blocked by content filtering",
 ]
 
 
@@ -89,25 +113,6 @@ def looks_like_malformed_conversation_history_error(exception: Exception) -> boo
     return any(p in s for p in MALFORMED_HISTORY_PATTERNS)
 
 
-# Vertex AI (Gemini) rejects context-caching requests when the cached content
-# is below the provider's minimum token threshold (currently 4096 tokens).
-# Example error: "The cached content is of 1171 tokens. The minimum token
-# count to start caching is 4096." — the `.lower()` comparison handles case
-# variation across providers but won't match reworded messages; update this
-# pattern if the API phrasing changes.
-PROMPT_CACHE_TOO_SMALL_PATTERNS: list[str] = [
-    "minimum token count to start caching",
-]
-
-AUTH_PATTERNS: list[str] = [
-    "invalid api key",
-    "unauthorized",
-    "missing api key",
-    "invalid authentication",
-    "access denied",
-]
-
-
 def is_prompt_cache_too_small(exception: Exception) -> bool:
     """Return True if the error indicates the prompt cache content is too small.
 
@@ -137,16 +142,6 @@ def looks_like_auth_error(exception: Exception) -> bool:
         if code in s:
             return True
     return False
-
-
-# Output/input content-policy blocks. Anthropic raises this as a 400 with
-# "Output blocked by content filtering policy"; the typed class may be flattened
-# to a plain BadRequestError by proxies/aliased providers, hence the text fallback.
-CONTENT_POLICY_PATTERNS: Final[list[str]] = [
-    "content_policy",
-    "content filtering policy",
-    "output blocked by content filtering",
-]
 
 
 def is_content_policy_violation(exception: Exception) -> bool:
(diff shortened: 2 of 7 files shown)


openhands-sdk/openhands/sdk/agent/agent.py
```diff
@@ -607,6 +608,28 @@ def step(
...
+                MessageEvent(
+                    source="user",
+                    llm_message=Message(
+                        role="user",
+                        content=[
+                            TextContent(
+                                text=(
+                                    "Your previous response was blocked by the "
+                                    "model's content filter. Please continue, "
+                                    "rephrasing to avoid the flagged content."
```
enyst, line 625: 😂

openhands-sdk/openhands/sdk/llm/exceptions/classifier.py
```diff
@@ -54,6 +57,30 @@
     "failed to parse tool call arguments as json",
 ]
 
+# Vertex AI (Gemini) rejects context-caching requests when the cached content
+# is below the provider's minimum token threshold (currently 4096 tokens).
+# Example error: "The cached content is of 1171 tokens. The minimum token
+# count to start caching is 4096." — the `.lower()` comparison handles case
+# variation across providers but won't match reworded messages; update this
+# pattern if the API phrasing changes.
+PROMPT_CACHE_TOO_SMALL_PATTERNS: Final[list[str]] = [
```
enyst, line 66: @openhands Look at this piece of code. This PR did not add it, just moved it around a bit. Look in history when was it added; I believe we added Vertex explicit caching at some point in time (possibly the same PR that added this, possibly a bit earlier) and then, we actually removed Vertex explicit caching (probably after this piece of code was added). Establish chronology here.

Look in the codebase in main branch: do we still have Vertex explicit caching for Gemini? If not, clean this out in a new branch from main.

Do not modify this PR, it’s a different issue: I just feel we might be able to clean up some tech debt.

enyst (approved):

openhands-sdk/openhands/sdk/llm/exceptions/classifier.py
enyst: Confirmed and cleaned up separately without modifying this PR.

Chronology I found:
- #3090 (`12ec83fd`, 2026-05-07) enabled Gemini explicit `cache_control` prompt-cache markers.
- #3480 (`29820ee7`, 2026-06-08, with earlier work in `6fb1bf54`/`8362654c`) added the Vertex/Gemini "minimum token count to start caching" classifier and retry-without-cache path.
- #3583 (`6f12ac88`) attempted a `vertex_cached_content` explicit Vertex caching config, but it was closed unmerged and is not in `main`.
- #3586 (`e6359545`/`c07f24c`, 2026-06-09) removed Gemini explicit `cache_control` markers from `main`; `model_features.py` now explicitly says not to add Gemini because it should use implicit prefix caching.

On current `main`, I found no `vertex_cached_content` support and no Gemini entry in `PROMPT_CACHE_MODELS`; only the orphaned prompt-cache-too-small classifier/retry/tests remained.

Cleanup is in new branch `cleanup/remove-obsolete-vertex-cache-retry` from `main`, commit `39c5a1ba`, with draft PR #3878: https://github.com/OpenHands/software-agent-sdk/pull/3878

_AI disclosure: This comment was written by an AI agent (OpenHands) on behalf of the user._

PR #3894, opened 2026-06-26
feat(mcp): subscribe to tools/list_changed for progressive-disclosure servers

### LINKED ISSUE #4059: Subscribe to MCP tools/list_changed notifications
## Summary

Support progressive-disclosure MCP servers by subscribing to tools/list_changed notifications, refreshing the advertised tool set, and registering newly available runtime tools safely.

## Tracking

PR: https://github.com/OpenHands/software-agent-sdk/pull/3894

## PR evidence acceptance criteria

- [ ] **Current `main` evidence:** Connect current main to a progressive-disclosure MCP server that advertises one tool, later emits tools/list_changed, and then exposes a second tool; show that the runtime tool set remains stale.
- [ ] **PR branch evidence:** Connect the PR branch to the same server, emit the notification, verify the advertised set refreshes exactly once, the new tool is registered and invocable, and removed/duplicate tools are handled safely.
- [ ] **Required evidence package:** The PR records both commit SHAs and exact environment/setup. Run the full relevant backend, service, infrastructure, or data pipeline live with PR code and attach commands plus output/logs; a movie or video is not a substitute for this live run. Run the real MCP server and SDK together and attach JSON-RPC traffic plus invocation output.

neubig:
HUMAN:

I have verified that this works with the datadog MCP in agent canvas

Screenshot 2026-07-13 at 9 39 21 PM

---

AGENT:

## Why

Some MCP servers (for example Datadog's hosted MCP) use progressive disclosure: they expose a small gateway toolset at connect time and register additional tools only after a skill-loading tool is invoked, signalling the change with `notifications/tools/list_changed`. The SDK previously listed tools once at connect time and never re-listed, so dynamically advertised tools stayed invisible and were never callable by the agent.

## Summary

- `create_mcp_tools` now accepts an `on_tools_changed` callback and subscribes to `notifications/tools/list_changed` through `_ToolListChangedHandler`. The conversation MCP provider forwards that callback to `AgentBase._on_mcp_tools_changed`, which registers newly advertised tools with `add_runtime_tools`.
- `_refresh_tools(client, on_tools_changed)` re-lists and reconciles `client._tools`, preserving existing tool definitions, adding newly advertised tools, dropping removed client-side definitions, and reporting only newly added tools to the callback.
- `tools_map` returns a snapshot copy so readers are isolated from runtime tool-map mutation on the MCP client's background event-loop thread.
- Follow-up fix in `1f772898`: schedule the list refresh outside the MCP receive-loop message handler. Awaiting `list_tools()` inline deadlocked because the same receive loop had to process the `tools/list` response.
- Added regression coverage in `tests/sdk/mcp/test_mcp_tool_list_changed.py`, including a real FastMCP streamable-HTTP server that sends `notifications/tools/list_changed`, refreshes the SDK client, and calls the newly advertised `extra` tool.

## Live evidence

Run on 2026-07-13 UTC on Linux 6.8.0-134-generic x86_64, Python 3.13.13, uv 0.11.8, and negotiated MCP protocol `2025-11-25`. The exact reusable harness is committed as `.pr/live_progressive_mcp_evidence.py`.

The harness starts a real progressive-disclosure FastMCP streamable-HTTP server under Uvicorn and a real SDK `Conversation`. The server initially advertises `gateway`, `register_extra_tool`, `repeat_list_changed`, and `remove_extra_tool`; `register_extra_tool` later registers `extra` and emits `notifications/tools/list_changed`. ASGI middleware records sanitized JSON-RPC directions, IDs, and methods. No credential values or tool arguments are logged.

Command in each checkout:

```bash
OPENHANDS_SUPPRESS_BANNER=1 uv run python .pr/live_progressive_mcp_evidence.py
```

For current main, the committed harness was invoked by its PR-worktree path while `uv run` supplied current-main packages.

### Current `main`: notification received, runtime remains stale

Commit: `777671766f4196da318b5b1e6179a6dc897cad36`

Sanitized JSON-RPC sequence:

```text
client->server id=0 initialize
client->server id=null notifications/initialized
client->server id=1 tools/list
client->server id=2 tools/call
server->client id=null notifications/tools/list_changed
```

Observed after waiting eight seconds:

```json
{
  "register_result": "[Tool 'register_extra_tool' executed.] | registered",
  "extra_appeared": false,
  "registration_logs": [],
  "after_register_agent_tools": [
    "finish",
    "gateway",
    "inspect_image_with_vision",
    "register_extra_tool",
    "remove_extra_tool",
    "repeat_list_changed",
    "think"
  ],
  "server_stopped": true
}
```

Current main receives the server notification but sends no second `tools/list`, so the conversation runtime never learns about `extra`.

### PR head: refresh, registration, invocation, duplicate, and removal

Commit: `60506089e8543fe23ec478ea8f73f7d85183e580`

The same initial sequence is immediately followed by `client->server id=3 tools/list`, then one runtime registration:

```text
Registered 1 dynamically advertised MCP tools: extra
```

Observed live behavior:

```json
{
  "extra_appeared": true,
  "extra_result": "[Tool 'extra' executed.] | 42",
  "extra_is_error": false,
  "registration_logs": [
    "Registered 1 dynamically advertised MCP tools: extra"
  ],
  "registration_logs_after_duplicate": [
    "Registered 1 dynamically advertised MCP tools: extra"
  ],
  "removed_from_client": true,
  "client_tools_after_removal": [
    "gateway",
    "register_extra_tool",
    "remove_extra_tool",
    "repeat_list_changed"
  ],
  "stale_removed_invocation_is_error": true,
  "stale_removed_invocation": "[Tool 'extra' executed.] | Unknown tool: 'extra'",
  "server_stopped": true
}
```

After a duplicate notification, JSON-RPC contains one additional `tools/list` but the registration-log list remains length one, proving the tool was not registered twice. After removal, the next notification produces another `tools/list`, drops `extra` from the live MCP client, and a retained agent snapshot definition fails safely as an error observation rather than raising or corrupting the session.

### Focused validation

```bash
uv run pytest -q tests/sdk/mcp/test_mcp_tool_list_changed.py
uv run pre-commit run --files .pr/live_progressive_mcp_evidence.py
```

Observed: `10 passed in 0.86s`; all pre-commit hooks passed.

## How to Test

1. `uv sync --group dev`
2. `uv run pytest tests/sdk/mcp/test_mcp_tool_list_changed.py -q` — 10 passed.
3. `uv run pytest tests/sdk/mcp/ tests/sdk/conversation/test_local_conversation_plugins.py -q` — 140 passed.
4. `env -u CODEX_HOME uv run pytest tests/sdk/agent/ -q` — 743 passed. Without unsetting `CODEX_HOME`, two ACP environment tests fail because this Codex runtime exports `CODEX_HOME`.
5. `uv run pre-commit run --files openhands-sdk/openhands/sdk/mcp/utils.py tests/sdk/mcp/test_mcp_tool_list_changed.py` — passed Ruff format/lint, pycodestyle, Pyright, import rules, and tool registration checks.

Closes #4059


neubig:
Re-running CI to check if sdk-tests failure was flaky

neubig:
📁 **PR Artifacts Notice**

This PR contains a `.pr/` directory with PR-specific documents. This directory will be **automatically removed** when the PR is approved.

> For fork PRs: Manual removal is required before merging.

malhotra5:
@openhands /codereview-roasted post back the results as a github comment

malhotra5:
# 🔥 Code Review (Roasted)

🟡 **Acceptable** — Works, human-verified against Datadog's progressive-disclosure MCP, and the core reconciliation logic is sound. But there's a real latent divergence bug hiding behind a `try/except Exception`, a docstring that lies, and duplicated list-and-create logic that should have been one function. Merge after addressing the divergence + a missing test.

_This review was generated by an AI agent (OpenHands) on behalf of the PR author, in response to @malhotra5's `/codereview-roasted` request._

---

## [CRITICAL ISSUES]

None that block the main feature — the happy path (add → register → invoke) is correct and proven.

### 1. `openhands-sdk/openhands/sdk/mcp/utils.py` — `_refresh_tools` docstring is wrong + DRY violation with `_connect_and_list_tools`

The `_refresh_tools` docstring opens with:

> Called both at connect time and whenever the server sends a `notifications/tools/list_changed` notification.

It isn't. Connect time still calls the **separate** `_connect_and_list_tools` (line 317):

```python
client.call_async_from_sync(_connect_and_list_tools, timeout=timeout, client=client)
```

So you now have two functions doing the same dance — `list_tools()` → `MCPToolDefinition.create()` → extend `client._tools`. With an empty initial `client._tools`, `_refresh_tools(client, None)` is **identical** to `_connect_and_list_tools` (everything is "added", callback is `None` → no-op). Delete `_connect_and_list_tools` and call `_refresh_tools` at connect. One path, one truth, and the docstring stops being a lie. Right now the docstring will mislead the next reader into assuming connect and refresh share reconciliation logic — they don't, and they can drift.

### 2. `openhands-sdk/openhands/sdk/mcp/utils.py` — handler is installed for *every* client, even when `on_tools_changed=None`

```python
handler = _ToolListChangedHandler(client=None, ..., on_tools_changed=on_tools_changed)
client = MCPClient(config, log_handler=log_handler, message_handler=handler)
```

`create_mcp_tools` unconditionally wires `_ToolListChangedHandler`. That means **every** MCP client in the codebase — including ones that never opted into progressive disclosure — now re-lists and rewrites `client._tools` on the background event-loop thread whenever a server happens to emit `tools/list_changed`. For the common agent path that snapshots `client.tools` once at connect, this is invisible; for any consumer that holds a live reference to `client.tools` (or iterates `client` directly), it's a silent behavior change that wasn't there before this PR. Either gate the handler on `on_tools_changed is not None`, or call this out explicitly in the docstring. Right now it's an undocumented side effect on a public factory.

### 3. `openhands-sdk/openhands/sdk/agent/base.py` — remove → re-add silently diverges agent and client tool maps (real bug, untested)

This is the one I'd actually want fixed. Trace the lifecycle:

- The agent **never** removes from `_tools` (by design — `verify()` enforces "tools may only be added, never removed"). Confirmed: no `_tools.pop`/`del` anywhere in `base.py`/`agent.py`/`local_conversation.py`.
- `_refresh_tools` **does** drop removed tools from `client._tools`.
- So after a server removes `extra`: `client._tools` no longer has it, but `agent._tools` **still does** (stale definition bound to the now-removed executor).
- Now the server re-adds `extra`. `_refresh_tools` sees `extra` not in `existing_by_name` (client dropped it) → it goes into `added` → callback fires → `add_runtime_tools([extra])` → hits:

```python
existing = set(self._tools) & set(tool_names)
if existing:
    raise ValueError(f"Duplicate tool names found: {existing}")
```

→ `ValueError` → swallowed by the broad `except Exception` in `_on_mcp_tools_changed` → logged as a warning → **the freshly re-advertised tool is never registered in the agent.** The client has the new executor; the agent keeps the stale one. They've diverged, silently.

The PR tests removal (stale invocation fails safely as an error observation — good), but **never tests remove-then-re-add**. That's the exact cycle that breaks. Progressive-disclosure servers that toggle a tool off and on (or that race a re-list) will hit this. At minimum: add a test, and decide on behavior — either let `add_runtime_tools` replace-on-duplicate for the MCP-changed path, or have `_on_mcp_tools_changed` reconcile against the client's live set instead of blindly adding. Don't leave it as a swallowed warning.

### 4. Comments narrate the fix history

The skill is blunt about this: comments that describe *why a workaround exists* drift. In `on_tool_list_changed`:

```python
# The message handler runs on the MCP receive loop; awaiting list_tools()
# here would block the same loop that must process the list_tools response.
```

This is a "we used to await inline and it deadlocked (commit 1f77289)" comment. The *why* (don't block the receive loop) is genuinely useful and I'd keep a one-liner; the rest belongs in the commit message, not the source. Similarly the 6-line block in `add_runtime_tools` explaining "frozen pydantic model… the underlying dict is mutable… tools_map returns a snapshot copy" restates what the code two methods down already does. Trim to the invariant, drop the restatement.

## [STYLE NOTES]

Skipping — linters handle the rest.

## [TESTING GAPS]

- `tests/sdk/mcp/test_mcp_tool_list_changed.py` — **no remove → re-add test.** This is the gap behind issue #3 above. The current removal test proves a stale definition errors safely, but not that a re-added tool re-registers (or fails to). Add: remove `extra`, notify, re-add `extra`, notify, assert the agent's `extra` is the *new* definition and is callable. This test should fail today.
- `test_create_mcp_tools_wires_message_handler` reaches into `client._session_kwargs.get("message_handler")` — a fastmcp private attribute. If fastmcp renames `_session_kwargs`, this test breaks without production code being broken. The e2e test (`test_list_changed_notification_refreshes_client_tools`) already proves wiring behaviorally; consider dropping the introspection assertion or loosening it.

On the positive side: the test suite is **good** where it counts. Real FastMCP streamable-HTTP server, real notification delivery, real tool invocation asserting `"42"` in the output — not a pile of mocks asserting `add_runtime_tools` was called. The `_FakeClient` unit tests cover the diff/reconcile logic deterministically without transport timing. That's the right split.

## [EVIDENCE]

The PR description's evidence is strong and meets the bar: real FastMCP server under Uvicorn, sanitized JSON-RPC direction/id/method trace, before/after comparison against `main` (notification received but no second `tools/list` → runtime stale) vs PR head (refresh + registration + invocation + duplicate-dedup + removal), committed reusable harness at `.pr/live_progressive_mcp_evidence.py`, and a human has verified against the real Datadog hosted MCP. Test counts and pre-commit results are listed. No gaps here.

## [RISK ASSESSMENT]

The feature is well-contained and the happy path is proven, but this PR changes a core factory (`create_mcp_tools`) that **all** MCP consumers route through, and it now mutates `client._tools` on a background thread for every client (issue #2) and introduces an agent/client tool-map divergence that fails silently (issue #3). The threading model (refresh scheduled off the receive loop, `_refresh_lock` serializing) is correct. The main residual risk is the untested remove→re-add path and the undocumented behavior change for callback-less clients. Neither is a merge blocker on its own, but #3 should be fixed or explicitly documented before this ships to every MCP user.

## [VERDICT]

✅ **Worth merging** — with #3 (divergence) addressed or explicitly scoped, and ideally #1 (DRY/docstring). Core logic is sound and the progressive-disclosure feature works end-to-end.

## [KEY INSIGHT]

The client reconciles tool additions *and* removals, but the agent only ever adds — so any remove-then-re-add cycle desynchronizes the two maps behind a swallowed `ValueError`, and that cycle is the one path this PR never tests.

---

>
>
>

malhotra5 (approved):
Non block issues outlined in review imho, the overall strategy lgtm

PR #3901, opened 2026-06-28
fix(sdk): disable streaming for summarizer LLM call (#3902)

### LINKED ISSUE #3902: Conversation crashes with `Summarization LLM call failed: Streaming requires an on_token callback` (standard OpenHands agent, not ACP)
## Summary

Long-running conversations using the standard OpenHands agent (not ACP) crash with
this server-side error:

```
Summarization LLM call failed: Streaming requires an on_token callback
```

After it fires the conversation stops progressing; the browser then also surfaces
downstream errors (HTTP 400 on `/api/file/search_subdirs?path=%2Fprojects` and the
events WebSocket closing before it finishes handshaking). Reproduces repeatedly
across multiple independent conversations in the same environment.

## Environment

- Stack: `npm run dev` (agent-canvas dev stack, `dev:automation` mode)
- Frontend: `http://localhost:3001` (Vite dev server)
- Agent-server: `http://localhost:18000` (uvx-installed `openhands-agent-server`)
- Agent type: **standard OpenHands agent** — **not** ACP, Claude Code, Codex, or
  Gemini. (Search hits for similar wording all point at the ACP epic OpenHands/OpenHands#15746, but
  those issues are unrelated to this report.)
- Browser console shows the failures against `127.0.0.1:8000` (the ingress
  proxy) and `127.0.0.1:18000`.

## Conversations where it was observed

- `c91af9a1-1706-4828-b753-f6c08f6563b4`
- `7412654a-c89e-4485-8b95-51b834e69d91`
- `b892321b-7c1a-4576-9a8f-befb692bf5ec`

## Repro context (browser console)

Captured verbatim while the summarization error was happening on
`b892321b-7c1a-4576-9a8f-befb692bf5ec`:

```
Summarization LLM call failed: Streaming requires an on_token callback

127.0.0.1:8000/api/file/search_subdirs?path=%2Fprojects:1  Failed to load resource: the server responded with a status of 400 (Bad Request)
… (same 400 repeats ~10x in quick succession) …
es-h-A2ly90.js?v=…:9159 You're attempting to animate multiple children within AnimatePresence, but its mode is set to "wait". This will lead to odd visual behaviour.
use-websocket.ts:149 WebSocket connection to 'ws://127.0.0.1:8000/sockets/events/b892321b-7c1a-4576-9a8f-befb692bf5ec?resend_mode=since&after_timestamp=…' failed: WebSocket is closed before the connection is established.
```

The `/api/file/search_subdirs?path=%2Fprojects` 400s and the events WebSocket
handshake failures are almost certainly downstream symptoms — once the agent
runtime throws on summarization, the conversation stops accepting new events and
the frontend's recurring file-change probe and WS reconnect attempts both fail.
The primary bug is the summarization exception itself.

## Why this looks like an SDK bug, not a configuration bug

The error string `Streaming requires an on_token callback` is raised by the
LLM client layer in `software-agent-sdk` (it is the standard guard message
when a streaming completion is started without supplying the `on_token`
callback the SDK requires). It is being triggered from the summarization path
inside the agent-server (the message is `"Summarization LLM call failed: …"`),
not from the main agent loop, which suggests the **condenser / summarizer is
calling the LLM in streaming mode but not registering the required callback**.

So this likely needs:

1. Confirmation in `openhands-agent-server` / `openhands-sdk` that the
   summarizer/compressor is making a non-streaming call, OR
2. A fix that supplies an `on_token` callback (or otherwise disables streaming)
   when the summarization LLM call is invoked.

It would be great to know whether this is reproducible on the latest released
agent-server (1.27.0) and on the corresponding `main` of `software-agent-sdk`,
since the user is on a local dev stack and the failure looks deterministic
once a conversation grows large enough to trigger summarization.

## Workaround

None known — once the error fires the conversation appears unrecoverable and a
new conversation has to be started, which makes long tasks impossible.

## Notes

- This is **not** the ACP-on-cloud streaming-redaction work tracked in OpenHands/OpenHands#15746 /
  OpenHands/agent-canvas#2110 / OpenHands/OpenHands#15720 — those issues are about masking secrets in streamed ACP
  tokens; this is the standard OpenHands agent and the failure is in the
  summarization path.
- Filing on behalf of the user; the content was drafted by an AI agent.

VascoSch92:
HUMAN:

Fixing a bug with condenser and streaming.

---

AGENT:

## Why

Long-running conversations using the standard OpenHands agent crash with a
server-side error once they grow large enough to trigger summarization:

```
Summarization LLM call failed: Streaming requires an on_token callback
```

After it fires the conversation stops progressing and becomes unrecoverable
(downstream `search_subdirs` 400s and events-WebSocket handshake failures are
secondary symptoms). Reported in #3902 (originally surfaced from
`OpenHands/agent-canvas#1514`).

Root cause: `LLMSummarizingCondenser` calls `self.llm.completion()` /
`self.llm.acompletion()` **without** an `on_token` callback. When the
conversation's LLM is configured with `stream=True` (e.g. the agent-canvas
`dev:automation` stack), `LLM.completion()` enables streaming because
`enable_streaming = bool(kwargs.get("stream", False)) or self.stream`, and then
raises `ValueError("Streaming requires an on_token callback")`. The condenser
re-wraps that as `Summarization LLM call failed: ...`.

Every other auxiliary (non-interactive) LLM call in the SDK already guards
against this by forcing non-streaming — e.g. `conversation/goal/judge.py`
(`judge_llm.model_copy(update={"stream": False})`) and the delegated
sub-agent paths (`tests/tools/delegate/test_delegation.py::test_spawn_disables_streaming_for_sub_agents`).
The summarizer was the one caller that was missed.

## Summary

- Force a non-streaming copy of the LLM for the summary completion in both
  `_generate_condensation` (sync) and `_agenerate_condensation` (async),
  mirroring the existing `judge.py` idiom. The copy is only made when
  `self.llm.stream` is `True`, so non-streaming setups are unaffected.
- The agent's own LLM is left untouched — `model_copy` is non-mutating and the
  copy keeps the same `usage_id` and shares the underlying `_metrics`/
  `_telemetry` object (we deliberately do **not** `reset_metrics()` here, unlike
  the sub-agent path which wants independent tracking), so summary token usage
  is still attributed to the conversation.
- Add regression tests covering the streaming (sync + async) and non-streaming
  paths, including an assertion that summary token usage is recorded on the
  original LLM's metrics.

## Issue Number

Fixes #3902

## How to Test

```bash
uv sync --dev
# Full condenser suite (39 tests, includes the new ones):
uv run pytest tests/sdk/context/condenser/test_llm_summarizing_condenser.py -q
```

To confirm the new tests actually reproduce the bug, revert the source change
and run only the streaming regression tests — they fail with the exact
production error before the fix:

```bash
git stash push -- openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
uv run pytest tests/sdk/context/condenser/test_llm_summarizing_condenser.py \
  -q -k streaming
git stash pop
```

## Video/Screenshots

Deterministic CLI reproduction (no GUI involved). Captured locally:

**1. New regression tests FAIL on `main` (fix reverted) — exact issue error:**

```
tests/.../test_llm_summarizing_condenser.py:390: in _agenerate_condensation
    raise NoCondensationAvailableException(
E   NoCondensationAvailableException: Summarization LLM call failed: Streaming requires an on_token callback
...
FAILED ...::test_summarization_disables_streaming_when_llm_streams
FAILED ...::test_async_summarization_disables_streaming_when_llm_streams
2 failed, 1 passed, 36 deselected
```

**2. With the fix — full condenser suite passes:**

```
collected 39 items
tests/sdk/context/condenser/test_llm_summarizing_condenser.py ............................. [100%]
39 passed in 0.24s
```

**3. Token usage is still counted (no metrics loss):** running the real
`completion()` path through the stream-disabled copy records the response's
`Usage(prompt_tokens=10, completion_tokens=5)` onto the **original** LLM's
`metrics.accumulated_token_usage`, with `usage_id` preserved. Asserted in
`test_summarization_disables_streaming_when_llm_streams`.

**4. Broader sweep + hooks green:**

```
tests/sdk/context/  ->  406 passed
pre-commit (Ruff format, Ruff lint, pycodestyle, pyright, import rules) -> all Passed
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- A truly live end-to-end repro (full agent-canvas dev stack + a real provider
  key, driving a conversation until it auto-summarizes) was **not** run here —
  it needs provider credentials and a long conversation. Instead the bug is
  reproduced deterministically at the SDK layer: the regression tests use a
  real `LLM(stream=True)` with the transport mocked, so the genuine streaming
  guard in `completion()`/`acompletion()` executes and raises the identical
  error without the fix.
- Metrics pattern note: the sub-agent path (`subagent/registry.py`) copies the
  condenser LLM with a **distinct** `usage_id` and calls `reset_metrics()` on
  purpose, so a sub-agent's condenser cost is tracked independently. This PR
  intentionally does the opposite (same `usage_id`, shared metrics), matching
  `judge.py`, so summarization cost rolls up to the conversation that incurred
  it rather than being split into a separate bucket.
- The fix is intentionally minimal and local to the condenser. An alternative
  is to disable streaming once where the condenser's LLM is constructed
  (agent-server wiring), but that is broader and not guaranteed to cover every
  code path that builds a condenser.

### DIFF
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -188,10 +188,19 @@ class LLMSummarizingCondenser(RollingCondenser):
 
         messages = [Message(role="user", content=[TextContent(text=prompt)])]
 
+        # Force non-streaming: the summary is consumed whole and no on_token
+        # callback is passed, which a streaming LLM would require. Sharing the
+        # same usage_id/metrics keeps summary tokens counted (model_copy is
+        # non-mutating; unlike sub-agents we do not reset_metrics here).
+        summarizing_llm = (
+            self.llm.model_copy(update={"stream": False})
+            if self.llm.stream
+            else self.llm
+        )
         # Do not pass extra_body explicitly. The LLM handles forwarding
         # litellm_extra_body only when it is non-empty.
         try:
-            llm_response = self.llm.completion(
+            llm_response = summarizing_llm.completion(
                 messages=messages,
             )
         except Exception as e:
@@ -384,8 +393,14 @@ class LLMSummarizingCondenser(RollingCondenser):
         )
 
         messages = [Message(role="user", content=[TextContent(text=prompt)])]
+        # Force non-streaming for the summary (see _generate_condensation).
+        summarizing_llm = (
+            self.llm.model_copy(update={"stream": False})
+            if self.llm.stream
+            else self.llm
+        )
         try:
-            llm_response = await self.llm.acompletion(messages=messages)
+            llm_response = await summarizing_llm.acompletion(messages=messages)
         except Exception as e:
             raise NoCondensationAvailableException(
                 f"Summarization LLM call failed: {e}"
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -80,6 +80,17 @@ class LLMSummarizingCondenser(RollingCondenser):
             )
         return self
 
+    @model_validator(mode="after")
+    def _disable_streaming_for_summary(self):
+        # Summaries are consumed whole with no on_token callback, which a
+        # streaming LLM requires (#3902). Disable streaming once so every
+        # summary path is covered. model_copy is non-mutating and shares
+        # usage_id/metrics, so summary tokens stay attributed to the
+        # conversation.
+        if self.llm.stream:
+            self.llm = self.llm.model_copy(update={"stream": False})
+        return self
+
     def handles_condensation_requests(self) -> bool:
         return True
 
@@ -188,19 +199,10 @@ class LLMSummarizingCondenser(RollingCondenser):
 
         messages = [Message(role="user", content=[TextContent(text=prompt)])]
 
-        # Force non-streaming: the summary is consumed whole and no on_token
-        # callback is passed, which a streaming LLM would require. Sharing the
-        # same usage_id/metrics keeps summary tokens counted (model_copy is
-        # non-mutating; unlike sub-agents we do not reset_metrics here).
-        summarizing_llm = (
-            self.llm.model_copy(update={"stream": False})
-            if self.llm.stream
-            else self.llm
-        )
         # Do not pass extra_body explicitly. The LLM handles forwarding
         # litellm_extra_body only when it is non-empty.
         try:
-            llm_response = summarizing_llm.completion(
+            llm_response = self.llm.completion(
                 messages=messages,
             )
         except Exception as e:
@@ -393,14 +395,8 @@ class LLMSummarizingCondenser(RollingCondenser):
         )
 
         messages = [Message(role="user", content=[TextContent(text=prompt)])]
-        # Force non-streaming for the summary (see _generate_condensation).
-        summarizing_llm = (
-            self.llm.model_copy(update={"stream": False})
-            if self.llm.stream
-            else self.llm
-        )
         try:
-            llm_response = await summarizing_llm.acompletion(messages=messages)
+            llm_response = await self.llm.acompletion(messages=messages)
         except Exception as e:
             raise NoCondensationAvailableException(
                 f"Summarization LLM call failed: {e}"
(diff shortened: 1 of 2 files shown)


csmith49 (approved):
LGTM, but it might be worth it to move the LLM settings tweak to a `post_init` function or something so all uses have streaming disabled

VascoSch92:
@csmith49 thanks for the hint. 

I changed ;-) the fix is much better now.

PR #2196, opened 2026-02-24
fix: redact credentials from URLs in logs, exceptions, and persisted state

jpshackelford:
## Summary

Fixes #2152

This PR addresses a security issue where credentials embedded in plugin source URLs (e.g., `https://oauth2:SECRET_TOKEN@gitlab.com/org/private-repo`) were being exposed in:

1. **Application logs** - Clone URLs and git commands were logged with credentials visible
2. **Exception messages** - Error messages included full URLs with credentials
3. **Persisted conversation state** - Plugin source URLs with credentials were stored in `ResolvedPluginSource`
4. **Installed extension metadata** - Source URLs with credentials were stored in `.installed.json`

### Changes

- **New utility function**: `redact_url_credentials(url)` in `git/utils.py` that replaces credentials in URLs with `****` (e.g., `https://****@gitlab.com/org/repo`)
- **Updated `run_git_command()`**: Redacts credentials from error logs, debug logs, and `GitCommandError` exception messages/command fields
- **Updated `cached_repo.py`**: Redacts credentials in clone log messages
- **Updated `extensions/fetch.py`**: Redacts credentials in `ExtensionFetchError` exceptions and warning logs (this is the new location after refactoring)
- **Updated `plugin/types.py`**: `ResolvedPluginSource.from_plugin_source()` now redacts credentials before persistence (credentials are only needed at fetch time)
- **Updated `plugin/loader.py`**: Redacts credentials in plugin loading log messages
- **Updated `local_conversation.py`**: Redacts credentials in plugin loaded debug messages
- **Updated `extensions/installation/info.py`**: Redacts credentials in `InstallationInfo.from_extension()` before persisting to `.installed.json`
- **Public API export**: `redact_url_credentials` is exported from `openhands.sdk` for cross-repo usage (related to OpenHands/OpenHands#12959)

### Example

Before:
```
INFO - Cloning repository from https://oauth2:SECRET_TOKEN@gitlab.com/org/repo.git
```

After:
```
INFO - Cloning repository from https://****@gitlab.com/org/repo.git
```

### Code Review Feedback Addressed

This update addresses all the issues identified in the [code review by @enyst](https://github.com/OpenHands/software-agent-sdk/pull/2196#pullrequestreview-3950231707):

- ✅ `plugin/loader.py` - now redacts `spec.source` in log messages
- ✅ `local_conversation.py` - now redacts `spec.source` in debug log messages  
[...]

PR #3909, opened 2026-06-29
fix(sdk): redact plugin and extension source credentials at the serialization boundary (#3752)

### LINKED ISSUE #3752: Redact plugin/extension source credentials at a single serialization choke point (follow-up to #2196 review, sibling of OpenHands#12959)

simonrosenberg:
HUMAN:

Taking over the stale community PR #3786 (thanks @haya-asif). After re-reading Vasco's review I went with the minimal version: one serializer that masks inline credentials but preserves `${VAR}` refs, instead of the secret-convention machinery (expose modes + encrypt-at-rest). Same security outcome for the paths that matter, far less surface area. Verified green locally on latest main.

---

AGENT:

## Why

Closes #3752. Supersedes #3786 (original work by @haya-asif).

Inline credentials in a plugin `source` URL (e.g. `https://oauth2:TOKEN@host/repo.git`)
leak into persisted state (conversation `meta.json`), the auto `plugins` tag, the
remote create payload, and ordinary model dumps, because `PluginSource.source`
serializes verbatim. This is the SDK-side sibling of APP-904
(OpenHands/OpenHands#14795, which added a `@field_serializer('source')` to the app
repo's `PluginSpec` with a `# TODO` to switch to the SDK's `redact_url_credentials`
once available — this closes that loop).

The one subtlety, per the #3786 / #2196 reviews: **`${VAR}` placeholders are not
secrets.** They are expanded from the secret registry at fetch time (including
server-side, where `_expand_plugin_source_ref` reads the registry), and they are
how private plugins are meant to authenticate. So they must survive serialization;
only genuine inline credentials should be masked.

## Summary

One choke point, no new secret API:

- **`redact_url_credentials(url, *, preserve_placeholders=False)`** gains a flag that
  leaves a `${VAR}`-bearing userinfo intact while still masking real inline
  credentials.
- **`PluginSource`** gets a single `@field_serializer("source")` that masks the
  userinfo (`https://****@host`) on every dump and keeps `${VAR}` refs. The raw value
  stays on the attribute, so fetch/clone is unaffected — only serialized forms
  (`meta.json`, the `plugins` tag, the remote payload) are masked. This covers all of
  Vasco's leak paths at once.
- The auto **`plugins` tag** is masked the same way.
- **`ResolvedPluginSource`** keeps redacting at construction, so direct attribute
  reads (example scripts, etc.) stay safe — no attribute-access leak introduced.

Deliberate non-changes:

- **Extensions (`.installed.json`) are left untouched.** `InstallationInfo.source`
  is serialized in exactly one place — `save_to_dir` → `.installed.json` — and
  `ExtensionManager.update()` re-clones from it. Extensions have no `${VAR}`
  mechanism and the layer has no cipher, so the source must stay intact at rest or
  private-extension updates break (Vasco's regression #2). Redacting it there would
  only break `update()` without closing the on-disk credential, so it's out of scope;
  a regression guard locks the re-clone behavior in. Encrypting extension metadata at
  rest is a separate follow-up.
- The plugin-load log leak (issue point 3) is already gone on `main` — the log sites
  print only `plugin.manifest.name` + `resolved_ref[:8]`. A guard test keeps it that
  way.

What this drops vs #3786: the `serialize_credential_url`/`validate_credential_url`
secret API, the `cipher` encrypt-at-rest mode, the before-validator decrypt, and the
`expose_secrets` plumbing on the remote payload and `save_to_dir`. The encrypt-at-rest
mode only helped *inline-credentialed* plugins survive a cacheless resume — but inline
credentials are exactly what we want out of persisted state, and `${VAR}` (which
survives) is the supported channel.

## Trade-off

An inline-credentialed *private* plugin will no longer re-clone on a remote/fresh-pod
resume — it must use a `${VAR}` reference instead. Those URLs were already leaking
their token into `meta.json`/tags/the payload before this PR, so the secure migration
is to `${VAR}`, which works end to end.

## Issue Number

Closes #3752.

## How to Test

```bash
uv run pytest tests/sdk/git/test_url_redaction.py tests/sdk/extensions/installation \
  tests/sdk/conversation/test_local_conversation_plugins.py \
  tests/workspace/test_cloud_workspace_automation_tags.py
uv run ruff check . && uv run pyright
```

New tests cover each live path: `${VAR}` survives the default dump (the case #3786
never guarded), inline creds masked while the URL shape is kept, the in-memory value
stays raw, non-credential sources unchanged, the redacted `plugins` tag, the remote
payload (mask inline / keep `${VAR}`), the plugin-load log guard, and the
`update()` re-clone guard. Locally: targeted + full `tests/sdk` and the agent-server
persistence/plugin suites pass; ruff + pyright clean.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:ce50187-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-ce50187-python \
  ghcr.io/openhands/agent-server:ce50187-python
```

```
ghcr.io/openhands/agent-server:ce50187-golang-amd64
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-golang-amd64
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-golang-amd64
ghcr.io/openhands/agent-server:ce50187-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:ce50187-golang-arm64
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-golang-arm64
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-golang-arm64
ghcr.io/openhands/agent-server:ce50187-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:ce50187-java-amd64
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-java-amd64
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-java-amd64
ghcr.io/openhands/agent-server:ce50187-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:ce50187-java-arm64
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-java-arm64
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-java-arm64
ghcr.io/openhands/agent-server:ce50187-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:ce50187-python-amd64
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-python-amd64
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-python-amd64
ghcr.io/openhands/agent-server:ce50187-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:ce50187-python-arm64
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-python-arm64
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-python-arm64
ghcr.io/openhands/agent-server:ce50187-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:ce50187-golang
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-golang
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-golang
ghcr.io/openhands/agent-server:ce50187-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:ce50187-java
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-java
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-java
ghcr.io/openhands/agent-server:ce50187-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:ce50187-python
ghcr.io/openhands/agent-server:ce50187ba047b2b4cbe8a057f133bb01c8fcd388-python
ghcr.io/openhands/agent-server:fix-redact-plugin-source-creds-python
ghcr.io/openhands/agent-server:ce50187-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `ce50187-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `ce50187-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/conversation.py
@@ -20,6 +20,7 @@ from openhands.sdk.logger import get_logger
 from openhands.sdk.plugin import PluginSource
 from openhands.sdk.secret import SecretValue
 from openhands.sdk.tool.client_tool import ClientToolSpec
+from openhands.sdk.utils.redact import redact_url_credentials
 from openhands.sdk.workspace import LocalWorkspace, RemoteWorkspace
 
 
@@ -174,7 +175,12 @@ class Conversation:
 
             # 2. Auto-generate plugins/skills tag from plugins parameter
             if plugins:
-                plugin_urls = [p.source_url for p in plugins if p.source_url]
+                # tags persist verbatim — mask inline creds (${VAR} refs survive).
+                plugin_urls = [
+                    redact_url_credentials(url, preserve_placeholders=True)
+                    for p in plugins
+                    if (url := p.source_url)
+                ]
                 if plugin_urls:
                     effective_tags["plugins"] = ",".join(plugin_urls)
                     logger.debug(f"Added plugins tag with {len(plugin_urls)} plugin(s)")
--- a/openhands-sdk/openhands/sdk/plugin/types.py
+++ b/openhands-sdk/openhands/sdk/plugin/types.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import frontmatter
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, field_serializer, field_validator
 
 from openhands.sdk.utils.path import to_posix_path
 from openhands.sdk.utils.redact import redact_url_credentials
@@ -63,6 +63,15 @@ class PluginSource(BaseModel):
                 "repo_path cannot contain '..' (parent directory traversal)"
             )
         return v
+
+    @field_serializer("source", when_used="always")
+    def _redact_source(self, source: str) -> str:
+        """Mask inline URL credentials on dump; ${VAR} refs survive (not secrets).
+
+        The raw value stays on the attribute for fetch/clone — only serialized
+        forms (persisted state, the plugins tag, the remote payload) are masked.
+        """
+        return redact_url_credentials(source, preserve_placeholders=True)
 
     @property
     def source_url(self) -> str | None:
--- a/openhands-sdk/openhands/sdk/utils/redact.py
+++ b/openhands-sdk/openhands/sdk/utils/redact.py
@@ -138,19 +138,17 @@ def http_error_log_content(response: httpx.Response) -> str | dict:
         return f"<non-JSON response body omitted ({body_len} chars)>"
 
 
-def redact_url_credentials(url: str) -> str:
+def redact_url_credentials(url: str, *, preserve_placeholders: bool = False) -> str:
     """Redact credentials embedded in a URL (e.g. https://user:token@host).
 
     Replaces the ``user:password`` or bare-token portion before the ``@`` with
     ``****`` so the URL is safe for logs, error messages, and persisted state.
     SSH URLs (``git@host``) and credential-free URLs are returned unchanged.
 
-    Args:
-        url: A URL that may contain embedded credentials.
-
-    Returns:
-        URL with the credential portion replaced by ``****``, or the original
-        URL if no embedded credentials are found.
+    With ``preserve_placeholders``, a userinfo holding a ``${VAR}`` reference is
+    left intact: it is expanded from the secret registry at fetch time and is
+    not itself a secret, so masking it would break the clone. Inline credentials
+    are still masked.
 
     Examples:
         >>> redact_url_credentials("https://oauth2:SECRET@gitlab.com/repo.git")
@@ -159,11 +157,17 @@ def redact_url_credentials(url: str) -> str:
         'https://github.com/owner/repo.git'
         >>> redact_url_credentials("git@github.com:owner/repo.git")
         'git@github.com:owner/repo.git'
+        >>> redact_url_credentials(
+        ...     "https://x-token-auth:${TOKEN}@host/r.git", preserve_placeholders=True
+        ... )
+        'https://x-token-auth:${TOKEN}@host/r.git'
     """
     match = re.match(r"^(https?://)([^@/]+)@(.+)$", url)
-    if match:
-        return f"{match.group(1)}****@{match.group(3)}"
-    return url
+    if not match:
+        return url
+    if preserve_placeholders and "${" in match.group(2):
+        return url
+    return f"{match.group(1)}****@{match.group(3)}"
 
 
 # Matches embedded ``http(s)://user:token@`` credentials anywhere within a
--- a/tests/sdk/conversation/remote/test_remote_conversation.py
+++ b/tests/sdk/conversation/remote/test_remote_conversation.py
@@ -256,6 +256,41 @@ class TestRemoteConversation:
     @patch(
         "openhands.sdk.conversation.impl.remote_conversation.WebSocketCallbackClient"
     )
+    def test_remote_conversation_plugin_source_redacted_placeholder_kept(
+        self, mock_ws_client
+    ):
+        """The create payload masks inline plugin-source creds but keeps ${VAR}
+        placeholders, so the server clones private plugins via secret expansion
+        without raw credentials crossing the wire."""
+        from openhands.sdk.plugin import PluginSource
+
+        conversation_id = str(uuid.uuid4())
+        mock_client_instance = self.setup_mock_client(conversation_id=conversation_id)
+        mock_ws_client.return_value = Mock()
+
+        placeholder = "https://x-token-auth:${MY_TOKEN}@host/org/repo.git"
+        RemoteConversation(
+            agent=self.agent,
+            workspace=self.workspace,
+            plugins=[
+                PluginSource(source="https://oauth2:LEAKME@host/org/priv.git"),
+                PluginSource(source=placeholder),
+            ],
+        )
+
+        create_call = next(
+            call
+            for call in mock_client_instance.request.call_args_list
+            if call[0][0] == "POST" and call[0][1] == "/api/conversations"
+        )
+        sources = [p["source"] for p in create_call.kwargs["json"]["plugins"]]
+        assert "https://****@host/org/priv.git" in sources
+        assert placeholder in sources
+        assert "LEAKME" not in str(create_call.kwargs["json"]["plugins"])
+
+    @patch(
+        "openhands.sdk.conversation.impl.remote_conversation.WebSocketCallbackClient"
+    )
     def test_remote_conversation_user_id_none_sends_explicit_null(self, mock_ws_client):
         """user_id=None sends an explicit null key (not omitted) so the server
         receives a consistent payload regardless of whether user_id was supplied."""
--- a/tests/sdk/conversation/test_local_conversation_plugins.py
+++ b/tests/sdk/conversation/test_local_conversation_plugins.py
@@ -1463,3 +1463,37 @@ class TestAmbientPluginAutoLoad:
         assert "ambient-skill" in skill_names
         assert conversation.resolved_plugins is None
         conversation.close()
+
+    def test_plugin_load_log_never_leaks_credentials(
+        self, tmp_path: Path, basic_agent, caplog: pytest.LogCaptureFixture
+    ):
+        """Plugin-load logs must never contain the source credential. A serializer
+        covers model dumps, not f-string log lines, so this guards against anyone
+        re-adding spec.source to that log (issue #3752)."""
+        plugin_dir = create_test_plugin(
+            tmp_path / "plugin",
+            name="test-plugin",
+            skills=[{"name": "s", "content": "c"}],
+        )
+        workspace = tmp_path / "workspace"
+        workspace.mkdir()
+        conversation = LocalConversation(
+            agent=basic_agent,
+            workspace=workspace,
+            plugins=[
+                PluginSource(source="https://oauth2:LEAKME@host.example.com/o/r.git")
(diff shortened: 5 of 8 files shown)


openhands-sdk/openhands/sdk/utils/pydantic_secrets.py
```diff
@@ -195,3 +196,31 @@ def _decrypt_env(cls, value, info):
...
+        return source
+    if mode == "encrypted":
+        cipher: Cipher | None = info.context.get("cipher") if info.context else None
+        if cipher is None:
+            raise MissingCipherError(
+                "Cannot encrypt credentialed source: no cipher configured. "
+                "Set OH_SECRET_KEY environment variable."
+            )
+        token = cipher.encrypt(SecretStr(source))
+        assert token is not None  # encrypt() returns None only for a None input
```
VascoSch92: Do we need this?

openhands-sdk/openhands/sdk/extensions/installation/metadata.py
```diff
@@ -159,7 +159,9 @@ def save_to_dir(self, installed_dir: Path) -> None:
         """Save metadata to the installed extensions directory."""
         metadata_path = self.get_metadata_path(installed_dir)
         metadata_path.parent.mkdir(parents=True, exist_ok=True)
-        metadata_path.write_text(self.model_dump_json(indent=2))
+        metadata_path.write_text(
```
VascoSch92: are we still storing credentials in plaintext?

VascoSch92 (approved):
Just a couple of comments.

openhands-sdk/openhands/sdk/utils/pydantic_secrets.py
simonrosenberg: Good call — no. I dropped the whole `serialize_credential_url`/`validate_credential_url` secret API in the latest push. `PluginSource.source` now just uses a single `@field_serializer` over `redact_url_credentials(..., preserve_placeholders=True)`: inline creds are masked on every dump, `${VAR}` refs survive (they're expanded server-side from the secret registry, so the recommended private-plugin path still clones), and the raw value stays on the attribute for fetch. No cipher / expose-mode plumbing. See the updated PR description.

openhands-sdk/openhands/sdk/extensions/installation/metadata.py
simonrosenberg: Yes — `.installed.json` stays plaintext, but that's now a deliberate, scoped decision rather than something this PR moves. I reverted the `save_to_dir` expose change. Extensions have no `${VAR}` mechanism and the installation layer has no cipher, so `ExtensionManager.update()` needs the real source at rest to re-clone — redacting it there would break private-extension updates (your regression #2) without actually removing the on-disk credential. I added a guard test (`test_update_reclones_with_credentialed_source`) to lock the re-clone behavior in, and called encrypt-at-rest for extension metadata out as a separate follow-up. Happy to file that issue if you'd like it tracked.

simonrosenberg:
@VascoSch92 Thanks for the review! I simplified the code

PR #3927, opened 2026-06-30
fix(sdk): accept 6+ digit event indices in EventLog filename regex

### LINKED ISSUE #3926: bug: EventLog silently drops events and corrupts the log once event index reaches 100000 (5-digit regex vs unbounded writer)
## Problem

Event files are persisted as `event-{idx}-{event_id}.json`. The **writer** and the **reader** disagree on how wide the `idx` field is, and the two diverge permanently at the 100,000th event:

`openhands-sdk/openhands/sdk/conversation/persistence_const.py:6-9`
```python
EVENT_NAME_RE = re.compile(
    r"^event-(?P<idx>\d{5})-(?P<event_id>[0-9a-fA-F\-]{8,})\.json$"  # reader: EXACTLY 5 digits
)
EVENT_FILE_PATTERN = "event-{idx:05d}-{event_id}.json"              # writer: MINIMUM 5 digits
```

`{idx:05d}` zero-pads to a **minimum** of 5 digits — it never caps — so `idx=100000` is written as `event-100000-...` (six digits). The reader regex `\d{5}` matches **exactly** five digits that must be immediately followed by `-`; for `event-100000-...` it consumes `10000`, then expects `-` but sees the sixth digit `0`, so the file **does not match**.

From the 100,000th event onward, event files are **written but cannot be read back**. The mismatch only bites on a **cold reload** (process restart → resume / fork), because in-process reads use the in-memory `_idx_to_id` map and never re-parse filenames. The cold path is `EventLog.__init__` → `_scan_and_build_index()`.

The event log is **append-only and never pruned** — condensation only marks events as forgotten in the `View`, it does not delete files (`grep` for `delete|prune|truncate|unlink` in `event_store.py` is empty). So a long-lived, persisted conversation grows monotonically and will eventually cross 100,000 events.

## What happens on reload

`openhands-sdk/openhands/sdk/conversation/event_store.py:229-261` (`_scan_and_build_index`)
```python
for p in paths:
    m = EVENT_NAME_RE.match(name)
    if m:
        by_idx[int(m.group("idx"))] = m.group("event_id")
    else:
        logger.warning(f"Unrecognized event file name: {name}")   # every idx >= 100000 lands here
...
n = 0
while True:
    if n not in by_idx:
        if any(i > n for i in by_idx.keys()):
            logger.warning("Event index gap detected: ...")        # it NOTICES, then proceeds anyway
        break
    n += 1
# rebuilds only range(n) -> length capped at 100000
```

1. Every file with `idx >= 100000` falls into the `else` branch and is silently excluded from `by_idx`.
2. The contiguity loop stops at `n = 100000`, so `len(EventLog)` returns **100000** even though 100002 files exist — silent data loss. Notably the code **detects the gap and logs a warning, then returns the truncated length anyway**.
3. The count path diverges from the scan path: `_count_events_on_disk()` (`event_store.py:172-186`) counts by **prefix** (`startswith("event-")`), so it still sees all 100002 files. On the next `append()`, `disk_length (100002) > self._length (100000)` triggers `_sync_from_disk(100002)`, which sets `self._length = max(scanned=100000, disk=100002)`. Length accounting and the index now permanently disagree — idx 100000/100001 remain on disk but are unreadable holes, and the log is irreversibly corrupted.

## Reproduction (deterministic, offline)

`tests/unit/bug_repro/poc_event_index_overflow.py` — writes 100,002 contiguous event files using the writer's own `EVENT_FILE_PATTERN`, then cold-loads them through a fresh `EventLog`. Real output:

```
PART 1 — writer pattern vs reader regex (pure, no I/O)
  idx=    99999  written=event-99999-...json    reader regex matches it? True
  idx=   100000  written=event-100000-...json   reader regex matches it? False
  idx=   100001  written=event-100001-...json   reader regex matches it? False
  idx=  1000000  written=event-1000000-...json  reader regex matches it? False

PART 2 — end-to-end: cold reload of a >100k-event log
  event files physically on disk : 100002
  len(EventLog) after cold scan  : 100000   (should be 100002)
  EVENTS LOST ON RELOAD          : 2
  read idx 0     -> OK
  read idx 99999 -> OK
  read idx 100000 -> LOST: IndexError: Event index out of range
  read idx 100001 -> LOST: IndexError: Event index out of range

PART 3 — appending after reload corrupts length accounting
  len(EventLog) after one append : 100003
  -> idx 100000/100001 remain permanent unreadable holes on disk
```

(`InMemoryFileStore`'s default LRU caps — `max_size=100_000`/20 MB, `io/memory.py:15-16` — are raised in the PoC; that configures only the storage test-double, not the code under test. A real on-disk `LocalFileStore` has no such cap.)

## Why this is a bug, not intended behavior

- **"100k events is unrealistic / the condenser trims history."** The condenser only marks `forgotten_event_ids` in the `View` (`event/condenser.py`); it never deletes event files. `EventLog` is strictly append-only (no delete/prune anywhere in `event_store.py`). A long-lived persisted server conversation — exactly the use case persistence/resume exists for — grows without bound and reaches 100k events over its lifetime.
- **"`:05d` and `\d{5}` are both just the 5-digit convention."** If a 5-digit cap were intended, the writer would truncate or assert, not use `:05d` (which pads but never caps). Both sides clearly meant "5-digit index"; the writer simply isn't bounded — an internal inconsistency, not a design that intends to explode at exactly the 100,000th event.
- **No safety net.** The stale-index rebuild in `_get_single_item` (`event_store.py:99-106`) pushes the read straight to `IndexError`, and `_count_events_on_disk`'s prefix count makes the length drift worse, not better.

## Severity

**Low** likelihood (needs ~100k events in a single conversation), but the failure mode is **silent, persistent, irreversible data loss** — no exception at load time, the history just comes back short, and subsequent appends corrupt the length accounting. A data-integrity bug in the persistence layer.

## Suggested fix

Relax the reader regex to a minimum width so it matches whatever the writer emits (simplest, fully backward-compatible — existing 5-digit files still match):
```python
EVENT_NAME_RE = re.compile(
    r"^event-(?P<idx>\d{5,})-(?P<event_id>[0-9a-fA-F\-]{8,})\.json$"
)
```
Additionally, align `_count_events_on_disk()` with the same regex so the count and scan paths can't disagree after the boundary.

## Related

- #3906 — same `EventLog`, different failure mode (append-path `listdir` O(N²)).

tomsen02:
HUMAN:

The writer pads the event index with `{idx:05d}` (a 5-digit *minimum*, uncapped) while the reader regex matched exactly `\d{5}`. So any event log that grows past 99,999 events writes 6-digit filenames the reader can no longer recognise, and the tail of the history is silently dropped on the next cold reload. This widens the reader to `\d{5,}` and adds regression tests across the boundary.

---

AGENT:

## Why

`EVENT_FILE_PATTERN = "event-{idx:05d}-{event_id}.json"` zero-pads the event index to a **minimum** of 5 digits and never caps it, so an append-only event log that grows past 99,999 events writes 6-digit filenames (`event-100000-...`). The reader `EVENT_NAME_RE` matched **exactly** 5 digits (`\d{5}`), so on a cold reload (process restart → resume/fork) every event with index ≥ 100000 was treated as "Unrecognized event file name" and excluded from the rebuilt index.

Consequences (silent, persistent, irreversible):
- `_scan_and_build_index()` stops at the first missing contiguous index, so `len(EventLog)` is truncated to 100000 — the tail of the history is silently lost on reload.
- The prefix-based `_count_events_on_disk()` still counts the 6-digit files, so it diverges from the scan; the next `append()` triggers `_sync_from_disk` and the length accounting becomes permanently inconsistent (idx 100000+ are on disk but unreadable holes).

Reported and analysed in #3926.

## Summary

- Relax the reader regex from `\d{5}` to `\d{5,}` so it accepts whatever width the `{idx:05d}` writer emits. One character; fully backward compatible (existing 5-digit names still match).
- Add regression tests: a parametrized writer↔reader agreement check across the 5/6-digit boundary, and an end-to-end cold-reload test of a >100k-event log.

## Issue Number

Closes #3926

## How to Test

Run the new + existing EventLog tests:

```
uv run pytest tests/sdk/conversation/test_event_store.py -q
# 29 passed
```

Guard verification — the new tests fail on the pre-fix regex and pass after the fix. Temporarily reverting `\d{5,}` → `\d{5}`:

```
FAILED ...::test_event_filename_writer_reader_agree[100000]
FAILED ...::test_event_filename_writer_reader_agree[100001]
FAILED ...::test_event_filename_writer_reader_agree[999999]
FAILED ...::test_event_filename_writer_reader_agree[1000000]
FAILED ...::test_event_log_cold_reload_past_100k_events
WARNING event_store.py:237 Unrecognized event file name: event-100000-...json
```

End-to-end reproduction before the fix (a 100,002-event log cold-reloads as length 100000, and idx 100000/100001 raise `IndexError: Event index out of range`); after the fix the same log reloads as 100002 with all events readable.

lint/format: `uv run ruff check` and `ruff format --check` clean on both files.

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/persistence_const.py
+++ b/openhands-sdk/openhands/sdk/conversation/persistence_const.py
@@ -3,7 +3,11 @@ import re
 
 BASE_STATE = "base_state.json"
 EVENTS_DIR = "events"
+# The index is zero-padded to a *minimum* of 5 digits by EVENT_FILE_PATTERN
+# ({idx:05d}); it is not capped, so logs past 99999 events emit 6+ digit names.
+# The reader must therefore accept 5 *or more* digits (\d{5,}) — matching exactly
+# 5 would silently drop every event with index >= 100000 on a cold reload.
 EVENT_NAME_RE = re.compile(
-    r"^event-(?P<idx>\d{5})-(?P<event_id>[0-9a-fA-F\-]{8,})\.json$"
+    r"^event-(?P<idx>\d{5,})-(?P<event_id>[0-9a-fA-F\-]{8,})\.json$"
 )
 EVENT_FILE_PATTERN = "event-{idx:05d}-{event_id}.json"
--- a/openhands-sdk/openhands/sdk/conversation/persistence_const.py
+++ b/openhands-sdk/openhands/sdk/conversation/persistence_const.py
@@ -3,10 +3,7 @@ import re
 
 BASE_STATE = "base_state.json"
 EVENTS_DIR = "events"
-# The index is zero-padded to a *minimum* of 5 digits by EVENT_FILE_PATTERN
-# ({idx:05d}); it is not capped, so logs past 99999 events emit 6+ digit names.
-# The reader must therefore accept 5 *or more* digits (\d{5,}) — matching exactly
-# 5 would silently drop every event with index >= 100000 on a cold reload.
+# Accept 5+ digits: the writer pads to a 5-digit minimum but does not cap width.
 EVENT_NAME_RE = re.compile(
     r"^event-(?P<idx>\d{5,})-(?P<event_id>[0-9a-fA-F\-]{8,})\.json$"
 )
--- a/tests/sdk/conversation/test_event_store.py
+++ b/tests/sdk/conversation/test_event_store.py
@@ -496,3 +496,65 @@ def test_event_cache_survives_across_multiple_iterations():
     first_pass = list(log)
     second_pass = list(log)
     assert all(a is b for a, b in zip(first_pass, second_pass, strict=True))
+
+
+# ---------------------------------------------------------------------------
+# Regression: event index width. The writer zero-pads to a *minimum* of 5
+# digits ({idx:05d}) and never caps, so a log past 99999 events emits 6+ digit
+# filenames. The reader regex must accept 5-or-more digits; a fixed-width
+# `\d{5}` silently dropped every event with index >= 100000 on a cold reload
+# (see issue #3926).
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("idx", [0, 1, 99999, 100000, 100001, 999999, 1_000_000])
+def test_event_filename_writer_reader_agree(idx):
+    """The reader regex must match the exact filename the writer emits."""
+    from openhands.sdk.conversation.persistence_const import (
+        EVENT_FILE_PATTERN,
+        EVENT_NAME_RE,
+    )
+
+    event_id = f"{idx:08x}-0000-0000-0000-000000000000"
+    name = EVENT_FILE_PATTERN.format(idx=idx, event_id=event_id)
+
+    m = EVENT_NAME_RE.match(name)
+    assert m is not None, f"reader regex failed to match writer output: {name!r}"
+    assert int(m.group("idx")) == idx
+
+
+def test_event_log_cold_reload_past_100k_events():
+    """A log with >100000 events reloads without dropping the tail.
+
+    Before the fix the cold scan truncated the index at 100000 (silent data
+    loss) and the count/scan paths diverged, corrupting length accounting.
+    """
+    from openhands.sdk.conversation.persistence_const import (
+        EVENT_FILE_PATTERN,
+        EVENTS_DIR,
+    )
+
+    # Raise the test-double LRU caps so the in-memory store keeps every file;
+    # this configures only the storage stand-in, not the code under test.
+    fs = InMemoryFileStore(max_size=500_000, max_memory=2 * 1024 * 1024 * 1024)
+    payload = create_test_event("seed", "seed").model_dump_json(exclude_none=True)
+
+    n = 100_002  # straddles the 5-/6-digit boundary
+    for i in range(n):
+        event_id = f"{i:08x}-0000-0000-0000-000000000000"
+        path = f"{EVENTS_DIR}/" + EVENT_FILE_PATTERN.format(idx=i, event_id=event_id)
+        fs.write(path, payload)
+
+    log = EventLog(fs)
+
+    # No silent truncation: every event survives the cold scan.
+    assert len(log) == n
+
+    # The events straddling the boundary are readable, not unreachable holes.
+    assert log[99999] is not None
+    assert log[100000] is not None
+    assert log[100001] is not None
+
+    # Appending after reload keeps length accounting consistent.
+    log.append(create_test_event("after-reload", "after"))
+    assert len(log) == n + 1
--- a/tests/sdk/conversation/test_event_store.py
+++ b/tests/sdk/conversation/test_event_store.py
@@ -6,6 +6,11 @@ from unittest.mock import Mock
 import pytest
 
 from openhands.sdk.conversation.event_store import EventLog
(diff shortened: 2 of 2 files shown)


openhands-sdk/openhands/sdk/conversation/persistence_const.py
```diff
@@ -3,7 +3,11 @@
 
 BASE_STATE = "base_state.json"
 EVENTS_DIR = "events"
+# The index is zero-padded to a *minimum* of 5 digits by EVENT_FILE_PATTERN
+# ({idx:05d}); it is not capped, so logs past 99999 events emit 6+ digit names.
+# The reader must therefore accept 5 *or more* digits (\d{5,}) — matching exactly
+# 5 would silently drop every event with index >= 100000 on a cold reload.
```
VascoSch92: Can you trim the comment? Just a minimal coincise one.

tests/sdk/conversation/test_event_store.py
```diff
@@ -496,3 +496,65 @@ def test_event_cache_survives_across_multiple_iterations():
...
+
+
+# ---------------------------------------------------------------------------
+# Regression: event index width. The writer zero-pads to a *minimum* of 5
+# digits ({idx:05d}) and never caps, so a log past 99999 events emits 6+ digit
+# filenames. The reader regex must accept 5-or-more digits; a fixed-width
+# `\d{5}` silently dropped every event with index >= 100000 on a cold reload
+# (see issue #3926).
+# ---------------------------------------------------------------------------
+
```
VascoSch92: We can delete this comment.

tests/sdk/conversation/test_event_store.py
```diff
@@ -496,3 +496,65 @@ def test_event_cache_survives_across_multiple_iterations():
...
+
+@pytest.mark.parametrize("idx", [0, 1, 99999, 100000, 100001, 999999, 1_000_000])
+def test_event_filename_writer_reader_agree(idx):
+    """The reader regex must match the exact filename the writer emits."""
+    from openhands.sdk.conversation.persistence_const import (
```
VascoSch92: import top of the file

tests/sdk/conversation/test_event_store.py
```diff
@@ -496,3 +496,65 @@ def test_event_cache_survives_across_multiple_iterations():
...
+    assert int(m.group("idx")) == idx
+
+
+def test_event_log_cold_reload_past_100k_events():
+    """A log with >100000 events reloads without dropping the tail.
+
+    Before the fix the cold scan truncated the index at 100000 (silent data
+    loss) and the count/scan paths diverged, corrupting length accounting.
+    """
+    from openhands.sdk.conversation.persistence_const import (
```
VascoSch92: import top of the file

VascoSch92 (commented):
LGTM!

If you can fix the couple of nits I pointed out, than we can merge. ;-)

tomsen02:
Thanks for the review! Addressed all four nits in 455b86d6:
- Trimmed the `persistence_const` comment to a single concise line.
- Removed the regression divider comment block.
- Hoisted both per-test imports to the top of the module.

Tests still green (29 passed), ruff check/format clean. Ready when you are 🙂

VascoSch92 (approved):
Thanks

PR #3935, opened 2026-07-01
ci: inline integration-test model config to drop private eval-repo checkout

VascoSch92:
HUMAN:

Small changes to don't use anymore the eval repo in the integration tests.

---

AGENT:

## Why

`integration-runner.yml`'s `setup-matrix` job cloned the private `OpenHands/evaluation`
repo (`eval-job/model-config`) to resolve model IDs, authenticating with
`OPENHANDS_BOT_GITHUB_PAT_PRIVATE`. That secret is not available to this public repo's
Actions, so `actions/checkout` received an empty token and failed immediately:

```
Error: Input required and not supplied: token
```

`setup-matrix` failed → `run-integration-tests` was skipped → `consolidate-results`
failed, blocking every integration/behavior run (including the v1.30.0 release checklist).

## Summary

- Remove the `Checkout evaluation model config` step that cloned the private eval repo with `OPENHANDS_BOT_GITHUB_PAT_PRIVATE`.
- Inline the default model configs (`gpt-5.5`, `deepseek-v4-flash`, `minimax-m2.7`, `gemini-3.1-pro`) directly in `setup-matrix`; `api_key`/`base_url` are still injected from env by `tests/integration/base.py`.
- Net one file, +20/-16 — no private-repo checkout, no private token, no org-secret dependency.

## Issue Number

No dedicated issue. Unblocks the v1.30.0 release (#3931) integration/behavior checks.

## How to Test

CI-only change (no runtime code), so I verified the generated matrix is equivalent to the
previous resolver and that the workflow is well-formed.

**1. The generated matrix is byte-for-byte identical to the eval-repo resolver.**
Ran the original `resolve_model_config.find_models_by_id` from `OpenHands/evaluation@main`
and the new inline dict over the 4 default IDs (Python 3.13) and diffed them:

```
$ diff orig_matrix.json inline_matrix.json
✅ IDENTICAL — inline output matches the eval-repo resolver byte-for-byte
```

Resulting matrix (unchanged from before):

```json
[
  {"id":"gpt-5.5","name":"GPT-5.5","run-suffix":"gpt_5_5_run",
   "llm-config":{"model":"litellm_proxy/openai/gpt-5.5","reasoning_effort":"high"}},
  {"id":"deepseek-v4-flash","name":"DeepSeek V4 Flash","run-suffix":"deepseek_v4_flash_run",
   "llm-config":{"model":"litellm_proxy/deepseek/deepseek-v4-flash"}},
  {"id":"minimax-m2.7","name":"MiniMax M2.7","run-suffix":"minimax_m2_7_run",
   "llm-config":{"model":"litellm_proxy/minimax/MiniMax-M2.7","temperature":1.0,"top_p":0.95}},
  {"id":"gemini-3.1-pro","name":"Gemini 3.1 Pro","run-suffix":"gemini_3_1_pro_run",
   "llm-config":{"model":"litellm_proxy/gemini-3.1-pro-preview","temperature":0.0}}
]
```

**2. Credentials still resolve.** The matrix `llm-config` only carries `model` + tuning
params; `api_key`/`base_url` come from env in `tests/integration/base.py:118-136`
(`os.getenv("LLM_API_KEY")` / `os.getenv("LLM_BASE_URL")`), so no credentials belong in the
inlined config.

**3. The workflow is well-formed.**

```
$ uvx pre-commit run yamlfmt --files .github/workflows/integration-runner.yml
Format YAML files........................................................Passed

$ python -c "import yaml; yaml.safe_load(open('.github/workflows/integration-runner.yml'))"
✅ YAML parses OK
```

Full end-to-end confirmation requires the `integration-test` / `behavior-test` label to
dispatch a real run on this base branch (which needs `LLM_API_KEY_EVAL`); that can't be
triggered locally.

## Video/Screenshots

N/A — CI/workflow-only change with no GUI. Command logs are included above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **`pull_request_target`**: this workflow reads its definition from the PR's **base**
  branch, so the fix applies to runs based on `rel-1.30.0`. The release PR #3931 (base
  `main`) won't turn green from this alone until the same change reaches `main` (on #3931
  merge).
- **Drift**: `OpenHands/evaluation` stays the source of truth; the inline table must be
  updated when a default model or proxy string changes there.
- **`workflow_dispatch` `model_ids`** now resolves only IDs present in the inline table;
  scheduled/label runs (the 4 defaults) are unaffected.

### DIFF
--- a/.github/workflows/integration-runner.yml
+++ b/.github/workflows/integration-runner.yml
@@ -70,17 +70,6 @@ jobs:
                   ref: ${{ github.event.pull_request.head.sha || github.ref }}
                   persist-credentials: false
 
-            - name: Checkout evaluation model config
-              uses: actions/checkout@v6
-              with:
-                  repository: OpenHands/evaluation
-                  ref: main
-                  token: ${{ secrets.OPENHANDS_BOT_GITHUB_PAT_PRIVATE }}
-                  path: .evaluation-model-config
-                  sparse-checkout: |
-                      eval-job/model-config
-                  persist-credentials: false
-
             - name: Set up Python
               uses: actions/setup-python@v5
               with:
@@ -109,15 +98,30 @@ jobs:
                   import json
                   import os
                   import sys
-                  sys.path.insert(0, '.evaluation-model-config/eval-job/model-config')
-                  from resolve_model_config import find_models_by_id
+                  # Inlined from OpenHands/evaluation (eval-job/model-config) so this
+                  # public workflow no longer clones the private eval repo. Only `model`
+                  # plus tuning params are needed here; api_key/base_url are injected from
+                  # env (LLM_API_KEY/LLM_BASE_URL) by tests/integration/base.py.
+                  MODELS = {
+                      "gpt-5.5": {"display_name": "GPT-5.5",
+                          "llm_config": {"model": "litellm_proxy/openai/gpt-5.5", "reasoning_effort": "high"}},
+                      "deepseek-v4-flash": {"display_name": "DeepSeek V4 Flash",
+                          "llm_config": {"model": "litellm_proxy/deepseek/deepseek-v4-flash"}},
+                      "minimax-m2.7": {"display_name": "MiniMax M2.7",
+                          "llm_config": {"model": "litellm_proxy/minimax/MiniMax-M2.7", "temperature": 1.0, "top_p": 0.95}},
+                      "gemini-3.1-pro": {"display_name": "Gemini 3.1 Pro",
+                          "llm_config": {"model": "litellm_proxy/gemini-3.1-pro-preview", "temperature": 0.0}},
+                  }
 
                   model_ids = os.environ["MODEL_IDS"].split(",")
                   model_ids = [m.strip() for m in model_ids if m.strip()]
 
-                  # find_models_by_id exits with code 1 and prints a helpful
-                  # message if any model ID cannot be resolved.
-                  resolved = find_models_by_id(model_ids)
+                  # Fail fast with a helpful message if any model ID is unknown.
+                  try:
+                      resolved = [MODELS[m] for m in model_ids]
+                  except KeyError as exc:
+                      print(f"Unknown model id {exc}; known ids: {', '.join(MODELS)}", file=sys.stderr)
+                      sys.exit(1)
 
                   matrix = []
                   for model_id, model in zip(model_ids, resolved):


enyst:
> Drift: OpenHands/evaluation stays the source of truth; the inline table must be
updated when a default model or proxy string changes there.

This is a bit of a problem, but maybe it’s OK:
- either we make a workflow to watch for changes and submit a PR here automatically (or an automation?)
- or, given they change relatively rarely, deal with it on spot (typically on releases PRs since that’s when we see them and can ask ourselves: “is this still good enough?”)

enyst:
@openhands read the PR, and all top level comments.
(1) Will this work well for all models, like GPT-5,5 comment shows for one? Will it work for behavior tests, wdyt? /codereview and post your review with gh api, you are allowed to use event approved, or any other you see fit.
(2) Find an open PR that re-adds Sonnet to the list of models, making them five, instead of four. Note that if it’s from a fork, set your remote there right. Merge main into it, it probably has conflicts right now. Then:
- if at (1) you approved, watch for this PR to be merged and then fix the other PR again accordingly
- if at (1) you commented or requested changes, then just pause.

enyst (approved):
🟢 **Good taste** — simple CI fix that removes a brittle private-repo dependency without changing the runner contract.

On the model question: this works for every model ID that is present in the inlined `MODELS` table. For this PR current default set, that means all four default scheduled/label-triggered models (`gpt-5.5`, `deepseek-v4-flash`, `minimax-m2.7`, `gemini-3.1-pro`), not just the GPT-5.5 run shown in the bot comment. The matrix code treats them uniformly: resolve ID → pass through `llm-config` → inject `LLM_API_KEY`/`LLM_BASE_URL` later in `tests/integration/base.py`.

For behavior tests: yes, this should work the same way. The `behavior-test` label only changes `TEST_TYPE_ARGS` to `--test-type behavior`; it still consumes the same matrix and `LLM_CONFIG` path as integration tests. So the inline config change is shared by both test types.

The important limitation is explicit and acceptable here: manual `workflow_dispatch model_ids` can no longer resolve arbitrary private evaluation-repo IDs. Any newly defaulted model, including the Sonnet PR mentioned in the thread, must be added to the inline table before it will work here. That is a drift/maintenance tradeoff, not a correctness blocker for this unblocker PR.

**[RISK ASSESSMENT]**
CI/workflow-only change; no runtime SDK behavior changes, no secrets added, and the private-token failure mode is removed. Main risk is model-config drift, which is already documented and bounded to this workflow table.

**VERDICT:**
✅ **Worth merging**: Approving.

**KEY INSIGHT:**
Inlining makes this public workflow self-contained; future default model additions just need to update the same table.

PR #3563, opened 2026-06-08
feat(security): supply-chain typosquat analyzer on shared shell parser

AyoubTadlaoui:
HUMAN:

 I personally tested the rebased head `2007bd7c` locally.

  - Supply-chain suite: 394 passed, 4 xfailed
  - Full security suite: 815 passed, 7 skipped, 17 xfailed
  - Changed-file pre-commit: all applicable checks passed

- [x] A human has tested these changes.

AGENT:

---

## Why

`PatternSecurityAnalyzer` and the policy-rail analyzers catch dangerous command shapes such as `rm -rf` or `curl | sh`. They do not catch a normal-looking `npm install` where the risky part is the dependency identity, for example `lodahs` instead of `lodash`.

This PR adds a deterministic, offline `SupplyChainSecurityAnalyzer` for that gap. It parses the shell command through the shared tree-sitter-bash `_shell_ast` view and raises `SecurityRisk.HIGH` for likely npm typosquat installs so `ConfirmRisky` can ask a human. It never hard-denies on its own.

## Summary

- Adds `openhands-sdk/openhands/sdk/security/supply_chain/` and exports `SupplyChainSecurityAnalyzer` from `openhands.sdk.security`.
- Detects candidates with capped Optimal-String-Alignment distance against a curated popular-package corpus.
- Flags only when the candidate is not itself popular, is exactly distance 1 from a popular package, and the popular package name is longer than 4 characters.
- Reads only `tool_call.arguments["command"]`, never tool name, thought, reasoning, or summary.
- Uses the shared `_shell_ast` command view for chained, piped, backgrounded, subshell, `$(...)`, path-qualified, wrapper, and runner installs.
- Handles static shell obfuscation for package operands and manager names: quote removal, backslash escapes, ANSI-C strings, adjacent concatenation, line continuations, and bounded comma brace expansion.
- Keeps runtime expansion, Unicode homoglyph/confusable matching, and inner command-string recursion out of scope.

## Issue Number

Implements #3560, aligned with #2721, and built on the shared shell AST from #3609.

## How to Test

Everything runs offline: no network, no model, and no new runtime dependency beyond the tree-sitter-bash view already in the SDK.

```
# new module + tests
uv run pytest tests/sdk/security/supply_chain/ -q
# => 364 passed, 3 xfailed

# full security suite
uv run pytest tests/sdk/security/ -q
# => 713 passed, 16 xfailed

# lint, types, and import layering
uv run pre-commit run --files \
[...]

PR #3944, opened 2026-07-01
feat(security): AST-backed shell command-name resolution (#2721 Phase 2b)

eeee2345:
HUMAN:
Scoped first cut for #2721 — parser-backed shell detection closing the three quoting / path / nesting bypasses agreed with @VascoSch92, kept deliberately small for review.

---

AGENT:

## Why

The HIGH shell-destructive detectors in `defense_in_depth` match flattened command text with regex, which cannot understand shell quoting, path qualification, or nesting. That leaves bypass classes no regex refinement can close: a quoted command name (`r"m" -rf /`), a path-qualified command (`/bin/rm -rf /`), and a nested command inside `-c` (`bash -c 'rm -rf /'`) all evade the `\brm\b` matcher. This implements #2721 Phase 2b for those three classes.

## Summary

- Add `shell_semantics.py`: AST-backed command-name resolution on the shared tree-sitter `_shell_ast` view — de-quotes the command name, reduces it to its POSIX basename, and recursively re-parses the `-c` operand of known shell runners (bounded depth).
- Wire it into `pattern.py` as a HIGH check after the regex HIGH pass, preserving the stable detector ID `exec.destruct.rm_rf`. On an unresolvable destructive-flag shape it emits UNKNOWN (fails safe under ConfirmRisky); benign / unparseable text stays LOW.
- Flip the quoted / path-qualified / nested-runner bypass tests to passing and add cases per class. Command-substitution and ANSI-C-quoting bypasses remain strict xfails for a follow-up. Injection and Python-code patterns stay regex — not shell syntax. No overlap with #3563.

## Issue Number

#2721

## How to Test

```
cd openhands-sdk && uv sync --group dev
uv run pytest tests/sdk/security/ -q
uv run pytest tests/sdk/security/defense_in_depth/test_shell_parser_bypasses.py -v
```

Full `tests/sdk/security/` suite: 363 passed, 10 xfailed. The bypass file shows the quoted (4) / path-qualified (5) / nested-runner (5) cases passing; command-substitution and ANSI-C classes remain xfail. `uv run pre-commit run --all-files`, whole-project `uv run pyright`, and `check_import_rules.py` (import-layering) are all green at this SHA.

Manual spot-check via `PatternSecurityAnalyzer.security_risk`:
- `r"m" -rf /`, `/bin/rm -rf /`, `/bin/"rm" -rf /`, `bash -c 'rm -rf /'`, `bash -c 'r"m" -rf /'` → all HIGH (caught).
- `evaluate()`, `subprocess.getoutput('id')`, `echo rm`, `git status`, `sh -c "$CMD"`, `tar -rf`, `grep -rf patterns.txt` → LOW (no false positive).
- `$(echo rm) -rf /`, `$FOO -rf /` → UNKNOWN (unresolvable verb); plain prose → LOW.

## Video/Screenshots

N/A — logic-only change; the pytest output above is the reproduction.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Scoped deliberately small. Follow-ups: command-substitution and ANSI-C-quoting bypasses (need a runtime-decode-vs-fail-closed policy call). cc @VascoSch92


openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py
```diff
@@ -229,6 +232,14 @@ def security_risk(self, action: ActionEvent) -> SecurityRisk:
                 logger.debug("Pattern matched: %s -> HIGH", det_id)
                 return SecurityRisk.HIGH
 
+        # HIGH: AST-backed shell resolution on executable fields only.
+        # Catches quoted, path-qualified, and nested command names that the
+        # regex word-boundary anchors cannot see (issue #2721, Phase 2b).
```
VascoSch92: don't write the commit number in the comment please :-)

openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py
```diff
@@ -241,4 +252,14 @@ def security_risk(self, action: ActionEvent) -> SecurityRisk:
                 logger.debug("Pattern matched: %s -> MEDIUM", det_id)
                 return SecurityRisk.MEDIUM
 
+        # No concrete signal fired. If the shell scan saw the destructive
```
VascoSch92: In general, search to make the comments precise but concise. Thanks

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+def _safe_parse(command: str) -> ShellProgram | None:
+    try:
+        return ast_view.parse_shell_program(command)
+    except (UnicodeEncodeError, ValueError):
+        return None
+
+
+def _program_signal(program: ShellProgram, depth: int) -> tuple[bool, bool]:
```
VascoSch92: instead of returning a tuple, could you not return an Object ? A small class which describe the output. Like

```python
class SignalOutput:
   matched: bool
   uncertain: bool
```

it is just more idiomatic

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+        elif ast_view.is_long_flag(word, _LONG_FORCE):
+            flags.add("f")
+    return flags
+
+
+def _nested_signal(
+    command: ShellCommand,
+    resolved_name: str | None,
+    depth: int,
+) -> tuple[bool, bool]:
```
VascoSch92: you could reuse the above object also here.

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+    if command.name is None:
+        return None
+    literal = _resolve_node_literal(command.name.node)
+    if literal is None:
+        return None
+    return _posix_basename(literal)
+
+
+def _resolve_word_literal(word: ShellWord) -> str | None:
```
VascoSch92: do we need this method? Can we not just use ` _resolve_node_literal(word.node)`

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+        parts.append(_decode(node.text))
+        return True
+
+    if node_type == "string":
+        # Double-quoted: recurse so an embedded expansion is caught as
+        # dynamic, while string_content fragments contribute literally.
+        for child in node.named_children:
+            if not _collect_literal_parts(child, parts):
+                return False
+        return True
```
VascoSch92: you can use a match statement here

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+def _resolve_node_literal(node: Node) -> str | None:
+    """Concatenate literal text from a name/word node, or None if dynamic.
+
+    Walks the concatenation/string/raw_string/word structure that
+    tree-sitter-bash produces for command names and arguments. Any dynamic
+    child (substitution, expansion, ANSI-C string) makes the value
+    unresolvable at analysis time and yields None.
+    """
+    parts: list[str] = []
+    if not _collect_literal_parts(node, parts):
```
VascoSch92: is it possible to be explicit here?

let pass to the method _collect_literal_parts the node and it return a list

tests/sdk/security/defense_in_depth/test_shell_parser_bypasses.py
```diff
@@ -192,3 +199,92 @@ def test_ansi_c_quoting_is_high(self, command: str):
...
+_LONG = "--" + "recursive " + "--" + "force"
+_RM = "r" + "m"
+_DEL_SHORT = _RM + " " + _RF + " /"
+_DEL_LONG = _RM + " " + _LONG + " /"
+
+
+class TestPathQualifiedCommand:
+    """Path-qualified command name breaks the ``\\brm\\s+`` anchor.
+
+    Vasco's #2721 catalog names the path-qualified delete: the directory
```
VascoSch92: you can delete my name :-)

tests/sdk/security/defense_in_depth/test_shell_parser_bypasses.py
```diff
@@ -192,3 +199,92 @@ def test_ansi_c_quoting_is_high(self, command: str):
...
+# contiguous literal only to keep local scanning tooling quiet; the strings
+# that reach the analyzer are exactly the documented bypass payloads.
+_RF = "-" + "rf"
+_LONG = "--" + "recursive " + "--" + "force"
+_RM = "r" + "m"
+_DEL_SHORT = _RM + " " + _RF + " /"
+_DEL_LONG = _RM + " " + _LONG + " /"
+
+
+class TestPathQualifiedCommand:
```
VascoSch92: you don't need a class. Just use the test method here

tests/sdk/security/defense_in_depth/test_shell_parser_bypasses.py
```diff
@@ -192,3 +199,92 @@ def test_ansi_c_quoting_is_high(self, command: str):
...
+    )
+    def test_path_qualified_is_high(self, command: str):
+        # Closed by #2721 Phase 2b: command-name resolution takes the POSIX
+        # basename after de-quoting, so path prefixes no longer hide the verb.
+        analyzer = PatternSecurityAnalyzer()
+        risk = analyzer.security_risk(make_action(command))
+        assert risk == SecurityRisk.HIGH
+
+
+class TestNestedCommandString:
```
VascoSch92: same as before

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+    if not has_flag or inner is None:
+        return False, False
+
+    inner_program = _safe_parse(inner)
+    if inner_program is None:
+        return False, False
+    return _program_signal(inner_program, depth + 1)
+
+
+def _extract_script_operand(
```
VascoSch92, line 303: This matches the script flag only when a word's text is exactly -c:

```
  if word.opaque or word.text != _SCRIPT_FLAG:   # _SCRIPT_FLAG = "-c"
      continue
```

POSIX shells allow clustering single-letter flags, so bash -xc '<script>' runs <script> just like bash -c — but the word is now -xc, so the scanner never descends. Combined with the inner-verb quoting this PR is specifically about, the regex layer is blind too:

```
  LOW   | bash -xc 'r"m" -rf /'     ← executes rm -rf /, classified LOW
  LOW   | bash -cx 'r"m" -rf /'     ← same
  HIGH  | bash -x -c 'r"m" -rf /'   ← control: separate flags → correctly caught
  LOW   | bash -c "bash -xc 'r"m" -rf /'"   ← same hole one level down
```

This is a working rm -rf / slipping through the exact "nested-runner bypass class" the summary says it closes.

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+)
+
+
+# Command basenames whose script operand is a nested shell program.
+# Resolving and re-parsing that operand closes the nested-runner bypass
+# class without hardcoding payloads.
+_SHELL_RUNNERS = frozenset({"sh", "bash", "dash", "zsh", "ksh", "ash"})
+
+# Basename of the recursive force-delete command resolved structurally.
+_RM_BASENAME = "rm"
```
VascoSch92: use from typing import Final everywhere here.

i.e.,` _RM_BASENAME: Final[str] = "rm"`

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+        "expansion",
+        "simple_expansion",
+        "arithmetic_expansion",
+        "ansi_c_string",
+    }
+)
+
+# How deep nested script operands are followed. Bounds work on adversarial
+# deeply-nested runner arguments.
+_MAX_NESTING_DEPTH = 8
```
VascoSch92: Bounding the work is right, but on hitting the bound the scanner should surface uncertain=True (→ UNKNOWN / fail-safe) rather than silently returning LOW.

VascoSch92 (changes requested):
We are going in the good direction ;-)

A couple of things to change.

VascoSch92:
@Fieldnote-Echo Could you also give a review please?

smolpaws (commented):

Taste rating: good taste

Findings:
- Changed area reviewed: openhands-sdk/openhands/sdk/security/defense_in_depth/pattern.py, openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py, tests/sdk/security/defense_in_depth/test_shell_parser_bypasses.py.

Risk assessment: low.

Verdict: worth merging

Fieldnote-Echo:
@VascoSch92 yes, I shall review this PR this afternoon.

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+    """Locate the script flag operand and resolve it to literal text.
+
+    Returns ``(has_flag, operand)``. ``has_flag`` is True when the script
+    flag is present. ``operand`` is the de-quoted literal text of the word
+    after the flag, or None when the flag has no operand or the operand is
+    dynamic (unresolvable at analysis time).
+    """
+    words = command.words
+    for index, word in enumerate(words):
+        if word.opaque or word.text != _SCRIPT_FLAG:
```
Fieldnote-Echo: This needs runner argv semantics, not an exact-word scan. Current behavior misses `bash -xc 'r"m" -rf /'` / `bash -cx ...` and over-descends on `bash -- -c ...` or `bash script -c ...`.

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+    script that no outer pattern can see. Returns ``(matched, uncertain)``.
+
+    A dynamic (unresolvable) script operand is deliberately NOT treated as
+    uncertain: passing a shell variable as the script is ordinary benign
+    usage, and flagging it would flood the caller with UNKNOWNs. Uncertainty
+    is reserved for the direct destructive-flags-on-unresolvable-verb shape
+    in ``_program_signal``.
+    """
+    if depth >= _MAX_NESTING_DEPTH:
+        return False, False
```
Fieldnote-Echo: Depth-bound hit should return uncertainty. At 9 nested `bash -c` levels, the analyzer stops scanning and returns LOW.

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+
+def _program_signal(program: ShellProgram, depth: int) -> tuple[bool, bool]:
+    """Return ``(matched, uncertain)`` for a parsed program.
+
+    ``matched`` is True when a recursive force delete is resolved. ``uncertain``
+    is True when the destructive flag shape is present on a command whose verb
+    cannot be resolved statically.
+    """
+    uncertain = False
+    for command in ast_view.iter_commands(program):
```
Fieldnote-Echo, line 185: `ShellCommand.has_error` is available but unused here. Either define parse-error spans as out of scope with coverage, or return UNKNOWN when ERROR/MISSING intersects command name, runner options, or rm flags.

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+
+    Returns True when the whole subtree is statically resolvable.
+    """
+    node_type = node.type
+
+    if node_type in _DYNAMIC_NAME_NODE_TYPES:
+        return False
+
+    if node_type == "word":
+        parts.append(_decode(node.text))
```
Fieldnote-Echo: Raw `word` text does not normalize shell escapes. `r\m -rf /` is still LOW; either handle escaped literals or mark them out of scope.

Fieldnote-Echo (changes requested):
What works well: the layering fits the existing defense-in-depth design: executable-field AST scan, stable detector ID, and UNKNOWN for unresolved destructive shapes.

What could improve: runner argv parsing and depth-bound handling need fail-safe behavior before merge; parse-error and escaped-literal scope should be explicit.

What gives me hope: this is the migration path I want to see for the shell analyzer: preserve detector identity and UNKNOWN semantics, move command resolution onto the AST incrementally, and let regression tests pin the shell cases that matter.

eeee2345:
Still working on it — sorry for the quiet stretch. For context on where this stands:

@Fieldnote-Echo's 2026-07-04 review flagged real correctness gaps (flag-clustering bypass like `bash -xc`, depth-bound silently returning LOW instead of failing safe to UNKNOWN, and escaped-literal handling). I pushed a fix on 2026-07-05 addressing those specifically — fail-safe UNKNOWN on depth-bound and parse-error, plus the POSIX flag-clustering case. @VascoSch92's earlier style/idiom comments (object instead of tuple return, Final typing, match statements) are still open on my end — happy to take another pass on those too.

@VascoSch92 @Fieldnote-Echo — could one of you take a look at the 2026-07-05 commit when you have a moment? Would rather get your read on whether the fail-safe approach actually closes the gaps you raised before touching anything else.

eeee2345:
Pushed `e2cd1c3` with the idiom pass I owed from my 7/11 comment — sorry it took the stale-bot ping.

@VascoSch92 — your review items:

- `_program_signal` / `_nested_signal` / `_scan_script_operand` return a `SignalOutput` dataclass now instead of `tuple[bool, bool]`, shared across all three as you suggested.
- `_resolve_word_literal` deleted; the single caller uses `_resolve_node_literal(word.node)` directly.
- `_collect_literal_parts` takes just the node and returns `list[str] | None` rather than appending into a passed-in list.
- The node-type dispatch is a `match` statement. `"string"` and `"command_name"`/`"concatenation"` had identical bodies, so they share one arm with the reason in a comment.
- The five single-test classes are plain test functions now, with the class docstrings moved onto the functions. The two classes holding more than one test kept their grouping. Your name is out of the docstrings.

`Final` typing and the commit-number comment were already handled in the 7/5 commit — flagging so you do not re-check them.

Behaviour is unchanged: 188 passed, 10 xfailed. `ruff format --check` is clean. The one `ruff check` finding (a StrEnum suggestion) also reproduces on `main` in a file this PR does not touch.

One thing I cannot do from here: the workflows on this branch are sitting at `action_required`. They have needed a maintainer to approve the run since the 7/5 push, so neither that commit nor this one has CI results attached. If one of you can kick that off, the checks would actually mean something.

@Fieldnote-Echo — the 7/5 fail-safe commit is still awaiting your read. That is the one that matters for correctness: fail-safe UNKNOWN on the depth bound and on parse errors, plus the POSIX flag-clustering case you raised.

openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py
```diff
@@ -0,0 +1,306 @@
...
+        return ShellScanResult(matched=False)
+
+    matched, uncertain = _program_signal(program, depth=0)
+    if matched:
+        return ShellScanResult(matched=True, detector_id=rm_detector_id)
+    return ShellScanResult(matched=False, uncertain=uncertain)
+
+
+def _safe_parse(command: str) -> ShellProgram | None:
+    try:
```
VascoSch92: use `contextlib.Suppress` here

eeee2345:
@VascoSch92 done — `_safe_parse` now uses `contextlib.suppress(UnicodeEncodeError, ValueError)` with a trailing `return None`, pushed in 104d8a6. Reads cleaner, same behaviour; 11 shell-semantics tests still pass and ruff is clean.

enyst (commented):

🔴 **Needs improvement** — the AST layer closes quoted command-name bypasses, but it still models quoted literal flags differently from the shell.

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/security/defense_in_depth/shell_semantics.py`, lines 218–226] **Security bypass: quoted literal flags are discarded.** POSIX quote removal makes both `rm "-rf" /` and `rm -r"f" /` execute with argv `rm`, `-rf`, `/`. Here, however, `_collect_flags()` delegates to helpers that reject an `opaque` (quoted/concatenated) word. The legacy regex does not rescue this because the quote remains in the scanned source. The result is another spelling of the same catastrophic delete that can avoid both rails and be classified LOW. Literal quoted fragments need to be resolved before option parsing just as literal quoted fragments are now resolved for the command name; dynamic expansions can remain unknown/fail-safe. Add analyzer-level regressions for fully quoted and concatenated short flags, not merely `split_short_flags()` unit expectations.

**[RISK ASSESSMENT]**
This code sits in the security decision path for destructive shell actions. A false negative can authorize irreversible filesystem deletion, and the bypass uses ordinary shell quoting rather than an exotic runtime expansion.

**Recommendation**: Do not auto-merge. Request review from a human security reviewer to validate that literal quote removal is applied consistently to command names and option words.

**VERDICT:**
❌ **Needs rework**: The new semantic rail still permits a straightforward equivalent spelling of the operation it is intended to catch.

**KEY INSIGHT:**
A shell analyzer cannot de-quote the executable while treating equally literal quoted argv words as opaque; command resolution and option resolution must share the same quote-removal semantics.

---

>
>
>

enyst (commented):

🟢 **Good taste**

The AST-backed detector resolves the actual shell argv semantics instead of growing another regex pile, keeps uncertainty narrow, and gives concrete matches precedence over uncertain parses. The implementation covers quoted/path-qualified verbs, nested shell runners, option parsing, malformed syntax, and bounded recursion with focused tests; the local security suite passed with 113 tests and 4 expected failures. I found no concrete correctness defect in the changed scanner.

**VERDICT:**
✅ **Worth merging**: The data model is substantially better than regex special cases, and the regression coverage matches the bypass classes being closed.

**KEY INSIGHT:**
Modeling shell commands as structured programs eliminates whole bypass classes that regex ordering can only chase one spelling at a time.

**[RISK ASSESSMENT]**
This changes a security control that can both permit destructive commands and interrupt benign commands, so impact is high even with strong focused coverage. **Recommendation**: Do not auto-merge; require a human security maintainer to validate the detector's uncertainty policy, nested-runner limits, and interaction with the analyzer ensemble.

eeee2345:
@enyst — the quoted-flag bypass is real and is fixed in `a39a199`. Thank you for
catching it; the asymmetry you named is exactly what it was.

I reproduced it first rather than fixing from the description, and it was wider
than the two spellings in the review. Every one of these came back LOW, and not
merely undetected — `uncertain` was false too, so nothing fell through to
UNKNOWN:

```
rm "-rf" /                      rm -r"f" /
rm '-rf' /                      rm -"r"f /
rm "-r" "-f" /                  rm -r\f /
rm "--recursive" "--force" /    rm --recur"sive" --force /
```

while `"rm" -rf /` matched. That is the asymmetry in one line: the verb was
de-quoted, the options were not.

**The fix.** `_collect_flags` now resolves each option word with
`_resolve_node_literal`, the same quote-removal already used for the command name
and for runner argv in `_extract_script_operand`, so command resolution and
option resolution share one semantics. `split_short_flags` and `is_long_flag`
take the resolved literal instead of a `ShellWord`; resolution is the caller's
job and an unresolvable word is skipped rather than parsed from raw source. The
end-of-options marker is compared against the de-quoted literal too, so a quoted
`--` terminates option parsing exactly as a bare one does.

Analyzer-level regressions are in `TestQuotedOptionWords`: all eight spellings
above plus the quoted-verb-and-quoted-flags combination, and the boundaries that
must stay LOW — both end-of-options forms, dynamic flag words, and either half of
the destructive pair on its own. One pre-existing unit expectation asserted that
a quoted flag group yields no flags; that had encoded the bypass as intended
behaviour, and it is replaced.

---

While verifying the fix I found three more things. None of them is closed by
`a39a199`, and I would rather put them in front of you than let them sit in my
notes. All were measured against `PatternSecurityAnalyzer`, not against
`scan_shell_command` alone, so these are end-to-end verdicts.

**1. Wrapper verbs plus quoting defeat both layers.**

| command | risk |
|---|---|
| `env rm -rf /` | HIGH |
| `env rm "-rf" /` | **LOW** |

Same for `nice`, `nohup`, `timeout`, `setsid`, `stdbuf`, `command`, `exec`,
`doas`, and for `echo / \| xargs rm "-rf"`, `find . -exec rm "-rf" {} +`, and a
heredoc feeding `rm "-rf" /` to `bash`. `sudo` is the exception and stays HIGH.

The shape is the same one you found. The unquoted form is caught only by the
flattened-text regex; quoting breaks the regex, and the AST layer never resolves
the wrapper's inner verb, so both layers miss together. The AST scanner only
fires when the resolved command *name* is `rm`, and a wrapper resolves to the
wrapper.

**2. GNU long-option abbreviation.**

`rm --recursive --force /` is HIGH; `rm --recu --fo /` is **LOW**. The long-flag
vocabulary is exact-match against two names. I checked this is a real delete
rather than a parsing curiosity, non-destructively against GNU coreutils 9.10:
`rm --recu --fo /nonexistent` exits 0 silently, `rm --recu /nonexistent` reports
`No such file or directory`, and `rm --bogusopt` is rejected as an unrecognized
option. So the abbreviations are accepted and `--fo` really is `--force`.

**3. A gap in the resolver I just touched, which I want to flag against my own
change.**

```
rm ''-rf /      HIGH
rm ""-rf /      HIGH
rm $''-rf /     LOW
```

An *empty* ANSI-C string carries no information, yet it nulls the whole word.
The same happens to a static fragment that shares a word with an ANSI-C one:
`rm -r$'f' /` and `rm $'-r'f /` are LOW, because `_collect_literal_parts` returns
`None` for a `concatenation` as soon as any child is dynamic, discarding the
perfectly resolvable `-r` alongside it.

I have deliberately not fixed this in this PR. ANSI-C in command-name position is
already an open `xfail` here with a documented policy choice between decoding the
escapes and failing closed, and flag position is the same question; picking an
answer unilaterally inside a review seems wrong. It is worth saying plainly
though that my docstring frames "dynamic words contribute no flags" as a safe
default, and for the destructive-flag case it is not safe — it silently produces
LOW rather than uncertain.

Happy to take any of the three in this PR if you want them here, or to open
follow-ups. My instinct is that 1 and 2 are separate pieces of work and 3 belongs
wherever the ANSI-C policy gets decided.

eeee2345:
@VascoSch92 @Fieldnote-Echo — I am going through the open review threads and
resolving them, and I owe you a summary rather than eighteen silent clicks.
`unresolved-review-threads` has been failing since 1 Aug purely because these
were never marked resolved, not because anything was outstanding.

Your structural and idiom points, all landed earlier in the branch:

- tuple return replaced by a `SignalOutput` dataclass, shared by
  `_program_signal`, `_nested_signal` and `_scan_script_operand`
- `_resolve_word_literal` deleted; the single caller uses
  `_resolve_node_literal(word.node)` directly
- `_collect_literal_parts` takes the node and returns `list[str] | None`
- node-type dispatch is a `match` statement
- `Final` on the module constants
- `_safe_parse` uses `contextlib.suppress`
- my name removed from the test file

Your two substantive ones:

- **the `-c` scan being exact-match.** You were right. Runner argv is now parsed
  with POSIX option semantics — short flags combine, an argument-taking option
  consumes the next word, `--` ends the option list, and the first operand ends
  it too. `TestRunnerArgvOptionParsing` covers `bash -xc`, `bash -cx`,
  `bash -o pipefail -c`, and the two shapes that must *not* descend.
- **the two comments in `pattern.py`.** Just pushed in `150f909`. The tracker
  reference is gone and both are cut to what they need to say.

@Fieldnote-Echo, your three:

- **`has_error` available but unused** — now used in both places it matters,
  `_program_signal` and `_nested_signal`, so an ERROR/MISSING node intersecting
  a span the detection relies on forces uncertainty instead of a silent LOW.
- **depth bound should surface uncertainty** — it does; hitting the bound with a
  script operand still unscanned returns uncertain.
- **raw word text does not normalize shell escapes** — handled;
  `r\m -rf /` resolves through `_unescape_unquoted` and is HIGH, and
  `"r\m" -rf /` correctly stays LOW, since inside double quotes that backslash
  is literal.

One place I did not follow you, @VascoSch92: you suggested dropping the test
class in favour of plain test methods. I kept classes, because the file already
groups by bypass family that way (`TestRunnerArgvOptionParsing`,
`TestEscapedVerb`) and the new quoted-option-word cases form another family.
Happy to flatten all of them if you would rather the file had no classes at all.

Resolving the threads now. Anything I have closed too eagerly, reopen and I will
pick it back up.

VascoSch92:
hey @eeee2345 

> you suggested dropping the test
> class in favour of plain test methods. I kept classes, because the file already
> groups by bypass family that way

I just think that methods are more clean and not so verbose as classes. But if the style of the tests are already in class format is not a problem.

Is it ready for another review?

eeee2345:
@VascoSch92 yes, ready.

Fair point on the classes — I will flatten them if you would rather the file had none, it is a mechanical change either way.

State as of `deb885b8` (main merged in): 475 passed, 7 skipped, 10 xfailed in `tests/sdk/security/`; ruff and ruff format clean; pyright clean on the four touched files. All eighteen review threads resolved. The eight quoted-flag spellings from @enyst's review are verified closed again after the merge, and the boundaries that must stay LOW — both end-of-options forms, dynamic flag words, either half of the destructive pair alone — still do.

Two things I would rather you knew before approving than discovered after.

First, the review-thread gate is the only thing left red, and it is stuck in `action_required` rather than failing: workflow runs on this PR need a maintainer to approve them. Nothing on my side unblocks that.

Second, scope. Three bypasses I found while verifying the fix are **not** closed by it, and I flagged them in a separate comment rather than quietly widening this PR:

- wrapper verbs plus quoting: `env rm -rf /` is HIGH, `env rm "-rf" /` is LOW, same for nice/nohup/timeout/setsid/stdbuf/command/exec/doas, xargs, find -exec and a heredoc. `sudo` is the exception.
- GNU long-option abbreviation: `rm --recu --fo /` is LOW, and I confirmed against real coreutils 9.10 that those abbreviations really are accepted.
- an ANSI-C gap in the resolver I touched: `rm ''-rf /` and `rm ""-rf /` are HIGH but `rm $''-rf /` is LOW — an empty ANSI-C string carries no information yet nulls the whole word.

I left all three out because ANSI-C handling is already an open policy question here (there is a strict xfail for it in command-name position) and picking an answer unilaterally inside someone else's review seemed wrong. Happy to take any of them in this PR if you would prefer them here.

VascoSch92:
LGTM

@Fieldnote-Echo What do you think? Can we merge?

VascoSch92 (approved):

PR #3953, opened 2026-07-02
fix(sdk): respect subscription validator composition

### LINKED ISSUE #3942: LLM._restore_is_subscription (mode=wrap) isn't composable with subclass before-validators, breaks extra='forbid' subclasses
## Summary

`LLM._restore_is_subscription` (`model_validator(mode="wrap")`, added in #3633) can't be safely overridden or worked around by a subclass, because it reads its own `data` argument — the pristine, unmodified raw input — rather than whatever a subclass's own `mode="before"` validator already transformed. This makes the `is_subscription` computed field effectively impossible to round-trip through any subclass that uses `extra='forbid'`, without duplicating the SDK's internal restoration logic by name.

Found while bumping the SDK pin in OpenHands/OpenHands (OpenHands/OpenHands#15079).

## Problem 1: `extra='forbid'` subclasses can't round-trip `is_subscription`

`is_subscription` is a `computed_field`, so it's never in `model_fields`. Any subclass with `model_config = ConfigDict(extra='forbid')` (a common pattern for strictly validating REST API input, e.g. to catch typos) will reject a `model_dump()` echoed straight back in, since the computed key looks like an unrecognized field:

```python
from pydantic import ConfigDict

class StrictLLM(LLM):
    model_config = ConfigDict(extra='forbid')

StrictLLM.model_validate({'model': 'openai/gpt-4o-mini', 'is_subscription': False})
# pydantic.ValidationError: is_subscription
#   Extra inputs are not permitted [type=extra_forbidden]
```

This is arguably an expected consequence of choosing `extra='forbid'` — computed fields are output-only by Pydantic design, and any consumer that wants strict validation needs to account for the base class's computed fields somehow.

## Problem 2 (the actual bug): the restore hook isn't composable with subclass validators

Naively, you'd expect a subclass `mode="before"` validator that strips `is_subscription` from the input dict to neutralize the restore behavior too, since it runs "before" in the chain. It doesn't:

```python
from pydantic import ConfigDict, model_validator

class StrictLLM(LLM):
    model_config = ConfigDict(extra='forbid')

    @model_validator(mode='before')
    @classmethod
    def _drop_is_subscription(cls, data):
        if isinstance(data, dict):
            data = {k: v for k, v in data.items() if k != 'is_subscription'}
        return data

m = StrictLLM.model_validate({'model': 'openai/gpt-4o-mini', 'is_subscription': True})
print(m.is_subscription)  # True  <-- the subclass's before-validator had no effect
```

The `is_subscription: True` from the raw input still survives onto the built model. This happens because `_restore_is_subscription` is `mode="wrap"`, and it captures its own `data` parameter — the original, untransformed input — independently of whatever `handler(data)` (which runs the subclass's `before` validators + core schema + `after` validators) does with it:

```python
@model_validator(mode="wrap")
@classmethod
def _restore_is_subscription(cls, data, handler):
    llm = handler(data)
    if isinstance(data, dict) and data.get("is_subscription"):
        llm._is_subscription = True
    return llm
```

Since `data` here is the wrap validator's own argument (not routed through the subclass's `before` validator's output), the subclass's attempt to strip the key never reaches this check. The *only* way to actually override this behavior is to redefine a method with the exact same name (`_restore_is_subscription`) in the subclass, so it shadows the parent's decorator registration entirely — which means depending on an internal, undocumented implementation name rather than any supported extension point.

## Impact

Any downstream subclass of `LLM` that:
1. Uses `extra='forbid'` (or otherwise wants to reject unknown input), and
2. Wants to prevent client input from setting `is_subscription` (since it's semantically only supposed to be set via `LLM.subscription_login()`, never via arbitrary deserialized JSON)

...cannot do so with normal Pydantic validator composition. It has to know and duplicate the private method name from `llm.py`.

## Suggested fixes (any one of these would help)

- Make the restoration logic operate on data that respects earlier-applied subclass transformations (e.g. thread it through `handler`'s result, or re-derive from the already-validated core-schema output) instead of the pristine `data` argument, so subclass `before` validators can actually neutralize it.
- Alternatively, expose the restore behavior as a small, named, documented extension point (e.g. a classmethod) subclasses can override cleanly, instead of relying on decorator-name shadowing.
- At minimum, document on `is_subscription` / `_restore_is_subscription` that: (a) it will break `extra='forbid'` subclasses on round-trip, and (b) subclass `before` validators cannot neutralize it — only re-declaring the same method name can.

## Current workaround (OpenHands/OpenHands)

`openhands/app_server/settings/llm_profiles.py`'s `StrictLLM` uses a generic `mode="before"` validator that strips any key in `cls.model_computed_fields` before validation, to avoid the 422. This doesn't attempt to block client-supplied `is_subscription` values (per problem 2 above, that would require replicating the private restore logic) — it only fixes the round-trip 422.

Sehlani042:
HUMAN:

This fixes subscription LLM serialization so strict downstream LLM subclasses can drop is_subscription during validation instead of having it restored from raw input. I reviewed the regression tests and this should keep base subscription round-trips working while making subclass validation predictable.

---

AGENT:

## Why

Fixes #3942.

`LLM._restore_is_subscription` used a `mode="wrap"` validator that read the original raw input after `handler(data)` returned. A subclass `mode="before"` validator could drop `is_subscription` before core validation, but the wrap validator still restored `_is_subscription` from the unmodified raw input. That made the computed field difficult to neutralize in strict downstream subclasses.

## Summary

- Replace the raw-input wrap restoration with a hidden `serialized_is_subscription` input field using `alias="is_subscription"`.
- Restore `_is_subscription` from the validated model state in `_post_init`, so subclass `before` validators can remove the serialized computed field before it is restored.
- Add a regression test for a strict `LLM` subclass that drops `is_subscription`, while preserving the existing base `LLM` serialization round trip.

## Issue Number

Fixes #3942

## How to Test

Regression test first failed on the old implementation:

```bash
OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/llm/test_subscription_mode.py::test_is_subscription_restore_respects_subclass_before_validator -q
# FAILED: restored.is_subscription was True
```

After the fix, I ran:

```bash
OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/llm/test_subscription_mode.py::test_is_subscription_restore_respects_subclass_before_validator tests/sdk/llm/test_subscription_mode.py::test_is_subscription_survives_serialization_round_trip -q
# 2 passed

OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/llm/test_subscription_mode.py -q
# 21 passed

OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/test_settings.py::test_llm_subscription_fields_roundtrip tests/sdk/test_settings.py::test_llm_create_agent_resolves_subscription_llm tests/sdk/test_settings.py::test_llm_from_persisted_rehydrates_subscription_runtime tests/sdk/test_settings.py::test_llm_load_from_env_rehydrates_subscription_runtime tests/sdk/test_settings.py::test_create_subscription_llm_from_config_preserves_runtime_llm tests/sdk/test_settings.py::test_async_subscription_api_key_uses_async_refresh tests/sdk/test_settings.py::test_sync_subscription_api_key_uses_valid_runtime_credentials tests/sdk/test_settings.py::test_openai_subscription_create_llm_serializes_subscription_auth tests/sdk/test_settings.py::test_create_subscription_llm_from_config_preserves_non_auth_options -q
# 9 passed

OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/conversation/test_switch_model.py::test_switch_llm_to_subscription_profile_disables_condenser -q
# 1 passed

OPENHANDS_SUPPRESS_BANNER=1 uv run pre-commit run --files openhands-sdk/openhands/sdk/llm/llm.py tests/sdk/llm/test_subscription_mode.py
# Ruff format, Ruff lint, pycodestyle, pyright, import rules, and tool registration all passed
```

I also checked that the hidden input field does not leak into `model_dump()` or `LLM.model_json_schema()`.

## Video/Screenshots

N/A; this is an SDK serialization/validation bug with unit-test coverage and command output above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

`uv` prints a repository settings warning about `exclude-newer = "7 days"` while running commands locally, but the test and pre-commit commands above completed with the pass counts shown.


enyst (commented):

**Taste Rating: 🟡 Acceptable** — the direction is right. A `mode="wrap"` validator that reaches past `handler(data)` into the raw dict is a validator that cannot be composed with, and deleting it in favor of a declared field is the correct instinct. Then the field it's replaced with grew a third copy of the same bit.

**[CRITICAL ISSUES]**

- [`openhands-sdk/openhands/sdk/llm/llm.py`, L544-550] **Data Structure — one fact, three homes, and one of them lies**: after this PR, "is this a subscription LLM?" is stored in `serialized_is_subscription` (field, L544), `_is_subscription` (private attr, L556), and `is_subscription` (property, L782). `model_post_init` copies field → private *once*, and then the field is never read or updated again. So they diverge the moment anything sets the flag at runtime, which is exactly what `subscription_login()` does:

  ```
  llm._is_subscription = True   # what subscription_login() effectively does
  is_subscription            -> True
  serialized_is_subscription -> False     # <-- public attribute, now wrong
  ```

  Serialization still round-trips correctly (`dump[is_subscription] -> True`, `round-trips? -> True`) because the dump goes through the computed property, so this is not a correctness bug today. It is a trap: `serialized_is_subscription` is a public, readable attribute whose name promises it describes the serialization state, and after any runtime transition it describes nothing. Someone will read it, believe it, and file a bug against you in six months.

  The shape with no special cases: let the field be the storage and delete `_is_subscription`, or keep the private attr and make the ingress a genuinely private one-shot (`model_config` alias into a `PrivateAttr`-backed setter). Either way the answer lives in one place.

- [same] **Three Flags To Hide One Field**: `SkipJsonSchema[...]` + `exclude=True` + `repr=False` + a name nobody would type. When a field needs four separate suppressions to stay out of sight, it's being used as a side channel rather than as state, and that's the signal to reconsider where it lives. (Credit where due: `populate_by_name` is not set, so `LLM(..., serialized_is_subscription=True)` correctly does *not* work — I checked, it silently resolves to `False` — and only the `is_subscription` alias is a real ingress. That's the right call, if slightly by accident.)

- [`tests/sdk/llm/test_subscription_mode.py`, L423-438] **Pragmatism**: the motivating test invents `StrictLLM` with `extra="forbid"` *and* a before-validator that strips `is_subscription`, purely to demonstrate the composition problem. Nobody does this. I went looking for whether subclassing `LLM` is real and it is — `RouterLLM(LLM)` (`llm/router/base.py:28`) and `TestLLM(LLM)` (`testing/test_llm.py:62`) both ship in-tree — so the *principle* holds and I'm not calling this imaginary. But neither of those subclasses would have hit the old bug. Say in the description what actually went wrong, or the next reader will assume this PR exists to satisfy its own test.

**[TESTING GAPS]**

- `uv run pytest tests/sdk/llm/test_subscription_mode.py` → **21 passed**, and the pre-existing `test_is_subscription_survives_serialization_round_trip` is retained, which is the assertion that matters most here. What's missing is a test pinning the divergence above — something that sets the flag at runtime and asserts the object doesn't report two different answers about itself.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟡 **MEDIUM** — touches `LLM` construction, which every conversation in the SDK funnels through, and changes how a flag survives the dump/validate hop to a remote agent-server. The round trip is covered by an existing test and still passes, which is what keeps this out of HIGH. Reviewers should focus on whether any caller reads `serialized_is_subscription` expecting it to be current.

**VERDICT:** ✅ **Worth merging** once the flag stops having two public spellings that can disagree — even collapsing it to a single source in this PR would be a smaller diff than what's here now.

**KEY INSIGHT:** Replacing a hack with a field was right; the mistake was keeping the thing the hack was working around, so the same bit now has a serialized copy, a runtime copy, and a property to paper over which one you meant.

enyst (commented):

🟢 **Good taste** — the current head fixes the validator-composition bug without keeping the duplicate public storage that Engel flagged on the previous revision. `is_subscription` now has one runtime source of truth, the `ContextVar` bridge is scoped and reset safely around validation, and the regression tests cover both runtime serialization and a strict subclass removing the computed input.

**[RISK ASSESSMENT]**
This changes `LLM` construction and serialization behavior used across local and remote conversations, so the blast radius is broader than the small diff suggests. The focused regressions, full green SDK suite, API-compatibility checks, and persisted-settings checks materially contain that risk.

**VERDICT:**
✅ **Worth merging**: The prior duplicate-state problem is gone, subclass validator composition is now respected, and the affected round-trip behavior is covered.

**KEY INSIGHT:**
The fix preserves one authoritative subscription flag while using validation-scoped transport state only to carry input through Pydantic’s inherited validator ordering.

enyst (approved):
Thank you!

PR #3954, opened 2026-07-02
fix(sdk): mark corrective nudge as environment event

### LINKED ISSUE #3883: Synthetic corrective nudge is emitted as user-authored MessageEvent
When the SDK handles an LLM response with no tool calls and no visible content, it injects the corrective nudge:

`Your last response did not include a function call or a message. Please use a tool to proceed with the task.`

This is currently emitted as a user-authored `MessageEvent` (`source="user"`, role `user`) with no stable metadata distinguishing it from real user input.

That creates two downstream problems:

1. UI/event consumers can display the synthetic nudge as if the human user wrote it.
2. Logic that treats `source="user"` as a real user turn can be affected, e.g. stuck/loop detection.

OpenHands docs already distinguish event attribution from LLM role and note that synthetic framework messages may use LLM role `user` while having non-user event origin.

Could the SDK mark this corrective nudge as framework-generated, for example by using `source="environment"` with LLM role `user`, or by adding a stable `sender`/metadata marker?

Related:
- #2512
- #2482
- OpenHands/OpenHands#13954

Sehlani042:
HUMAN:

This makes framework-generated corrective nudges appear as environment-origin events while preserving the user-role message sent to the model. I reviewed the follow-up scope, including visualizer/event attribution, so the PR should no longer treat synthetic recovery nudges as human input.

---

AGENT:
This PR was created by an AI agent (OpenHands) on behalf of Sehlani042.

## Why

The corrective nudge emitted after an empty or reasoning-only model response is synthetic framework feedback. It should remain a user-role LLM message so the model receives the same instruction, but the event source should not be `user`; otherwise UI/event consumers and stuck detection can treat it as human input.

## Summary

- Emit corrective nudge `MessageEvent` with `source="environment"` while keeping `llm_message.role == "user"`.
- Update regression coverage for empty and reasoning-only responses to lock the event-source / LLM-role split.
- Update the docstring to match the new origin semantics.

## Issue Number

Closes #3883

## How to Test

Baseline before the fix:

- `OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/agent/test_response_dispatch.py::test_empty_response_sends_nudge tests/sdk/agent/test_response_dispatch.py::test_reasoning_only_sends_nudge -q` failed with `assert "user" == "environment"` on both regression cases.

After the fix:

- `OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/agent/test_response_dispatch.py::test_empty_response_sends_nudge tests/sdk/agent/test_response_dispatch.py::test_reasoning_only_sends_nudge -q` passed: 2 passed.
- `OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/sdk/agent/test_response_dispatch.py tests/cross/test_stuck_detector.py -q` passed: 29 passed.
- `OPENHANDS_SUPPRESS_BANNER=1 uv run pre-commit run --files openhands-sdk/openhands/sdk/agent/response_dispatch.py tests/sdk/agent/test_response_dispatch.py` passed Ruff format, Ruff lint, pycodestyle, pyright, import dependency rules, and tool registration checks.

Note: local `uv run` still prints the existing settings-discovery warning about `exclude-newer = "7 days"`, but the commands exit successfully.

## Video/Screenshots

N/A - SDK behavior change covered by regression tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This keeps the LLM-visible message role unchanged and only changes event attribution, matching the existing stop-hook feedback pattern of `source="environment"` plus `role="user"`.


VascoSch92:
Hey @Sehlani042 

thanks for the PR. 

Can you complete the HUMAN part? Moreover, the change is not complete

neubig (approved):
LGTM, thank you!

neubig (changes requested):
Actually tests, are failing. Could you please fix them and re-request my review?

VascoSch92:
@Sehlani042 tests are failing

VascoSch92:
Thanks for the contribution

The code looks good, but there is still one related case to address before this is complete.

`DelegationVisualizer` overrides the default message-rendering path and still classifies messages using `event.llm_message.role`. Because the corrective nudge now has `source="environment"` while intentionally retaining `role="user"`, delegated conversations still render it as human input (for example, `User Message to Worker Agent`). This preserves the original false-attribution problem in delegation views.

Could you update `openhands-tools/openhands/tools/delegate/visualizer.py` to identify human messages using `event.source == "user"`, give environment/hook messages an appropriate framework title/color, and add a regression test under `tests/tools/delegate/test_visualizer.py` for an environment-source, user-role message?

neubig (approved):

neubig:
Thank you!

PR #3960, opened 2026-07-02
fix(llm): raise retryable LLMNoResponseError when stream yields no chunks

tomsen02:
HUMAN:

When a provider opens a streaming response but closes it after zero chunks (overloaded backend, truncated SSE body), `litellm.stream_chunk_builder([])` returns `None` and the transport layer dies on a bare `AssertionError`. Since `AssertionError` is not in `LLM_RETRY_EXCEPTIONS`, the call is never retried — while the non-streaming path raises the retryable `LLMNoResponseError` for the equivalent "empty choices" case. This PR makes the streaming path raise the same retryable error, so the existing retry + temperature-bump recovery applies. Sync and async transports are both fixed, symmetrically.

---

AGENT:

## Why

`_transport_call` / `_atransport_call` rebuild the final response from collected stream chunks:

```python
ret = litellm.stream_chunk_builder(chunks, messages=messages)
assert isinstance(ret, ModelResponse), f"Expected ModelResponse, got {type(ret)}"
```

If the provider yields **zero chunks** (stream opens, then closes immediately — the degenerate case behind "streaming hangs/crashes" reports), `stream_chunk_builder([])` returns `None`, and the assert raises:

```
AssertionError: Expected ModelResponse, got <class 'NoneType'>
```

Two problems with that:

1. **Not retried.** `AssertionError` is not in `LLM_RETRY_EXCEPTIONS` and is not mapped by `map_provider_exception`, so a transient provider hiccup kills the whole agent step. This is inconsistent with the non-streaming path, which raises the *retryable* `LLMNoResponseError` for empty `choices` (`_validate_chat_response`), and with the Responses streaming path (`_finalize_stream_response`), which already raises `LLMNoResponseError` when the stream finishes without a completed response.
2. **Unactionable error.** A bare `AssertionError` about types tells the user nothing about what actually happened (provider returned no content).

## Summary

- When `stream_chunk_builder` returns `None` (empty stream), raise `LLMNoResponseError` inside the retry boundary — same treatment as empty choices in the non-streaming path.
- Applied symmetrically to `_transport_call` (sync) and `_atransport_call` (async).

## Issue Number

No SDK issue filed; the failure mode is the degenerate case behind streaming-crash reports such as OpenHands/software-agent-sdk#4260.

## How to Test

```bash
uv run pytest tests/sdk/llm/test_llm_no_response_retry.py -v
```

4 new tests (sync/async × two scenarios). On `main` without this fix, `test_streaming_empty_stream_raises_llm_no_response` fails with the exact `AssertionError` above and `litellm_completion` is called only once (no retry); with the fix, the empty stream is retried and `..._retries_then_succeeds` shows a transient empty stream recovering transparently.

## Video/Screenshots

End-to-end behavior demo (mocked transport: first call returns an empty stream, second returns a normal one — simulating a transient provider hiccup):

**Before (main):**
```
💥 AssertionError: Expected ModelResponse, got <class 'NoneType'>
   litellm called 1 time — never retried
```

**After (this branch):**
```
✅ response recovered on second attempt
   litellm called 2 times — empty stream retried transparently
```

## Type

- [x] Bug fix

## Notes

The fix intentionally reuses `LLMNoResponseError` (rather than a new exception type) so the existing temperature-bump recovery in `RetryMixin` engages — some providers return empty output at `temperature=0` and succeed after the bump. Full test file (17 tests), `tests/sdk/llm/test_llm_completion.py`, and pre-commit (ruff/pycodestyle/pyright) all pass.

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -2036,6 +2036,13 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
                     on_token(chunk)
                     chunks.append(chunk)
                 ret = litellm.stream_chunk_builder(chunks, messages=messages)
+                # stream_chunk_builder returns None for an empty stream; raise
+                # LLMNoResponseError inside the retry boundary so it is retried,
+                # matching the empty-choices handling in the non-streaming path.
+                if ret is None:
+                    raise LLMNoResponseError(
+                        "Streaming completion finished without any chunks"
+                    )
 
             assert isinstance(ret, ModelResponse), (
                 f"Expected ModelResponse, got {type(ret)}"
@@ -2068,6 +2075,13 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
                     await _invoke_token_callback(on_token, chunk)
                     chunks.append(chunk)
                 ret = litellm.stream_chunk_builder(chunks, messages=messages)
+                # stream_chunk_builder returns None for an empty stream; raise
+                # LLMNoResponseError inside the retry boundary so it is retried,
+                # matching the empty-choices handling in the non-streaming path.
+                if ret is None:
+                    raise LLMNoResponseError(
+                        "Streaming completion finished without any chunks"
+                    )
 
             assert isinstance(ret, ModelResponse), (
                 f"Expected ModelResponse, got {type(ret)}"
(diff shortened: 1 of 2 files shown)


smolpaws (commented):

Taste rating: good taste

Findings:
- Changed area reviewed: openhands-sdk/openhands/sdk/llm/llm.py, tests/sdk/llm/test_llm_no_response_retry.py.

Risk assessment: low.

Verdict: worth merging

tomsen02:
Still active — this is complete and ready for review. A streaming call that yields no chunks now raises the retryable `LLMNoResponseError` instead of failing the whole run, matching the existing non-streaming behavior. CI is green and the diff-review above found no blocking issues.

tomsen02:

PR #3961, opened 2026-07-02
fix(conversation): hold state lock in update_secrets

tomsen02:
HUMAN:

`update_secrets` is the only conversation state mutator that doesn't take the state lock — it mutates `secret_registry.secret_sources` in place. When called from a user/HTTP thread while the run thread's autosave is serializing the same dict (any state field assignment triggers a full `model_dump_json`), pydantic-core panics with "dictionary changed size during iteration". The resulting `PanicException` is a `BaseException`, so it escapes the run loop's `except Exception` — the conversation never transitions to ERROR and the thread dies silently. The fix takes the state lock, matching the neighboring mutators.

---

AGENT:

## Why

Every other mutator (`send_message`, `pause`, `set_security_analyzer`, the run loop itself) acquires `with self._state:`. `update_secrets` doesn't:

```python
secret_registry = self._state.secret_registry
secret_registry.update_secrets(secrets)   # in-place dict.update, no lock
```

Both ends of the race exist in a stock agent-server deployment: the run loop executes in a dedicated executor thread and assigns state fields on every step (each assignment autosaves via `model_dump_json`, iterating `secret_sources`), while `POST .../secrets` dispatches `update_secrets` to another executor thread. A user updating secrets while a conversation runs hits it.

The failure is worse than a normal crash: pydantic-core surfaces the mutation as `pyo3_runtime.PanicException` — a `BaseException` — which bypasses `except Exception` in both `ConversationState.__setattr__` and the run loop. No ERROR status, no `ConversationErrorEvent`; the conversation just hangs forever from the client's perspective.

## Summary

- Acquire the state lock in `update_secrets` before mutating the registry, matching the neighboring mutators.

## Issue Number

None filed.

## How to Test

```bash
uv run pytest tests/sdk/conversation/local/test_update_secrets_thread_safety.py -v
```

The new test freezes an autosave mid-iteration of `secret_sources` (blocking secret serializer), then calls the real `conversation.update_secrets()` from another thread. On `main` without this fix it fails with:

```
AssertionError: autosave crashed while update_secrets ran concurrently:
[PanicException('dictionary changed size during iteration')]
```

With the fix, `update_secrets` waits for the lock and both threads complete safely; the injected secret is present afterwards.

## Video/Screenshots

Red run (fix stashed, test on main behavior):

```
E   AssertionError: autosave crashed while update_secrets ran concurrently:
E   [PanicException('dictionary changed size during iteration')]
1 failed in 0.07s
```

Green run (with fix): `1 passed`. Related suites (`test_secrets_manager.py`, `test_conversation_secrets_constructor.py`, `test_conversation_core.py`): 36 passed. pre-commit (ruff / pycodestyle / pyright): all pass.

## Type

- [x] Bug fix

## Notes

`switch_llm` has a related (milder) pattern — it calls `llm_registry.add` before taking the lock and writes `stats.usage_to_metrics` unlocked; left out of scope here to keep this PR minimal, can file separately if useful.

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -2272,8 +2272,12 @@ class LocalConversation(BaseConversation):
                      SecretValue = str | Callable[[], str]. Callables are invoked lazily
                      when a command references the secret key.
         """
-        secret_registry = self._state.secret_registry
-        secret_registry.update_secrets(secrets)
+        # Hold the state lock: mutating secret_sources while another thread's
+        # state autosave is serializing it panics pydantic-core ("dictionary
+        # changed size during iteration").
+        with self._state:
+            secret_registry = self._state.secret_registry
+            secret_registry.update_secrets(secrets)
         logger.info(f"Added {len(secrets)} secrets to conversation")
 
     def set_security_analyzer(self, analyzer: SecurityAnalyzerBase | None) -> None:
--- a/tests/sdk/conversation/local/test_update_secrets_thread_safety.py
+++ b/tests/sdk/conversation/local/test_update_secrets_thread_safety.py
@@ -1,0 +1,93 @@
+"""update_secrets must hold the state lock like every other state mutator.
+
+Without the lock, an update_secrets call from a user/HTTP thread mutates
+``state.secret_registry.secret_sources`` in place while another thread's
+state-field assignment is autosaving (``_save_base_state`` ->
+``model_dump_json``), which iterates that same dict during pydantic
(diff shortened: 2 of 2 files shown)


smolpaws (commented):

Taste rating: acceptable

Findings:
- Changed area reviewed: openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py, tests/sdk/conversation/local/test_update_secrets_thread_safety.py.

Risk assessment: medium.

tomsen02:
Still active — ready for review. Small locking fix: `update_secrets` now holds the state lock like every other state mutation, closing a race with a concurrently stepping conversation. CI is green, no open findings.

tomsen02:

PR #3968, opened 2026-07-02
fix(tools): default toolset at create_agent — SDK-owned names, profile passthrough (#3967)

### LINKED ISSUE #3967: Profile-resolved OpenHands agents have no exec tools (and no public skills)
### Summary

When a conversation is started from an **AgentProfile** (`agent_profile_id`) instead of an inline `agent_settings` payload, the resolved OpenHands agent has **no exec tools** — no `TerminalTool`, `FileEditorTool`, or `TaskTrackerTool`. It gets only the built-ins (`FinishTool`, `ThinkTool`, and `SwitchLLMTool` when enabled). Public skills are likewise not loaded by default. The agent can't run shell commands or edit files.

This is the profile-launch path exercised by agent-canvas #1571 (and the SaaS cloud backend OpenHands/OpenHands#15060). It affects **local agent-server, cloud, and eval** identically, since all three build the agent purely from the stored profile fields.

### Root cause (traced)

- `OpenHandsAgentProfile` (`openhands-sdk/openhands/sdk/profiles/agent_profile.py`) has **no `tools` field**.
- `_build_openhands_settings` (`openhands-sdk/openhands/sdk/profiles/resolver.py`) never sets `tools`, so the resolved `OpenHandsAgentSettings.tools` defaults to `[]`.
- `OpenHandsAgentSettings.create_agent()` (`openhands-sdk/openhands/sdk/settings/model.py:1116`) does `tools=self.tools` (i.e. `[]`) and only adds `include_default_tools = [FinishTool, ThinkTool]` (+ `SwitchLLMTool`).
- Exec tools (`TerminalTool`/`FileEditorTool`/`TaskTrackerTool`) live in `openhands-tools` and are attached **only** when the caller passes them in `agent_settings.tools`. On the profile path there is no such caller — agent-canvas sends `agent_profile_id` XOR `agent_settings`, so its `getAgentTools()` list is dropped.
- Similarly, skill discovery is gated on `skill_refs`; a profile with `skill_refs=[]` gets no public skills.

Agent-server resolution entry point: `openhands-agent-server/openhands/agent_server/conversation_service.py:_resolve_agent_from_profile` → `resolve_agent_profile(...)` → `settings_config.create_agent()`.

### Proposed fix

Make OpenHands profile resolution attach a default agent toolset when the profile doesn't specify one:

- In `_build_openhands_settings`, default `tools` to the standard exec set — `TerminalTool`, `FileEditorTool`, `TaskTrackerTool` — and add the sub-agent tool set when `enable_sub_agents` is set (mirroring what a normal `agent_settings` launch carries). Consider making this overridable so a profile can still narrow/replace the toolset (e.g. an optional `tools` field on `OpenHandsAgentProfile`, defaulting to the standard set).
- Default public-skill discovery on the profile path (so `skill_refs` unset ⇒ public skills load), matching non-profile launches.

Client-owned enrichments that legitimately can't be reconstructed server-side (agent-canvas's `canvas_ui` tool and the dev `RUNTIME_SERVICES` system-message-suffix) are out of scope — those stay canvas-only and only apply on the `agent_settings` path.

### Context / tracking

- Epic: #3713 (Agent Profiles)
- Consumer that surfaced this: OpenHands/agent-canvas#1571 (a scoping comment there references this issue)
- Cloud backend: OpenHands/OpenHands#15060 (delegates to the same resolver)

simonrosenberg:
HUMAN:

Found while addressing the code review on the canvas Agent Profiles PR (OpenHands/agent-canvas#1571): a conversation launched from an agent profile came up with an agent that couldn't run shell commands or edit files. Reworked after the review discussion on the hardcoded tool names: the default toolset is now SDK-owned data materialized at `create_agent()` (design: #3978), instead of string literals in the profile resolver. A human should sanity-check the chosen default toolset, the `tools=None` vs `tools=[]` semantics, and the final browser placement (deterministic SDK default; the agent-server injects browser on profile launches where chromium is usable — settled with Vasco across two review iterations).

AGENT:

## Why

A conversation started from an **AgentProfile** (`agent_profile_id`, instead of an inline `agent_settings` payload) resolved to an agent with **no `terminal` / `file_editor` / `task_tracker`** — `OpenHandsAgentProfile` has no `tools` field and `OpenHandsAgentSettings.tools` defaulted to `[]`, so `create_agent()` only attached the Finish/Think built-ins (#3967). This hit local agent-server, cloud, and eval identically.

The first iteration of this PR fixed it with hardcoded tool-name literals in the resolver. Review pushback (correct): those literals drift-race the real `.name` attributes and re-derive a set that already exists in `openhands.tools.preset.default.get_default_tools`. The deeper problem is that the default toolset existed in **four copies** — canvas `getAgentTools()` (client-side), `get_default_tools()`, the resolver literals, and the cloud caller-side override — so the same profile produced different agents on different surfaces. Full design: **#3978**.

Both the drift objection and the "don't touch callers in other repos" constraint are satisfiable at once: `Tool` is `{name, params}` and `create_agent()` only builds *specs* (implementations resolve later via the runtime registry), so the default tool **names are a wire contract** — data the SDK can own without importing `openhands-tools`.

## Summary

- **`openhands/sdk/tool/defaults.py` (new)**: `DEFAULT_EXEC_TOOL_NAMES` (`terminal`, `file_editor`, `task_tracker`), `SUB_AGENT_TOOL_NAME` (`task_tool_set`), `BROWSER_TOOL_NAME` (`browser_tool_set`), and the pure `default_tool_specs(enable_sub_agents=..., enable_browser=...)`. A new `tests/cross/test_default_tool_names.py` pins these to the real `TerminalTool.name` etc. and to `get_default_tools()` — a tool rename now breaks CI in this repo instead of silently drifting.
- **`OpenHandsAgentSettings.tools` is now `list[Tool] | None = None`**, and `create_agent()` is the **single defaulting point**: `None` → the standard exec set, plus `task_tool_set` when `enable_sub_agents` is set (that field was previously **dead server-side** — nothing consumed it); `[]` → explicitly bare (old persisted payloads keep their exact behavior); explicit list → used as given.
- **The resolver literals are gone.** `OpenHandsAgentProfile` gains the same tri-state `tools: list[Tool] | None = None`, passed through verbatim — `None` inherits the server default at `create_agent`, an explicit list enables per-profile tool selection (the `mcp_server_refs` pattern; catalog for a picker already exists via `GET /tools`).
- **`build_seed_profile` copies `agent_settings.tools`** instead of dropping it, so a global config with e.g. browser tools seeds a genuinely behavior-preserving profile.
- **Browser is injected by the serving layer, not the default** (review follow-up, superseding an adaptive-default iteration): the SDK default stays deterministic (exec set + optional `task_tool_set`; `enable_browser` is an explicit bool, off by default), and the **agent-server** appends `browser_tool_set` on a default-toolset (`tools=None`) OpenHands profile launch when the new `registry.is_tool_usable()` confirms the chromium stack — the layer that *is* the runtime makes the environment call, mirroring cloud's existing caller-side injection and canvas's client-side behavior on the settings path. An explicit `profile.tools` list (`[]` included) is authoritative and never amended; ACP profiles are never touched. Net: profile launches get browser wherever it can actually run, with zero client changes and no environment-dependent SDK defaults.

Follow-ups tracked in #3978: `StartConversationRequest.extra_tools` for client tools (`canvas_ui`), canvas dropping its client-side `DEFAULT_TOOL_NAMES`, and narrowing cloud's caller-side `get_default_tools` override to the tools-unset case (its environment injection is now the sanctioned pattern, not a wart to delete).

Closes #3967. Design: #3978. Epic: #3713.

## How to Test

- `uv run pytest tests/sdk/profiles/ tests/sdk/test_settings.py tests/sdk/tool/test_defaults.py tests/cross/test_default_tool_names.py` — profile resolution passes `tools` through as tri-state (`None` default / explicit list / `[]` bare), `create_agent()` materializes the default exec set and honors `enable_sub_agents`, the seed copies explicit toolsets, and the cross-package parity tests pin the SDK names to `openhands-tools`. (269 passed, incl. `tests/sdk/tool/test_defaults.py`)
- `tests/sdk/tool/test_defaults.py` pins the deterministic default (no registry dependence) + the `is_tool_usable` contract; `tests/agent_server/test_agent_profile_conv_start.py` covers the serving-layer injection (usable → injected; unusable → not; explicit `tools` incl. `[]` → never amended; ACP → never touched); the cross parity test runs the full `enable_browser` × `enable_sub_agents` matrix against `get_default_tools`.
- Full suites: `uv run pytest tests/sdk` (5111 passed) and `tests/agent_server/test_agent_profile_conv_start.py tests/agent_server/test_agent_profiles_router.py tests/sdk/test_apply_agent_settings_diff.py tests/cross/test_check_persisted_settings_compat.py` (104 passed).

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:91e24e3-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-91e24e3-python \
  ghcr.io/openhands/agent-server:91e24e3-python
```

```
ghcr.io/openhands/agent-server:91e24e3-golang-amd64
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-golang-amd64
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-golang-amd64
ghcr.io/openhands/agent-server:91e24e3-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:91e24e3-golang-arm64
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-golang-arm64
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-golang-arm64
ghcr.io/openhands/agent-server:91e24e3-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:91e24e3-java-amd64
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-java-amd64
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-java-amd64
ghcr.io/openhands/agent-server:91e24e3-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:91e24e3-java-arm64
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-java-arm64
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-java-arm64
ghcr.io/openhands/agent-server:91e24e3-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:91e24e3-python-amd64
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-python-amd64
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-python-amd64
ghcr.io/openhands/agent-server:91e24e3-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:91e24e3-python-arm64
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-python-arm64
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-python-arm64
ghcr.io/openhands/agent-server:91e24e3-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:91e24e3-golang
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-golang
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-golang
ghcr.io/openhands/agent-server:91e24e3-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:91e24e3-java
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-java
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-java
ghcr.io/openhands/agent-server:91e24e3-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:91e24e3-python
ghcr.io/openhands/agent-server:91e24e3bea37979752f88f675dc9a8b504d28305-python
ghcr.io/openhands/agent-server:fix-profile-resolve-default-tools-python
ghcr.io/openhands/agent-server:91e24e3-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `91e24e3-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `91e24e3-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/tool/defaults.py
+++ b/openhands-sdk/openhands/sdk/tool/defaults.py
@@ -22,8 +22,14 @@ DEFAULT_EXEC_TOOL_NAMES: tuple[str, ...] = (
 """Names of the standard exec tools every default OpenHands agent gets."""
 
 BROWSER_TOOL_NAME = "browser_tool_set"
-"""Name of the browser tool set, included in the default when the runtime
-has it registered and usable (chromium available)."""
+"""Name of the browser tool set.
+
+Not part of the deterministic default: browser is an environment-dependent
+capability, so the serving layer that knows its runtime injects it — the
+agent-server appends it on profile launches when ``is_tool_usable`` says the
+chromium stack is present, and the cloud conversation-builder does its own
+injection. Clients (canvas) add it themselves on the settings launch path.
+"""
 
 SUB_AGENT_TOOL_NAME = "task_tool_set"
 """Name of the sub-agent delegation tool set, gated on ``enable_sub_agents``."""
@@ -32,25 +38,16 @@ SUB_AGENT_TOOL_NAME = "task_tool_set"
 def default_tool_specs(
     *,
     enable_sub_agents: bool = False,
-    enable_browser: bool | None = None,
+    enable_browser: bool = False,
 ) -> list[Tool]:
     """Default tool specs for an OpenHands agent whose settings carry no tools.
 
-    ``enable_browser=None`` (the default) is adaptive: browser tools are
-    included exactly when the current runtime has them registered and usable
-    (chromium present) — the same bar canvas applies via ``GET /tools`` on the
-    settings launch path — so a default-toolset agent gets browser wherever it
-    can actually run, and resolving the spec never raises on a runtime without
-    it. Pass True/False to force.
+    Deterministic: the same inputs yield the same specs on every runtime.
+    Browser is off by default (see :data:`BROWSER_TOOL_NAME` — the serving
+    layer injects it where it can actually run); pass ``enable_browser=True``
+    to include it explicitly.
     """
     names = list(DEFAULT_EXEC_TOOL_NAMES)
-    if enable_browser is None:
-        # Local import: the registry is populated by openhands-tools at import
-        # time in the serving process; consulted lazily so this module stays
-        # import-light and cycle-free.
-        from openhands.sdk.tool.registry import is_tool_usable
-
-        enable_browser = is_tool_usable(BROWSER_TOOL_NAME)
     if enable_browser:
         names.append(BROWSER_TOOL_NAME)
     if enable_sub_agents:
--- a/openhands-sdk/openhands/sdk/tool/defaults.py
+++ b/openhands-sdk/openhands/sdk/tool/defaults.py
@@ -1,0 +1,38 @@
+"""Canonical default tool names for the standard OpenHands agent.
+
+Tool *names* are a wire contract: they are persisted in settings/profile JSON
+and sent by clients, independently of where the implementations live. Keeping
+the canonical defaults here lets ``openhands-sdk`` (which must not import
+``openhands-tools``) default a toolset from data alone — ``Tool`` is a spec
+(name + params) resolved to an implementation only at runtime via the registry.
+
+``openhands.tools.preset.default.get_default_tools`` remains the constructor
+that also registers the implementations; ``tests/cross`` asserts it stays in
+lockstep with these names.
+"""
+
+from openhands.sdk.tool.spec import Tool
+
+
+DEFAULT_EXEC_TOOL_NAMES: tuple[str, ...] = (
+    "terminal",
+    "file_editor",
+    "task_tracker",
+)
+"""Names of the standard exec tools every default OpenHands agent gets."""
+
+SUB_AGENT_TOOL_NAME = "task_tool_set"
+"""Name of the sub-agent delegation tool set, gated on ``enable_sub_agents``."""
+
+
+def default_tool_specs(*, enable_sub_agents: bool = False) -> list[Tool]:
+    """Default tool specs for an OpenHands agent whose settings carry no tools.
+
+    Browser tools are not included: they are not registered on every runtime
+    and resolving an unknown tool raises at conversation start. A caller (or a
+    profile) that wants browser opts in with an explicit tools list.
+    """
+    names = list(DEFAULT_EXEC_TOOL_NAMES)
+    if enable_sub_agents:
+        names.append(SUB_AGENT_TOOL_NAME)
+    return [Tool(name=name) for name in names]
--- a/openhands-sdk/openhands/sdk/tool/defaults.py
+++ b/openhands-sdk/openhands/sdk/tool/defaults.py
@@ -21,18 +21,38 @@ DEFAULT_EXEC_TOOL_NAMES: tuple[str, ...] = (
 )
 """Names of the standard exec tools every default OpenHands agent gets."""
 
+BROWSER_TOOL_NAME = "browser_tool_set"
+"""Name of the browser tool set, included in the default when the runtime
+has it registered and usable (chromium available)."""
+
 SUB_AGENT_TOOL_NAME = "task_tool_set"
 """Name of the sub-agent delegation tool set, gated on ``enable_sub_agents``."""
 
 
-def default_tool_specs(*, enable_sub_agents: bool = False) -> list[Tool]:
+def default_tool_specs(
+    *,
+    enable_sub_agents: bool = False,
+    enable_browser: bool | None = None,
+) -> list[Tool]:
     """Default tool specs for an OpenHands agent whose settings carry no tools.
 
-    Browser tools are not included: they are not registered on every runtime
-    and resolving an unknown tool raises at conversation start. A caller (or a
-    profile) that wants browser opts in with an explicit tools list.
+    ``enable_browser=None`` (the default) is adaptive: browser tools are
+    included exactly when the current runtime has them registered and usable
+    (chromium present) — the same bar canvas applies via ``GET /tools`` on the
+    settings launch path — so a default-toolset agent gets browser wherever it
+    can actually run, and resolving the spec never raises on a runtime without
+    it. Pass True/False to force.
     """
     names = list(DEFAULT_EXEC_TOOL_NAMES)
+    if enable_browser is None:
+        # Local import: the registry is populated by openhands-tools at import
+        # time in the serving process; consulted lazily so this module stays
+        # import-light and cycle-free.
+        from openhands.sdk.tool.registry import is_tool_usable
+
+        enable_browser = is_tool_usable(BROWSER_TOOL_NAME)
+    if enable_browser:
+        names.append(BROWSER_TOOL_NAME)
     if enable_sub_agents:
         names.append(SUB_AGENT_TOOL_NAME)
     return [Tool(name=name) for name in names]
--- a/examples/01_standalone_sdk/46_agent_settings.py
+++ b/examples/01_standalone_sdk/46_agent_settings.py
@@ -45,7 +45,7 @@ print()
 restored = OpenHandsAgentSettings.model_validate(payload)
 assert restored.condenser.enabled is True
 assert restored.condenser.max_size == 50
-assert len(restored.tools) == 2
+assert restored.tools is not None and len(restored.tools) == 2
 print("✓ Roundtrip deserialization successful — all fields preserved")
 print()
 
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -313,15 +313,12 @@ def _resolve_agent_from_profile(
         raise ValueError(f"Profile '{profile_name}' failed to resolve: {exc}") from exc
 
     agent = settings_config.create_agent()
-    # Browser is an environment-dependent capability, deliberately absent from
-    # the SDK's deterministic default (#3978): this server knows its own
-    # runtime, so it injects browser onto a default-toolset OpenHands profile
-    # when the chromium stack is actually present. An explicit profile.tools
-    # list is authoritative and never amended; the cloud conversation-builder
-    # does its own equivalent injection.
+    # Browser is deliberately absent from the deterministic SDK default
+    # (environment-dependent); this server knows its runtime, so it injects
+    # browser when usable. An explicit profile.tools list is authoritative.
     if (
-        getattr(profile, "tools", None) is None
-        and profile.agent_kind == "openhands"
+        profile.agent_kind == "openhands"
+        and profile.tools is None
         and is_tool_usable(BROWSER_TOOL_NAME)
     ):
         agent = agent.model_copy(
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -48,6 +48,7 @@ from openhands.sdk.event import MessageEvent
 from openhands.sdk.event.conversation_state import ConversationStateUpdateEvent
 from openhands.sdk.git.exceptions import GitCommandError, GitRepositoryError
 from openhands.sdk.git.utils import run_git_command, validate_git_repository
+from openhands.sdk.tool import BROWSER_TOOL_NAME, Tool, is_tool_usable
 from openhands.sdk.tool.client_tool import register_client_tools
 from openhands.sdk.utils.cipher import Cipher
 from openhands.sdk.workspace import LocalWorkspace
@@ -312,6 +313,20 @@ def _resolve_agent_from_profile(
         raise ValueError(f"Profile '{profile_name}' failed to resolve: {exc}") from exc
 
     agent = settings_config.create_agent()
+    # Browser is an environment-dependent capability, deliberately absent from
+    # the SDK's deterministic default (#3978): this server knows its own
+    # runtime, so it injects browser onto a default-toolset OpenHands profile
+    # when the chromium stack is actually present. An explicit profile.tools
+    # list is authoritative and never amended; the cloud conversation-builder
+    # does its own equivalent injection.
+    if (
+        getattr(profile, "tools", None) is None
+        and profile.agent_kind == "openhands"
+        and is_tool_usable(BROWSER_TOOL_NAME)
+    ):
+        agent = agent.model_copy(
+            update={"tools": [*agent.tools, Tool(name=BROWSER_TOOL_NAME)]}
+        )
     launched = LaunchedAgentProfile(
         agent_profile_id=profile.id,
         revision=profile.revision,
--- a/openhands-sdk/openhands/sdk/profiles/agent_profile.py
+++ b/openhands-sdk/openhands/sdk/profiles/agent_profile.py
@@ -31,6 +31,7 @@ from openhands.sdk.settings.model import (
     VerificationSettings,
 )
 from openhands.sdk.skills import Skill
+from openhands.sdk.tool import Tool
 
 
 AGENT_PROFILE_SCHEMA_VERSION = 1
@@ -159,6 +160,17 @@ class OpenHandsAgentProfile(AgentProfileBase):
     agent: str = Field(
         default="CodeActAgent",
         description="Agent class to build.",
+    )
+    # Same tri-state as the resolved settings' ``tools``: passed through
+    # verbatim by the resolver, so ``create_agent`` is the single defaulting
+    # point (#3978). Secret-free by construction (``Tool`` is name + params).
+    tools: list[Tool] | None = Field(
+        default=None,
+        description=(
+            "Tool selection for the resolved agent. None (the default) = the "
+            "server's standard tool set; [] = an explicitly bare agent; a "
+            "non-empty list is used exactly as given."
+        ),
     )
     skills: list[Skill] = Field(
         default_factory=list,
--- a/openhands-sdk/openhands/sdk/profiles/resolver.py
+++ b/openhands-sdk/openhands/sdk/profiles/resolver.py
@@ -44,10 +44,32 @@ from openhands.sdk.settings.model import (
     validate_agent_settings,
 )
 from openhands.sdk.skills import Skill, merge_skills_by_name
+from openhands.sdk.tool import Tool
 from openhands.sdk.utils.pydantic_secrets import (
     REDACTED_SECRET_VALUE,
     decrypt_str_with_cipher_or_keep,
 )
+
+
+# Default exec tools for a profile-resolved agent: a profile carries no ``tools``,
+# so without this ``create_agent`` gets ``tools=[]`` and only the Finish/Think
+# built-ins (#3967). String literals, not ``openhands-tools`` imports (this
+# package can't depend on it); browser is omitted since it isn't registered on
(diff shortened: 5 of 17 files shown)


VascoSch92 (commented):
The hardcoded tool names are avoidable — openhands.tools.preset.default.get_default_tools(enable_browser=..., enable_sub_agents=...) is already the canonical source for this exact set, uses the real .name attributes (no drift), and the caller in conversation_service.py lives in agent_server, which is allowed to import openhands.tools (and already does — see conversation_router.py). Cleaner fix would be to inject it from the caller (or set settings.tools post-resolve) rather than re-deriving a subset as string literals in the SDK resolver, which can't validate them and silently breaks if a tool is ever renamed.

The one tradeoff for keeping it here: the resolver is the single chokepoint, so hardcoding fixes local/cloud/eval in one contained SDK change without touching each caller. If we keep it, we should at least add a comment pointing at get_default_tools as the source these literals mirror.

simonrosenberg:
> The hardcoded tool names are avoidable — openhands.tools.preset.default.get_default_tools(enable_browser=..., enable_sub_agents=...) is already the canonical source for this exact set, uses the real .name attributes (no drift), and the caller in conversation_service.py lives in agent_server, which is allowed to import openhands.tools (and already does — see conversation_router.py). Cleaner fix would be to inject it from the caller (or set settings.tools post-resolve) rather than re-deriving a subset as string literals in the SDK resolver, which can't validate them and silently breaks if a tool is ever renamed.
> 
> The one tradeoff for keeping it here: the resolver is the single chokepoint, so hardcoding fixes local/cloud/eval in one contained SDK change without touching each caller. If we keep it, we should at least add a comment pointing at get_default_tools as the source these literals mirror.

you're right its a regression, Im still looking for a solution that doesnt require updating the callers in different repos

openhands-sdk/openhands/sdk/tool/defaults.py
```diff
@@ -0,0 +1,38 @@
...
+    "file_editor",
+    "task_tracker",
+)
+"""Names of the standard exec tools every default OpenHands agent gets."""
+
+SUB_AGENT_TOOL_NAME = "task_tool_set"
+"""Name of the sub-agent delegation tool set, gated on ``enable_sub_agents``."""
+
+
+def default_tool_specs(*, enable_sub_agents: bool = False) -> list[Tool]:
```
VascoSch92: still web-broswer is lost here. We do have web-broswer in agent-canvas now.

simonrosenberg: Fixed in 5b4ffdb39 — you're right that dropping browser on the profile path while canvas has it was the remaining gap. Rather than hardcoding browser in (which would make `resolve_tool` raise at conversation start on runtimes without it), the default is now **adaptive**: `default_tool_specs(enable_browser=None)` includes `browser_tool_set` exactly when the current runtime has it registered **and usable** (chromium present), via a new `registry.is_tool_usable()`. That's the same bar canvas itself applies through `GET /tools` on the settings launch path — so a profile-launched agent on the agent-server (browser registered + chromium in the image) now gets browser, matching canvas, while a bare-SDK/CLI context without chromium keeps a working default instead of a KeyError.

Exec tools stay unconditional on purpose: they're the required baseline of the standard experience, so a runtime that lacks them should fail loudly rather than silently produce a can't-edit-files agent (the #3967 bug class). Browser is an optional capability, hence include-when-available.

Cross-tests pin `BROWSER_TOOL_NAME == BrowserToolSet.name` and full parity with `get_default_tools(enable_browser=True/False)`; unit tests cover registered+usable / registered-but-no-chromium / unregistered.

simonrosenberg: Follow-up: reworked in 6ded5aecb after discussion — same outcome (browser present on profile launches wherever it can run), cleaner layering than the adaptive default above:

- The SDK default is **deterministic** again: `default_tool_specs(enable_browser: bool = False)` — no registry probing inside the SDK, same inputs → same specs everywhere.
- The **agent-server** injects `browser_tool_set` at profile launch (`_resolve_agent_from_profile`) when the profile has no explicit `tools` and `is_tool_usable` confirms chromium — i.e. the layer that *is* the runtime makes the environment call. This mirrors what cloud's conversation-builder already does and what canvas does client-side on the settings path, and it fixes a subtle flaw in the adaptive version: `create_agent()` can run far from the runtime (e.g. the cloud app-server), so probing there could test the wrong machine's chromium.
- An explicit `profile.tools` list (including `[]`) is authoritative and never amended; ACP profiles are never touched.

Net for canvas: unchanged — profile launches on the agent-server get browser, no client work needed.

simonrosenberg:
@openhands fix tests

simonrosenberg:
Fixed the CI failure in `c0d6359`.

**Root cause (unrelated to this PR's changes):** the only failing check was `agent-server-tests`, where all 6 tests in `test_agent_server_wsproto.py` errored on the session-scoped `agent_server` fixture with *"Agent server failed to start on port"*. This is a pre-existing flaky integration test — it never touches the code changed in this PR, starts fine locally (~26s under `-n auto` + coverage), but reproduced 6/6 on a CI re-run because of the full-suite parallel load. (Same diagnosis as the now-stale #1278.)

Two failure modes:
- `find_free_port()` closes the socket before the spawned server rebinds it, leaving a TOCTOU window where another process on a busy CI host can steal the port and crash startup.
- The readiness probe hit the heavy `/docs` (swagger) page with a fixed 60s budget and only caught `ConnectionError`.

**Fix (single test file, no product code):** poll the lightweight `/alive` endpoint, fail fast when the server process exits instead of waiting out the timeout, broaden the caught request errors, and retry on a fresh port to absorb the TOCTOU race.

All of this PR's own suites remain green locally (`tests/sdk/profiles/`, `tests/sdk/test_settings.py`, `tests/cross/test_default_tool_names.py`, `tests/sdk/tool/test_defaults.py`, `tests/agent_server/test_agent_profile_conv_start.py`, `tests/agent_server/test_agent_profiles_router.py` → 350 passed), and the hardened wsproto file passes under coverage + xdist.

_This comment was created by an AI agent (OpenHands) on behalf of the requester._

simonrosenberg:
Update: found the real root cause (fixed in `601020f`), which is deeper than the timeout I first hardened for.

The spawned wsproto server was **crashing on import**, not merely starting slowly. I reproduced the full `agent-server-tests` conditions locally (`pytest -n 2 --cov ... tests/agent_server`) and captured the child traceback:

```
ImportError: cannot import name 'ClientSession' from partially initialized
module 'mcp' (circular import) (.../site-packages/browser_use/mcp/__init__.py)
```

Dumping the child's `sys.path` showed `sys.path[0] == .../site-packages/browser_use`. The fixture uses the `spawn` start method from inside a pytest-xdist worker, so the child inherits the worker's `sys.path`; under the full suite another test leaves that stray `browser_use` package dir on the path. That makes `browser_use/mcp` importable as the top-level `mcp`, so `fastmcp`'s eager `import mcp` (at import time) hits the `browser_use.mcp` → `from mcp import ClientSession` circular import and the server dies before binding the port — hence "failed to start". This is env-specific (only surfaces under the full-suite spawn+xdist), which is why it passed in isolation and never reproduced from a clean interpreter.

**Fix:** sanitize the spawned child's `sys.path` by dropping entries located directly inside a `site-packages/` directory, so top-level `mcp` resolves normally. Kept the readiness hardening too (poll `/alive`, fail fast on process exit, retry on a fresh port for the `find_free_port()` TOCTOU window).

Verified locally: the 6 wsproto tests now pass under the full `-n 2 --cov tests/agent_server` run (circular-import count went 3 → 0). Note this is entirely a test-infra fix — none of this PR's product changes touch server startup or `mcp`.

_This comment was created by an AI agent (OpenHands) on behalf of the requester._

simonrosenberg:

simonrosenberg (commented):
🔥 **Roasted code review** (requested via `/codereview-roasted`)

## Taste Rating: 🟡 Acceptable, leaning 🟢

Honestly, this is mostly good taste. The v1 hardcoded-literals fix was garbage (four copies of the same toolset, guaranteed to drift), and the rework fixes the *actual* design problem instead of the symptom: one defaulting point (`create_agent`), a clean tri-state (`None` = server default / `[]` = deliberately bare / list = verbatim), and the default *names* owned by the SDK as the wire contract they already are. The cross-package parity test is the star of the show — it turns "these literals will silently drift from `.name`" into a hard CI failure. That's the right answer to the review objection, not a band-aid over it.

So no fundamental redesign needed. A few things to clean up.

- **`conversation_service.py` (~L320) — `getattr(profile, "tools", None)` fights the type system.** `profile` is a discriminated union `OpenHandsAgentProfile | ACPAgentProfile` keyed on `agent_kind`, and this repo explicitly says *avoid `getattr`/`hasattr` guards; narrow the type instead*. You're leaning on `getattr` only because `ACPAgentProfile` has no `tools`. Reorder the conjunction so the discriminant comes first and pyright narrows the union for you:

  ```python
  if (
      profile.agent_kind == "openhands"
      and profile.tools is None
      and is_tool_usable(BROWSER_TOOL_NAME)
  ):
  ```

  Identical behavior (the AND short-circuits the same way), but `profile.tools` is now a typed attribute access on the narrowed `OpenHandsAgentProfile` instead of a runtime string probe. Bonus: it reads in the order you actually reason about it ("is this an OpenHands profile with no explicit tools, and can we run a browser?").

- **Comment-to-code ratio is high, and a lot of it is PR-description narration.** The diff adds a fair amount of prose whose job is to cite issue numbers (`#3967` / `#3978`) and re-explain the design decision — e.g. the 6-line block above the browser injection, and the passthrough comments in `resolver.py` / `seed.py` / `settings/model.py`. `git blame` + the PR already own the "why we did this". The *one* comment that genuinely earns its place is "browser is deliberately absent from the deterministic default because it's environment-dependent" — that's a non-obvious invariant a future reader would otherwise try to 'fix'. Keep that one; trim the rest to a line. Non-blocking, but it's noise.

## [TESTING]

- Cross parity test (`tests/cross/test_default_tool_names.py`): 👌 pins the SDK names to the real `TerminalTool.name` etc. **and** runs the full `enable_browser × enable_sub_agents` matrix against `get_default_tools`. This is exactly the regression proof the drift objection demanded.
- SDK-side tests (`test_settings.py`, `test_resolver.py`, `test_defaults.py`) exercise the real `create_agent()` path and assert on resulting tool names — real code, not mock theater.
- The agent-server injection tests are mock-heavy (`MockResolve` / `mock_config.create_agent`), but the branch that actually matters — `model_copy(update={"tools": [*agent.tools, Tool(...)]})` — runs for real, and `is_tool_usable` is a legitimate environment probe to pin. The `usable → injected` / `unusable → skipped` / `explicit [] → never amended` / `ACP → never touched` quartet covers the real decision table. Acceptable.

## [BACKWARD COMPATIBILITY — confirm, don't fix]

Flipping the default from `[]` → `None` changes behavior for any **persisted `OpenHandsAgentSettings` that omitted `tools`**: previously builtins-only (bare), now the exec set. That's the whole point of #3967, the API-breakage bot already caught it and applied `release-note-required`, and the persisted-settings compat suite passes — so this is deliberate and tracked, not a lurking regression. Calling it out only so it's on the record: an explicit `tools: []` in old payloads still round-trips to a bare agent, which is the contract you actually need to preserve, and the test proves it.

## [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟢 **LOW** — Additive SDK data module + one serving-layer injection, guarded on an environment probe. The only behavior change is the intended fix, it's explicitly release-noted, and backward compat for the load-bearing `[]` case is tested.

## VERDICT
✅ **Worth merging** after the `getattr` reorder (and optionally trimming comments).

⚠️ Because this changes which tools a profile-launched / tools-unset agent materializes — i.e. **agent behavior** — I'm leaving this as a **COMMENT**, not an approval, per this repo's review policy: behavior-affecting changes should get a human maintainer sign-off (lightweight eval sanity-check) before merge. The PR checklist says a human tested it; a maintainer approve after that is all that's missing.

## KEY INSIGHT
The winning move was recognizing that `Tool` is `{name, params}` and `create_agent()` only builds *specs* — so the default tool **names are a wire contract the SDK can own without importing `openhands-tools`**, killing the drift race and the four-copies problem in one shot.

---

simonrosenberg:
Addressed the roasted review in 91e24e3be:

- **`getattr` reorder**: the browser-injection guard now checks `profile.agent_kind == "openhands"` first, so pyright narrows the union and `profile.tools` is a typed attribute access (verified 0 pyright errors on the file). Behavior identical.
- **Comment trim**: the PR-description narration in `resolver.py` / `seed.py` / `settings/model.py` / `conversation_service.py` is cut to the invariants the code can't show; the one comment you called load-bearing (browser deliberately absent from the deterministic default because it's environment-dependent) stays, in two lines.
- **Backward compat**: no action, as you said — the `[]`-stays-bare contract is what old payloads rely on, and `test_create_agent_empty_tools_stays_bare` pins it; the `None` flip is the intended, release-noted fix.

Remaining per your verdict: human maintainer sign-off.

VascoSch92 (approved):

simonrosenberg:
@openhands please kick off a new release of the SDK to include this PR (use github actions)

simonrosenberg:
@openhands  please kick off a new release of the SDK to include this PR (use github actions)

simonrosenberg:
🚀 Kicked off a new SDK release via GitHub Actions.

- **Prepare Release** workflow run: https://github.com/OpenHands/software-agent-sdk/actions/runs/28796856874 ✅
- **Release PR:** #4002 — `Release v1.32.0`

**Version choice:** `v1.32.0` (minor bump from the current latest `v1.31.1`). This PR is the only substantive change on `main` since the last release, and it introduces a behavioral default change (`OpenHandsAgentSettings.tools`: `list` → `None`) that CI auto-labeled `release-note-required` — a minor bump signals that behavior change while remaining API-compatible.

**Next steps:** the release PR runs the integration / behavior / example test suites. Once it's merged into `main`, `create-release.yml` will automatically tag `v1.32.0`, generate release notes (with the `release-note-required` preamble), and dispatch the PyPI publish, agent-server image build, and release-binaries workflows.

If you'd prefer a patch bump (`v1.32.0` → `v1.31.2`) instead, let me know and I can close #4002 and re-run for that version.

_This comment was created by an AI agent (OpenHands) on behalf of the requester._

PR #3970, opened 2026-07-02
fix(agent-server): honor explicit initial message run flag

### LINKED ISSUE #3940: initial_message.run is ignored on conversation start (hardcoded to True)
### Summary

When a conversation is started with an `initial_message`, the SDK ignores the client-provided `SendMessageRequest.run` flag and always runs the agent loop. The `run` argument is hardcoded to `True` at the call site.

### Location

`openhands-agent-server/openhands/agent_server/conversation_service.py` (around L800–L805 on `main`):

```python
initial_message = request.initial_message
if initial_message:
    message = Message(
        role=initial_message.role, content=initial_message.content
    )
    await event_service.send_message(message, True)  # <-- run hardcoded to True
```

`send_message` (in `event_service.py`) takes `run` as its second positional argument:

```python
async def send_message(
    self, message: Message, run: bool = False, _from_goal_loop: bool = False
):
```

But `initial_message.run` is defined and documented on the request model (`openhands-sdk/openhands/sdk/conversation/request.py`) and is silently discarded:

```python
class SendMessageRequest(BaseModel):
    ...
    run: bool = Field(
        default=False,
        description="Whether the agent loop should automatically run if not running",
    )
```

Note that only `role` and `content` are forwarded when constructing the `Message`; `run` is dropped.

### Expected behavior

The initial message should honor `initial_message.run`, e.g.:

```python
await event_service.send_message(message, initial_message.run)
```

A client that starts a conversation with `initial_message.run=False` (to seed/queue the first message without immediately kicking off the agent loop) should not have the agent run automatically.

### Actual behavior

The agent loop always runs on start, regardless of `initial_message.run`.

### Reproduction

Start a conversation via the agent-server (or app-server `POST /api/v1/app-conversations`) with:

```json
{
  "initial_message": {
    "role": "user",
    "content": [{"type": "text", "text": "Reply with exactly: NORUN_OK"}],
    "run": false
  }
}
```

The agent still produces an assistant response (verified on the on-disk event log: user `MessageEvent` is followed by an assistant `MessageEvent` with no further client action). Setting `run: true` behaves identically.

### Environment

- `openhands-sdk` / agent-server version **1.29.0**
- Also present on `main` as of filing.

### Impact / severity

Low. The dominant UI flow is "send first message → run", which matches the hardcoded `True`, so most users are unaffected. It only surprises API/SDK clients that intentionally want a no-run initial message. Filing for correctness since the field is public and documented.

### Possible consideration

If starting a conversation is intended to *always* run the initial message by design, then `SendMessageRequest.run` should arguably be ignored/omitted on the start path documentation (or the model should note that `run` has no effect for `initial_message`), to avoid the silent-discard footgun.

Sehlani042:
HUMAN:

---

AGENT:

## Why

Fixes #3940.

`ConversationService.start_conversation()` currently forwards an initial message with `run=True` unconditionally. That ignores clients that explicitly send `initial_message.run=false`, but simply switching to `initial_message.run` would regress existing first-party callers that omit `run` and currently expect startup auto-run behavior.

## Summary

- Honor explicit `initial_message.run=false` on conversation start.
- Preserve existing auto-run behavior when the `run` field is omitted.
- Add regression coverage for both explicit `run=false` and omitted `run`.

## Issue Number

Fixes #3940

## How to Test

```bash
uv run python -m pytest -q \
  tests/agent_server/test_conversation_service.py::TestConversationServiceStartConversation::test_start_conversation_honors_explicit_initial_message_run_false \
  tests/agent_server/test_conversation_service.py::TestConversationServiceStartConversation::test_start_conversation_autoruns_when_initial_message_run_omitted \
  tests/agent_server/test_conversation_service.py::TestConversationServiceStartConversation::test_start_conversation_with_secrets
```

Result: `3 passed`.

```bash
uv run ruff format --check \
  openhands-agent-server/openhands/agent_server/conversation_service.py \
  tests/agent_server/test_conversation_service.py
uv run ruff check \
  openhands-agent-server/openhands/agent_server/conversation_service.py \
  tests/agent_server/test_conversation_service.py
uv run pyright \
  openhands-agent-server/openhands/agent_server/conversation_service.py \
  tests/agent_server/test_conversation_service.py
  openhands-agent-server/openhands/agent_server/conversation_service.py \
  tests/agent_server/test_conversation_service.py
```

Result: Ruff format, Ruff lint, Pyright, pycodestyle, import dependency rules, and Tool subclass registration passed.

## Video/Screenshots

N/A, agent-server behavior covered by regression tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This keeps the PR draft because the HUMAN section still needs human-authored context before review.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -820,7 +820,14 @@ class ConversationService:
             message = Message(
                 role=initial_message.role, content=initial_message.content
             )
-            await event_service.send_message(message, True)
+            # Preserve existing startup behavior for first-party callers that
+            # omit run, while honoring clients that explicitly set run=False.
+            run_initial_message = (
+                initial_message.run
+                if "run" in initial_message.model_fields_set
+                else True
+            )
+            await event_service.send_message(message, run_initial_message)
 
         state = await event_service.get_state()
         conversation_info = _compose_conversation_info(event_service.stored, state)
--- a/tests/agent_server/test_conversation_service.py
+++ b/tests/agent_server/test_conversation_service.py
@@ -28,6 +28,7 @@ from openhands.agent_server.models import (
     ConversationInfo,
     ConversationPage,
     ConversationSortOrder,
+    SendMessageRequest,
     StartConversationRequest,
     StoredConversation,
     UpdateConversationRequest,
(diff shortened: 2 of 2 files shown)


VascoSch92 (changes requested):
This will be a breaking change in agent-canvasa and cloud. Therefore, it should be discussed first.

PR #3976, opened 2026-07-03
fix(tools/glob): don't report truncation at exactly 100 matches in ripgrep backend

tomsen02:
HUMAN:

Found that the glob tool reports results as truncated when there are
exactly 100 matches, even though nothing was actually dropped. The
ripgrep and Python backends were inconsistent on this boundary, and the false flag misleads the agent into doing extra re-searches. I went
through the root cause and the before/after repro myself.

---

AGENT:

## Why

The glob tool's ripgrep backend stops collecting at 100 files and then flags `truncated = len(file_paths) >= 100`. A result set of **exactly 100** files is therefore reported as truncated even though nothing was dropped:

- `impl.py` Python fallback uses a strict `> 100` (`_execute_with_glob`), and the grep tool does too — the ripgrep path is the only `>=`.
- `definition.py`'s field doc says *"Whether results were truncated to 100 files"* — false at exactly 100.
- The flag is not cosmetic: when set, the observation appends *"[Results truncated to first 100 files. Consider using a more specific pattern.]"*, instructing the agent to re-search for files that don't exist.

Because the backend is chosen by whether `rg` is installed, the same directory + pattern currently yields different `truncated` answers on different machines.

## Summary

- Collect all ripgrep stdout lines first (already sorted by `--sortr=modified`), set `truncated = len(lines) > 100`, return the first 100 — matching the Python fallback exactly.
- `Path.resolve()` now runs only on the returned slice instead of every line.
- Add a red→green boundary test asserting both backends return `len == 100` and `truncated is False` at exactly 100 matches.

## Issue Number

N/A (searched open/closed issues and PRs for glob truncation — no prior report).

## How to Test

```bash
uv run pytest tests/tools/glob/ -q
```

The new test fails on main (ripgrep backend reports `truncated=True` at 100 matches) and passes with this change. End-to-end repro against the real executor (ripgrep installed):

```
# main
matches=100: returned=100 truncated=True   ← false positive
# this branch
matches=99:  returned=99  truncated=False
matches=100: returned=100 truncated=False
matches=101: returned=100 truncated=True
```

## Video/Screenshots

Agent-visible observation text on main at exactly 100 matches (real `GlobExecutor` run, not mocked):

```
Found 100 file(s) matching pattern '*.py' in '/tmp/...':
...
[Results truncated to first 100 files. Consider using a more specific pattern.]
```

After this change the misleading instruction no longer appears; the 101-match case still produces it correctly.

### DIFF
--- a/openhands-tools/openhands/tools/glob/impl.py
+++ b/openhands-tools/openhands/tools/glob/impl.py
@@ -168,17 +168,14 @@ class GlobExecutor(ToolExecutor[GlobAction, GlobObservation]):
             env=sanitized_env(),
         )
 
-        # Parse output into file paths
-        file_paths = []
+        # Parse output into file paths (ripgrep already sorted by --sortr=modified)
+        lines = []
         if result.stdout:
-            for line in result.stdout.strip().split("\n"):
-                if line:
-                    file_paths.append(str(Path(line).resolve()))
-                    # Limit to first 100 files
-                    if len(file_paths) >= 100:
-                        break
+            lines = [line for line in result.stdout.strip().split("\n") if line]
 
-        truncated = len(file_paths) >= 100
+        # Limit to first 100 files; truncated only if matches were actually dropped
+        truncated = len(lines) > 100
+        file_paths = [str(Path(line).resolve()) for line in lines[:100]]
 
         return file_paths, truncated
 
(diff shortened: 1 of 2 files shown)


tomsen02:

PR #3991, opened 2026-07-04
fix(agent-server): survive webhook rate limiting and reap stale run tasks

### LINKED ISSUE #3842: agent-server: conversation wedges at execution_status=idle while /run returns 409 (stale _run_task never reaped, no self-heal)
## Summary

A conversation can end up permanently wedged: `execution_status` reports `idle`, but `POST /api/conversations/{id}/run` returns `409 conversation_already_running`, and no incoming user message ever starts a run again. The only recovery is restarting the agent-server process. Every new user message is appended but never answered.

This is the same family as #3363 (fixed by #3364), but the self-heal that #3364 added does **not** fire here, because the lingering `_run_task` never completes. In #3363 the task was in its `wait_for_pending` tail and cleared within ~30s, so the next `send_message` re-ran the stranded input. Here the task stays `not done()` indefinitely, so the rerun is never honoured.

## Environment

- `openhands-agent-server` 1.29.0 (`ghcr.io/openhands/agent-server:1.29.0-python`)
- `openhands-sdk` 1.29.0
- Self-hosted, agent-server reached over the events WebSocket through a link that drops/reconnects periodically (flaky VPN). The drops are the trigger, not the root cause.

## Symptom

1. Agent finishes a turn, conversation goes `idle` (normal).
2. User sends a follow-up. Server logs `Received message: <id>` (`sockets.py`), the `MessageEvent` is persisted, then nothing. No LLM call, no `ActionEvent`.
3. Sending further messages does not help. They all persist and none are processed.
4. `GET /api/conversations?ids=<id>` keeps reporting `execution_status: idle`.
5. `POST .../run` returns `409`. `POST .../pause` then `POST .../run` still returns `409` (status flips to `paused`, run still refused).
6. Recovery only via restarting the agent-server process. State reloads from the persisted event store and the queued message can finally be run.

So the UI/health shows a ready (`idle`) conversation that is actually dead.

## Root cause

`execution_status` and `_run_task` liveness can diverge.

In `EventService.run()` (`openhands-agent-server/openhands/agent_server/event_service.py`):

```python
async with self._run_lock:
    if await self._get_execution_status() == ConversationExecutionStatus.RUNNING:
    ...
    if self._run_task is not None and not self._run_task.done():
```

The `409` we see is the **second** guard: status is `idle` (so the first guard passes) but `_run_task` is still alive. An incoming user message resets the terminal status to `IDLE` via `conversation.send_message(...)` independently of the run loop, so once the underlying `_run_task` hangs (never returns from `arun()`/`run()`, or its cleanup never completes), you get `status == IDLE` with `_run_task` not done.

The rerun safety net added in #3364 is only consumed inside `_run_and_publish`'s `finally`:

```python
finally:
    ...
    self._run_task = None
    ...
    if rerun_requested:
        ...
        if should_restart:
            await self.run(...)
```

If `_run_task` never reaches its `finally`, `_rerun_requested` is set on every subsequent `send_message(run=True)` but never acted on. Deadlock. There is no reaper that reconciles "status is IDLE/PAUSED but a stale `_run_task` is still alive" or that ever cancels the stale task.

## Why the WebSocket churn matters

The trigger here is a connection that drops and reconnects around the end of a run (or around stuck-detection firing). A robust server should reconnect and carry on. Instead, one badly-timed drop leaves the run task orphaned and the conversation permanently unrunnable. The networking is flaky; the bug is that flaky networking can permanently brick a conversation with no self-recovery.

## Suggested direction

Any one of these would break the deadlock:

1. In `run()`, if the status guard passes (`!= RUNNING`) but `_run_task is not None and not _run_task.done()`, treat it as an inconsistent state: cancel/replace the stale task instead of raising `409`. A run task that is not reflected in the execution status should not be allowed to block a new run forever.
2. Add a bounded timeout / cancellation around `_run_and_publish` (and its `wait_for_pending` tail) so a task can never stay `not done()` indefinitely while the status has moved on.
3. Add a reconciler/reaper: if `execution_status` is `IDLE`/`PAUSED` with a trailing unprocessed user message and no live step in progress, start a run (or surface a non-`idle` status so health checks can detect the wedge).

Happy to put up a PR for option 1 if that is the preferred approach.

## Repro notes

Hard to reproduce deterministically since it needs a drop at the wrong moment, but it is reliably reachable on a long-running conversation over a flaky link. The diagnostic that uniquely identifies this state: `execution_status == "idle"` while `POST .../run` returns `409`.

shanemort1982:
HUMAN:

Follow-up to my report in #3842, which VascoSch92 reproduced. We caught the trigger in the wild on our fleet: the app server rate-limits its own sandbox's webhook posts (429), the stock poster gives up, the app-side transcript freezes, and the lingering run task answers 409 forever with no self-heal. Fix validated end to end on a self-hosted box (evidence below) on top of the unit suites.

---

AGENT:

## Why

Fixes #3842.

Two defects compound into the wedge:

1. `WebhookSubscriber._post_events` retries on a fixed delay with no backoff, does not honour `Retry-After`, fires overlapping volleys (each buffer-full flush posts concurrently), and after exhausting retries only tries again when a NEW event arrives. Against a rate-limited downstream (the OpenHands app server applies `InMemoryRateLimiter(requests=10, seconds=1)` per client IP to ALL routes, including `/api/v1/webhooks/events/...`) the concurrent volleys trip the limiter, keep it tripped, and delivery stalls; the queue trim then drops events permanently. Production signature: hundreds of `Webhook post attempt 4 failed: 429` / `Failed to post events ... after 4 attempts` per second for 90+ minutes while the app-side transcript sat frozen (the agent's local store reached 8,622 events; the app side had 270).

2. `EventService.run()` raises `conversation_already_running` whenever `_run_task` is alive, even while `execution_status` reports non-RUNNING. A run task that never completes therefore wedges the conversation forever - the self-heal added in #3364 never fires because the task never reaches its cleanup tail.

## Summary

- Webhook poster: serialise flushes behind a lock; exponential backoff between attempts, honouring `Retry-After`; after a failed volley enter an exponentially growing cooldown (1s floor, 60s cap) instead of hammering; re-arm the flush timer on failure so delivery resumes WITHOUT requiring a new event; preserve delivery order by prepending re-queued events.
- `EventService.run()`: when a live `_run_task` coexists with non-RUNNING `execution_status`, start a staleness clock; if the SAME task is still lingering on a later attempt past `STALE_RUN_TASK_REAP_SECONDS` (120s, comfortably beyond the 30s-bounded `wait_for_pending` tail), cancel and reap it, then start the requested run. The task's `finally` now only clears its own task reference so a reaped task cannot clobber a replacement run.

## How to Test

Commands run in a clean checkout with the patch applied:

- `uv run pytest tests/agent_server/test_webhook_subscriber.py tests/agent_server/test_event_service.py` -> **150 passed**. 5 new tests cover Retry-After/backoff with zero event loss and order preservation, cooldown queueing without posts, and the stale-run-task reaper (first observation never reaps; same lingering task past the grace period is cancelled and cleared; a different task restarts the clock). 5 existing assertions updated where the fixed-delay retry cadence intentionally changed to exponential.
- `uv run ruff check` / `uv run ruff format` on the four touched files: clean.

End to end on a self-hosted deployment (agent-server image built from this branch's changes applied to v1.30.0, `--target source`; OpenHands app at current main). The app's global rate limit was tightened to `requests=1/second` to force the storm, then a long multi-step streaming conversation was driven through the full app -> sandbox -> webhook path:

- **Stock agent-server (1.30.0)**: webhook volleys tripped repeated 429s and logged the permanent give-up (`Failed to post events to webhook ... after 4 attempts`); the app-side transcript stalls until unrelated later events happen to re-trigger a flush - or forever when the storm persists (the production case above).
- **Patched agent-server**: the serialised, paced poster stayed under the SAME limiter - zero 429s - and the full transcript including the final assistant message reached the app store; the conversation completed normally. Under a separately induced sustained-failure scenario, the cooldown machinery held: re-arms progressed 10s -> 20s -> 40s -> 60s (capped), no hot loop, no drops within the queue bound.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -2,6 +2,7 @@ import asyncio
 import importlib
 import json
 import logging
+import time
 from collections.abc import Awaitable, Callable
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import suppress
@@ -1598,6 +1599,11 @@ class AutoTitleSubscriber(Subscriber):
             return None
 
 
+# Cap for the exponential retry/cooldown backoff so a long outage retries at a
+# steady cadence instead of growing unboundedly.
+_WEBHOOK_MAX_BACKOFF_SECONDS = 60.0
+
+
 @dataclass
 class WebhookSubscriber(Subscriber):
     conversation_id: UUID
@@ -1606,6 +1612,15 @@ class WebhookSubscriber(Subscriber):
     session_api_key: str | None = None
     queue: list[Event] = field(default_factory=list)
     _flush_timer: asyncio.Task | None = field(default=None, init=False)
+    # Serializes flushes: without it a burst of events (e.g. streaming deltas)
+    # fires overlapping _post_events volleys that hammer a rate-limited
+    # downstream and deliver out of order.
+    _post_lock: asyncio.Lock = field(default_factory=asyncio.Lock, init=False)
+    # Cooldown circuit: monotonic deadline before which no post is attempted.
+    # Set after a flush exhausts its retries; grows exponentially with
+    # consecutive failures so a downstream returning 429 gets room to recover.
+    _backoff_until: float = field(default=0.0, init=False)
+    _consecutive_failures: int = field(default=0, init=False)
     # Per-instance sleep seam so tests override delays without patching the
     # global asyncio.sleep. default_factory (not default) keeps it an instance
     # attribute, else the function would be descriptor-bound as a method.
@@ -1617,7 +1632,20 @@ class WebhookSubscriber(Subscriber):
         """Add event to queue and post to webhook when buffer size is reached."""
         self.queue.append(event)
 
-        if len(self.queue) >= self.spec.event_buffer_size:
+        # Enforce the queue bound on enqueue as well as on failed-flush
+        # re-queue: during a failure cooldown events accumulate here without
+        # passing through _post_events, and the bound must hold throughout.
+        overflow = len(self.queue) - self.spec.max_queue_size
+        if overflow > 0:
+            del self.queue[:overflow]
+            logger.warning(
+                f"Webhook queue exceeded max_queue_size="
+                f"{self.spec.max_queue_size}; dropped {overflow} oldest "
+                f"event(s) for conversation {self.conversation_id}."
+            )
+
+        in_cooldown = time.monotonic() < self._backoff_until
+        if len(self.queue) >= self.spec.event_buffer_size and not in_cooldown:
             # Cancel timer since we're flushing due to buffer size
             self._cancel_flush_timer()
             await self._post_events()
@@ -1630,11 +1658,47 @@ class WebhookSubscriber(Subscriber):
         self._cancel_flush_timer()
 
         if self.queue:
+            # Best effort on shutdown: skip any active cooldown so the tail of
+            # the conversation gets one final delivery attempt.
+            self._backoff_until = 0.0
             await self._post_events()
+
+    def _retry_wait(self, error: Exception, attempt: int) -> float:
+        """Exponential backoff wait, honouring a Retry-After header when present."""
+        wait = min(self.spec.retry_delay * (2**attempt), _WEBHOOK_MAX_BACKOFF_SECONDS)
+        if isinstance(error, httpx.HTTPStatusError):
+            retry_after = error.response.headers.get("Retry-After")
+            if retry_after:
+                # TypeError guards non-string values (e.g. mocked responses).
+                with suppress(TypeError, ValueError):
+                    wait = max(wait, float(retry_after))
+        return wait
+
+    def _schedule_retry_flush(self):
+        """Arm a flush timer so delivery resumes after the cooldown expires.
+
+        Without this, events re-queued after a failed flush would only be
+        retried when a NEW event arrived - the tail of a conversation could
+        stay undelivered forever after a downstream outage.
+        """
+        if self._flush_timer:
+            return
+        delay = max(self._backoff_until - time.monotonic(), 0.0)
+        self._flush_timer = asyncio.create_task(self._flush_after_delay(delay))
 
     async def _post_events(self):
         """Post queued events to the webhook with retry logic."""
+        async with self._post_lock:
+            await self._post_events_locked()
+
+    async def _post_events_locked(self):
         if not self.queue:
+            return
+        if time.monotonic() < self._backoff_until:
+            # Still cooling down after exhausted retries (e.g. the downstream
+            # is rate limiting with 429s). Leave events queued; the retry
+            # flush timer will deliver them once the cooldown expires.
+            self._schedule_retry_flush()
             return
 
         events_to_post = self.queue.copy()
@@ -1662,7 +1726,9 @@ class WebhookSubscriber(Subscriber):
             f"{self.spec.base_url.rstrip('/')}/events/{self.conversation_id.hex}"
         )
 
-        # Retry logic
+        # Retry logic: exponential backoff (honouring Retry-After) between
+        # attempts; after the final failure, enter a cooldown and re-arm the
+        # flush timer instead of hammering a struggling downstream.
         for attempt in range(self.spec.num_retries + 1):
             try:
                 async with httpx.AsyncClient() as client:
@@ -1678,17 +1744,32 @@ class WebhookSubscriber(Subscriber):
                         f"Successfully posted {len(event_data)} events "
                         f"to webhook {events_url}"
                     )
+                    self._consecutive_failures = 0
+                    self._backoff_until = 0.0
                     return
             except Exception as e:
                 logger.warning(f"Webhook post attempt {attempt + 1} failed: {e}")
                 if attempt < self.spec.num_retries:
-                    await self._sleep(self.spec.retry_delay)
+                    await self._sleep(self._retry_wait(e, attempt))
                 else:
+                    self._consecutive_failures += 1
+                    # Floor of 1s so retry_delay=0 cannot produce a hot loop.
+                    cooldown = max(
+                        min(
+                            self.spec.retry_delay * (2**self._consecutive_failures),
+                            _WEBHOOK_MAX_BACKOFF_SECONDS,
+                        ),
+                        1.0,
+                    )
+                    self._backoff_until = time.monotonic() + cooldown
                     logger.error(
                         f"Failed to post events to webhook {events_url} "
-                        f"after {self.spec.num_retries + 1} attempts"
+                        f"after {self.spec.num_retries + 1} attempts; "
+                        f"retrying in {cooldown:.0f}s"
                     )
-                    self.queue.extend(events_to_post)
+                    # Prepend so delivery order is preserved when events
+                    # queued during this flush are eventually posted.
+                    self.queue = events_to_post + self.queue
                     overflow = len(self.queue) - self.spec.max_queue_size
                     if overflow > 0:
                         del self.queue[:overflow]
@@ -1697,6 +1778,7 @@ class WebhookSubscriber(Subscriber):
                             f"{self.spec.max_queue_size}; dropped {overflow} "
                             f"oldest event(s) for {events_url}."
                         )
+                    self._schedule_retry_flush()
 
     def _cancel_flush_timer(self):
         """Cancel the current flush timer if it exists."""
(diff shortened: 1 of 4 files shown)


shanemort1982:
@VascoSch92 rebased onto current main and reconciled with the `webhook_subscriber` refactor. The PR was showing `mergeable: dirty` purely because two files (`conversation_service.py`, `test_webhook_subscriber.py`) had been committed with CRLF line endings, so every line read as changed and the three-way merge blew up. Normalised to LF and re-applied; the real change is back to +369/-22 across 4 files and now merges clean.

Green locally against main: `test_event_service.py` + `test_webhook_subscriber.py` = 152 passed, ruff clean. CI on the new head is passing so far.

Since you already reproduced #3842, could you give this a review? It's biting a self-hosted fleet daily: a run task that never reaches its cleanup tail leaves the conversation permanently answering 409 with no self-heal, and the webhook poster stalls forever against the app server's per-IP rate limit on `/api/v1/webhooks/events/*`. The reaper half is the important one for the wedge.

shanemort1982:
Field update: this has been live in production across our self-hosted fleet for several days now (6 hosts on SDK 1.36.0 + this agent-server build) and the #3842 wedge has not reoccurred.

We're now removing the watchdog shim we'd built to auto-POST `/run` on `idle`+`RUNNING` conversations as a stopgap for #3842, because the reaper makes it redundant. That removal is the cleanest confirmation the fix is doing its job in practice: every conversation the shim was "recovering" turned out to be a fresh conversation waiting for its first user message, not a stale run task.

The webhook backoff + cooldown circuit has also held up under real sustained-429 load that previously orphaned run tasks.

Still showing `mergeable: dirty`. Happy to rebase onto current main and normalise line-endings on `conversation_service.py` / `test_webhook_subscriber.py` if that helps move it forward. @enyst, a review would be great when you have a moment.

enyst (commented):

🔴 **Needs improvement** - the webhook backoff work is sensible, but the stale-run recovery can permit two runs to mutate one conversation concurrently.

**[CRITICAL ISSUES]**
- [`openhands-agent-server/openhands/agent_server/event_service.py`, Line 953] **Concurrency / Data Ownership**: `_maybe_reap_stale_run_task()` cancels the asyncio wrapper, drops `_run_task`, and immediately lets a replacement run start. For the synchronous fallback at line 1033, cancellation does not stop `conversation.run()` already executing in `ThreadPoolExecutor`; it only stops awaiting that worker. The old worker can therefore continue mutating conversation state and persistence while the replacement run starts against the same `Conversation`. Do not make the conversation runnable again until the underlying worker has actually exited, or discard/recreate the old conversation instance before starting a replacement.

**[TESTING GAPS]**
- [`tests/agent_server/test_event_service.py`, Line 3424] **The Test Models the Wrong Thing**: the reaper test cancels a plain `asyncio.Event().wait()` task, which is cooperatively cancellable and does not exercise the `run_in_executor` path. Add a blocking synchronous `conversation.run()` regression test and prove a second run cannot start until that worker exits.

**KEY INSIGHT:**
An asyncio task reference is not ownership of the executor worker it awaits; clearing one does not terminate the other.

**[RISK ASSESSMENT]**
This changes core conversation execution and recovery behavior, with a realistic overlapping-run path that can corrupt state or duplicate actions. **Recommendation**: Do not auto-merge. Request review from a human concurrency/agent-runtime maintainer to validate worker lifetime and replacement semantics.

**VERDICT:**
❌ **Needs rework**: stale-run recovery must not overlap an uncancelled synchronous run with its replacement.

---

>
>
>

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -1598,6 +1599,11 @@ def _load_title_llm(self) -> LLM | None:
             return None
 
 
+# Cap for the exponential retry/cooldown backoff so a long outage retries at a
+# steady cadence instead of growing unboundedly.
+_WEBHOOK_MAX_BACKOFF_SECONDS = 60.0
```
VascoSch92, line 1604: use Final[float] from typing

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -1630,12 +1658,48 @@ async def close(self):
...
 
+    def _retry_wait(self, error: Exception, attempt: int) -> float:
+        """Exponential backoff wait, honouring a Retry-After header when present."""
+        wait = min(self.spec.retry_delay * (2**attempt), _WEBHOOK_MAX_BACKOFF_SECONDS)
+        if isinstance(error, httpx.HTTPStatusError):
+            retry_after = error.response.headers.get("Retry-After")
+            if retry_after:
+                # TypeError guards non-string values (e.g. mocked responses).
+                with suppress(TypeError, ValueError):
+                    wait = max(wait, float(retry_after))
```
VascoSch92, line 1674: `_retry_wait` does `wait = min(..., _WEBHOOK_MAX_BACKOFF_SECONDS)` then `wait = max(wait, float(retry_after))` the header value overrides the 60s cap with no ceiling. 

That sleep runs at line 1753 inside `_post_events_locked`, which holds `_post_lock`. 

A numeric Retry-After: 3600 stalls all delivery for an hour. 

Real, highest-impact (`misconfig/attacker-triggerable`). Note: an HTTP-date Retry-After is harmless (float() raises → suppressed), so only numeric-seconds values trigger it — still a real bug.

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -67,6 +68,11 @@
 # Bounds initial-state push so subscribe_to_events does not stall on a
 # subscriber whose __call__ blocks (e.g. WS with a full TCP send buffer).
 INITIAL_STATE_PUSH_TIMEOUT_SECONDS = 0.5
+# How long a lingering _run_task may stay not-done while execution_status
+# reports non-RUNNING before run() reaps it and starts fresh (#3842). Must
+# comfortably exceed the legitimate wrap-up tail (wait_for_pending is bounded
+# at 30s) so a normal finishing run is never cancelled.
+STALE_RUN_TASK_REAP_SECONDS = 120.0
```
VascoSch92, line 75: as before

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -90,6 +96,15 @@ class EventService:
         default_factory=lambda: PubSub[Event](max_subscribers=50), init=False
     )
     _run_task: asyncio.Task | None = field(default=None, init=False)
+    # Stale run-task detection (#3842): a _run_task that never completes while
```
VascoSch92, line 99:
```suggestion
    # Stale run-task detection: a _run_task that never completes while
```

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -992,8 +1051,12 @@ async def _run_and_publish():
...
                         )
 
-                    # Clear task reference and publish state update
-                    self._run_task = None
+                    # Clear task reference and publish state update. Only
+                    # clear our own reference: a task reaped as stale (#3842)
+                    # may reach this finally after a replacement run has
+                    # already been started, and must not clobber it.
+                    if self._run_task is asyncio.current_task():
+                        self._run_task = None
```
VascoSch92, line 1059: Both guards are literally `if (self._run_task is not None and not self._run_task.done()) or (...): raise ValueError("conversation_already_running")` with no `_maybe_reap_stale_run_task()` call. 

Only `run()` self-heals. 

This can confirmed with the goal command. In fact,  /goal start/resume stay permanently 409 on a wedged task.

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -902,6 +917,45 @@ def _token_streaming_callback(chunk: LLMStreamChunk | str) -> None:
...
+        ):
+            return False
+        logger.warning(
+            "Reaping stale _run_task for conversation %s: task still not done "
+            "%.0fs after execution_status first reported non-RUNNING; "
+            "cancelling it so the conversation can run again.",
+            self.stored.id,
+            now - self._stale_run_first_seen,
+        )
+        task.cancel()
```
VascoSch92, line 953: Sync path is `await loop.run_in_executor(self._run_executor, conversation.run)`; reaper does only `task.cancel(); _run_task = None`, then `run()` starts a fresh task. `cancel()` cannot stop the running thread → two runs on one conversation. 

Real for the sync-executor wedge (which is exactly the #3842 scenario; the async path can't wedge without freezing the loop that runs the reaper).

VascoSch92 (commented):
Left a couple of comments. 

Could you also trim the comments and not reference the issue :-) ?

Thanks

PR #3994, opened 2026-07-05
feat(skills): add path-triggered rules

### LINKED ISSUE #3984: [Investigation] Discussion of Claude Rules
## Summary

This issue captures some primary analysis of **Claude Code Rules** and opens a discussion about whether/how to add similar functionality to `software-agent-sdk`. It is intended as a starting point for broader community discussion rather than a concrete implementation proposal.

Motivation comes from user feedback like this:

> We are currently seeing huge improvements in our coding results by migrating a share of our skills into Claude Code rules. The rules work so great because they come with a **deterministic trigger**, such as filename or path patterns. If a file matching that pattern is touched, the rule content is **guaranteed** to be loaded and processed by the LLM. The deterministic nature of it solves the problem of the model failing to make consistent skill-usage decisions. With the reduced number of skills, the amount of frontmatter loaded into the context has reduced significantly as well.

## What are Claude Code Rules?

Rules are markdown files (with YAML frontmatter) that live under `.claude/rules/` (project level) and `~/.claude/rules/` (user level):

```markdown
---
paths:
  - "src/api/**/*.ts"
  - "**/*.test.tsx"
---
When touching API or test files, follow these conventions: ...
```

Two flavors:

- **Path-scoped rules** (`paths:` globs) — content is injected **deterministically** when the agent *touches a file matching the glob* (read/edit). No model decision is involved.
- **Always rules** (no `paths:`) — loaded every session, like `CLAUDE.md`.

The same `paths` field has also been extended to Skills, so rules and skills increasingly share the path-scoping mechanism.

## Why it matters (UX / DevEx)

- **Determinism** — the biggest draw. Context is injected based on *what files are being touched*, not on the model semantically deciding to invoke a skill. This eliminates the "guessing game" where an under-specified prompt (e.g. *"implement the backend API"*) causes the model to skip a relevant skill entirely.
- **Token efficiency** — path-scoped content stays out of the always-on context/frontmatter catalog and only loads when relevant. Users report meaningfully smaller baseline context after migrating skills to rules.
- **Granularity across directories** — a single rule with a glob like `**/*.test.ts` applies to test files everywhere, avoiding a `CLAUDE.md`-in-every-directory maintenance burden.
- **Simple authoring** — plain markdown + a `paths` list; easy to share across repos (e.g. via symlinks).

### Known rough edges (worth learning from)

- In Claude Code, path rules reliably fire on **Read/Edit** but **not on Write/create** of a new matching file — the exact "enforce conventions on file creation" case where they'd be most valuable ([anthropics/claude-code#23478](https://github.com/anthropics/claude-code/issues/23478)). A clean-room implementation could do better here.
- Overly broad globs (e.g. `**/*.ts` in a TS-heavy repo) effectively become "always" rules, undermining the token-efficiency benefit.

---

Follow-up comments below cover **(1)** what `software-agent-sdk` already has that we can build on, and **(2)** the gaps and a recommended approach. Feedback and alternative framings welcome.

_This issue was created by an AI agent (OpenHands) on behalf of @jpshackelford._

VascoSch92:
HUMAN:

Add path-triggered skills, which behave like claude code rules.
You can fin a ready to run event in `.pr`

---

AGENT:

## Why

Claude Code "Rules" (issue #3984) inject guidance **deterministically**: when a
file matching a glob is touched, the rule is guaranteed to load — no reliance on
the model choosing a skill, and no always-on frontmatter cost. This adds that to
the SDK by extending the existing Skill machinery rather than building a parallel
Rules subsystem.

## Summary

- **New `PathTrigger`** — a skill with `paths:` glob frontmatter. Its content is
  injected when the agent reads/edits/creates a matching file (gitignore-style
  `**` globs, matched against the workspace-relative POSIX path). Fires on
  `create` too, beating Claude Code's Write/create gap.
- **Zero baseline cost** — path rules are excluded from `<available_skills>` and
  are not model-invocable (`disable_model_invocation` is forced), so they never
  sit in always-on context.
- **Reuses existing machinery** — a rule is just a skill, so loading, precedence,
  and serialization are unchanged (no new directory or concept). Injection rides
  on a new `ObservationEvent.extended_content`, deduped once per conversation via
  `ConversationState.activated_path_rules`.

### REST API contract changes

Compared with base OpenAPI `61b7db85f6ea` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -1730,0 +1731 @@
+schema ObservationEvent property extended_content optional schema=type="array" items=TextContent
@@ -1752,0 +1754,3 @@
+schema PathTrigger property paths required schema=type="array" items=type="string"
+schema PathTrigger property type optional schema=type="string" const="path" default="path"
+schema PathTrigger type="object"
@@ -1968 +1972 @@
-schema Skill-Input property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger],type="null"]
+schema Skill-Input property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger,PathTrigger],type="null"]
@@ -1984 +1988 @@
-schema Skill-Output property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger],type="null"]
+schema Skill-Output property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger,PathTrigger],type="null"]
```

### REST API contract changes

Compared with base OpenAPI `61b7db85f6ea` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -1730,0 +1731 @@
+schema ObservationEvent property extended_content optional schema=type="array" items=TextContent
@@ -1752,0 +1754,3 @@
+schema PathTrigger property paths required schema=type="array" items=type="string"
+schema PathTrigger property type optional schema=type="string" const="path" default="path"
+schema PathTrigger type="object"
@@ -1968 +1972 @@
-schema Skill-Input property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger],type="null"]
+schema Skill-Input property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger,PathTrigger],type="null"]
@@ -1984 +1988 @@
-schema Skill-Output property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger],type="null"]
+schema Skill-Output property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger,PathTrigger],type="null"]
```

### REST API contract changes

Compared with base OpenAPI `ebdb4566aca5` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -1818,0 +1819 @@
+schema ObservationEvent property extended_content optional schema=type="array" items=TextContent
@@ -1840,0 +1842,3 @@
+schema PathTrigger property paths required schema=type="array" items=type="string"
+schema PathTrigger property type optional schema=type="string" const="path" default="path"
+schema PathTrigger type="object"
@@ -2056 +2060 @@
-schema Skill-Input property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger],type="null"]
+schema Skill-Input property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger,PathTrigger],type="null"]
@@ -2072 +2076 @@
-schema Skill-Output property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger],type="null"]
+schema Skill-Output property trigger optional schema=anyOf=[oneOf=[KeywordTrigger,TaskTrigger,PathTrigger],type="null"]
```

## Issue Number

#3984

## How to Test

**End-to-end** — real `file_editor` tool + real `Conversation.run()`, scripted
`TestLLM` (no API key). Committed at `.pr/demo_path_rules.py`:

```
uv run python .pr/demo_path_rules.py
```

Output:

```
====================================================================
rule loaded          : trigger=PathTrigger paths=['src/api/**/*.ts']
advertised in catalog: False  (should be False)
invocable by model   : False  (should be False)
activated_path_rules : ['api-rule']
LLM sees the rule after touching src/api/users.ts: True
====================================================================
```

The agent creates `src/api/users.ts` through the real file-editor tool; the rule
body (`API RULE: validate all request inputs with zod.`) is folded into the tool
observation the LLM reads on its next step, while the rule stays absent from the
catalog and non-invocable.

**Tests**

```
uv run pytest tests/sdk/skills/test_path_trigger.py \
  tests/sdk/conversation/test_path_rules_injection.py \
  tests/tools/test_path_rules_file_editor.py
# 61 passed
```

Regression (unchanged behavior): the full `tests/sdk/{skills,context,event,tool}`
and `tests/sdk/conversation` suites pass. `pre-commit` (ruff, ruff-format,
pyright, import-dependency rules) is clean on every changed file.

## Video/Screenshots

CLI feature — text demo output shown above (no GUI).

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- A skill is **either** path-triggered **or** model-invocable, not both: if a file
  declares both `paths:` and `triggers:`, `paths:` wins. This is deliberate (keeps
  rules deterministic and out of the catalog); easy to make them combinable if
  that's preferred.
- ACP conversations do not get injection (ACP servers own tool execution, so file
  edits don't flow through the SDK observation seam) — a known boundary, not a
  regression.
- Follow-ups intentionally out of scope: "always" rules (empty `paths:` →
  `REPO_CONTEXT`) and packaging rules inside Plugins.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:39997a3-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-39997a3-python \
  ghcr.io/openhands/agent-server:39997a3-python
```

```
ghcr.io/openhands/agent-server:39997a3-golang-amd64
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-golang-amd64
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-golang-amd64
ghcr.io/openhands/agent-server:39997a3-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:39997a3-golang-arm64
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-golang-arm64
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-golang-arm64
ghcr.io/openhands/agent-server:39997a3-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:39997a3-java-amd64
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-java-amd64
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-java-amd64
ghcr.io/openhands/agent-server:39997a3-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:39997a3-java-arm64
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-java-arm64
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-java-arm64
ghcr.io/openhands/agent-server:39997a3-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:39997a3-python-amd64
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-python-amd64
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-python-amd64
ghcr.io/openhands/agent-server:39997a3-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:39997a3-python-arm64
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-python-arm64
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-python-arm64
ghcr.io/openhands/agent-server:39997a3-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:39997a3-golang
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-golang
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-golang
ghcr.io/openhands/agent-server:39997a3-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:39997a3-java
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-java
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-java
ghcr.io/openhands/agent-server:39997a3-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:39997a3-python
ghcr.io/openhands/agent-server:39997a3ca6b09005c36cb7bb7f17964574a9aab3-python
ghcr.io/openhands/agent-server:3984-investigation-discussion-of-claude-rules-python
ghcr.io/openhands/agent-server:39997a3-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `39997a3-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `39997a3-python-amd64`) are also available if needed


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -465,6 +465,84 @@ def wrapped(event: Event) -> None:
...
+
+    def _touched_rule_path(self, event: ObservationEvent) -> str | None:
+        """Return the workspace-relative POSIX path a tool observation touched.
+
+        Correlates the observation to its ``ActionEvent`` and reads the action's
+        ``path`` field generically (no dependency on the tools package). Returns
+        None when the action has no file ``path`` or the path is outside the
+        workspace.
+        """
+        try:
```
VascoSch92: with contextlib.SUppression.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -465,6 +465,84 @@ def wrapped(event: Event) -> None:
...
+        if not isinstance(raw_path, str) or not raw_path:
+            return None
+
+        raw = PurePosixPath(raw_path.replace("\\", "/"))
+        if not raw.is_absolute():
+            return str(raw)
+        root = PurePosixPath(str(self.workspace.working_dir).replace("\\", "/"))
+        try:
+            return str(raw.relative_to(root))
+        except ValueError:
```
VascoSch92: as above

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -465,6 +465,84 @@ def wrapped(event: Event) -> None:
 
         return cast(ConversationCallbackType, wrapped)
 
+    def _rules_injecting(
+        self, inner: ConversationCallbackType
+    ) -> ConversationCallbackType:
```
VascoSch92, line 474: trim all the docstring.

tests/sdk/conversation/test_path_rules_injection.py
```diff
@@ -0,0 +1,277 @@
+"""Tests for the path-rule injection seam in LocalConversation.
+
+When the agent touches a file whose path matches a PathTrigger skill ("rule"),
+the rule content is appended to the resulting ObservationEvent's
+``extended_content`` and deduped via ``state.activated_path_rules``.
+"""
+
+from __future__ import annotations
```
VascoSch92: we don't need this

tests/sdk/skills/test_path_trigger.py
```diff
@@ -0,0 +1,248 @@
+"""Tests for path-scoped skills ("rules"): PathTrigger, glob matching, loading,
+partition exclusion, and the AgentContext tool-use injection matcher."""
+
+from __future__ import annotations
```
VascoSch92: delete

tests/tools/test_path_rules_file_editor.py
```diff
@@ -0,0 +1,146 @@
+"""End-to-end check that path rules fire on the REAL file-editor action.
+
+The injection seam reads the touched path from the action's ``path`` field
+(``local_conversation._touched_rule_path``). These tests use the actual
+``FileEditorAction`` / ``FileEditorObservation`` from openhands-tools so a
+rename of that field (which would silently no-op the whole feature) is caught.
+"""
+
+from __future__ import annotations
```
VascoSch92: delete

jpshackelford (commented):
# Path-Triggered Skills Testing Summary

## 🔬 How I Tested

### Test 1: Plugin Integration Demo
- **Created** a plugin with 2 path-triggered skills
- **Loaded** the plugin using `Plugin.load()`
- **Executed** file operations that matched the path patterns
- **Tool**: TestLLM (scripted responses, no API calls)

### Test 2: Keyword Deception Test
- **Created** 3 rules with DISTINCT themes (animals, colors, numbers)
- **Used** prompts with MISLEADING keywords (mention animals, but create frontend file)
- **Verified** only path-matching rule fires (not keyword-matching)
- **Tool**: TestLLM (full control over which files are touched)

### Test 3: Real LLM Verification
- **Used** generic prompts with NO theme keywords
- **Let LLM decide** how to create files
- **Verified** correct rules fire based on file paths
- **Tool**: Real LLM via LiteLLM proxy (gpt-4o-mini)

## ✅ Results

### Plugin Integration - **PASS** ✅
```
✓ Created plugin with 2 path-triggered skills
✓ Loaded plugin successfully
✓ Both skills activated when matching files touched
✓ Rules injected into LLM context via extended_content
✓ Deduplication worked correctly

Activated path rules: ['typescript-rules', 'test-rules']
```

### Keyword Deception Test - **ALL PASS** ✅
```
Test 1: Prompt with ANIMAL keywords → created frontend/app.js
        Result: COLORS rule fired (not animals!) ✅

Test 2: Prompt with COLOR keywords → created backend/api.py  
        Result: ANIMALS rule fired (not colors!) ✅

Test 3: No keywords → created src/utils.py
        Result: NO rules fired ✅

Test 4: Numbers keywords → created data/schema.sql
        Result: NUMBERS rule fired ✅
```

**PROVEN**: Keywords in prompts have ZERO effect - only file paths trigger rules!

### Real LLM Test - **4/5 PASS** ✅
```
Test 1: Generic prompt → LLM created backend/api.py
        Result: ANIMALS rule fired ✅

Test 2: Generic prompt → LLM created frontend/app.js
        Result: COLORS rule fired ✅

Test 3: Generic prompt → LLM created data/schema.sql
        Result: NUMBERS rule fired ✅

Test 4: Generic prompt → LLM created src/utils.py
        Result: NO rules fired ✅

Test 5: Nested path backend/nested/deep/file.py
        Result: Not fired (already activated - deduplication) ℹ️
```

## 🎓 Key Findings

1. **✅ Plugin Support**: Path-triggered skills work with plugins **out of the box** - no additional code needed

2. **✅ Hook Compatibility**: Hooks and path-triggered skills are **complementary** - they work together without conflicts

3. **✅ Path-Based Triggering**: Rules fire based on **glob pattern matching** on file paths, NOT semantic analysis or keyword detection

4. **✅ Deterministic**: Same path always triggers the same rule(s)

5. **✅ Zero Baseline Cost**: Rules excluded from `<available_skills>` catalog until triggered

6. **✅ Deduplication**: Each rule fires once per conversation (prevents redundant injections)

## 🚀 Conclusion

**Path-triggered skills work exactly as designed:**
- Triggers are 100% path-based (proven with keyword deception tests)
- Works seamlessly with plugins and hooks
- Tested with both scripted and real LLM
- Deterministic, efficient, and reliable

openhands-sdk/openhands/sdk/context/agent_context.py
```diff
@@ -296,6 +296,8 @@ def _partition_skills(self) -> tuple[list[Skill], list[Skill]]:
         available_skills: list[Skill] = []
         for s in self.skills:
             if s.is_agentskills_format or s.trigger is not None:
+                # Path rules force disable_model_invocation (Skill validator), so
+                # they fall out here — injected only on file-touch, never listed.
```
enyst, line 318: Do I understand correctly, they cannot be set on directories? If so, I think per directories is actually more useful... 🤔

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -465,6 +465,77 @@ def wrapped(event: Event) -> None:
...
+        """Return the workspace-relative POSIX path a tool observation touched.
+
+        Correlates the observation to its ``ActionEvent`` and reads the action's
+        ``path`` field generically (no dependency on the tools package). Returns
+        None when the action has no file ``path`` or the path is outside the
+        workspace.
+        """
+        action_event: Event | None = None
+        with contextlib.suppress(KeyError):
+            idx = self._state.events.get_index(event.action_id)
```
enyst, line 523: Does this search across all events? @openhands I think we need lower complexity than O(n) here, what is the case?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -1016,7 +1090,7 @@ def _ensure_plugins_loaded(self) -> None:
                 visualizer=self._visualizer,
                 conversation_stats=self._state.stats,
             )
-            self._on_event = self._tree_stamping(raw_on_event)
+            self._on_event = self._tree_stamping(self._rules_injecting(raw_on_event))
```
enyst, line 1111: This reads a bit strange. Could it be triggered `on_event` ... rather than outside it, if that makes sense?

openhands-sdk/openhands/sdk/event/llm_convertible/observation.py
```diff
@@ -46,12 +53,19 @@ def visualize(self) -> Text:
...
+                "\n\nPrompt Extension based on Agent Context:\n", style="bold"
+            )
+            content.append(" ".join(content_to_str(self.extended_content)))
         return content
 
     def to_llm_message(self) -> Message:
         return Message(
             role="tool",
-            content=self.observation.to_llm_content,
+            content=list(self.observation.to_llm_content) + list(self.extended_content),
```
enyst, line 68: Maybe the `extended_content` belongs in `to_llm_content`, rather than added from outside? I think `to_llm_content` means "transform this Event's _everything_ into a message for the LLM"

openhands-sdk/openhands/sdk/context/agent_context.py
jpshackelford: The system uses .gitignore type glob patterns and I verified that it does work with paths such as:

```
docs/**
"src/**/*.py" 
```

jpshackelford: Oh I see what you mean -- yes it only works on file operations.  A directory operation like the following doesn't trigger the rule.

```
cd docs/  
```

VascoSch92: Do we want also directories?

enyst: Actually, maybe not! All I think matters… is that we insert something like this:
https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-agent-server/AGENTS.md
… when it uses a file in `agent-server`. 🤔

So I think it’s fine as is. If it doesn’t edit/view a file, why would it matter?

enyst: This means we need to … transform AGENTS.md into a skill with frontmatter? Or make a skill that redirects to it 🤔

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: No, it does not scan all events. 

This is already O(1), not O(n).

`self._state.events` is an `EventLog`, and `get_index` is a plain dict lookup:

```python
class EventLog(EventsListBase):
    def __init__(self, ...):
        self._id_to_idx: dict[EventID, int] = {}   # built once at scan time
        self._idx_to_id: dict[int, EventID] = {}
        self._event_cache: dict[int, Event] = {}
        ...

    def get_index(self, event_id: EventID) -> int:
        try:
            return self._id_to_idx[event_id]     # O(1) hash lookup
        except KeyError:
            raise KeyError(f"Unknown event_id: {event_id}")
```

`_id_to_idx` is a `dict[EventID, int]` populated in `_scan_and_build_index()` and kept current as events are appended, so mapping an event_id → index is a single hash lookup, O(1).

The follow-up `self._state.events[idx]` (`__getitem__` → `_get_single_item`) is also O(1): it hits _`event_cache.get(i)`, or on a miss reads the single file at `_path(i)` computed directly from the index, again no scan.

So the whole action→observation correlation in `_touched_rule_path` is O(1). The O(n) concern doesn't apply here; nothing walks the event list.

openhands-sdk/openhands/sdk/event/llm_convertible/observation.py
VascoSch92: It's actually already at the right layer, and it isn't "from outside." 

The subtlety is which object owns which method:

- `to_llm_content` is a property of the inner `Observation`: the tool's own result schema, produced by tool execution.
- `to_llm_message()` is the method on the `ObservationEvent`: the Event-level "turn this whole event into a Message" transform enyst is describing.
- `extended_content` is a field on the `Event`, and it's merged inside the Event's own `to_llm_message()`:

```python
def to_llm_message(self) -> Message:
    return Message(
        role="tool",
        content=list(self.observation.to_llm_content) + list(self.extended_content),
        ...
    )
```

So the merge already happens inside the Event's transform, not outside it. Pushing `extended_content` down into `Observation.to_llm_content` would be the wrong layer: the `Observation` is the tool's result payload (owned by the tool executor), whereas path-rule content is agent-context injected at the conversation layer — it isn't tool output, and folding it in would pollute str(), the observation schema, and serialization.

Moreover, this exactly mirrors the existing MessageEvent (message.py), which has done the same thing for keyword/task-triggered skills since before this PR:

```python
# MessageEvent.to_llm_message()
msg.content = list(msg.content) + list(self.extended_content)   # activated_skills + extended_content
```

`ObservationEvent` is deliberately built as the tool-observation twin of that pattern (same extended_content field, same "Prompt Extension based on Agent Context" visualize block). 

So the current placement is the consistent choice, not an outlier.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: `_rules_injecting` is part of `on_event` — it's a link in the callback chain, not something wrapped around the outside. `self._on_event` is composed from three concerns layered as decorators:

- `raw_on_event` (innermost): from the shared `create_hook_callback` factory: runs hooks, persists the event, notifies subscribers.
- `_rules_injecting`: transforms the event (`inner(self._maybe_inject_path_rules(event))`) so a matching file-touch gets extended_content before it's persisted/broadcast.
- `_tree_stamping` (outermost): stamps `parent_id` / advances HEAD.

This follows the pattern `_tree_stamping` already established in the same file, path-rule injection just adds a second decorator in the same idiom. The reason it isn't folded into raw_on_event: that callback comes from a shared factory (create_hook_callback) that knows nothing about path rules and shouldn't; injection needs conversation self (agent_context, the activated_path_rules dedup set, the state lock). Keeping it as a conversation-level wrapper keeps each concern independently testable and out of the hook factory.

openhands-sdk/openhands/sdk/context/agent_context.py
VascoSch92: I think what do you mean is another feature. If you enter a repo you should load all the AGENTS.md file which are inside and inject them into the agent context.

I believe this is what CC is doing.

enyst: You mean codex :) but no. I don't think so:
- all we need is a way to load the .md files when the agent enters the respective directory 
- we could add frontmatter to use this paths PR, in this repo
- the cwd based loading is not great, it's partial, and it seems redundant with this paths feature

VascoSch92: Ok but this is a different feature right?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
enyst: There is no suggestion to insert it in the hook factory, so the last para doesn't seem applicable, sorry.

The line diff above has a smell: for every detail we want to happen on event, we'd come and add to callees here. I think maybe this can be redesigned nicer... But maybe we can think about it

openhands-sdk/openhands/sdk/event/llm_convertible/observation.py
enyst: Very verbose explanation but I think it's correct ! Thank you 🙏

openhands-sdk/openhands/sdk/context/agent_context.py
enyst: No? The functionality is: insert sets of instructions in LLM's context but only for certain files / dirs

VascoSch92: OK but is this still a `PathTriggered` skill? So it is just a change to make it also detect directories?

enyst: And to migrate the contents of the inner AGENTS.md.

So my thinking is: CC and Codex are solving the same problem, with different solutions. OpenHands SDK also semi-solves this problem today, with a Codex-like solution (the prompt-based half of it).

What can we do? A few alternatives:
- (1) implement both solutions
- (2) implement the best solution (we seem to agree it's the CC solution), and migrate our old to it
- (3) implement the CC solution, and don't migrate, just leave our existing stuff in the codebase

IMO, (3) is not good... we want to load these instructions files, and I don't think the agent can be trusted here. I was thinking before that we need SOME path enforcement deterministically

In my reading, you, Vasco, seem to prefer (1). 😸 

I can live with (1), I think it's worth thinking about (2) though; it's clean, and simpler over time IMHO.

enyst: Ref: [what are these solutions](https://openhands-ai.slack.com/archives/C06R25BT5B2/p1783413537905639?thread_ts=1783329156.524489&cid=C06R25BT5B2)

VascoSch92: @enyst there are no number in your answer. So i don't know exactly at what you are pointing.

enyst: Fixed 🙏

jpshackelford:
## Suggestion: Add log warning for dropped triggers

When a skill defines both `paths:` and `triggers:`, the `paths:` takes precedence and `triggers:` is silently ignored. This could be confusing for users creating skills who don't realize one of their fields is being ignored.

### Minimal in-scope fix

Add a `logger.warning()` in `_create_skill_from_metadata` when this happens:

```python
if paths:
    # Warn if both paths and triggers/inputs are defined
    if keywords or "inputs" in metadata_dict:
        ignored = []
        if keywords:
            ignored.append(f"'triggers': {keywords}")
        if "inputs" in metadata_dict:
            ignored.append("'inputs'")
        
        logger.warning(
            f"Skill '{agent_name}' has both 'paths:' and {', '.join(ignored)} defined. "
            f"Using 'paths:' (path-triggered rule). {', '.join(ignored)} will be ignored. "
            f"A skill can be either path-triggered OR model-invocable, not both."
        )
```

While this only appears in server logs (not the conversation UI), it at least makes the behavior **discoverable** for debugging.

### Follow-up issue

For user-visible skill loading warnings in the conversation UI (similar to `HookExecutionEvent`), I've opened #4015 which describes the broader infrastructure needed. That would be out of scope for this PR but valuable for future work.

---

*This comment was created by an AI agent (OpenHands) on behalf of the reviewer.*

openhands-sdk/openhands/sdk/skills/skill.py
```diff
@@ -635,6 +636,26 @@ def _handle_third_party(cls, path: Path, file_content: str) -> Union["Skill", No
...
 
+    @classmethod
+    def _handle_nested_third_party(
+        cls, path: Path, file_content: str, rel_dir: PurePosixPath
+    ) -> Union["Skill", None]:
+        """Build a ``PathTrigger`` rule scoped to ``rel_dir`` from a nested
+        third-party file, so its guidance is injected only when a file under that
+        directory is touched. ``rel_dir`` is workspace-relative (the matcher's
+        base)."""
+        skill_name = cls.PATH_TO_THIRD_PARTY_SKILL_NAME.get(path.name.lower())
```
enyst, line 647: Just a side note here, I think `AGENTS.md` is enough, TBH. I would hope that someday the other two are not necessary at all, but for now they’re only necessary in root directory. e.g. I don’t think we need to support `CLAUDE.md` in subdirectories, unless maybe we know they happen in Claude Code

enyst:
@openhands Describe to me the behavior of this PR when the repository looks like this repository itself : that is, as you can see, it has nested AGENTS.md files, not only in root dir, but also in some subdirs. This PR attempts to load them sometimes: describe to me when it loads e.g. AGENTS.md from openhands-agent-server.

Respond directly as a comment on github here.

Then. Once you responded. Take a deep breath, and /codereview this PR. Post your feedback also on the PR.

enyst:
@enyst Good question — this repo is actually a perfect stress test, because it has an `AGENTS.md` at the root **and** five nested ones. Here's exactly what this PR does with them (numbers below are from running the real loader against this checkout).

### Two different code paths for root vs. nested

The root `AGENTS.md` is **not** a path rule. It's picked up by `find_third_party_files()` and loaded as an always-on repo skill (`trigger=None`), exactly as before this PR. It sits in permanent context regardless of what files you touch.

The **nested** ones are the new behavior. `find_nested_third_party_files()` walks the worktree (via `git ls-files --cached --others --exclude-standard`, i.e. tracked + non-ignored files), finds every `AGENTS.md` that is *not* at the top level, and `_handle_nested_third_party()` turns each into a `PathTrigger` "rule" scoped to its own directory. On this checkout that produces:

```
agents:openhands-agent-server                  -> ['openhands-agent-server/**']
agents:openhands-sdk/openhands/sdk             -> ['openhands-sdk/openhands/sdk/**']
agents:openhands-sdk/openhands/sdk/subagent    -> ['openhands-sdk/openhands/sdk/subagent/**']
agents:openhands-tools/openhands/tools         -> ['openhands-tools/openhands/tools/**']
agents:openhands-workspace/openhands/workspace -> ['openhands-workspace/openhands/workspace/**']
```

The rule name encodes the directory (`agents:<rel_dir>`) so the five files that all map to the `agents` third-party name stay distinct for dedup/activation tracking.

### When does `openhands-agent-server/AGENTS.md` load?

That file becomes the rule `agents:openhands-agent-server` with glob `openhands-agent-server/**`. It is injected the **first time the agent touches a file whose workspace-relative POSIX path matches that glob** — read, edit, *or* create through the file-editor tool (any tool whose action carries a `path`). Verified against the real matcher:

| Touched path | Injected? |
|---|---|
| `openhands-agent-server/openhands/agent_server/config.py` | ✅ yes |
| `openhands-agent-server/pyproject.toml` | ✅ yes |
| `openhands-agent-server/AGENTS.md` (the file itself) | ✅ yes |
| `openhands-agent-server` (the bare dir path) | ❌ no — `**` requires a segment after the slash |
| `openhands-sdk/openhands/sdk/llm/llm.py` | ❌ no |
| `openhands-agent-server-other/x.py` | ❌ no — not fooled by the shared prefix |

Mechanically: when the tool `ObservationEvent` comes back, `_maybe_inject_path_rules()` correlates it to its `ActionEvent`, reads the action's `path`, relativizes it against the workspace root, and asks the agent context for matching `PathTrigger` skills. The matched rule's **entire `AGENTS.md` body** is folded into that observation's new `extended_content`, so the LLM sees it appended right after the tool result on its next step — wrapped in an `<EXTRA_INFO>` block explaining the rule fired because a touched file matched `openhands-agent-server/**`.

### Three behaviors worth calling out for this repo specifically

1. **Once per conversation.** Injection is deduped via `state.activated_path_rules`. The first touch under `openhands-agent-server/` injects the whole file; every subsequent touch under that dir is silent. So it isn't re-paid on every edit.

2. **Nested overlap = multiple rules fire.** The globs are prefix-scoped, so touching `openhands-sdk/openhands/sdk/subagent/foo.py` matches **both** `openhands-sdk/openhands/sdk/**` *and* `openhands-sdk/openhands/sdk/subagent/**` (I confirmed both match). Editing a subagent file therefore injects *both* the `sdk` AGENTS.md and the `sdk/subagent` AGENTS.md (each once). That's arguably the intended "parent + child guidance" behavior, but worth being deliberate about — the parent file is large.

3. **Never advertised, never model-invocable.** These rules are forced to `disable_model_invocation=True` and excluded from `<available_skills>`, so they cost nothing until a matching file is actually touched. Injection only happens through the SDK observation seam, so ACP-backed conversations don't get it (a known boundary noted in the PR).

One caveat: matching is entirely on the **workspace-relative** path, and only fires when the touched file resolves *inside* the workspace root (absolute paths are relativized; symlinked roots are handled via a `resolve()` fallback). If the workspace working dir were pointed at a subdirectory rather than the repo root, the relative paths — and thus which nested files load — would shift accordingly.

---
*This comment was generated by an AI agent (OpenHands) on behalf of @enyst.*

enyst (commented):
## Code Review — feat(skills): add path-triggered rules

🟡 **Acceptable** — the core design is genuinely good taste: a "rule" is just a `Skill` with a `PathTrigger`, so loading, precedence, serialization, and dedup all ride on machinery that already exists instead of a parallel Rules subsystem. The injection seam (`ObservationEvent.extended_content`, mutated on a frozen model via `model_copy`, appended in `to_llm_message`) is clean and mirrors the existing message-suffix path. Tests exercise real code paths (real `LocalConversation`, real events, real file-editor tool — no "assert the mock was called"), and they pass locally (76 passed). Nice work.

I'm **not approving**, for one policy reason and a couple of small nits — details below.

- **[local_conversation.py:490] `getattr` guard on a declared field.** `agent_context` is a declared field on `AgentBase` (`agent/base.py:209`), so `getattr(self.agent, "agent_context", None)` can just be `self.agent.agent_context`. The repo conventions explicitly discourage `getattr`/`hasattr` guards in favor of typed access. I see this mirrors the pre-existing pattern at line 418, so it's not net-new — but since you're adding a fresh call site, worth doing it the typed way (and maybe cleaning up 418 while you're here).

- **[local_conversation.py:525] `getattr(action_event.action, "path", None)` is an implicit contract.** This is a deliberate, pragmatic decoupling from the tools package (you call it out in the code comment, 👍). The flip side is that *any* action with a `path` field now drives rule matching. Today that's fine (file-editor's `path` is a workspace file path), but it silently assumes "a field named `path` == a workspace file path." If a future tool adds a `path` field meaning something else (a URL path, a registry key), it would spuriously trigger rules. Not blocking, but the assumption deserves to be explicit — either a short note in the docstring or a follow-up when a second path-bearing tool appears.

- **[agent_context.py:536] `if not isinstance(skill, Skill)`** on a `list[Skill]` is a redundant defensive guard (mirrors the existing one at 489). Minor; skip if you're matching the sibling method's style deliberately.

### [BEHAVIORAL NOTE — nested/overlapping rules]

Because rule globs are directory-prefix-scoped (`<dir>/**`), **overlapping nested rules all fire**. In a repo shaped like this one, touching `openhands-sdk/openhands/sdk/subagent/foo.py` matches both `openhands-sdk/openhands/sdk/**` *and* `openhands-sdk/openhands/sdk/subagent/**`, so it injects **both** the parent `sdk/AGENTS.md` and the child `sdk/subagent/AGENTS.md` (each once). That's a reasonable "parent + child guidance" semantic, but it's an intentional-looking design decision with real context-cost implications (the parent `AGENTS.md` here is large), and I don't see a test pinning that behavior. Consider adding one so the "inherit up the tree" contract is explicit and can't silently regress into "only the deepest match fires" (or vice-versa). See my [walkthrough comment](https://github.com/OpenHands/software-agent-sdk/pull/3994#issuecomment-4917245399) for the full breakdown of how this repo's six AGENTS.md files get partitioned.

### [TESTING]

Real tests, no mock-theater — good. Two gaps worth a line each, neither blocking:
- No test for the nested-overlap double-injection described above.
- No test that a non-file tool carrying an unrelated `path` field does / doesn't trigger (locks in the implicit contract from the `getattr` note).

Backward-compat on the new `ObservationEvent.extended_content` field is fine: it has a `default_factory`, and `extra="forbid"` only rejects *unknown* keys, so old persisted observations load unchanged.

### [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟡 **MEDIUM**

This PR deterministically injects new content into the LLM's context on file-touch — that is a change to what the model sees on every matching step, which squarely lands in the eval/benchmark-risk category for this repo. The mechanism itself is well-gated (deduped once per conversation, excluded from `<available_skills>`, forced non-invocable, backward-compatible event field), so the *engineering* risk is low-to-medium. The *behavioral* risk is the open question: injecting large nested `AGENTS.md` bodies mid-conversation can shift agent behavior and benchmark scores in ways unit tests can't catch.

Per this repo's review policy, I can't approve an agent-behavior/prompt-injection change without eval evidence. I don't see an `openhands-eval-monitor.vercel.app` link in the description or comments.

**Recommendation:** Do not auto-merge. A human maintainer should run a lightweight eval and confirm no regression before this ships. Once there's an eval-monitor link + maintainer sign-off, this is close to mergeable modulo the two small nits above.

**VERDICT:** 🟣 **Commenting, deferring to human maintainer** — solid, well-tested implementation; blocked only on eval validation (repo policy) plus a couple of optional cleanups.

**KEY INSIGHT:** Modeling a "rule" as a path-triggered skill is the right call — it eliminates a whole parallel subsystem — but the moment rules inject full nested `AGENTS.md` bodies into live context, this becomes a benchmark-sensitive change that needs eval proof, not just green unit tests.

---
> **Improve this review?** If any feedback here seems incorrect or irrelevant to this repository, add/edit `.agents/skills/custom-codereview-guide.md` on your branch with the `/codereview` trigger and the missing context, then re-request review. See the [customization docs](https://docs.openhands.dev/openhands/usage/use-cases/code-review#customization).
>

---
*This review was generated by an AI agent (OpenHands) on behalf of @enyst.*

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -465,6 +465,84 @@ def wrapped(event: Event) -> None:
...
+        """Return ``event`` with matching path-rule content, or unchanged.
+
+        Only ``ObservationEvent``s carrying a file path are considered. Matching
+        rules already injected in this conversation are skipped via
+        ``state.activated_path_rules``.
+        """
+        if not isinstance(event, ObservationEvent):
+            return event
+
+        if (agent_context := getattr(self.agent, "agent_context", None)) is None:
```
enyst:
```suggestion
        if (agent_context := self.agent.agent_context) is None:
```
Probably?

enyst:
@openhands Read the above comment (the last comment) and see this :

> local_conversation.py:525] getattr(action_event.action, "path", None) is an implicit contract

and all the improvement opportunity discussed about `path`. Open a new issue where you document it, and propose a few alternative options (not for this PR, for a follow up clean up). Note that importing tools is not an option; but surely we can think of a better code design pattern?

Do not modify this PR.

enyst (approved):
Looks good to me, thank you for addressing the alternative AGENTS.md issue nicely! 

(Not for this PR, maybe then we can also clean up the root file.)

enyst:
@enyst Done — opened #4038 to document the implicit `getattr(action_event.action, "path", None)` contract and propose follow-up alternatives (not for this PR).

It captures:
- **Why it's fragile**: the SDK depends on the *string literal* `"path"` matching a field on tool actions it can't import/type-check, and it **already silently misses tools** — Gemini file tools use `file_path` and apply_patch uses patch text / `move_path`, so path rules don't fire there today.
- **The constraint**: no importing `openhands-tools` from `openhands-sdk`.
- **Four alternatives**: (A) a `runtime_checkable` `Protocol` (`touched_paths()`) tools opt into; (B) a field-level `Annotated[..., FsPath()]` semantic marker; (C) *preferred* — the producer reports `affected_paths` on the `Observation` (typed, multi-path, no correlation lookup); (D) a dependency-inverted extractor registry.

Recommendation leans to **(C)** with **(A)** as a lighter option. This PR is left unchanged.

_This comment was created by an AI agent (OpenHands) on behalf of @enyst._

enyst:
cc @VascoSch92 just for visibility:
> **it already silently misses tools** — Gemini file tools use `file_path` and (GPT-5.x) `apply_patch` uses patch text / `move_path`, so path rules don't fire there today.

openhands-sdk/openhands/sdk/skills/skill.py
VascoSch92: Claude Code support CLAUDE.md in sub-directories. So I thoughtthat thi was nice to have easy migration. (actually it is just a rename of a file)

VascoSch92:
> cc @VascoSch92 just for visibility:
> 
> > **it already silently misses tools** — Gemini file tools use `file_path` and (GPT-5.x) `apply_patch` uses patch text / `move_path`, so path rules don't fire there today.

Yes I know. I was just overfitting on the default tools that we are using. People using OH have access just to these tools. But we can have a follow-up for that.

PR #4003, opened 2026-07-06
fix: normalize content to "" instead of dropping it for assistant too…

kanxiao:
HUMAN:

Fix assistant tool-call messages returning HTTP 422 on some LLMs due to content=null.

---

AGENT:

## Why

Some LLMs (e.g. Gause) return HTTP 422 when an assistant message has `tool_calls` and `content=null`. The `_remove_content_if_empty()` method was dropping the `content` key entirely, producing `content=null` in the serialized payload. Some providers require `content` to be present (even if empty string).

## Summary

- Changed `_remove_content_if_empty()` to set `content=""` instead of `message_dict.pop("content", None)`, in both the string and list content branches.
- Updated two unit tests to assert `result["content"] == ""` instead of `assert "content" not in result`.

## Issue Number
N/A

## How to Test

```bash
# Run the unit tests for message serialization
cd openhands-sdk
pytest tests/sdk/llm/test_message.py::test_message_tool_calls_drop_empty_string_content -v
pytest tests/sdk/llm/test_message.py::test_message_tool_calls_strip_blank_list_content -v
pytest tests/sdk/llm/test_message.py::test_message_with_tool_calls -v
pytest tests/sdk/llm/test_message.py::test_empty_assistant_message_uses_string_content_in_list_serializer -v
pytest tests/sdk/llm/test_message.py::test_empty_assistant_from_llm_chat_message_uses_string_content_in_list_serializer -v
```

Expected: all 5 tests pass, and assistant messages with tool_calls + empty content now serialize as {"role": "assistant", "content": "", "tool_calls": [...]} instead of having the content key absent.

## Video/Screenshots

N/A

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This is consistent with the existing _normalize_empty_assistant_content() behavior which already sets content="" for non-tool-call assistant messages. The change only affects the wire format sent to LLM providers — internal Message objects are unchanged.


enyst (commented):

**Taste Rating: 🔴 Needs improvement** — not because the code is bad. The code is four characters. It's because those four characters revert a deliberate fix for a different model, globally, and nobody checked on that model on the way past.

**[CRITICAL ISSUES]**

- [`openhands-sdk/openhands/sdk/llm/message.py`, L411 & L437] **Breaking Change — "we don't break userspace"**: Dropping the `content` key was not an accident anyone left lying around. `git log -S"_remove_content_if_empty"` gives exactly one commit:

  ```
  6b2b671a  Remote empty text content for tool calls & support `kimi-k2-thinking` (#1093)
  ```

  It was added *to make kimi-k2-thinking work*. You are now unconditionally reverting it to make a different provider work. kimi is not a historical curiosity either — it's still first-class in `verified_models.py:68-70,120-121` (`kimi-k2-thinking`, `kimi-k2.5`, `kimi-k2.6`), and `message.py:322` still carries a `# Required for model like kimi-k2-thinking` special case ten lines above the code you changed.

  So the question this PR has to answer, and currently doesn't: **does kimi-k2-thinking still accept `content: ""` alongside `tool_calls`?** If yes, say so in the description and this becomes a trivial merge. If no, you have moved the 422 from one vendor to another and the next PR will move it back. That's not a fix, that's a see-saw.

  The shape that ends the see-saw is a capability check keyed on the model, next to the other model-conditional serialization already living in this file — not a global flip.

- [`openhands-sdk/openhands/sdk/llm/message.py`, L399-401] **Unverifiable Comment**: `Some LLMs (e.g. Gause) return HTTP 422...`. "Gause" appears nowhere else in this repository:

  ```
  $ grep -rni "gause" --include='*.py' --include='*.md' .
  $ # (no matches)
  ```

  I can't tell whether that's a provider I don't know, a codename, or a typo. A source comment that names a specific vendor's HTTP status code has no mechanism to stay true — it'll be wrong within two quarters and nobody will dare delete it. Put the vendor detail in the PR description and the commit message, where `git blame` will find it, and leave the code comment saying *why the shape matters*, not *who complained*.

**[TESTING GAPS]**

- The two updated tests just flipped `assert "content" not in result` → `assert result["content"] == ""`. They now assert the new behavior, which means they can no longer catch the regression they were originally written to catch. `uv run pytest tests/sdk/llm/test_message.py` → **30 passed**, which proves the assertions match the code and nothing else. There is no test that would fail if kimi breaks, because there was never a test for that either — it was only ever encoded in that one commit title. Worth adding one now that you're standing on it.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🔴 **HIGH** — small diff, wide blast radius. This changes the wire format of *every* assistant tool-call message for *every* provider, in the one function that all chat serialization funnels through, to satisfy one unnamed endpoint. **Recommendation**: Do not auto-merge. A human who knows the kimi/Moonshot integration should confirm `content: ""` is accepted there before this lands.

**VERDICT:** ❌ **Needs rework** — or, at minimum, needs one sentence in the description proving kimi-k2-thinking survives it.

**KEY INSIGHT:** Two providers disagree about whether `content` may be absent; a global default can only ever satisfy one of them, so this belongs in the model-capability layer that this file already has, not in a shared serializer.

Issue #4017, opened 2026-07-07
Simplify AgentProfile resolution: defaults-then-override + collapse the skills model (one root fix for #3967/#4014/#4016/#3979/#3959)


simonrosenberg:
### Summary

Conversations launched from an **AgentProfile** (`agent_profile_id`) keep coming up subtly wrong — no exec tools (#3967), no public/user skills (the `skill_refs` default), no project skills (#4016), no streaming / `on_token` crash (#4014). These share **one root**: the profile resolver (`_build_openhands_settings` in `openhands-sdk/openhands/sdk/profiles/resolver.py`) **constructs a minimal agent from scratch** and omits defaults the inline `agent_settings` launch path carries. Each omission gets rediscovered later as its own "gap."

This issue proposes fixing the shared root — profile resolution should establish the same baseline a default `agent_settings` launch produces, with profile fields as overrides — and, in the same pass, collapse the redundant, secret-bearing skills model that is the largest single source of surrounding complexity.

**Two important scoping notes** (thanks to a code-grounded review of the first draft):

- **This is a resolver-scoped fix, not a "fix the class" fix.** OpenHands/OpenHands#3967 was fixable at the shared `create_agent()` chokepoint because `OpenHandsAgentSettings.tools: list[Tool] | None = None` has a `None` sentinel to detect "caller didn't opt in." The skill flags have no such sentinel — `agent_context: AgentContext = Field(default_factory=AgentContext)` is *always* constructed (settings/model.py:1064), and there are ~17 non-test `AgentContext(...)` sites (examples, `subagent/registry`, `workspace/remote/base`, and agent-server's own `agent.agent_context or AgentContext()` fallbacks). Flipping `AgentContext`'s class-level `load_*` defaults `False→True` would silently make all of them network-fetch/non-deterministic. So the fix is **resolver-local** (patch the resolver's `AgentContext` construction), plus one agent-server-local change for streaming (below) — explicitly *not* a class-default change. The tools fix is a conceptual precedent for "default it in resolution," not a structural template.

[...]

Issue #4014, opened 2026-07-07
Profile-launched conversations never enable LLM streaming — on_token crash on switch_llm


simonrosenberg:
### Summary

A conversation started from an **AgentProfile** (`agent_profile_id`) instead of an inline `agent_settings` payload never gets LLM token streaming wired up — and a later `switch_llm` call that assumes streaming is already enabled (a legitimate, pre-existing behavior on the `agent_settings` path) crashes the run with:

```
ValueError: Streaming requires an on_token callback
```

This is the same profile-launch path exercised by agent-canvas#1571, and the same category of gap as #3967 (exec tools / public skills silently dropped on `agent_profile_id` launches) — this time for streaming.

### Root cause (traced)

- `LocalConversation.__init__` decides **once**, at conversation construction, whether to wire the `on_token` streaming callback, via `streaming_enabled = isinstance(agent, ACPAgent) or any(llm.stream for llm in agent.get_all_llms())` (`openhands-agent-server/openhands/agent_server/event_service.py`). This is **never re-evaluated** afterward — confirmed by reading `LocalConversation.switch_llm` (`openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py`): it swaps `self.agent`'s LLM via `model_copy(update={"llm": new_llm, ...})` but never touches `self._on_token`.
- On the legacy `agent_settings` path, a consuming client can force `stream=True` on the *inline* LLM config it sends at conversation-start (agent-canvas does this — see below), so `streaming_enabled` is correctly `True` from construction.
- On the `agent_profile_id` path, `_build_openhands_settings` (`openhands-sdk/openhands/sdk/profiles/resolver.py`) does `"llm": llm` — the loaded LLM profile object **as-is**, with **no forced `stream=True`** override. So `streaming_enabled` depends entirely on whatever `stream` value happens to be persisted on the referenced LLM profile (SDK default: `False`).
- Consequence: any profile-launched conversation whose LLM profile was never explicitly saved with `stream: true` gets `on_token=None` for its **entire lifetime**. The first turn's completion is fine (no streaming requested), but a client that later calls `switch_llm` with `stream: true` (assuming parity with how conversation-start behaves) crashes the very next completion, since `on_token` can never be (re-)wired post-construction.

[...]

PR #4018, opened 2026-07-07
fix(profiles): deny-list skill selection (disabled_skills) + drop embedded skills

### LINKED ISSUE #4016: Profile-launched conversations never load project skills (load_project_skills stuck False in resolver)
### Summary

A conversation started from an **AgentProfile** (`agent_profile_id`) instead of an inline `agent_settings` payload **never loads project skills** — the workspace's `.agents/skills/`, `.openhands/skills/`, and `AGENTS.md`/third-party skill files are silently ignored. The agent can't activate any skill that lives in the repo it's working on.

This is deterministic (not flaky): the profile-resolution path never enables `load_project_skills`, so it is off for 100% of profile-launched conversations. It affects **local agent-server, cloud, and eval** identically, since all three build the agent purely from the stored profile fields. Same category as #3967 (exec tools / public skills dropped on the profile path) and #4014 (streaming never enabled on the profile path) — this time for project skills.

### Root cause (traced)

- `AgentContext.load_project_skills` (`openhands-sdk/openhands/sdk/context/agent_context.py:126`) defaults to **`False`**, and — unlike `load_user_skills` / `load_public_skills`, which `AgentContext` resolves itself — it is resolved lazily by `LocalConversation` on the first `send_message()` / `run()`, gated on this flag:

  ```python
  # local_conversation.py:833
  if merged_context is not None and merged_context.load_project_skills:
      project_skills = load_available_skills(
          work_dir=self.workspace.working_dir,
          include_user=False, include_project=True, include_public=False,
      )
      # merge_skills_by_name(project_skills, merged_context.skills) — authoritative
  ```

- So project-skill loading happens **only when `AgentContext.load_project_skills is True`**. Nothing in the SDK ever sets it to `True` programmatically — it's read-only in the codebase. A consuming client has to construct the `AgentContext` with it enabled.

- On the legacy `agent_settings` path, the client does exactly that (agent-canvas's `buildAgentContext()` sets `load_project_skills: true`). But `_build_openhands_settings` (`openhands-sdk/openhands/sdk/profiles/resolver.py:251`), which builds the settings for **every `agent_profile_id` launch**, constructs the context **without it**:

  ```python
  "agent_context": AgentContext(
      skills=skills,
      system_message_suffix=profile.system_message_suffix,
      # load_project_skills defaults to False → project skills never load
  ),
  ```

Agent-server resolution entry point: `openhands-agent-server/openhands/agent_server/conversation_service.py:_resolve_agent_from_profile` → `resolve_agent_profile(...)` → `_build_openhands_settings(...)`.

### Why the API-level skill catalog masks this

`load_all_skills()` / `discover_profile_skills()` in `openhands-agent-server/.../skills_service.py` (which back the `/api/skills` browser endpoint) **do** discover project skills independently — so a project skill shows up in the UI catalog and in a raw `GET /api/skills`, while the agent's own per-turn `get_user_message_suffix()` matching (which stamps `MessageEvent.activated_skills`) never sees it. Confirmed in a local agent-server debug run: `Loaded 1 project skills: ['e2e-test-project-skill']` appears from the API path, but the agent's message pipeline emits `activated_skills=[]` for the same conversation because its `AgentContext.load_project_skills` was `False`.

### Proposed fix

- `_build_openhands_settings` should construct the resolved `AgentContext` with `load_project_skills=True` (matching what `agent_settings`-launching clients already set), so profile-launched conversations pick up workspace/repo skills the same way. Consider also defaulting `load_user_skills` / `load_public_skills` consistently with the non-profile path, and/or exposing these as profile-model fields if per-profile control is wanted — but the immediate regression is project skills being unconditionally off.
- More broadly (cross-cutting with #3967 / #4014): the profile resolver keeps dropping `AgentContext`/settings enrichments that the `agent_settings` path carries. A single audit of "what does a normal `agent_settings` launch set on `AgentContext` that `_build_openhands_settings` doesn't" would catch these as a class rather than one at a time.

### Impact

Real functional regression, not just a test artifact: a user who launches from an agent profile (the default home-composed launch path in agent-canvas#1571) loses all project/repo skill support — `.agents/skills/`, `.openhands/skills/`, `AGENTS.md`. No client-side workaround is possible: `load_project_skills` is neither a stored-profile field nor a request-payload field — it's hardcoded (absent) inside the resolver, so only an SDK fix can restore it. (Contrast #4014, where canvas could self-heal a persisted profile field.)

### Context / tracking

- Epic: #3713 (Agent Profiles)
- Sibling gaps on the same profile-launch path: #3967 (exec tools / public skills), #4014 (streaming / on_token)
- Surfaced by: OpenHands/agent-canvas#1571's `mock-llm-skills.spec.ts` "project skill in workspace/.agents/skills/ triggers on matching keyword" — fails deterministically on the profile-launch path

simonrosenberg:
HUMAN:

Reviewed SDK issue #4017 against the code and implemented it, then — after an architectural review found skills were being modeled like MCP servers when they structurally can't be — pivoted this PR to the deny-list design (Option A from the discussion in the comments). The full reasoning is in the architecture comment and the Option-A pivot note below. Both consumer PRs (agent-canvas#1571, OpenHands#15060) are updated to match.

---

AGENT:

## Why

Conversations launched from an **AgentProfile** kept coming up subtly wrong — no project skills (#4016), no streaming / an `on_token` crash (#4014), and a recurring class of **launch-breaking `DanglingSkillRef` → 422**. These share one root: skills were modeled like MCP servers — `skill_refs`, a name **allow-list** over a "catalog." But a skill catalog is discovered from many incomplete, drifting sources (user / public / org / project / marketplace) and is recomputed per launch, so an allow-list dangles the moment the catalog a profile was authored against differs from the one resolved at launch. Three successive regressions (most recently the seed's freeze-by-name) were all facets of that one mismatch.

Separately, the profile carried an embedded `skills: list[Skill]` whose `mcp_tools` can hold a secret — the only thing that forced cipher / mask / restore machinery through the resolver, the store, and both routers.

## Summary

**Skills → a `disabled_skills` deny-list** (the model the product already runs at the member level — canvas `settings.disabled_skills`, cloud `user.disabled_skills`, migrations 102/104):
- `OpenHandsAgentProfile.skill_refs` → `disabled_skills: list[str] = []`. A deny-list over the discovered catalog: `[]` = all skills; a listed name absent from the catalog is a no-op, so a profile **can never fail launch on catalog drift**. `ACPAgentProfile` drops the skill field entirely (the subprocess owns its context).
- `AgentContext.disabled_skills`, applied in `_load_auto_skills` and the `LocalConversation` project-skill merge — one drift-tolerant chokepoint honored by every skill source (explicit, auto-loaded, lazy project).
- Deleted `_compute_skill_filter`, `DanglingSkillRef`, and `discover_profile_skills_if_needed` (the `None`/`[]`/freeze tri-state); added `_apply_disabled_skills` (catalog minus disabled, never raises). Dry-run diagnostics report `disabled_skills` / `resolved_skills`.
- The migration **seed sets `disabled_skills=[]`** — the freeze-by-name launch break is structurally gone (regression test `test_seed_then_resolve_with_narrower_catalog_does_not_dangle`).

**Drop embedded skills + the secret machinery:**
- Removed `skills: list[Skill]` from the profile. It now carries only references + the `disabled_skills` deny-list, so it is genuinely secret-free at rest. Removed the `cipher` parameter from `AgentProfileStore.save`/`load` and the `X-Expose-Secrets` handling from the local `agent_profiles_router`.

**Resolver defaults (closes #4016; profile-launch side of #4014):**
- `load_project_skills=True` on the resolved `AgentContext` (OpenHands *and* ACP) so profile-launched conversations get repo-scoped project skills.
- `conversation_service` forces `llm.stream=True` for OpenHands profile launches — the agent-server layer, which guarantees the `on_token` wiring, not the SDK resolver (which runs for headless callers too).
- **`LLM` now gracefully degrades streaming without an `on_token` callback** instead of hard-raising `ValueError` across all four entrypoints (`completion` / `acompletion` / `responses` / `aresponses`). This hardens the class so a stream/callback mismatch can never crash a run — covering the `switch_llm` side of #4014, where `on_token` is frozen at `LocalConversation` construction and never re-evaluated. Subscription mode stays exempt (it streams internally without an `on_token`).

**Schema / API:** the profile model has never shipped (no prod, no client, no persisted data), so this collapses the pre-ship schema history into a clean **v1 baseline** — `AGENT_PROFILE_SCHEMA_VERSION = 1`, no migrators (the versioning scaffold stays as the home for the first *real* post-ship change). A stray removed `skills` / `skill_refs` key is now rejected by `extra="forbid"` rather than migrated away. `safe_validation_error_detail` returns `msg` (no secret-bearing field remains to redact).

Net: a deletion overall — the deny-list and the clean-baseline collapse remove far more than they add.

### REST API contract changes

Compared with base OpenAPI `bd8a846c894f` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -684,0 +685 @@
+schema AgentContext-Input property disabled_skills optional schema=type="array" items=type="string"
@@ -695,0 +697 @@
+schema AgentContext-Output property disabled_skills optional schema=type="array" items=type="string"
@@ -751 +753 @@
-schema AgentProfileDiagnostics property dangling_skill_refs optional schema=type="array" items=type="string"
+schema AgentProfileDiagnostics property disabled_skills optional schema=type="array" items=type="string"
@@ -761 +762,0 @@
-schema AgentProfileDiagnostics property skill_refs optional schema=anyOf=[type="array" items=type="string",type="null"]
```

## Where this differs from #4017

#4017 proposed keeping `skill_refs` as an allow-list with the default flipped to `None`. That doesn't fix the root problem — an allow-list over a *discovered* catalog still dangles. The deny-list does (a name that isn't there is a no-op) and it makes the profile speak the same skill-selection language the product already uses at the global level.

## Downstream (both updated; land after this)

- **agent-canvas#1571** — already deny-list-native (user-level `disabled_skills` UI + a generic profile merge that carries the field), so only stale comments needed fixing. Per-profile skill picker stays out of scope for the minimal editor.
- **OpenHands#15060** — unions the launched profile's `disabled_skills` with the member's (`effective_disabled_skills`) into the existing deny filter. Needs this SDK released and the `openhands-sdk` pin bumped before its CI is green.

## Known follow-ups (not blocking)

- The lazy `LocalConversation` project-skill deny path is exercised via the eager and resolver paths but lacks a dedicated runtime test.
- ts-client / OpenAPI regen for the changed `AgentProfileDiagnostics` shape (`disabled_skills` in, `skill_refs` / `dangling_skill_refs` out) at the consumer pin bump.

## How to Test

```
uv run pytest tests/sdk/profiles/ \
  tests/agent_server/test_agent_profile_conv_start.py \
  tests/agent_server/test_agent_profiles_router.py \
  tests/agent_server/test_skills_service.py \
  tests/sdk/context/test_agent_context.py
```

308 pass, including `test_seed_then_resolve_with_narrower_catalog_does_not_dangle` (proves the launch break is fixed) and the deny-list no-op-on-missing-name tests. `ruff check` / `ruff format --check` / `pyright` clean.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:780de42-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-780de42-python \
  ghcr.io/openhands/agent-server:780de42-python
```

```
ghcr.io/openhands/agent-server:780de42-golang-amd64
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-golang-amd64
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-golang-amd64
ghcr.io/openhands/agent-server:780de42-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:780de42-golang-arm64
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-golang-arm64
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-golang-arm64
ghcr.io/openhands/agent-server:780de42-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:780de42-java-amd64
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-java-amd64
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-java-amd64
ghcr.io/openhands/agent-server:780de42-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:780de42-java-arm64
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-java-arm64
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-java-arm64
ghcr.io/openhands/agent-server:780de42-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:780de42-python-amd64
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-python-amd64
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-python-amd64
ghcr.io/openhands/agent-server:780de42-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:780de42-python-arm64
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-python-arm64
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-python-arm64
ghcr.io/openhands/agent-server:780de42-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:780de42-golang
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-golang
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-golang
ghcr.io/openhands/agent-server:780de42-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:780de42-java
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-java
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-java
ghcr.io/openhands/agent-server:780de42-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:780de42-python
ghcr.io/openhands/agent-server:780de42de18ff960dc9d409a041473d9f2bc1134-python
ghcr.io/openhands/agent-server:simplify-agent-profile-resolution-python
ghcr.io/openhands/agent-server:780de42-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `780de42-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `780de42-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -316,18 +316,11 @@ def _resolve_agent_from_profile(
         raise ValueError(f"Profile '{profile_name}' failed to resolve: {exc}") from exc
 
     if isinstance(settings_config, OpenHandsAgentSettings):
-        # Guarantee the on_token wiring this launch path needs (#4014): unlike
-        # an inline agent_settings launch, a client can't set llm.stream on a
-        # profile's referenced LLM ahead of time. This layer — not the SDK
-        # resolver, which runs for every caller including headless/scripted
-        # ones — is where the guarantee holds: this agent-server always wires
-        # _token_streaming_callback based on streaming_enabled (event_service.py,
-        # `any(llm.stream for llm in agent.get_all_llms())`), so setting it here
-        # is never a stream/callback mismatch. Forcing it any earlier (e.g. in
-        # the SDK resolver) would risk crashing a caller that resolves a
-        # profile without ever wiring on_token — acompletion now degrades
-        # gracefully in that case too (llm.py), but this is the layer that can
-        # make streaming actually happen instead of just failing safe.
+        # Force streaming so this launch path wires on_token: a client can't set
+        # llm.stream on a profile's referenced LLM ahead of time. Safe at this
+        # layer (not the SDK resolver) because this server wires the token
+        # callback whenever any llm.stream is set; a headless resolver caller
+        # that never wires on_token is covered by LLM's graceful degradation.
         settings_config = settings_config.model_copy(
             update={"llm": settings_config.llm.model_copy(update={"stream": True})}
         )
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -32,7 +32,7 @@ from openhands.agent_server.models import (
 )
 from openhands.agent_server.pub_sub import Subscriber
 from openhands.agent_server.server_details_router import update_last_execution_time
-from openhands.agent_server.skills_service import discover_profile_skills_if_needed
+from openhands.agent_server.skills_service import discover_profile_skills
 from openhands.agent_server.utils import safe_rmtree, utc_now
 from openhands.sdk import LLM, AgentContext, Event, Message
 from openhands.sdk.agent.base import AgentBase
@@ -267,7 +267,6 @@ def _resolve_agent_from_profile(
     Raises:
         ProfileNotFound: No stored profile has ``profile_id``.
         DanglingMcpServerRef: A referenced MCP server is absent from the global config.
-        DanglingSkillRef: A referenced skill is absent from the discovered catalog.
         ValueError: Profile load or settings validation failure.
     """
     from openhands.agent_server.persistence.store import (
@@ -293,15 +292,18 @@ def _resolve_agent_from_profile(
             f"Failed to load agent profile '{profile_name}': {exc}"
         ) from exc
 
-    # Both variants honor ``skill_refs``; the helper skips discovery when it
-    # selects none. A genuine discovery failure fails the launch loudly rather
-    # than silently producing a zero-skill agent.
-    try:
-        available_skills = discover_profile_skills_if_needed(profile)
-    except Exception as exc:
-        raise ValueError(
-            f"Skill discovery failed for profile '{profile_name}': {exc}"
-        ) from exc
+    # OpenHands profiles get the discovered catalog minus their ``disabled_skills``
+    # deny-list; ACP profiles carry no user/public skills so discovery is skipped.
+    # A genuine discovery failure fails the launch loudly rather than silently
+    # producing a zero-skill agent.
+    available_skills = None
+    if profile.agent_kind == "openhands":
+        try:
+            available_skills = discover_profile_skills()
+        except Exception as exc:
+            raise ValueError(
+                f"Skill discovery failed for profile '{profile_name}': {exc}"
+            ) from exc
 
     llm_store = get_llm_profile_store()
     try:
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -275,6 +275,7 @@ def _resolve_agent_from_profile(
         get_llm_profile_store,
     )
     from openhands.sdk.profiles.resolver import ProfileNotFound, resolve_agent_profile
+    from openhands.sdk.settings.model import OpenHandsAgentSettings
 
     store = get_agent_profile_store()
     profile_name = store.name_for_id(profile_id)
@@ -282,7 +283,7 @@ def _resolve_agent_from_profile(
         raise ProfileNotFound(f"Agent profile with id '{profile_id}' not found")
 
     try:
-        profile = store.load(profile_name, cipher=cipher)
+        profile = store.load(profile_name)
     except FileNotFoundError:
         raise ProfileNotFound(
             f"Agent profile '{profile_name}' (id={profile_id}) not found"
@@ -313,6 +314,23 @@ def _resolve_agent_from_profile(
         )
     except (TypeError, ValueError) as exc:
         raise ValueError(f"Profile '{profile_name}' failed to resolve: {exc}") from exc
+
+    if isinstance(settings_config, OpenHandsAgentSettings):
+        # Guarantee the on_token wiring this launch path needs (#4014): unlike
+        # an inline agent_settings launch, a client can't set llm.stream on a
+        # profile's referenced LLM ahead of time. This layer — not the SDK
+        # resolver, which runs for every caller including headless/scripted
+        # ones — is where the guarantee holds: this agent-server always wires
+        # _token_streaming_callback based on streaming_enabled (event_service.py,
+        # `any(llm.stream for llm in agent.get_all_llms())`), so setting it here
+        # is never a stream/callback mismatch. Forcing it any earlier (e.g. in
+        # the SDK resolver) would risk crashing a caller that resolves a
+        # profile without ever wiring on_token — acompletion now degrades
+        # gracefully in that case too (llm.py), but this is the layer that can
+        # make streaming actually happen instead of just failing safe.
+        settings_config = settings_config.model_copy(
+            update={"llm": settings_config.llm.model_copy(update={"stream": True})}
+        )
 
     agent = settings_config.create_agent()
     # Browser is deliberately absent from the deterministic SDK default
--- a/openhands-sdk/openhands/sdk/profiles/agent_profile_store.py
+++ b/openhands-sdk/openhands/sdk/profiles/agent_profile_store.py
@@ -50,8 +50,9 @@ class AgentProfileStore:
     JSON file per profile under ``~/.openhands/agent-profiles``, the filename is
     the (renameable) profile ``name``, and the stable ``id`` (uuid) lives inside
     the file. The profile is secret-free at rest — every field is a reference
-    (``llm_profile_ref``, ``mcp_server_refs``, ``skill_refs``) or a plain value,
-    so no cipher/encryption is needed to persist or load one (#4017).
+    (``llm_profile_ref``, ``mcp_server_refs``), a deny-list of names
+    (``disabled_skills``), or a plain value, so no cipher/encryption is needed to
+    persist or load one (#4017).
     """
 
     def __init__(self, base_dir: Path | str | None = None) -> None:
--- a/openhands-sdk/openhands/sdk/profiles/agent_profile_store.py
+++ b/openhands-sdk/openhands/sdk/profiles/agent_profile_store.py
@@ -24,7 +24,6 @@ if TYPE_CHECKING:
         ACPAgentProfile,
         OpenHandsAgentProfile,
     )
-    from openhands.sdk.utils.cipher import Cipher
 
 _DEFAULT_PROFILE_DIR: Final[Path] = Path.home() / ".openhands" / "agent-profiles"
 _LOCK_TIMEOUT_SECONDS: Final[float] = 30.0
@@ -50,9 +49,9 @@ class AgentProfileStore:
     Mirrors :class:`~openhands.sdk.llm.llm_profile_store.LLMProfileStore`: one
     JSON file per profile under ``~/.openhands/agent-profiles``, the filename is
     the (renameable) profile ``name``, and the stable ``id`` (uuid) lives inside
-    the file. The profile is secret-free at rest except for
-    ``skills[].mcp_tools`` env/headers, which are masked by default and
-    encrypted when a ``cipher`` is supplied.
+    the file. The profile is secret-free at rest — every field is a reference
+    (``llm_profile_ref``, ``mcp_server_refs``, ``skill_refs``) or a plain value,
+    so no cipher/encryption is needed to persist or load one (#4017).
     """
 
     def __init__(self, base_dir: Path | str | None = None) -> None:
@@ -136,17 +135,14 @@ class AgentProfileStore:
         self,
         profile: OpenHandsAgentProfile | ACPAgentProfile,
         *,
-        cipher: Cipher | None = None,
         max_profiles: int | None = None,
     ) -> None:
         """Save a profile under its own ``name``, overwriting any namesake.
 
-        With no ``cipher`` the profile is secret-free at rest:
-        ``skills[].mcp_tools`` env/headers are redacted. With a ``cipher`` those
-        values are encrypted (recoverable) rather than redacted; every other
-        field is a reference and dumps in the clear regardless. When
-        ``max_profiles`` is set, creating a *new* profile beyond the cap raises
-        ``ProfileLimitExceeded`` under the same lock as the write.
+        The profile is secret-free at rest, so it dumps in the clear with no
+        cipher/encryption needed. When ``max_profiles`` is set, creating a
+        *new* profile beyond the cap raises ``ProfileLimitExceeded`` under the
+        same lock as the write.
 
         Raises:
             ValueError: If ``profile.name`` is not a valid profile name.
@@ -154,12 +150,7 @@ class AgentProfileStore:
             TimeoutError: If the lock cannot be acquired.
         """
         profile_path = self._get_profile_path(profile.name)
-
-        # Cipher present => encrypt mcp_tools secrets; absent => redact them.
-        context: dict[str, Any] = (
-            {"cipher": cipher, "expose_secrets": "encrypted"} if cipher else {}
-        )
-        payload = profile.model_dump(mode="json", context=context)
+        payload = profile.model_dump(mode="json")
 
         with self._acquire_lock():
             if max_profiles is not None and not profile_path.exists():
@@ -186,15 +177,11 @@ class AgentProfileStore:
     def load(
         self,
         name: str,
-        *,
-        cipher: Cipher | None = None,
     ) -> OpenHandsAgentProfile | ACPAgentProfile:
         """Load and validate the profile stored under ``name``.
 
-        A ``cipher`` is threaded through the validation context for parity with
-        ``save``, so encrypted ``skills[].mcp_tools`` secrets are restored as
-        ``SecretStr`` values. Reference fields (``llm_profile_ref`` etc.) carry
-        no secret and load unchanged.
+        All fields are references or plain values, so no cipher is needed to
+        load one.
 
         Raises:
             FileNotFoundError: If ``name`` does not exist.
@@ -213,8 +200,7 @@ class AgentProfileStore:
 
             try:
                 data = json.loads(profile_path.read_text())
-                context = {"cipher": cipher} if cipher else None
-                profile = validate_agent_profile(data, context=context)
+                profile = validate_agent_profile(data)
             except Exception as e:
                 raise ValueError(f"Failed to load profile `{name}`: {e}") from e
 
@@ -244,8 +230,7 @@ class AgentProfileStore:
         """Atomically rename a profile, keeping the in-file ``name`` in sync.
 
         Unlike a bare file move this also rewrites the persisted ``name`` field
-        (a surgical raw-JSON edit, so encrypted ``mcp_tools`` are untouched and
-        no cipher is needed). The stable ``id`` is preserved.
+        (a surgical raw-JSON edit). The stable ``id`` is preserved.
 
         Raises:
             FileNotFoundError: If ``old_name`` is missing.
@@ -279,9 +264,9 @@ class AgentProfileStore:
         The single-profile write primitive behind ``profile_refs.cascade_rename``.
         Self-locks (re-entrant), so the read-modify-write is atomic whether called
         standalone or nested under the FK helpers' :meth:`lock`. A raw-JSON edit
-        (only the ref field changes), so encrypted ``skills[].mcp_tools`` and the
-        stable ``id`` are untouched and no cipher is needed. No-op when the profile
-        is missing, non-dict, or an ACP profile (which carries no ref).
+        (only the ref field changes), so the stable ``id`` is untouched. No-op
+        when the profile is missing, non-dict, or an ACP profile (which carries
+        no ref).
         """
         path = self._get_profile_path(name)
         with self._acquire_lock():
@@ -380,13 +365,10 @@ class AgentProfileStoreProtocol(Protocol):
         self,
         profile: OpenHandsAgentProfile | ACPAgentProfile,
         *,
-        cipher: Cipher | None = ...,
         max_profiles: int | None = ...,
     ) -> None: ...
 
-    def load(
-        self, name: str, *, cipher: Cipher | None = ...
-    ) -> OpenHandsAgentProfile | ACPAgentProfile: ...
+    def load(self, name: str) -> OpenHandsAgentProfile | ACPAgentProfile: ...
 
     def delete(self, name: str) -> None: ...
 
@@ -422,7 +404,6 @@ def save_profile_preserving_identity(
     store: AgentProfileStoreProtocol,
     profile: OpenHandsAgentProfile | ACPAgentProfile,
     *,
-    cipher: Cipher | None = None,
     max_profiles: int | None = None,
 ) -> OpenHandsAgentProfile | ACPAgentProfile:
     """Save ``profile`` with the server-managed id/revision policy.
@@ -445,5 +426,5 @@ def save_profile_preserving_identity(
             )
         else:
             profile = profile.model_copy(update={"id": uuid4()})
-        store.save(profile, cipher=cipher, max_profiles=max_profiles)
+        store.save(profile, max_profiles=max_profiles)
     return profile
--- a/openhands-agent-server/openhands/agent_server/agent_profiles_router.py
+++ b/openhands-agent-server/openhands/agent_server/agent_profiles_router.py
@@ -27,7 +27,7 @@ from openhands.agent_server.persistence import (
     get_settings_store,
 )
 from openhands.agent_server.profiles_router import MAX_PROFILES
-from openhands.agent_server.skills_service import discover_profile_skills_if_needed
+from openhands.agent_server.skills_service import discover_profile_skills
 from openhands.sdk.llm import LLM
 from openhands.sdk.llm.llm_profile_store import (
     ProfileLimitExceeded as LLMProfileLimitExceeded,
@@ -491,20 +491,20 @@ async def materialize_agent_profile(
     settings = get_settings_store(config).load() or PersistedSettings()
     mcp_config = settings.agent_settings.mcp_config
 
-    # Discover skills off the event loop so the dry-run can report which
-    # ``skill_refs`` resolve vs. dangle. A discovery failure must not 500 the
-    # preview: pass ``available_skills=None`` (catalog unknown — the dry-run then
-    # reports nothing dangling, rather than flagging every selected skill) and
-    # surface the failure as its own diagnostic below.
+    # Discover skills off the event loop so the dry-run can report which skills
+    # (catalog minus ``disabled_skills``) resolve. Only OpenHands profiles carry
+    # user/public skills; ACP profiles do not. A discovery failure must not 500
+    # the preview: pass ``available_skills=None`` and surface the failure as its
+    # own diagnostic below.
     discovery_error: str | None = None
-    try:
-        available_skills = await asyncio.to_thread(
-            discover_profile_skills_if_needed, profile
-        )
-    except Exception as exc:
-        available_skills = None
-        discovery_error = str(exc)
-        logger.warning("Skill discovery failed during materialize: %s", exc)
+    available_skills = None
+    if profile.agent_kind == "openhands":
+        try:
+            available_skills = await asyncio.to_thread(discover_profile_skills)
+        except Exception as exc:
+            available_skills = None
+            discovery_error = str(exc)
+            logger.warning("Skill discovery failed during materialize: %s", exc)
 
     llm_store = get_llm_profile_store()
     diagnostics = resolve_agent_profile_dry_run(
--- a/openhands-agent-server/openhands/agent_server/agent_profiles_router.py
+++ b/openhands-agent-server/openhands/agent_server/agent_profiles_router.py
@@ -300,8 +300,8 @@ async def save_agent_profile(
         profile = validate_agent_profile({**body, "name": name})
     except ValidationError as e:
         # Match FastAPI's request-validation shape (``detail`` is a list of
-        # error objects), but surface only ``loc``/``type`` (see
-        # safe_validation_error_detail's docstring).
+        # error objects): ``loc``/``type``/``msg`` (``input`` dropped — see
+        # safe_validation_error_detail's docstring for why msg is now safe).
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail=safe_validation_error_detail(e),
--- a/openhands-agent-server/openhands/agent_server/agent_profiles_router.py
+++ b/openhands-agent-server/openhands/agent_server/agent_profiles_router.py
@@ -19,13 +19,7 @@ from typing import Annotated, Any
 from fastapi import APIRouter, HTTPException, Path, Request, status
 from pydantic import BaseModel, Field, ValidationError
 
-from openhands.agent_server._secrets_exposure import (
-    build_expose_context,
-    get_cipher,
-    get_config,
-    parse_expose_secrets_header,
-    translate_missing_cipher,
-)
+from openhands.agent_server._secrets_exposure import get_cipher, get_config
 from openhands.agent_server.persistence import (
     PersistedSettings,
     get_agent_profile_store,
@@ -213,7 +207,7 @@ def _seed_default_profile(
                 settings.agent_settings.llm, cipher
             )
         profile = build_seed_profile(settings.agent_settings, active_llm_profile)
-        store.save(profile, cipher=cipher, max_profiles=MAX_AGENT_PROFILES)
+        store.save(profile, max_profiles=MAX_AGENT_PROFILES)
 
         profile_id = str(profile.id)
         settings_store = get_settings_store(get_config(request))
@@ -266,31 +260,24 @@ async def list_agent_profiles(request: Request) -> AgentProfileListResponse:
 
 
 @agent_profiles_router.get("/{name}", response_model=AgentProfileDetailResponse)
-async def get_agent_profile(
-    request: Request, name: ProfileName
-) -> AgentProfileDetailResponse:
+async def get_agent_profile(name: ProfileName) -> AgentProfileDetailResponse:
     """Get a stored profile.
 
-    Use the ``X-Expose-Secrets`` header to control ``skills[].mcp_tools`` secret
-    exposure (``encrypted`` / ``plaintext``); absent, those values are masked.
+    A profile is secret-free at rest (#4017), so there is nothing to mask or
+    expose — unlike the LLM ``/api/profiles`` router, this endpoint has no
+    ``X-Expose-Secrets`` behavior.
     """
-    expose_mode = parse_expose_secrets_header(request)
-    cipher = get_cipher(request)
-
     store = get_agent_profile_store()
     try:
         with _store_errors():
-            profile = store.load(name, cipher=cipher)
+            profile = store.load(name)
     except FileNotFoundError:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
             detail=f"Agent profile '{name}' not found",
         )
 
-    context = build_expose_context(expose_mode, cipher)
-    with translate_missing_cipher():
-        payload = profile.model_dump(mode="json", context=context)
-
+    payload = profile.model_dump(mode="json")
     return AgentProfileDetailResponse(name=name, profile=payload)
 
 
@@ -300,31 +287,29 @@ async def get_agent_profile(
     status_code=status.HTTP_201_CREATED,
 )
 async def save_agent_profile(
-    request: Request, name: ProfileName, body: dict[str, Any]
+    name: ProfileName, body: dict[str, Any]
 ) -> AgentProfileMutationResponse:
     """Save an ``AgentProfile`` under ``name`` (overwriting a namesake).
 
     The path ``name`` is authoritative — it overrides any ``name`` in the body.
-    With ``OH_SECRET_KEY`` configured, ``skills[].mcp_tools`` secrets are
-    encrypted at rest; otherwise they are redacted. Returns 409 if creating a
-    new profile would exceed ``MAX_AGENT_PROFILES``.
+    The profile is secret-free at rest (#4017), so no cipher/encryption is
+    involved. Returns 409 if creating a new profile would exceed
+    ``MAX_AGENT_PROFILES``.
     """
-    cipher = get_cipher(request)
     try:
         profile = validate_agent_profile({**body, "name": name})
     except ValidationError as e:
-        # Match FastAPI's request-validation shape (``detail`` is a list of error
-        # objects), but surface only ``loc``/``type`` — a nested mcp_tools
-        # FastMCP config error embeds the input (which may carry secrets) in
-        # ``msg``.
+        # Match FastAPI's request-validation shape (``detail`` is a list of
+        # error objects), but surface only ``loc``/``type`` (see
+        # safe_validation_error_detail's docstring).
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail=safe_validation_error_detail(e),
         )
     except Exception:
-        # Any other validation failure (e.g. SkillValidationError from a
-        # malformed mcp_tools, or a schema/migration error) is a client error,
-        # never a 500. Stay generic — these messages can embed the input.
+        # Any other validation failure (e.g. a schema/migration error) is a
+        # client error, never a 500. Stay generic — these messages can embed
+        # the input.
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail="Invalid agent profile",
@@ -339,7 +324,7 @@ async def save_agent_profile(
     try:
         with _store_errors():
             save_profile_preserving_identity(
-                store, profile, cipher=cipher, max_profiles=MAX_AGENT_PROFILES
+                store, profile, max_profiles=MAX_AGENT_PROFILES
             )
     except ProfileLimitExceeded:
         raise HTTPException(
@@ -489,18 +474,19 @@ async def materialize_agent_profile(
     than raising — the only error status is 404 (unknown profile name).
     resolved_settings is redacted (api_key_set booleans; no raw secrets).
     """
-    cipher = get_cipher(request)
-
     store = get_agent_profile_store()
     try:
         with _store_errors():
-            profile = store.load(name, cipher=cipher)
+            profile = store.load(name)
     except FileNotFoundError:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
             detail=f"Agent profile '{name}' not found",
         )
 
+    # Still needed here (unlike the profile load above): resolve_agent_profile_
+    # dry_run uses it to decrypt the *referenced LLM profile's* own secret.
+    cipher = get_cipher(request)
     config = get_config(request)
     settings = get_settings_store(config).load() or PersistedSettings()
     mcp_config = settings.agent_settings.mcp_config
--- a/openhands-agent-server/openhands/agent_server/conversation_router.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_router.py
@@ -51,7 +51,6 @@ from openhands.sdk.marketplace.registry import (
 from openhands.sdk.plugin import PluginFetchError
 from openhands.sdk.profiles.resolver import (
     DanglingMcpServerRef,
-    DanglingSkillRef,
     ProfileNotFound,
 )
 from openhands.sdk.tool.client_tool import ClientToolRegistrationError
@@ -216,11 +215,6 @@ async def start_conversation(
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail={"message": str(e), "dangling_mcp_server_refs": e.missing},
-        ) from e
-    except DanglingSkillRef as e:
-        raise HTTPException(
-            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail={"message": str(e), "dangling_skill_refs": e.missing},
         ) from e
     except ClientToolRegistrationError as e:
         raise HTTPException(
--- a/openhands-agent-server/openhands/agent_server/skills_service.py
+++ b/openhands-agent-server/openhands/agent_server/skills_service.py
@@ -21,7 +21,6 @@ import tempfile
 from dataclasses import dataclass
 from pathlib import Path
 from time import monotonic
-from typing import TYPE_CHECKING
 
 from pydantic import BaseModel, ValidationError
 
@@ -55,10 +54,6 @@ from openhands.sdk.skills.utils import (
 )
 from openhands.sdk.utils import sanitized_env
 from openhands.sdk.utils.path import to_posix_path
-
-
-if TYPE_CHECKING:
-    from openhands.sdk.profiles import ACPAgentProfile, OpenHandsAgentProfile
 
 
 logger = get_logger(__name__)
@@ -458,20 +453,21 @@ def load_all_skills(
 
 
 def discover_profile_skills() -> list[Skill]:
-    """Skill catalog for ``AgentProfile.skill_refs`` resolution (#3868).
+    """Skill catalog an OpenHands profile launches with (#4017).
 
     Returns the merged user + public skills — the deterministic sources of
-    :func:`load_all_skills` that ``resolve_agent_profile`` filters by name.
-    ``load_all_skills`` already absorbs and logs benign per-source failures, so
-    this does not swallow errors: an unexpected failure propagates rather than
-    silently resolving the profile to a zero-skill agent.
+    :func:`load_all_skills`. ``resolve_agent_profile`` keeps all of them except
+    the profile's ``disabled_skills`` deny-list. ``load_all_skills`` already
+    absorbs and logs benign per-source failures, so this does not swallow errors:
+    an unexpected failure propagates rather than silently resolving the profile
+    to a zero-skill agent.
 
     Org / project skills need auth / workspace context not available at resolve
-    time, so they are not in this catalog (a follow-up). Because the resolver
-    now hard-fails dangling ``skill_refs`` (mirroring MCP), a profile that
-    selects an org/project skill via the picker would fail launch here until the
-    catalog is broadened — the SaaS app-server, which has that context, is
-    expected to supply the fuller catalog the same way it does for the picker.
+    time, so they are not in this catalog. That is safe under the deny-list
+    model: a skill missing from the catalog is simply never disabled (it stays
+    on when it does load, e.g. project skills loaded lazily by
+    ``LocalConversation``) — the old allow-list would instead have dangled and
+    failed the launch.
     """
     return list(
         load_all_skills(
@@ -481,21 +477,6 @@ def discover_profile_skills() -> list[Skill]:
             load_project=False,
         ).skills
     )
-
-
-def discover_profile_skills_if_needed(
-    profile: "OpenHandsAgentProfile | ACPAgentProfile",
-) -> list[Skill] | None:
-    """Discover the skill catalog a profile needs, or ``None`` to skip discovery.
-
-    ``skill_refs == []`` selects no discovered skills, so the (potentially
-    network-bound) discovery is skipped and the resolver receives ``None``. Any
-    other value — ``None`` (all discovered) or a name list — needs the catalog.
-    Centralizes the skip guard shared by conversation start and the dry-run.
-    """
-    if profile.skill_refs == []:
-        return None
-    return discover_profile_skills()
 
 
 def sync_public_skills() -> tuple[bool, str]:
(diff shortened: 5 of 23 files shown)


openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -314,6 +315,23 @@ def _resolve_agent_from_profile(
     except (TypeError, ValueError) as exc:
         raise ValueError(f"Profile '{profile_name}' failed to resolve: {exc}") from exc
 
+    if isinstance(settings_config, OpenHandsAgentSettings):
+        # Guarantee the on_token wiring this launch path needs (#4014): unlike
```
VascoSch92: can we trimmer the comment?

simonrosenberg:
Two follow-ups from an xhigh review pass. Neither is a hard blocker on this PR, but both are behaviors introduced here that are worth surfacing / deciding deliberately.

### 1. ACP profiles now inject workspace project skills → possible duplication for CLIs that read `AGENTS.md` themselves

`_build_acp_settings` now *always* builds `AgentContext(load_project_skills=True)` (previously `agent_context=None` when no skills were selected). Consistent with the OpenHands path and correct for #4016 — but ACP CLIs (Claude Code, Codex, Gemini) already ingest workspace context files (`AGENTS.md`, `CLAUDE.md`, …) natively, so a workspace containing those hands the ACP subprocess the same content twice. `skill_refs` doesn't gate project skills, so there's no per-profile opt-out today, and the docstring's "behavior-preserving for the rendered ACP prompt" claim only holds for workspaces with **no** project skills.

Filed for a design call: #4019.

### 2. `skill_refs=None` (new OpenHands default) runs skill discovery on every launch *and* every `/materialize` dry-run

`discover_profile_skills_if_needed` skips discovery only when `skill_refs == []`. With the OpenHands default now `None`, every profile-launched conversation start — and every `/materialize` dry-run — performs a user + public skills catalog scan: a `~/.openhands/skills` walk plus a public `OpenHands/extensions` catalog load (potentially a git fetch), even for users who configured no skills. The previous `[]`-default seed skipped this entirely, so this adds network/IO latency to the launch hot path for the common case.

Not necessarily wrong (it's the cost of "all skills by default"), but worth deciding whether launch/materialize should block on a live fetch, or whether discovery should be cached / lazier.

_(Surfaced by a workflow-backed review; both were independently verified.)_

simonrosenberg:
## xhigh code-review findings

Workflow-backed review (6 finders → 18 verified candidates → 9 distinct findings, 0 refuted), then the top finding re-verified by hand against the code. Ranked most-severe first.

### 🔴 1. BLOCKER — the migration seed can hard-fail the default profile's launch (`seed.py:74`)

The seed freezes the global's resolved skills by name (`skill_refs=[s.name for s in context.skills]`). At launch, `discover_profile_skills()` re-resolves those names against a **narrower** catalog — `load_all_skills(load_user=True, load_public=True, load_org=False, load_project=False)` at `DEFAULT_MARKETPLACE_PATH` (`skills_service.py:476-483`). Any frozen name that doesn't reappear there hits `resolve_agent_profile`'s `raise DanglingSkillRef` (`resolver.py:344-345`) and the conversation fails to start.

`AgentContext._load_auto_skills` (`agent_context.py:219-243`) eagerly materializes user + public + **inline/explicit** skills into `context.skills`, so the seeded `default` profile bricks its own launch when the global `agent_context.skills` contains:

- an **inline / hand-authored skill** (name absent from any discoverable repo) — this is exactly the config `test_seed_freezes_resolved_skills_by_name` sets up (`agent_context.skills=[{name:'alpha'}]`); the test only asserts `skill_refs == ['alpha']` and never launches, so the break is invisible in CI;
- an **org skill** (`load_org=False` at discovery) — a real enterprise config, directly relevant to the cloud rollout (#15060);
- a **public skill under a non-default `marketplace_path`** (discovery uses the default path).

`origin/main`'s seed embedded full `Skill` objects, so launch used them directly and **could not** dangle. The freeze-by-name change traded the earlier "seed expands a partial-source global" concern for a strictly worse outcome: a hard launch failure. `discover_profile_skills`'s own docstring self-documents this (`skills_service.py:471-474`): *"a profile that selects an org/project skill via the picker would fail launch here until the catalog is broadened."*

**Fix required before merge** — see the root-cause note; this is a design choice, not a one-liner.

### 🟠 Root cause tying #1–#3 together

The resolver hard-fails dangling `skill_refs` for **MCP-parity**, but the two aren't symmetric: `mcp_config` is the complete authoritative map, whereas the **launch skill-discovery catalog is narrower than the set of skills that can legitimately name a profile** (inline, org, project, custom-marketplace). So *any* path that puts such a name into `skill_refs` — the seed-freeze (#1), a future picker, or a "fold embedded names into refs" migration (#2/#3) — detonates at launch. This contract should be decided deliberately: either **discovery covers every source a ref can come from**, or **dangling skill refs become non-fatal** (drop-with-log, unlike MCP). Note this is why the obvious fix for #2/#3 (fold `skills[].name` → `skill_refs`) is **not** safe as-is — it would recreate #1.

### 🟡 2 & 3. v1→v2 migration silently drops embedded skills (`agent_profile.py:323, 331`)

`_migrate_v1_to_v2` pops `skills` without folding names into `skill_refs`, and leaves a v1 `skill_refs=[]` untouched (`[]` = none) → zero-skill agent. Two facets:

- **Schema-less POST silently returns 201** — a body with `skills:[…]` and no `schema_version` is coerced to v1, migrated (skills dropped), and returns 201; the client believes it saved skills. Only an *explicit* current-schema body gets a 422. That asymmetry is a live-API footgun (`test_save_v1_body_with_stray_skills_key_is_migrated_and_saved` encodes it) — worth a 422 or an explicit warning.
- **Persisted v1 profiles** lose skills on load — but **no prod v1 data exists** (neither consumer PR merged), so this facet is theoretical.

Low severity given no prod data; mind the #1 interaction before "fixing" by folding names in.

### 🟢 4 & 6. Known / already tracked

- **#4** `skill_refs=None` (new OpenHands default) runs user+public discovery — incl. a public-repo fetch — on **every launch and every `/materialize` dry-run** (`discover_profile_skills_if_needed` only skips when `skill_refs == []`). Already surfaced in the follow-up comment above. It's the cost of "all skills by default"; the open question is whether to cache/lazy it, not a bug.
- **#6** ACP path forces `load_project_skills=True` → double-ingestion for CLIs that read `AGENTS.md` themselves. Already filed as **#4019** and self-caveated in the `_build_acp_settings` docstring.

### 🟢 5 & 7. Mostly intended / parity-with-main

- **#5** OpenHands `load_project_skills=True` (`resolver.py:256`) "diverges from the settings-driven launch." Per #4017's table the `agent_settings` (canvas) path *already* sets `load_project_skills=true`, so forcing it in the resolver is the intended #4016 fix and matches that baseline — the finder compared against a bare `AgentContext()` default rather than the client-enriched path. Only real sub-point: the seed can't preserve a deliberate global `load_project_skills=False` (minor; project-skills-always-on is the documented design).
- **#7** Forced `stream=True` (`conversation_service.py:324`) can override an explicit `stream=False`. Verified **PLAUSIBLE**, not confirmed. The non-profile path also forces `llm.stream=true` for OpenHands client-side, so this is faithful parity with `main`; a capability gate here would *diverge* from it. Low.

### 🔵 8 & 9. Cleanup

- **#8** `skill_refs` is re-declared on `ACPAgentProfile` (`agent_profile.py:226`) only to flip the default to `[]`, duplicating the entire Field description (base at line 121). Cleaner: put `None` on `OpenHandsAgentProfile` and `[]` on `ACPAgentProfile`, leaving no divergent copy to drift.
- **#9** `safe_validation_error_detail` (`agent_profile.py:424`) still strips `msg`/`input` from every profile-save 422, but its own docstring now admits no secret-bearing field remains — so it's pure UX degradation (clients can't see *why* a save failed). With `mcp_server_refs`/`llm_profile_ref` being by-name refs carrying no inline secret, restoring full pydantic detail is safe and strictly better.

---

### Recommendation

**#1 is the sole blocker.** Two viable directions:

1. **Revert the seed to `skill_refs=None`** (all-discovered) — it can never dangle; the "expands a partial-source global" concern that motivated the freeze is far milder than a bricked launch.
2. **Make dangling `skill_refs` non-fatal on the launch path** (drop-with-log), keeping the dry-run's dangling *report* for the editor. More general — also unblocks org/project skills and a future picker.

I lean toward (2), plus a launch test that actually starts a seeded-inline-skill profile (the current test stops at seed construction), and cleanups #8/#9 folded in.

<sub>Surfaced by a workflow-backed xhigh review; findings independently verified.</sub>

enyst:
> > The local agent-server's agent_profiles_router.py no longer threads a cipher through profile load/save or does any X-Expose-Secrets handling for the GET/POST endpoints 

This sounds like the right thing to do! Simplification ❤️ 

---

@openhands Understand this PR and its referenced/linked issues and PRs. Take your time. Clone yourself the referenced repos e.g. openhands/openhands, openhands/agent-canvas and investigate deeply.

Note: do NOT assume that the text is accurate, the issues and PRs have been written by an AI, and you know those can hallucinate: you need to verify for yourself.

Then do a /codereview-roasted on it. Post your feedback on this PR.

Then take a step back and read this comment, the first issue: the blocker, and its root cause. https://github.com/OpenHands/software-agent-sdk/pull/4018#issuecomment-4909512373 . Note: do NOT assume that the text is accurate, this comment too was written by an AI, so verify and restate as you think appropriate. And tell us WDYT about this blocker?

enyst:
Re: the blocker.

I'm not sure of all details, I do think the Dangling error doesn't need to be fatal though. Same for MCP. I think we can just give the user a warning. That's what `codex` does from what I recall, it's up to the user to continue or fix their MCP/whatever and restart. It's the friendly alternative? 😅 

As noted in the linked issue, I do think maybe we don't need to push extra/duplicate skills on the ACP applications running in the background. That way lies a lot of headaches...

openhands-sdk/openhands/sdk/profiles/agent_profile_store.py
```diff
@@ -186,15 +177,11 @@ def save(
...
-        cipher: Cipher | None = None,
     ) -> OpenHandsAgentProfile | ACPAgentProfile:
         """Load and validate the profile stored under ``name``.
 
-        A ``cipher`` is threaded through the validation context for parity with
-        ``save``, so encrypted ``skills[].mcp_tools`` secrets are restored as
-        ``SecretStr`` values. Reference fields (``llm_profile_ref`` etc.) carry
-        no secret and load unchanged.
+        All fields are references or plain values, so no cipher is needed to
+        load one.
```
enyst, line 186:
```suggestion
```

LLMs love to say what is not happening anymore 😅

enyst:
## `/codereview-roasted` — verified against the checked-out branch

I read the actual code at PR head (not just the PR body) and independently reproduced the key failure. Where the PR body and I disagree, I trust the code.

**Taste rating: 🟡 Acceptable, but one real blocker.** The core idea — inherit the same baseline a default `agent_settings` launch gets, and delete the embedded `skills` field so the profile is genuinely secret-free — is good taste. It removes a whole cipher/mask/`X-Expose-Secrets` code path from the profile router, and the code that's left is simpler and honest about it. The problem is a single-line consequence of the seed change.

### 🔴 CRITICAL — the migration seed can hard-fail a profile launch (`seed.py`)

This is finding #1 from @simonrosenberg's xhigh pass, and it holds up. I reproduced it end-to-end:

- `build_seed_profile` freezes `skill_refs=[s.name for s in context.skills]`.
- `AgentContext` materializes **inline/hand-authored** skills (and, in enterprise, org skills) into `context.skills` — the `skills` field has no source restriction, only a duplicate-name check (`agent_context.py:_validate_skills` / `_load_auto_skills`).
- At launch, `discover_profile_skills()` only loads `load_user=True, load_public=True` (`skills_service.py`). Any frozen name not in that narrower catalog hits `resolve_agent_profile`'s `raise DanglingSkillRef` — a **fatal** launch failure.

Minimal repro I ran against this branch:

```
global agent_context.skills = [Skill(name="alpha", ...)]   # inline, not in any repo
build_seed_profile(...).skill_refs == ["alpha"]
resolve_agent_profile(..., available_skills=[])  # user+public catalog lacks "alpha"
=> DanglingSkillRef(['alpha'])   # conversation fails to start
```

Scope matters and cuts both ways:
- The **default zero-skill global is safe**: `context.skills == []` → `skill_refs=[]` → `discover_profile_skills_if_needed` returns `None` → discovery skipped → no dangling. I verified this too.
- But **any global carrying an inline / org / custom-marketplace skill bricks its own seeded `default` profile.** That's a real enterprise config (org skills, `load_org=False` at discovery) and directly relevant to the #15060 cloud rollout.

Why `main` didn't have this: `main`'s seed embeds full `Skill` objects with `skill_refs=[]`, so the resolver used them directly and *could not* dangle. Trading embedded objects for frozen-names is what introduces the failure mode.

**Why CI is green anyway (testing gap):** `test_seed_freezes_resolved_skills_by_name` sets up exactly the dangerous config (`agent_context.skills=[{name:'alpha'},...]`) but only asserts `skill_refs == ['alpha']` and **never launches**. The 141 profile tests pass and the break is invisible. If you keep freeze-by-name, add a test that actually resolves/launches a seeded inline-skill profile.

### 🟠 Root cause — MCP-parity is the wrong parity for skills

The resolver hard-fails dangling `skill_refs` "to mirror MCP." But the two aren't symmetric: `mcp_config` is the *complete authoritative map*, whereas the launch skill-discovery catalog is **strictly narrower** than the set of skills that can legitimately name a profile (inline, org, project, custom-marketplace). So any path that puts such a name into `skill_refs` detonates at launch. This is a contract decision, not a one-liner — see my separate WDYT comment on the blocker.

### 🟡 v1→v2 migration silently drops embedded skills — with an API asymmetry

`_migrate_v1_to_v2` pops `skills` without folding names into `skill_refs`. Combined with the discriminator defaulting a schema-less POST to v1:
- A body with `skills:[…]` and **no** `schema_version` → coerced to v1 → migrated (skills dropped) → **201 Created**. The client thinks it saved skills; it saved none. (`test_save_v1_body_with_stray_skills_key_is_migrated_and_saved` encodes exactly this.)
- A body that *does* set the current `schema_version` with a stray `skills` → 422.

No prod v1 data exists, so the persisted-migration half is theoretical, but the **live-API footgun** (silent 201 that drops a field the client sent) is worth a 422 or at least a warning. Do **not** "fix" it by folding `skills[].name` into `skill_refs` — that recreates the blocker above.

### 🔵 Minor cleanups (non-blocking)

- **#8** `ACPAgentProfile.skill_refs` is re-declared solely to flip the default to `[]`, duplicating the base's entire Field description — two copies to drift. Cleaner: default `None` on `OpenHandsAgentProfile`, `[]` on `ACPAgentProfile`, no shared base declaration to diverge from. Small.
- **#9** `safe_validation_error_detail` still strips `msg`/`input` from every profile-save 422, but its own updated docstring now says no secret-bearing field remains. It's now pure UX degradation (clients can't see *why* a save failed). Restoring full pydantic detail is safe once `skills[].mcp_tools` is gone.

### Things the PR body claims that I could **not** confirm (FYI, not blockers)

- The named downstream shims **`withDefaultSkillRefs` and `ensureLlmProfileStreams` do not exist** anywhere in `OpenHands/agent-canvas` (code search + PR #1571 diff). The *concept* is real though — canvas#1571 has an e2e path that omits `skill_refs` and comments that it relies on this PR's unset→`null` default. So the dependency is genuine; the specific function names in the description are not.
- OpenHands#15060 is **not** "actively building ~150 lines of embedded-skill mask/restore." Its current head (`feat-agent-profiles-cloud`, updated 2026-07-07) has **already been rewritten to the secret-free model** — `enterprise/server/routes/agent_profiles.py` explicitly says "secret-free at rest, nothing to mask … no `X-Expose-Secrets` handling here at all." So that consumer already depends on this PR's design; the "building against a model this PR deletes" framing is stale.

These don't change the merits — they just mean the "downstream impact" section describes a past state.

### Things I verified as **correct**

- `AgentContext.load_project_skills` exists, defaults `False`, and is lazily resolved by `LocalConversation` (`local_conversation.py`) gated on that flag. The old OpenHands build path left it unset (=False), so profile launches really did miss project skills. ✅ genuine #4016 fix.
- `_build_acp_settings` really did return `agent_context=None` when no skills were selected; now it always builds one. ✅
- Forcing `llm.stream=True` in `_resolve_agent_from_profile` does light up `on_token`: `event_service.py` wires the token callback whenever `any(llm.stream for llm in agent.get_all_llms())`. ✅
- The "don't also flip `load_user_skills`/`load_public_skills`" argument is correct: `discover_profile_skills()` already loads user+public, so those skills arrive via `filtered_skills`; flipping the flags would double-load. ✅

### [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟡 **MEDIUM**. Mostly a simplification (deletes a secret-handling path — good), but it changes the OpenHands `skill_refs` default and the seed's skill mechanism, and it touches agent launch behavior (project skills + forced streaming) that could plausibly affect eval/benchmark runs. The seed blocker is a real launch-time regression for skill-carrying globals. Per repo review policy I'm leaving this as a **COMMENT**, not an approval — the launch-behavior + skills-default changes are eval-relevant and there's an unresolved blocker.

**VERDICT:** ❌ Needs rework — resolve the dangling-`skill_refs` contract before merge (see WDYT comment); fold in the 201-vs-422 asymmetry and a launch-level test.

**KEY INSIGHT:** Deleting embedded skills is the right call, but `skill_refs` inherited MCP's *hard-fail-on-dangling* contract without MCP's *complete-catalog* guarantee — so the seed's freeze-by-name can name a skill launch-discovery will never find.

---
_This review was generated by an AI agent (OpenHands) on behalf of @enyst._

enyst:
## WDYT on the blocker (re: [#issuecomment-4909512373](https://github.com/OpenHands/software-agent-sdk/pull/4018#issuecomment-4909512373))

I verified the blocker instead of taking the comment at its word, and I agree with @enyst's instinct: **the fatal `DanglingSkillRef` is the actual problem, not the seed.**

### Is the blocker real? Yes — reproduced end-to-end.

The described mechanism checks out against the code and I reproduced it on this branch:

- `AgentContext.skills` can hold **inline / hand-authored** skills (and org skills in enterprise) — the field only rejects duplicate names, nothing else.
- `build_seed_profile` freezes those as `skill_refs=[names]`.
- Launch discovery (`discover_profile_skills` → `load_all_skills(load_user=True, load_public=True, load_org=False, load_project=False)`) is a **narrower** catalog.
- A frozen name that isn't in user+public → `resolve_agent_profile` raises `DanglingSkillRef` → the conversation fails to start.

```
global skills = [Skill(name="alpha")]      # inline, in no repo
seed.skill_refs == ["alpha"]
resolve(..., available_skills=[])           # user+public lacks "alpha"
=> DanglingSkillRef(['alpha'])              # fatal
```

**But the scope is narrower than "blocker" suggests:** the *default* zero-skill global is completely safe (`skill_refs=[]` → discovery skipped → nothing to dangle). It only bites a global that carries inline/org/custom-marketplace skills. So it's not "every launch breaks" — it's "skill-carrying setups can't launch their seeded default," which is still bad (it's the exact enterprise shape #15060 targets), just not universal.

One correction to the finding's framing: the CI-invisibility claim is right — `test_seed_freezes_resolved_skills_by_name` sets up the dangerous config but stops at seed construction and never launches, so all 141 profile tests pass while the break is live.

### Root cause — agreed, and it's a contract mismatch, not a seed bug

The finding's "root cause" section is correct and it's the important part. The resolver borrowed MCP's **hard-fail-on-dangling** contract without MCP's **complete-authoritative-catalog** guarantee:

- `mcp_config` *is* the full map → a dangling MCP ref is genuinely a user error.
- The skill discovery catalog is **strictly narrower** than the set of skills that can legitimately name a profile (inline, org, project, custom-marketplace). So a dangling skill ref is often *not* an error — it's just "not in this particular catalog."

Given that asymmetry, hard-failing skill refs punishes legitimate configs.

### What I'd do

I'm with @enyst — **make dangling refs non-fatal (drop-with-log) on the launch path**, and I'd apply the same friendliness to MCP. This is option (2) in the finding, and it's the more general fix:

- It fixes the seed blocker without reverting the freeze-by-name behavior-preservation.
- It unblocks org/project skills and a future picker, which would otherwise re-trip the same wire.
- It matches how `codex` behaves (warn + let the user continue/fix/restart), which is the friendlier UX enyst described.
- **Keep the dry-run's dangling *report*** (`AgentProfileDiagnostics.dangling_skill_refs`) exactly as-is — the editor/`/materialize` should still surface set/missing honestly. Only the *launch* path should degrade to a warning instead of raising.

I'd lean away from option (1) (revert seed to `skill_refs=None`): it dodges *this* dangle, but `None` = "all discovered," which silently *adds* skills a partial-source global never had (the very thing the freeze was introduced to prevent), and it leaves the fatal-dangling contract in place to bite the picker later.

On the ACP double-ingestion point (#4019): agreed too — pushing duplicate/AGENTS.md content into an ACP CLI that already reads it natively is a headache with little upside. A per-profile opt-out (or just not forcing `load_project_skills=True` for ACP) seems saner than shipping the duplication and tracking it as a follow-up.

Net: the PR's *direction* is right (secret-free profile, shared baseline), the seed change is fine **once dangling stops being fatal**. I'd gate merge on flipping that contract + a launch-level test, and fold the 201-vs-422 migration asymmetry and the two cleanups (#8/#9) in while here.

---
_This comment was generated by an AI agent (OpenHands) on behalf of @enyst._

simonrosenberg:
> **Stepping back from the line-level review to the architecture.** This PR's cleanup — removing embedded `skills` and the cipher machinery — is correct and should land. But it doesn't touch the deeper problem, and the seed → `DanglingSkillRef` blocker in the review above is a *symptom* of it, not a one-off. Posting the root-cause analysis here for the record.

## Skills is built on the wrong substrate: reference-and-resolve (like MCP) over a *discovered* catalog

**The wrong assumption, in one sentence:** `skill_refs` — names filtered against a "catalog" — copies the MCP-server design (`mcp_server_refs` filtered against `mcp_config`). But MCP servers are **user-configured and persisted**, while skills are **discovered from many lazy, external, nondeterministic sources**. You can safely reference-filter a persisted set. You cannot safely reference-filter a discovered one — because the set you reference doesn't exist as a single, stable object.

Every recurring symptom in this subsystem — the successive launch-breaking regressions, the cipher/mask/restore machinery, the `None`/`[]`/freeze tri-state default, the seed migration that bricks launches — descends from that one mismatch.

### MCP and skills look like twins but have opposite substrate

The resolver literally shares `_partition_refs` between them and defines parallel `DanglingMcpServerRef` / `DanglingSkillRef` exceptions (`resolver.py:142–200`). But:

| | `mcp_server_refs` → `mcp_config` | `skill_refs` → skill catalog |
|---|---|---|
| Where the set lives | The user's **persisted settings** (`agent_settings.mcp_config`) | **Nowhere** — assembled on demand |
| How it's produced | The user typed it in | Discovered from ~7 sources |
| Complete? | Yes — every referenceable server is in it | No — org / project / inline / marketplace routinely excluded |
| Deterministic? | Yes | No — network fetches, cache TTLs, workspace/auth context |
| Same set at pick-time and launch-time? | Yes, same object | **No — different computations entirely** |

Hard-failing on a dangling ref is *correct* for MCP (a missing server is a genuine config error) and *catastrophic* for skills (a legitimately-authored name goes fatal the moment the two catalogs drift). That is the direct cause of the seed → `DanglingSkillRef` blocker in the review above: the seed freezes names resolved from one catalog that a different, narrower launch catalog can't reproduce.

### There is no "the catalog" — there are ~8 of them

Tracing skill enumeration across the SDK, agent-server, canvas (#1571), and cloud (#15060) surfaced these distinct "what skills exist" computations, **none of which agree**:

1. Canvas frontend `SKILLS_CATALOG` — a **build-time snapshot** from the pinned `@openhands/extensions` package; shipped *inline* on a normal launch (`agent-server-adapter.ts:659`).
2. Canvas `SkillsService.getSkills` — server user+project + bundled public (feeds the skills-settings modal, **not** the profile editor).
3. agent-server `discover_profile_skills` — **user+public only**, default marketplace; resolves `skill_refs` on a *local* profile launch (`skills_service.py:460`).
4. agent-server `load_all_skills` / `/api/skills` picker — **all** sources incl. org/project/marketplace/sandbox (`skills_service.py:342`).
5. agent-server `service_get_marketplace_catalog` — `default.json`, the install UI.
6. SDK `AgentContext._load_auto_skills` — user+public, but respects `marketplace_path` + `registered_marketplaces`; resolves skills on the *normal* launch (`agent_context.py:219`).
7. Cloud app-server picker `search_skills` — **two local dirs only** (`GLOBAL_SKILLS_DIR` + `~/.openhands/microagents`); no org, project, or marketplace.
8. Cloud launch discovery `/api/skills` — public+user+project+**org**+sandbox+marketplace.

A user picks from one of these; the runtime resolves against a *different* one. Divergence axes include `marketplace_path`, registered-marketplaces, org tier, project tier, public-cache TTL, server-vs-frontend `@openhands/extensions` version, and host FS. Since the resolver hard-fails unknown refs, every axis is a latent launch failure.

### Two facts that show the model isn't load-bearing anywhere today

1. **In production cloud, `skill_refs` is inert.** The cloud resolver is called with `available_skills=None` (`saas_settings_store.py:215`, `agent_profiles.py:509`), which the SDK maps to `return [], [], []` (`resolver.py:194`) — refs resolve to nothing and nothing is flagged dangling. Cloud actually selects skills with a **`disabled_skills` deny-list** (`app_conversation_service_base.py:253`): attach all discovered skills, turn some off. **Production already uses a different, opposite selection model than the allow-list this whole edifice is built around.**
2. **In canvas, there is no skill picker.** The profile editor exposes name, agent kind, sub-agents, concurrency, and an LLM dropdown — and no skill selection (`agent-profiles-local-view.tsx`). Every profile is created with `skill_refs: null` = "all discovered." A subset can only *survive* an edit; it can never be *authored*.

So the "subset of skills" capability — the reason `skill_refs`, the tri-state, the `DanglingSkillRef` contract, the seed-freeze, and the cipher-adjacent complexity all exist — **is unbuilt in the UI, inert in cloud, and crash-prone locally.** The machinery has negative value: it can't deliver the feature and it breaks launches trying.

### A compounding, separate smell: `Skill.mcp_tools`

Embedded skills dragged in cipher/mask/restore (~130 lines in the cloud router alone, plus store-cipher plumbing) only because a `Skill` can carry `mcp_tools: dict[str, MCPServer]` — i.e. a "skill" (a prompt/context extension) is conflated with executable tooling that carries credentials. That conflation is what made "secret-free at rest" impossible and forced the encryption apparatus. This PR correctly deletes the *symptom* (embedded skills), but the underlying "a skill can embed a secret-bearing MCP server" conflation remains in the `Skill` model and is worth splitting out.

### Fix direction (a design decision, not another resolver patch)

- **Option A — adopt the deny-list model everywhere (recommended).** It's what cloud already does, and it's the naturally correct model for a *discovered* set: "all discovered skills, minus the ones this profile turns off." An unknown/removed skill simply isn't there to disable — **no dangling, no crash, graceful drift.** This unifies local + cloud on the mechanism production already uses, lets us delete `skill_refs`, `DanglingSkillRef`, the tri-state default, and the seed-freeze entirely, and makes the canvas UI a checklist over the *live* discovered catalog.
- **Option B — keep allow-list refs but stop pretending it's MCP.** Make refs **non-fatal by construction** (drop-with-log, never raise), and make the picker and the resolver the **same computation**: the picker resolves the catalog, the user selects, and the *resolved snapshot the user actually saw* is what launch uses — resolve against that snapshot, never re-discover.

Non-negotiables either way: **one catalog computation shared by authoring and launch; skill selection tolerant of a drifting/incomplete catalog (never fatal); and stop sharing the reference-resolution abstraction with MCP.** Separately, split `Skill.mcp_tools` out — a context extension shouldn't carry credentials.

<sub>Root-cause traced across the SDK profiles/resolver, the agent-server skills service, agent-canvas #1571, and OpenHands #15060.</sub>

simonrosenberg:
## Re: "Skills is built on the wrong substrate" — verified, and I shipped the non-fatal-refs half

I didn't take the comment at its word (@enyst's standing instruction). I read the code at PR head, cloned `OpenHands/OpenHands` and `OpenHands/agent-canvas`, and reproduced the failure. **The core thesis holds and the immediate blocker is real** — so I've landed the one fix everyone here already converged on. Details below, including where I'd restate the framing.

### What I verified true (against the code, not the prose)

- **`skill_refs` copies MCP's reference-and-resolve shape.** `resolver.py` shares `_partition_refs` between the two and defines parallel `DanglingMcpServerRef` / `DanglingSkillRef`. Confirmed.
- **The two substrates really are opposite.** `mcp_config` is the user's complete, persisted map — a dangling MCP ref *is* a config error. The skill "catalog" is assembled on demand: `discover_profile_skills()` (`skills_service.py:460`) is **user+public only** (`load_org=False, load_project=False`), while the picker `load_all_skills()` (`skills_service.py:342`) defaults **all** sources on. So the launch catalog is strictly *narrower* than the authoring catalog. Confirmed.
- **The blocker reproduces end-to-end.** A global carrying an inline/org skill → `build_seed_profile` freezes `skill_refs=[names]` → launch discovery (user+public) lacks the name → `DanglingSkillRef` → `conversation_router.py` maps it to **HTTP 422** → the conversation can't start. Minimal repro:
  ```python
  from openhands.sdk.profiles.resolver import _compute_skill_filter
  _compute_skill_filter([], ["alpha"])   # -> ([], [], ["alpha"]); resolver then raised -> 422
  ```
  The default zero-skill global stays safe (`skill_refs=[]` skips discovery), so the scope is "skill-carrying setups can't launch their seeded default" — narrower than "every launch breaks," but it's exactly the enterprise shape #15060 targets.
- **Cloud already uses the opposite (deny-list) model.** `app_conversation_service_base.py:254` attaches all discovered skills and filters by `user.disabled_skills` — an allow-list `skill_refs` is not how production selects skills. Confirmed on `main`.
- **Canvas has no skill picker.** The profile editor's sections are overview/general/model/tools-mcp/condenser/provider-model/launch/mcp/authentication (`editor/sections.ts`) — no skills section; `skills` rides through as opaque `extras`. So a skill *subset* can't be authored in the UI today. Confirmed on the editor branch.
- **`Skill.mcp_tools`** (`skill.py:157`) genuinely lets a "skill" carry a secret-bearing `MCPServer` — that conflation is what forced the cipher/mask apparatus this PR deletes. Confirmed.

### Where I'd restate the framing

- **"~8 catalogs, none agree" is directionally right but partly counts unmerged code.** The specific cloud line refs (`agent_profiles.py:509`, `saas_settings_store.py:215` resolving with `available_skills=None`) are from **PR #15060, which isn't merged** — on `main` there is no profile-resolve path in production cloud at all. The *substrate* point stands without them; I'd just not cite unmerged lines as current-state evidence.
- **"The machinery has negative value" — I'd soften to "net-negative *as currently wired*."** The allow-list can't be authored (no UI) and is inert in cloud (deny-list), while the *fatal* contract actively breaks launches. That's a real net-negative **today**. But a future picker that resolves-then-launches against the same snapshot (the comment's own Option B) would give the subset feature value — so it's the fatal coupling that's wrong, not "subsetting skills" as a concept.

### WDYT / what I did

I agree with @enyst ("the Dangling error doesn't need to be fatal... just give the user a warning") and with the earlier on-behalf-of-@enyst conclusion (gate on flipping the contract + a launch-level test). That's **Option B's non-negotiable core**, it's separable from the bigger A-vs-B design call, and it unblocks this PR. So I shipped exactly that and nothing more:

- `resolve_agent_profile` now **drops** a dangling `skill_refs` entry with a `logger.warning` instead of raising `DanglingSkillRef`.
- The dry-run still reports `dangling_skill_refs` honestly but **no longer flips `valid`**, so `/materialize` agrees with launch (editor still sees set/missing).
- **Dangling MCP refs stay fatal** — `mcp_config` is authoritative, so that's still a genuine config error.
- `DanglingSkillRef` is retained/exported (API compat) but is no longer raised; the dead 422 handler in the start path is removed.
- Tests updated + launch-level coverage that a dangling ref resolves (not 422). `tests/sdk/profiles/` (141) + the two agent-server profile suites pass; ruff/pyright clean.

Pushed as `3d4be43`.

### What I deliberately did **not** do here (agreeing it's a design call, not a resolver patch)

- **Option A (deny-list everywhere) + deleting `skill_refs`/the tri-state/seed-freeze.** It's the right long-term unification with what cloud already does, but it spans SDK + agent-server + canvas + cloud and changes the wire contract — too big to fold into this PR silently. Worth its own issue.
- **Splitting `Skill.mcp_tools` out** of the context model. Agreed it's the deeper conflation; separate change.
- **Same friendliness for MCP** (warn-not-fail). @enyst floated it; I kept MCP fatal for now since it's a different substrate and out of this PR's scope.
- **ACP `load_project_skills` duplication (#4019)** — already tracked; agree a per-profile opt-out beats shipping duplication.

Net: this PR's direction is right, the seed change is fine now that dangling stops being fatal, and the broad substrate rework is a real follow-up I'd file rather than block this on.

_This comment was generated by an AI agent (OpenHands) on behalf of @simonrosenberg's review._

simonrosenberg:
Pivoted this PR to **Option A (the deny-list)** from the architecture note above, and force-pushed to supersede the interim Option B commit (`3d4be43fc`, "make dangling skill_refs non-fatal (drop-with-log)"). The two are mutually-exclusive edits to the same resolver, and Option A structurally subsumes B — a deny-list can't dangle in the first place, so there's nothing to make non-fatal.

**What changed (`1986d9562`, net −117 lines):**
- `OpenHandsAgentProfile.skill_refs` → `disabled_skills: list[str] = []`: a deny-list over the discovered catalog. `[]` = all skills; a name absent from the catalog is a no-op, so a profile can never fail launch on catalog drift. `ACPAgentProfile` drops the skill field entirely.
- `AgentContext.disabled_skills`, applied in `_load_auto_skills` and the `LocalConversation` project-skill merge — one drift-tolerant chokepoint honored by every skill source.
- Deleted `_compute_skill_filter` + `DanglingSkillRef`; added `_apply_disabled_skills` (never raises). Removed `discover_profile_skills_if_needed` (the `[]`-skip tri-state).
- Seed sets `disabled_skills=[]` — the freeze-by-name launch break is structurally gone (regression test `test_seed_then_resolve_with_narrower_catalog_does_not_dangle` seeds an inline skill absent from the launch catalog and asserts resolve succeeds).
- Schema v2→v3 migration drops `skill_refs`; `safe_validation_error_detail` now returns `msg` (no secret-bearing field remains to redact).

308 profiles/context tests pass; ruff + pyright clean.

**Consumer PRs updated to match:**
- **agent-canvas#1571** — already deny-list-native (user-level `disabled_skills` UI + generic profile merge), so only stale comments needed fixing. Per-profile skill picker stays out of scope for the minimal editor.
- **OpenHands#15060** — unions the launched profile's `disabled_skills` with the member's (`effective_disabled_skills`) into the existing deny filter; needs this SDK released + the pin bumped before its CI is green.

neubig (approved):
This seems cleaner than before, thanks!

simonrosenberg:
Review this PR as well as the consumer PRs
 https://github.com/OpenHands/agent-canvas/pull/1571
      https://github.com/OpenHands/OpenHands/pull/15060

simonrosenberg:
## `/codereview-roasted` — reviewed against the checked-out branch (`1986d95`) + both consumer PRs

I read the actual code at PR head, ran the profile/context/agent-server suites (all green), and cross-checked agent-canvas#1571 and OpenHands#15060. Where the prose and the code disagreed, I trusted the code.

**Taste rating: 🟢 Good taste.** The pivot from a `skill_refs` allow-list to a `disabled_skills` deny-list is the right call and — importantly — it *dissolves* the earlier blocker rather than patching around it. A deny-list can't dangle: a disabled name that isn't in the discovered catalog is a harmless no-op, so `DanglingSkillRef → 422` (the launch-break enyst and I flagged) is gone by construction, not by adding an error path. This is exactly the "make the special case disappear" move. Net −117 lines, a whole cipher/mask/`X-Expose-Secrets` apparatus deleted, and the profile is now genuinely secret-free at rest. I'd merge this.

Below is the roast anyway — none of it is a hard blocker on the SDK code.

### What's genuinely good

- **One drift-tolerant chokepoint.** `AgentContext.disabled_skills` is applied in `_load_auto_skills` *and* in the `LocalConversation` project-skill merge, so every skill source (explicit, auto-loaded, lazy project) is filtered by the same deny-list. The comment on the `local_conversation.py` branch correctly calls out *why* it's needed there (`model_copy` bypasses the validator). Good.
- **Secret-free-at-rest is now true, not aspirational.** Dropping `skills[].mcp_tools` removes the only secret-bearing field, which is what lets `save`/`load`/`AgentProfileStoreProtocol`/`save_profile_preserving_identity` shed the `cipher` param and the router shed `X-Expose-Secrets`. The `materialize` path correctly *keeps* `get_cipher` (it decrypts the referenced LLM profile, unrelated). That distinction is subtle and the code got it right.
- **Migrations are honest.** v1→v2 drops `skills`, v2→v3 drops `skill_refs`; both are documented as belt-and-suspenders since no prod data exists. `test_save_v1_body_with_stray_skills_key_is_migrated_and_saved` vs `test_save_current_schema_version_rejects_stray_skills_key` nicely pins the "migrate old / reject current-version stray key" boundary.

- **`agent_profiles_router.py` (~L302) — stale comment.** The `except ValidationError` block still says *"surface only `loc`/`type`"*, but `safe_validation_error_detail` now also returns `msg` (you changed it deliberately, and the docstring explains why it's safe). The inline comment contradicts the function it calls. One-line fix: point it at the docstring or say `loc`/`type`/`msg`.

- **Redundant double-filter in `_build_openhands_settings`.** The resolver computes `filtered_skills = _apply_disabled_skills(catalog, disabled)` *and* stamps `disabled_skills=profile.disabled_skills` on the `AgentContext`, whose validator filters again. It's idempotent so it's harmless, but passing the already-filtered list plus the deny-list is slightly confusing to the next reader (looks like belt-and-suspenders where one belt would do). Not worth a code change if it complicates the None-catalog handling — just flagging that the two mechanisms overlap.

### [TESTING GAPS]

- **The new `local_conversation.py` deny-list branch has no direct test.** `tests/sdk/conversation/test_repo_root_project_skills.py` covers `load_project_skills` injection/precedence/failure, but nothing exercises a **project skill whose name is in `disabled_skills`** being dropped during the lazy merge. That's a real code path added in this PR (the `model_copy`-bypasses-validator case). `AgentContext.disabled_skills` and the resolver path are both well covered; this one branch isn't. A ~15-line test in that file would close it.

### Consumer PRs

**OpenHands#15060 (cloud backend)** — ✅ aligned. Updated today; it consumes `disabled_skills`/`#4017` throughout (`effective_disabled_skills()` unions member- and profile-level deny-lists, `GET`/`POST` documented as needing no masking, `cipher=None`). This is the PR that most needed the SDK to land first, and it reads as if it did. Good.

**agent-canvas#1571 (frontend)** — ⚠️ mostly aligned, but carries stale `skill_refs` residue that will confuse the next reader and *could* rot:
  - Production types/service are on the deny-list model (`agent-profiles-service.api.ts` comments `disabled_skills`, `exposeSecrets` vestigial). Good.
  - But `tests/e2e/mock-llm/utils/mock-llm-helpers.ts` still comments *"the server now defaults an unset OpenHands profile's `skill_refs` to null (all discovered skills)"* — that's the **old allow-list design**; the field no longer exists.
  - Several test fixtures still set `skill_refs: [...]` / `skill_refs: null` (`agent-profiles-local-view.test.tsx`, `merge-agent-profile-save-input.test.ts`). Because a POST without `schema_version` migrates and drops `skill_refs`, these won't fail at runtime — but they're asserting against a field this PR removed, so they document a contract that's gone. Worth a cleanup pass on the canvas PR before it merges, and I'd double-check the pinned ts-client actually ships the `disabled_skills` field the frontend now depends on.

Since the canvas PR is the one still holding stale `skill_refs`, I'd make sure it's rebased on this design (not the pre-pivot one) before it merges, so the "downstream can delete the shims" claim actually holds.

### [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟢 **LOW**. Server/SDK-scoped config plumbing; the behavior change (deny-list, forced `llm.stream=True` on the profile launch path) is well-isolated and covered by `test_openhands_profile_forces_llm_stream_true` / `test_acp_profile_does_not_force_llm_stream`. Streaming is a transport toggle wired to `on_token`, not a prompt/loop/tool change, so eval-behavior risk is negligible. No secrets leave the environment; the change *removes* a secret-handling surface. All CI green except two process gates: `Validate PR description` (the template heading must be `## How to Test`, not `## How to test`) and `unresolved-review-threads`.

**VERDICT:** ✅ **Worth merging.** Address the stale router comment + add the one project-skill deny-list test if you want it airtight; neither blocks. Fix the `## How to Test` heading and resolve the open review threads to clear CI, and make sure agent-canvas#1571 is on the deny-list design before it lands.

**KEY INSIGHT:** Modeling skills as a deny-list instead of an allow-list is the correct data-structure choice — it makes the "catalog drift" edge case structurally impossible instead of turning it into a 422, which is why this PR can delete more than it adds.

---

> **Improve this review?** If any feedback above seems incorrect or irrelevant, add/edit `.agents/skills/custom-codereview-guide.md` on the branch with the `/codereview` trigger and re-request a review.
>

simonrosenberg:
Thanks for the careful pass — addressed the actionable items.

**SDK (`a65a40d9c`):**
- ✅ **Stale 422 comment** (`agent_profiles_router.py`): now reads `loc`/`type`/`msg` (`input` dropped), matching `safe_validation_error_detail`.
- ✅ **Testing gap**: added `test_disabled_skills_drops_project_skill_at_lazy_merge` in `test_repo_root_project_skills.py` — it exercises the one deny-list branch outside the validated path (the `LocalConversation` `model_copy` project-skill merge that bypasses the `AgentContext` validator): a disabled project skill (`AGENTS.md` → `agents`) is dropped, and an absent disabled name is a no-op.
- **Double-filter** in `_build_openhands_settings`: left as-is (you flagged it as not worth a change). The resolver pre-filter produces the `resolved_skills` diagnostic and the pre-loaded user/public skill set; the `disabled_skills` stamped on the context is what the *lazy project-skill* merge needs. The overlap is only on the already-loaded user/public set and is an idempotent no-op — removing either belt would drop coverage on one of those two paths.

**CI gate:**
- ✅ Fixed the `## How to Test` heading (was lowercase `test`).
- The `unresolved-review-threads` gate is a manual resolve — leaving those to you.

**agent-canvas#1571 (`10b7820f`)** — cleaned up the stale `skill_refs` residue you found:
- `merge-agent-profile-save-input.test.ts` + `agent-profiles-local-view.test.tsx`: `skill_refs` → `disabled_skills`, dropped the removed embedded `skills`, `schema_version: 3`, ACP fixtures drop the skill field (ACP has none), and corrected the stale `exposeSecrets`/`mcp_tools` comment (profiles are secret-free now). 10/10 tests pass.
- `mock-llm-helpers.ts`: the omitted-field comment now describes the deny-list default.
- **ts-client check**: the pinned client (`1.32.0`) types neither `skill_refs` nor `disabled_skills`, and canvas never references either by name (the profile merge is a generic spread + the fixtures ride untyped via `as unknown` casts), so there's no runtime dependency — the field rides untyped until the ts-client regen at the SDK release / pin bump.

Everything green locally: SDK `308 + 59` profile/router/conversation tests, canvas `10/10` on the touched files, ruff/pyright clean.

VascoSch92:
`_apply_disabled_skills` drops the catalog de-dup that `_compute_skill_filter` had — duplicate-named catalog now crashes resolution

resolver.py — the old _compute_skill_filter de-duplicated by name (by_name = {s.name: s for s in available_skills}). Its replacement _apply_disabled_skills returns list(available_skills) / a plain comprehension with no de-dup, so duplicates flow into AgentContext(skills=…), whose _validate_skills validator raises ValueError: Duplicate skill name found → resolve_agent_profile throws ValidationError.

```
cat = [Skill(name="review",...), Skill(name="lint",...), Skill(name="review",...)]  # dup
resolve_agent_profile(prof, llm_store=ls, mcp_config={}, available_skills=cat, cipher=None)
```

Not reachable via the in-repo agent-server flow (both callers pass `discover_profile_skills()`, which pre-dedups through `merge_skills`). It bites any external caller of the exported resolve_agent_profile that passes a non-deduped catalog — which this PR's design explicitly anticipates (the app-server supplying a "fuller catalog" whose sources `load_all_skills` documents as colliding by name; #15060 is a direct caller). Looks like an unintended drop during the dict→comprehension refactor — the docstring still claims "this can never raise."

Fix
```python
if not available_skills:
    return []
by_name = {s.name: s for s in available_skills}
denied = set(disabled)
return [s for s in by_name.values() if s.name not in denied]
```

simonrosenberg:
Good catch — confirmed and fixed in `780de42de`.

You're right that the dict→comprehension refactor dropped `_compute_skill_filter`'s by-name de-dup, so a colliding catalog flowed into `AgentContext(skills=…)` and tripped `_validate_skills` → the resolve raised, contradicting the docstring's "can never raise." Not reachable via the in-repo callers (`discover_profile_skills()` / `None` are both de-duped through `merge_skills`), but `resolve_agent_profile` is exported and the design does anticipate the app-server passing a merged multi-source catalog (#15060 is a direct caller), so restoring the de-dup is the right defensive fix.

Applied essentially your patch — by-name dict, last occurrence wins to match `load_all_skills`'s later-source-overrides precedence — updated the docstring to state the de-dup, and added a regression test (`test_duplicate_named_catalog_is_deduped`) that resolves a catalog with a duplicate name plus a disabled name. Thanks for the precise repro.

PR #4028, opened 2026-07-08
fix(agent-server): persist LLM/profile switch to meta.json so it survives restart

### LINKED ISSUE #4032: [Bug]: LLM profile timeout is reset after agent-server restart

## Description

When `agent-server` restarts, the configured LLM timeout value is not preserved. The system resets the timeout to a default value (or to a value that does not exist in any configured LLM profile).

## Steps to reproduce

1. Change the active LLM profile timeout value to `1234`.
2. Restart the `agent-server` container.
3. Reconnect OH to the `agent-server`.
4. Check the timeout value being used by the agent.

## Expected behavior

The agent should continue using the LLM profile timeout value configured before the restart (`1234`) or the one of the active profile.

## Actual behavior

After reconnecting, the agent uses a different timeout value. The configured value is lost after the `agent-server` restart.

## Environment

* OH commit: `2abf0b1303c8d3c8dcd9fb080507395afbaa21d4`
  Date: Wed Jul 8 21:17:44 2026 +0700

* agent-server commit: `65ee52f817f5b30292abe5f2ac620f84576eb91f`
  Date: Wed Jul 8 16:16:47 2026 +0200

VascoSch92:
HUMAN:

Fix bug with restart.

---

AGENT:

## Why

When the agent-server restarts (e.g. the container is stopped and the conversation is reopened), a conversation's LLM config — including `timeout` — silently reverts to the values captured at conversation *creation*. Switching the LLM profile appears to fix it, but the fix is lost again on the next restart.

Root cause: switching the LLM updates only the live `ConversationState` (persisted to `base_state.json`). On resume, `EventService.start()` rebuilds the agent from `meta.json` (`self.stored.agent`) and `ConversationState.create()` overwrites the just-loaded `base_state.json` agent with it (`state.agent = agent`). Because the switch was never mirrored into `meta.json`, the creation-time LLM wins on every restart.

There are three ways the live LLM changes, and all three were affected: the `/switch_llm` endpoint (app-servers, #3017), the `/switch_profile` endpoint (agent-canvas), and the agent's own built-in `switch_llm` tool (`SwitchLLMExecutor` → `conversation.switch_profile`, which bypasses the HTTP layer entirely). The ACP model-switch path (`switch_acp_model`) already handles exactly this by writing the change back to `meta.json`; the LLM/profile paths were missing it.

## Summary
- Add a single `EventService._sync_llm_to_meta()` helper that writes a swapped LLM back into `meta.json` (`save_meta`), matching the existing `switch_acp_model` precedent.
- Route the `/switch_llm` and `/switch_profile` endpoints through new `EventService` methods that call the helper, and call it from the post-run hook so the agent's in-run `switch_llm` tool is covered too — all three switch paths now survive a restart.
- Only the stored agent's `llm`/`condenser` are updated (not the whole live agent), keyed on LLM object identity. This keeps the stored agent **plugin-unmerged** — `_ensure_plugins_loaded` merges plugin `agent_context`/`mcp_config` into the live agent on first run, and persisting that would double-merge on resume — and makes the post-run hook a no-op for ordinary/plugin-loading runs.

## Issue Number

Fix #4032 
Addresses the report: *"agent-server sets wrong LLM timeouts when it starts; switching the LLM profile sets the correct values."* Reproduction below.

## How to Test

Automated (real resume from disk / real agent run, not just unit mocks):

```
uv run pytest \
  tests/agent_server/test_conversation_service.py::test_switch_llm_survives_restart \
  "tests/agent_server/test_event_service.py::TestEventServiceSaveMeta" \
  tests/agent_server/test_conversation_router.py -k switch
```

- `test_switch_llm_survives_restart` (endpoint path) starts a real `ConversationService`, creates a conversation whose LLM has `timeout=300`, switches to an LLM with `timeout=600`, tears the service down, then starts a fresh `ConversationService` on the same directory (a genuine resume from disk) and asserts the resumed conversation still reports `timeout=600`.
- `test_in_run_switch_llm_persists_to_meta` (tool path) starts a real `EventService`, swaps the live LLM directly on the conversation (simulating the agent's `switch_llm` tool, bypassing the endpoints), drives a real run to completion with a scripted `TestLLM`, and asserts the post-run hook wrote `timeout=600` into `meta.json`.
- `test_sync_llm_to_meta_ignores_plugin_merge` guards the plugin case: a plugin-style agent swap (new agent object, same LLM) must **not** be persisted, so the stored agent stays unmerged. (Verified it fails if the whole live agent is mirrored instead.)

Manual reproduction (matches the report):
1. Give a conversation an active LLM profile with a non-default `timeout`.
2. Stop the agent-server / container, reopen the conversation so it restarts and reconnects.
3. `cat /workspace/conversations/<SESSION>/*.json | jq '.. | .timeout? // empty'` — before this change the timeout reverts to the creation-time value; after it, the switched value persists.

## Video/Screenshots

End-to-end evidence that the fix is exercised (not unit-mocked): each regression test fails when its persistence hook is removed (pre-fix behavior) and passes with the fix.

Endpoint path — pre-fix (mirror step reverted):
```
$ uv run pytest tests/agent_server/test_conversation_service.py::test_switch_llm_survives_restart -q --tb=line
tests/agent_server/test_conversation_service.py:582: AssertionError: assert 300 == 600
FAILED ...::test_switch_llm_survives_restart
```

Tool path — pre-fix (post-run hook removed):
```
$ uv run pytest "...::test_in_run_switch_llm_persists_to_meta" -q
FAILED ...::test_in_run_switch_llm_persists_to_meta
```

Both pass with the fix. Full suites for the touched files pass locally (`test_event_service.py`, `test_conversation_service.py`, `test_conversation_router.py`, `test_goal_loop.py` — 306 total). `ruff check`, `ruff format`, and `pyright` are clean (all pre-commit hooks pass).

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes
- Updates `llm` **and** `condenser` (the fields a `switch_llm` changes: condenser is disabled for subscription LLMs / rebuilt for the new model), so the condenser LLM's `timeout` is fixed on resume too — not just the primary LLM. `save_meta` serializes with the cipher context, so at-rest secret encryption is unchanged.
- The post-run hook is best-effort (failures logged, never break run completion) and a no-op unless the LLM object actually changed, so ordinary and plugin-loading runs don't rewrite `meta.json`.
- ACP model switches keep their existing dedicated mirroring (`switch_acp_model`) — untouched.
- No API/schema changes: both endpoints keep their request/response shapes and status codes (`404` unknown conversation, `404` unknown profile, `400` corrupted profile).


openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -1415,6 +1431,57 @@ async def switch_acp_model(self, model: str) -> None:
         )
         await self.save_meta()
 
+    async def _sync_agent_to_meta(self) -> None:
+        """Mirror the live agent into ``meta.json`` when it has changed.
```
VascoSch92: trimmer the docstring

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -996,6 +1002,16 @@ async def _run_and_publish():
                     self._run_task = None
                     await self._publish_state_update()
 
+                    # Persist an in-run agent swap (the agent's switch_llm tool)
```
VascoSch92: trimmer comment

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -86,6 +86,10 @@ class EventService:
     owner_instance_id: str = field(default_factory=lambda: uuid4().hex)
     lease_ttl_seconds: float = DEFAULT_LEASE_TTL_SECONDS
     _conversation: LocalConversation | None = field(default=None, init=False)
+    # The agent object last mirrored into meta.json. Lets _sync_agent_to_meta
```
VascoSch92: trim comment

VascoSch92: Trimmed to a single line.

openhands-agent-server/openhands/agent_server/event_service.py
VascoSch92: Trimmed to two lines.

openhands-agent-server/openhands/agent_server/event_service.py
VascoSch92: Trimmed the docstring. It's shorter but I kept the two load-bearing points — *why* we persist (resume rebuilds the agent from meta.json) and *why* only llm/condenser (so plugin-merged context isn't baked in). Also reworked this method to fix a real bug that surfaced here: mirroring the whole live agent would double-merge plugins on resume.

simonrosenberg (approved):

enyst (changes requested):
Are we sure this is the best code design? The tool firing is an event we could capture with a hook or maybe a new listener, rather than attempting to always save 🤔

Could we analyze the options a bit here?

enyst (commented):
Issue #4032 notes:

>  The agent should continue using the LLM profile timeout value configured before the restart (1234) or the one of the active profile.

I see from this PR that what happened is that the entire profile switch may have not been persisted as expanded LLM. That makes me wonder, could we maybe save the profile, rather than the expanded LLM? And load it on restore.

Alternatively, since that is quite involving, for the bug fix maybe we could read the active profile at restore time?

Please consider it just quick thoughts, open for discussion.

openhands-agent-server/openhands/agent_server/conversation_router.py
```diff
@@ -465,9 +465,8 @@ async def switch_conversation_profile(
     event_service = await conversation_service.get_event_service(conversation_id)
     if event_service is None:
         raise HTTPException(status.HTTP_404_NOT_FOUND)
-    conversation = event_service.get_conversation()
     try:
-        conversation.switch_profile(profile_name)
+        await event_service.switch_profile(profile_name)
```
kripper, line 476: 👍

openhands-agent-server/openhands/agent_server/conversation_router.py
```diff
@@ -499,11 +498,10 @@ async def switch_conversation_llm(
     event_service = await conversation_service.get_event_service(conversation_id)
     if event_service is None:
         raise HTTPException(status.HTTP_404_NOT_FOUND)
-    conversation = event_service.get_conversation()
     cipher = get_cipher(request)
     if cipher is not None:
         llm = decrypt_incoming_llm_secrets(llm, cipher)
-    conversation.switch_llm(llm)
+    await event_service.switch_llm(llm)
```
kripper, line 511: 👍

VascoSch92:
> Issue #4032 notes:
> 
> > The agent should continue using the LLM profile timeout value configured before the restart (1234) or the one of the active profile.
> 
> I see from this PR that what happened is that the entire profile switch may have not been persisted as expanded LLM. That makes me wonder, could we maybe save the profile, rather than the expanded LLM? And load it on restore.
> 
> Alternatively, since that is quite involving, for the bug fix maybe we could read the active profile at restore time?
> 
> Please consider it just quick thoughts, open for discussion.

@enyst I did consider it. 

Two reasons I went with persisting the expanded LLM instead:

1. The `/switch_llm` endpoint passes an inline LLM with no profile on disk to reload, so a profile-reference scheme can't cover all three switch paths uniformly. It'd need two persistence formats. (which is not so nice)
2. More importantly, a running conversation already freezes the profile at switch time (the expanded LLM is cached in the registry, first-write-wins, later edits to the profile file aren't picked up until an explicit re-switch). Persisting the expanded LLM keeps a restarted conversation behaving identically to one that was never restarted. Reloading the profile at restore would instead make a restart silently pull in profile edits the live process was ignoring.

Make it sense, or?

kripper:
Makes sense. It's a fix for the current expanded LLM config. Thanks!

It also makes sense to deprecate the redundancy, but the name "expanded" sounds more like an expanded per-conversation config override.

In this case, the live in-memory config (of a session that has never been restarted) should be:

`llm_profile_config + convo_expanded_config`

And the configs should always be written to the location where the agent can load them again after a restart.

kripper:
and architectural changes are better handled in a separate PR

kripper:
I cherry-picked the PR and on my env the issue is still present, ie. agent-server resets the LLM timeout once OH reconnects.
I will continue testing...

enyst:
Just to note, for investigation of the best solution here:

> The /switch_llm endpoint passes an inline LLM with no profile on disk to reload, so a profile-reference scheme can't cover all three switch paths uniformly. It'd need two persistence formats. (which is not so nice)

Doesn’t the endpoint change the “active” profile? If not, I think maybe it should. The human-triggered switch does… 🤔 

On a side note, we already have two persistence formats, that is why I think it’s worth thinking how to reduce them to one: profiles, as much as possible. Maybe the agent is confused here, though maybe it’s me. 😅

kripper:
> I cherry-picked the PR and on my env the issue is still present, ie. agent-server resets the LLM timeout once OH reconnects.
> I will continue testing...

I tested again with latest version of OH:
- with agent-server (latest version) => issue is present
- with agent-server (this PR) => issue is solved

So the PR solves the issue.

enyst:
@openhands describe what does the PR do, if the user has a conversation which is started, works, then later restored, but they never used any of switch llm profile during this conversation.

reply directly on gh.

enyst:
If the conversation never switches LLM/profile, this PR does not change its behavior.

At conversation creation, the expanded LLM config is stored in `meta.json`. On the first start, this PR records that LLM object as the already-synced baseline. Ordinary runs (including plugin loading) keep the same LLM object, so the new sync path is a no-op and does not rewrite `meta.json`. When the conversation is restored, it is rebuilt from the same expanded LLM snapshot in `meta.json`.

Therefore:
- if the desired timeout was already in `meta.json` when the conversation was created, that value is used after restore;
- if the profile file/active profile was changed later, but the conversation never switched to it, this PR does **not** reload that profile on restore—the conversation continues using its creation-time expanded config.

So this PR specifically makes an actual live LLM/profile switch durable; it does not change profile resolution for conversations that never switched.

_This reply was generated by an AI agent (OpenHands) on behalf of the user._

kripper:
FYI, I depend on this PR (or a similar solution).

VascoSch92:
@openhands solve the conflicts

VascoSch92:
@enyst what do you think?

VascoSch92:
Resolved the merge conflicts with `main` (pushed as b8f4ca6a).

The only conflict was in `tests/agent_server/test_conversation_service.py`, where both branches added a new test at the same spot:
- kept this PR's `test_switch_llm_survives_restart`
- kept main's `test_startup_and_search_do_not_hydrate_idle_conversation`

One semantic adjustment was needed: main's #4100 ("Lazily hydrate persisted conversations") means idle conversations are no longer eagerly loaded on restart, so the restart assertion now explicitly hydrates via `await restarted.get_event_service(conversation_id)` (the genuine resume-from-`meta.json` path) instead of reading `restarted._event_services.get(...)` directly.

Verification:
- `test_switch_llm_survives_restart` passes.
- Full touched suites pass: `test_conversation_service.py`, `test_event_service.py`, `test_conversation_router.py` — 297 passed.
- `ruff check` and `ruff format --check` clean on the touched file.

The PR is now mergeable.

VascoSch92:
@enyst :-)

enyst (commented):

🔴 **Needs improvement** - the synchronization marker is committed before the durable state it is meant to represent.

**[CRITICAL ISSUES]**
- [`openhands-agent-server/openhands/agent_server/event_service.py`, Line 1676] **Persistence Correctness**: `_synced_llm` advances before `save_meta()` succeeds. The callback deliberately catches persistence errors, so one transient failed write leaves `meta.json` stale while the in-memory marker says this exact LLM is already synchronized; every later agent update then returns at the identity guard and never retries. Update the marker only after a successful save. Keep the candidate `StoredConversation` local until persistence succeeds as well, so the marker, in-memory metadata, and disk commit advance together.

**[TESTING GAPS]**
- [`tests/agent_server/test_event_service.py`] **Missing Failure-Retry Proof**: add a test where the first `save_meta()` fails and a later matching agent-state update retries and persists the switched LLM. The current exception tests do not prove recovery after a transient disk failure.

**KEY INSIGHT:**
A synchronization watermark must represent committed durable state, not an attempted write.

**[RISK ASSESSMENT]**
The change is localized, but it modifies restart persistence for live LLM/profile switching; a failed write currently defeats the PR's core guarantee until another distinct switch occurs.

**VERDICT:**
❌ **Needs rework**: make the durable write succeed before advancing the synchronization marker.

---

>
>
>

Issue #3978, opened 2026-07-03
Tools: SDK-owned default tool names, single defaulting point in create_agent, per-profile tool selection


simonrosenberg:
## Problem

The default toolset is defined in **four independent copies across three repos**, and none of them is authoritative:

| Where | Contents | Used by |
| -- | -- | -- |
| canvas `getAgentTools()` (`agent-server-adapter.ts:531`) | terminal, file_editor, task_tracker, **canvas_ui**, browser (env-gated), task_tool_set (toggle-gated) | settings-path launches only |
| `openhands-tools` `get_default_tools()` | terminal, file_editor, task_tracker, browser, task_tool_set | direct SDK users |
| PR [#3968](<https://github.com/OpenHands/software-agent-sdk/issues/3968>) resolver literals (`profiles/resolver.py`) | terminal, file_editor, task_tracker, task_tool_set — **no browser** | profile-path launches |
| cloud caller override (`live_status_app_conversation_service.py:1746`) | `get_default_tools(enable_browser=True, ...)` — **browser always on** | every cloud launch, profile or not |

Consequences:

1. **The same agent profile produces three different agents** depending on where it launches: local profile launch (no browser, no canvas_ui) vs. local settings launch (canvas_ui + maybe browser) vs. cloud (browser always; the profile can never narrow the toolset because cloud overrides tools unconditionally after resolution).
2. `OpenHandsAgentSettings.tools` defaults to `[]` → any API caller that forgets to send tools gets an agent with only Finish/Think (OpenHands/software-agent-sdk#3967 was exactly this, surfaced through the profile path).
3. `OpenHandsAgentSettings.enable_sub_agents` is **dead code server-side** — `create_agent()` never reads it; only canvas's client-side list-building honors the toggle.
4. `build_seed_profile` (`profiles/seed.py`) drops `agent_settings.tools` on the floor (profiles have no tools field), so the "behavior-preserving" migration profile silently loses browser/custom tools.

The [#3968](<https://github.com/OpenHands/software-agent-sdk/issues/3968>) review objection (hardcoded literals drift from the real `.name` attributes) and the counter-constraint ("don't touch callers in other repos") are **both satisfiable at once**, because of one observation: `Tool` is `{name, params}` and `create_agent()` only constructs *specs* — resolution to implementations happens later in the runtime via the registry. Default tool **names are a wire contract** (persisted in settings JSON, sent by clients), i.e. data the SDK can own without importing `openhands-tools`.

[...]

Issue #1650, opened 2026-01-08
feat(agent-server): Support plugin loading when starting conversations


jpshackelford:
## Summary

Extend the agent server to support loading plugins when starting conversations. When a `plugin_source` is provided in the `StartConversationRequest`, the agent server will fetch the plugin (using `Plugin.fetch()` from #1645) and load its skills, hooks, and MCP configuration into the conversation context.

**Dependency:** Requires `Plugin.fetch()` from #1645 / PR #1647.

## Context

This is part of the Plugin Directory feature ([OpenHands/OpenHands#12088](https://github.com/OpenHands/OpenHands/issues/12088)).

The key architectural insight is that plugin fetching must happen **inside the sandbox/runtime** (where the agent server runs), not on the app server. This is because:
- Plugins may contain scripts or hooks that need to execute in the sandbox
- Plugin MCP servers need to run inside the sandbox
- Skills may reference files that need to exist in the sandbox filesystem

See [OpenHands/OpenHands#12085 comment](https://github.com/OpenHands/OpenHands/issues/12085#issuecomment-3725059694) for how this fits with other components.

### 1. Extend `StartConversationRequest`

In `openhands-agent-server/openhands/agent_server/models.py`:

```python
class StartConversationRequest(BaseModel):
    # ... existing fields ...

    plugin_source: str | None = Field(
        default=None,
        description="Plugin source to fetch and load. Can be 'github:owner/repo', "
                    "a git URL, or a local path."
    )
    plugin_ref: str | None = Field(
        default=None,
        description="Optional branch, tag, or commit for the plugin."
    )
```

### 2. Add Plugin Loading in `ConversationService.start_conversation()`

In `openhands-agent-server/openhands/agent_server/conversation_service.py`:

```python
from openhands.sdk.plugin import Plugin, PluginFetchError

async def start_conversation(self, request: StartConversationRequest) -> tuple[ConversationInfo, bool]:
    # ... existing code ...

    # Load plugin if specified
    if request.plugin_source:
        try:
            plugin_path = Plugin.fetch(
                request.plugin_source,
                ref=request.plugin_ref
            )
            plugin = Plugin.load(plugin_path)

            # Merge plugin skills into agent context
            request = self._merge_plugin_into_request(request, plugin)
        except PluginFetchError as e:
            logger.error(f"Failed to fetch plugin: {e}")
            raise

    # ... continue with existing conversation start logic ...
```

### 3. Helper Method to Merge Plugin Content

[...]

PR #4030, opened 2026-07-08
feat(agent-server): support deployment context on profile launches

### LINKED ISSUE #4029: Allow client-supplied launch-time agent overrides on the `agent_profile_id` path (carry canvas-only enrichments like `<RUNTIME_SERVICES>`)
### Summary

A conversation launched via `agent_profile_id` is rebuilt **entirely server-side** from the stored `AgentProfile` — the client's `agent_settings` (`request.agent`) is discarded. That's correct for the agent's *configuration*, but it also drops **client-owned, deployment-specific enrichments that have no server-side representation** and that the SDK/resolver therefore cannot reconstruct. There is currently no channel to carry them onto the profile-launch path.

This proposes a narrow addition: let the client supply **launch-time agent overrides** (initially a `system_message_suffix` to *append*, and optionally an `agent_context` overlay) that the server applies to the resolved agent **after** profile resolution — reusing the exact seam where the server already appends its own worktree guidance.

### Concrete manifestation

- **agent-canvas#1571.** The canvas injects a `<RUNTIME_SERVICES>…</RUNTIME_SERVICES>` block into the agent's system prompt (via `agent_context.system_message_suffix`) describing the deployment's live service topology — the Automation backend URL (`/api/automation`), VSCode/browser ports, which services are up. It also adds a canvas-only `canvas_ui` tool. Both are computed **client-side at build/serve time** and have no server representation.
- On the `agent_settings` launch path they ride along fine. On the `agent_profile_id` path they are silently dropped, because the server replaces `request.agent` with the profile-resolved agent.
- This surfaced as a real, deterministic e2e failure: `mock-llm-automation.spec.ts` asserts `<RUNTIME_SERVICES>` is in the LLM system prompt; it fails whenever the home conversation launches from the seeded profile.
- Canvas currently works around it (agent-canvas#1571) by routing the seeded **`default`** profile through the `agent_settings` path instead of `agent_profile_id`. That keeps the enrichments for the common home-launch, but it means **named/custom profiles still lose them**, and it's a client-side special-case that a proper server channel would let us delete.

### Root cause (code anchors, v1.33.0 / `origin/main`)

`conversation_service._start_conversation` — when `agent_profile_id` is set, the resolved agent **overwrites** whatever the client sent:

```python
resolved_agent, launched_agent_profile = await asyncio.to_thread(
    _resolve_agent_from_profile, request.agent_profile_id, self.cipher, mcp_config,
)
request = request.model_copy(update={"agent": resolved_agent})  # client agent_settings discarded
```

`_resolve_agent_from_profile` builds the agent from `stored profile + server-persisted mcp_config + discovered skills + create_agent() defaults` — there is no seam for a client-supplied overlay.

This is by design for config (the profile is the source of truth). The gap is specifically for **enrichments the profile can't and shouldn't persist** — a `<RUNTIME_SERVICES>` block is *dynamic and per-deployment* (ports, backend URLs, live services), so baking it into a stored profile's `system_message_suffix` would be stale/wrong across deployments. It belongs at **launch time**, not at rest.

### Why the #4017/#4018 fixes don't cover this

[#4018](https://github.com/OpenHands/software-agent-sdk/pull/4018) (and [#4017](https://github.com/OpenHands/software-agent-sdk/issues/4017)) made the resolver reproduce the SDK's **default baseline** — exec tools (`create_agent()`), project/user/public skills, `llm.stream=True`. Those all have a server-side or SDK representation, so the server can rebuild them. `<RUNTIME_SERVICES>` / `canvas_ui` do not: they are **client-computed, deployment-specific**, and nothing in `agent_profile_id` resolution can reconstruct them. `defaults-then-override` there = *SDK defaults overridden by stored-profile fields*, all server-side — it is not a client-supplied override channel.

### Proposed fix

Add launch-time overrides to `StartConversationRequest` and apply them to the resolved agent **after** `_resolve_agent_from_profile`, at the same place the server already appends its own suffix.

The precedent already exists — `_prepare_request_workspace` → `_append_worktree_guidance` (`conversation_service.py:239`) appends a `system_message_suffix` to `request.agent.agent_context` *after* profile resolution:

```python
agent = _append_worktree_guidance(request.agent, ...)   # server-side post-resolution suffix append
```

Concretely:

1. Add a field to `StartConversationRequest` — e.g. `agent_launch_overrides` with (at minimum) `system_message_suffix_append: str | None`, and optionally a small `agent_context` overlay (extra tools, extra context fields). Keep it strictly additive/override — it must **not** be able to swap the LLM or replace the resolved config (that's what the profile is for).
2. In `_start_conversation`, after the `agent_profile_id` block sets `request.agent = resolved_agent`, apply the overrides (append the suffix to `agent.agent_context.system_message_suffix`, merge any extra tools) — reusing the `_append_worktree_guidance` pattern.
3. Applies to **both** launch paths uniformly (on the `agent_settings` path it's a no-op since the client already folds these in, but a single code path is cleaner).

### Design considerations / scope guards

- **Additive only.** Overrides append/extend; they can't redefine the agent's LLM, base tools, or skills — the profile stays authoritative for configuration. This keeps the "profile is the source of truth" guarantee from #4017/#4018 intact.
- **Suffix append, not replace**, mirroring `_append_worktree_guidance` (which concatenates onto any existing suffix).
- Consider whether the override belongs in the request body or is registered like `client_tools` / `tool_module_qualnames` (existing client-supplied launch inputs that already survive resolution) — the `canvas_ui` tool half may fit the `client_tools` mechanism, leaving only the `system_message_suffix` append genuinely new.

### Downstream

- **agent-canvas#1571.** With this landed, canvas can carry `<RUNTIME_SERVICES>` on the `agent_profile_id` path and **remove the `default`-profile → `agent_settings` special-case** (so named/custom profiles also get the deployment services block). Tracked against the RUNTIME_SERVICES fix there.

### Relationship

Follow-up to [#4018](https://github.com/OpenHands/software-agent-sdk/pull/4018) / [#4017](https://github.com/OpenHands/software-agent-sdk/issues/4017); epic #3713. Complements the resolver-defaults work by covering the enrichments that are structurally client-only.

---
<sub>Filed via Claude Code at the request of @simonrosenberg.</sub>

simonrosenberg:
HUMAN:

Allow Canvas to launch the real default or named Agent Profile without losing deployment context or Canvas-owned tools.

---

AGENT:

## Why

Agent Canvas needs to launch the selected Agent Profile by ID while adding deployment-specific runtime-service instructions. Canvas-owned UI tools already have a dedicated transport in `client_tools`, so the server only needs a narrow launch-context contract rather than a second default-tool system.

## Summary

- Add `agent_launch_additions.system_message_suffix_append` for deployment context applied after inline-agent or Agent Profile resolution.
- Keep application-owned tools on the existing `client_tools` path; this PR adds no tool-registration defaults and no `/server_info.default_tools` field.
- Preserve the profile contract: `tools: null` uses the standard OpenHands tool set, while any list is exact, including `tools: []` for no profile tools.
- Migrate only a schema-v1, untouched revision-0 OpenHands profile named `default` from the legacy `tools: []` seed to `tools: null`; user-created profiles and explicit empty lists remain unchanged.
- Persist the materialized agent while excluding launch-only context from request metadata, so the appended suffix survives restore exactly once.
- Test launch composition with an actual `ClientToolSpec` to verify client tools remain additive after profile resolution and survive persistence.

### REST API contract changes

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@
+schema AgentLaunchAdditions property system_message_suffix_append optional schema=anyOf=[type="string" maxLength=32768,type="null"]
+schema AgentLaunchAdditions type="object" additionalProperties=false
+schema StartConversationRequest property agent_launch_additions optional schema=anyOf=[AgentLaunchAdditions,type="null"]
```

## Lifecycle

1. A client starts a conversation with either an inline agent or an Agent Profile ID, plus any `client_tools` and deployment context.
2. The agent server resolves the selected agent source.
3. It appends the deployment suffix to the resolved OpenHands agent.
4. The existing client-tool path injects application-owned tools.
5. The server stores the materialized agent, client tool specifications, and profile provenance.

The stored Agent Profile is never modified, and resume does not reapply the launch suffix.

## How to Test

- `178 passed` across the focused profile, launch-addition, router, and store suites.
- Ruff lint and formatting checks pass for all changed Python files.
- Pyright reports zero errors or warnings for all changed Python files.

## Type

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs / chore

## Downstream

- OpenHands/agent-canvas#1650 supplies `canvas_ui_control` through `client_tools` and uses launch additions only for runtime-service context.
- OpenHands/docs#625 documents the resulting architecture and ownership boundaries.

Fixes #4029.

_This PR was updated by an AI agent on behalf of the user._

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:a115652-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-a115652-python \
  ghcr.io/openhands/agent-server:a115652-python
```

```
ghcr.io/openhands/agent-server:a115652-golang-amd64
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-golang-amd64
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-golang-amd64
ghcr.io/openhands/agent-server:a115652-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:a115652-golang-arm64
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-golang-arm64
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-golang-arm64
ghcr.io/openhands/agent-server:a115652-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:a115652-java-amd64
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-java-amd64
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-java-amd64
ghcr.io/openhands/agent-server:a115652-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:a115652-java-arm64
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-java-arm64
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-java-arm64
ghcr.io/openhands/agent-server:a115652-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:a115652-python-amd64
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-python-amd64
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-python-amd64
ghcr.io/openhands/agent-server:a115652-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:a115652-python-arm64
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-python-arm64
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-python-arm64
ghcr.io/openhands/agent-server:a115652-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:a115652-golang
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-golang
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-golang
ghcr.io/openhands/agent-server:a115652-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:a115652-java
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-java
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-java
ghcr.io/openhands/agent-server:a115652-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:a115652-python
ghcr.io/openhands/agent-server:a115652ebc7839fb0057c678f593a2dfa2bf82e4-python
ghcr.io/openhands/agent-server:openhands-launch-time-agent-overrides-python
ghcr.io/openhands/agent-server:a115652-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `a115652-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `a115652-python-amd64`) are also available if needed


enyst (changes requested):
Are we sure this design is the best? Maybe we could consider first if we can detach all/most of the runtime config from the profile or even from the agent context?

simonrosenberg:
@enyst — good call, I went and traced this end-to-end rather than just wiring the field, and I think you're right: runtime/deployment topology doesn't belong on `agent_context`. This PR reuses the worktree-guidance seam, which is itself the exact config/runtime conflation you're pointing at — `_append_worktree_guidance` mutates `agent_context.system_message_suffix` (a persisted, profile-round-tripped *config* field) with per-launch data and then freezes the mutated agent into `meta.json`. This PR just extends that to client-supplied text.

What I found on the way:

**The two "obvious" clean fixes don't apply here:**
- *Server computes it* (the way it computes worktree paths/branch): can't. `<RUNTIME_SERVICES>` is canvas-**layer** topology — the automation backend at `/api/automation`, the `X-Session-API-Key` / `OPENHANDS_AUTOMATION_API_KEY` convention, ingress/frontend ports. The agent-server only knows *itself*; canvas is the authoritative source, so the channel has to accept client-supplied data.
- *Deliver it as an environment event/observation*: there's no environment-event channel in the SDK today — the only agent-injected-at-start event is `SystemPromptEvent`. That'd be net-new infra and cuts against the current grain.

**The channel the framework actually wants isn't `agent_context` — it's the `secret_registry` pattern.** `ConversationState.secret_registry` is a *conversation-scoped* object (not agent config, not persisted into the agent record) that `Agent.get_dynamic_context(state)` merges into the DYNAMIC (uncached) system-prompt tier and renders via a dedicated `<CUSTOM_SECRETS>` `PromptSection`. `<RUNTIME_SERVICES>` is the same shape.

**Proposed redesign (replaces the `agent_launch_overrides` approach in this PR):**
- Add a conversation-scoped `runtime_context` to `StartConversationRequest` → `ConversationState` — **not** on `agent_context`, **not** on the profile, **not** frozen into `meta.json`.
- New DYNAMIC-tier `RuntimeServicesSection` renders it as `<RUNTIME_SERVICES>…`, mirroring `CustomSecretsSection`.
- Because it never touches the agent, it works uniformly on **both** the `agent_profile_id` and `agent_settings` paths — so canvas can delete the agent-canvas#1571 `default`-profile → `agent_settings` special-case, and named/custom profiles get it too.
- Stays a system-role block (DYNAMIC renders as a second system content block), so the `mock-llm-automation.spec.ts` assertion still holds.
- Bonus: worktree guidance could later migrate off the suffix-mutation onto the same mechanism.

(The `canvas_ui` *tool* half is orthogonal and already tracked by #3978 — this is only the prompt-text half.)

Tradeoff to flag: this is a real SDK-core change (agent / prompt-registry / request / state) + a canvas change, versus the minimal PR here that's already green. I think it's worth it — it's the correct layering and it lets canvas drop the special-case. Before I rebuild: does conversation-scoped `runtime_context` → dynamic `<RUNTIME_SERVICES>` section match what you had in mind, or were you picturing pushing it even further off the prompt (e.g. a tool-provided resource the agent queries on demand)?

enyst:
@openhands understand this PR’s goal and the discussion here in top comments.

(clone yourself also smolpaws/smolpaws and look at the skills .agents/skills/ teach-me and show-me)

I want to understand the alternative designs if we move runtime services out of agent context; and their implications. Where else could/should it be, how will it work then for prompt construction (also including at conversation restore time, not only new conversation)

Make an explainer for this or these alternative designs, on enyst.github.io, using those skills as you see fit. Goal is think with me.

Do not modify this PR. Push to main on enyst io the explainer.

simonrosenberg:
@enyst I traced this through current SDK + Canvas main, including create, persistence, cold restore, prompt rendering, plugins/hooks, and both tool-registration paths. My recommendation is: **do not merge the generic `system_message_suffix_append` API as-is. Detach deployment/runtime facts from both `AgentProfile` and `AgentContext`, but keep durable agent instructions in `AgentContext`.**

The important lifecycle split is:

- **Profile / durable agent config:** agent kind, LLM/profile refs, base server tools, MCP refs, disabled skills, condenser/verification, personality/custom instructions. A profile's existing `system_message_suffix` is valid for durable instructions.
- **Conversation runtime binding:** workspace/worktree and the deployment service topology reachable by this conversation. `<RUNTIME_SERVICES>` belongs here.
- **Client UI capability:** `canvas_ui`. This exists only while a compatible Canvas client is attached.
- **Server capability:** registered executable tools, browser usability, MCP providers, module/plugin implementations.

Why #4030 is the wrong boundary even though it works: the override field is excluded, but its arbitrary text is folded into `agent.agent_context.system_message_suffix`; that mutated agent is then persisted in `meta.json` and `base_state.json`, and the rendered `SystemPromptEvent` is persisted too. On cold restore the server reconstructs the stored agent, then `Agent.init_state()` sees the existing `SystemPromptEvent` and deliberately does not render again. So this is not really an ephemeral launch override: it permanently converts deployment data into agent config, with no typed ownership or refresh semantics. The existing worktree-guidance mutation has the same layering debt; it is precedent for behavior, not for the boundary.

I suggest replacing this PR's model with a narrow typed `ConversationRuntimeContext` / `RuntimeServices` input and state field:

1. Carry structured facts (service kind, URL from the agent's network namespace, API prefix/docs URL, auth env-var name), not pre-rendered prompt text or a generic AgentContext overlay.
2. Populate it from a server/deployment provider where possible. Pure agent-server discovery is insufficient—the server cannot infer Canvas ingress/automation topology—but Canvas's launcher already owns that topology and can configure the server. If a request-supplied form is retained for separately deployed clients, make that an explicit authenticated/trusted-deployment policy; do not create a raw system-prompt injection bypass.
3. Add a dedicated dynamic-tier prompt section that renders `<RUNTIME_SERVICES>` from the typed conversation context. This works identically after either `agent_profile_id` or `agent_settings` resolution and keeps profiles portable.
4. Persist the resolved context as a **conversation snapshot**, separate from the agent/profile. On restore, the persisted `SystemPromptEvent` remains the exact-once prompt snapshot. If deployments need rebinding to changed URLs, make that an explicit runtime-context update/versioned event (or a live provider/tool); silently re-reading a profile or appending again would make history nondeterministic.

For `canvas_ui`, use the existing `client_tools` channel rather than adding `extra_tools` or keeping `tool_module_qualnames`. Its current Python executor is already a no-op acknowledgment while Canvas handles the `ActionEvent`, which is exactly `ClientTool`'s contract. `client_tools` is applied after profile resolution, persisted in conversation metadata, re-registered on cold restore, and preserved on forks. It also has defined collision behavior: duplicate request names fail, collisions with server tools fail, and same-name/different-schema reuse fails. By contrast, `tool_module_qualnames` merely imports an implementation; it does not attach the tool to a profile-resolved agent, and global `register_tool` currently warns then last-write-wins on collisions. Canvas should migrate the JSON schema/description to `client_tools` and remove the Python module registration.

Alternatives are weaker:

- A generic append-only or additive launch-augmentation model is cheap but preserves the ownership/trust/persistence bug.
- Pure server discovery cannot see Canvas-owned services; server-side **provisioning** can.
- A runtime-services tool would stay fresh and reduce permanent prompt text, but the model must know to call it and ACP agents do not receive SDK tools. It is a reasonable later redesign, not a parity-safe replacement today.
- Plugins/hooks have the wrong lifecycle and semantics: plugins merge skills/MCP/hooks at first use; SessionStart output is not prompt context, and UserPromptSubmit context is user-message augmentation.
- Removing `<RUNTIME_SERVICES>` in favor of stable env vars or an MCP/tool resource may ultimately be cleaner, but the automation skills currently depend on deterministic prompt discovery; remove it only with an end-to-end replacement.

This boundary gives profile portability and local/cloud parity, prevents deployment data from entering stored profiles, keeps resume deterministic, preserves API compatibility through optional additive fields, and uses the stricter existing collision rules for UI tools. Cost is higher than this green PR (SDK request/state/prompt section + Canvas migration), but it fixes the model rather than adding another generic escape hatch.

simonrosenberg:
@enyst I rebuilt the PR around the detached boundary we discussed.

The generic `AgentLaunchOverrides` / raw suffix append is gone. `ConversationRuntimeContext` is now typed conversation state, persisted as a deterministic snapshot and rendered by a dedicated dynamic `<RUNTIME_SERVICES>` prompt section. It never becomes an `AgentProfile` or `AgentContext` field; for ACP, the conversation context is combined by the shared prompt builder only after `AgentContext` has produced its own inputs. Profile resolution therefore remains authoritative and portable.

The request accepts constrained service facts rather than prompt prose (machine identifier, URL from the agent namespace, API/docs locations, optional auth header/env-var metadata), with validation against extra fields, duplicate/prompt-shaped names, multiline URLs, and invalid auth identifiers. Restore retains the stored context and existing `SystemPromptEvent`, so it does not re-render or pick up silently changed deployment data. Local, remote, profile, legacy-agent, ACP, and fork paths share the same lifecycle.

I left `canvas_ui` out of this new model: it belongs on existing `client_tools`, not in runtime topology or `tool_module_qualnames`. Canvas can migrate that independently while removing the default-profile `agent_settings` special-case.

Local verification: 268 selected profile/service/prompt/ACP/remote/OpenAPI tests pass; pre-commit formatting, lint, typing, import rules, and tool-registration checks pass. CI is now running on `e77ec5ff4`.

simonrosenberg:
@enyst I simplified the implementation after weighing the migration cost more heavily.

The final diff is now **two production files plus one focused test file**, not a new conversation-runtime subsystem. `runtime_services` is a narrow typed launch input; after profile/direct-agent resolution the server renders it through the existing suffix seam into the resolved conversation agent. The stored Agent Profile remains untouched, while the existing resolved-agent persistence gives deterministic cold restore and exact-once prompt behavior for both OpenHands and ACP without new state/prompt/factory plumbing.

This keeps the important boundary—deployment data cannot be saved into a profile and the client cannot submit arbitrary system-prompt prose—while accepting the pragmatic compromise that the rendered launch snapshot lives in the conversation's resolved `AgentContext`.

Canvas's `canvas_ui` capability remains orthogonal: the follow-up uses existing `client_tools` for OpenHands profile launches. ACP subprocesses own their tools and reject SDK custom tools, so they keep their existing tool behavior.

Local verification: 160 profile/service/OpenAPI tests pass; pre-commit and Python API compatibility checks pass. Head: `a267c1da9`.

simonrosenberg:
Canvas follow-up: OpenHands/agent-canvas#1650. It removes the local default-profile workaround, sends typed runtime services on profile launches, and uses client_tools for OpenHands canvas_ui. It is draft and explicitly depends on this SDK change.

simonrosenberg:
@enyst — follow-up after tracing the alternatives and reconsidering the consumer boundary: I changed the PR to remove the SDK-level `RuntimeService(s)` model.

The boundary is now:

- `AgentProfile` owns durable, portable agent configuration.
- `agent_launch_overrides.system_message_suffix_append` is a trusted, append-only launch mechanism applied only after profile/inline-agent resolution.
- The resolved conversation agent persists the rendered result, so cold resume reproduces the original prompt exactly once; the launch field itself is not persisted or reapplied, and the profile is never mutated.
- Canvas owns `RuntimeServicesInfo` and renders `<RUNTIME_SERVICES>` itself. Another SDK consumer can use the mechanism without adopting Canvas's topology, auth conventions, or class shape.
- `canvas_ui` remains a UI capability and uses `client_tools`; ACP subprocess tools remain server/runtime-owned.

This keeps deployment data out of durable profile configuration while avoiding a Canvas-tailored runtime-context type in the SDK. It does still intentionally freeze the rendered launch context into the conversation's resolved agent snapshot: that is the resume contract, not a claim that the text is durable agent configuration.

The weaker alternatives were: server discovery cannot observe Canvas-owned services; `client_tools` cannot represent prompt context; a tool makes required topology available only on demand; plugins/hooks require matching runtime code; and a new typed conversation-state/prompt-section subsystem adds broad lifecycle plumbing while still needing a consumer-specific schema or generic renderer. A replacement-style override would also be too permissive, so this API is additive-only, size-limited, rejects unknown fields, and is documented as trusted control-plane input rather than end-user input.

The final SDK diff is back to 2 production files plus 1 focused test file. Canvas #1650 contains the consumer-side change and removes the `default`-profile workaround.

simonrosenberg:
Split the independent client-tool lifecycle bug into #4078. This PR now contains only the trusted launch-augmentation API and its tests (2 production files + 1 test file). The Canvas integration depends on both PRs, but they can now be reviewed, merged, or reverted independently.

openhands-sdk/openhands/sdk/conversation/request.py
```diff
@@ -74,6 +75,21 @@ def create_message(self) -> Message:
...
+class AgentLaunchOverrides(BaseModel):
+    """Append trusted deployment context after agent resolution."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    system_message_suffix_append: str | None = Field(
+        default=None,
+        max_length=32768,
+        description=(
+            "Trusted deployment-controlled text appended to the resolved agent's "
```
enyst: It’s coming from the outside, in the request, right? Idk, maybe we shouldn’t call it “trusted”

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -745,6 +749,17 @@ async def _start_conversation(
             )
             request = request.model_copy(update={"agent": resolved_agent})
 
+        addition = (
+            request.agent_launch_overrides.system_message_suffix_append.strip()
+            if request.agent_launch_overrides
```
enyst: I think maybe if it’s an addition, we could call it addition, not override… I’m following up

enyst (approved):
Simon, please correct if I misunderstand: we are adding some lines to the system prompt, coming from canvas. They go in as suffix, as part of the suffix… 🤔 And they’re going to be saved in the system prompt event on disk; on restore, these lines cannot change.

I think this is cleaner than override! I also think maybe it isn’t ideal because then we will need to figure out how to… either:
- replace them, or
- tell the agent we override them with newer instructions

when we want to move a conversation from a runtime to another… I mean, it seems to me that if they’re coming from a running canvas, that’s a particular backend; so freezing them in the system prompt might give us headaches later. But maybe I’m wrong! And I like the simplicity here. Up to you

I left a few tiny nits on naming; I think we will be happier - and so will be the agent - if we name things for what they are 🙏

simonrosenberg:
@enyst maybe we should 1. remove this completely from SDK 2. have clients add this data in the first user message?  no need for special wiring

simonrosenberg:
@enyst — follow-up after thinking through the simpler client-owned path: I think we should remove this PR's launch-override channel.

The high-level ownership should be:

- `AgentProfile`: durable, portable agent configuration.
- Conversation transcript: user intent and historical context.
- Runtime binding: mutable Canvas/deployment facts for the currently attached runtime.
- Tools/resources/environment: live capabilities.

`agent_launch_overrides.system_message_suffix_append` is uncomfortable for two reasons. First, it lets a conversation-creation request inject arbitrary system-role text; append-only validation and a length limit constrain its shape, but do not make the caller trusted. Second, it folds the Canvas topology into the resolved agent and persisted `SystemPromptEvent`, so a restored conversation permanently retains the old binding and a later user instruction has lower authority than that stale system context.

Having Canvas attach `<RUNTIME_SERVICES>` to the first user turn is a better boundary: it leaves the profile authoritative, works uniformly for profile/direct-agent/ACP launches, needs no launch-specific SDK mechanism, and lets a later same-role context block supersede the old binding after a runtime move.

I would not literally concatenate it into the visible user string, though. Raw `initial_message.content` is also used for chat rendering, autotitling, UserPromptSubmit hooks, and skill-trigger matching. The runtime block itself contains phrases such as “Automation backend,” so it could activate automation skills for unrelated tasks. Empty-conversation and attachment flows also have no creation-time message, so Canvas must intercept the first later send anyway.

The SDK already has the useful distinction internally: `MessageEvent.extended_content` reaches the model while remaining separate from user-authored content. The clean small API, if we want to avoid raw concatenation, would be a generic message-envelope field such as `client_context` on `SendMessageRequest` / `initial_message`, mapped to `MessageEvent.extended_content`:

```text
user turn
├── content: actual user-authored task
└── client_context: Canvas-rendered <RUNTIME_SERVICES> block
```

That is not a profile or system-prompt override. Skill matching, hooks, titles, and UI continue to see the real user content; the model receives both.

I would also define this as the first user turn **per runtime binding**, not merely the first-ever conversation turn. Give the binding an ID/version; after relocation, Canvas attaches a new block on the next real user turn stating that it replaces the previous binding. We keep history immutable while making rebinding explicit.

If the agent must cold-resume and act before another user message, then neither first-turn context nor this PR is sufficient. At that point topology needs to be live runtime state—ideally standardized environment variables or a runtime-services tool/resource. That seems like the actual long-term architecture.

So my recommendation is: close/remove the generic system-suffix API; use client-owned first-turn context as the tactical solution, with a small generic message-context field only if we want it hidden from user-message semantics.

simonrosenberg:
Superseded by #4110, which replaces the generic launch-time system-suffix augmentation with client-owned context carried on the initial or next user message and persisted as MessageEvent.extended_content. Canvas adoption: OpenHands/agent-canvas#1782. Documentation: OpenHands/docs#622. The independent conversation-scoped client-tools work remains in #4078.

simonrosenberg:
@enyst honestly i dont know... maybe still worth merging this one so we have a "cleaner" solution now and we can continue looking for a better one

openhands-sdk/openhands/sdk/conversation/request.py
simonrosenberg: Addressed in 747f3ff0d: removed “trusted” from the API name and descriptions. The request now exposes deployment-controlled launch additions.

openhands-agent-server/openhands/agent_server/conversation_service.py
simonrosenberg: Addressed in 747f3ff0d: renamed the model and request field to AgentLaunchAdditions / agent_launch_additions. The implementation remains additive-only.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -89,19 +89,41 @@ def _append_worktree_guidance(
...
 
 
+def _append_agent_tools(agent: AgentBase, additions: list[Tool]) -> AgentBase:
+    tools_by_name = {tool.name: tool for tool in agent.tools}
+    new_tools: list[Tool] = []
+    for tool in additions:
+        existing = tools_by_name.get(tool.name)
+        if existing is not None:
+            if existing != tool:
+                raise ValueError(
```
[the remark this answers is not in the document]
simonrosenberg: Fixed in 1b5bab83f: launch-tool conflicts now use a dedicated error mapped to HTTP 409, with router coverage.

simonrosenberg:
Architecture update in `30a50d812`: removed launch-time tool appending. Tool implementations now declare whether they are runtime defaults; the agent-server registers runtime modules before resolving a profile and adds usable runtime defaults only for OpenHands profiles with `tools: null`. Explicit `[]` and explicit tool lists remain authoritative. `agent_launch_additions` now carries deployment prompt context only.

simonrosenberg:
Architecture simplification in `f698a22ab`: the SDK `runtime_default` marker and registry bookkeeping are gone. The agent-server now owns default selection through `extra_default_tools`, resolves `tools: null` before agent construction, and publishes the same resolved deployment list through `/server_info.default_tools`. Explicit empty and non-empty profile lists remain authoritative, and request-level module imports cannot alter startup defaults. Canvas consumes this contract in OpenHands/agent-canvas#1650.

simonrosenberg:
Migration follow-up in `f0555d6ee` and `1f6307f20`: the risk is real because an older SDK represented global default tools as `[]`, and the profile seed copied that value. Profile schema v2 now migrates only a v1, untouched revision-0 OpenHands profile named `default` from `[]` to `null`. New/unversioned/v2 `[]`, revised v1 profiles, and custom v1 profiles remain explicitly bare. A cross-repository launch smoke test with the Canvas module produced legacy/default → `terminal,file_editor,task_tracker,canvas_ui`, explicit `[]` → no tools, and `[terminal]` → only terminal.

simonrosenberg:
Superseding the earlier extra_default_tools proposal: default membership now lives beside tool registration via register_tool(..., default=True). The agent-server composes canonical defaults, delegation/browser policy, and usable registered defaults; it has no Canvas-specific name or deployment tool list. Canvas only preloads its module. The cross-repo smoke test resolves the standard tools plus canvas_ui through this registration path.

PR #4035, opened 2026-07-08
fix(conversation): read persisted base_state.json as UTF-8 on resume

### LINKED ISSUE #4034: Windows: UnicodeDecodeError on conversation resume — base_state.json read with OS default codec instead of UTF-8
## Summary

On Windows, resuming a persisted conversation crashes with
`UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f ...` whenever
`base_state.json` contains any non-ASCII byte (e.g. Cyrillic/accented
characters in agent instructions, tool descriptions, or message content).

`base_state.json` is **written as UTF-8** by `LocalFileStore.write`
(`openhands-sdk/openhands/sdk/io/local.py` — `open(..., "w", encoding="utf-8")`),
but `LocalConversation._recover_persisted_client_tools` reads it back **without
an encoding**, bypassing the FileStore:

```python
# openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py  (main, ~line 498)
data = json.loads(base_path.read_text())        # <-- no encoding=
```

`Path.read_text()` with `encoding=None` uses `locale.getpreferredencoding()`,
which on Windows is **cp1252**, not UTF-8. So the SDK reads back a file it wrote
as UTF-8 using a different codec and raises. This is a write-UTF-8 /
read-locale asymmetry, not a user misconfiguration.

## Impact

- Deterministic crash on Windows for any conversation whose persisted state
  contains non-ASCII content, **on resume** (create-or-resume path). The first
  turn of a session is fine; the failure appears on the next turn/reload once
  `base_state.json` exists.
- Because it only triggers on resume, it looks intermittent and is easy to
  misdiagnose — the crash often surfaces downstream as a secondary error
  (e.g. a broken pipe / `[Errno 22]` when writing to the failed runner),
  masking the real cause.

## Environment

- `openhands-sdk` 1.31.0 (also confirmed present on `main`)
- Windows 11, CPython 3.12 (default file codec cp1252)
- Persistence: `LocalFileStore`, `persistence_dir` + `conversation_id`

## Reproduction

```python
import locale, uuid, tempfile
from pathlib import Path
from openhands.sdk import LLM, Agent, Conversation
from openhands.sdk.tool import Tool
from openhands.sdk.tool.client_tool import ClientToolSpec

def mk():
    return Agent(llm=LLM(model="gpt-4o", usage_id="t"), tools=[Tool(name="terminal")])

tmp = Path(tempfile.mkdtemp())
# non-ASCII content lands in base_state.json as UTF-8
specs = [ClientToolSpec(name="persist_note",
                        description="тепловая матрица — аномалии ±0.5 A",
                        parameters={"type": "object", "properties": {}})]
cid = uuid.uuid4()
Conversation(agent=mk(), workspace=str(tmp/"ws"), persistence_dir=str(tmp/"persist"),
             conversation_id=cid, client_tools=specs, delete_on_close=False).close()

convo = Conversation(agent=mk(), workspace=str(tmp/"ws2"), persistence_dir=str(tmp/"persist2"),
                     conversation_id=uuid.uuid4(), delete_on_close=False)
# simulate a non-UTF-8 default locale (as on Windows)
locale.getpreferredencoding = lambda *a, **k: "ascii"
convo._recover_persisted_client_tools(str(tmp/"persist"), cid)   # raises UnicodeDecodeError
```

On a UTF-8-default host (typical Linux CI) the bare read happens to work, which
is why this has gone unnoticed in CI.

## Root cause

`_recover_persisted_client_tools` constructs `base_path` directly and calls
`read_text()` instead of going through the FileStore (`fs.read()`), which is
already UTF-8-correct. The one read path that bypasses the abstraction is the
one that crashes.

## Fix

Pass `encoding="utf-8"` on the persisted-state read so it matches the UTF-8
write. PR opened: adds the encoding + a regression test that simulates a
non-UTF-8 locale (deterministic on every platform).

## Related (same class, not in the PR)

Several other stores also read persisted JSON with a bare `read_text()`
(`profiles/agent_profile_store.py`, `llm/llm_profile_store.py`,
`extensions/installation/metadata.py`, `agent/acp_agent.py`). Unlike
`base_state.json`, those files are also **written** with the platform default
codec (`NamedTemporaryFile(mode="w")` / bare `write_text()`), so a read-only
UTF-8 change there would introduce a *new* write/read asymmetry. Making them
UTF-8 correctly needs a coordinated read+write pass — happy to follow up in a
separate PR if you'd like.

shneydermanmm:
## What

Read the persisted `base_state.json` with `encoding="utf-8"` on resume.

## Why

`base_state.json` is **written** as UTF-8 by `LocalFileStore.write`
(`io/local.py`: `open(..., "w", encoding="utf-8")`), but
`LocalConversation._recover_persisted_client_tools` reads it back with a bare
`Path.read_text()`, which uses the OS default codec
(`locale.getpreferredencoding()` — cp1252 on Windows).

So on Windows, resuming a conversation whose state contains any non-ASCII byte
crashes in the resume path:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f ... maps to <undefined>
```

It only manifests on resume (once the state file exists), so it reads as
intermittent and is easy to misattribute to a downstream error. This is a
write-UTF-8 / read-locale asymmetry — the fix makes the read match the write.

Fixes #4034.

## Change

```diff
- data = json.loads(base_path.read_text())
+ data = json.loads(base_path.read_text(encoding="utf-8"))
```

Plus a regression test in the existing
`tests/sdk/conversation/local/test_client_tools_persistence.py`. It persists a
conversation with non-ASCII content, then recovers it while simulating a
non-UTF-8 default locale (`monkeypatch` of `locale.getpreferredencoding`), so
the failure is **deterministic on every platform** — it fails before this
change (`UnicodeDecodeError`) and passes after.

## Scope

Intentionally minimal — only the read of `base_state.json`, whose write side is
already UTF-8 (via `FileStore`). Other stores also read persisted JSON with a
bare `read_text()`, but they write with the platform default codec too, so a
read-only UTF-8 change there would introduce a *new* asymmetry; those need a
coordinated read+write pass. Called out in #4034 as a possible follow-up.

### Alternative considered

Route this read through `FileStore.read()` (already UTF-8) instead of
constructing the path and calling `read_text()` directly. That's the cleaner
long-term shape but touches more surface; happy to switch to it if preferred.

## Testing

- New regression test fails before the change, passes after (verified against
  1.31.0 on Windows/cp1252).
- No behavior change on UTF-8-default hosts; purely widens correctness on
  Windows / non-UTF-8 locales. Reader-side only.


neubig (approved):
LGTM, thank you!

PR #4042, opened 2026-07-09
ci: add release security-scan

smolpaws:
HUMAN:

I asked smolpaws to add a fourth release-PR check for security: an approval-drift (TOCTOU) guard and a supply-chain dependency diff. Starting with these two deterministic checks; more to follow.

---

AGENT:

## Why

We run three labeled suites on release PRs today — `integration-test`, `behavior-test`, `test-examples`. This adds a **fourth**, `security-scan`, for a gap those don't cover: the integrity of *what actually ships* versus *what a human reviewed*.

As agents (ours included) get faster and some PRs start merging with lighter or automated review, a real window opens up:

- a PR is approved, then commits land **after** the approval and still get merged;
- a PR is merged with **no human approval** at all;
- a review/QA agent runs while an **injected comment** sits on the PR, and whatever it did ends up on `main`.

This is a classic **TOCTOU** (time-of-check → time-of-use) gap. Cutting a release from that `main` inherits it. The right place to catch it is the release diff — *last release tag → this PR* — audited as one unit.

## Summary

- Adds `.github/workflows/security-scan.yml`: a new opt-in job gated on a `security-scan` label (plus `workflow_dispatch`), mirroring the existing labeled-suite pattern. Posts one PR comment and fails on any blocking finding.
- Adds `check_approval_drift.py` (TOCTOU): for every PR merged since the last `v*` tag, flags those merged with **no human approval**, or where commits landed **after** the last human approval (reviewed sha ≠ merged sha). Trusted bots (e.g. `dependabot[bot]`) are configurable and exempt.
- Adds `check_dependency_diff.py` (supply chain): diffs `uv.lock`, checks new/bumped packages against **OSV** for known vulns, and flags any resolved from a **non-PyPI source** (git/url/alt-index — dependency-confusion surface).
- Adds `security_scan_common.py` shared helpers. **Deterministic by design**: the scanners only read git history and query read-only APIs (GitHub REST, OSV); they never install or execute PR code, so the gate itself is not an injection target — hence plain `pull_request` (not `pull_request_target`) and no secrets beyond the default `GITHUB_TOKEN`. An **OSV lookup failure blocks** (fail-closed: a release diff that can't be verified against known vulns shouldn't ship — a maintainer re-runs or clears it), as does any non-PyPI / mirror / unrecognized source. Un-auditable PRs in the approval-drift check (transient GitHub API errors) remain warnings so a flaky call doesn't wedge a release.

## Issue Number

N/A (follow-up to the OpenHands/release-actions#18 freeze-on-ready discussion).

## How to Test

Run either scanner locally against real history:

```bash
# Supply-chain diff (OSV + non-PyPI source), stdlib only
SECURITY_SCAN_BASELINE_TAG=v1.28.0 python .github/scripts/check_dependency_diff.py

# Approval drift (needs a token to read reviews)
GITHUB_TOKEN=<token> GITHUB_REPOSITORY=OpenHands/software-agent-sdk \
  SECURITY_SCAN_BASELINE_TAG=v1.31.0 python .github/scripts/check_approval_drift.py
```

Or add the `security-scan` label to a PR / trigger via `workflow_dispatch`.

Verified locally against this repo's real history (evidence):
- `check_dependency_diff.py` over `v1.28.0..HEAD` correctly surfaced a known-vulnerable transitive `starlette==1.0.1` with real advisory IDs (`GHSA-82w8-qh3p-5jfq, GHSA-jp82-jpqv-5vv3, GHSA-wqp7-x3pw-xc5r, GHSA-x746-7m8f-x49c, PYSEC-2026-248, PYSEC-2026-249`) and listed 9 added / 11 bumped deps; over `v1.32.0..HEAD` it reports ✅ no findings (only the SDK's own version bumps).
- `check_approval_drift.py` over `v1.31.0..HEAD` flagged a real `changed-after-approval` merge (#4018: last approval by @neubig on `1986d9562`, merged head `780de42de`), exit code 1; over `v1.33.0..HEAD` it reports ✅ no findings across 6 PRs.
- `ruff check` and `ruff format` clean on all three scripts.

## Type

- [x] Feature (CI)

## Notes

- Scoped to the **two** deterministic checks we agreed to start with. Natural follow-ups (separate PRs): a prompt-injection / agent-influence scan of PR content + surrounding comments; a general static "weirdness" pass (egress / dangerous-exec / CI-workflow tampering); and running the scanners from a trusted ref rather than PR head for extra hardening.
- Not yet dogfooded on a live release PR — suggest enabling on the next release cut (or a follow-up integration test) before relying on it as a gate.

*Opened by 🐾 smolpaws on Engel's behalf.*

### DIFF
--- a/.github/scripts/check_approval_drift.py
+++ b/.github/scripts/check_approval_drift.py
@@ -1,0 +1,204 @@
+"""Release security check #1 — approval drift (a TOCTOU guard).
+
+The window we care about: a human reviews a PR (time-of-check), but what is
+actually merged and shipped (time-of-use) is not what they reviewed — because
+commits were pushed after approval, or the PR was merged with no human approval
+at all (e.g. auto-merge, or an agent acting on an injected review comment).
+
+For every PR merged into this line of history since the last release tag, this
+compares the PR's merged head against its last *human* APPROVED review:
+
+  * merged with **no human approval**            -> blocking
+  * commits pushed **after** the last approval    -> blocking (the reviewed
+    diff is not the shipped diff)
+
+Deterministic: reads ``git log`` for the merged-PR list and the read-only
+GitHub REST API for reviews. It never runs the PR's code.
+
+Env:
+  GITHUB_TOKEN       (recommended; raises rate limits, reads private repos)
+  GITHUB_REPOSITORY  (owner/name; falls back to the origin remote)
+  SECURITY_SCAN_BASELINE_TAG  (optional; overrides the auto-detected last tag)
+  SECURITY_SCAN_TRUSTED_BOTS  (optional CSV of bot logins whose merges are
+                               allowed without human approval, e.g. dependabot)
+"""
+
+from __future__ import annotations
+
+import os
+import re
+import sys
+
+from security_scan_common import (
+    Report,
+    github_request,
+    last_release_tag,
+    resolve_repo,
+    run_git,
+)
+
+
+# Squash-merge subjects on this repo end with "(#1234)"; that is the PR number.
+_PR_NUM_RE = re.compile(r"\(#(\d+)\)\s*$")
+
+# GitHub marks these association values as coming from outside the project.
+_OUTSIDE_ASSOC = {"NONE", "FIRST_TIMER", "FIRST_TIME_CONTRIBUTOR", "MANNEQUIN"}
+
+
+def merged_pr_numbers(baseline: str) -> list[tuple[int, str]]:
+    """(pr_number, subject) for commits in ``baseline..HEAD`` that name a PR."""
+    log = run_git(
+        "log", "--first-parent", "--pretty=format:%H%x1f%s", f"{baseline}..HEAD"
+    )
+    found: list[tuple[int, str]] = []
+    for line in log.splitlines():
+        if "\x1f" not in line:
+            continue
+        _, subject = line.split("\x1f", 1)
+        match = _PR_NUM_RE.search(subject)
+        if match:
+            found.append((int(match.group(1)), subject))
+    return found
+
+
+def _is_bot(user: dict[str, object]) -> bool:
+    return user.get("type") == "Bot" or str(user.get("login", "")).endswith("[bot]")
+
+
+def audit_pr(
+    repo: str, number: int, token: str | None, trusted_bots: set[str]
+) -> tuple[str, str | None]:
+    """Return (status, detail) for one PR.
+
+    status is one of: "ok", "no-human-approval", "changed-after-approval",
+    "trusted-bot", "error".
+    """
+    try:
+        pr = github_request(f"/repos/{repo}/pulls/{number}", token)
+        reviews = github_request(
+            f"/repos/{repo}/pulls/{number}/reviews?per_page=100", token
+        )
+    except Exception as exc:  # noqa: BLE001 - report, do not crash the gate
+        return "error", str(exc)
+
+    merge_sha = pr.get("merge_commit_sha")
+    head_sha = (pr.get("head") or {}).get("sha")
+    author = (pr.get("user") or {}).get("login", "?")
+
+    approvals = [
+        r
+        for r in reviews
+        if r.get("state") == "APPROVED"
+        and not _is_bot(r.get("user") or {})
+        and (r.get("author_association") or "NONE") not in _OUTSIDE_ASSOC
+    ]
+
+    if not approvals:
+        merged_by = (pr.get("merged_by") or {})
+        if _is_bot(merged_by) and merged_by.get("login") in trusted_bots:
+            return "trusted-bot", f"merged by trusted bot {merged_by.get('login')}"
+        return (
+            "no-human-approval",
+            f"author @{author}; merged={_short(merge_sha)}; no human APPROVED review",
+        )
+
+    # The reviewed commit is the head each approval was submitted against.
+    approved_shas = {r.get("commit_id") for r in approvals if r.get("commit_id")}
+    reviewed_head = head_sha in approved_shas
+    if not reviewed_head:
+        latest = approvals[-1]
+        reviewer = (latest.get("user") or {}).get("login", "?")
+        approved_on = _short(latest.get("commit_id"))
+        return (
+            "changed-after-approval",
+            (
+                f"last approval by @{reviewer} was on {approved_on}, but merged "
+                f"head was {_short(head_sha)} — commits landed after review"
+            ),
+        )
+
+    return "ok", None
+
+
+def _short(sha: object) -> str:
+    return str(sha)[:9] if sha else "?"
+
+
+def main() -> int:
+    token = os.environ.get("GITHUB_TOKEN") or os.environ.get("GH_TOKEN")
+    repo = resolve_repo()
+    baseline = os.environ.get("SECURITY_SCAN_BASELINE_TAG") or last_release_tag()
+    trusted_bots = {
+        b.strip()
+        for b in os.environ.get("SECURITY_SCAN_TRUSTED_BOTS", "").split(",")
+        if b.strip()
+    }
+
+    report = Report("🔒 Approval drift (time-of-check vs time-of-use)")
+
+    if not baseline:
+        report.add("No previous `v*` release tag found; nothing to compare against.")
+        print(report.render())
+        return 0
+
+    try:
+        run_git("rev-parse", "--verify", f"{baseline}^{{commit}}")
+    except Exception:  # noqa: BLE001
+        report.warn("baseline unavailable")
+        report.add(
+            f"Baseline tag `{baseline}` is not present locally "
+            "(fetch tags with `fetch-depth: 0`). Skipping."
+        )
+        print(report.render())
+        return 0
+
+    prs = merged_pr_numbers(baseline)
+    report.add(f"Baseline: `{baseline}` — {len(prs)} merged PR(s) in range.")
+    report.add("")
+
+    blocking_rows: list[str] = []
+    ok = 0
+    errors = 0
+    for number, _subject in prs:
+        status, detail = audit_pr(repo, number, token, trusted_bots)
+        if status == "ok":
+            ok += 1
+        elif status == "trusted-bot":
+            ok += 1
+        elif status == "error":
+            errors += 1
+            report.warn(f"could not audit #{number}")
+        else:
+            blocking_rows.append(f"| #{number} | `{status}` | {detail} |")
+            report.block(f"#{number}: {status}")
+
+    if blocking_rows:
+        report.add("| PR | finding | detail |")
+        report.add("| -- | ------- | ------ |")
+        report.lines.extend(blocking_rows)
+        report.add("")
+    report.add(
+        f"Audited {len(prs)} PR(s): {ok} clean, {len(blocking_rows)} flagged, "
+        f"{errors} un-auditable."
+    )
+    if errors:
+        report.add(
+            "\n> Un-auditable PRs are reported as warnings, not blockers, so a "
+            "transient API error does not wedge a release."
+        )
+
+    sys.stdout.write(report.render())
+    _emit_summary_file(report)
+    return 1 if report.blocking else 0
+
+
+def _emit_summary_file(report: Report) -> None:
+    out = os.environ.get("SECURITY_SCAN_OUTPUT")
+    if out:
+        with open(out, "a", encoding="utf-8") as handle:
+            handle.write(report.render())
+            handle.write("\n")
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/.github/scripts/check_approval_drift.py
+++ b/.github/scripts/check_approval_drift.py
@@ -94,7 +94,7 @@ def audit_pr(
     ]
 
     if not approvals:
-        merged_by = (pr.get("merged_by") or {})
+        merged_by = pr.get("merged_by") or {}
         if _is_bot(merged_by) and merged_by.get("login") in trusted_bots:
             return "trusted-bot", f"merged by trusted bot {merged_by.get('login')}"
         return (
--- a/.github/scripts/check_approval_drift.py
+++ b/.github/scripts/check_approval_drift.py
@@ -32,6 +32,7 @@ import sys
 from security_scan_common import (
     Report,
     github_request,
+    github_request_all,
     last_release_tag,
     resolve_repo,
     run_git,
@@ -85,7 +86,11 @@ def audit_pr(
     """
     try:
         pr = github_request(f"/repos/{repo}/pulls/{number}", token)
-        reviews = github_request(
+        # Paginate: the reviews endpoint returns every review (COMMENTED too),
+        # so a heavily-discussed PR can exceed one page. A human APPROVED past
+        # the first page must not be missed, or the PR is misclassified as
+        # unapproved and blocks a clean release.
+        reviews = github_request_all(
             f"/repos/{repo}/pulls/{number}/reviews?per_page=100", token
         )
     except Exception as exc:  # noqa: BLE001 - report, do not crash the gate
--- a/.github/scripts/check_approval_drift.py
+++ b/.github/scripts/check_approval_drift.py
@@ -45,20 +45,30 @@ _PR_NUM_RE = re.compile(r"\(#(\d+)\)\s*$")
 _OUTSIDE_ASSOC = {"NONE", "FIRST_TIMER", "FIRST_TIME_CONTRIBUTOR", "MANNEQUIN"}
 
 
-def merged_pr_numbers(baseline: str) -> list[tuple[int, str]]:
-    """(pr_number, subject) for commits in ``baseline..HEAD`` that name a PR."""
+def merged_pr_numbers(baseline: str) -> tuple[list[tuple[int, str]], list[str]]:
+    """PRs named in ``baseline..HEAD`` plus the subjects that map to none.
+
+    Returns ``(matched, unmapped)`` where ``matched`` is ``(pr_number,
+    subject)`` for first-parent commits whose subject ends in ``(#N)``, and
+    ``unmapped`` is the subjects that do not — so the caller can surface the
+    blind spot (a merge that skipped the squash convention is exactly the one
+    an approval-drift guard must not silently ignore) instead of dropping it.
+    """
     log = run_git(
         "log", "--first-parent", "--pretty=format:%H%x1f%s", f"{baseline}..HEAD"
     )
-    found: list[tuple[int, str]] = []
+    matched: list[tuple[int, str]] = []
+    unmapped: list[str] = []
     for line in log.splitlines():
         if "\x1f" not in line:
             continue
         _, subject = line.split("\x1f", 1)
         match = _PR_NUM_RE.search(subject)
         if match:
-            found.append((int(match.group(1)), subject))
-    return found
+            matched.append((int(match.group(1)), subject))
+        else:
+            unmapped.append(subject)
+    return matched, unmapped
 
 
 def _is_bot(user: dict[str, object]) -> bool:
@@ -152,8 +162,19 @@ def main() -> int:
         print(report.render())
         return 0
 
-    prs = merged_pr_numbers(baseline)
+    prs, unmapped = merged_pr_numbers(baseline)
     report.add(f"Baseline: `{baseline}` — {len(prs)} merged PR(s) in range.")
+    if unmapped:
+        report.warn(f"{len(unmapped)} commit(s) not mapped to a PR")
+        report.add(
+            f"> ⚠️ {len(unmapped)} first-parent commit(s) in range did not match "
+            "the `(#N)` squash convention and were **not** audited (a merge/rebase "
+            "merge, or a lost `(#N)` suffix). These are a blind spot — inspect them:"
+        )
+        for subject in unmapped[:20]:
+            report.add(f">   - {subject}")
+        if len(unmapped) > 20:
+            report.add(f">   - …and {len(unmapped) - 20} more")
     report.add("")
 
     blocking_rows: list[str] = []
--- a/.github/scripts/check_dependency_diff.py
+++ b/.github/scripts/check_dependency_diff.py
@@ -1,0 +1,191 @@
+"""Release security check #2 — supply-chain dependency diff.
+
+Compares the resolved dependency set (``uv.lock``) between the last release tag
+and HEAD, then for everything new or bumped:
+
+  * looks it up in **OSV** for known vulnerabilities (blocking on a hit), and
+  * flags packages resolved from a **non-PyPI source** — git URLs, direct
+    URLs, or alternative indexes — which is the classic supply-chain and
+    dependency-confusion surface (blocking; a release should pin to the
+    trusted registry).
+
+Deterministic: parses the two lockfile revisions from git and queries the
+public OSV API. It never installs or executes any dependency.
+
+Env:
+  SECURITY_SCAN_BASELINE_TAG  (optional; overrides the auto-detected last tag)
+  SECURITY_SCAN_OUTPUT        (optional; append the rendered report here too)
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+import tomllib
+
+from security_scan_common import (
+    Report,
+    last_release_tag,
+    osv_query_batch,
+    run_git,
+)
+
+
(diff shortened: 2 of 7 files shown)


enyst:
@openhands Do a /codereview on this PR. Post your feedback as a review with gh api.

.github/scripts/check_dependency_diff.py
```diff
@@ -0,0 +1,191 @@
...
+        return f"registry:{source['registry']}"
+    return str(source)
+
+
+def _is_trusted_registry(source: object) -> bool:
+    """A first-party workspace member or a plain PyPI registry pin is trusted."""
+    if not isinstance(source, dict):
+        return False
+    if "registry" in source:
+        return "pypi.org" in str(source["registry"])
```
enyst: **Robustness / potential release-blocking false positive.** Trust here hinges on `"pypi.org" in str(source["registry"])`. Verified against this repo's current `uv.lock` (409/413 packages are `{registry: "https://pypi.org/simple"}`, 4 are `{editable}`), so it's correct today. But: (1) uv can omit the `source` table or emit a mirror/alt-index for default packages depending on `[tool.uv]` config — those would fall through to the non-registry branch and get `report.block(...)`, failing the release gate on ordinary PyPI deps; (2) the substring check treats `https://pypi.org.evil.example/simple` as trusted. Prefer an exact host/URL allowlist, and decide deliberately whether an empty/missing source means *default = trusted* or *unknown = warn* (rather than a hard block).

.github/scripts/check_approval_drift.py
```diff
@@ -0,0 +1,204 @@
...
+_PR_NUM_RE = re.compile(r"\(#(\d+)\)\s*$")
+
+# GitHub marks these association values as coming from outside the project.
+_OUTSIDE_ASSOC = {"NONE", "FIRST_TIMER", "FIRST_TIME_CONTRIBUTOR", "MANNEQUIN"}
+
+
+def merged_pr_numbers(baseline: str) -> list[tuple[int, str]]:
+    """(pr_number, subject) for commits in ``baseline..HEAD`` that name a PR."""
+    log = run_git(
+        "log", "--first-parent", "--pretty=format:%H%x1f%s", f"{baseline}..HEAD"
```
enyst, line 59: **Silent under-reporting = false negative in a security gate.** Coverage depends on `--first-parent` + the squash subject ending in `(#N)`. A PR merged via merge/rebase commit, or a squash whose title dropped the `(#N)` suffix, is skipped and never audited — which is exactly the merge an approval-drift guard should catch. Consider comparing the first-parent commit count in range against the matched count and surfacing "N commits could not be mapped to a PR" as a warning so the blind spot is visible.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+                      if [ -f "$BODY_FILE" ]; then
+                          cat "$BODY_FILE"
+                      else
+                          echo "_No report produced._"
+                      fi
+                      echo
+                      echo "<sub>Deterministic scanners: approval-drift + supply-chain dependency diff. Read-only; no PR code executed.</sub>"
+                  } > comment.md
+                  jq -Rs '{body: .}' comment.md > payload.json
+                  curl -sSf -X POST \
```
enyst: Two minor notes on this POST: (1) it creates a fresh comment on every `labeled` event — re-adding the label spams duplicates; an upsert-by-marker (find → PATCH) keeps one comment. (2) On a `pull_request` from a fork, `GITHUB_TOKEN` is read-only regardless of the `permissions:` block, so `curl -sSf` 403s and (no `continue-on-error` here) fails the job. Not an issue for release PRs cut from internal branches, but a stray label on a fork PR would produce a confusing red check — worth a guard or note.

enyst (commented):
## 🟡 Code Review — Taste Rating: Acceptable

Solid, well-scoped work. The design is genuinely good: both scanners are deterministic (read git history + read-only APIs, never execute PR code), fail *safe* (network/OSV errors and un-auditable PRs degrade to warnings, not blockers), and the `pull_request`-not-`pull_request_target` choice is correct and well-justified. I fetched the branch and ran both scanners against this repo's real history:

- `check_dependency_diff.py` over `v1.32.0..HEAD` → ✅ no findings, correctly identifies the 4 first-party version bumps (exit 0).
- `check_approval_drift.py` over `v1.33.0..HEAD` → ✅ 6 PRs, 6 clean, 0 flagged (exit 0).

Both match the evidence in the PR description. `actions/checkout@v7` and `actions/setup-python@v6` both exist and match the repo's dominant pinning (41× `checkout@v7`, 9× `setup-python@v6`).

The blocking-vs-acceptable items below are about robustness and test coverage, not fundamental design.

---

### [TESTING GAPS] — the main thing to address

This repo has a clear, established convention: `.github/scripts/check_*.py` scripts get real unit tests in `tests/cross/` — e.g. `tests/cross/test_check_version_bumps.py` and `tests/cross/test_check_deprecations.py`, which exercise the real logic against temp git repos and real parsing (not mocks). This PR adds three new scripts with pure, highly-testable functions (`_load_lock`, `_is_trusted_registry`, `merged_pr_numbers`, `audit_pr` split from I/O) and ships **zero tests**.

For a *security gate*, the parsing/classification logic is exactly what must not silently regress. Please add `tests/cross/test_check_dependency_diff.py` and `tests/cross/test_check_approval_drift.py` following the existing pattern, covering at minimum:
- `_load_lock` diffing (added / bumped / removed) from two lockfile revisions in a temp repo;
- `_is_trusted_registry` for `{registry: pypi}`, `{editable}`, `{git}`, `{url}`, and the empty/missing-source case (see the robustness note below);
- `merged_pr_numbers` extraction, including subjects that do **not** match `(#N)`;
- `audit_pr` status derivation (`no-human-approval`, `changed-after-approval`, `ok`, `trusted-bot`) from small fixture dicts — no network needed since it's pure once the two API payloads are passed in.

The PR description evidence (real commands + outputs) is good and appreciated, but per this repo's convention it doesn't substitute for the missing tests.

---

**1. `_is_trusted_registry` is brittle and can produce release-blocking false positives** — `.github/scripts/check_dependency_diff.py` L69–76.
It trusts a package only if `source["registry"]` contains the substring `"pypi.org"`, else (for non-registry dicts) only `virtual`/`editable`. Two concerns:
- I verified this repo's current `uv.lock`: all 413 packages carry either `{registry: "https://pypi.org/simple"}` (409) or `{editable: ...}` (4), so it works **today**. But uv can omit the `source` table for default-index packages (or emit a mirror/alt-index registry) depending on `[tool.uv]` index config. If that ever happens, every normal added/bumped dep falls through to "non-PyPI source" → `report.block(...)` → **the release gate blocks on ordinary PyPI packages**. That's a false-positive that fails closed on the wrong signal.
- `"pypi.org" in registry` is a substring match, so `https://pypi.org.evil.example/simple` would be treated as trusted. An `==`/host-parse against the known-good index URL(s) is safer and clearer.

Suggest: match the registry host exactly (allowlist the canonical PyPI simple URL), and decide deliberately what an empty/missing `source` means — treat it as *default registry = trusted* rather than *unknown = block*, or at least emit a warning instead of a hard block for the "unrecognized source shape" case.

**2. `merged_pr_numbers` can silently under-report — a false negative in a security gate** — `.github/scripts/check_approval_drift.py` L48–61.
Coverage depends on `--first-parent` + the squash-merge subject ending in `(#N)`. Any PR merged without that convention (a rebase/merge-commit merge, or a squash whose title lost the `(#N)` suffix) is skipped entirely and never audited. For an approval-drift guard, a skipped merge is exactly the merge you'd want to catch. Consider cross-checking the count of first-parent commits in range against the number that matched, and surfacing "N commits in range could not be mapped to a PR" as a warning so the blind spot is visible rather than silent.

---

### [STYLE / MINOR — non-blocking]

- **Comment posting is fire-and-new every time** (`security-scan.yml` L82–108): each `labeled` event POSTs a fresh issue comment with no upsert/dedup. Re-adding the label (or multiple runs) spams duplicate comments. Consider updating a single marker comment (find-by-marker → PATCH) as some of the existing labeled suites may already do.
- **`workflow_dispatch` produces no visible artifact**: the "Post results" step is gated on `github.event_name == 'pull_request'`, and the report is written to `security-report.md` in the workspace, not `$GITHUB_STEP_SUMMARY`. On manual dispatch you only get raw logs. Writing the rendered report to `$GITHUB_STEP_SUMMARY` would make both trigger paths legible.
- **Fork-PR caveat** (`security-scan.yml` L82–108): on a `pull_request` from a fork, `GITHUB_TOKEN` is read-only regardless of the `permissions:` block, so the `curl -sSf` comment POST 403s and (no `continue-on-error`) fails the job. Not a problem for the intended use — release PRs are cut from internal branches — but worth a one-line note/guard so a stray label on a fork PR doesn't produce a confusing red check.

---

### [RISK ASSESSMENT]

New opt-in CI job, label-gated, additive only (639 insertions, 0 deletions), touches no product/runtime code. Scanners are read-only and never execute PR content. Worst realistic failure mode is a false-positive block from the `_is_trusted_registry` edge case (fails closed — annoying, not dangerous) or a false-negative skip in `merged_pr_numbers` (a coverage gap in a brand-new, not-yet-relied-upon gate). No secrets beyond the default token. The PR itself already flags it isn't dogfooded on a live release yet — enabling it on the next release cut before treating it as a hard gate is the right call.

---

**VERDICT:** ✅ Worth merging once the tests are added (this repo's own convention) and the `_is_trusted_registry` missing/mirror-source case is made deliberate. The two `[IMPROVEMENT OPPORTUNITIES]` are robustness fixes, not redesigns.

**KEY INSIGHT:** The design is sound and fails safe; the remaining risk is entirely in two input-classification assumptions (`uv.lock` source shape, and the squash `(#N)` merge convention) that are correct for this repo *today* but silently degrade the gate if either drifts — which is precisely why the missing `tests/cross/` coverage matters most.

---

>
>
>

<sub>This review was generated by an AI agent (OpenHands) on behalf of the requester.</sub>

.github/scripts/check_dependency_diff.py
smolpaws: Fixed in `48e37cc`. Replaced the substring check with an exact-host allowlist (`pypi.org`, `files.pythonhosted.org`) via a parsed-URL hostname compare, so `pypi.org.evil.example` no longer passes. On the missing/empty-source question I made a deliberate call the *opposite* way to the suggestion: **unknown/empty source fails closed (blocks)**, not trusted-default. Rationale: on a release PR an unexpected/unclassifiable source is exactly the supply-chain signal, and a BLOCKED is cheap — it just parks the release for a maintainer to look at (and re-run/clear on the fly), where a silent trust would be the dangerous default. First-party workspace members (`editable`/`virtual`) stay trusted. Covered by `tests/cross/test_check_dependency_diff.py` (lookalike host, mirror, git/url, empty/missing shapes).

.github/scripts/check_approval_drift.py
smolpaws: Fixed in `48e37cc`. `merged_pr_numbers` now returns `(matched, unmapped)`: every first-parent commit whose subject doesn't end in `(#N)` (a merge/rebase merge, or a squash that lost the suffix) goes into `unmapped`, and `main` surfaces it as a visible warning — "N first-parent commit(s) in range did not match the `(#N)` convention and were **not** audited" with the subjects listed — instead of silently dropping it. So the blind spot is now legible. Tests in `tests/cross/test_check_approval_drift.py` cover the matched/unmapped split, a trailing-whitespace subject, and the `(#N)`-not-at-end case.

.github/workflows/security-scan.yml
smolpaws: Both fixed in `48e37cc`. (1) The POST is now an **upsert**: it searches existing PR comments for a `<!-- release-security-scan -->` marker and PATCHes that one if present, only POSTing a fresh comment when none exists — so re-labeling / re-runs update in place instead of spamming. (2) The comment step is now guarded with `github.event.pull_request.head.repo.full_name == github.repository`, so on a fork PR (where `GITHUB_TOKEN` is read-only) it's skipped rather than 403-ing the job. The full report is always written to `$GITHUB_STEP_SUMMARY` in a separate always-run step, so fork PRs and `workflow_dispatch` still get legible output.

smolpaws:
Pushed `48e37cc` addressing the review 🐾

**Robustness (the two improvement points):**
- **Exact-host registry allowlist** — `pypi.org` / `files.pythonhosted.org` matched on parsed hostname, not substring, so `pypi.org.evil.example` no longer slips through; mirror / alt-index / git / url / unknown source all block.
- **Approval-drift blind spot surfaced** — `merged_pr_numbers` returns `(matched, unmapped)`; commits that don't fit the `(#N)` squash convention are now reported as a visible warning instead of silently skipped.

**Deliberate design call (flags a divergence from the review):** the review suggested degrading the unknown-source and OSV-failure cases to *warnings* (fail-safe). I went **fail-closed** instead — both now **block**. On a release gate, "can't verify ⇒ don't ship" is the safer default, and a BLOCKED is cheap: it just parks the release for a human to look at and re-run/clear on the fly. Updated the PR body to match. (Transient GitHub-API errors in approval-drift stay warnings, so a flaky call doesn't wedge a release.)

**Style notes:** comment POST is now upsert-by-marker (no dup spam), guarded to non-fork PRs, and the report always goes to `$GITHUB_STEP_SUMMARY` so `workflow_dispatch`/fork runs are legible.

**Tests (the merge condition):** added `tests/cross/test_check_dependency_diff.py` + `test_check_approval_drift.py` — 17 tests over the source classifier and the matched/unmapped split, following the `tests/cross/` convention. `ruff check`/`format` clean; both scanners still run green against this repo's real `v1.33.0..HEAD`.

enyst:
@openhands Do a /codereview on this PR. Post your feedback as a review with gh api. Feel free to read all previous discussions / inline threads here too.

.github/scripts/check_dependency_diff.py
smolpaws: To be explicit about ownership: the fail-closed choice here is **the maintainer's decision** (Engel's), not just my preference — flagging it since it deliberately overrides the review's fail-safe suggestion.

The reasoning, in his framing: for *us* a BLOCKED is cheap. This gate only runs on release PRs, which are cut from internal branches by a human who's already in the loop — so a false block isn't a silent failure that ships bad bits, it's just a signal that says "a human should look." We can re-run it, or clear/waive it on the fly in seconds. The asymmetry is the whole point: the cost of an unnecessary block is a minute of a maintainer's attention; the cost of silently trusting an unverified/unexpected source on a release is a supply-chain incident. At our size and cadence we'd rather pay the cheap side every time. Fail-safe would be the right call for a high-volume gate on every PR where blocks are expensive and frequent — this isn't that.

enyst (commented):
## 🟢 Code Review (re-review of `48e37cc`) — Taste Rating: Good taste

The three items from the previous review are all addressed, and I re-verified each against this repo's real state rather than taking the reply at face value:

**Verification I ran on the current branch:**
- **Tests exist and pass** — `tests/cross/test_check_approval_drift.py` + `tests/cross/test_check_dependency_diff.py`, **17 passed**. They exercise the real pure logic (`merged_pr_numbers` matched/unmapped split, `_is_trusted_registry`, `_registry_host`, `_source_label`, `_is_bot`) with no mocks-as-tests — exactly the parsing/classification surface a security gate must not silently regress. This closes the main gap from last round.
- **Supply-chain diff** over `v1.32.0..HEAD` → ✅ no findings, 4 first-party bumps, live OSV query succeeded (exit 0).
- **Approval-drift** over `v1.33.0..HEAD` → 6 PRs, 6 clean, 0 flagged, and the **new unmapped-commit warning** correctly surfaces the 3 not-yet-squash-merged commits in range as a visible blind-spot warning (exit 0, non-blocking) rather than silently dropping them.
- **Exact-host allowlist fix confirmed**: `_is_trusted_registry({"registry": "https://pypi.org.evil.example/simple"})` → `False`; mirror / git / url / empty all block; `editable`/`virtual` stay trusted. I also re-checked the live `uv.lock`: 409 `registry` + 4 `editable`, **zero** `<no source>` packages, so the deliberate "unknown source fails closed" decision won't false-positive on ordinary PyPI deps today.
- **Workflow hardening confirmed**: comment step is now an upsert-by-marker (`<!-- release-security-scan -->`), guarded on `head.repo.full_name == github.repository` so fork PRs skip the POST instead of 403-ing, and the full report is always written to `$GITHUB_STEP_SUMMARY`. `actions/checkout@v7` / `actions/setup-python@v6` match the repo's dominant pinning.

I want to explicitly endorse the one place smolpaws deliberately went *against* the prior suggestion: making an **unknown/empty `uv.lock` source fail closed (block)** rather than defaulting to trusted. On a *release* diff that's the correct call — an unclassifiable source is precisely the supply-chain signal, and a block is cheap (it parks the release for a maintainer) whereas silent trust is the dangerous default. Good taste.

---

### [IMPROVEMENT OPPORTUNITIES] — non-blocking

**1. Reviews are fetched as a single un-paginated page** — `.github/scripts/check_approval_drift.py` L88–89.
`audit_pr` requests `/pulls/{number}/reviews?per_page=100` with no pagination loop. The reviews endpoint returns *every* review object (COMMENTED included, not just APPROVED), so a heavily-discussed PR can exceed 100. If the human `APPROVED` review sits past the first 100, the PR is misclassified as `no-human-approval` → a **false-positive block**. Rare, and it fails safe (blocks a clean release rather than passing a dirty one), but on a security gate it's worth following the `Link: rel="next"` header or looping until a short page. Not blocking for merge.

---

### [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟢 **LOW**

New opt-in/label-gated CI job, additive only (946 insertions, 0 deletions), no product/runtime code touched. Scanners are read-only, never execute PR content, and use `pull_request` (not `pull_request_target`) with only the default `GITHUB_TOKEN`. Both fail safe (network/OSV errors and un-auditable PRs degrade to warnings/blocks, never to a silent pass). Worst realistic failure is a false-positive block (annoying, not dangerous) from the pagination edge above. The PR is honest that it isn't dogfooded on a live release yet — enabling it on the next release cut before treating it as a hard gate remains the right sequencing.

---

**VERDICT:** ✅ **Worth merging.** The prior round's testing gap and both robustness items are genuinely fixed and verified against real history. The single remaining note (reviews pagination) is a minor, fail-safe edge case that can land here or as a follow-up.

**KEY INSIGHT:** The gate's integrity now rests on well-tested, deterministic input classification, and every failure mode is fail-closed/fail-safe — the design's original strength, now backed by the `tests/cross/` coverage that makes those assumptions regression-proof.

---

>
>
>

<sub>This review was generated by an AI agent (OpenHands) on behalf of the requester.</sub>

smolpaws:
Fixed the pagination edge case in `e9a989e` — folding it in here rather than a follow-up, since a silently-wrong gate on a big PR is exactly the kind of thing this PR is meant to prevent.

**What changed:** `audit_pr` fetched `/pulls/{n}/reviews?per_page=100` as a single page. Since that endpoint returns *every* review object (COMMENTED too), a human `APPROVED` past the first 100 would be missed → false-positive `no-human-approval` block. Added `github_request_all()` to `security_scan_common` which follows the `Link: rel="next"` header and concatenates all pages; `audit_pr` now uses it. A non-list response is wrapped as a single item so an error object isn't iterated element-wise.

**Tests:** new `tests/cross/test_security_scan_common.py` covers `_next_link` parsing and multi-page / single-page / non-list pagination. Full cross suite now **23 passed**. Smoke-ran approval-drift over `v1.33.0..HEAD` with a real token: 6 PRs, 6 clean, exit 0.

You called it right — it fails safe, but on a security gate a silent false block is worth closing. 🐾

enyst:
@openhands Do a /codereview on this PR. Post your feedback as a review with gh api. Feel free to read all previous discussions / inline threads here too. Post as review with approved or comment or request changes as you think fit.

enyst (approved):
## 🟢 Code Review (re-review of `e9a989e`) — Taste Rating: Good taste

This is the third pass, focused on the one commit that landed **after** the last 🟢 re-review: `e9a989e` — *"paginate PR reviews to avoid false-positive block."* I read all six changed files at PR head, re-read the two prior reviews and the three (now-resolved) inline threads so I'm not re-litigating settled feedback, and **verified every claim against this repo's real history** rather than taking the description at face value.

### What I verified on the current branch (`e9a989e`)

- **Tests pass** — `tests/cross/test_check_approval_drift.py`, `test_check_dependency_diff.py`, and the new `test_security_scan_common.py` → **23 passed**. They exercise the real pure logic (pagination `_next_link` / `github_request_all` multi-page + non-list wrapping, `merged_pr_numbers` matched/unmapped split, `_is_trusted_registry`, `_registry_host`, `_source_label`, `_is_bot`) with no mocks-as-tests — exactly the parsing/classification surface a security gate must not silently regress.
- **Supply-chain diff** over `v1.32.0..HEAD` → ✅ no findings; correctly identifies the 4 first-party version bumps; **live OSV query succeeded** (exit 0).
- **Approval-drift** over `v1.33.0..HEAD` → 6 PRs, 6 clean, 0 flagged, 0 un-auditable; the unmapped-commit warning correctly surfaces this PR's own 4 not-yet-squash-merged commits as a visible blind-spot warning (exit 0, non-blocking).
- **Pagination fix is correct.** `audit_pr` now calls `github_request_all(.../reviews?per_page=100)`, which follows `Link: rel="next"` and concatenates pages. This closes a genuine false-positive: the `/reviews` endpoint returns *every* review object (COMMENTED included), so a human `APPROVED` past the first 100 would previously be missed and the PR misclassified as `no-human-approval` — a spurious release block. The refactor cleanly splits `_github_get` (one page + Link header) from `github_request` (single) / `github_request_all` (paginated), and the non-list-response guard means an error object is never iterated element-wise.
- **Lint clean** — `ruff check` and `ruff format --check` pass on all three scripts and the test files.
- **Pinned actions** `actions/checkout@v7` and `actions/setup-python@v6` exist and match the repo's dominant pinning (41× checkout@v7, 9× setup-python@v6).

### Design remains sound

Both scanners are deterministic (read git history + read-only GitHub REST / OSV, never execute PR code), which correctly justifies plain `pull_request` over `pull_request_target` and no secrets beyond the default `GITHUB_TOKEN`. The fail-closed choices (OSV lookup failure and unrecognized dep source both block; transient per-PR API errors degrade to warnings so a flaky call can't wedge a release) are deliberate, documented, and owner-endorsed. No data-structure smells, no >3-level nesting, no breaking changes to existing workflows.

### One minor, non-blocking note (no action required to merge)

The opt-in path (behavior B: a `security-scan` label on a non-`release/**` PR) runs on the `labeled` event but not on subsequent `synchronize` pushes — so a labeled feature PR that gets new commits won't automatically re-scan until re-labeled or re-dispatched. This is fine because the *actual* gate (behavior A, `release/**`) does re-run on `synchronize`, and behavior B is explicitly "best-effort opt-in." Flagging only so it's a conscious choice; not worth changing here.

**[RISK ASSESSMENT]**
  New CI-only, opt-in/label-gated workflow plus three read-only stdlib scripts and their tests. No production code paths, no runtime dependencies added, no execution of PR-controlled content, no new secrets. Worst-case failure mode is a spurious release-gate block that a maintainer re-runs or clears — no data exposure or privilege surface. Not yet dogfooded on a live release PR (author notes this); enabling on the next release cut is the natural validation.

**VERDICT:**
✅ **Worth merging** — core logic is sound, well-tested against real history, and all prior review feedback is addressed and verified.

**KEY INSIGHT:**
Auditing the release diff (last tag → PR) as one unit, with read-only deterministic scanners, is the right place and the right mechanism to catch approval-drift/TOCTOU and supply-chain risk without turning the gate itself into an injection target.

<sub>Reviewed by an AI agent (OpenHands) on behalf of @enyst.</sub>

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,165 @@
...
+# injection target. We therefore use plain `pull_request` (not
+# pull_request_target) and need no secrets beyond the default GITHUB_TOKEN. On
+# fork PRs that token is read-only, so results always go to the job summary and
+# the PR comment is posted only when the token can write (non-fork).
+
+on:
+    pull_request:
+        # `opened/synchronize/reopened` drive behavior A on release/** PRs;
+        # `labeled` drives behavior B anywhere.
+        types: [opened, synchronize, reopened, labeled]
```
enyst:
```suggestion
        types: [labeled]
```

PR #4050, opened 2026-07-09
fix: isolate parallel example tests to remove infra flakiness

### LINKED ISSUE #4049: Flaky example tests: port collision (14_client_defined_tools) and shared tmux socket (50_async_cancellation) under parallel test-examples run
## Summary

Two example scripts flake in the `test-examples` CI job because they contend over **global OS resources** when examples run in parallel. The job runs:

```
pytest tests/examples/test_examples.py --run-examples --examples-results-dir "$RESULTS_DIR" -n 4
```

so up to 4 examples run concurrently, each as its own subprocess. Both failures below happened in ~11s — i.e. before any real model work — so this is **infra flakiness, not LLM flakiness**. Re-running usually passes, which is why they read as flaky.

Observed on the v1.34.0 release PR (#4048), job [`test-examples`](https://github.com/OpenHands/software-agent-sdk/actions/runs/29004245673/job/86071896895):

```
Total: 65 | Passed: 63 | Failed: 2
- examples/01_standalone_sdk/50_async_cancellation.py: Exit code 1
- examples/02_remote_agent_server/14_client_defined_tools.py: Exit code 1
```

> Note: an earlier revision of this issue contained two inaccuracies (the port-collision inventory, and the cited tmux file/line + a claim that async cancellation was involved). Both are corrected below and explained in a follow-up comment.

---

## Flaky test 1 — `examples/02_remote_agent_server/14_client_defined_tools.py`

**Error** (raised at `14_client_defined_tools.py:113`)
```
[SERVER] ... [Errno 98] error while attempting to bind on address ('127.0.0.1', 8001): [errno 98] address already in use
RuntimeError: Server process exited before becoming healthy. Check the server logs above.
```

**Root cause:** two examples hardcode the **same** `ManagedAPIServer` port, so under `-n 4` if they run concurrently the second can't bind and its server exits. Actual ports in use across the `02_remote_agent_server` examples:

| Example | Port |
|---|---|
| `01_convo_with_local_agent_server.py` | **8001** |
| `14_client_defined_tools.py` | **8001** ← collides with `01` |
| `11_conversation_fork.py` | 8002 |
| `16_deferred_init.py` | 8003 |
| `12_settings_and_secrets_api.py` | 8765 |
| `13_workspace_get_llm.py` | 8766 |
| `15_openai_compatible_gateway.py` | 8770 |

`8001` is the only duplicated port — the collision is specifically **`01` ↔ `14`**. All other examples already use distinct ports.

**Minimal fix:** give `14_client_defined_tools.py` (or `01`) a unique port (e.g. `8767`) so no two examples share one.

**Robust fix (preferred, removes the whole class of bug):** make `ManagedAPIServer` grab a free ephemeral port instead of a fixed one. In `examples/02_remote_agent_server/scripts/utils.py`:

```python
import socket

def _pick_free_port(host: str) -> int:
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        s.bind((host, 0))
        return s.getsockname()[1]
```

Change `__init__(self, port: int = 8000, ...)` → `port: int | None = None`, then:

```python
self.port = port if port is not None else _pick_free_port(host)
```

Every caller already reads `server.base_url` (never the literal port), so nothing downstream breaks. Then the explicit `port=` args can be dropped from the examples.

---

## Flaky test 2 — `examples/01_standalone_sdk/50_async_cancellation.py`

**Error**
```
ValueError: tmux set-environment stderr: ['no server running on /tmp/tmux-1001/openhands']
```

**Full traceback (key frames):**
```
File "examples/01_standalone_sdk/50_async_cancellation.py", line 84, in <module>
    conversation.send_message(...)
File "openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py", line 1564, in send_message
    self._ensure_agent_ready()
File ".../local_conversation.py", line 1283, in _ensure_agent_ready
    self.agent._initialize(self._state)
File "openhands-sdk/openhands/sdk/agent/base.py", line 537, in _initialize
    result = future.result()
File "openhands-tools/openhands/tools/terminal/definition.py", line 323, in create
    executor = TerminalExecutor(...)
File "openhands-tools/openhands/tools/terminal/impl.py", line 144, in _initialize_pool
    self._pool.initialize()
File "openhands-tools/openhands/tools/terminal/terminal/tmux_pane_pool.py", line 131, in initialize
    self._session.set_environment(k, v)
File ".../libtmux/common.py", line 92, in set_environment
    raise ValueError(msg)
ValueError: tmux set-environment stderr: ['no server running on /tmp/tmux-1001/openhands']
```

**Root cause:** the tmux terminal shares **one global socket**. `TMUX_SOCKET_NAME: Final[str] = "openhands"` (`openhands-tools/openhands/tools/terminal/constants.py`), and with no `TMUX_TMPDIR` the socket resolves to `/tmp/tmux-<uid>/openhands` — the same path for every process on the runner. Session *names* are unique, but the *server* is shared. `TmuxPanePool.initialize()` does `new_session(...)` and then loops `set_environment(...)`; if a concurrent example's terminal teardown kills the last remaining session on that shared server in between, the server exits and this process's `set_environment` sees "no server running".

Two important clarifications vs. the original issue:
- **Async cancellation is not involved.** The failure happens at module load during `conversation.send_message()` → agent/tool initialization — i.e. *before* `main()`, `arun()`, or the auto-cancel timer ever run. It is not specific to the cancellation example.
- **It affects any standalone example that uses the terminal.** 34 examples under `examples/01_standalone_sdk/` use `TerminalTool` and none set `TMUX_TMPDIR`, so all share `/tmp/tmux-<uid>/openhands`. `50` just happened to lose the race in this run.

**This is already solved for the agent-server, and the same approach applies here.** `agent-server` defaults `TMUX_TMPDIR` to a per-PID directory when unset (`openhands-agent-server/openhands/agent_server/api.py:78-95`, documented in `AGENTS.md`), which is why remote-server examples don't hit this. The same CI log shows it working: `TMUX_TMPDIR not set; defaulting to per-server tmux directory /tmp/openhands-agent-server-29036`. Standalone in-process terminals get no such isolation.

**Proposed fix (immediate, low-risk, test-only):** give each example subprocess its own tmux socket dir. In `tests/examples/test_examples.py::test_example_scripts`, where `env` is built:

```python
import tempfile
...
env.setdefault("PYTHONUNBUFFERED", "1")
env.setdefault("PYTHONIOENCODING", "utf-8")
# Isolate each example's tmux server so parallel workers can't tear down a shared one.
env["TMUX_TMPDIR"] = tempfile.mkdtemp(prefix="tmux-")  # mkdtemp is 0700, which tmux requires
```

**Proposed fix (deeper, consistent with existing code):** have the standalone terminal tool default `TMUX_TMPDIR` per-process when unset — mirroring exactly what `agent_server/api.py` already does — so every standalone SDK user (not just CI) gets isolation. Sessions are already uniquely named and cleaned up, so nothing relies on a *shared* server.

---

## Optional safety net

Independently of the fixes, add `pytest-rerunfailures` and run the examples suite with `--reruns 1` so a genuinely transient failure doesn't red a release. This is a backstop, **not** a substitute for the two root-cause fixes above.

VascoSch92:
HUMAN:

Fixing some flake test.

---

AGENT:

## Why

The `test-examples` CI job runs the example scripts in parallel
(`pytest tests/examples/test_examples.py --run-examples -n 4`), and two examples
intermittently fail because they contend over **global OS resources** — this is
infra flakiness, not LLM flakiness (both fail in ~11s, before any model call).
Observed on the v1.34.0 release PR #4048; full analysis in #4049.

- `01_standalone_sdk/50_async_cancellation.py` →
  `ValueError: tmux set-environment stderr: ['no server running on /tmp/tmux-1001/openhands']`.
  All standalone examples share one tmux socket (`$TMUX_TMPDIR/tmux-<uid>/openhands`,
  a fixed socket name). When one example's terminal teardown kills the last session,
  the shared tmux **server** exits, and another example that is mid-`TmuxPanePool.initialize()`
  hits `set_environment` on a now-dead server. (The failure happens during
  `send_message()` → agent init; the async-cancellation aspect is unrelated. Any of
  the 34 standalone examples that use `TerminalTool` can lose this race.)
- `02_remote_agent_server/14_client_defined_tools.py` →
  `[Errno 98] address already in use ('127.0.0.1', 8001)`. It hardcoded the same
  `ManagedAPIServer` port as `01_convo_with_local_agent_server.py` (both `8001`).

## Summary

- **tmux isolation:** give each example subprocess its own `TMUX_TMPDIR` in
  `tests/examples/test_examples.py`, so parallel workers can't tear down a shared
  tmux server. This mirrors the per-process `TMUX_TMPDIR` default the agent-server
  already applies (`openhands-agent-server/openhands/agent_server/api.py`).
- **port collision:** move `14_client_defined_tools.py` to a unique port (`8004`);
  `8001` was the only duplicated port across the `02_remote_agent_server` examples.

## Issue Number

#4049

## How to Test

The authoritative end-to-end test is re-running the `test-examples` workflow (it
needs LLM credentials and the `-n 4` parallel runner); with these changes the two
examples should no longer flake.

Local checks performed:

- `ruff check` and `ruff format --check` pass on both edited files.
- Port `8004` is unique across all `examples/` (no other example uses it).
- Demonstrated that per-`TMUX_TMPDIR` isolation prevents the cross-teardown that
  caused the tmux failure — two servers using the same socket name are isolated,
  and killing one leaves the other alive:

```
$ TMUX_TMPDIR=$D1 tmux -L openhands new-session -d -s s1
$ TMUX_TMPDIR=$D2 tmux -L openhands new-session -d -s s2
# sockets land under separate dirs:
#   $D1/tmux-<uid>/openhands
#   $D2/tmux-<uid>/openhands
$ TMUX_TMPDIR=$D1 tmux -L openhands kill-server
$ TMUX_TMPDIR=$D2 tmux -L openhands list-sessions
s2: 1 windows (created ...)      # D2 unaffected
```

I could not run the full `--run-examples` suite locally (it requires API keys and
the parallel runner), so CI is the real validation.

## Video/Screenshots

N/A — CLI/CI-only change; the tmux isolation evidence is included above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Intentionally minimal. #4049 also describes more robust follow-ups that are out of
scope here: (1) have `ManagedAPIServer` pick a free ephemeral port instead of fixed
ports, and (2) default `TMUX_TMPDIR` per-process inside the standalone terminal tool
itself (not just the CI harness), matching what the agent-server already does.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:716b468-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-716b468-python \
  ghcr.io/openhands/agent-server:716b468-python
```

```
ghcr.io/openhands/agent-server:716b468-golang-amd64
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-golang-amd64
ghcr.io/openhands/agent-server:fix-flaky-example-tests-golang-amd64
ghcr.io/openhands/agent-server:716b468-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:716b468-golang-arm64
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-golang-arm64
ghcr.io/openhands/agent-server:fix-flaky-example-tests-golang-arm64
ghcr.io/openhands/agent-server:716b468-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:716b468-java-amd64
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-java-amd64
ghcr.io/openhands/agent-server:fix-flaky-example-tests-java-amd64
ghcr.io/openhands/agent-server:716b468-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:716b468-java-arm64
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-java-arm64
ghcr.io/openhands/agent-server:fix-flaky-example-tests-java-arm64
ghcr.io/openhands/agent-server:716b468-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:716b468-python-amd64
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-python-amd64
ghcr.io/openhands/agent-server:fix-flaky-example-tests-python-amd64
ghcr.io/openhands/agent-server:716b468-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:716b468-python-arm64
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-python-arm64
ghcr.io/openhands/agent-server:fix-flaky-example-tests-python-arm64
ghcr.io/openhands/agent-server:716b468-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:716b468-golang
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-golang
ghcr.io/openhands/agent-server:fix-flaky-example-tests-golang
ghcr.io/openhands/agent-server:716b468-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:716b468-java
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-java
ghcr.io/openhands/agent-server:fix-flaky-example-tests-java
ghcr.io/openhands/agent-server:716b468-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:716b468-python
ghcr.io/openhands/agent-server:716b4688947cff42aefdbf1e338a7940a1a0ffb4-python
ghcr.io/openhands/agent-server:fix-flaky-example-tests-python
ghcr.io/openhands/agent-server:716b468-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `716b468-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `716b468-python-amd64`) are also available if needed


VascoSch92:
@enyst Should we run the test-examples?

enyst (approved):

enyst:
> @enyst Should we run the test-examples?

I would say no need, under the assumption that, during a Release PR, we do look at failing tests if any, and solve before releasing.

This is my normal working assumption, and we basically always do it. So we're good.

PR #3998, opened 2026-07-05
ci: don't block OpenHands bump PR when OpenHands-CLI PR creation fails

aivong-openhands:
## Why
The OpenHands-CLI bump step currently aborts the whole job when it fails, which blocks the OpenHands PR from ever being created — as happened for v1.31.1, where the CLI's `agent-client-protocol<0.9.0` pin conflicts with the SDK's `>=0.10.1` floor (OpenHands/OpenHands-CLI#786) and will keep failing until that migration lands. Marking the CLI step `continue-on-error` lets the OpenHands PR proceed independently, and the job summary now flags a CLI failure explicitly so it doesn't fail silently behind a green check.

## Validation
- YAML parses (`python3 -c "yaml.safe_load(...)"`).
- Re-traced the v1.31.1 failure path: with `continue-on-error: true` on `cli_pr`, the OpenHands step and Summary still run, and `steps.cli_pr.outcome == 'failure'` selects the warning line in the summary.

This PR was drafted by an AI agent on behalf of the user.

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
| python | amd64, arm64 | `nikolaik/python-nodejs:python3.13-nodejs22-slim` | [Link](https://hub.docker.com/_/nikolaik/python-nodejs:python3.13-nodejs22-slim) |
| golang | amd64, arm64 | `golang:1.21-bookworm` | [Link](https://hub.docker.com/_/golang:1.21-bookworm) |

**Pull (multi-arch manifest)**
```bash
# Each variant is a multi-arch manifest supporting both amd64 and arm64
docker pull ghcr.io/openhands/agent-server:8541979-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-8541979-python \
  ghcr.io/openhands/agent-server:8541979-python
```

**All tags pushed for this build**
```
ghcr.io/openhands/agent-server:8541979-golang-amd64
ghcr.io/openhands/agent-server:8541979b172f7db8ca6fa24a00a1dab609643156-golang-amd64
ghcr.io/openhands/agent-server:av-version-bump-prs-cli-nonblocking-golang-amd64
[...]

PR #4051, opened 2026-07-09
ci(version-bump-prs): make PR-creation steps independent

VascoSch92:
HUMAN:

If one step is failing, the other are still running.

---

AGENT:

## Why

In the `create-version-bump-prs` job, a step with no `if:` condition implicitly
requires all prior steps to have succeeded. In [run 29010691311](https://github.com/OpenHands/software-agent-sdk/actions/runs/29010691311/job/86093299051),
the "Create PR for OpenHands-CLI repo" step failed, which caused the
independent "Create PR for automation repo", "Summary", and "Notify Slack"
steps to be skipped — even though they bump unrelated repos and don't depend
on that step's outcome.

## Summary
- Give the "Wait for packages to be available on PyPI" step an `id: wait_for_packages`.
- Change the three "Create PR for..." steps, "Summary", and "Notify Slack" to
  `if: ${{ !cancelled() && steps.wait_for_packages.outcome == 'success' }}`
  (Notify Slack keeps its existing `SLACK_BOT_TOKEN` check too), so each only
  depends on the shared setup steps, not on its sibling steps.

## Issue Number

N/A

## How to Test

- Validated the YAML with `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/version-bump-prs.yml'))"` — parses cleanly.
- Printed each step's resolved `if:` condition via PyYAML to confirm the three
  "Create PR" steps, "Summary", and "Notify Slack" now depend on
  `wait_for_packages` rather than each other, while "Checkout", "Get version",
  "Validate version", and "Install uv" keep their default (all-must-succeed)
  behavior.
- Did not trigger a real `workflow_dispatch` run, since this workflow creates
  live PRs against `OpenHands/OpenHands`, `OpenHands/openhands-cli`, and
  `OpenHands/automation` — not safe to exercise end-to-end outside of an
  actual release.

## Video/Screenshots

N/A — CI workflow change, no UI.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Only the `create-version-bump-prs` job is touched; `bump-typescript-client`
  is a separate job with genuinely sequential steps and is unaffected.
- The job's overall conclusion is still `failure` if any step fails (no
  `continue-on-error` is used) — this only stops one failing step from
  skipping its independent siblings.

### DIFF
--- a/.github/workflows/version-bump-prs.yml
+++ b/.github/workflows/version-bump-prs.yml
@@ -52,7 +52,6 @@ jobs:
                   enable-cache: false
 
             - name: Wait for packages to be available on PyPI
-              id: wait_for_packages
               env:
                   VERSION: ${{ steps.get_version.outputs.version }}
               run: |
@@ -99,9 +98,12 @@ jobs:
                   echo "✅ All packages are resolvable on PyPI!"
 
             - name: Create PR for OpenHands repo
-              # Runs independently of the other "Create PR" steps below: each
-              # bumps an unrelated repo, so one failing shouldn't skip the rest.
-              if: ${{ !cancelled() && steps.wait_for_packages.outcome == 'success' }}
+              id: openhands_pr
+              # continue-on-error keeps the sibling "Create PR" steps below
+              # independent: each bumps an unrelated repo, so one failing must
+              # not skip the rest. Failures are surfaced in the Summary step via
+              # steps.<id>.outcome.
+              continue-on-error: true
               env:
                   VERSION: ${{ steps.get_version.outputs.version }}
               run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/version-bump-prs.yml
```diff
@@ -98,6 +99,9 @@ jobs:
                   echo "✅ All packages are resolvable on PyPI!"
 
             - name: Create PR for OpenHands repo
+              # Runs independently of the other "Create PR" steps below: each
+              # bumps an unrelated repo, so one failing shouldn't skip the rest.
+              if: ${{ !cancelled() && steps.wait_for_packages.outcome == 'success' }}
```
aivong-openhands: We can use `continue-on-error: true`. See this PR for an example https://github.com/OpenHands/software-agent-sdk/pull/3998/changes. I had considered that approach as well prior but went for the simpler reordering to unblock the openhands PR creation

aivong-openhands (approved):

PR #4053, opened 2026-07-09
refactor(skills): report touched paths on observations for path-rule injection

### LINKED ISSUE #4038: Cleanup: replace implicit `getattr(action, "path")` contract for path-rule injection with a typed design
## Summary

While reviewing #3994 (path-triggered rules), we flagged an **implicit contract** in
the way the conversation layer discovers which filesystem path a tool touched:

> `local_conversation.py:525] getattr(action_event.action, "path", None)` is an implicit contract

This issue documents the problem and proposes a few alternative designs for a
**follow-up cleanup** (not for #3994 itself). The hard constraint is that
importing `openhands-tools` from `openhands-sdk` is **not** an option — the SDK
core must not depend on the tools package. So we need a code-design pattern that
keeps that layering intact while making the coupling explicit and typed.

Discussion context: https://github.com/OpenHands/software-agent-sdk/pull/3994#issuecomment-4917245399

## Where it lives

`LocalConversation._touched_rule_path()` correlates a tool `ObservationEvent`
back to its `ActionEvent` and then reads the touched path generically:

```python
# openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
raw_path = getattr(action_event.action, "path", None)
if not isinstance(raw_path, str) or not raw_path:
    return None
```

`Action` (the base schema in `openhands-sdk/openhands/sdk/tool/schema.py`) has
**no** `path` field. The `path` attribute only exists on *some* concrete tool
actions defined in `openhands-tools`. The SDK reaches for it by string name
precisely so it doesn't have to import the tools package.

## Why this is fragile

1. **Name coupling with no interface.** The SDK depends on the *string literal*
   `"path"` matching a field on tool actions it cannot see or type-check. Nothing
   enforces the contract; a rename on either side breaks it silently.

2. **It already misses tools today.** The heuristic assumes every file tool names
   its field `path`. That is not true across the built-in tools:

   | Tool action | Field carrying the file path | Picked up by `getattr(action, "path")`? |
   |---|---|---|
   | `FileEditorAction` (`file_editor`) | `path` | ✅ yes |
   | Gemini `read_file` / `write_file` / `edit` | `file_path` | ❌ no |
   | `apply_patch` | patch text / `move_path` | ❌ no |

   So path rules silently fail to fire for the Gemini file tools and
   apply-patch-based edits, with no error and no log — exactly the class of bug an
   implicit contract produces.

3. **Semantic ambiguity.** "Any action with a non-empty string field named
   `path`" is a guess about semantics. A future non-file tool that happens to have
   a `path` field (a URL path, an HTTP route, a registry path, …) would falsely
   trigger rule injection.

4. **Single path only.** `getattr` returns one value. Real file operations can
   touch *more* than one path (a move/rename touches source **and** destination; a
   patch or glob-driven edit can touch many). The current shape can't express that.

## Constraint

- `openhands-sdk` must **not** import `openhands-tools`. Any solution has to
  express the contract from the SDK side (or via inversion), not by importing
  concrete action classes.

### Option A — Explicit `Protocol` / marker on the producer side
Define a typed contract in the SDK that tool actions opt into:

```python
# in openhands-sdk (tool/schema.py or a new module)
@runtime_checkable
class TouchesFiles(Protocol):
    def touched_paths(self) -> list[str]: ...
```

Tool actions in `openhands-tools` implement `touched_paths()`. The SDK checks
`isinstance(action, TouchesFiles)` — no import of concrete classes, name coupling
gone, multiple paths supported, and it fails *loudly* (type checker) if a tool
claims to touch files but doesn't implement the method.

**Trade-off:** every file tool must implement the method (explicit opt-in), which
is arguably the point.

### Option B — Field-level semantic annotation
Keep it declarative: mark the path field with metadata instead of relying on its
name.

```python
# tool side
path: Annotated[str, FsPath()] = Field(...)
# or: Field(json_schema_extra={"semantic": "fs_path"})
```

The SDK scans an action's model fields for the `FsPath` marker (via
`model_fields` / annotations) and reads whatever field carries it, regardless of
name. Decouples semantics from the field name and naturally supports several
annotated fields.

**Trade-off:** a bit of reflection in the SDK; the marker type still needs to live
in the SDK so tools can import it (correct direction of dependency).

### Option C — Producer reports touched paths on the `Observation` (preferred)
The tool executor already *knows* exactly what it touched. Have it say so, instead
of making the consumer reverse-engineer it from the action:

```python
class Observation(Schema, ABC):
    affected_paths: list[str] = Field(default_factory=list)
```

Path-rule injection then reads `observation.affected_paths` — typed data, no
`getattr`, no action↔observation correlation lookup, supports multiple paths,
moves, and directory operations, and works uniformly for every current and future
file tool. This is the cleanest separation: the layer that performs the effect
declares the effect.

**Trade-off:** touches the `Observation` base schema (a serialized event type, so
it needs backward-compatible handling) and requires each file tool to populate the
field.

### Option D — Dependency-inverted extractor registry
The SDK exposes a registration hook; `openhands-tools` registers, at import time,
extractor callables keyed by action type:

```python
# SDK exposes: register_path_extractor(action_type, fn)
# tools call it in their definition modules
```

The SDK calls the registered extractor for the action's type. Keeps the SDK free
of tool imports (inversion), but adds global mutable registration state and
import-ordering considerations — generally less attractive than A/C.

## Recommendation

Lean toward **Option C** (producer reports `affected_paths` on the observation) as
the primary target, with **Option A** (`Protocol`) as a lighter alternative if we
want to keep the correlation-from-action shape. Both eliminate the string-name
guess, are statically checkable, and fix the "silently misses Gemini/apply_patch"
bug for free. Option C additionally unlocks multi-path and directory-scoped
triggers that came up in the #3994 discussion.

## Out of scope

- This is a **follow-up cleanup**; #3994 intentionally keeps the pragmatic
  `getattr` seam to stay minimal.
- Importing `openhands-tools` into the SDK is explicitly off the table.

Related: #3984, #3994

---
*This issue was created by an AI agent (OpenHands) on behalf of @enyst.*

VascoSch92:
HUMAN:

Refactor path-rule injection.

---

AGENT:

## Why

Path-rule injection discovered the touched file via
`getattr(action_event.action, "path", None)` in
`LocalConversation._touched_rule_path` — an implicit, untyped contract flagged
while reviewing #3994. It fails in two directions with tools that ship today:

- **Misses tools.** The Gemini file tools name the field `file_path`, and
  `apply_patch` carries only a patch string — so `getattr(action, "path")`
  returns `None` and rules silently never fire.
- **Fires spuriously.** `grep` / `glob` / `list_directory` carry a `path` that
  is a *search directory*, not a modified file, so a directory a tool merely
  searched can trigger `{dir}/**` or `**` rules.

It also can't express an operation that touches several files (a rename touches
source **and** destination; a patch can touch many).

## Summary

- Add an `affected_paths` property to the `Observation` base (empty by default);
  file-touching observations override it to declare the paths they modified.
- The injection seam now reads `observation.affected_paths` directly — dropping
  the action-correlation lookup and the `getattr` string-name guess, and
  supporting multiple paths.
- Producers populate it: `file_editor`, Gemini `read`/`write`/`edit`, and
  `apply_patch` (changed files + rename destinations). `grep`/`glob`/`list`
  report nothing, so rules no longer fire on a directory that was only searched.

This follows **Option C** from the issue, adapted: the value is a *derived*
property, not a stored/serialized field, so it needs no event-schema migration
and recomputes correctly after deserialization (the base is frozen with
`extra="forbid"`, so a serialized computed field would break round-trip
validation).

## Issue Number

Closes #4038

## How to Test

Unit + producer coverage:

```
uv run pytest tests/sdk/conversation/test_path_rules_injection.py \
              tests/tools/test_path_rules_file_editor.py \
              tests/tools/test_affected_paths.py -q
# 22 passed
```

Broader regression sweep (base-schema + producer changes):

```
uv run pytest tests/sdk/tool tests/tools/{file_editor,apply_patch,gemini,grep,glob} \
              tests/sdk/conversation -q
# 1226 passed
```

End-to-end (the previously-missed case) — drive a **real Gemini `write_file`**
observation through a live `LocalConversation` and confirm the rule fires:

```
getattr(action, 'path'): None            <- what the old seam saw (missed it)
observation.affected_paths: ['/.../src/api/users.ts']
...
RESULT: rule fired for a Gemini write_file. activated_path_rules = ['api']
```

`pre-commit run --all-files` scoped to the changed files passes (ruff, pyright,
import-dependency rules, tool-registration).

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Producer + consumer land together (by design).** Once the seam reads
  `affected_paths`, any file-mutating observation that doesn't override the
  property reports nothing — so all producers are updated in this PR. Third-party
  tools that want path rules should override `affected_paths` on their
  observation.
- **Reads count as touches.** `read_file` / file-editor `view` report their path
  (consistent with the pre-existing file-editor behaviour). Erroring ops report
  nothing. Happy to gate reads out if reviewers prefer writes-only.
- The core `openhands-sdk` → `openhands-tools` layering is preserved (no new
  cross-package imports; verified by the import-rules hook).

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -5,7 +5,7 @@ import copy
 import json
 import uuid
 from collections.abc import Mapping, Sequence
-from pathlib import Path, PurePath
+from pathlib import Path
 from typing import Any, Final, TypeGuard, cast
 
 from openhands.sdk.agent.acp_agent import ACPAgent
@@ -72,6 +72,7 @@ from openhands.sdk.security.confirmation_policy import (
     ConfirmationPolicyBase,
 )
 from openhands.sdk.skills import load_available_skills, merge_skills_by_name
+from openhands.sdk.skills.skill import normalize_rule_path
 from openhands.sdk.skills.utils import (
     expand_mcp_variables,
     expand_variable_references,
@@ -495,8 +496,9 @@ class LocalConversation(BaseConversation):
             return event
 
         contents: list[TextContent] = []
+        working_dir = self.workspace.working_dir
         for raw_path in event.observation.affected_paths:
-            if (file_path := self._normalize_rule_path(raw_path)) is None:
+            if (file_path := normalize_rule_path(raw_path, working_dir)) is None:
                 continue
             result = agent_context.get_tool_use_suffix(
                 file_path=file_path,
@@ -513,32 +515,6 @@ class LocalConversation(BaseConversation):
         return event.model_copy(
             update={"extended_content": list(event.extended_content) + contents}
         )
-
-    def _normalize_rule_path(self, raw_path: str) -> str | None:
-        """Return the workspace-relative POSIX form of ``raw_path``.
-
-        Returns None when the path is empty or resolves outside the workspace
-        (rules are repo-scoped).
-        """
-        if not raw_path:
-            return None
-
-        # Use the native path flavour so Windows drive paths (e.g. ``D:\\...``) are
-        # recognized as absolute; emit a POSIX string for glob matching.
-        raw = PurePath(raw_path)
-        if not raw.is_absolute():
-            return raw.as_posix()
-        root = PurePath(self.workspace.working_dir)
-        with contextlib.suppress(ValueError):
-            return raw.relative_to(root).as_posix()
-        # Fall back to filesystem resolution so a symlinked workspace root (e.g.
-        # macOS /tmp -> /private/tmp) still matches; strict=False keeps a
-        # not-yet-created leaf.
-        with contextlib.suppress(ValueError, OSError):
-            resolved = Path(raw_path).resolve()
-            resolved_root = Path(self.workspace.working_dir).resolve()
-            return resolved.relative_to(resolved_root).as_posix()
-        return None  # touched a file outside the workspace; rules are repo-scoped
 
     def _recover_persisted_client_tools(
         self,
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -484,8 +484,8 @@ class LocalConversation(BaseConversation):
     def _maybe_inject_path_rules(self, event: Event) -> Event:
         """Return ``event`` with matching path-rule content, or unchanged.
 
-        Only ``ObservationEvent``s carrying a file path are considered. Matching
-        rules already injected in this conversation are skipped via
+        Only ``ObservationEvent``s that report ``affected_paths`` are considered.
+        Matching rules already injected in this conversation are skipped via
         ``state.activated_path_rules``.
         """
         if not isinstance(event, ObservationEvent):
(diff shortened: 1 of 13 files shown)


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -494,40 +494,33 @@ def _maybe_inject_path_rules(self, event: Event) -> Event:
         if (agent_context := self.agent.agent_context) is None:
             return event
 
-        if (file_path := self._touched_rule_path(event)) is None:
-            return event
+        contents: list[TextContent] = []
+        for raw_path in event.observation.affected_paths:
+            if (file_path := self._normalize_rule_path(raw_path)) is None:
```
enyst: I wonder, do we have to do this in local_conversation? It’s becoming a fat all-things class; also below, maybe a function to normalize paths doesnt belong here?

tests/sdk/conversation/test_path_rules_injection.py
```diff
@@ -35,15 +38,29 @@ class _NoPathAction(Action):
     command: str = "ls"
 
 
-def _conversation(tmp_path: Path, rule: Skill) -> LocalConversation:
+class _FileObservation(Observation):
+    """Minimal path-bearing observation (stands in for a file-tool result).
```
enyst, line 42: should this be a protocol…? 🤔

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
VascoSch92: extracted :-)

tests/sdk/conversation/test_path_rules_injection.py
VascoSch92: @enyst 

Kept it as a property on the `Observation` base rather than a Protocol. 

A Protocol makes sense when types don't share a base, but every observation already subclasses `Observation`, and the seam already gets it typed (`ObservationEvent.observation: Observation`). 

The base default (`return []`) makes `affected_paths` total, so the consumer just iterates with no isinstance/hasattr guard. 

Or wdyt?

PR #4068, opened 2026-07-10
fix(sdk): keep legacy history on resume when the stored tail is a non-tree artifact

### LINKED ISSUE #4057: Resuming a pre-event-tree conversation on 1.33.0 orphans prior history (stamps a false `__root__`)
## Summary

Resuming a conversation created by an **older agent-server (pre event-tree, e.g. 1.29.3)** under **1.33.0** can make the agent lose all prior history: it rebuilds context from only the events written after resume, and behaves as if the conversation is brand-new. **No data is lost** — every event file is intact on disk (the web UI still shows them) — but the agent's `path_to_root(leaf)` walk stops early, so the older events are orphaned from the LLM's view.

## Root cause

1.33.0 builds the agent's view by walking the event tree from the leaf up via `parent_id` (`EventStore.path_to_root` → `_effective_parent_id`). Legacy events (no `parent_id`) are meant to fall back to the previous file, so old flat logs load unbranched:

```python
# conversation/event_store.py  _effective_parent_id
if event.parent_id == ROOT_PARENT_ID:   # "__root__" → hard stop
    return None
if event.parent_id is not None:
    return event.parent_id
if idx == 0:
    return None
return self.get_id(idx - 1)              # legacy linear fallback
```

The break happens when the **first event written after resume** is stamped `parent_id="__root__"` instead of being chained onto the legacy tail. That comes from `ConversationState` on resume:

```python
# conversation/state.py  _resolve_active_leaf
if self.leaf_event_id is not None: return self.leaf_event_id
if self.head_is_empty:             return None
if (n := len(self._events)) == 0:  return None
if self._events[n - 1].parent_id is not None:   # ← trap
    return None
return self._events.get_id(n - 1)

# conversation/state.py  _stamp_parent_id
parent = self._resolve_active_leaf()
if parent is None and len(self._events) > 0:
    parent = ROOT_PARENT_ID          # ← stamps a false root
```

On resume `leaf_event_id` is unset (older versions didn't persist it). If the **last stored event is not a bare legacy event** — e.g. it's a `ConversationStateUpdateEvent` or a `ConversationErrorEvent` that already carries a `parent_id` (such artifacts get written during a failed/interrupted startup) — then `_events[n-1].parent_id is not None` is true, `_resolve_active_leaf` returns `None`, and `_stamp_parent_id` concludes "empty head over a non-empty log ⇒ new root" and stamps `__root__`. The tree walk then halts at that event and never reaches (or falls back into) the older events.

## Reproduction (observed)

- Conversation created under 1.29.3 (5,566 flat events, no `parent_id`); last stored event was a state/error artifact carrying a `parent_id`.
- Resumed under 1.33.0; first new user message stamped `parent_id="__root__"`.
- Agent's reachable branch = 165 events (2 user messages); the other 5,566 events are on disk but unreachable by `path_to_root`.

## Impact

Any conversation created before the event-tree feature and resumed on 1.33.0 can silently present as brand-new if its stored tail isn't a bare legacy event. Data is recoverable — it's a link, not a delete — but the agent operates without its history until repaired.

## Suggested fix

On resume with an unset `leaf_event_id`, resolve the active leaf to the **last real event** (skipping `ConversationStateUpdateEvent` / error artifacts) and chain the first new event onto it, rather than treating "last event already has a `parent_id`" as grounds for a brand-new root. That keeps legacy tails linked and preserves the existing legacy fallback.

## Recovery for an already-affected conversation

Re-link the offending event's `parent_id` from `"__root__"` to the id of the last pre-resume event; the legacy fallback then chains the rest. (In one real case this restored the reachable branch from 165 → 5,731 events, back to index 0.)

## Detailed walkthrough (before → after, with diagrams)

Full explanation with the linear-log → event-tree contrast, a worked trace, and the exact stop point:
**https://enyst.github.io/arch/eventlog-resume-orphaned-history.html**

_Environment: `openhands-sdk` / `openhands-agent-server` 1.33.0; conversation originally created under 1.29.3._

VascoSch92:
HUMAN:

Fixing resume of convo.

---

AGENT:

## Why

Resuming a conversation created by an older agent-server (pre event-tree, e.g. `1.29.3`) under `1.33.0+` can make the agent lose all prior history: it rebuilds context only from events written after resume and behaves as if the conversation is brand-new. No data is lost on disk — every event file is intact — but the agent's `path_to_root(leaf)` walk stops early, orphaning the older events from the LLM's view.

Root cause: on a legacy resume `leaf_event_id` is unset (older servers didn't persist it), so `ConversationState._resolve_active_leaf` falls back to the last stored event. It treated *"the last stored event has a `parent_id`"* as grounds for a brand-new root. When the stored tail is a **non-tree artifact** — a `ConversationStateUpdateEvent` or `ConversationErrorEvent` a newer server writes onto the legacy tail during a failed/interrupted startup — that check is true, so `_resolve_active_leaf` returns `None` and `_stamp_parent_id` stamps the first post-resume event with `parent_id="__root__"`. `_effective_parent_id` then halts at that false root and never reaches (or falls back into) the older events.

## Summary

- `_resolve_active_leaf`: on a legacy resume (unset `leaf_event_id`), skip trailing non-tree artifacts (`ConversationStateUpdateEvent` / `ConversationErrorEvent`) and resolve to the last **real** event, so the first new event chains onto the legacy tail instead of a false `__root__`. Keeps the existing legacy linear fallback; no disk rewrite.
- Added parametrized unit regression tests: the two named artifact-tail types, plus multiple stacked artifacts and a trailing event type `_resolve_active_leaf` does not special-case (e.g. `PauseEvent`) — all fail on the pre-fix code.
- Added a **realistic** restore test in `tests/cross/` (per review): a conversation persisted in the pre-event-tree shape *derived from a real run* (strip `parent_id` off every event, drop `leaf_event_id`/`head_is_empty` from `base_state`) with an artifact tail, resumed and continued.

## Issue Number

Fixes #4057

## How to Test

**End-to-end reproduction through a real `LocalConversation` resume** (not a unit test). Save as `repro.py` and run `uv run python repro.py`:

```python
import tempfile

from openhands.sdk.event.conversation_error import ConversationErrorEvent
from openhands.sdk.event.llm_convertible import MessageEvent
from openhands.sdk.llm import LLM, Message, TextContent


def msg(t):
    return MessageEvent(
        source="user",
        llm_message=Message(role="user", content=[TextContent(text=t)]),
    )


with tempfile.TemporaryDirectory() as tmp:
    agent = Agent(
        llm=LLM(model="gpt-4o-mini", api_key=SecretStr("x"), usage_id="t"), tools=[]
    )
    conv = Conversation(
        agent=agent, persistence_dir=tmp, workspace=tmp,
        visualizer=None, delete_on_close=False,
    )
    legacy = [msg(f"legacy-{i}") for i in range(12)]
    for e in legacy:
        conv.state.events.append(e)  # pre-tree flat log (no parent_id)
    conv.state.events.append(  # artifact tail from an interrupted startup
        ConversationErrorEvent(
            source="environment", parent_id=legacy[-1].id,
            code="RuntimeError", detail="boom",
        )
    )
    cid = conv.id
    conv.close()

    resumed = Conversation(
        agent=agent, persistence_dir=tmp, workspace=tmp,
        conversation_id=cid, visualizer=None, delete_on_close=False,
    )
    with resumed._state:
        resumed._on_event(msg("first message after resume"))
    leaf = resumed.state.leaf_event_id
    reachable = len(resumed.state.events.path_to_root(leaf))
    print(f"reachable from agent's view = {reachable} / {len(resumed.state.events)} on disk")
```

- On `main` (pre-fix): `reachable from agent's view = 1 / 14 on disk` — the 12 legacy events + artifact are orphaned; the agent sees only the new message.
- On this branch (fixed): `reachable from agent's view = 13 / 14 on disk` — the full legacy history is back on the active branch. (The non-tree artifact is intentionally kept off the LLM branch — it is not an LLM event — so nothing is lost on disk.)

**Regression tests** (fail on `main`, pass here):

```
uv run pytest tests/sdk/conversation/local/test_conversation_tree.py -k legacy_resume -q
```

- With the fix: `4 passed` — two named artifact-tail types, multiple stacked artifacts, and a non-special-cased `PauseEvent` tail.
- Reverting only `_resolve_active_leaf`: those `4 failed` (the tail carries a `parent_id`, so the pre-fix code stamps a false `__root__`).

**Realistic restore test** in `tests/cross/` (per review — real run downgraded to the pre-event-tree on-disk shape, then resumed):

```
uv run pytest tests/cross/test_conversation_restore_behavior.py::test_restore_pre_event_tree_conversation_with_artifact_tail_keeps_history -q
```

- With the fix: `1 passed`. Reverting `_resolve_active_leaf`: `1 failed` (resumes with an empty active branch — the orphaning).

**Broader suite + lint**, all green:

```
uv run pytest tests/sdk/conversation/local/test_conversation_tree.py tests/sdk/conversation/test_event_tree.py tests/cross/test_conversation_restore_behavior.py -q
# 53 passed

uv run pre-commit run --files openhands-sdk/openhands/sdk/conversation/state.py tests/sdk/conversation/local/test_conversation_tree.py
# Ruff format / Ruff lint / pycodestyle / pyright / import rules — all Passed
```

## Video/Screenshots

N/A — SDK library change; terminal logs are inline under **How to Test**.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Scope: preventive fix.** It stops new orphaning for any legacy conversation resumed from now on. A conversation that was **already** corrupted by an earlier resume has `leaf_event_id` persisted pointing at the false `__root__`, so `_resolve_active_leaf` returns early and it stays orphaned until repaired. Recovery for those: re-link the offending event's `parent_id` from `"__root__"` to the id of the last pre-resume event; the legacy fallback then chains the rest. A conservative load-time auto-heal for already-affected conversations is feasible as a follow-up (keyed on the corruption fingerprint: the false `__root__` sits immediately after a non-tree artifact over a legacy prefix).
- **The hardcoded artifact set is not load-bearing for data safety.** `_resolve_active_leaf` skips `ConversationStateUpdateEvent` / `ConversationErrorEvent` to keep known non-tree artifacts *off* the active lineage, but preventing orphaning does not depend on that list being exhaustive: returning *any* trailing event as the leaf lets `path_to_root` walk back through the full legacy history. Verified that a `PauseEvent` / `InterruptEvent` tail (not in the list) and several stacked artifacts all keep history intact — orphaning is only reachable in the degenerate case where the log contains no real event at all (nothing to lose).
- Detailed walkthrough of the bug: https://enyst.github.io/arch/eventlog-resume-orphaned-history.html

### DIFF
--- a/tests/cross/test_conversation_restore_behavior.py
+++ b/tests/cross/test_conversation_restore_behavior.py
@@ -35,6 +35,8 @@ from openhands.sdk.context.condenser.llm_summarizing_condenser import (
 )
 from openhands.sdk.conversation.impl.local_conversation import LocalConversation
 from openhands.sdk.event import ActionEvent, MessageEvent
+from openhands.sdk.event.conversation_state import ConversationStateUpdateEvent
+from openhands.sdk.event.types import ROOT_PARENT_ID
 from openhands.sdk.llm import LLM
 from openhands.sdk.llm.utils.openhands_provider import (
     LITELLM_PROXY_PREFIX,
@@ -829,5 +831,94 @@ def test_restore_reasoning_effort_none_strips_temperature(mock_completion):
             assert last_call["reasoning_effort"] == "none"
             assert "temperature" not in last_call
             assert "top_p" not in last_call
+        finally:
+            restored.close()
+
+
+@patch("openhands.sdk.llm.llm.litellm_completion")
+def test_restore_pre_event_tree_conversation_with_artifact_tail_keeps_history(
+    mock_completion,
+):
+    """Resuming a pre-event-tree conversation whose tail is a startup artifact
+    must keep the full history (#4057).
+
+    Older agent-servers (pre event-tree) wrote events without ``parent_id`` and
+    never persisted ``leaf_event_id``. We reproduce that shape from a *real* run:
+    strip ``parent_id`` off every event and drop the tree HEAD markers from
+    ``base_state``, then append the kind of non-tree artifact a newer server
+    writes onto the tail during an interrupted startup (a
+    ``ConversationStateUpdateEvent`` carrying a ``parent_id``). On resume the
+    agent must load the whole prior history and chain onto its real tail, rather
+    than stamping the next event a false ``__root__`` that orphans everything.
+    """
+    mock_completion.return_value = create_mock_litellm_response(
+        content="Acknowledged.", finish_reason="stop"
+    )
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        base = Path(temp_dir)
+        lifecycle = RestoreLifecycle(
+            workspace_dir=base / "workspace",
+            persistence_base_dir=base / "persist",
+        )
+        lifecycle.workspace_dir.mkdir(parents=True, exist_ok=True)
+        lifecycle.persistence_base_dir.mkdir(parents=True, exist_ok=True)
+
+        tools = [Tool(name="TerminalTool"), Tool(name="FileEditorTool")]
+        agent = _agent(
+            llm_model="gpt-4o-mini",
+            tools=tools,
+            condenser_max_size=80,
+            skill_name="skill-v1",
+            skill_keyword="alpha",
+        )
+
+        # 1) Produce a realistic event log by actually running a session.
+        lifecycle.run_initial_session(agent)
+        assert lifecycle.conversation_id is not None
+        events_dir = (
+            lifecycle.persistence_base_dir / lifecycle.conversation_id.hex / "events"
+        )
+        event_files = sorted(
+            events_dir.glob("event-*.json"),
+            key=lambda p: int(p.name.split("-")[1]),
+        )
+        assert len(event_files) >= 3
+        original_ids = [json.loads(p.read_text())["id"] for p in event_files]
+
+        # 2) Downgrade to the pre-event-tree shape: strip parent_id off every
+        #    event and drop the tree HEAD markers older servers never wrote.
+        for p in event_files:
+            data = json.loads(p.read_text())
+            data.pop("parent_id", None)
+            p.write_text(json.dumps(data))
+        base_state = lifecycle.read_base_state()
+        base_state.pop("leaf_event_id", None)
+        base_state.pop("head_is_empty", None)
+        lifecycle.write_base_state(base_state)
+
+        # 3) Append the interrupted-startup artifact: a non-tree event a newer
+        #    server chained onto the legacy tail (it carries a parent_id).
+        artifact = ConversationStateUpdateEvent(
+            parent_id=original_ids[-1], key="execution_status", value="idle"
+        )
+        artifact_path = events_dir / f"event-{len(event_files):05d}-{artifact.id}.json"
+        artifact_path.write_text(artifact.model_dump_json(exclude_none=True))
+
+        # 4) Resume under the current (tree-aware) code.
+        restored = lifecycle.restore(agent)
+        try:
+            # The whole prior history is the active branch (nothing orphaned);
+            # the non-tree artifact is not on it.
+            active_ids = [e.id for e in restored.state.active_branch()]
+            assert active_ids == original_ids
+
+            # Continuing chains onto the real legacy tail, not a false __root__.
+            restored.send_message("Third message")
+            new_event = restored.state.events[-1]
+            assert new_event.parent_id == original_ids[-1]
+            assert new_event.parent_id != ROOT_PARENT_ID
+            reachable = {e.id for e in restored.state.events.path_to_root(new_event.id)}
+            assert set(original_ids).issubset(reachable)
         finally:
             restored.close()
--- a/tests/cross/test_conversation_restore_behavior.py
+++ b/tests/cross/test_conversation_restore_behavior.py
@@ -922,3 +922,73 @@ def test_restore_pre_event_tree_conversation_with_artifact_tail_keeps_history(
             assert set(original_ids).issubset(reachable)
         finally:
             restored.close()
+
+
+@patch("openhands.sdk.llm.llm.litellm_completion")
+def test_restore_event_tree_conversation_keeps_history(mock_completion):
+    """Control for the pre-event-tree case: a normal 1.33.0+ restore is unaffected.
+
+    A conversation persisted by an event-tree server (events carry ``parent_id``,
+    ``base_state`` persists ``leaf_event_id``) is restored *without* any downgrade
+    and must keep its full history and chain onto the real tail. Because
+    ``leaf_event_id`` is set, ``_resolve_active_leaf`` returns it directly and the
+    legacy-resume path the #4057 fix touches is never entered — this guards that
+    normal restore stays intact.
+    """
+    mock_completion.return_value = create_mock_litellm_response(
+        content="Acknowledged.", finish_reason="stop"
+    )
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        base = Path(temp_dir)
+        lifecycle = RestoreLifecycle(
+            workspace_dir=base / "workspace",
+            persistence_base_dir=base / "persist",
+        )
+        lifecycle.workspace_dir.mkdir(parents=True, exist_ok=True)
+        lifecycle.persistence_base_dir.mkdir(parents=True, exist_ok=True)
+
+        tools = [Tool(name="TerminalTool"), Tool(name="FileEditorTool")]
+        agent = _agent(
+            llm_model="gpt-4o-mini",
+            tools=tools,
+            condenser_max_size=80,
+            skill_name="skill-v1",
+            skill_keyword="alpha",
+        )
+
+        lifecycle.run_initial_session(agent)
+        assert lifecycle.conversation_id is not None
+        events_dir = (
+            lifecycle.persistence_base_dir / lifecycle.conversation_id.hex / "events"
+        )
+        event_files = sorted(
+            events_dir.glob("event-*.json"),
+            key=lambda p: int(p.name.split("-")[1]),
+        )
+        assert len(event_files) >= 3
+        original_ids = [json.loads(p.read_text())["id"] for p in event_files]
+
+        # This IS the modern (event-tree) on-disk shape — leaf persisted and
+        # events carry parent_id (contrast the pre-event-tree test above). No
+        # downgrade, no injected artifact.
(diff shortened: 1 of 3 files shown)


smolpaws (commented):
Reviewing as the reporter of #4057 — I diagnosed this exact bug from a live incident, so I can speak to whether the fix addresses the real root cause. **It does.** Leaving this as a COMMENT rather than an approval because it changes history-reconstruction behavior (eval-risk category per the repo's review guide) — flagging for a maintainer to run lightweight evals before merge.

### The fix is correct and minimal

The root cause was precise: on a legacy resume `leaf_event_id` is unset, so `_resolve_active_leaf` fell back to the last stored event — but if that tail was a **non-tree artifact carrying a `parent_id`** (a `ConversationStateUpdateEvent` or `ConversationErrorEvent` a newer server wrote onto the legacy tail during an interrupted startup), the old check `self._events[n-1].parent_id is not None → return None` fired, and `_stamp_parent_id` then stamped the first post-resume event with `parent_id="__root__"`. `_effective_parent_id` halts at that false root, stranding all prior history.

Walking back from the last *real* event instead of bailing on "tail has a parent_id" is exactly right, and it preserves the existing legacy linear fallback with no disk rewrite. In my real case the orphaned conversation had 5,566 pre-tree events behind such an artifact tail (the tail was literally a `ConversationErrorEvent: "Streaming requires an on_token callback"` from the failed old-server startup) — this change would have chained onto it correctly.

### One consistency nit (non-blocking)

The new docstring says these artifacts "are not tree nodes (see `append_event`)". That's only strictly true for `ConversationStateUpdateEvent` — `append_event` excludes *only* that type from advancing `leaf_event_id`:

```python
if not isinstance(event, ConversationStateUpdateEvent):
    self.leaf_event_id = event.id
```

A `ConversationErrorEvent` **does** advance HEAD in a live session, so it isn't a "non-tree node" in the same sense — it's a real tree node that you're (correctly) choosing to skip *only* in the unset-leaf resume path. The behavior is right; the wording just slightly overstates the parallel to `append_event`. Consider tightening the comment to "artifacts we skip when resolving a resume leaf" rather than implying `append_event` already treats both as non-nodes. Worth a one-line follow-up thought: if the two lists (HEAD-skip vs resume-skip) drift, a shared predicate/constant would keep them honest.

### Test coverage

The parametrized regression test covers both artifact-tail types and asserts the important invariants: the artifact is skipped, the active branch is the full legacy history, and the first post-resume event chains onto the last real legacy event (not a false `__root__`) with `path_to_root` returning the whole chain. That's the right shape and it exercises a real `LocalConversation` resume, not a mock. 

One suggestion: add a case where the tail is **two** stacked artifacts (e.g. an error followed by a state-update, which is exactly the ordering I saw on disk: `ConversationErrorEvent` at index N, then a `ConversationStateUpdateEvent` at N+1). The current loop handles it (it walks back over any run of artifacts), but a test would lock that in.

### Recovery note for anyone already affected

For conversations already stranded by this bug, the on-disk data is intact — it's recoverable by re-linking the offending event's `parent_id` from `"__root__"` to the last pre-resume event id; the legacy fallback then chains the rest. (In my case that restored the reachable branch from 165 → 5,731 events.)

Detailed before→after walkthrough with diagrams: https://enyst.github.io/arch/eventlog-resume-orphaned-history.html

**Verdict:** correct fix for the root cause, good test, minimal surface. Recommend a maintainer confirm evals given it touches what the agent sees in context, then merge. 🐾

enyst (commented):
We have in `cross` test directory some tests that are specifically for conversation restore, with realistic old version / new version events. Maybe we can make one for this?

enyst:
@openhands look at the linked issue and this PR. /codereview this PR and post your feedback as review with gh api.

Note that the issue was written by an AI, and you know those can hallucinate, verify it yourself. You are an agent, you can install yourself carefully everything, the former version, the new version., and you can use openai with gpt-5-nano I think, for a real conversation with a message (e.g. read the readme) and responses and restore and message after restore. You can enable telemetry/logging and save as json so you can see what exactly happened.

tests/cross/test_conversation_restore_behavior.py
```diff
@@ -831,3 +833,92 @@ def capture_completion(*_args: Any, **kwargs: Any):
             assert "top_p" not in last_call
         finally:
             restored.close()
+
+
+@patch("openhands.sdk.llm.llm.litellm_completion")
+def test_restore_pre_event_tree_conversation_with_artifact_tail_keeps_history(
+    mock_completion,
```
enyst, line 840: I like this test! How about we add a similar one but for normal restore (on 1.33.0+), to make sure we didn’t break anything?

enyst (approved):
🟢 **Taste Rating: Good taste**

This is the right shape of fix: small, localized, and preserves the existing legacy-linear fallback instead of rewriting persisted histories or adding a migration step. I independently verified the issue is real rather than just trusting the AI-written report:

- On `origin/main`, the local reproduction resumes with `reachable=1 total=14 parent=__root__`, confirming the false-root orphaning.
- On this PR branch, the same reproduction resumes with `reachable=13 total=14` and chains the next event to the legacy tail.
- Focused regression tests passed: `uv run pytest tests/sdk/conversation/local/test_conversation_tree.py -k legacy_resume -q` → 4 passed.
- Broader focused suite passed: `uv run pytest tests/sdk/conversation/local/test_conversation_tree.py tests/sdk/conversation/test_event_tree.py tests/cross/test_conversation_restore_behavior.py -q` → 53 passed.
- Targeted pre-commit passed on the touched files: Ruff format/lint, pycodestyle, pyright, import rules, and tool registration checks.

I do not see a blocking correctness issue. The hardcoded skip list is a little special-casey, but it is constrained to resume-time legacy compatibility and the tests cover the important safety property: even unknown trailing event types do not orphan the old history. No dependencies or public API surface changes are involved.

**[RISK ASSESSMENT]**

This touches core conversation resume/tree lineage logic, so the blast radius is higher than a leaf-level bug fix. The implementation is nevertheless localized, backward-compatible, and covered by both unit-level legacy-tail regressions and an integration-like restore test derived from a real persisted run shape. No security-sensitive logic, dependency changes, or external-service integrations are introduced.

**VERDICT:**
✅ **Worth merging**: Core logic is sound and the regression evidence is strong.

**KEY INSIGHT:**
The important architectural choice is preserving legacy flat-log semantics at leaf resolution time, without mutating historical event files or widening the persistence contract.

---

_This review was created by an AI agent (OpenHands) on behalf of @openhands._

PR #4090, opened 2026-07-11
fix: heal history orphaned by pre-#4068 legacy-resume __root__ bug

smolpaws:
HUMAN:
This PR proposes a one-time migration of the affected conversations fixed by 4068

---

AGENT:

## Why

Before #4068, resuming a **pre-event-tree** conversation whose stored tail was a
non-tree artifact (a `ConversationStateUpdateEvent` / `ConversationErrorEvent`
written during an interrupted startup) could stamp the first post-resume event
with `parent_id == "__root__"`. `_effective_parent_id` treats `"__root__"` as a
hard root, so `path_to_root` stops there and **every older event is orphaned from
the LLM's view** — even though all event files are intact on disk.

#4068 stops *new* resumes from producing this, but conversations **already written**
with the bad `__root__` stay broken until repaired. Users are hitting this in the
wild (reported by @kripper: orphaned history on resume across the 1.29.3→1.33.x
boundary). There is currently no automatic recovery — it requires a hand-edit of a
single event file.

## Summary

- **`EventLog.rewrite_event(idx, event)`** — overwrite an already-persisted event in
  place (same idx + id, filename unchanged, under the write lock) so a one-time
  migration can repair a single field without appending or reordering the log.
- **`ConversationState._migrate_orphaned_root_tail()`** — a tightly-guarded on-disk
  repair, run once on the resume path (before `rebuild_view`). It heals exactly the
  bug's fingerprint and re-links the wrongly-rooted event to the legacy tail.
- Tests: 2 unit + 1 cross (see below).

### What gets healed (and what does not)

An event at `idx > 0` with `parent_id == "__root__"` is repaired **only if**:
1. it sits directly after a **non-tree artifact** (`ConversationStateUpdateEvent` /
   `ConversationErrorEvent`) — the stale tail that caused the bug; a deliberate
   `navigate_to(None)` root sits after a *real* event, so this excludes it; **and**
2. every real (non-artifact) event before it carries **no** `parent_id` (a genuine
   legacy log, not a real tree with branches).

When both hold, the event is re-linked to the nearest preceding real event,
reconnecting history. Genuine tree roots, deliberate `navigate_to(None)` roots, and
real branch points are left untouched.

## Issue Number

Related to the orphaned-history-on-resume reports (OpenHands/OpenHands#15247 is a
neighbouring resume-robustness issue in the same area). This PR is the one-time
migration for the `__root__`-at-the-tail case.

## How to Test

```bash
# The cross test reproduces the bug from a REAL session, then resumes and proves the heal:
uv run python -m pytest tests/cross/test_conversation_restore_behavior.py \
  -k test_restore_migrates_orphaned_root_and_restores_history -q

# Unit tests: heal the damaged shape, and leave a deliberate root untouched:
uv run python -m pytest tests/sdk/conversation/local/test_conversation_tree.py \
  -k "migration_heals or migration_leaves" -q

# Full relevant suites:
uv run python -m pytest tests/sdk/conversation/local/test_conversation_tree.py \
  tests/cross/test_conversation_restore_behavior.py -q
```

**Result on this branch:**
- `test_restore_migrates_orphaned_root_and_restores_history` — PASS
- `test_migration_heals_orphaned_root_over_legacy_tail` — PASS
- `test_migration_leaves_deliberate_feature_root_untouched` — PASS
- Full suites: **42 passed**
- `pyright`: 0 errors; `ruff` lint + format: clean; all pre-commit hooks pass.

The cross test is not a mock: it runs a real `LocalConversation` session to produce
a genuine event log, downgrades it to the pre-tree shape, appends the interrupted
artifact + the wrongly-`__root__` post-resume event, asserts history is orphaned
*before* migration, then resumes under this code and asserts (a) the event is
re-linked to the legacy tail, (b) all legacy events are reachable again in order,
(c) continuing the conversation chains onto the healed tail, and (d) the repair is
durable on disk after reopening the log.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The migration is idempotent: after healing, the event no longer has `__root__`, so
  a subsequent resume is a no-op.
- It only touches the resume path, never fresh conversation creation.
- Only rewrites the single offending event's `parent_id`; it never bulk-strips
  `parent_id`s or reorders the log.

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/event_store.py
+++ b/openhands-sdk/openhands/sdk/conversation/event_store.py
@@ -222,6 +222,34 @@ class EventLog(EventsListBase):
             )
             raise
 
+    def rewrite_event(self, idx: int, event: Event) -> None:
+        """Overwrite the already-persisted event at ``idx`` in place.
+
+        The event keeps its index and id (the filename is unchanged); only its
+        stored payload is replaced. Used by one-time on-disk migrations that
+        repair a single event's fields (e.g. re-linking a wrongly-rooted
+        ``parent_id``) without appending or reordering the log.
+
+        Raises:
+            IndexError: If ``idx`` is out of range.
+            ValueError: If ``event.id`` does not match the event stored at ``idx``.
+        """
+        if idx < 0 or idx >= self._length:
+            raise IndexError(f"Event index out of range: {idx}")
+        existing_id = self._idx_to_id[idx]
+        if event.id != existing_id:
+            raise ValueError(
+                f"rewrite_event id mismatch at idx {idx}: stored '{existing_id}', "
+                f"got '{event.id}'"
+            )
+        payload = event.model_dump_json(exclude_none=True)
+        write_guard = (
+            nullcontext() if self._write_guard is None else self._write_guard()
+        )
+        with write_guard:
+            self._fs.write(self._path(idx, event_id=existing_id), payload)
+        self._event_cache[idx] = event
+
     def _count_events_on_disk(self) -> int:
         """Count event files on disk."""
         try:
--- a/openhands-sdk/openhands/sdk/conversation/state.py
+++ b/openhands-sdk/openhands/sdk/conversation/state.py
@@ -260,6 +260,80 @@ class ConversationState(OpenHandsModel):
     def events(self) -> EventLog:
         return self._events
 
+    def _migrate_orphaned_root_tail(self) -> None:
+        """One-time on-disk repair for history orphaned by the pre-#4068 bug.
+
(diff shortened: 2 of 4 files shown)


smolpaws:
Closing this for now. The migration rewrites an already-persisted event's `parent_id` during restore via `EventLog.rewrite_event()`. Event history is append-only: once an event is persisted, we should not mutate it in place. We'll first verify whether affected histories remain after agent-canvas moved to SDK 1.35.0, then revisit with an append-only repair if one is still needed.

PR #4100, opened 2026-07-13
Lazily hydrate persisted conversations

### LINKED ISSUE #3140: perf: agent server loads all conversations eagerly at startup
## Problem

On startup, `ConversationService.__aenter__` iterates every conversation directory on disk and calls `_start_event_service` for each one. Each call creates a `LocalConversation` (with agent, LLMs, event log), a lease-renew background task, and subscriber registrations. With 1000 persisted conversations, startup creates 1000 full conversation objects serially.

**Location:** `conversation_service.py:742–785`

## Evidence

Code review of `conversation_service.py:742–785`. The startup loop is serial with no concurrency limit or lazy loading.

**Severity:** High

## Impact of fix

Startup time drops from O(N) to O(1). Memory at startup is bounded regardless of how many conversations are persisted on disk.

## Suggested fix

Load only conversation metadata at startup. Hydrate the full `LocalConversation` on first access (lazy loading). Consider LRU eviction for idle conversations.

## Related issues and PRs

- [OpenHands#13193](https://github.com/OpenHands/OpenHands/pull/13193) — add auto_pause_existing flag and fix sandbox eviction order (open PR, eviction for sandboxes)
- [OpenHands#13892](https://github.com/OpenHands/OpenHands/pull/13892) — lazy-load models per provider (closed PR, similar lazy-loading pattern)
- [OpenHands#9156](https://github.com/OpenHands/OpenHands/issues/9156) — Improve runtime startup performance with container reuse strategies (open issue)

neubig:
HUMAN:
I (@neubig) have tested this and it is working quite well, much faster conversation load times and lower memory usage.

AGENT:

## Why

Persisted conversation startup was hydrating every event history, which raised memory use and delayed readiness for large local corpora.

## How to Test

Run the focused conversation-service and lease tests from the Validation section below. The Live acceptance testing section records a real Agent Server run against 120 persisted conversations and 55200 production-format event files.

## Summary

- load a lightweight persisted-conversation catalog at agent-server startup
- hydrate an `EventService` and event history only when an operation needs a live runtime
- serialize hydration so concurrent requests cannot create duplicate runtimes
- serve search, count, and conversation metadata reads without hydrating event history
- preserve startup recovery for conversations persisted in the running state
- prevent forks from overwriting an unloaded persisted conversation

This is a smaller replacement for #3895.
Part of #3140 

## Root cause

The agent server currently reconstructs every persisted conversation and its event history during startup. On the local corpus used to reproduce the watchdog failures, 136 conversations containing 55,312 event files raise the server to roughly 3 GiB RSS before any active work. A long-running npm test or coverage process can then exhaust the host's 7.8 GiB memory limit, causing an exit-137/OOM kill. The watchdog health failure is a downstream symptom of that memory pressure and process death.

With the catalog-only startup path, the same corpus uses roughly 355 MiB RSS before runtime hydration.

## Impact

Idle historical conversations no longer consume event-history memory. Search and count remain available from the catalog, while runtime operations transparently hydrate the requested conversation. A single hydration lock keeps check-and-load atomic and bounds transient memory pressure.

## Validation

- `uv run pytest -q tests/agent_server/test_conversation_service.py tests/cross/test_conversation_lease_behavior.py -x` — 90 passed
- router, event-service, websocket, socket, and stress regression selection — 264 passed
- repository pre-commit checks for both changed files — passed
- real-corpus startup memory comparison — approximately 355 MiB lazy versus 3 GiB eager

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:808bcbd-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-808bcbd-python \
  ghcr.io/openhands/agent-server:808bcbd-python
```

```
ghcr.io/openhands/agent-server:808bcbd-golang-amd64
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-golang-amd64
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-golang-amd64
ghcr.io/openhands/agent-server:808bcbd-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:808bcbd-golang-arm64
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-golang-arm64
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-golang-arm64
ghcr.io/openhands/agent-server:808bcbd-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:808bcbd-java-amd64
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-java-amd64
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-java-amd64
ghcr.io/openhands/agent-server:808bcbd-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:808bcbd-java-arm64
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-java-arm64
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-java-arm64
ghcr.io/openhands/agent-server:808bcbd-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:808bcbd-python-amd64
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-python-amd64
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-python-amd64
ghcr.io/openhands/agent-server:808bcbd-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:808bcbd-python-arm64
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-python-arm64
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-python-arm64
ghcr.io/openhands/agent-server:808bcbd-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:808bcbd-golang
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-golang
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-golang
ghcr.io/openhands/agent-server:808bcbd-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:808bcbd-java
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-java
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-java
ghcr.io/openhands/agent-server:808bcbd-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:808bcbd-python
ghcr.io/openhands/agent-server:808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb-python
ghcr.io/openhands/agent-server:codex-simplify-lazy-conversation-loading-python
ghcr.io/openhands/agent-server:808bcbd-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `808bcbd-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `808bcbd-python-amd64`) are also available if needed

Closes #3140

## Live evidence

On 2026-07-14 UTC I launched real `python -m openhands.agent_server` processes against a deterministic production-format corpus: 120 persisted conversations, 460 events each, 55,200 event JSON files total, corpus SHA-256 `c6a5867f0bb11bf9d64cd28ab296a657b8ae22bcd2035aeb6ccbb4b8099e64c4`. The benchmark used fetched `origin/main` `56ac31719f91059dc9b319dbc6bb79f17ef60cd7` as the before ref and current PR head `808bcbd5254ee9c7b65fae40147aaa8e8a66b2bb` as the after ref. The harness used real HTTP endpoints plus production `owner_lease.json` files and resume logs; no test-only endpoint or monkeypatch was used.

| Check | Current main | PR head |
| --- | ---: | ---: |
| Startup time | 12.555 s | 5.625 s |
| Startup RSS | 594,624 KiB | 388,560 KiB |
| Hydrated runtimes at readiness | 120 | 0 |
| Resume logs at readiness | 120 | 0 |
| Hydrated runtimes after search/count | 120 | 0 |
| Eight concurrent event-count requests | all returned 460 | all returned 460 |
| Hydrated runtimes after target event-count access | 120 | 1 |

Additional PR checks: duplicate fork to an existing unloaded conversation ID returned HTTP 409 with the target `meta.json` unchanged. Startup with one persisted `running` record hydrated only that record, retained all 120 catalog records, and reported one recovered `error` record.


neubig:
✅ **PR Artifacts Cleaned Up**

The `.pr/` directory has been automatically removed.

enyst:
@openhands Check this PR, see its temporary `.pr/` directory for evidence; /codereview and consider simplicity and maintainability in your review. Post it as a review on gh api on this PR; feel free to comment inline; and to add the event you see fit, approved, comment, request changes; only important stuff, not nits.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -493,31 +502,196 @@ class ConversationService:
...
+        self, conversation_id: UUID, record: _ConversationRecord
+    ) -> ConversationExecutionStatus:
+        event_services = self._event_services
+        if event_services is None:
+            raise ValueError("inactive_service")
+        event_service = event_services.get(conversation_id)
+        if event_service is not None and event_service.is_open():
+            state = await event_service.get_state()
+            record.execution_status = state.execution_status
+        return record.execution_status
```
enyst, line 595: 🔴 **Stale status filters:** unloaded records return the startup-cached status here. With two real services sharing the persistence directory, I reproduced `get_conversation(id)` reading `running` from disk while `count_conversations(RUNNING)` returned `0` and filtered search returned no items on the secondary. Refresh `base_state.json` for unloaded records before status filtering; that remains lightweight and keeps these APIs consistent.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -894,18 +1070,16 @@ async def interrupt_conversation(self, conversation_id: UUID) -> bool:
...
 
     async def delete_conversation(self, conversation_id: UUID) -> bool:
-        if self._event_services is None:
+        event_services = self._event_services
+        if event_services is None:
             raise ValueError("inactive_service")
-        event_service = self._event_services.pop(conversation_id, None)
+        event_service = await self._get_or_load_event_service(conversation_id)
         if event_service:
+            event_services.pop(conversation_id, None)
```
enyst, line 1090: 🔴 **Delete can be undone by a waiting hydration.** A concurrent getter can capture the record before delete, wait on `_hydration_lock`, then acquire it after these pops and hydrate/reinsert the conversation. I reproduced `delete_conversation()` returning `True` while the getter returned a second open service left in both maps. Removal and hydration need the same lifecycle synchronization/reservation boundary.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -1269,6 +1386,7 @@ async def __aexit__(self, exc_type, exc_value, traceback):
         if event_services is None:
             return
         self._event_services = None
+        self._conversation_records = {}
```
enyst: 🔴 **Shutdown can leak an in-flight hydration.** `_start_event_service()` captures the current runtime dict before awaiting startup. If shutdown snapshots an empty dict and clears service state while that await is blocked, hydration later inserts into the detached dict, repopulates the catalog, and returns an open service that teardown never closes. I reproduced this deterministically. Serialize teardown with hydration and re-check active state before publishing; close the just-started service if shutdown won.

enyst (commented):
_This review was generated by an OpenHands AI agent on behalf of @enyst._

The catalog/runtime split is a good simplification and the live evidence supports the startup-memory win, but lifecycle synchronization is incomplete. I left three important inline findings, each reproduced deterministically against the current head:

- status-filtered search/count can disagree with fresh persisted state in shared-storage deployments;
- delete can report success while a waiting request resurrects the runtime and catalog entry;
- shutdown can finish while in-flight hydration publishes an open, unclosed runtime afterward.

Focused regression suites still pass (`90 passed`), so the existing tests do not cover these interleavings. The required `Validate PR description` check is also failing: the human author needs to restore the `HUMAN:` / `AGENT:` template structure and provide the human-written note; an AI agent must not fill that field.

**[RISK ASSESSMENT]**
This changes core persisted-runtime ownership and teardown behavior. The performance approach is sound, but the confirmed races can leak runtimes or make a deleted conversation reappear. **Recommendation**: do not auto-merge; have a human maintainer validate the lifecycle-locking fix and shared-storage semantics.

**VERDICT:**
❌ **Needs rework**: make hydration, removal, and shutdown share one lifecycle synchronization boundary, and refresh unloaded status from `base_state.json` when filtering.

**KEY INSIGHT:**
The hydration lock must protect runtime publication against every competing lifecycle transition, not only duplicate hydration.

---

>
> 1. Add a `.agents/skills/custom-codereview-guide.md` file to your branch (or edit it if one already exists) with the `/codereview` trigger and the context the reviewer is missing. See the [customization docs](https://docs.openhands.dev/openhands/usage/use-cases/code-review#customization).
> 3. When merged, the guideline goes through normal maintainer review.
>
> **Resolve with AI?** Install the [iterate skill](https://github.com/OpenHands/extensions/tree/main/skills/iterate) and run `/iterate` to drive the PR through CI, review, and QA.
>

enyst:
@openhands I’m curious about some parts of the code in this PR: is it all necessary, is it introducing redundancies, does it follow the patterns in the codebase, or, can we think about achieving its goal simpler?

Examples to look into:
- “direct embedders and existing test fixtures” — if test fixtures need updated for what the PR is doing, we should update them, we should NOT introduce code in the main codebase just to keep the same test structures. Is this the case here?
- ConversationRecord dataclass: is it necessary? I seem to recall we have some class for the `meta.json` or do I misremember? can’t we reuse what exists
- `if event_service is not None and event_service.is_open()` — this is repeated in the same method, didn’t we return already? tell me if we can simplify
- is `_reconcile_active_records()` necessary? I saw code that adds to records a conversation in use. If you think it’s necessary, can you make a test to prove that, if this method is not called, then some conversation could, in real-world use, be missed?

Do not modify the PR. Just answer me here on github. Intro yourself first so that readers know who is speaking (you are OpenHands-GPT-5.6)

openhands-agent-server/openhands/agent_server/conversation_service.py
neubig: Addressed in 2e97329c3: status-filtered search and count now refresh base_state.json for unloaded records without hydrating an EventService. Added a shared-storage regression covering both APIs.

openhands-agent-server/openhands/agent_server/conversation_service.py
neubig: Addressed in 2e97329c3: hydration and deletion now share one lifecycle lock, and deletion performs lookup, removal, close, and directory cleanup within that boundary. Added a deterministic waiting-getter regression.

openhands-agent-server/openhands/agent_server/conversation_service.py
neubig: Addressed in 2e97329c3: hydration publication and shutdown now share the lifecycle lock, startup re-checks the active runtime map before publishing, and startup cleanup closes the service if shutdown already won. Added a deterministic in-flight hydration regression.

smolpaws:
I’m **OpenHands-GPT-5.6**, commenting through the `smolpaws` account. I reviewed the current PR head (`2e97329c3`), its base, the follow-up lifecycle fix, and the relevant tests.

Short version: the **lazy catalog/runtime split is the right design**, but the implementation carries compatibility code for tests that mutate private state. I would remove that compatibility rather than preserve it in production.

<details>
<summary>1. “Direct embedders and existing test fixtures” — ❌ Test-only compatibility</summary>

In this repository, `_reconcile_active_records()` accommodates test structure, not a demonstrated production path.

I searched non-test callers. The server constructs `ConversationService` normally and uses public methods; I found no production embedder that injects `_event_services`. In contrast, `test_conversation_service.py` has many tests that write mocked services directly into that private dict.

I also replaced `_reconcile_active_records()` with a no-op at runtime and ran two paths:

- the persisted lifecycle test `test_startup_and_search_do_not_hydrate_idle_conversation` passed;
- `test_search_conversations_basic` failed because it inserts a mock only into `_event_services`, leaving the catalog empty.

That is strong evidence that reconciliation repairs an invalid fixture rather than a real lifecycle gap. Since `_event_services` is private, unknown external mutation should not define the service contract. Update those tests to use the public lifecycle or explicitly establish both sides of the invariant in narrow unit tests.

The missing-record fallback in `get_conversation()` has the same issue: normal lifecycle paths already publish both entries. Quiet repair makes the invariant harder to reason about and can hide real bugs.

</details>

<details>
<summary>2. Is <code>_ConversationRecord</code> necessary? ✅</summary>

You remembered correctly: `StoredConversation` is already the typed model for `meta.json`, and this PR reuses it.

`_ConversationRecord` is not another meta model. It combines:

- `stored: StoredConversation` from `meta.json`; and
- `execution_status`, which lives in `base_state.json`, not `meta.json`.

The dataclass is therefore reasonable if status is intentionally cached for startup recovery and filtering. Replacing it with `StoredConversation` alone would lose the second value or move it to another parallel structure.

The larger redundancy is the **two parallel maps**, `_conversation_records` and `_event_services`. They can drift, which is why reconciliation appeared. The cleanest eventual model is one authoritative record per conversation with an optional runtime, conceptually:

```python
_ConversationRecord(stored, execution_status, event_service=None)
```

That removes cross-map reconciliation. It is a larger refactor, though. The minimal clean change here is to keep the dataclass and two maps, enforce their invariant in lifecycle methods, and fix tests instead of repairing drift during reads.

</details>

<details>
<summary>3. Repeated open-service check ✅</summary>

At the head that existed when the question was posted, the repeated check inside `_get_or_load_event_service()` was necessary double-checking:

1. check before waiting for `_hydration_lock`;
2. await the lock;
3. check again because another coroutine may have hydrated the conversation while this one waited.

The first return does not make the second check unreachable across an `await`.

The follow-up commit `2e97329c3` replaced that hydration lock with a lifecycle lock acquired before the check and moved the body into `_get_or_load_event_service_locked()`. On the current head, that control flow has only one check, so it has already been simplified correctly. Similar guards in `_conversation_info()` and `_execution_status()` are separate read paths, not duplicate branches in one method.

</details>

<details>
<summary>4. Is <code>_reconcile_active_records()</code> necessary? ❌ No real-world need found</summary>

I do **not** think it is necessary under the intended production invariant, and I cannot honestly propose a real-world test proving otherwise from this codebase.

Such a test would need a supported production path where a live `EventService` appears without a catalog record. I found none. `_start_event_service()` publishes both; delete removes both; startup creates the catalog before hydration; shutdown clears both under `_lifecycle_lock`.

The only reproducible missing-record path is direct mutation of `_event_services` in tests. That belongs at the test boundary, not in runtime reconciliation. Reconciliation also runs only in search/count while `get_conversation()` has a separate repair path; this asymmetry is a design smell.

</details>

### Recommended simplification

For this PR, I would:

1. keep `_ConversationRecord`;
2. remove `_reconcile_active_records()` and its search/count calls;
3. remove read-time repair of missing catalog entries;
4. update tests that directly inject `_event_services` to establish valid state;
5. document and enforce: every live runtime has exactly one catalog record, and lifecycle mutation happens under `_lifecycle_lock`.

One additional simplification: `delete_conversation()` currently hydrates an unloaded conversation — including event history and lease — only to close and delete it. An unloaded delete should operate from the catalog/persisted state; only an already-live runtime needs closing. That better follows this PR’s central rule: hydrate only when an operation actually needs a runtime.

I ran seven focused current-head regressions covering lazy search, concurrent hydration, shared-state status refresh, delete/hydration ordering, shutdown/hydration ordering, duplicate fork protection, and the legacy direct-injection search path; all seven pass as written.

**Bottom line:** good performance architecture, but `_reconcile_active_records()` is production complexity serving private test setup. Keep the record abstraction; remove reconciliation/repair behavior; make the catalog authoritative.

_This comment was generated by OpenHands-GPT-5.6, an AI agent, on behalf of Engel Nyst._

enyst (approved):
Thank you for this! We definitely shouldn’t load entire conversations just for a few details like knowing how many they are or their titles. It looks like this gives the largest gain in performance.

My agent verified a few potential nits, and it seems that 2 of them are happening (which come down to the same root cause I think). I think maybe they are opportunities for simpler code: https://github.com/OpenHands/software-agent-sdk/pull/4100#issuecomment-4961497968

enyst:
@openhands pick the latest nits from here https://github.com/OpenHands/software-agent-sdk/pull/4100#issuecomment-4961497968 and open a new PR from main (we merged this one) as a follow up to address the simplification suggested there

enyst:
Opened follow-up PR #4122 from current `main`: https://github.com/OpenHands/software-agent-sdk/pull/4122

It removes read-time catalog reconciliation/repair, makes test fixtures establish the catalog/runtime invariant explicitly, and deletes unloaded conversations without hydrating event history while respecting leases. Focused conversation/lease tests pass (98), and adjacent router/event/plugin/restore regressions pass (262).

_This comment was created by an AI agent (OpenHands) on behalf of the requester._

PR #4108, opened 2026-07-13
fix(task): propagate interruptions to active subagents

### LINKED ISSUE #4107: [Bug]: TaskToolSet does not propagate parent interruption to active subagents

### Bug Description

When `LocalConversation.arun()` is interrupted while the `task` tool is executing, the parent conversation transitions to `PAUSED`, but the active subagent started by `TaskToolSet` continues running.

The SDK already propagates interruption from `LocalConversation.interrupt()` to the active tool by calling `ToolExecutor.interrupt()`. However, `TaskExecutor` does not override `interrupt()`, so it inherits the default no-op implementation. `TaskManager` also has no method for interrupting the child conversations stored in `_tasks`.

As a result, the parent can be considered interrupted while a subagent continues making LLM requests, executing tools, consuming tokens, and potentially modifying the shared workspace.

### Expected Behavior

Interrupting a parent conversation should propagate the interruption to every active subagent started by its `TaskToolSet`.

At minimum, active child conversations should be paused before they can start another iteration. Ideally, an in-flight child LLM request or tool execution should also be interrupted when supported.

### Actual Behavior

The parent conversation transitions to `PAUSED` and emits an interruption event, but the `TaskExecutor` worker thread continues waiting for the child conversation.

The child runs until it finishes, reaches its iteration limit, or fails independently.

### Steps to Reproduce

1. Create a parent agent with `TaskToolSet`.
2. Register a subagent that performs a sufficiently long task.
3. Start the parent conversation using `await conversation.arun()`.
4. Wait until the parent invokes the `task` tool and the subagent starts running.
5. Call `conversation.interrupt()`.
6. Observe that the parent becomes `PAUSED` while the child conversation continues producing LLM or tool activity.

### Installation Method

Editable repository installation using uv

_No response_

### SDK Version

Latest upstream main. The affected implementation is also present in the local 1.34.0 development checkout.

### Version Confirmation

### Python Version

3.13.14

Not model-specific

### Operating System

Linux

### Logs and Error Messages

No exception is emitted. The parent conversation reports PAUSED, while the child conversation continues running.

### Minimal Code Sample

```
from uuid import uuid4

from openhands.tools.task.impl import TaskExecutor
from openhands.tools.task.manager import Task, TaskManager, TaskStatus

manager = TaskManager()
child = MagicMock()

manager._tasks["task_1"] = Task.model_construct(
    id="task_1",
    status=TaskStatus.RUNNING,
    conversation_id=uuid4(),
    conversation=child,
    result=None,
    error=None,
)

TaskExecutor(manager).interrupt()

# Currently 0 because TaskExecutor inherits the default no-op interrupt().
assert child.interrupt.call_count == 1
```

The generic cancellation chain already exists:

`LocalConversation.interrupt()` -> `ParallelToolExecutor` -> `ToolExecutor.interrupt()`

The chain stops at `TaskExecutor`.

A focused fix could:

1. Implement `TaskExecutor.interrupt()` and delegate to `TaskManager.interrupt()`.
2. Make `TaskManager.interrupt()` take a thread-safe snapshot of active child conversations and interrupt them outside `_tasks_lock`.
3. Make `TaskManager.close()` interrupt active children before clearing `_tasks`.
4. Add tests for repeated interruption and interruption racing with normal child completion.

Because child conversations currently use blocking `conversation.run()`, the first change may only stop them after the current synchronous step. Full cancellation of an in-flight LLM or tool call may require running child conversations asynchronously and can be handled separately.

This is distinct from #2056: the desired behavior is cancellation propagation, not moving a blocking task to the background.

bozhnyukAlex:
HUMAN:
I reviewed the implementation. Need this thing for production use.

---

AGENT:

## Why

Interrupting a parent `LocalConversation` while its `task` tool was running only paused the parent. `TaskExecutor` inherited the no-op `ToolExecutor.interrupt()`, so the active subagent continued consuming iterations and model calls.

## Summary

- Delegate `TaskExecutor.interrupt()` to `TaskManager`.
- Interrupt every active subagent conversation without holding the task registry lock during callbacks.
- Track interrupt generations so cancellation is not lost while a subagent is starting.
- Run subagent conversations through `arun()` so an in-flight model request can be cancelled.
- Interrupt active subagents during manager shutdown.
- Add manager, executor, and end-to-end parent/subagent interruption coverage.

## Issue Number

Closes #4107.

## How to Test

- `uv run pre-commit run --all-files --show-diff-on-failure`
- `uv run pytest tests/tools/task -q`
- `uv run pytest tests/agent_server/test_openapi_discriminator.py tests/sdk/context/test_agent_context_serialization.py tests/sdk/conversation/local/test_conversation_core.py tests/sdk/conversation/local/test_state_serialization.py tests/tools/file_editor/test_memory_usage.py -q`
- `uv run pytest tests/tools/terminal -q` (311 passed; two environment-contaminated cases passed when rerun in isolation after cleaning a stale tmux server)

End-to-end evidence: `test_parent_interrupt_stops_active_subagent` starts a real parent conversation with `TaskToolSet`, waits for a blocking subagent LLM call, interrupts the parent, and verifies both cancellation of the child request and the parent's `PAUSED` state.

## Video/Screenshots

Not applicable; this changes runtime cancellation behavior and is covered by an end-to-end test.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Notes

The synchronous `TaskManager.start_task()` contract is unchanged. Only the internal subagent run loop is asynchronous so SDK interruption can cancel in-flight work.

Cancellation propagation applies when the parent conversation is driven through `LocalConversation.arun()`. Parents using synchronous `run()` retain cooperative pause semantics and cannot interrupt an in-flight subagent call.

Task subagents now run through `arun()` and therefore use the separate `LLM.acompletion()` path instead of `LLM.completion()`. Custom LLM subclasses used by task subagents must support that asynchronous method.

### DIFF
--- a/openhands-tools/openhands/tools/task/impl.py
+++ b/openhands-tools/openhands/tools/task/impl.py
@@ -67,3 +67,7 @@ class TaskExecutor(ToolExecutor):
 
     def close(self) -> None:
         self._manager.close()
+
+    def interrupt(self) -> None:
+        """Interrupt all active sub-agent tasks."""
+        self._manager.interrupt()
--- a/openhands-tools/openhands/tools/task/manager.py
+++ b/openhands-tools/openhands/tools/task/manager.py
@@ -45,6 +45,9 @@ ConfirmationHandler = Callable[[str, list["ActionEvent"]], bool]
 logger = get_logger(__name__)
 
 _SUBAGENTS_DIR: Final[str] = "subagents"
+_INTERRUPTED_BEFORE_START_ERROR: Final[str] = (
+    "Sub-agent task was interrupted before it started."
+)
 
 
 class TaskStatus(StrEnum):
@@ -103,6 +106,7 @@ class TaskManager:
 
         self._tasks: dict[str, Task] = {}
         self._tasks_lock = threading.Lock()
+        self._interrupt_generation = 0
 
         # Set once in _ensure_parent: uses the parent's subagents dir
         # when the parent persists, otherwise a temporary directory.
@@ -180,6 +184,7 @@ class TaskManager:
         Returns:
             TaskState with the final result.
         """
+        interrupt_generation = self._get_interrupt_generation()
         if conversation:
             self._ensure_parent(conversation)
 
@@ -197,7 +202,16 @@ class TaskManager:
         return self._run_task(
             task=task,
             prompt=prompt,
+            interrupt_generation=interrupt_generation,
         )
+
+    def _get_interrupt_generation(self) -> int:
+        with self._tasks_lock:
+            return self._interrupt_generation
+
+    def _was_interrupted(self, generation: int) -> bool:
+        with self._tasks_lock:
+            return generation != self._interrupt_generation
 
     def _resume_task(self, resume: str, subagent_type: str) -> Task:
         """Resume a sub-agent task."""
@@ -344,8 +358,15 @@ class TaskManager:
         )
         return sub_agent
 
-    def _run_task(self, task: Task, prompt: str) -> Task:
+    def _run_task(
+        self,
+        task: Task,
+        prompt: str,
+        interrupt_generation: int | None = None,
+    ) -> Task:
         """Run a task synchronously."""
+        if interrupt_generation is None:
+            interrupt_generation = self._get_interrupt_generation()
         if task.conversation is None:
             raise RuntimeError(f"Task '{task.id}' has no conversation to run.")
         # Get parent name for sender info
@@ -355,6 +376,11 @@ class TaskManager:
             parent_name = getattr(parent._visualizer, "_name", None)
 
         try:
+            if self._was_interrupted(interrupt_generation):
+                task.set_error(_INTERRUPTED_BEFORE_START_ERROR)
+                logger.info("Task '%s' was interrupted before it started.", task.id)
+                return task
+
             task.conversation.send_message(prompt, sender=parent_name)
             self._run_until_finished(task.id, task.conversation)
             status = task.conversation.state.execution_status
@@ -453,6 +479,7 @@ class TaskManager:
     def interrupt(self) -> None:
         """Interrupt all active sub-agent conversations."""
         with self._tasks_lock:
+            self._interrupt_generation += 1
             active_tasks = tuple(
                 (task.id, task.conversation)
                 for task in self._tasks.values()
--- a/openhands-tools/openhands/tools/task/manager.py
+++ b/openhands-tools/openhands/tools/task/manager.py
@@ -10,6 +10,7 @@ a temporary directory, ensuring the state can be restored
 if the task is resumed for further work later.
 """
 
+import asyncio
 import shutil
 import tempfile
 import threading
@@ -400,7 +401,13 @@ class TaskManager:
         self, task_id: str, conversation: LocalConversation
     ) -> None:
         """Run a sub-agent conversation to completion, handling confirmations."""
-        conversation.run()
+        asyncio.run(self._arun_until_finished(task_id, conversation))
+
+    async def _arun_until_finished(
+        self, task_id: str, conversation: LocalConversation
+    ) -> None:
+        """Run a sub-agent asynchronously so in-flight work can be interrupted."""
+        await conversation.arun()
         while (
             conversation.state.execution_status
             == ConversationExecutionStatus.WAITING_FOR_CONFIRMATION
@@ -412,10 +419,10 @@ class TaskManager:
             if self._confirmation_handler is None or self._confirmation_handler(
                 task_id, pending
             ):
-                conversation.run()
+                await conversation.arun()
             else:
                 conversation.reject_pending_actions("User rejected the actions")
-                conversation.run()
+                await conversation.arun()
 
     def _set_confirmation_policy(
         self,
@@ -443,8 +450,24 @@ class TaskManager:
                 task.conversation.conversation_stats.get_combined_metrics()
             )
 
+    def interrupt(self) -> None:
+        """Interrupt all active sub-agent conversations."""
+        with self._tasks_lock:
+            active_tasks = tuple(
+                (task.id, task.conversation)
+                for task in self._tasks.values()
+                if task.status == TaskStatus.RUNNING and task.conversation is not None
+            )
+
+        for task_id, conversation in active_tasks:
+            try:
+                conversation.interrupt()
+            except Exception:
+                logger.warning("Error interrupting task '%s'", task_id, exc_info=True)
+
     def close(self) -> None:
         """Clean up temporary directory (if used) and remove all created tasks."""
+        self.interrupt()
         # Only clean up when using a temp dir (parent had no persistence).
         # When the parent persists, subagent data lives under its directory.
         parent_persists = (
--- a/tests/tools/task/test_task_manager.py
+++ b/tests/tools/task/test_task_manager.py
@@ -1,7 +1,7 @@
 import uuid
 from pathlib import Path
 from typing import Any, cast
-from unittest.mock import MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 from pydantic import SecretStr
@@ -323,6 +323,33 @@ class TestTaskManager:
         assert not manager._persistence_dir.exists()
         assert len(manager._tasks) == 0
 
+    def test_interrupt_pauses_only_running_tasks(self, tmp_path):
+        manager, _ = _manager_with_parent(tmp_path)
+        register_builtins_agents()
+        running = manager._create_task(
+            subagent_type="general-purpose",
+            description="running task",
+        )
+        completed = manager._create_task(
+            subagent_type="general-purpose",
+            description="completed task",
+        )
+        completed.set_result("done")
+        assert running.conversation is not None
+        assert completed.conversation is not None
+
+        manager.interrupt()
+        manager.interrupt()
+
+        assert (
+            running.conversation.state.execution_status
+            == ConversationExecutionStatus.PAUSED
+        )
+        assert (
+            completed.conversation.state.execution_status
+            == ConversationExecutionStatus.IDLE
+        )
+
     def test_returns_local_conversation(self, tmp_path):
         manager, _ = _manager_with_parent(tmp_path)
         register_builtins_agents()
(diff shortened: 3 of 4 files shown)


openhands-tools/openhands/tools/task/manager.py
```diff
@@ -443,8 +450,24 @@ def _update_parent_metrics(self, parent: LocalConversation, task: Task) -> None:
                 task.conversation.conversation_stats.get_combined_metrics()
             )
 
+    def interrupt(self) -> None:
+        """Interrupt all active sub-agent conversations."""
+        with self._tasks_lock:
+            active_tasks = tuple(
+                (task.id, task.conversation)
+                for task in self._tasks.values()
+                if task.status == TaskStatus.RUNNING and task.conversation is not None
```
VascoSch92, line 486: This does the right thing for tasks that are already running, but it means an interrupt landing while a subagent is still starting up just disappears. Two variants:

- arrives before `_create_task` registers the task, so nothing matches and `interrupt()` is a no-op
- arrives after registration but before `arun()` starts, so `_arun_task` is still `None`, `LocalConversation.interrupt()` falls back to `pause()`, and then `arun()` flips `PAUSED` straight back to `RUNNING` and the pause is gone

I poked at the second one and the child does go on to make a real LLM call. Realistically the window is only ~20ms (`_ensure_agent_ready` runs after `_arun_task` is set, so it doesn't widen it), so I don't think this blocks anything. If you do want to close it, a latch on the manager would be enough: set it here, check it in `_run_task` before `_run_until_finished`, clear it when the parent starts a new run.

openhands-tools/openhands/tools/task/manager.py
```diff
@@ -400,7 +401,13 @@ def _run_until_finished(
...
     ) -> None:
         """Run a sub-agent conversation to completion, handling confirmations."""
-        conversation.run()
+        asyncio.run(self._arun_until_finished(task_id, conversation))
+
+    async def _arun_until_finished(
+        self, task_id: str, conversation: LocalConversation
+    ) -> None:
+        """Run a sub-agent asynchronously so in-flight work can be interrupted."""
+        await conversation.arun()
```
VascoSch92, line 436: Heads up that `acompletion` isn't a thin wrapper over `completion`, it's a separate path built on `litellm.acompletion`. So subagents are now going through code they never hit before.

Your own `InterruptibleSubagentLLM` needing to override both methods is a good illustration of the split: anyone with a custom `LLM` subclass that only overrides `completion()` will quietly get different behavior under the task tool now. I'm not saying it's wrong, just that it's a wider surface change than "propagate interrupts" suggests, so it deserves a sentence in the description.

openhands-tools/openhands/tools/task/impl.py
```diff
@@ -67,3 +67,7 @@ def __call__(
 
     def close(self) -> None:
         self._manager.close()
+
+    def interrupt(self) -> None:
+        """Interrupt all active sub-agent tasks."""
+        self._manager.interrupt()
```
VascoSch92, line 73: Might be worth noting in the description that this only helps async parents. `executor.interrupt()` only ever gets called from `_arun_safe`'s `CancelledError` handler; the sync `execute_batch` / `_run_safe` path never touches it. So anyone driving the parent with `conversation.run()` still can't interrupt subagents. Fine as a limitation, I just read the description as promising a bit more than that on first pass.

tests/tools/task/test_task_manager.py
```diff
@@ -323,6 +323,33 @@ def test_close(self, tmp_path):
...
+        )
+        completed = manager._create_task(
+            subagent_type="general-purpose",
+            description="completed task",
+        )
+        completed.set_result("done")
+        assert running.conversation is not None
+        assert completed.conversation is not None
+
+        manager.interrupt()
```
VascoSch92, line 343: Small thing: neither conversation has actually run at this point, so `_arun_task` is `None` on both and this ends up exercising the `pause()` fallback rather than the task-cancellation path the fix actually depends on.

The e2e test covers the real path so coverage is fine overall. If you wanted this one sharper, you could push a task through `_run_until_finished` on a thread and interrupt while `arun()` is in flight.

VascoSch92 (commented):
Went through this carefully and the bug is definitely real. I ran the repro from #4107 on both branches and got `call_count` 0 on main vs 1 here, and your new `test_parent_interrupt_stops_active_subagent` fails on main and passes on this branch. Task suite is green for me too (68/68).

One thing I want to highlight because it's easy to miss on a skim: swapping `run()` for `arun()` isn't cosmetic, it's the part that actually makes this work. `LocalConversation.interrupt()` only cancels `_arun_task`, and that's only ever set by `arun()`. If the child had stayed on sync `run()` you'd have fallen through to `pause()`, which doesn't take effect until between iterations, so it would never cancel an in-flight LLM call. Nice catch.

A few comments inline, none of them blocking.

Also worth recording two things I went looking for and didn't find: nesting `asyncio.run()` inside a running loop isn't a problem here, since tool calls always get dispatched to a worker thread via `run_in_executor`; and I wondered about `_tasks_lock` contention while `_create_task` holds it, but it only measures 1-3ms because tools initialize lazily outside the lock.

Happy to approve once the description picks up the sync-parent and `acompletion` points.

VascoSch92 (commented):
A couple of comments :-)

tests/tools/task/test_task_manager.py
bozhnyukAlex: Agreed. The end-to-end test covers the actual in-flight cancellation path, while this unit test only verifies that the manager targets running tasks. I renamed it to `test_interrupt_only_targets_running_tasks` in 7ad3e8795 to make that scope explicit.

openhands-tools/openhands/tools/task/manager.py
bozhnyukAlex: Good point. I updated the PR description to state explicitly that task subagents now use the separate `LLM.acompletion()` path instead of `LLM.completion()`, and that custom LLM subclasses used by task subagents must support the asynchronous method.

openhands-tools/openhands/tools/task/impl.py
bozhnyukAlex: Agreed. I clarified the PR description that immediate cancellation propagation applies when the parent is driven through `LocalConversation.arun()`. A parent using synchronous `run()` retains cooperative pause semantics and cannot interrupt an in-flight subagent call.

openhands-tools/openhands/tools/task/manager.py
bozhnyukAlex: Implemented in 7e9e5074b using an interrupt generation counter rather than a boolean latch, so no explicit reset is needed. `start_task()` snapshots the generation before creating or resuming the task, `interrupt()` increments it under the task lock, and `_run_task()` rejects stale generations before sending the prompt or starting `arun()`. I added deterministic coverage for interrupts before registration, after registration, and for a new task starting successfully after an interrupt.

VascoSch92 (approved):
One thing I want to confirm before approving: routing subagents through `arun()` means they now always use `LLM.acompletion()`, even when the parent runs synchronously. Since `acompletion()` is an independent litellm path and does not delegate to `completion()`, any custom LLM subclass that overrides only `completion()` will have that override silently bypassed for subagents. 

It's documented in the Notes, but should we also surface it in the changelog / release notes so downstream users with custom sync-only LLMs aren't caught out? @enyst 

Otherwise LGTM

bozhnyukAlex:
Still active and ready to merge. All review comments have been addressed, the review threads are resolved, and CI is green.  
@enyst, could you please confirm whether a changelog entry is needed for the LLM.acompletion() behavior change? If not, this should be ready to merge. Thanks!

enyst (commented):

🔴 **Needs improvement** - The direction is right, but the cancellation handoff still loses an interrupt in one real race window and the async conversion breaks an existing customization contract.

**[CRITICAL ISSUES]**
- [`openhands-tools/openhands/tools/task/manager.py`, Line 379] **Lost interrupt between the generation check and `arun()` registration**: `_was_interrupted()` is checked before `send_message()` and `_run_until_finished()`. If `TaskManager.interrupt()` lands after that check but before `LocalConversation.arun()` sets `_arun_task` and `_cancel_token`, `conversation.interrupt()` falls back to `pause()`. `arun()` then treats `PAUSED` as resumable and sets the child back to `RUNNING`, so the interrupt this PR is meant to make durable is lost. Close the handoff by registering the async run before the final generation check (or otherwise carrying a durable cancellation token into startup), and add a deterministic barrier test for this exact interval; the current tests cover interruption before `_run_task()` and after the child LLM call starts, but not between them.
- [`openhands-tools/openhands/tools/task/manager.py`, Line 430] **Breaking change for sync-only custom LLM subclasses**: every task subagent now goes through `arun()`/`LLM.acompletion()`. The inherited `acompletion()` path calls the async LiteLLM transport directly; it does not delegate to a subclass's overridden `completion()`. Existing custom task agents that intentionally implement only the long-standing synchronous override will now bypass their implementation and may fail or call an unintended provider. The PR description acknowledges the new requirement, but documentation does not make a userspace break acceptable. Preserve the old override contract through a compatibility fallback/adapter or a deprecation path, and add a regression test with an `LLM` subclass that overrides only `completion()`.

**[TESTING GAPS]**
- [Overall PR] **Eval-sensitive task behavior**: this changes tool execution and subagent orchestration, and I found no completed eval-monitor run plus human confirmation. Per this repository's review guide, a human maintainer should decide after lightweight eval coverage once the correctness issues above are fixed.

**[RISK ASSESSMENT]**
This touches core subagent orchestration and changes which public LLM method downstream customizations execute. The remaining startup race can leak model work after an interrupt, while the compatibility break can fail existing custom agents. **Recommendation**: Do not auto-merge. Request review from a human SDK architect and validate task-tool behavior with lightweight evals after fixing the startup handoff and sync-only LLM compatibility.

**VERDICT:**
❌ **Needs rework**: The current implementation does not make startup cancellation fully durable and breaks existing custom LLM subclasses.

**KEY INSIGHT:**
Cancellation is only reliable when ownership transfers atomically from the manager's generation token to the child's registered async task.

---

>
>
>

Issue #3629, opened 2026-06-10
ACPAgent init hangs silently when authenticate() blocks (no startup timeout)


simonrosenberg:
## Summary

When an ACP provider's `authenticate()` call hangs (e.g. codex-acp with an expired `id_token`), the conversation freezes indefinitely with no error surfaced to the client.

## Root cause

`_start_acp_server` runs the async `_init()` coroutine via `self._executor.run_async(_init)` (line ~1785 of `acp_agent.py`). Inside `_init()`, the call to `await conn.authenticate(method_id=method_id, **auth_kwargs)` has no timeout — if the underlying ACP server hangs on its auth network call, `run_async` blocks forever.

`init_state` wraps `_start_acp_server` in a `try/except` that emits a `ConversationErrorEvent` and flips the conversation to ERROR status — but that only fires on exceptions. A silent hang never raises, so the except block is never reached, no event is sent, and the UI renders nothing.

The same gap applies to `conn.initialize()` and `conn.new_session()` / `conn.load_session()` inside `_init()`.

## Observed symptom

- User has an expired `id_token` in their Codex `auth.json` (`last_refresh` was 5 days ago)
- Conversation is created, first message is received, file-secret is materialised, ACP server starts
- Log shows `Authenticating with ACP method: chatgpt` — then silence, indefinitely
- UI shows a spinner / no response, no error banner, forever

## Fix

Wrap the entire `_init()` coroutine in `asyncio.wait_for` before passing it to `run_async`, e.g.:

```python
self._executor.run_async(asyncio.wait_for(_init(), timeout=acp_startup_timeout))
```

When it fires, `TimeoutError` propagates to `init_state`'s except block, which already emits `ConversationErrorEvent` via `_classify_acp_init_error` and sets the conversation to ERROR — so the UI renders its existing error banner.

A new `acp_startup_timeout` field on `ACPAgentSettings` (mirroring the existing `acp_prompt_timeout`) would let callers tune this. Reasonable default: 60–90 s. `_classify_acp_init_error` should recognise `TimeoutError` and return a descriptive code/message ("ACP startup timed out — if using chatgpt auth, your id\_token may be expired").

rsd-darshan:
A fix has been submitted here: #4126

[...]

PR #4126, opened 2026-07-16
fix(acp): bound ACP server startup with a timeout

rsd-darshan:
HUMAN:

Picked this up from the open issue tracker, fix verified locally.

---

AGENT:

## Why

`ACPAgent._start_acp_server` runs its spawn + `initialize()`/`authenticate()`/`new_session()`/`load_session()` coroutine (`_init()`) via `self._executor.run_async(_init)` with no timeout. If any of those ACP calls hang — e.g. codex-acp blocking indefinitely on `authenticate()` when the local `id_token` is expired — the coroutine never returns and never raises, so `init_state()` freezes forever. Because a silent hang never raises, the existing `try/except` in `init_state()` that emits a typed `ConversationErrorEvent` is never reached, and the client sees no error at all — just a spinner forever.

## Summary

- Added `acp_startup_timeout: float` (default 90s) to `ACPAgent`, mirroring the existing `acp_prompt_timeout` idle-timeout field.
- Passed `timeout=self.acp_startup_timeout` to the existing `run_async(...)` call (it already supports a `timeout` kwarg via `anyio.fail_after`, already used elsewhere in this file for the prompt-turn idle timeout).
- The resulting bare `TimeoutError` (`anyio.fail_after` raises one with no message) is caught and re-raised with a descriptive message, mirroring `_idle_timeout_message()`'s pattern for the turn-loop path.
- Added an `ACPStartupTimeout` branch to `_classify_acp_init_error()` so `init_state()`'s existing cold-start error handler emits a typed `ConversationErrorEvent` instead of the generic `ACPInitError`.
- Wired the new field through every place that already mirrors `acp_prompt_timeout`: `ACPAgentSettings` (+ its `create_agent()` pass-through), `ACPAgentProfile`, and the two settings↔profile conversion helpers (`profiles/seed.py`, `profiles/resolver.py`).

## Issue Number

#3629

## How to Test

Ran the full local test suite after the change:

```
uv run pytest tests/sdk/agent/test_acp_agent.py tests/sdk/test_settings.py tests/sdk/profiles/ \
  tests/agent_server/test_settings_router.py tests/agent_server/test_conversation_router.py -q
# 832 passed

uv run pyright openhands-sdk/openhands/sdk/agent/acp_agent.py openhands-sdk/openhands/sdk/settings/model.py \
  openhands-sdk/openhands/sdk/profiles/agent_profile.py openhands-sdk/openhands/sdk/profiles/seed.py \
  openhands-sdk/openhands/sdk/profiles/resolver.py
# 0 errors

make lint
# All checks passed!
```

Added two new regression tests that reproduce the reported hang directly: `test_start_acp_server_bounds_a_hung_handshake_call` makes `conn.initialize()` hang for 10s with `acp_startup_timeout=0.05`, and asserts `_start_acp_server` raises a descriptive `TimeoutError` well before the 10s would elapse (test runs in well under a second). `test_init_state_surfaces_startup_timeout` drives the same hang through the full `init_state()` path and asserts the client-facing `ConversationErrorEvent` carries code `ACPStartupTimeout` and the descriptive detail message. Also added `test_timeout_error_is_startup_timeout` for the classifier unit.

Pre-commit hooks (ruff format, ruff lint, pycodestyle, pyright, import-dependency rules, tool-subclass registration) all passed on commit.

## Video/Screenshots

Not applicable — backend Python logic fix with no UI surface; verified via the tests above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

New field defaults to 90s, within the 60-90s range suggested in the issue. No breaking changes — existing callers get the new timeout automatically with a sane default; no config migration needed.

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/acp_agent.py
+++ b/openhands-sdk/openhands/sdk/agent/acp_agent.py
@@ -1021,15 +1021,20 @@ def _classify_acp_init_error(exc: BaseException) -> str:
     - ``ACPAuthRequired``: a credential failure — the explicit ``-32000`` auth code,
       or a ``-32603`` whose message/data reveals an upstream 401/403 (see
       :func:`_acp_error_indicates_auth`).  The most actionable cloud failure.
+    - ``ACPStartupTimeout``: startup did not complete within
+      ``acp_startup_timeout`` — e.g. the server hung on ``authenticate()``
+      without ever returning an error (see :meth:`ACPAgent._start_acp_server`).
     - ``ACPSpawnError``: the subprocess could not be launched — the CLI binary is
       missing or not executable (``FileNotFoundError`` / ``PermissionError`` from
       ``create_subprocess_exec``).
     - ``ACPInitError``: anything else during the protocol handshake or session
-      creation (timeouts, transport drops, unexpected protocol errors, cwd
-      mismatch surfaced by the server).
+      creation (transport drops, unexpected protocol errors, cwd mismatch
+      surfaced by the server).
     """
     if _acp_error_indicates_auth(exc):
         return "ACPAuthRequired"
+    if isinstance(exc, TimeoutError):
+        return "ACPStartupTimeout"
     if isinstance(exc, (FileNotFoundError, PermissionError)):
         return "ACPSpawnError"
     return "ACPInitError"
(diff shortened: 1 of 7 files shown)


neubig (approved):
Thank you!

PR #4132, opened 2026-07-17
fix(llm): skip unauthenticated managed proxy model-info lookup

### LINKED ISSUE #4098: Avoid unauthenticated model-info lookup for managed OpenHands models during LLM initialization
## Summary

Constructing or validating an LLM configured with an `openhands/*` model can synchronously call the managed LiteLLM proxy's `GET /v1/model/info` endpoint even when no API key is available.

The SDK correctly omits the Authorization header when the key is empty, but it still sends the request. The managed proxy returns 401, the SDK silently falls back to static LiteLLM metadata, and production accumulates roughly 14K–30K unauthenticated model-info responses per day.

For the known managed OpenHands proxy, a request without credentials cannot return key-scoped proxy metadata and should be skipped.

## Current behavior

The initialization path is:

1. `LLM._post_init()`
2. `LLM._init_model_info_and_caps()`
3. `get_litellm_model_info()`
4. `_get_model_info_from_litellm_proxy()`
5. `httpx.get(f"{base_url}/v1/model/info", headers=headers)`

For a missing key, `headers` remains empty but step 5 still runs.

This commonly occurs during startup, profile loading, validation, or deserialization—not only before a completion. The process-local LRU cache reduces identical calls, but pod restarts, hourly cache keys, concurrency, and key/model cardinality still produce substantial traffic.

PR #1022 previously fixed a blank `Authorization: Bearer ` header by omitting it. That changed the failure from 422 to 401 but retained the request. PR #1148 added caching, but the lookup remains eager during LLM initialization.

## Expected behavior

When all of the following are true:

- the model resolves to the managed OpenHands LiteLLM proxy;
- the proxy URL is the default or recognized managed OpenHands proxy, rather than an arbitrary custom proxy; and
- no API key is available;

the SDK should skip `/v1/model/info` discovery and use its existing static LiteLLM metadata fallback without making an HTTP request.

Unauthenticated discovery must remain supported for generic `litellm_proxy/*` configurations and custom proxy URLs, because private proxies may intentionally expose model information without authentication.

Authenticated managed-proxy discovery must also remain unchanged because it supplies key-scoped aliases and capability/token-limit overrides.

## Suggested implementation

- Preserve enough pre-translation context to distinguish the managed OpenHands provider from a generic LiteLLM proxy.
- Before the HTTP request, return `None` when the target is the managed OpenHands proxy and the resolved API key is empty.
- Preserve the current static fallback.
- Consider explicitly handling non-2xx responses instead of treating arbitrary error JSON as an empty model list, without propagating discovery failures from initialization.
- Coordinate with draft PR #3912, which adds model-info timeouts but does not address missing credentials or 401 volume.

## Acceptance criteria

- [ ] Constructing `LLM(model="openhands/<model>")` without an API key and without a custom base URL performs no model-info HTTP request.
- [ ] A recognized managed OpenHands proxy URL without an API key also skips the request.
- [ ] An authenticated managed OpenHands configuration still calls `/v1/model/info` with its Bearer token and applies returned overrides.
- [ ] A custom or authless `litellm_proxy/*` configuration without an API key can still query model info.
- [ ] A custom `openhands/*` proxy URL is not incorrectly classified as the managed production proxy.
- [ ] Regression tests cover each case above.

## Related work

- #3495 documents silent model-info lookup failure affecting capability detection, but its authenticated eval-proxy case has different behavior and acceptance criteria.
- #1022 removed the blank Bearer header while retaining missing-key discovery.
- #1148 introduced the current model-info cache.
- #3912 adds a timeout to eager discovery.
- BerriAI/litellm#27923 and BerriAI/litellm#29483 address the separate problem of budget enforcement blocking authenticated discovery. That fix ships in LiteLLM v1.89.0+ and does not change missing-key 401 behavior.

josephsenior:
HUMAN:

I reviewed the implementation and ran the focused model-info proxy tests,
Ruff lint, and Ruff formatting checks locally. All passed.

AGENT:

---

## Why

Constructing an `LLM` for an `openhands/*` model translates it to the managed
LiteLLM proxy before model-info discovery. When no API key is available, the SDK
currently still calls `/v1/model/info` without authentication. The managed proxy
cannot return key-scoped metadata in that case, responds with 401, and the SDK
falls back to static LiteLLM metadata anyway.

This creates avoidable unauthenticated traffic during startup, profile loading,
validation, and deserialization.

## Summary

- Recognize the managed OpenHands proxy using the existing normalized URL helper.
- Skip proxy model-info discovery only when that managed proxy has no API key.
- Continue using the existing static LiteLLM metadata fallback.
- Preserve authenticated managed-proxy discovery and Bearer-token behavior.
- Preserve unauthenticated discovery for generic or custom LiteLLM proxies.
- Preserve custom `openhands/*` proxy URLs instead of treating them as managed.
- Add focused regression coverage for each behavior.

## Issue Number

Closes #4098

## How to Test

```bash
uv run pytest tests/sdk/llm/test_model_info_proxy_lookup.py
uv run ruff check   openhands-sdk/openhands/sdk/llm/utils/model_info.py   tests/sdk/llm/test_model_info_proxy_lookup.py
uv run ruff format --check   openhands-sdk/openhands/sdk/llm/utils/model_info.py   tests/sdk/llm/test_model_info_proxy_lookup.py
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/utils/model_info.py
+++ b/openhands-sdk/openhands/sdk/llm/utils/model_info.py
@@ -36,7 +36,10 @@ def _get_model_info_from_litellm_proxy(
         if api_key:
             headers["Authorization"] = f"Bearer {api_key}"
 
-        response = httpx.get(f"{base_url}/v1/model/info", headers=headers)
+        normalized_base_url = base_url.rstrip("/")
+        if normalized_base_url.endswith("/v1"):
+            normalized_base_url = normalized_base_url[:-3]
+        response = httpx.get(f"{normalized_base_url}/v1/model/info", headers=headers)
         data = response.json().get("data", [])
         # Match against either the public alias (`model_name`) or the
         # underlying provider/model_name form (`litellm_params.model`). The proxy itself
--- a/openhands-sdk/openhands/sdk/llm/utils/model_info.py
+++ b/openhands-sdk/openhands/sdk/llm/utils/model_info.py
@@ -7,10 +7,19 @@ from litellm.types.utils import ModelInfo
 from litellm.utils import get_model_info
 from pydantic import SecretStr
 
-from openhands.sdk.llm.utils.openhands_provider import litellm_call_kwargs
+from openhands.sdk.llm.utils.openhands_provider import (
+    is_openhands_proxy_base_url,
+    litellm_call_kwargs,
+)
 
 
 logger = getLogger(__name__)
+
+
+def _get_api_key_value(secret_api_key: SecretStr | str | None) -> str | None:
+    if isinstance(secret_api_key, SecretStr):
+        return secret_api_key.get_secret_value()
+    return secret_api_key
 
 
 @lru_cache
@@ -23,10 +32,9 @@ def _get_model_info_from_litellm_proxy(
     logger.debug(f"Get model_info_from_litellm_proxy:{cache_key}")
     try:
         headers = {}
-        if isinstance(secret_api_key, SecretStr):
-            secret_api_key = secret_api_key.get_secret_value()
-        if secret_api_key:
-            headers["Authorization"] = f"Bearer {secret_api_key}"
+        api_key = _get_api_key_value(secret_api_key)
+        if api_key:
+            headers["Authorization"] = f"Bearer {api_key}"
 
         response = httpx.get(f"{base_url}/v1/model/info", headers=headers)
         data = response.json().get("data", [])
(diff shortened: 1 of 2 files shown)


enyst (commented):

🟢 **Good taste** — this is a narrow fix at the correct boundary: managed-proxy identity is reused from the canonical URL helper, missing keys skip only the request that cannot succeed, and authenticated/custom proxy behavior remains intact. The URL normalization also avoids duplicating `/v1`, and the focused tests cover the meaningful policy branches.

**[RISK ASSESSMENT]**
The change affects startup metadata discovery only, preserves the static fallback, defaults custom proxies to their previous behavior, and has targeted regression coverage for authenticated and unauthenticated paths.

**VERDICT:**
✅ **Worth merging**: The behavior change is minimal, backward-compatible, and correctly scoped.

**KEY INSIGHT:**
Managed service policy belongs behind canonical endpoint identity, not behind model-name heuristics.

PR #4135, opened 2026-07-17
Add injectable remote WebSocket client factory

p1c2u:
HUMAN:

I reviewed the change and its test evidence and consider it ready for review.

---
AGENT:

## Why

`RemoteConversation` currently constructs `WebSocketCallbackClient` directly. This prevents callers from supplying a transport-specific callback client without replacing module globals.

An injectable factory is needed for transports such as WebSocket-over-Unix-domain-socket while preserving the SDK’s existing readiness, reconnect, reconciliation, cleanup, and lifecycle behavior. OpenHands 1.36 also passes an `on_reconnect` callback, so the factory contract must include and forward it.

The default behavior remains unchanged when no factory is provided.

## Summary

- Add a typed `WebSocketClientFactory` protocol accepting `host`, `conversation_id`, `callback`, `api_key`, and `on_reconnect`.
- Add an optional `websocket_client_factory` argument to `RemoteConversation`; default construction still uses `WebSocketCallbackClient`.
- Add focused coverage proving factory selection, keyword propagation, startup, readiness waiting, and shutdown behavior.

## Issue Number

#4708

### SDK tests

From the repository root:

```bash
env OPENHANDS_SUPPRESS_BANNER=1 \
  .venv/bin/python -m pytest \
  tests/sdk/conversation/remote/test_websocket_client.py \
  -q
```

Result:

```text
55 passed in 2.97s
```

Repository checks:

```bash
pre-commit run --files \
  openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py \
  tests/sdk/conversation/remote/test_remote_conversation.py

git diff --check
```

Result: all formatter, lint, type, import, registration, and whitespace checks passed.

### End-to-end transport verification

The patched SDK was also exercised from an external OpenLoop integration harness against the exact OpenHands 1.36 agent-server Docker image and a real HAProxy relay.

The controller had `--network none` and communicated with the relay only over a Unix-domain socket. The test:

1. constructed `RemoteConversation` using the injected factory;
2. established REST and WebSocket traffic through the UDS;
3. authenticated every WebSocket connection using the first message;
4. removed and replaced only the HAProxy relay;
5. observed the callback client reconnecting without replacing the agent;
6. verified `on_reconnect` reconciliation completed before the replacement connection’s state event was delivered;
7. verified REST remained usable after reconnection; and
8. validated checkpoint restrictions, archive streaming, no published ports, credential redaction, and resource cleanup.

Command run from the OpenLoop repository:

```bash
env \
  OPENHANDS_RELAY_LIVE=1 \
  OPENHANDS_RELAY_SDK_SOURCE=/private/tmp/openloop-software-agent-sdk-1.36.0/openhands-sdk \
  OPENHANDS_SUPPRESS_BANNER=1 \
  PYTHONPATH=/private/tmp/openloop-software-agent-sdk-1.36.0/openhands-sdk \
  .venv/bin/python -m pytest \
  tests/integration/test_openhands_relay_live.py \
  -q -s --tb=short
```

Result:

```text
3 passed in 84.94s
```

Post-test cleanup reported zero labeled containers, networks, and volumes.

A wider external regression run also completed with:

```text
681 passed, 10 skipped, 9 warnings in 7.24s
```

No model-provider request was made during the end-to-end test.

## Video/Screenshots

N/A — this is a backend SDK construction seam with no visual interface. Reproduction steps and test results are included above.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- This is backward compatible: omitting `websocket_client_factory` preserves the existing `WebSocketCallbackClient` path.
- The factory is selected only at the existing callback-client construction point.
- The existing client still owns startup, readiness waiting, initial reconciliation, failure cleanup, and shutdown.
- The change avoids module-global replacement and allows multiple conversations to use independently bound transports.

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.54s · commit 264b028  
**Coverage:** complete supplied coverage; 8/8 hunks, 2/2 files.

<details>

| --- | --- | --- |
| Command injection | 3.0% | No direct hunk selected |
| Weakened authentication | 5.0% | No direct hunk selected |
| Weakened authorization | 6.0% | No direct hunk selected |
| Contract regression | 8.0% | No direct hunk selected |
| Sensitive data disclosure | 5.0% | No direct hunk selected |
| Unexpected data transfer | 4.0% | No direct hunk selected |
| Credential misuse | 7.0% | No direct hunk selected |
| Package source redirection | 4.0% | No direct hunk selected |
| Security assessment bypass | 4.0% | No direct hunk selected |
| Primary concern | None selected; confidence 93.0% | No primary concern to locate |

</details>

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -7,7 +7,7 @@ import time
 import uuid
 from collections.abc import Callable, Mapping
 from queue import Empty, Queue
-from typing import TYPE_CHECKING, SupportsIndex, overload
+from typing import TYPE_CHECKING, Protocol, SupportsIndex, overload
 from urllib.parse import quote, urlparse
 
 import httpx
@@ -277,6 +277,18 @@ class WebSocketCallbackClient:
             if remaining <= 0:
                 return
             await asyncio.sleep(min(0.1, remaining))
+
+
+class WebSocketClientFactory(Protocol):
+    def __call__(
+        self,
+        *,
+        host: str,
+        conversation_id: str,
+        callback: ConversationCallbackType,
+        api_key: str | None,
+        on_reconnect: Callable[[], object] | None,
+    ) -> WebSocketCallbackClient: ...
 
 
 class RemoteEventsList(EventsListBase):
@@ -725,6 +737,7 @@ class RemoteConversation(BaseConversation):
         observability_metadata: dict[str, TraceMetadataValue] | None = None,
         observability_tags: list[str] | None = None,
         observability_span_name: str = "conversation",
+        websocket_client_factory: WebSocketClientFactory | None = None,
         **_: object,
     ) -> None:
         """Remote conversation proxy that talks to an agent server.
@@ -762,6 +775,8 @@ class RemoteConversation(BaseConversation):
             observability_tags: Optional root span tags for observability backends.
             observability_span_name: Optional child span name for observability
                       backends. The root span remains named "conversation".
+            websocket_client_factory: Optional callback-client factory. Uses the
+                      default WebSocket callback client when omitted.
         """
         super().__init__()  # Initialize base class with span tracking
         self.agent = agent
@@ -988,7 +1003,12 @@ class RemoteConversation(BaseConversation):
         composed_callback = BaseConversation.compose_callbacks(all_callbacks)
 
         # Initialize WebSocket client for callbacks
-        self._ws_client = WebSocketCallbackClient(
+        factory = (
+            WebSocketCallbackClient
+            if websocket_client_factory is None
+            else websocket_client_factory
+        )
+        self._ws_client = factory(
             host=self.workspace.host,
             conversation_id=str(self._id),
             callback=composed_callback,
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -7,7 +7,7 @@ import time
 import uuid
 from collections.abc import Callable, Mapping
 from queue import Empty, Queue
-from typing import TYPE_CHECKING, Final, SupportsIndex, overload
+from typing import TYPE_CHECKING, Final, Protocol, SupportsIndex, overload
 from urllib.parse import urlparse
 
 import httpx
@@ -284,6 +284,18 @@ class WebSocketCallbackClient:
             if remaining <= 0:
                 return
             await asyncio.sleep(min(0.1, remaining))
+
+
+class WebSocketClientFactory(Protocol):
+    def __call__(
+        self,
+        *,
+        host: str,
+        conversation_id: str,
+        callback: ConversationCallbackType,
+        api_key: str | None,
+        on_reconnect: Callable[[], object] | None,
+    ) -> WebSocketCallbackClient: ...
 
 
 class RemoteEventsList(EventsListBase):
@@ -720,6 +732,7 @@ class RemoteConversation(BaseConversation):
         observability_metadata: dict[str, TraceMetadataValue] | None = None,
         observability_tags: list[str] | None = None,
         observability_span_name: str = "conversation",
+        websocket_client_factory: WebSocketClientFactory | None = None,
         **_: object,
     ) -> None:
         """Remote conversation proxy that talks to an agent server.
@@ -757,6 +770,8 @@ class RemoteConversation(BaseConversation):
             observability_tags: Optional root span tags for observability backends.
             observability_span_name: Optional child span name for observability
                       backends. The root span remains named "conversation".
+            websocket_client_factory: Optional callback-client factory. Uses the
+                      default WebSocket callback client when omitted.
         """
         super().__init__()  # Initialize base class with span tracking
         self.agent = agent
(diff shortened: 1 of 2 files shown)


openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -725,6 +737,7 @@ def __init__(
         observability_metadata: dict[str, TraceMetadataValue] | None = None,
         observability_tags: list[str] | None = None,
         observability_span_name: str = "conversation",
+        websocket_client_factory: WebSocketClientFactory | None = None,
```
neubig, line 740: Could we just do this and remove the none handling later?

```suggestion
        websocket_client_factory: WebSocketClientFactory = WebSocketCallbackClient,
```

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -988,7 +1003,12 @@ def run_complete_callback(event: Event) -> None:
         composed_callback = BaseConversation.compose_callbacks(all_callbacks)
 
         # Initialize WebSocket client for callbacks
-        self._ws_client = WebSocketCallbackClient(
+        factory = (
+            WebSocketCallbackClient
+            if websocket_client_factory is None
+            else websocket_client_factory
+        )
+        self._ws_client = factory(
```
neubig, line 1011:
```suggestion
        self._ws_client = websocket_client_factory(
```

neubig (commented):
Thanks! I think this can be helpful. One small request to make this simpler.

Please click the github "re-request review" button when you'd like me to take another look!

enyst (commented):

**Taste Rating: 🟡 Acceptable** — a clean seam, typed with a `Protocol` instead of `Callable[..., Any]`, defaulting to today's behavior. This is how you add an injection point without anyone noticing. Two small things and one question.

First, housekeeping so nobody blames you for it: `uv run pytest tests/sdk/conversation/remote/test_remote_conversation.py` gives **1 failed, 45 passed**, and the failure is `test_remote_conversation_raises_when_websocket_never_ready`. I checked it against `main` with your `remote_conversation.py` reverted — it fails identically there. **Pre-existing, not yours.**

- [`openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py`, L1006-1011] **Simplification**: five lines to express a default.

  ```python
  factory = (
      WebSocketCallbackClient
      if websocket_client_factory is None
      else websocket_client_factory
  )
  ```

  is `factory = websocket_client_factory or WebSocketCallbackClient`. Since the parameter is typed `WebSocketClientFactory | None`, the only falsy value it can hold is `None`, so `or` is safe here and says the same thing in one line.

- [same file, L740 `**_: object`] **Pragmatism — the signature silently eats typos**: this isn't introduced by you, but you're adding a keyword-only parameter to a constructor that swallows every unrecognized kwarg. `websocket_client_factory=...` misspelled as `websocket_factory=...` will not raise, will not warn, and will silently use the default — which for an injection seam is the single most confusing possible failure. Worth a follow-up issue to at least log unexpected kwargs at DEBUG.

**[TESTING GAPS]**

- [`tests/.../test_remote_conversation.py`, L225-252] The test is almost entirely `assert_called_once` on mocks, which is normally where I start complaining. I'll allow it here: the *feature is the seam*, so "was the injected factory used instead of the default, with these arguments" is the actual behavior under test, not a proxy for it. `default_factory.assert_not_called()` is the assertion that makes it meaningful — without it you'd only know your factory ran, not that it replaced anything.

  What's missing is the negative case: no test asserts that omitting `websocket_client_factory` still constructs the real `WebSocketCallbackClient`. That's the path 100% of existing users take, and right now nothing pins it.

**[QUESTION]**

- [L282-291] `WebSocketClientFactory` declares `__call__` returning `WebSocketCallbackClient`, and the default value is the *class* `WebSocketCallbackClient`. Structurally that's fine at runtime, but does pyright accept `type[WebSocketCallbackClient]` where `WebSocketClientFactory` is expected? If pre-commit is green, ignore me. If you had to think about it, a docstring note would save the next person the same thought.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 **LOW** — purely additive, keyword-only, defaults to current behavior. Nothing changes for any existing caller. Adds a public API surface (`WebSocketClientFactory`) that you'll have to keep, so it's worth being sure the signature is the one you want before it ships.

**KEY INSIGHT:** The `Protocol` is the right call over a bare `Callable` — it makes the five keyword arguments part of the contract, so an incompatible factory fails at type-check time instead of at WebSocket-connect time in production.

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
p1c2u: I tried this and it breaks the existing tests, so I've pushed a middle ground.

A default argument is bound once at function-definition time. The suite patches the module global in 87 patch() sites across 10 test files:

```python
@patch("openhands.sdk.conversation.impl.remote_conversation.WebSocketCallbackClient")
```

That rebinds the module attribute after the default has already been captured, so the default stops seeing the mock and each of those tests constructs the real client instead.

Applying both suggestions verbatim, two files with 7 patch sites between them give 7 failures — one per site, each burning the full 30s readiness timeout:

```
tests/sdk/conversation/test_conversation_factory.py:72
  conversation = Conversation(agent=agent, workspace=remote_workspace)
E openhands.sdk.conversation.exceptions.WebSocketConnectionError:
  WebSocket subscription did not complete within 30.0 seconds
7 failed, 4 passed in 210.52s
```
Extrapolating to all 87 sites, the suite doesn't realistically finish.

So I kept the | None = None sentinel but collapsed the five-line conditional to one, which removes the branch you were objecting to and keeps the WebSocketCallbackClient lookup late-bound:

```
factory = websocket_client_factory or WebSocketCallbackClient
self._ws_client = factory(
```
Since the parameter is typed WebSocketClientFactory | None, None is the only falsy value it can hold, so or is exact.

Happy to go with your original version instead if you'd prefer the cleaner signature — it just needs the 87 patch sites migrated to injecting the factory explicitly rather than patching the module global. Let me know which you'd rather have and I'll do it.

PR #4146, opened 2026-07-17
fix(visualizer): show per-request token usage alongside cumulative to…

### LINKED ISSUE #4105: Visualizer metrics subtitle shows only cumulative token usage, easily mistaken for per-request usage
**Problem**
The metrics subtitle rendered by `_format_metrics_subtitle()` in
`openhands-sdk/openhands/sdk/conversation/visualizer/default.py` only
displays `accumulated_token_usage`. Since the line appears under each
event, it's easy to misread the cumulative total as the size of the
latest request.

Example: first request ↑4.39K, second request ↑4.63K → the UI shows
↑9.02K under the second event, which looks like a huge context jump.

**Proposal**
Show both per-request and cumulative usage, e.g.:

`Tokens: request ↑4.63K ↓114 • total ↑9.02K ↓258 • cache hit 39.76% • $0.0015`

**Implementation notes**
- The per-request data already exists: `TokenUsage.per_turn_token` is
  preserved on the accumulated usage (`__add__` keeps the last turn's
  value). For a split ↑/↓ per request, `metrics.token_usages[-1]` works,
  but note `get_combined_metrics()` merges usages across all LLM
  usage_ids (agent, condenser, ...), so the merged list is not
  guaranteed chronological — the per-request value should come from the
  agent LLM's own metrics or be selected by response ordering.
- Tests: assert per-request and cumulative values are rendered
  separately.

I'd like to work on this — happy to open a PR.

luciobaiocchi:
HUMAN:

While debugging LLM costs I kept misreading the running totals shown under
each event as per-request token usage. This PR makes the visualizer show
both, so the numbers finally mean what they look like.

---

AGENT:

## Why

Fixes #4105. The metrics subtitle under each event showed only cumulative
totals ("↑ input 54k"), which read as per-request usage. Readers (including
while debugging cost issues) misinterpret running totals as single-request
numbers, and there was no way to tell the two apart.

## Summary

- `_format_metrics_subtitle()` now takes the event being rendered and shows
  per-request usage alongside the running total: `↑ input 4.63K (total 9.02K)`.
  Reasoning tokens and cache hit rate get the same per-request treatment.
- Per-request usage is found by matching `event.llm_response_id` against
  `TokenUsage.response_id` (`_find_request_usage()`), not by taking
  `token_usages[-1]` — the combined list merges agent/condenser usages and is
  not chronological across them.
- Events with no attributable request (user messages, `AgentErrorEvent`,
  unknown or per-session response ids as in ACP) fall back to totals-only —
  now labeled explicitly: `↑ input 54K (total) • … • $ (total 0.4231)`.
  Labeling was chosen over skipping because ACP events never match, so
  skipping would remove the subtitle entirely for ACP transcripts.
- Cost is always shown as `$ (total …)`: `Cost` entries carry no
  `response_id` (unlike `TokenUsage`), so per-request cost cannot be
  attributed today (see Notes).
- `DelegationVisualizer` forwards the event into `_format_metrics_subtitle`
  in both of its render paths, so delegate transcripts use the same format.

## Issue Number

Fixes #4105

## How to Test

```
uv run pytest tests/sdk/conversation/test_visualizer.py tests/tools/delegate/test_visualizer.py -v
```

7 new tests: per-request vs cumulative placement (full-string assertion),
per-request reasoning/cache-hit, response_id lookup vs naive last-index
(agent + condenser interleaved), labeled totals-only fallback (no event /
unknown response id), user-message fallback, delegate event forwarding, and
delegate labeled fallback. All 40 visualizer tests pass; `ruff` and
`pyright` clean.

Manual check: run any example (e.g.
`uv run python examples/01_standalone_sdk/01_hello_world.py`) and compare
the subtitle under an agent action (per-request + total) with the one under
a user message (labeled totals).

## Video/Screenshots

Subtitle for an event with attributable usage:

    Tokens: ↑ input 4.63K (total 9.02K) • cache hit 50.00% (total 36.67%) • ↓ output 114 (total 258) • $ (total 0.4231)

Fallback (user message / ACP / error events):

    Tokens: ↑ input 9.02K (total) • cache hit 36.67% (total) • ↓ output 258 (total) • $ (total 0.4231)

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Output change for previously totals-only subtitles: they now carry
  explicit `(total)` labels. No API changes; visualizer-only.
- Follow-up candidate: add `response_id` to `Cost` (as `ResponseLatency`
  already has) to enable true per-request cost. Kept out of this PR per
  review discussion.


openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -364,7 +365,33 @@ def _create_event_block(self, event: Event) -> Group | None:
...
+
+        Matches on ``TokenUsage.response_id``, not on list position: telemetry
+        records ``TokenUsage.response_id = resp.id`` and events carry
+        ``llm_response_id = llm_response.id``, so these ids correspond
+        one-to-one. We deliberately do NOT use
+        ``get_combined_metrics().token_usages[-1]`` because that list merges
+        usages across every usage id (agent, condenser, ...) and is not
+        chronological across them.
+        """
+        response_id = getattr(event, "llm_response_id", None)
```
VascoSch92: do we need getatts here?

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -364,7 +365,33 @@ def _create_event_block(self, event: Event) -> Group | None:
...
+        if not response_id:
+            return None
+
+        stats = self.conversation_stats
+        if not stats:
```
VascoSch92: you can use walrus operator here :-)

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -364,7 +365,33 @@ def _create_event_block(self, event: Event) -> Group | None:
...
+            return None
+
+        for metrics in stats.usage_to_metrics.values():
+            for usage in reversed(metrics.token_usages):
+                if usage.response_id == response_id:
+                    return usage
+
+        return None
```
VascoSch92:
```suggestion
```

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -355,7 +356,7 @@ def _create_event_block(self, event: Event) -> Group | None:
         title_color = config.color(event) if callable(config.color) else config.color
 
         # Build subtitle if needed
-        subtitle = self._format_metrics_subtitle() if config.show_metrics else None
+        subtitle = self._format_metrics_subtitle(event) if config.show_metrics else None
```
VascoSch92, line 359: `DelegationVisualizer` overrides `_create_event_block`, so it never reaches this line — it still calls `_format_metrics_subtitle()` with no event, at `openhands-tools/openhands/tools/delegate/visualizer.py:168` and `:247`. Since the new parameter defaults to `None` there's no type error and no test failure; it just silently does nothing.

What makes it worse is that it's only partial: `Condensation` and `AgentErrorEvent` fall through to `super()._create_event_block(event)` and do get per-request numbers, while `ActionEvent` and `MessageEvent` don't. So a single delegate transcript ends up mixing both formats. I rendered both to confirm:

```
--- Default ---
  ActionEvent : ↑ input 12K (total 52K) … ↓ output 350 (total 1.35K)
--- Delegation ---
  ActionEvent : ↑ input 52K … ↓ output 1.35K
```

This is a live path — `DelegationVisualizer` is exported from `openhands.tools.__init__` and used via `create_sub_visualizer` in `hooks/executor.py:252` and `delegate/impl.py:202`. Should just need `self._format_metrics_subtitle(event)` at both call sites.

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -364,7 +365,33 @@ def _create_event_block(self, event: Event) -> Group | None:
...
         )
 
-    def _format_metrics_subtitle(self) -> str | None:
+    def _find_request_usage(self, event: Event | None) -> TokenUsage | None:
+        """Find the TokenUsage for the LLM request that produced ``event``.
+
+        Matches on ``TokenUsage.response_id``, not on list position: telemetry
+        records ``TokenUsage.response_id = resp.id`` and events carry
+        ``llm_response_id = llm_response.id``, so these ids correspond
+        one-to-one. We deliberately do NOT use
```
VascoSch92: Small thing: the one-to-one claim doesn't hold on the ACP path. `acp_agent.py:1759` records `response_id=session_id`, which is shared across every call in a session rather than being per-request, while the `ActionEvent` at `:3113` carries an unrelated `uuid.uuid4()`. So ACP always falls back to the cumulative-only format.

That's fine as behaviour, but worth softening the wording here so nobody later relies on the guarantee.

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -404,11 +442,11 @@ def abbr(n: int | float) -> str:
 
         # Build with fixed color scheme
         parts: list[str] = []
-        parts.append(f"[cyan]↑ input {input_tokens}[/cyan]")
+        parts.append(f"[cyan]{input_str}[/cyan]")
         parts.append(f"[magenta]cache hit {cache_rate}[/magenta]")
         if reasoning_tokens > 0:
             parts.append(f"[yellow] reasoning {abbr(reasoning_tokens)}[/yellow]")
```
VascoSch92: This still reads from the cumulative `usage`, so together with the change above we end up with per-request input/output but a cumulative reasoning count sandwiched between them:

```
↑ input 12K (total 52K) • cache hit 0.00% • reasoning 8.2K • ↓ output 350 (total 1.35K)
```

Per-request reasoning there is actually 200, not 8.2K. Unlabelled and sitting between two explicitly per-request numbers it reads as per-request, which is the exact confusion #4105 is about — so for reasoning models this is arguably worse than before the PR.

`reasoning_tokens` is recorded per call (`telemetry.py:244`), so `request_usage.reasoning_tokens` is already available. Can we give it the same `X (total Y)` treatment inside the `request_usage is not None` branch? Same question for `cache hit` two lines up — lower stakes since it's a rate, but worth making it a deliberate choice either way.

tests/sdk/conversation/test_visualizer.py
```diff
@@ -609,6 +609,157 @@ def test_metrics_abbreviation_formatting():
...
+        tool_name="test",
+        tool_call_id="call_1",
+        tool_call=tool_call,
+        llm_response_id="response_2",
+    )
+
+    subtitle = visualizer._format_metrics_subtitle(event)
+    assert subtitle is not None
+    # Per-request usage (the second request only).
+    assert "4.63K" in subtitle
```
VascoSch92: Minor, but these assertions can't distinguish per-request from total. The rendered subtitle is:

```
Tokens: ↑ input 4.63K (total 9.02K) • cache hit 0.00% • ↓ output 114 (total 258) • $ 0.00
```

If the two numbers were swapped — `input 9.02K (total 4.63K)` — all four `in subtitle` checks would still pass, since both tokens are present either way. Nothing pins the placement.

Asserting the full formatted string instead would fix that, and would also have caught the reasoning-token issue I flagged in `default.py`, which no test currently covers.

VascoSch92 (changes requested):

luciobaiocchi:
Thanks for the thorough review, @VascoSch92 — addressed all of it in 6c7d08fc:

  - **`getattr` on `_find_request_usage`**: replaced with an explicit `isinstance(event, (ActionEvent, MessageEvent, Condensation))` check against the classes that
  actually declare `llm_response_id`, so it's properly typed instead of silently swallowing mismatches.
  - **Walrus operator**: applied to the `conversation_stats` check.
  - **Redundant `return None`**: removed the trailing one; the declared `TokenUsage | None` return type already covers the implicit fallthrough (confirmed clean 
  with `pyright`).
  - **`DelegationVisualizer` not forwarding `event`**: this was the real bug — fixed both call sites (`visualizer.py:168` and `:247`) to pass 
  `self._format_metrics_subtitle(event)`. Added a regression test (`test_delegation_visualizer_action_event_shows_per_request_usage`) since nothing previously 
  caught this.
  - **Docstring overclaiming a 1:1 `response_id` correspondence**: reworded to call out that the ACP path does not guarantee it and falls back to cumulative-only,
  matching your note about `acp_agent.py`.
  - **Reasoning tokens / cache hit staying cumulative-only**: both now get the same `X (total Y)` treatment as input/output when a per-request match exists (cache
  hit reuses the existing `MetricsSnapshot.cache_hit_rate` property instead of duplicating the formula). Added
  `test_metrics_subtitle_reasoning_and_cache_hit_are_per_request` to cover the exact case you flagged as untested.
  - **Weak test assertions**: replaced the loose `in subtitle` checks in `test_metrics_subtitle_shows_per_request_and_cumulative` with a single exact full-string
  assertion, so number placement is actually pinned down.

  `ruff check`, `ruff format --check`, and `pyright` are clean on the changed files; the visualizer test suites (38 tests) pass.

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -364,7 +365,36 @@ def _create_event_block(self, event: Event) -> Group | None:
...
+            return None
+
+        response_id = event.llm_response_id
+        if not response_id:
+            return None
+
+        if not (stats := self.conversation_stats):
+            return None
+
+        for metrics in stats.usage_to_metrics.values():
```
VascoSch92: nit

```
return next(
    (
        usage
        for metrics in stats.usage_to_metrics.values()
        for usage in reversed(metrics.token_usages)
        if usage.response_id == response_id
    ),
    None,
)
```

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -391,24 +421,52 @@ def abbr(n: int | float) -> str:
...
-        parts.append(f"[cyan]↑ input {input_tokens}[/cyan]")
+        parts.append(f"[cyan]{input_str}[/cyan]")
         parts.append(f"[magenta]cache hit {cache_rate}[/magenta]")
-        if reasoning_tokens > 0:
-            parts.append(f"[yellow] reasoning {abbr(reasoning_tokens)}[/yellow]")
-        parts.append(f"[blue]↓ output {output_tokens}[/blue]")
+        if show_reasoning:
+            parts.append(f"[yellow] {reasoning_str}[/yellow]")
+        parts.append(f"[blue]{output_str}[/blue]")
         parts.append(f"[green]$ {cost_str}[/green]")
```
VascoSch92: Cost is still bare cumulative:

```
↑ input 12K (total 54K) • cache hit 75.00% (total 90.74%) • ↓ output 350 (total 650) • $ 0.4231
```

Everything else on the line is per-request now, so the `$` at the end reads as per-request too. Same thing we just fixed for reasoning, and it's the number people actually care about when they're chasing spend.

Can't do the same lookup for it though, `Cost` has no `response_id` (`ResponseLatency` right below it does). Adding one feels like a separate PR. Can we just print `$ (total 0.4231)` for now so nothing on the line is naked?

VascoSch92 (commented):
Rebase looks good and last round's stuff is all handled: reasoning and cache hit are per-request now, delegate visualizer passes the event through in both spots.

I did go check whether the response_id lookup is actually safe after a resume, since that was my worry. It is. `restore_metrics` shares the object by reference so `usage_to_metrics` doesn't go stale, and `merge` concatenates per usage-id so `token_usages[-1]` genuinely would have been wrong. Good call there.

Two things left and they're really the same thing: numbers on that line that aren't labelled. Inline.

Also description still says three new tests, there's five.

tests/tools/delegate/test_visualizer.py
```diff
@@ -188,6 +189,60 @@ def test_delegation_visualizer_action_event():
     assert "Lodging Expert Agent Action" in str(block.renderables[0])
 
 
+def test_delegation_visualizer_action_event_shows_per_request_usage():
+    """Regression: DelegationVisualizer overrides _create_event_block and must
+    forward the event into _format_metrics_subtitle. Without it, ActionEvent
+    subtitles silently stay cumulative-only while Condensation/AgentErrorEvent
```
VascoSch92: `AgentErrorEvent` doesn't show per-request numbers. No `llm_response_id`, and it's not in the isinstance tuple either. Only `Condensation` does.

Test is fine, it's just this comment. Worth fixing because it reads like the `AgentErrorEvent` case is already handled when it isn't (other comment).

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -364,7 +365,36 @@ def _create_event_block(self, event: Event) -> Group | None:
...
+        ``llm_response_id = llm_response.id``, which line up for most LLM
+        paths. That correspondence isn't guaranteed everywhere though -- the
+        ACP path records one ``response_id`` per session rather than per
+        request, so it never matches here and the caller falls back to the
+        cumulative-only format. We deliberately do NOT use
+        ``get_combined_metrics().token_usages[-1]`` because that list merges
+        usages across every usage id (agent, condenser, ...) and is not
+        chronological across them.
+        """
+        if not isinstance(event, (ActionEvent, MessageEvent, Condensation)):
```
VascoSch92, line 382: `AgentErrorEvent` and user messages both have `show_metrics=True` but neither can ever match here. `AgentErrorEvent` has no `llm_response_id` field at all, user `MessageEvent`s leave it `None`. So with `skip_user_messages` off, which is the default:

```
ActionEvent       : ↑ input 12K (total 54K) • cache hit 75.00% (total 90.74%) • ↓ output 350 (total 650)
user MessageEvent : ↑ input 54K             • cache hit 90.74%                • ↓ output 650
AgentErrorEvent   : ↑ input 54K             • cache hit 90.74%                • ↓ output 650
```

That bare 54K is the original bug, still there, now sitting directly under the fixed version. Same mixing problem as the delegate one from last round, just reached a different way.

Either skip metrics on events we can't attribute, or have the fallback say `(total 54K)` too.

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
luciobaiocchi: Done $ (total 0.4231) as suggested. 

Agreed that a real per-request cost means adding response_id to Cost (like ResponseLatency has); happy to open a follow-up PR for that if you think it’s worth it :)

tests/tools/delegate/test_visualizer.py
luciobaiocchi: fixed:

The docstring now says only Condensation carries an llm_response_id. AgentErrorEvent goes through the fallback, which now labels everything (total) so its numbers can’t be misread as per-request.

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
VascoSch92: i think for now is good.

VascoSch92 (approved):
LGTM. 

There is just one thing with parallel tool execution that I saw just now.  `response_dispatch.py:205-210` emits one `ActionEvent` per tool call, all sharing the same `llm_response_id`, so a parallel batch repeats the identical per-request numbers.

Therefore, you will have duplication in the output (but this is just missleading is not affecting the final results). 

It is not a blocker. But if you could open a follow up PR would be great :-) (but up to you)

luciobaiocchi:
> LGTM. 
> 
> There is just one thing with parallel tool execution that I saw just now.  `response_dispatch.py:205-210` emits one `ActionEvent` per tool call, all sharing the same `llm_response_id`, so a parallel batch repeats the identical per-request numbers.
> 
> Therefore, you will have duplication in the output (but this is just missleading is not affecting the final results). 
> 
> It is not a blocker. But if you could open a follow up PR would be great :-) (but up to you)

yes for sure, i will open an issue first

PR #4154, opened 2026-07-20
feat(agent-server): add GET /api/llm/balance endpoint (OpenRouter credit balance)

NelsonScott:
## Motivation

Users running OpenHands against OpenRouter want to see their remaining credits in the UI without leaving the app. The agent-server already stores the (encrypted) API key for the active LLM and for saved profiles, so it can query the provider's balance endpoint server-side and expose a safe, key-free summary. A companion agent-canvas UI PR that consumes this endpoint is coming.

## API contract

`GET /api/llm/balance?profile=<name>`

- `profile` (optional): LLM profile name to check. When omitted, the endpoint uses the currently active agent settings LLM (the same config the server resolves for conversations).
- The provider is detected as OpenRouter when the resolved LLM has a `base_url` containing `openrouter.ai` or a model prefixed with `openrouter/`. The server then calls `GET https://openrouter.ai/api/v1/key` with the stored API key (decrypted server-side, 10s timeout). The key is never returned to the client.

**200 response:**

```json
{
  "provider": "openrouter",
  "limit": 25.0,
  "limit_remaining": 18.5,
  "usage": 6.5,
  "usage_daily": 0.5,
  "usage_weekly": 2.0,
  "usage_monthly": 6.5,
  "is_free_tier": false
}
```

`limit`, `limit_remaining`, and the `usage_*` fields are nullable floats.

**Errors:**

- `404` — the named profile does not exist, the resolved provider does not support balance lookup, or no API key is configured for it (clear `detail` message in each case).
- `502` — the upstream OpenRouter call failed (HTTP error, timeout, or unexpected response shape).

## Implementation notes

- Lives in `llm_router.py` next to the other `/api/llm` endpoints, following the surrounding conventions (pydantic response model, `httpx.AsyncClient`, `HTTPException` with `status` constants).
- A small provider -> fetcher mapping (`_BALANCE_FETCHERS`) makes it easy to add other providers with balance/credits endpoints later without changing the route.
- Tests mock the OpenRouter HTTP call via `httpx.MockTransport` and cover: success (including no-key-leak assertion), base_url-based detection, unsupported provider, missing API key, upstream HTTP/network failures, `profile` lookup, and unknown profile.

## Testing

- `uv run pytest tests/agent_server/test_llm_router.py` — 24 passed.
- Full `tests/agent_server/` suite passes except a pre-existing, unrelated `test_terminal_service` failure that also fails on a clean `main` checkout in my environment.
- `pre-commit` (ruff format/lint, pycodestyle, pyright, import rules) passes on the changed files.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/llm_router.py
+++ b/openhands-agent-server/openhands/agent_server/llm_router.py
@@ -5,11 +5,20 @@ from __future__ import annotations
 import asyncio
 import secrets
 import time
+from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 
-from fastapi import APIRouter, HTTPException, Query, status
-from pydantic import BaseModel, Field
+import httpx
+from fastapi import APIRouter, HTTPException, Query, Request, status
+from pydantic import BaseModel, Field, SecretStr
 
+from openhands.agent_server._secrets_exposure import get_cipher, get_config
+from openhands.agent_server.persistence import (
+    PersistedSettings,
+    get_llm_profile_store,
+    get_settings_store,
+)
+from openhands.sdk.llm import LLM
 from openhands.sdk.llm.auth.openai import (
     DEVICE_CODE_TIMEOUT_SECONDS,
     OPENAI_CODEX_MODELS,
@@ -58,6 +67,19 @@ class VerifiedModelsResponse(BaseModel):
     """Response containing verified LLM models organized by provider."""
 
     models: dict[str, list[str]]
+
+
+class BalanceResponse(BaseModel):
+    """Credit balance information for the resolved LLM's provider."""
+
+    provider: str
+    limit: float | None = None
+    limit_remaining: float | None = None
+    usage: float
+    usage_daily: float | None = None
+    usage_weekly: float | None = None
+    usage_monthly: float | None = None
+    is_free_tier: bool = False
 
 
 class SubscriptionStatusResponse(BaseModel):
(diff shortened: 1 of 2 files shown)


neubig (commented):
Thanks for putting this together. After looking at the API shape, we think upstream provider budget/quota probing belongs in LiteLLM because discovery, credentials, and response semantics will differ across providers. We opened BerriAI/litellm#34734 to propose a capability-based abstraction.

We'd like to wait a little and see whether LiteLLM is willing to consider this before adding an OpenRouter-specific endpoint here. If they aren't, we can implement the abstraction in software-agent-sdk, but we'd still like the public contract and routing to be provider-neutral: optional provider adapters returning a normalized quota snapshot and a clear unsupported result.

Initial support should cover at least LiteLLM Proxy (`litellm_proxy`, using its configured provider budgets) and OpenRouter (using its key endpoint), while preserving their differing scope, unit, and reset semantics. That would let us add other providers later without changing the agent-server or UI contract.

Could we pause this PR briefly while we see how LiteLLM responds?

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-agent-server/openhands/agent_server/llm_router.py`, lines 250–254] **Credential routing**: provider detection uses a case-sensitive substring search over the entire `base_url`. It classifies both `https://openrouter.ai.evil.example/v1` and `https://evil.example/openrouter.ai/v1` as OpenRouter, while rejecting the legitimate `https://OPENROUTER.AI/api/v1`. For a custom endpoint whose URL merely contains that text, this sends the endpoint's stored API key to OpenRouter's `/api/v1/key`, crossing provider trust boundaries. Parse the URL and compare a normalized hostname to the exact allowed OpenRouter host (or an explicit allowlist), then add spoofed-host and case-normalization tests.

**[RISK ASSESSMENT]**
This adds an outbound third-party integration that handles plaintext API keys. The current provider classifier can route a custom provider credential to the wrong external service. **Recommendation**: Do not auto-merge. Request review from a human security reviewer to validate URL/provider binding and credential handling.

**VERDICT:**
❌ **Needs rework**: Secrets must be routed from parsed, exact provider identity—not substring coincidence.

**KEY INSIGHT:**
Provider detection is a security boundary once it decides where a credential will be transmitted.

---

>
>
>

NelsonScott:
> neubig

hey @neubig, that makes sense, and if I followed the references right it seems like this/something like it may effectively be in place https://github.com/OpenHands/OpenHands/pull/16311 
Haven't looked extensively so let me know if this PR still makes sense or if its better to close, either way is fine

PR #4155, opened 2026-07-20
fix(terminal): submit multiline PowerShell commands on Windows

### LINKED ISSUE #4133: [BUG]: WindowsTerminal does not submit multiline PowerShell commands, preventing PS1 metadata execution
## Summary

On Windows, multiline commands executed through `WindowsTerminal` remain at PowerShell's `>>` continuation prompt and eventually time out.

The original command is not executed, and the PS1 metadata suffix appended by OpenHands is echoed as pending input instead of being executed.

This affects multiline commands generally and is not specific to Python or invalid quoting.

## Environment

- OpenHands SDK: 1.36.1
- OS: Windows
- Terminal backend: `WindowsTerminal`
- Shell: `powershell.exe`
- Python: 3.12

## Minimal reproduction

```python
from openhands.tools.terminal.definition import TerminalAction
from openhands.tools.terminal.impl import TerminalExecutor

executor = TerminalExecutor(
    working_dir=".",
    terminal_type="powershell",
    shell_path="powershell.exe",
)

try:
    observation = executor(
        TerminalAction(
            command='Write-Output "a\nb"',
            timeout=2,
        )
    )

    print("exit_code:", observation.exit_code)
    print("output:", observation.text)
finally:
    executor.close()
```

## Actual behavior

The command remains at the PowerShell continuation prompt:

```text
Write-Output "a
>> b"; $oh1 = $?; $oh2 = $LASTEXITCODE; ...
>>
```

The observation eventually returns:

```text
exit_code: -1
```

The output contains OpenHands' internal metadata suffix, but no new `###PS1JSON###` metadata block is produced because the suffix was never executed.

## Expected behavior

The command should execute and return:

```text
a
b
```

with exit code `0`, followed by a valid PS1 metadata block.

More generally, multiline commands should behave the same way as when they are submitted directly to the same PowerShell session.

## Root cause

`WindowsTerminal.send_keys()` currently appends the metadata suffix and sends only one trailing newline:

```python
stripped_text = text.rstrip()

if stripped_text:
    self._command_running_event.set()
    command = f"{stripped_text}; {self._metadata_suffix()}"
else:
    command = text

if enter and not command.endswith("\n"):
    command += "\n"

self._write_to_stdin(command)
```

When PowerShell receives a multiline statement through the persistent redirected stdin session, one trailing newline leaves it at the `>>` continuation prompt.

As a result:

1. The original command is not executed.
2. The appended metadata suffix is not executed.
3. No new PS1 metadata marker is generated.
4. `TerminalSession` waits until timeout.
5. The echoed internal suffix is returned as command output.

## Verified workaround

Submitting an additional blank line for multiline commands fixes the issue:

```python
if enter:
    if "\n" in stripped_text or "\r" in stripped_text:
        command += "\n\n"
    elif not command.endswith("\n"):
        command += "\n"
```

After this change:

- The original multiline command executes.
- Its stdout/stderr is preserved.
- `$LASTEXITCODE` is captured correctly.
- The PS1 metadata suffix executes normally.
- The terminal no longer times out.

I also verified this using a multiline `python -c` command containing invalid Python. Before the change, the command timed out and exposed the internal metadata suffix. After the change, the original Python `SyntaxError` was returned with exit code `1`, followed by a valid metadata block.

## Suggested regression tests

1. A one-line PowerShell command still executes with one trailing newline.
2. A multiline PowerShell string executes successfully.
3. A multiline `python -c` command returns its real stdout/stderr.
4. A failing multiline command returns its real non-zero exit code.
5. Internal PS1 metadata code is not exposed in the observation.

VascoSch92:
HUMAN:

Fixing a bug with the PowerShell.

---

AGENT:

## Why

On Windows, multiline commands executed through `WindowsTerminal` never run.
`send_keys()` appends the PS1 metadata suffix and terminates the input with a
single trailing newline. That is enough for a one-line command, but PowerShell's
interactive REPL keeps a multiline statement at the `>>` continuation prompt
until it receives a **blank line**. With only one newline, the session sits at
`>>`, so neither the original command nor the appended metadata suffix ever
executes: no `###PS1END###` marker is produced, `TerminalSession` waits out the
timeout, returns `exit_code: -1`, and echoes OpenHands' internal metadata suffix
as the command output.

This affects multiline commands in general (not specific to Python or quoting).

## Summary

- `WindowsTerminal.send_keys()` now appends a blank line (`\n\n`) when the
  command contains an embedded newline, so PowerShell submits the continuation
  buffer and executes the command plus its metadata suffix.
- Single-line commands and control-key input are unchanged (still one trailing
  newline / no newline for `enter=False`).
- Added a Windows regression test that runs a multiline `Write-Output` command
  and asserts it executes with `exit_code == 0`.

## Issue Number

Closes #4133

## How to Test

Reviewer test (Windows, PowerShell), using the reproduction from the issue:

```python
from openhands.tools.terminal.definition import TerminalAction
from openhands.tools.terminal.impl import TerminalExecutor

executor = TerminalExecutor(
    working_dir=".",
    terminal_type="powershell",
    shell_path="powershell.exe",
)
try:
    obs = executor(TerminalAction(command='Write-Output "a\nb"', timeout=2))
    print("exit_code:", obs.exit_code)   # before: -1  |  after: 0
    print("output:", obs.text)           # after: contains "a" and "b"
finally:
    executor.close()
```

Automated test (runs on the `windows-tests` / `windows-latest` CI job, which
covers `tests/tools/`):

```
uv run pytest tests/tools/terminal/test_windows_terminal.py -k multiline
```

### Evidence gathered on this (non-Windows) machine

I don't have a Windows/PowerShell host, so I could not drive the real backend
end-to-end locally; the behavioral test is gated to Windows and runs in CI. What
I verified locally:

- Drove the **real** patched `send_keys()` (with a stubbed live process) and
  confirmed the exact bytes written to stdin:
  - multiline `Write-Output "a\nb"` → ends with `\n\n` (blank line submits the
    `>>` continuation buffer)
  - single-line `Get-ChildItem` → ends with exactly one `\n` (unchanged)
- Confirmed the byte stream before the fix ended in a single `\n` after the
  multiline block, matching the hang described in the issue.
- `uv run pre-commit run --files ...` passes on both changed files (ruff format,
  ruff lint, pycodestyle, pyright).
- The new test collects and skips cleanly on non-Windows.

## Video/Screenshots

N/A — terminal backend change with no GUI surface. See the byte-level evidence
above; end-to-end PowerShell behavior is exercised by the Windows CI job.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Backward compatible: only commands with an embedded newline change behavior;
  the common single-line path is byte-identical to before.
- A uniform "always append `\n\n`" approach was intentionally avoided: for a
  single-line command the extra empty line is not consumed by a continuation
  submit and would trigger a spurious extra prompt. The conditional sends
  exactly the newlines PowerShell consumes to return to a fresh prompt.


neubig (commented):
On unix, I don't think that we run multi-line commands either, but rather throw an error telling the agent not to run multi-line commands. Could you check that this behavior is consistent with the unix terminal, and if not fix it to be consistent?

VascoSch92:
@neubig 

I checked, and the unix terminal does run multi-line commands. What it rejects is multiple statements.

A single statement spanning several lines returns one element and runs; two statements joined by a newline are what produce "Cannot execute multiple commands at once." Newlines split statements, `&&` and `;` don't, which is what the tool description already tells the agent.

The existing tests pin this behaviour.

Windows is consistent with this

Moreover, I added coverage in response to this review:

Following the issue, I also added a small script `.pr/check_multiline_consistency.py` reproduces the full analysis and exits non-zero on mismatch.

PS: a PowerShell here-string (`@" … "@`) gets split into two statements by the bash-based parser and is rejected on Windows. But this is another problem. I can fill an issue for that and fix separately.

neubig (approved):
OK, LGTM thanks!

PR #4163, opened 2026-07-21
fix(file_editor): stop insert gluing text onto a newline-less last line

JiataiWang:
HUMAN:

Inserting at the end of a file without a trailing newline was corrupting the last line by gluing the new text onto it. I reproduced the bug before/after the fix and added regression tests for both the insert and the full-file view behavior.

---

AGENT:

## Why

Two related trailing-newline defects in the file editor:

1. `insert` reads existing lines with their newlines preserved. When the file's last line has no trailing newline and the insert position is the end of the file, the first inserted line is concatenated onto that last line (data corruption). Files without a trailing newline are common (many are produced by the agent itself).
2. The full-file `view` path numbered `content.split("\n")` directly, so a file ending in `\n` rendered a phantom extra empty numbered line — a different line count than the equivalent range `view` (which already normalizes via `splitlines()`) and than `cat -n`.

## Summary

- `insert`: terminate the carried-over last line before appending, so text isn't glued onto a newline-less last line.
- `view` (full file): normalize trailing newlines like the range path, removing the phantom line.
- Add regression tests for both.

## Issue Number

## How to Test

Reproduced directly against the tool before/after the fix:

```python
from openhands.tools.file_editor.editor import FileEditor
ed = FileEditor()
# file content "a\nb\nc" (no trailing newline)
ed.insert(path, insert_line=3, new_str="X")
# BEFORE: "a\nb\ncX\n"  (c and X glued)   AFTER: "a\nb\nc\nX\n"
```
Full-file `view` of `"line1\nline2\nline3\n"` showed a phantom line `4` before the fix; after, it matches the `view_range=[1,3]` output and `cat -n`.

Tests: `.venv/bin/python -m pytest tests/tools/file_editor/ -q` — 158 passed (incl. the 2 new regression tests). `ruff check` / `ruff format --check` clean.

## Type

- [x] Bug fix

## Notes

### DIFF
--- a/openhands-tools/openhands/tools/file_editor/editor.py
+++ b/openhands-tools/openhands/tools/file_editor/editor.py
@@ -369,7 +369,12 @@ class FileEditor:
         start_line = 1
         if not view_range:
             file_content = self.read_file(path)
-            output = self._make_output(file_content, str(path), start_line)
+            # Normalize trailing newlines the same way the range path does, so a
+            # file ending in "\n" doesn't render a phantom extra numbered line
+            # (this path emulates `cat -n`, which shows no such trailing line).
+            output = self._make_output(
+                "\n".join(file_content.splitlines()), str(path), start_line
+            )
 
             return FileEditorObservation.from_text(
                 text=output,
@@ -573,6 +578,12 @@ class FileEditor:
                     break
                 new_lines.append(line)
                 history_lines.append(line)
+
+        # When inserting at the end of a file whose last line has no trailing
+        # newline, terminate that carried-over line first so the inserted text
+        # starts on its own line instead of being glued onto it.
+        if new_lines and not new_lines[-1].endswith("\n"):
+            new_lines[-1] += "\n"
 
         # Insert new content
         for line in new_str_lines:
--- a/openhands-tools/openhands/tools/file_editor/editor.py
+++ b/openhands-tools/openhands/tools/file_editor/editor.py
@@ -369,12 +369,7 @@ class FileEditor:
         start_line = 1
         if not view_range:
             file_content = self.read_file(path)
-            # Normalize trailing newlines the same way the range path does, so a
-            # file ending in "\n" doesn't render a phantom extra numbered line
-            # (this path emulates `cat -n`, which shows no such trailing line).
-            output = self._make_output(
-                "\n".join(file_content.splitlines()), str(path), start_line
-            )
+            output = self._make_output(file_content, str(path), start_line)
 
             return FileEditorObservation.from_text(
                 text=output,
--- a/tests/tools/file_editor/test_basic_operations.py
+++ b/tests/tools/file_editor/test_basic_operations.py
@@ -324,6 +324,28 @@ def test_view_with_a_specific_range(editor):
     assert "101" not in result.text
 
 
+def test_view_full_file_with_trailing_newline_has_no_phantom_line(tmp_path):
+    """A full-file view of a file ending in a newline must not render a phantom
+    extra numbered line (it emulates ``cat -n``), and must match the equivalent
+    explicit range view."""
+    editor = FileEditor()
+    test_file = tmp_path / "trailing.txt"
+    test_file.write_text("line1\nline2\nline3\n")
+
+    full = editor(command="view", path=str(test_file))
+    assert isinstance(full, FileEditorObservation)
+    assert full.text == (
+        f"Here's the result of running `cat -n` on {test_file}:\n"
+        "     1\tline1\n"
+        "     2\tline2\n"
+        "     3\tline3\n"
+    )
+
+    # The full view and the equivalent range view must agree on line count.
+    ranged = editor(command="view", path=str(test_file), view_range=[1, 3])
+    assert full.text == ranged.text
+
+
 def test_create_file(editor):
     editor, test_file = editor
     new_file = test_file.parent / "new_file.txt"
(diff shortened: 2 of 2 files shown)


enyst (commented):

**Taste Rating: 🔴 Needs improvement** — the fix in the title is correct and I'd merge it today. The *other* fix, the one not in the title, makes the tool lie about file contents.

**[CRITICAL ISSUES]**

- [`openhands-tools/openhands/tools/file_editor/editor.py`, L372-376] **Data Corruption In The View Path**: `"\n".join(file_content.splitlines())` is not a trailing-newline normalizer. `str.splitlines()` breaks on `\v`, `\f`, `\r`, `\x1c`-`\x1e`, `\x85`, ` `, ` ` — and rejoining with `\n` silently rewrites every one of them into a newline.

  Concretely, with a file containing a form feed (`\x0c` — common in Python, Lisp, and C sources as a page separator):

  ```
  $ cat -n /tmp/ff.txt | cat -v          # ground truth
       1  def a():
       2      pass
       3  ^Ldef b():
       4      pass
  ```

  ```
  === main, full view ===              === this PR, full view ===
       1  def a():                          1  def a():
       2      pass                          2      pass
       3  ^Ldef b():                        3
       4      pass                          4  def b():
       5                                    5      pass
  ```

  `main` is content-faithful and has one cosmetic bug (the phantom line 5). Your branch fixes the phantom line and, in exchange, destroys the form feed and invents a line. The header still says "the result of running `cat -n`", which is now false.

  This matters more than usual for this specific tool: the agent reads the view output and then issues `str_replace` against what it saw. If the view shows `\n` where the file has `\x0c`, the replacement won't match — or worse, will match the wrong thing.

- [`tests/tools/file_editor/test_basic_operations.py`, L344] **The Test Pins The Wrong Invariant**: `assert full.text == ranged.text` looks like the obviously-correct consistency check. It isn't, because the range path is *already* the broken one:

  ```
  === main, range view [1,4] ===
       1  def a():
       2      pass
       3
       4  def b():
       5      pass
  ```

  That's `main`, unmodified, corrupting the form feed **and** returning five numbered lines for a four-line request. So this PR reaches consistency by moving the good path down to the bad one. If consistency is the goal — and it should be — fix `_make_output`/the range path to split on `\n` only, and let the full path keep its fidelity.

- [whole PR] **Two Fixes, One Title**: "stop insert gluing text onto a newline-less last line" describes the second hunk. The first hunk changes the rendering of *every full-file view in the product* and is not mentioned in the title. Split it. The insert fix can merge immediately; the view change needs the discussion above.

**[TESTING GAPS]**

- `uv run pytest tests/tools/file_editor/test_basic_operations.py` → **43 passed**, and the insert test is a good one: it asserts on the resulting file bytes rather than on a mock, and it would fail without your `new_lines[-1] += "\n"`.
- Missing: any test with a non-`\n` line terminator in the content. That's the gap that let this through, and one `write_bytes(b'a\n\x0cb\n')` case would close it permanently.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🔴 **HIGH** — the view path is how the agent perceives every file it edits, so a fidelity bug here degrades every downstream edit rather than failing loudly. The insert half is 🟢 low risk in isolation. **Recommendation**: Do not auto-merge as one unit. Land the insert fix; hold the view change for a human decision on which path should be normative.

**VERDICT:** ❌ **Needs rework** — split the PR. The insert fix is genuinely good work and shouldn't be held hostage by the view change.

**KEY INSIGHT:** "Make the two code paths agree" is only an improvement if you check which one was right first — here the path with the cosmetic bug was the one telling the truth.

JiataiWang:
You're right, and the form-feed demo settled it — I picked the wrong path as normative. `splitlines()` treating `\x0c`/`\v`/`\x85` as line breaks means the "normalization" was actually rewriting file content in the view the agent edits against.

Pushed 3ab0b42b, splitting as you recommended:
- **Reverted** the full-view hunk. The PR is now insert-fix only (title finally matches the diff).
- **Dropped** the `full == ranged` consistency test — as you showed, it reached consistency by dragging the faithful path down to the broken one.
- **Added** the missing regression test: a file containing `\x0c` must render it as-is, glued to its numbered line (`write_bytes(b"...\x0cdef b():...")` → asserts `"3\t\x0cdef b():"` appears verbatim).

The phantom-trailing-line cosmetic issue and the range path's own splitlines corruption are now untouched pre-existing behavior — I can file a follow-up issue laying out the "which path should be normative" question for a human call, if useful.

enyst (commented):

🟢 **Good taste**

The change fixes the corruption at the single ownership boundary where preserved file lines meet inserted text. It leaves history content untouched, writes atomically through the existing path, and the regression test proves the exact newline-less EOF case; the additional view test protects the fidelity issue discovered during revision.

**Repository policy note:** This changes file-editor tool behavior and can plausibly affect agent/evaluation performance. No completed eval-monitor run with human confirmation is linked, so a human maintainer should make the merge decision after lightweight eval validation.

**[RISK ASSESSMENT]**
The production change is only two effective lines and is well targeted, but file editing is benchmark-sensitive core tool behavior. Focus validation on insert positions around EOF and files with unusual line terminators.

**VERDICT:**
✅ **Worth merging**: The implementation is minimal, behavior-preserving outside the broken EOF case, and covered by a real file round trip, subject to the repository's human eval check.

**KEY INSIGHT:**
Terminate the carried-over record before appending new records; do not normalize the whole file to repair one boundary.

PR #4165, opened 2026-07-21
fix(llm): don't crash prompt caching on an empty-content trailing message

JiataiWang:
HUMAN:

An empty-content trailing tool/user message crashed prompt caching with an IndexError that aborted the whole completion. I reproduced it and added a regression test covering the empty-message case.

---

AGENT:

## Why

`_apply_prompt_caching` marks the last user/tool message's final content block with `message.content[-1].cache_prompt = True`. When that message has empty content — e.g. a tool observation that produced no text or image, or an empty trailing user message — `content[-1]` raised `IndexError` and aborted the entire completion. The system-message branch above already length-checks its content; the user/tool loop did not.

## Summary

- Skip empty-content user/tool messages so the cache breakpoint falls through to the previous non-empty message instead of crashing.
- Add a regression test.

## Issue Number

## How to Test

Reproduced before/after:

```python
llm._apply_prompt_caching([
    Message(role="system", content=[TextContent(text="sys")]),
    Message(role="user", content=[TextContent(text="hi")]),
    Message(role="tool", content=[]),   # empty tool observation
])
# BEFORE: IndexError: list index out of range
# AFTER: no error; breakpoint set on the previous non-empty message
```
Tests: `.venv/bin/python -m pytest tests/sdk/llm/test_prompt_caching_cross_conversation.py -q` — 7 passed. `ruff` clean.

## Type

- [x] Bug fix

## Notes

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -2579,7 +2579,11 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         # extends every turn. Anthropic-only; Gemini is excluded from
         # PROMPT_CACHE_MODELS because its cache can't extend this way.
         for message in reversed(messages):
-            if message.role in ("user", "tool"):
+            # Skip empty-content user/tool messages (e.g. a tool observation that
+            # produced no text/image): there is no content block to mark, and
+            # indexing content[-1] on an empty list would raise IndexError. Fall
+            # through to the previous non-empty user/tool message instead.
+            if message.role in ("user", "tool") and message.content:
                 message.content[
                     -1
                 ].cache_prompt = True  # Last item inside the message content
--- a/tests/sdk/llm/test_prompt_caching_cross_conversation.py
+++ b/tests/sdk/llm/test_prompt_caching_cross_conversation.py
@@ -144,6 +144,35 @@ def test_end_to_end_caching_flow(tmp_path, dynamic_context, expect_dynamic):
     assert messages[1].content[-1].cache_prompt is True
 
 
+def test_apply_prompt_caching_skips_empty_content_trailing_message():
+    """REGRESSION: an empty-content trailing user/tool message (e.g. a tool
+    observation that produced no text/image) must not crash prompt caching.
+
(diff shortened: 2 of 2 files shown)


enyst (commented):

**Taste Rating: 🟢 Good taste** — a one-condition guard that eliminates the special case instead of wrapping it in a `try/except IndexError`, plus a regression test that actually regresses. Nothing to fix.

I don't take "REGRESSION:" in a docstring on faith, so I reverted `llm.py` to `main` and ran your test against the unfixed code:

```
--- PR test against UNFIXED llm.py ---
openhands-sdk/openhands/sdk/llm/llm.py:2628: in _apply_prompt_caching
    message.content[
E   IndexError: list index out of range
FAILED ...::test_apply_prompt_caching_skips_empty_content_trailing_message

--- with the fix ---
tests/sdk/llm/test_prompt_caching_cross_conversation.py .......
7 passed in 0.03s
```

That is the whole job: fails before, passes after, for the stated reason. The `assert messages[1].content[-1].cache_prompt is True` is the part that earns its keep — it pins the *fall-through* behavior, not just the absence of a crash, so a future "fix" that swallows the error with `try/except` and marks nothing would still fail this test. The `break` on L2590 means the fall-through is genuinely to the previous non-empty message rather than to all of them.

None worth your time. The comment is four lines for a two-token change, which is more than I'd usually tolerate, but it explains a non-obvious invariant (*why* the empty case exists and *what* the fallback is) rather than restating `and message.content`, so it stays.

**[TESTING GAPS]**

None.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 **LOW** — strictly narrows when a line executes, on a path that previously raised. There is no caller that was depending on `IndexError`.

**KEY INSIGHT:** The bug was an unstated assumption that a `user`/`tool` message always has at least one content block; the fix makes the assumption explicit and cheap to check, which is strictly better than making it true everywhere else.

JiataiWang:
@enyst, when you have a moment, could you take a final look at this one? Your review found no code or test gaps and rated it low risk / worth merging. The substantive fork workflows are still awaiting maintainer approval; no changes are pending from my side. Thanks!

PR #4166, opened 2026-07-21
fix(conversation): persist secrets added via update_secrets

JiataiWang:
HUMAN:

update_secrets was mutating the registry in place and never persisting it, so secrets (including constructor-seeded ones) were lost on restart. I reproduced the missing write and added tests for both the update and constructor paths.

---

AGENT:

## Why

`LocalConversation.update_secrets` mutated `secret_registry` in place. Because that touches a nested object rather than reassigning a `ConversationState` field, it never went through `ConversationState.__setattr__` and so never triggered the base_state autosave. Secrets added this way — including those seeded via the constructor's `secrets=` argument, which routes through this same method — were held only in memory and silently dropped on restart unless some later, unrelated field change happened to persist the state.

## Summary

- Build a modified copy of the registry and reassign it inside `with self._state:` so the autosave fires exactly once.
- Add regression tests for both the `update_secrets` and constructor seeding paths.

## Issue Number

## How to Test

Reproduced before/after:

```python
conv = Conversation(agent=agent, workspace=d, persistence_dir=d, secrets={"SEED": "s"})
conv.update_secrets({"MY_TOKEN": "tok"})
persisted = conv.state._fs.read(BASE_STATE)
# BEFORE: "MY_TOKEN" not in persisted (and "SEED" not persisted either)
# AFTER: both keys present in base_state.json immediately
```
Tests: `.venv/bin/python -m pytest tests/sdk/conversation/test_conversation_secrets_constructor.py -q` — 12 passed; related secret/conversation suites (31 tests) also pass. `ruff` clean.

## Type

- [x] Bug fix

## Notes

Found via code review; no pre-existing issue. Most impactful with a cipher configured to preserve secrets across restarts.

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -2582,11 +2582,8 @@ class LocalConversation(BaseConversation):
                      when a command references the secret key.
         """
         # Reassign a modified copy so ConversationState.__setattr__ fires the
-        # autosave. An in-place `secret_registry.update_secrets(...)` mutates a
-        # nested object without touching any state *field*, so it never triggers
-        # persistence: secrets added here (or seeded through the constructor,
-        # which also routes through this method) would be dropped on restart
-        # unless some unrelated later field change happened to save the state.
+        # autosave; an in-place registry mutation never touches a state field
+        # and so never persists.
         with self._state:
             registry = self._state.secret_registry.model_copy(deep=True)
             registry.update_secrets(secrets)
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -2581,13 +2581,11 @@ class LocalConversation(BaseConversation):
                      SecretValue = str | Callable[[], str]. Callables are invoked lazily
                      when a command references the secret key.
         """
-        # Reassign a modified copy so ConversationState.__setattr__ fires the
-        # autosave; an in-place registry mutation never touches a state field
-        # and so never persists.
+        # Long-lived consumers bind methods on this registry, so preserve its
+        # identity and explicitly schedule autosave for the nested mutation.
         with self._state:
-            registry = self._state.secret_registry.model_copy(deep=True)
-            registry.update_secrets(secrets)
-            self._state.secret_registry = registry
+            self._state.secret_registry.update_secrets(secrets)
+            self._state._mark_dirty()
         logger.info(f"Added {len(secrets)} secrets to conversation")
 
     def set_security_analyzer(self, analyzer: SecurityAnalyzerBase | None) -> None:
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -2581,8 +2581,16 @@ class LocalConversation(BaseConversation):
                      SecretValue = str | Callable[[], str]. Callables are invoked lazily
                      when a command references the secret key.
         """
-        secret_registry = self._state.secret_registry
-        secret_registry.update_secrets(secrets)
+        # Reassign a modified copy so ConversationState.__setattr__ fires the
+        # autosave. An in-place `secret_registry.update_secrets(...)` mutates a
+        # nested object without touching any state *field*, so it never triggers
+        # persistence: secrets added here (or seeded through the constructor,
+        # which also routes through this method) would be dropped on restart
+        # unless some unrelated later field change happened to save the state.
+        with self._state:
+            registry = self._state.secret_registry.model_copy(deep=True)
+            registry.update_secrets(secrets)
+            self._state.secret_registry = registry
         logger.info(f"Added {len(secrets)} secrets to conversation")
 
     def set_security_analyzer(self, analyzer: SecurityAnalyzerBase | None) -> None:
--- a/openhands-sdk/openhands/sdk/conversation/state.py
+++ b/openhands-sdk/openhands/sdk/conversation/state.py
@@ -417,6 +417,16 @@ class ConversationState(OpenHandsModel):
         self._write_guard = write_guard
         self._events.set_write_guard(write_guard)
 
+    def _mark_dirty(self) -> None:
+        """Schedule a save after mutating a nested field in place.
+
+        The caller must hold this state as a context manager so the mutation
+        and the deferred save remain under the same lock.
+        """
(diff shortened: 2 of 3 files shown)


enyst (commented):

**Taste Rating: 🟡 Acceptable** — the diagnosis is exactly right and the fix is the correct shape. The test claims more than it proves, and the gap matters because it's the difference between "your secret is saved" and "the *name* of your secret is saved."

The underlying bug is real and well-described: mutating a nested object never touches a state *field*, so `ConversationState.__setattr__` never fires and autosave never happens. Reassigning a modified copy is the right fix, and doing it inside `with self._state:` is the right scope.

**[CRITICAL ISSUES]**

- [`tests/sdk/conversation/test_conversation_secrets_constructor.py`, L175-177] **The Assertion Doesn't Test What The Docstring Promises**: `assert "MY_TOKEN" in persisted` greps the raw JSON for the *key name*. That string is present even though the value on disk is a redaction placeholder. Here's the actual end-to-end round trip on your branch:

  ```
  BEFORE restart -> 'super-secret-value-12345'
  AFTER  restart -> None
  ```

  And the SDK says so itself while saving:

  ```
  WARNING openhands.sdk.conversation.state: Saving conversation state without cipher -
  1 secret(s) will be redacted and lost on restore. Consider providing a cipher to preserve secrets.
  ```

  So without a cipher, this PR moves the outcome from "nothing persisted" to "an entry persisted with `value=None`". That is a genuine improvement — the registry structure now survives, and with a cipher configured the value should too — but a reader of that test will conclude secrets round-trip, and in the default no-cipher path they do not.

  Fix the test to assert what you actually achieved: reopen the conversation and assert on `secret_registry.secret_sources`, ideally once with a cipher (value preserved) and once without (entry present, value `None`). That test would also document the cipher requirement, which is currently only discoverable by reading a log line at WARNING.

- [`local_conversation.py`, L2507-2515] **Comment Length**: eight lines of comment for four lines of code. Most of it is change-narration — "an in-place `update_secrets(...)` mutates a nested object without touching any state *field*" describes what the code *used to do*, which is what the commit message and `git blame` are for. The load-bearing sentence is the first one: reassign so `__setattr__` fires the autosave. Keep that; the rest belongs in the PR body.

**[TESTING GAPS]**

- `uv run pytest tests/sdk/conversation/test_conversation_secrets_constructor.py` → **12 passed**. Both new tests pass, but per the above they'd also pass if the value were silently dropped, which is precisely the failure mode you're fixing.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟡 **MEDIUM** — adds a deep copy plus a state write to every `update_secrets` call, including the constructor path, so every conversation now writes `base_state.json` at least once more during startup. Correctness-wise that's the point; just be aware it's on the hot construction path. `model_copy(deep=True)` is safe for what the registry actually holds today.

  Unrelated to your change but worth knowing while you're in here: the docstring you're sitting under advertises `SecretValue = str | Callable[[], str]`, and passing a bare `lambda: "computed"` raises `ValueError: Invalid SecretValue`. I confirmed that's **pre-existing** — it fails identically with your fix neutralized — so it's not yours to fix, but the docstring is lying and you're the last person to touch it.

**VERDICT:** ✅ **Worth merging** — the fix is correct. Please strengthen the test to assert post-restart values rather than substring presence, so the next person doesn't inherit false confidence.

**KEY INSIGHT:** `assert "KEY" in json_blob` is the testing equivalent of checking that the envelope arrived without opening it — the bug you're chasing lives entirely in the contents.

JiataiWang:
Fair point about the envelope vs. the contents — the substring assertion did prove much less than the docstring promised.

Pushed b94cb027: the tests now do the real round trip you described. `test_update_secrets_registry_survives_restart_without_cipher` reopens the conversation and asserts the registry entry survives with `get_value() is None` (documenting the no-cipher redaction explicitly), and `test_update_secrets_value_round_trips_with_cipher` asserts the value itself survives when a `Cipher` is configured. The constructor-seeded test got the same restart treatment. Also trimmed the update_secrets comment down to the load-bearing sentence.

Noted on the pre-existing `SecretValue` callable docstring mismatch — I'll leave that out of this PR but can file it separately.

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py`, line 2511] **Secret-source identity and compatibility regression**: `model_copy(deep=True)` recursively clones every existing `SecretSource` just to make the registry assignment observable. `SecretSource` is an extensible public abstraction and the repository already uses stateful custom implementations; cloning one silently forks counters/caches and breaks identity with the source object the caller supplied. A perfectly valid implementation with a private `threading.Lock` cannot be deep-copied at all (`TypeError: cannot pickle '_thread.lock' object`), so a later `update_secrets()` would now fail before adding anything. Copy the registry shell and its `secret_sources` mapping while preserving each source object, then update the copied mapping and assign the registry. Add a regression test that updates a registry containing a stateful or lock-backed custom `SecretSource` and verifies both successful persistence and source identity/state continuity.

**[RISK ASSESSMENT]**
This touches persisted secrets and fixes a real loss-on-restart bug, but the deep copy introduces a new failure mode for supported custom secret providers and can alter their runtime state semantics.

**Recommendation**: Do not auto-merge. Request a human reviewer to validate the copy/ownership model for custom `SecretSource` instances and the persistence round trip.

**VERDICT:**
❌ **Needs rework**: Trigger persistence without recursively cloning caller-owned secret providers.

**KEY INSIGHT:**
The registry mapping is state-owned, but the dynamic secret providers inside it are caller-owned runtime objects whose identity must be preserved.

---

>
>
>

JiataiWang:
Good catch — `deep=True` was the wrong ownership boundary. I replaced it with a shallow registry copy plus a copied `secret_sources` mapping, so reassignment still triggers autosave while existing providers retain identity and runtime state. The new lock-backed regression fails on the old deep-copy path, checks live identity/state continuity, and then reopens the conversation to verify the updated registry persisted. Rebasing onto current main stayed clean; the isolated secrets/state/plugin suites pass 123/123, along with all pre-commit hooks.

JiataiWang:
A final runtime pass exposed one more ownership edge: the shallow registry copy preserved each SecretSource, but it still replaced the registry object after ACP had bound its lifetime masker to that object. A later secret could therefore persist correctly while remaining invisible to the old masking callback. The latest update keeps the registry, its mapping, and existing sources stable, and marks the nested mutation dirty under the ConversationState lock so the existing encrypted, write-guarded autosave still runs once on context exit. The regression now captures the registry, mapping, lookup, and masker before both updates and verifies that all remain live through restart. The full conversation suite passes (770 collected), focused state/plugin/secret checks are green, and all pre-commit hooks pass.

PR #4168, opened 2026-07-21
fix(llm): clamp thinking budget and sibling key in joint-budget clamp

JiataiWang:
HUMAN:

On Bedrock the joint-budget clamp lowered max_tokens but left thinking.budget_tokens above it, producing a request the provider rejects. I reproduced the scenario and added regression tests for the thinking-budget and both-keys cases.

---

AGENT:

## Why

`_clamp_max_tokens_for_joint_budget` lowers the per-request output budget for joint-window providers (Bedrock-Anthropic) when the input is large, but only rewrote the single budget key it found. On the extended-thinking path this caused two follow-on defects: (1) `thinking.budget_tokens` was left just under the full 64k output, so once `max_tokens` is clamped below it, `budget_tokens >= max_tokens`, which Bedrock/Anthropic reject; (2) chat_options sets `max_tokens` without popping the default `max_completion_tokens`, so a stale full-size sibling key could defeat the clamp downstream.

## Summary

- Re-apply the `budget_tokens < max_tokens` invariant against the clamped value.
- Clamp the sibling budget key too (only when already present, so no new key is introduced).
- Rebuild (not mutate) the nested `thinking` dict so caller kwargs are untouched.
- Add regression tests.

## Issue Number

## How to Test

Reproduced against the clamp with a Bedrock thinking request (200k window, 180k input):

```
call_kwargs = {"max_tokens": 64000, "max_completion_tokens": 64000,
               "thinking": {"type": "enabled", "budget_tokens": 63999}}
# AFTER: max_tokens=19744, max_completion_tokens=19744, thinking.budget_tokens=19743  (< max_tokens)
# BEFORE: budget_tokens stayed 63999 >= clamped max_tokens 19744 -> provider rejects
```
Tests: `.venv/bin/python -m pytest tests/sdk/llm/test_joint_budget_clamp.py -q` — 12 passed. `ruff` clean.

## Type

- [x] Bug fix

## Notes

Found via code review; no pre-existing issue. Triggers on Bedrock Claude thinking models with large inputs.

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -2342,7 +2342,34 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
                 input_tokens,
                 context_window,
             )
-        return {**call_kwargs, budget_key: clamped}
+
+        updated = {**call_kwargs, budget_key: clamped}
+
+        # The extended-thinking path in chat_options sets ``max_tokens`` without
+        # popping the default ``max_completion_tokens``, so both budget keys can
+        # be present. Clamp the sibling too (only when already present) so a stale
+        # full-size copy can't defeat the clamp downstream.
+        sibling_key = (
+            "max_completion_tokens" if budget_key == "max_tokens" else "max_tokens"
+        )
+        sibling = updated.get(sibling_key)
+        if isinstance(sibling, int) and sibling > clamped:
+            updated[sibling_key] = clamped
+
+        # Preserve the extended-thinking invariant ``budget_tokens < max_tokens``
+        # (Anthropic/Bedrock reject a thinking budget that meets or exceeds
+        # max_tokens). chat_options applied this against the original
+        # max_output_tokens; re-apply it against the clamped budget so lowering
+        # max_tokens for a large input can't invert it into an invalid request.
+        thinking = updated.get("thinking")
+        if (
+            isinstance(thinking, dict)
+            and isinstance(thinking.get("budget_tokens"), int)
+            and thinking["budget_tokens"] >= clamped
+        ):
+            updated["thinking"] = {**thinking, "budget_tokens": max(clamped - 1, 1)}
+
+        return updated
 
     def _init_model_info_and_caps(self) -> None:
         self._model_info = get_litellm_model_info(
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -160,6 +160,10 @@ DEFAULT_MAX_OUTPUT_TOKENS_CAP: Final[int] = 16384
 # default change that made this manifest.
 JOINT_BUDGET_SAFETY_MARGIN_TOKENS: Final[int] = 256
 JOINT_BUDGET_MIN_OUTPUT_TOKENS: Final[int] = 1024
+# Anthropic rejects ``thinking.budget_tokens`` below this minimum, so a clamped
+# ``max_tokens`` that cannot fit ``budget_tokens >= 1024`` while keeping
+# ``budget_tokens < max_tokens`` leaves no legal thinking budget at all.
+ANTHROPIC_MIN_THINKING_BUDGET_TOKENS: Final[int] = 1024
 # Provider name prefixes known to enforce a joint input/output token budget.
 # Kept as a narrow allowlist so direct providers (Anthropic, OpenAI, etc.) --
 # which have independent input/output windows -- are not affected. We match by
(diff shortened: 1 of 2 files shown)


enyst (commented):

**Taste Rating: 🟡 Acceptable** — the diagnosis is right, I verified the premise, and the tests are unusually thoughtful. There's one arithmetic edge that swaps the error you're fixing for a different one.

**Premise verified.** You claim `chat_options` sets `max_tokens` without popping the default `max_completion_tokens`. It does — L30 sets `max_completion_tokens`, and the extended-thinking branch at L85 adds `max_tokens` on top without removing it:

```python
30:        "max_completion_tokens": max_output_tokens,
...
85:            out["max_tokens"] = max_output_tokens
```

So both keys really are live on that path, and clamping only one really does leave a stale full-size copy. Good catch — that's not obvious from either file alone.

**[CRITICAL ISSUES]**

- [`openhands-sdk/openhands/sdk/llm/llm.py`, L2371] **Off-By-One Into A Different Invalid Request**: `max(clamped - 1, 1)` assumes any positive thinking budget is legal. Anthropic's minimum is **1024**. And `clamped` has a floor of exactly `JOINT_BUDGET_MIN_OUTPUT_TOKENS = 1024` (L2318), so the worst case lands precisely one token short:

  ```
  # input 199,500 tokens against a 200k window — almost no headroom
  clamped max_tokens      : 1024
  thinking.budget_tokens  : 1023
  Anthropic minimum is 1024 -> valid? False
  ```

  Before your change that request failed with "budget_tokens must be less than max_tokens". After it, the same request fails with "budget_tokens must be >= 1024". Still a 400, just a different one — and now with a number the user can't trace back to anything they configured.

  The floor of `1` makes it explicit: there is no such thing as a valid 1-token thinking budget, so if the arithmetic can ever produce one, the arithmetic is wrong. When `clamped` is at or near the minimum there is no legal thinking budget at all, and the honest options are to **drop the `thinking` block entirely** (degrade to non-thinking for this call) or to raise `clamped` so both constraints fit. Silently emitting 1023 picks neither.

- [L2359-2371] **Duplicated Invariant**: `budget_tokens < max_tokens` is now enforced in two places with two different spellings — `min(llm.extended_thinking_budget, max_output_tokens - 1)` in `chat_options.py:69-72`, and `max(clamped - 1, 1)` here. They will drift. One helper (`_fit_thinking_budget(max_tokens) -> int | None`, returning `None` when nothing fits) used by both call sites would hold the rule, the 1024 minimum, and the degrade decision in exactly one place.

**[TESTING GAPS]**

- `uv run pytest tests/sdk/llm/test_joint_budget_clamp.py` → **12 passed**. Both new tests are real: they call the actual method, assert on returned values, and `test_thinking_budget_preserved_when_no_clamp_needed` pins the negative case so the clamp can't start firing unconditionally. The `assert call_kwargs["thinking"]["budget_tokens"] == 63_999` line — checking you didn't mutate the caller's nested dict — is the kind of assertion most people don't think to write. Credit.
- Missing: the tight-headroom case above. Your two tests both use comfortable headroom, which is exactly why the 1023 case slipped through. Add one at `token_counter=199_500` and decide what it should assert.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟡 **MEDIUM** — this sits on the request-construction path for every Anthropic/Bedrock extended-thinking call. It strictly improves the common case (large-but-not-extreme inputs now produce valid requests where they previously didn't) and regresses only the near-exhausted-context case, which was already failing. No API or dependency changes.

**VERDICT:** ✅ **Worth merging** once the near-minimum case is handled — even just dropping the `thinking` key when `clamped <= 1024` would close it, and it's a two-line change.

**KEY INSIGHT:** You clamped a value against one constraint while a second constraint on the same value lives in another file — which is why the fix and the bug are now both one token wide.

JiataiWang:
Thanks for the sharp catch on the arithmetic edge — you're right that `max(clamped - 1, 1)` just traded one 400 for another at the floor.

Pushed 0ca2cda3: when the clamped budget can't fit `1024 <= budget_tokens < max_tokens`, the `thinking` block is now dropped entirely (degrade to non-thinking, with a warning) instead of emitting an illegal 1023. Added the tight-headroom regression test you suggested (`token_counter=199_500` → thinking dropped, caller kwargs untouched) plus the boundary case where `clamped = 1025` still fits the minimum legal budget of exactly 1024.

On the duplicated-invariant point (`chat_options.py` vs the clamp): agreed a shared `_fit_thinking_budget()` helper would be the right home for the rule — happy to do that as a follow-up PR so this one stays focused on the correctness fix.

enyst (commented):

🟢 **Good taste**

The fix restores all three request invariants at the point where the output budget is actually clamped: both budget keys agree, the thinking budget remains strictly below `max_tokens`, and an impossible sub-1024 thinking budget is removed rather than sent to the provider. The boundary tests cover the exact failure modes without mutating caller-owned kwargs.

**Repository policy note:** This changes LLM request budgeting and can plausibly affect agent/evaluation behavior. No completed eval-monitor run with human confirmation is linked, so a human maintainer should make the merge decision after lightweight eval validation.

**[RISK ASSESSMENT]**
The patch is narrow and well tested, but it changes core LLM request construction for large-context Bedrock/Anthropic calls. Review should focus on provider compatibility and the deliberate fallback from extended thinking to non-thinking when no legal budget fits.

**VERDICT:**
✅ **Worth merging**: The clamp logic is coherent and the regression boundaries are convincing, subject to the repository's human eval check.

**KEY INSIGHT:**
Budget invariants belong after dynamic clamping, because setup-time validation cannot account for per-request input headroom.

PR #4172, opened 2026-07-21
feat(agent-server): sanitized product-analytics telemetry with split consent policy

VascoSch92:
HUMAN:

Add telemetry in agent-server.

---

AGENT:

## Why

We need server-side visibility into local Agent Canvas execution and errors. The repo has no PostHog client, dependency, or event emitter today, and the two existing signals are unsuitable for product analytics: `LLM.log_completions` writes full prompts, responses and tracebacks to disk, and Laminar/OpenTelemetry produces distributed traces. Neither may be forwarded wholesale.

This adds a versioned, allowlisted diagnostic-event contract owned by `openhands-agent-server`, behind an injectable sink that defaults to a no-op, with the PostHog exporter enabled only at the composition boundary.

**Design note:** the original issue framed consent plumbing as a design blocker, and this PR first resolved it with a typed settings field plus a deployment "mode". The OSS-5715 follow-up thread overruled that. Consent now lives in `misc_settings.telemetry.consent`, there is no cloud-mode concept, and writes go through the existing `/api/settings` endpoint. The event schema, sanitizer, sink and subscriber layers were compatible with either design and are unchanged.

## Summary

- **Consent lives in `misc_settings.telemetry.consent`**, written through the existing `PATCH /api/settings`. No settings schema change, no dedicated endpoint, and it persists across restarts for free. `misc_settings.telemetry` is documented as the single sanctioned exception to the opaque-container rule. Precedence: `DO_NOT_TRACK` → `OH_TELEMETRY_CONSENT` in `override` mode → persisted value → legacy Canvas key (read-only fallback) → `OH_TELEMETRY_CONSENT` as a seed → `unset`.
- **No cloud mode.** Nothing in agent-server special-cases a hosted deployment; enforcement is seeding the sandbox's misc settings at runtime init, which lives outside this repo.
- **Exporters are selectable** — `OH_TELEMETRY_EXPORTER` = `none` (default) | `posthog` | `http`. The new generic HTTP exporter POSTs sanitized batches to a configured endpoint, so no vendor credentials need to live in the sandbox.
- **Privacy enforced by the type system.** Every property is a constrained scalar — never a bare `str` — so a prompt or traceback is a construction-time `ValidationError`. The sanitizer never reads `str(exc)` nor calls into `traceback`. Ids are keyed digests, magnitudes bucketed, and the exporter only calls `capture()` so it cannot mint or merge an identity.
- **Delivery cannot block or fail a conversation.** `emit()` is synchronous, because `PubSub.__call__` awaits its subscribers. Events go onto a `deque(maxlen=…)` drained by one retained task; failed batches are dropped, not re-queued; shutdown is hard-capped.

## Issue Number

No issue number was supplied with this work item. Happy to link one.

## Acceptance criteria

Every item from the issue is addressed:

| Requirement | Where |
|---|---|
| Keep `openhands-sdk` vendor-neutral | No PostHog dep or vendor API; the only SDK change is a docstring on `user_id` |
| Sanitized diagnostic-event contract in agent-server | `telemetry/models.py` |
| Injectable no-op sink + optional PostHog exporter at the composition boundary | `telemetry/sink.py`, `telemetry/posthog_exporter.py`, wired in `api.py` / `init_router.py` |
| Sanitize at event construction, before any exporter | Pydantic validation on construction; sinks accept only `DiagnosticEvent`, never a dict |
| Consent explicitly resolved, including revocation | `telemetry/policy.py`, full truth table tested |
| Versioned schema over minimal lifecycle + failure events | `TELEMETRY_SCHEMA_VERSION`; 7 events |
| Properties allowlisted; no prompts, messages, source, secrets, full paths, bodies, tracebacks | `EXPECTED_PROPERTY_NAMES` + deny test; leak audit below |
| Core SDK has no PostHog dependency or vendor API | Asserted in a subprocess test |
| Exporter optional, disabled by default for library/headless | `[project.optional-dependencies] posthog`; default exporter `none` |
| Consent states explicitly represented and tested | `test_telemetry_policy.py`, `test_telemetry_settings_path.py` |
| No telemetry before consent; revocation stops delivery and discards queue | Nothing is buffered while disabled; the settings path re-resolves and clears the queue before returning |
| Delivery async, bounded, cannot block or fail a conversation | Sync `emit()` + bounded deque + retained drain task |
| Release/version, coarse runtime props, normalized error class/fingerprint, privacy-safe correlation ids | `RuntimeProperties`, `normalize_exception`, keyed `conversation_ref` |
| Canvas ↔ agent-server correlation without duplicate identities | `distinct_id` reuses host `user_id` verbatim; `capture()` only |
| Tests for sanitization, consent transitions, queue limits, exporter failure, disabled behavior | 234 tests in `tests/agent_server/telemetry/` |
| Docs distinguish analytics from completion logging and Laminar/OTel | Comparison table in the agent-server README |

## How to Test

```bash
uv run pytest tests/agent_server/telemetry/ -v     # 234 passed
uv run pytest tests/agent_server/ -q               # 1714 passed
uv run pytest tests/cross/ -q                      # 360 passed, 1 pre-existing failure
uv run pre-commit run --all-files                  # all hooks pass
```

Two pre-existing failures you may hit locally, both of which reproduce unchanged on a clean `main` worktree:

- `tests/cross/test_remote_conversation_live_server.py::test_openai_chat_completions_gateway_over_real_server` needs a real LLM provider.
- Running `tests/agent_server/` and `tests/cross/` **in one process** yields 25 failures from cross-suite state pollution (`test_conversation_restore_behavior`, `test_agent_loading`, `test_agent_secrets_integration`). Verified identical on clean `main`. CI runs them as separate jobs, which is why it is green — run them separately locally too.

### End-to-end against a real server

Unit tests alone are not sufficient, so the three required scenarios were driven against a real `uvicorn` process with a capturing stand-in for the exporter, via `curl`. Re-run against the new consent path after the follow-up rework.

**1. Opted-in produces sanitized events.** Exact captured payload:

```json
{
  "event": "agent_server.request_failed",
  "distinct_id": "anon:855e4331473d401587ab806f547d892f",
  "properties": {
    "schema_version": 1, "server_version": "1.36.1", "python_version": "3.13",
    "platform": "darwin", "deferred_init": false,
    "source": "openhands-agent-server",
    "route_template": "/api/_boom", "method": "GET", "status_code": 500,
    "error_class": "ValueError", "error_category": "config",
    "error_fingerprint": "445aa04176d34c95", "error_id": "7b474744..."
  }
}
```

**Leak audit.** The route raises `ValueError(f"secret={CANARY} path=/Users/alice/x.py")`; a parameterized route was hit at `/api/_boom/CONV-SENSITIVE-999/detail/TOK-SECRET`:

```
absent   sk-ant-CANARY-DO-NOT-LEAK       absent   CONV-SENSITIVE-999
absent   'secret=' (message fragment)    absent   TOK-SECRET
absent   /Users/alice                    absent   Traceback
route_template: '/api/_boom/{conversation_id}/detail/{secret_token}'
```

**2. Opted-out produces nothing.**

Driven entirely through `PATCH /api/settings` — there is no telemetry endpoint:

```
OPTED OUT : 500 raised                            -> captured events: 0
GRANT     : misc_settings.telemetry.consent=granted
            500 raised x2                         -> captured events: 2
REVOKE    : misc_settings.telemetry.consent=denied
            500 raised x3                         -> before 2, after 2  (discarded)
```

Consent survives a process restart, since `misc_settings` is already persisted.

**3. Exporter failure leaves execution unaffected.** With `ConnectionError` on every send:

```
server still serving      -> HTTP 200
'export failing' warnings -> 1     (per streak, not per failure)
shutdown took 0.5s                 (bounded)
```

**4. Disabled by default.** `create_app(Config())` yields `NoOpTelemetrySink` and never imports the exporter module (subprocess-asserted). The test asserts *our* module stays unimported rather than `"posthog" not in sys.modules`, because the pre-existing `browser-use` dependency imports posthog transitively — unrelated to this change.

## Video/Screenshots

Not applicable — no GUI surface. The transcripts above are the equivalent evidence.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**Follow-up alignment (latest commit).** Reverted from this PR's original design: the typed `PersistedSettings.telemetry_consent` field, the v2→v3 schema bump and its baseline fixture, the `cloud_locked`/`local_opt_in`/`disabled` mode concept, `OH_TELEMETRY_MODE`, the `deployment_mode` event property, and the dedicated consent router.

### ⚠️ Blocking follow-up outside this repo

Cloud gets **no telemetry** until OpenHands/runtime-api seeds the sandbox agent-server with `misc_settings.telemetry.consent = granted` and `managed = true` at runtime init, before any conversation starts. That is the direct consequence of removing the cloud-mode concept: nothing in this repo special-cases Cloud any more, so enforcement has to happen at seeding time.

Similarly out of scope: the backend endpoint the `http` exporter targets — the one that revalidates runtime auth and TOS/consent before forwarding to PostHog — lives in the OpenHands main app. Its payload contract is documented in the agent-server README so it can be built against independently.

**Confirmed during review:** `OH_TELEMETRY_CONSENT_MODE` semantics (`seed` default, `override` opt-in); keeping `DO_NOT_TRACK` and dropping the redundant `OH_TELEMETRY_DISABLED` alias; dual-**read** of the legacy Canvas consent key without dual-write.

**Reviewer attention:** `telemetry/service.py` must call `get_settings_store(config)`, never the no-arg form — that singleton's directory and cipher are fixed by the first call, and telemetry initialises before `ConversationService` primes it. An earlier revision left the process with `cipher=None`. Regression-tested.

**Not covered:** no test drives a full LLM-backed conversation; lifecycle is verified against the real event object and a real `StoredConversation`, not a live model run. `request_failed` / `server_*` always use `anon:` since no user context is available at those points.


openhands-agent-server/openhands/agent_server/telemetry/factory.py
```diff
@@ -0,0 +1,136 @@
+"""Builds :class:`DiagnosticEvent` objects.
+
+Centralising construction here means the runtime envelope, the pseudonym salt
+and the identity rules are decided in exactly one place, so a caller cannot
+accidentally assemble an event with a raw conversation id or an unbucketed
+count.
+"""
+
+from __future__ import annotations
```
VascoSch92: do you need that?

openhands-agent-server/openhands/agent_server/telemetry/factory.py
```diff
@@ -0,0 +1,136 @@
...
+from openhands.agent_server.telemetry.sanitizer import (
+    UNKNOWN_TOKEN,
+    pseudonymize,
+    safe_token,
+    safe_version,
+)
+from openhands.sdk.utils import utc_now
+
+
+ANONYMOUS_PREFIX = "anon:"
```
VascoSch92: use Final from typing

openhands-agent-server/openhands/agent_server/telemetry/models.py
```diff
@@ -0,0 +1,324 @@
...
+The single intentional exception is :attr:`DiagnosticEvent.distinct_id`, which
+carries the caller-supplied ``user_id`` verbatim so events correlate with the
+identity the hosting deployment already owns. It is documented as a
+pass-through and excluded from the "no bare str" rule by design, not by
+oversight.
+
+This module must not import any analytics vendor SDK.
+"""
+
+from __future__ import annotations
```
VascoSch92: do we need that? In general search for other of this kind to see if we need them

openhands-agent-server/openhands/agent_server/telemetry/models.py
```diff
@@ -0,0 +1,324 @@
...
+
+from datetime import datetime
+from typing import Annotated, Literal
+
+from pydantic import BaseModel, ConfigDict, Field, StringConstraints
+
+
+TELEMETRY_SCHEMA_VERSION = 1
```
VascoSch92: typing. In general use typing

openhands-agent-server/openhands/agent_server/telemetry/models.py
```diff
@@ -0,0 +1,324 @@
...
+"""A bucketed magnitude such as ``11-50``. Never a raw count."""
+
+
+SERVER_STARTED = "agent_server.server_started"
+SERVER_STOPPED = "agent_server.server_stopped"
+CONVERSATION_STARTED = "agent_server.conversation_started"
+CONVERSATION_FINISHED = "agent_server.conversation_finished"
+CONVERSATION_FAILED = "agent_server.conversation_failed"
+CONVERSATION_ERROR = "agent_server.conversation_error"
+REQUEST_FAILED = "agent_server.request_failed"
```
VascoSch92: Can this be a StrEnum?

openhands-agent-server/openhands/agent_server/telemetry/policy.py
```diff
@@ -0,0 +1,94 @@
...
+* **Consent** is supplied by the person using a local installation
+  (``PersistedSettings.telemetry_consent``) and only matters in
+  ``local_opt_in``.
+
+Consent is *not* read from ``misc_settings``: that container is documented as
+opaque and never interpreted by the agent-server, and this value is very much
+interpreted by the agent-server.
+"""
+
+from __future__ import annotations
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/policy.py
```diff
@@ -0,0 +1,94 @@
...
+
+import os
+from dataclasses import dataclass
+from typing import Literal
+
+
+TelemetryMode = Literal["cloud_locked", "local_opt_in", "disabled"]
+TelemetryConsent = Literal["granted", "denied", "unset"]
+
+DO_NOT_TRACK_ENV = "DO_NOT_TRACK"
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/policy.py
```diff
@@ -0,0 +1,94 @@
...
+        return TelemetryDecision(enabled=False, reason="mode_disabled")
+
+    if mode == "cloud_locked":
+        return TelemetryDecision(enabled=True, reason="cloud_locked")
+
+    if consent == "granted":
+        return TelemetryDecision(enabled=True, reason="consent_granted")
+    if consent == "denied":
+        return TelemetryDecision(enabled=False, reason="consent_denied")
+    return TelemetryDecision(enabled=False, reason="consent_unset")
```
VascoSch92: use a match sttement

openhands-agent-server/openhands/agent_server/telemetry/posthog_exporter.py
```diff
@@ -0,0 +1,94 @@
...
+2. ``alias()`` is irreversible in PostHog — a single mistake permanently
+   merges two identities.
+3. Person profiles turn cheap events into billed ones, and these are
+   diagnostics rather than per-person product usage.
+
+Correlation is achieved by *reusing* the identity the deployment already
+established, never by asserting a new one.
+"""
+
+from __future__ import annotations
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/posthog_exporter.py
```diff
@@ -0,0 +1,94 @@
...
+import asyncio
+
+from openhands.agent_server.telemetry.factory import ANONYMOUS_PREFIX
+from openhands.agent_server.telemetry.models import DiagnosticEvent
+from openhands.sdk.logger import get_logger
+
+
+logger = get_logger(__name__)
+
+DEFAULT_POSTHOG_HOST = "https://us.i.posthog.com"
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/router.py
```diff
@@ -0,0 +1,136 @@
...
+two reasons:
+
+1. ``SettingsResponse`` / ``SettingsUpdateRequest`` live in ``openhands-sdk``,
+   which stays vendor-neutral and free of telemetry concepts. Routing consent
+   through them would push a telemetry concept into the core SDK.
+2. Consent should be an explicit, auditable act. A deep-merged diff blob is the
+   wrong shape for something that must be *granted*, not accidentally set.
+"""
+
+from __future__ import annotations
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/router.py
```diff
@@ -0,0 +1,136 @@
...
+)
+from openhands.agent_server.telemetry.models import TELEMETRY_SCHEMA_VERSION
+from openhands.agent_server.telemetry.policy import TelemetryMode, resolve
+from openhands.agent_server.telemetry.service import notify_consent_changed
+from openhands.sdk.logger import get_logger
+
+
+logger = get_logger(__name__)
+
+CONSENT_PATH = "/consent"  # -> /api/telemetry/consent
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+   forbidden here because :class:`traceback.FrameSummary` reads the actual
+   *source line* off disk into the object, and ``f_locals`` obviously holds
+   live values.
+
+The fingerprint is a grouping key, not a redaction mechanism: its preimage is
+already PII-free, so the hash is defence in depth rather than the thing being
+relied on.
+"""
+
+from __future__ import annotations
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+from types import TracebackType
+
+from openhands.agent_server.telemetry.models import (
+    ERROR_CATEGORY_BY_CLASS_NAME,
+    TELEMETRY_SCHEMA_VERSION,
+    ErrorCategory,
+)
+
+
+_FIRST_PARTY_ROOT = "openhands"
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+_MAX_FINGERPRINT_FRAMES = 5
+
+_SAFE_TOKEN_RE = re.compile(r"^[a-z0-9][a-z0-9_.:\-]{0,63}$")
+_SAFE_IDENTIFIER_RE = re.compile(
+    r"^[A-Za-z_][A-Za-z0-9_]*([.:][A-Za-z_][A-Za-z0-9_]*)*$"
+)
+_SAFE_IDENTIFIER_MAX_LEN = 96
+_VERSION_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.+\-/]{0,63}$")
+
+UNKNOWN_TOKEN = "unknown"
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+    return bucket(count, _COUNT_BOUNDS)
+
+
+def token_bucket(tokens: int | None) -> str:
+    return bucket(tokens, _TOKEN_BOUNDS)
+
+
+def cost_bucket(cost: float | None) -> str:
+    return bucket(cost, _COST_BOUNDS)
+
```
VascoSch92: do we need these methods? just use the bucket(....)

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+    ("claude", "anthropic"),
+    ("gpt", "openai"),
+    ("o1", "openai"),
+    ("o3", "openai"),
+    ("gemini", "gemini"),
+    ("mistral", "mistral"),
+    ("deepseek", "deepseek"),
+    ("llama", "ollama"),
+    ("grok", "xai"),
+)
```
VascoSch92: can we get these from another place?

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+    lowered = model.strip().lower()
+
+    prefix = lowered.split("/", 1)[0] if "/" in lowered else ""
+    if prefix in MODEL_FAMILIES:
+        return prefix
+
+    for hint, family in _MODEL_FAMILY_HINTS:
+        if hint in lowered:
+            return family
+    return "other"
```
VascoSch92, line 215: use an iterator

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+@dataclass(frozen=True)
```
VascoSch92: slots

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,356 @@
...
+    return safe_identifier(f"{top_level}:{name}", default=name), False
+
+
+def _categorize(exc: BaseException) -> ErrorCategory:
+    """Category from the exception *type* only, walking the MRO."""
+    for klass in type(exc).__mro__:
+        category = ERROR_CATEGORY_BY_CLASS_NAME.get(klass.__name__)
+        if category is not None:
+            return category
+    return "unknown"
```
VascoSch92: iterator

openhands-agent-server/openhands/agent_server/telemetry/sink.py
```diff
@@ -0,0 +1,310 @@
...
+
+* the queue is bounded **on ingest** via ``deque(maxlen=…)`` rather than only
+  on the retry path, so a wedged exporter cannot grow memory without limit;
+* a batch that exhausts its retries is **dropped, not re-queued** — re-queueing
+  converts a downstream outage into unbounded memory plus a queue permanently
+  full of stale events;
+* revocation **discards** the queue instead of draining it.
+"""
+
+from __future__ import annotations
```
VascoSch92: as above

openhands-agent-server/openhands/agent_server/telemetry/sink.py
```diff
@@ -0,0 +1,310 @@
...
+    TelemetryMode,
+    resolve,
+)
+from openhands.sdk.logger import get_logger
+
+
+logger = get_logger(__name__)
+
+#: Consent cache TTL. Revocation does not wait for it; the endpoint pushes.
+CONSENT_REFRESH_SECONDS = 5.0
```
VascoSch92: as above

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
```diff
@@ -0,0 +1,275 @@
...
+from openhands.sdk.event.conversation_error import ConversationErrorEvent
+from openhands.sdk.logger import get_logger
+from openhands.sdk.utils import utc_now
+
+
+logger = get_logger(__name__)
+
+_TERMINAL_STATUSES = frozenset({"finished", "error", "stuck"})
+_FAILURE_STATUSES = frozenset({"error", "stuck"})
+_EXECUTION_STATUS_KEY = "execution_status"
```
VascoSch92: as above

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
```diff
@@ -0,0 +1,275 @@
...
+  awaits an initial state push and only catches ``TimeoutError``, so an
+  exception escaping this subscriber during registration would fail
+  conversation *startup*.
+* **Non-blocking.** ``__call__`` must be ``async def`` to satisfy the
+  ``Subscriber`` ABC, but its body contains **zero ``await``** — it terminates
+  in the synchronous ``sink.emit``. ``PubSub.__call__`` awaits its subscribers,
+  so anything slower would stall event fan-out for the whole conversation.
+"""
+
+from __future__ import annotations
```
VascoSch92: as above

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
```diff
@@ -0,0 +1,275 @@
...
+
+        ``full_state`` carries ``stats.usage_to_metrics``, a map of usage id to
+        a ``MetricsSnapshot``. Summing it gives the conversation totals, which
+        are then bucketed — the raw figures are never reported. Best-effort:
+        any shape change degrades to ``unknown`` rather than raising.
+        """
+        try:
+            value = getattr(event, "value", None)
+            if not isinstance(value, dict):
+                return
```
VascoSch92: return None

openhands-agent-server/openhands/agent_server/telemetry/factory.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/models.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/policy.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/posthog_exporter.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/router.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/sink.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
VascoSch92: Removed — the package targets 3.12+ and had no forward references that needed it. Dropped from all nine telemetry modules.

openhands-agent-server/openhands/agent_server/telemetry/factory.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/models.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/policy.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/posthog_exporter.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/router.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/sink.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
VascoSch92: Done — now `typing.Final`. Applied to every module constant in the package.

openhands-agent-server/openhands/agent_server/telemetry/models.py
VascoSch92: Yes — now a `StrEnum`:

```python
class EventName(StrEnum):
    SERVER_STARTED = "agent_server.server_started"
    ...
```

This also removes a real risk: the seven constants had a parallel `EventNameLiteral` alias repeating the same strings, so the two could drift. `DiagnosticEvent.event_name` is now typed `EventName`, and the wire value is unchanged since `StrEnum` serialises to its member value.

openhands-agent-server/openhands/agent_server/telemetry/policy.py
VascoSch92: Done — `resolve()` now uses match statements for both the mode and the consent dispatch. The kill-switch check stays a guard clause above them, since it must short-circuit before any mode is considered.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Removed. The four wrappers are gone and callers pass the bounds directly, e.g. `bucket(elapsed, DURATION_BOUNDS)`. The bounds constants are now public (`DURATION_BOUNDS`, `COUNT_BOUNDS`, `TOKEN_BOUNDS`, `COST_BOUNDS`) instead of underscore-private.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Good catch — yes. Now sourced from `litellm.provider_list` (133 providers) via a cached helper, instead of a hand-maintained set that would drift:

```python
@lru_cache(maxsize=1)
def _known_providers() -> frozenset[str]:
    from litellm import provider_list
    return frozenset(str(getattr(p, "value", p)).lower() for p in provider_list)
```

This also made the output more accurate: `litellm_proxy/my-internal-deployment` now reports `litellm_proxy` rather than collapsing to `other` — still without the deployment name. I updated the test that asserted the old behaviour. Verified the deployment name and custom hostnames are still stripped.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Done — `_walk_frames` is now a generator yielding `(module, lineno)` instead of building a list.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Added `slots=True`. Also applied to `TelemetryDecision`, `ConversationTelemetryContext` and `TelemetrySubscriber`.

Note this has a test-visible consequence: `slots` removes `__dict__`, and two leak-audit tests were introspecting `fingerprint.__dict__`. They now use `dataclasses.asdict()`, which is the more correct assertion anyway.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Done — `_categorize` now builds a generator over the MRO and consumes it.

One caveat worth flagging: I first wrote it as `next((c for c in candidates if c is not None), _UNKNOWN_CATEGORY)`, but pyright widens `next()` over a generator expression to `str` and loses the `Literal` type, failing the type check. Confirmed with `reveal_type` → `Unknown | str`. It is consumed with a `for` loop over the generator instead, with a comment recording why.

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
VascoSch92: Done — explicit `return None` on all three early-exit branches in `_capture_usage`.

VascoSch92:
Audited the package against the rest of the repo for duplication and hardcoded data. Found six real duplications — pushed in `f2e16c25b`.

### Duplicated definitions, now single-sourced

| Was | Now |
|---|---|
| `TelemetryMode` declared **3×** — `config.py`, `telemetry/policy.py`, and inline on `RuntimeProperties` | `config.py` is canonical; the other two import it |
| `TelemetryConsent` declared **2×** — `persistence/models.py` and `telemetry/policy.py` | `persistence` owns it (it is the persisted value); policy re-exports |
| `_TERMINAL_STATUSES` / `_FAILURE_STATUSES` hardcoded as `{"finished","error","stuck"}` | Derived from `ConversationExecutionStatus.is_terminal()` |
| `_FULL_STATE_KEY = "full_state"` | Imported from the SDK, which already exports `FULL_STATE_KEY` |
| `_package_version` — my docstring literally said *"Mirror of `server_details_router._package_version`"* | Removed; versions and build metadata now come from `ServerInfo` |
| `OPENHANDS_BUILD_GIT_SHA/_REF` env reads duplicating `ServerInfo` | Same — read once, via `ServerInfo` |

The `ServerInfo` change has a side benefit: `/server_info` and telemetry can no longer disagree about what is running.

### Hardcoded values

- `_FIRST_PARTY_ROOT` is now derived from the module’s own package rather than the literal `"openhands"`.
- Provider names already come from `litellm.provider_list` (previous round).
- **The error-category table stays hardcoded, deliberately.** Mapping an exception class to an analytics category is a policy decision — no library exposes it. But I verified all 36 entries against their sources and locked that in: 11 exist in `litellm.exceptions`, 12 in `builtins`, 3 in `httpx`; the rest come from git/mcp/asyncio/pydantic. A rename upstream now fails a test instead of silently degrading the error to `unknown`.
- Bucket bounds stay constants — changing them changes what the analytics *mean*, so they belong with the schema, not in config.

### New test

`test_telemetry_no_duplication.py` (24 tests) fails if a duplicate definition reappears, if the derived status sets drift from the enum, if telemetry versions diverge from `ServerInfo`, or if a hardcoded exception name stops existing upstream.

1672 agent-server tests pass (+24), pre-commit green.

openhands-agent-server/openhands/agent_server/telemetry/factory.py
```diff
@@ -0,0 +1,125 @@
...
+from openhands.agent_server.telemetry.sanitizer import (
+    UNKNOWN_TOKEN,
+    pseudonymize,
+    safe_token,
+    safe_version,
+)
+from openhands.sdk.utils import utc_now
+
+
+ANONYMOUS_PREFIX: Final = "anon:"
```
VascoSch92: use FInal str here

openhands-agent-server/openhands/agent_server/telemetry/http_exporter.py
```diff
@@ -0,0 +1,81 @@
...
+from openhands.agent_server.telemetry.models import (
+    TELEMETRY_SCHEMA_VERSION,
+    DiagnosticEvent,
+)
+from openhands.sdk.logger import get_logger
+
+
+logger = get_logger(__name__)
+
+DEFAULT_TIMEOUT_SECONDS = 15.0
```
VascoSch92: use Final int here

openhands-agent-server/openhands/agent_server/telemetry/models.py
```diff
@@ -0,0 +1,315 @@
...
+"""
+
+from datetime import datetime
+from enum import StrEnum
+from typing import Annotated, Final, Literal
+
+from pydantic import BaseModel, ConfigDict, Field, StringConstraints
+
+
+TELEMETRY_SCHEMA_VERSION: Final = 1
```
VascoSch92: final int here

openhands-agent-server/openhands/agent_server/telemetry/policy.py
```diff
@@ -0,0 +1,199 @@
...
+MISC_MANAGED_KEY: Final = "managed"
+
+#: Pre-namespace key Canvas used. Read as a fallback so an existing user is not
+#: silently reset to no-consent; never written back.
+LEGACY_MISC_NAMESPACE: Final = "app_preferences"
+LEGACY_CONSENT_KEY: Final = "user_consents_to_analytics"
+
+DO_NOT_TRACK_ENV: Final = "DO_NOT_TRACK"
+CONSENT_ENV: Final = "OH_TELEMETRY_CONSENT"
+CONSENT_MODE_ENV: Final = "OH_TELEMETRY_CONSENT_MODE"
```
VascoSch92: Can we use StrEnum here?

openhands-agent-server/openhands/agent_server/telemetry/policy.py
```diff
@@ -0,0 +1,199 @@
...
+            reason="env_seed",
+            managed=managed,
+        )
+
+    return TelemetryDecision(
+        consent="unset", enabled=False, reason="default", managed=managed
+    )
+
+
+__all__ = [
```
VascoSch92: This should be top of the file

openhands-agent-server/openhands/agent_server/telemetry/posthog_exporter.py
```diff
@@ -0,0 +1,93 @@
...
+from typing import Final
+
+from openhands.agent_server.telemetry.factory import ANONYMOUS_PREFIX
+from openhands.agent_server.telemetry.models import DiagnosticEvent
+from openhands.sdk.logger import get_logger
+
+
+logger = get_logger(__name__)
+
+DEFAULT_POSTHOG_HOST: Final = "https://us.i.posthog.com"
```
VascoSch92: Final str here

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,333 @@
...
+_SAFE_TOKEN_RE: Final = re.compile(r"^[a-z0-9][a-z0-9_.:\-]{0,63}$")
+_SAFE_IDENTIFIER_RE: Final = re.compile(
+    r"^[A-Za-z_][A-Za-z0-9_]*([.:][A-Za-z_][A-Za-z0-9_]*)*$"
+)
+_SAFE_IDENTIFIER_MAX_LEN: Final = 96
+_VERSION_RE: Final = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.+\-/]{0,63}$")
+
+UNKNOWN_TOKEN: Final = "unknown"
+UNKNOWN_ERROR_CLASS: Final = "UnknownError"
+_UNKNOWN_CATEGORY: Final[ErrorCategory] = "unknown"
```
VascoSch92: Final and the type of the varaible

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,333 @@
...
+    Bounds cardinality and guarantees no free-form string escapes.
+    """
+    token = safe_token(value, default=default)
+    return token if token in allowed else default
+
+
+DURATION_BOUNDS: Final[tuple[float, ...]] = (1, 5, 15, 60, 300, 1800)
+COUNT_BOUNDS: Final[tuple[float, ...]] = (1, 5, 20, 100, 500)
+TOKEN_BOUNDS: Final[tuple[float, ...]] = (1_000, 10_000, 50_000, 200_000, 1_000_000)
+COST_BOUNDS: Final[tuple[float, ...]] = (0.01, 0.1, 1, 10, 100)
```
VascoSch92: Can this be in an object Bounds ? make it faster and clear

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
```diff
@@ -0,0 +1,333 @@
...
+
+@lru_cache(maxsize=1)
+def _known_providers() -> frozenset[str]:
+    """Provider names, sourced from litellm rather than hand-maintained."""
+    from litellm import provider_list
+
+    return frozenset(str(getattr(p, "value", p)).lower() for p in provider_list)
+
+
+_MODEL_FAMILY_HINTS: Final[tuple[tuple[str, str], ...]] = (
```
VascoSch92, line 185: do we need to hardcode this here?

openhands-agent-server/openhands/agent_server/telemetry/sink.py
```diff
@@ -0,0 +1,297 @@
...
+logger = get_logger(__name__)
+
+#: Consent cache TTL. Revocation does not wait for it; the endpoint pushes.
+CONSENT_REFRESH_SECONDS: Final = 5.0
+
+#: Caps the final flush so a hung endpoint cannot extend shutdown.
+SHUTDOWN_FLUSH_TIMEOUT_SECONDS: Final = 5.0
+
+_MAX_BACKOFF_SECONDS: Final = 60.0
+
```
VascoSch92: FInal with int

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
```diff
@@ -0,0 +1,281 @@
...
+
+# Derived from the SDK enum rather than hardcoded, so a new terminal status
+# cannot silently stop being reported.
+_TERMINAL_STATUSES: Final = frozenset(
+    s.value for s in ConversationExecutionStatus if s.is_terminal()
+)
+_FAILURE_STATUSES: Final = _TERMINAL_STATUSES - {
+    ConversationExecutionStatus.FINISHED.value
+}
+_EXECUTION_STATUS_KEY: Final = "execution_status"
```
VascoSch92: FInal str

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
```diff
@@ -0,0 +1,281 @@
...
+                if not self._seeded:
+                    # The attach-time push is a baseline, not a transition;
+                    # otherwise every rehydration re-emits a terminal event.
+                    self._seeded = True
+                    if status in _TERMINAL_STATUSES:
+                        self._terminal_emitted = True
+                    return
+
+                if status in _TERMINAL_STATUSES:
+                    self._capture_usage(event)
```
VascoSch92: use a match statement here

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
```diff
@@ -0,0 +1,281 @@
...
+    if key == _EXECUTION_STATUS_KEY and isinstance(value, str):
+        return value
+    if key == FULL_STATE_KEY and isinstance(value, dict):
+        status = value.get(_EXECUTION_STATUS_KEY)
+        if isinstance(status, str):
+            return status
+    return None
+
+
+__all__ = [
```
VascoSch92: top of the file

openhands-agent-server/openhands/agent_server/telemetry/factory.py
VascoSch92: Done — now `Final[str]`. Applied the explicit type to every module constant in the package (`Final[str]`, `Final[int]`, `Final[float]`, `Final[frozenset[str]]`, `Final[re.Pattern[str]]`); a grep for bare `: Final =` now returns nothing.

openhands-agent-server/openhands/agent_server/telemetry/posthog_exporter.py
VascoSch92: Done — now `Final[str]`. Applied the explicit type to every module constant in the package (`Final[str]`, `Final[int]`, `Final[float]`, `Final[frozenset[str]]`, `Final[re.Pattern[str]]`); a grep for bare `: Final =` now returns nothing.

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
VascoSch92: Done — now `Final[str]`. Applied the explicit type to every module constant in the package (`Final[str]`, `Final[int]`, `Final[float]`, `Final[frozenset[str]]`, `Final[re.Pattern[str]]`); a grep for bare `: Final =` now returns nothing.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Done — `sanitizer.py` constants now carry explicit types: `Final[str]` for the tokens, `Final[int]` for the length cap, and `Final[re.Pattern[str]]` for the compiled regexes. Applied package-wide, not just here.

openhands-agent-server/openhands/agent_server/telemetry/http_exporter.py
VascoSch92: Done — now `DEFAULT_TIMEOUT_SECONDS: Final[float] = 15.0`.

Flagging one deviation: I used `float` rather than `int`, because the value is `15.0` and it feeds `httpx.AsyncClient(timeout=...)`, which is float-typed. Happy to change it to `Final[int] = 15` if you would rather — `httpx` accepts either.

openhands-agent-server/openhands/agent_server/telemetry/models.py
VascoSch92: Done — `TELEMETRY_SCHEMA_VERSION: Final[int] = 1`.

openhands-agent-server/openhands/agent_server/telemetry/sink.py
VascoSch92: Done, with the same caveat as the HTTP timeout: these are `Final[float]` rather than `Final[int]` because the values are `5.0` / `60.0` and feed `asyncio.sleep`, and the backoff computes `retry_delay * 2**streak` which stays float. `CONSENT_REFRESH_SECONDS`, `SHUTDOWN_FLUSH_TIMEOUT_SECONDS` and `_MAX_BACKOFF_SECONDS` are all annotated now. Say the word if you want them as ints.

openhands-agent-server/openhands/agent_server/telemetry/policy.py
VascoSch92: Done — both string-valued types are now `StrEnum`, matching `EventName`:

```python
class TelemetryConsent(StrEnum):
    GRANTED = "granted"
    DENIED = "denied"
    UNSET = "unset"

class ConsentMode(StrEnum):
    SEED = "seed"
    OVERRIDE = "override"
```

The resolution path now compares by identity (`is TelemetryConsent.GRANTED`) instead of string equality, and `_coerce_consent` uses `TelemetryConsent(value)` with a `ValueError` guard rather than a hand-maintained `_VALID_CONSENT` set, which is gone.

The comment was anchored on the env-var name constants, so if you actually meant *those* should be an enum rather than the consent values, say so and I will move it — I read it as the consent/mode literals since that mirrors the `EventName` change.

openhands-agent-server/openhands/agent_server/telemetry/policy.py
VascoSch92: Done — `__all__` now sits directly after the imports (line 24 of 216).

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
VascoSch92: Done — same here, `__all__` moved to just after the imports (line 43 of 281).

openhands-agent-server/openhands/agent_server/telemetry/subscriber.py
VascoSch92: Done:

```python
match event:
    case AgentErrorEvent():
        self._emit_error_from_agent_event(event)
    case ConversationErrorEvent():
        self._emit_error_from_conversation_event(event)
    case ConversationStateUpdateEvent():
        self._handle_state_update(event)
```

The state-update branch also moved into its own method — it was three levels of nesting inside the dispatch, and the early-return reads better on its own.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: Done — the edges and the lookup now live together:

```python
class Bounds(NamedTuple):
    edges: tuple[float, ...]

    def label(self, value: float | None) -> str:
        ...
        index = bisect.bisect_right(self.edges, numeric)
```

Clearer, because a caller can no longer pair a value with the wrong scale — `DURATION_BOUNDS.label(x)` instead of `bucket(x, SOME_TUPLE)`. Faster, because the linear scan became a binary search. `bucket(value, bounds)` is kept as a thin wrapper so the call sites read the same.

openhands-agent-server/openhands/agent_server/telemetry/sanitizer.py
VascoSch92: I tried to remove it and could not — reporting what I found rather than forcing it.

`litellm` models **providers**, not model **families**, and the same family is served by many providers. Deriving the mapping from `litellm.models_by_provider` gives an ambiguous answer for every family that matters:

```
claude    -> anthropic, azure_ai, bedrock, deepinfra
gpt       -> aiml, azure, azure_ai, azure_text
gemini    -> deepinfra, fal_ai, gemini, github_copilot
deepseek  -> azure_ai, bedrock, clarifai, cloudflare
llama     -> anyscale, azure_ai, cerebras, cloudflare
```

`litellm.get_llm_provider()` does not help either — it raises `BadRequestError` for bare names like `claude-sonnet-4`, `deepseek-chat`, `llama-3` and `grok-2`.

So this table only exists for the case where the model string has **no** provider prefix. The prefixed path already uses `litellm.provider_list` (from the previous round), and there is a test asserting every hint target here is a real litellm provider, so it cannot silently drift. I have recorded that reasoning inline so the next reader does not repeat the experiment.

If you would rather drop the fallback entirely, bare model names would just report `other` — happy to do that if the reduced fidelity is acceptable.

malhotra5:
NIT: for the telemetry service would be nice to automatically parse a dedicated header like `X-OpenHands-Telemetry-Distinct-Id`, this way we can attribute other endpoint activities like llm profile creation 


The FE would be responsible for attaching the header properly so that the BE can appropriately attribute

```
headers["X-OpenHands-Telemetry-Distinct-Id"] = posthog.get_distinct_id()
```

openhands-agent-server/openhands/agent_server/api.py
```diff
@@ -528,6 +599,7 @@ async def _unhandled_exception_handler(
             error_id,
             exc_info=(type(exc), exc, exc.__traceback__),
         )
+        _emit_request_failed(request, exc, error_id)
         return JSONResponse(status_code=500, content=content)
```
malhotra5, line 612: Any idea why we don't emit request failed on the Base exception group errors above ?

openhands-agent-server/openhands/agent_server/init_router.py
```diff
@@ -203,6 +218,11 @@ async def initialize(self, req: InitRequest) -> InitStatus:
                 for key, value in req.env.items():
                     os.environ[key] = value
 
+            # Must precede get_instance(), which captures the sink.
+            await shutdown_telemetry_sink()
+            self._app.state.telemetry_sink = await build_telemetry_sink(new_config)
+            emit_server_started()
```
malhotra5: This emits `server_started` before deferred init has actually reached `ready`. If any later step fails (`ConversationService.get_instance`, bash service enter, service enter, root path update, etc.), the handler rolls back to `dormant` without emitting `server_stopped` or resetting the telemetry lifecycle flag. A retry can then emit another `server_started`, because `emit_server_started()` does not early-return when `_server_started_emitted` is already true.

Can we either move this until after the init transition succeeds, or make the failure path stop/reset the telemetry lifecycle state? A focused test for “init fails after telemetry start, then retry” would catch this.

_This inline review comment was created by an AI agent (OpenHands) on behalf of the user._

malhotra5 (approved):
LGTM

Couple of nits, and would be great to test this in canvas

VascoSch92:
Good call — done in `dc11aafef`.

The agent-server now reads `X-OpenHands-Telemetry-Distinct-Id` and uses it as the PostHog `distinct_id` for request-scoped events. `request_failed` is wired up as the first consumer (it previously always fell back to `anon:` since there is no conversation `user_id` at that point), so the pattern is in place for LLM-profile creation and other endpoint events as they start emitting.

```python
# frontend
headers["X-OpenHands-Telemetry-Distinct-Id"] = posthog.get_distinct_id()
```

It is trusted the same way as `user_id` — the frontend owns setting it — and sanitized the same way: control characters stripped, bounded to 256 chars, empty falls back to the anonymous id. Conversation-scoped events are unchanged; they keep using the `user_id` from the stored conversation. Documented in the agent-server README under Telemetry → Identity.

openhands-agent-server/openhands/agent_server/api.py
VascoSch92: Good catch — no principled reason, it was an oversight in that branch. The `BaseExceptionGroup`-with-no-`HTTPException` path returns a genuine 500 but fell through to `return JSONResponse(...)` without calling `_emit_request_failed`, so those failures were invisible to telemetry. Fixed in `922df5e17` — it now emits `request_failed` like the plain-unhandled path, with a regression test that fails against the old branch.

openhands-agent-server/openhands/agent_server/init_router.py
VascoSch92: You are right on every point, and the retry double-emit is the real problem. Fixed in `922df5e17` with your option (a): `emit_server_started()` now runs **after** the `ready` transition, not before `get_instance()`. So a failed init emits no start, rolls back to dormant cleanly, and a successful retry emits exactly one.

I also made `emit_server_started()` idempotent as defense in depth, since you flagged that it did not early-return when `_server_started_emitted` was already true — a start while one is outstanding is now a no-op.

Added the focused test you asked for: init forced to fail in `get_instance()` after the sink is built, then a retry that succeeds. It asserts zero starts after the failure and exactly one after the retry, and it fails against the previous code (verified by reverting each fix in isolation).

PR #4176, opened 2026-07-21
feat(marketplace): auto-load standalone marketplace skills

ak684:
HUMAN:

I would like conversation auto-load for standalone skill marketplaces  in cloud/ohe to work instead of not work like it does today.

---

AGENT:

## Why

Marketplace **auto-load** only loads plugins. A `marketplace.json` can also declare standalone `skills` (each a `MarketplaceEntry` with a `source`), and those are parsed and surfaced in the catalog/preview — but `load_registered_marketplace_skills` and the SDK-native `LocalConversation` auto-load iterate only `marketplace.plugins`. So a skills-only marketplace registered with `auto_load=True` loads **nothing**. The canonical `examples/01_standalone_sdk/43_mixed_marketplace_skills` marketplace is skills-only, so we ship an example that appears to work but auto-loads no skills. This is purely an auto-load gap: standalone skills already parse and appear in the catalog; they just never reach a conversation.

## Summary

- Load standalone `marketplace.skills` in both auto-load paths (`openhands-agent-server` `load_registered_marketplace_skills` and `openhands-sdk` `LocalConversation`), reusing the install flow's resolution: `resolve_source_path` → `find_skill_md` → `Skill.load(strict=False)`. Standalone skills merge at low precedence so plugins loaded afterwards override same-named skills (matching the catalog's plugin priority).
- Add `MarketplaceRegistration.auto_loads_skill` (mirrors `auto_loads_plugin`); `auto_load` now covers plugins **and** standalone skills, and a name list selects skills the same way it selects plugins.

## Issue Number

N/A (found while validating Bitbucket Data Center personal-repo marketplaces for a customer).

## How to Test

Unit: `uv run pytest tests/agent_server/test_skills_service.py tests/sdk/marketplace/ tests/sdk/conversation/test_local_conversation_plugins.py` (148 pass; new tests cover skills-only auto-load, selective-by-name, mixed plugins+skills, and SDK-native `LocalConversation` standalone-skill auto-load). Full `tests/sdk/conversation/` suite: 737 pass. Lint/format clean.

End-to-end against a real private Bitbucket Data Center personal repo (`~alona/marketplace-skills-only`, skills-only `marketplace.json`):

```
$ python -c "load_registered_marketplace_skills([MarketplaceRegistration(name=..., source=<authed bbdc url>, auto_load=True)])"
INFO  Fetching marketplace 'alona-skills-only' from https://****@bitbucket-02.../scm/~alona/marketplace-skills-only.git
INFO  Cloning repository ...
STANDALONE-SKILLS loaded: 2
  - greeting-style (trigger=KeywordTrigger)
  - commit-format (trigger=always/None)
selective ['greeting-style']: ['greeting-style']
```

Before this change the same call returned 0 skills (plugins-only).

## Video/Screenshots

CLI reproduction included above (headless SDK change; no GUI surface).

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Marketplaces only declare `plugins` and `skills`; plugins already auto-load and everything else (commands/agents/hooks/mcp) lives inside a plugin, so standalone skills are the complete gap. Standalone skills whose source is a non-github remote URL resolve to `None` and skip with a warning — the same limitation the existing `install_skills_from_marketplace` flow has.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:4f943b1-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-4f943b1-python \
  ghcr.io/openhands/agent-server:4f943b1-python
```

```
ghcr.io/openhands/agent-server:4f943b1-golang-amd64
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-golang-amd64
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-golang-amd64
ghcr.io/openhands/agent-server:4f943b1-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:4f943b1-golang-arm64
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-golang-arm64
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-golang-arm64
ghcr.io/openhands/agent-server:4f943b1-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:4f943b1-java-amd64
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-java-amd64
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-java-amd64
ghcr.io/openhands/agent-server:4f943b1-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:4f943b1-java-arm64
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-java-arm64
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-java-arm64
ghcr.io/openhands/agent-server:4f943b1-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:4f943b1-python-amd64
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-python-amd64
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-python-amd64
ghcr.io/openhands/agent-server:4f943b1-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:4f943b1-python-arm64
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-python-arm64
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-python-arm64
ghcr.io/openhands/agent-server:4f943b1-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:4f943b1-golang
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-golang
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-golang
ghcr.io/openhands/agent-server:4f943b1-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:4f943b1-java
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-java
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-java
ghcr.io/openhands/agent-server:4f943b1-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:4f943b1-python
ghcr.io/openhands/agent-server:4f943b16ae717ba01b382db6f135181a0bb50241-python
ghcr.io/openhands/agent-server:alona-marketplace-autoload-standalone-skills-python
ghcr.io/openhands/agent-server:4f943b1-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `4f943b1-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `4f943b1-python-amd64`) are also available if needed


VascoSch92:
**1. The new standalone-skills block is copy-pasted into both auto-load paths.**

The loop added in `skills_service.py` (`load_registered_marketplace_skills`, lines ~336-361) and the one in `local_conversation.py` (`_ensure_plugins_loaded`, lines ~893-919) are the same ~26 lines, identical apart from the accumulator variable name and some log line-wrapping:

```python
for skill_entry in marketplace.skills:
    if not registration.auto_loads_skill(skill_entry.name):
        continue
    try:
        skill_dir = resolve_source_path(skill_entry.source, base_path=marketplace_path)
        skill_md = find_skill_md(skill_dir) if skill_dir else None
        if skill_md is None:
            logger.warning(...); continue
        <acc>.append(Skill.load(skill_md, strict=False))
    except Exception:
        logger.warning(...)
```

This also sits right on top of the plugin loop, which was already duplicated across these two files. Rather than extend both copies, could we pull this out into a single helper (something like `load_marketplace_standalone_skills(marketplace, marketplace_path, registration) -> list[Skill]`) and call it from both sites? That collapses the plugin+skills duplication instead of doubling it.

**2. `find_skill_md` + `Skill.load` already exists as a helper.**

`installed.py` has `_load_skill_from_dir` which does exactly this:

```python
def _load_skill_from_dir(skill_root: Path) -> Skill:
    skill_md = find_skill_md(skill_root)
    if not skill_md:
        raise SkillValidationError(...)
    return Skill.load(skill_md, strict=True)
```

The only real difference here is `strict=False` + warn-and-skip vs. `strict=True` + raise. Might be worth parameterizing that helper (e.g. a `strict` arg) and reusing it, so the `find_skill_md` → `Skill.load` step lives in one place — right now this pattern would exist in three spots (the install flow at `installed.py:185-192` plus the two new copies).

VascoSch92 (changes requested):
see comments above

ak684:
Thanks @VascoSch92 — both good calls, addressed in 3b16326:

**1 & 2 — shared helper + reuse `_load_skill_from_dir`.** Extracted `load_marketplace_standalone_skills(marketplace, marketplace_path, registration) -> list[Skill]` into `openhands-sdk` (`skills/installed.py`, next to `install_skills_from_marketplace`, since the agent-server depends on the SDK and both call sites now import it from there). It reuses `_load_skill_from_dir`, which I parameterized with `strict: bool = True` — so the `find_skill_md` → `Skill.load` step lives in one place, and the auto-load path passes `strict=False` + warn-and-skip while the install path keeps `strict=True` + raise. Both `load_registered_marketplace_skills` (agent-server) and `LocalConversation._ensure_plugins_loaded` (SDK) now just call the helper. Added direct unit tests for it (all-load / select-by-name / not-auto-load / missing-SKILL.md-skips).

**On collapsing the plugin loops too:** I deliberately left those separate — they only *look* alike. In the agent-server path the plugin loop fetches and loads plugins into skills inline (`Plugin.load(...).get_all_skills()`), whereas in `LocalConversation` it collects `PluginSource` specs into `plugins_to_load` for a later loop that also pulls hooks, agents, MCP config, and records resolved commit SHAs for persistence. Unifying those would mean forcing one path to do work it doesn't need (or a callback that hurts more than it helps), so I scoped the dedup to the standalone-skills block, which genuinely was identical. Happy to revisit if you'd still like a shared plugin helper.

No behavior change — full skills/conversation/marketplace suites pass, lint/format clean.

VascoSch92:
There is still a thing but then I think we are good to go.

The problem is that precedence is inverted in the agent-server path

The catalog builder establishes the rule (skills_service.py:742): `# Plugins take priority: if a name appears in both plugins and skills, the plugin version is used`

...and enforces it with if skill_entry.name not in entries at line 759. 

The SDK path does match this but the agent-server not

At `skills_service.py:336` the standalone skills are appended after the plugin loop into one flat `list. load_all_skills` (line 379, the only caller) passes that list to `merge_skills` as a single tier, and merge_skills is last-wins within a tier — so appending last means winning.

I suggest to move the standalone `all_skills.extend(...)` block above the for entry in `marketplace.plugins`: loop. 

Moreover, we have a test gap, flipping the precedence in either direction leaves all 170 tests passing. The new tests only use disjoint names, so the rule the PR explicitly claims to implement isn't pinned by anything. Worth a test asserting plugin-over-standalone in both paths.

ak684:
Good catch @VascoSch92 — fixed in eeb5743.

**Precedence:** you're right, the agent-server path had it inverted. `load_registered_marketplace_skills` returns one flat tier and `merge_skills` is last-wins, so appending standalone *after* the plugin loop let a standalone skill win over a same-named plugin skill. Moved the standalone `extend(...)` **above** the plugin loop so plugins are appended last and win — matching the catalog rule (and the SDK path, which was already correct via the low-precedence merge + later `add_skills_to`).

**Test gap:** added `test_plugin_wins_over_standalone_on_name_collision` for **both** paths — a marketplace declaring a plugin that bundles skill `shared` (`FROM_PLUGIN`) *and* a standalone skill `shared` (`FROM_STANDALONE`), asserting the loaded `shared` is the plugin's. I verified these actually pin the rule: with the order flipped they fail (`FROM_STANDALONE` wins), and `merge_skills([[standalone, plugin]])` vs `[[plugin, standalone]]` confirms the last-wins mechanism.

512 tests pass, lint/format clean.

VascoSch92 (approved):

PR #4181, opened 2026-07-22
fix(agent-server): /api/vscode/url without base_url advertises the configured VSCode port

harish-chandramowli:
HUMAN:
  I ran into the issue due to port collision as i had another service in the hardcoded ports .

---

AGENT:

This PR was prepared with an AI agent (Claude Code). Evidence and reproduction details below; the underlying bug was observed on a real self-hosted deployment before any code was changed.

## Why

`GET /api/vscode/url` without a `base_url` query param builds the URL from the route's hardcoded string default `"http://localhost:8001"`, which is decoupled from the port the VSCode server actually binds (`config.vscode_port` / `OH_VSCODE_PORT`). When the server is launched on any other port — agent-canvas launches it on `backend + 1000` = 19000 — a caller that omits `base_url` receives a URL pointing at a port nothing listens on, and there is no API through which a client can discover the real port.

Observed live on a self-hosted agent-canvas 1.5.2 + openhands-agent-server 1.36.1 VM (Ubuntu 24.04, install per the official backend-setup docs): openvscode-server bound `0.0.0.0:19000` (confirmed via `ss -ltnp`), while `curl http://127.0.0.1:18000/api/vscode/url` (no param, valid session key header) returned `{"url":"http://localhost:8001/?tkn=…&folder=workspace"}` — port 8001 had no listener. Full analysis incl. the client-side half: OpenHands/OpenHands#15434.

## Summary

- Default the route's `base_url` param to `None` instead of `"http://localhost:8001"`, so `VSCodeService.get_vscode_url` falls through to its existing `http://localhost:{self.port}` branch (previously dead code on this route) and the advertised port matches the real bind.
- Callers that pass `base_url` explicitly are unaffected — passthrough behavior is unchanged.
- Added `test_get_vscode_url_default_uses_configured_port` locking in the new no-param behavior.

## Issue Number

OpenHands/OpenHands#15434

## How to Test

Unit tests:

```bash
uv sync
uv run pytest tests/agent_server/test_vscode_router.py tests/agent_server/test_vscode_service.py -q
# 42 passed
```

Live repro of the fix (any machine with the agent-server installed and an openvscode-server at the expected root):

```bash
OH_VSCODE_PORT=19000 uv run agent-server --host 127.0.0.1 --port 18000 &
curl -s -H "X-Session-API-Key: <key>" http://127.0.0.1:18000/api/vscode/url
# before this PR: {"url":"http://localhost:8001/?tkn=..."}   (nothing listens on 8001)
# after this PR:  {"url":"http://localhost:19000/?tkn=..."}  (matches the actual bind)
curl -s -H "X-Session-API-Key: <key>" "http://127.0.0.1:18000/api/vscode/url?base_url=http://example.com"
# unchanged in both: {"url":"http://example.com/?tkn=..."}
```

`ruff check` and `ruff format --check` are clean on the changed files.

## Video/Screenshots

N/A (REST endpoint behavior; before/after curl transcript above).

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The strict-default behavior only changes for callers that omit `base_url`; the shipped agent-canvas frontend always sends `base_url` explicitly, so this is additive for current clients. The canvas-side design choice (advertise the port vs. proxy VSCode through the ingress at a base path) is deliberately left to maintainers in OpenHands/OpenHands#15434.

https://claude.ai/code/session_015So55gyabKA29kzuQwXQKR


enyst:
Thank you for the PR! I think the REST API compatibility checks failure is actually acceptable… it’s a good reason but a release note should do to inform users.

@openhands WDYT? /codereview this PR, post your feedback with GH API and feel free to approve or whatever event you want to send.

enyst (approved):
🟢 **Good taste**

The change is the minimal fix: an omitted `base_url` now reaches the existing `VSCodeService` fallback, which uses the configured VSCode port, while explicit `base_url` values remain unchanged. The REST API compatibility failure is expected because OpenAPI no longer advertises the stale `http://localhost:8001` default and now models omission as nullable; the `release-note-required` label is already present to communicate that client-visible contract change. Focused validation passed: 42 tests, Ruff lint, and Ruff formatting.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 LOW — two localized files, no dependency or security changes, explicit-call behavior is preserved, and the changed omitted-value path is covered.

**VERDICT:**
✅ **Worth merging**: Core logic is sound and the intentional compatibility change is release-noted.

**KEY INSIGHT:**
The route should not duplicate a deployment-specific default already owned by the configured service.

PR #4185, opened 2026-07-22
fix(sdk): accept boolean JSON Schema nodes in _process_schema_node

### LINKED ISSUE #4182: [Bug]: Boolean MCP schemas still crash `_process_schema_node` (`items: true`)

### Bug Description

This appears related to #1923 , which was closed without a code fix.

I am opening a new report because the issue is still reproducible on OpenHands SDK `1.36.1` with the minimal example below.

For example, the following MCP input property causes the crash:

  ```json
  {
    "cursor": {
      "description": "Pagination cursor",
      "anyOf": [
        {
          "type": "array",
          "items": true
        },
        {
          "type": "null"
        }
      ]
    }
  }
```

Boolean schemas (true and false) are valid JSON Schema values. In this case, items: true means that every array item is accepted.

The SDK recursively passes true to _process_schema_node(), which assumes that every schema node is a dictionary. It then evaluates:

```python
if "$ref" in node:
````
Because node is a Python boolean, this raises: TypeError: argument of type 'bool' is not iterable.

The failure occurs while converting MCP tool schemas for an LLM request. The same conversion is also used by token counting.

### Expected Behavior

The SDK should accept boolean JSON Schema nodes.

For this input:

  ```json
  {
    "type": "array",
    "items": true
  }
```

the generated tool schema should remain valid, for example by normalizing it to the equivalent object schema:

```json
  {
    "type": "array",
    "items": {}
  }
```

The MCP tool should then be included in the LLM request successfully, and token counting should complete without raising an exception.

### Actual Behavior

Tool schema conversion fails before the LLM request is sent:

```text
  TypeError: argument of type 'bool' is not iterable
```

As a result:

 - the conversation run fails;

- the LLM request is not sent;

### Steps to Reproduce

1. Install `openhands-sdk==1.36.1`.
2. Create an MCP-style object input schema containing an array with `"items": true`.
3. Pass the schema to `_process_schema_node()`, or expose it through an MCP tool and let the SDK prepare the tool schema for an LLM request.
4. Observe that schema processing raises `TypeError`.

The following minimal code reproduces the underlying failure directly:

```python
from openhands.sdk.tool.schema import _process_schema_node

schema = {
    "type": "object",
    "properties": {
        "cursor": {
            "description": "Pagination cursor",
            "anyOf": [
                {
                    "type": "array",
                    "items": True,
                },
                {
                    "type": "null",
                },
            ],
        }
    },
}

_process_schema_node(schema, {})
```

### Installation Method

Installed from PyPI using uv with `openhands-sdk==1.36.1` pinned in `pyproject.toml`.

_No response_

### SDK Version

1.36.1

### Version Confirmation

### Python Version

3.12.3

Not model-specific. The failure occurs before the LLM request is sent

### Operating System

Linux

### Logs and Error Messages

```text
  File "<reproduction>", line 20, in <module>
    _process_schema_node(schema, {})
  File ".../site-packages/openhands/sdk/tool/schema.py", line 154, in _process_schema_node
    result["properties"][prop_name] = _process_schema_node(
  File ".../site-packages/openhands/sdk/tool/schema.py", line 140, in _process_schema_node
    processed = _process_schema_node(non_null_types[0], defs, _visiting)
  File ".../site-packages/openhands/sdk/tool/schema.py", line 165, in _process_schema_node
    result["items"] = _process_schema_node(node["items"], defs, _visiting)
  File ".../site-packages/openhands/sdk/tool/schema.py", line 108, in _process_schema_node
    if "$ref" in node:
TypeError: argument of type 'bool' is not iterable
```

### Minimal Code Sample

```python
from openhands.sdk.tool.schema import _process_schema_node

schema = {
    "type": "object",
    "properties": {
        "cursor": {
            "description": "Pagination cursor",
            "anyOf": [
                {
                    "type": "array",
                    "items": True,
                },
                {
                    "type": "null",
                },
            ],
        }
    },
}

_process_schema_node(schema, {})
```

_No response_

VascoSch92:
HUMAN:

Fix a small bug with MCP server and JSON schema

---

AGENT:

## Why

JSON Schema allows a boolean wherever a schema is expected: `true` accepts any
value, `false` accepts none. MCP servers emit these in practice, e.g.

```json
{"cursor": {"description": "Pagination cursor",
            "anyOf": [{"type": "array", "items": true}, {"type": "null"}]}}
```

`_process_schema_node` assumed every node was a dict, so `"$ref" in node`
raised `TypeError: argument of type 'bool' is not iterable`.

This is not confined to the private helper. `MCPToolDefinition._get_tool_schema`
feeds the **raw, server-provided** `inputSchema` straight into it
(`openhands-sdk/openhands/sdk/mcp/tool.py:374-377`), so a single MCP server
advertising `items: true` breaks the request before it is sent. The tool
conversion in `LLM.get_token_count` (`llm.py:2764`) sits *outside* the
`try/except` at `llm.py:2786`, so the `TypeError` propagates out of token
counting rather than being swallowed.

## Summary

- Normalize boolean schema nodes to their object equivalents (`true` → `{}`,
  `false` → `{"not": {}}`) at the top of `_process_schema_node`.
- Guard the `anyOf` member filter, which raised a *separate*
  `AttributeError: 'bool' object has no attribute 'get'` on the same input.
- Add `TestBooleanSchemaHandling` covering `items: true` / `items: false`,
  booleans inside `anyOf`, boolean property schemas, and the exact schema
  from the issue.

## Issue Number

Fixes #4182

## How to Test

**1. Unit tests**

```
uv run pytest tests/sdk/tool/test_mcp_schema.py -q
# 17 passed
```

Each new test was confirmed to fail on `main` before the fix (HEAD's function
loaded via `importlib`, working tree untouched):

```
FAILS on HEAD: test_items_true_normalized_to_empty_schema -> TypeError: argument of type 'bool' is not iterable
FAILS on HEAD: test_items_false_normalized_to_never_schema -> TypeError: argument of type 'bool' is not iterable
FAILS on HEAD: test_boolean_inside_any_of              -> AttributeError: 'bool' object has no attribute 'get'
FAILS on HEAD: test_boolean_property_schema            -> TypeError: argument of type 'bool' is not iterable
FAILS on HEAD: test_issue_4182_reported_schema         -> TypeError: argument of type 'bool' is not iterable
```

**2. End-to-end — real `MCPToolDefinition` through the LLM request path**

Unit tests alone are not sufficient here, so this drives an actual MCP tool
built from the issue's schema through the code that prepares the LLM request:

```python
tool = MCPToolDefinition(
    description="list items", action_type=MCPToolAction, observation_type=None,
    executor=DummyExecutor(),
    mcp_tool=mcp.types.Tool(name="list_items", description="d", inputSchema={
        "type": "object",
        "properties": {"cursor": {"description": "Pagination cursor",
            "anyOf": [{"type": "array", "items": True}, {"type": "null"}]}},
    }),
)
llm = LLM(model="gpt-4o", api_key=SecretStr("test"), service_id="t")
llm.get_token_count([Message(role="user", content=[TextContent(text="hello")])],
                    tools=[tool])
```

| Path | `main` | This PR |
|---|---|---|
| `to_openai_tool()` | `TypeError` | `{'type': 'array', 'items': {}, 'description': 'Pagination cursor'}` |
| `to_responses_tool()` | `TypeError` | same |
| `llm.get_token_count()` | `TypeError` | returns `84` |

The emitted schema matches what the issue asks for: `items: true` → `items: {}`.

**3. Normalization semantics verified against the `jsonschema` library**

Both rewrites are valid Draft 2020-12 and behaviourally identical to the
booleans they replace:

```
items:{}        accepts [1,"a"] -> True    (orig items:true  -> True)
items:{"not":{}} accepts [1]    -> False   (orig items:false -> False)
items:{"not":{}} accepts []     -> True
```

**4. Full suite + lint**

```
uv run pytest tests/sdk -q
# 5431 passed, 9 skipped, 13 xfailed

uv run pre-commit run --files openhands-sdk/openhands/sdk/tool/schema.py tests/sdk/tool/test_mcp_schema.py
# ruff format / ruff lint / pycodestyle / pyright / import rules / Tool registration -> all Passed
```

## Video/Screenshots

Not applicable — this is a library-internal schema conversion fix with no UI
surface. The reproduction and before/after output are in "How to Test" above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Deliberately scoped to `_process_schema_node`, per the issue. Three adjacent
problems were found and verified while investigating but are **not** fixed here:

1. `fn_call_converter.py:296` — `if "enum" in param_info:` is unguarded while
   line 292 immediately above already does `isinstance(param_info, dict)`.
   Raises the same `TypeError` on a boolean sub-schema.
2. `client_tool.py:291` — `ClientTool` copies `input_schema` verbatim, bypassing
   `_process_schema_node` entirely, which is what keeps (1) reachable.
3. Non-dict, non-bool nodes still crash: tuple-form `items` (draft-07,
   `{"items": [{...}]}`) raises `AttributeError: 'list' object has no attribute
   'get'`. Widening the guard to `not isinstance(node, dict)` would cover it.

Happy to fold any of these in if reviewers prefer a broader fix.

One judgment call: `false` is mapped to `{"not": {}}` (semantically correct)
rather than `{}`. If a provider is used in OpenAI **strict** function-calling
mode, `not` may be rejected — trading a `TypeError` for an API 400. `items:
false` is rare enough that correctness seemed the better default, but this is
easy to change.

Note that `_process_schema_node` is lossy by design (it keeps only the first
`anyOf` branch), so `anyOf: [false, {"type": "null"}]` normalizes to
`{"not": {}}` rather than "must be null". That is consistent with the existing
behaviour for `anyOf: [string, integer]` and is not addressed here.

### DIFF
--- a/openhands-sdk/openhands/sdk/tool/schema.py
+++ b/openhands-sdk/openhands/sdk/tool/schema.py
@@ -69,7 +69,7 @@ def _shallow_expand_circular_ref(ref_def: dict[str, Any]) -> dict[str, Any]:
 
 
 def _process_schema_node(
-    node: dict[str, Any],
+    node: dict[str, Any] | bool,
     defs: dict[str, Any],
     _visiting: frozenset[str] | None = None,
 ) -> dict[str, Any]:
@@ -104,6 +104,12 @@ def _process_schema_node(
     if _visiting is None:
         _visiting = frozenset()
 
+    # JSON Schema allows boolean sub-schemas: `true` accepts any value and
+    # `false` accepts none. Normalize them to their object equivalents so the
+    # rest of this function can assume a dict node.
+    if isinstance(node, bool):
+        return {} if node else {"not": {}}
+
     # Handle $ref references
     if "$ref" in node:
         ref_path = node["$ref"]
@@ -134,7 +140,11 @@ def _process_schema_node(
 
     # Handle anyOf (often used for optional fields with None)
     if "anyOf" in node:
-        non_null_types = [t for t in node["anyOf"] if t.get("type") != "null"]
+        non_null_types = [
+            t
+            for t in node["anyOf"]
+            if not isinstance(t, dict) or t.get("type") != "null"
+        ]
         if non_null_types:
             # Process the first non-null type
             processed = _process_schema_node(non_null_types[0], defs, _visiting)
--- a/openhands-sdk/openhands/sdk/tool/schema.py
+++ b/openhands-sdk/openhands/sdk/tool/schema.py
@@ -104,9 +104,7 @@ def _process_schema_node(
     if _visiting is None:
         _visiting = frozenset()
 
-    # JSON Schema allows boolean sub-schemas: `true` accepts any value and
-    # `false` accepts none. Normalize them to their object equivalents so the
-    # rest of this function can assume a dict node.
+    # JSON Schema booleans: `true` accepts any value, `false` accepts none.
     if isinstance(node, bool):
         return {} if node else {"not": {}}
 
--- a/tests/sdk/tool/test_mcp_schema.py
+++ b/tests/sdk/tool/test_mcp_schema.py
@@ -3,6 +3,7 @@
 import json
 from collections.abc import Sequence
 
+import pytest
 from pydantic import Field
 
 from openhands.sdk.llm import ImageContent, TextContent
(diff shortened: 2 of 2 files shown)


tests/sdk/tool/test_mcp_schema.py
```diff
@@ -386,3 +386,69 @@ class LinkedListNode(Schema):
         assert result["properties"]["next"]["type"] == "object"
 
         json.dumps(result)
+
+
+class TestBooleanSchemaHandling:
```
VascoSch92: Make the tests like methods. No need to a class here.

tests/sdk/tool/test_mcp_schema.py
```diff
@@ -386,3 +386,69 @@ class LinkedListNode(Schema):
...
+    """Tests for JSON Schema boolean sub-schemas.
+
+    JSON Schema allows ``true``/``false`` wherever a schema is expected:
+    ``true`` accepts any value, ``false`` accepts none. MCP servers emit
+    these (e.g. ``{"type": "array", "items": true}``), and the SDK used to
+    crash with ``TypeError: argument of type 'bool' is not iterable``.
+
+    Regression coverage for:
```
VascoSch92: I don't want the issue here

tests/sdk/tool/test_mcp_schema.py
```diff
@@ -386,3 +386,69 @@ class LinkedListNode(Schema):
         assert result["properties"]["next"]["type"] == "object"
 
         json.dumps(result)
+
```
VascoSch92, line 390: in general, you you can use pytest-mark-parametrize

openhands-sdk/openhands/sdk/tool/schema.py
```diff
@@ -104,6 +104,12 @@ def _process_schema_node(
     if _visiting is None:
         _visiting = frozenset()
 
+    # JSON Schema allows boolean sub-schemas: `true` accepts any value and
```
VascoSch92: trim this comment. Just the essential

neubig (commented):
An alternative solution here would be to never call `process_schema_node` on a boolean value (short-circuit the recursion if the child value is a boolean). That seems like it might be simpler and more transparent logic. WDYT?

VascoSch92:
Booleans are valid schemas for [JSON-schemas](https://json-schema.org/draft/2020-12/json-schema-core#section-4.3.2), so I think `dict | bool` is actually the correct annotation — the old one was too narrow.

Short-circuiting at the callers means guarding all four recursion points ($ref, anyOf, properties, items) with the same mapping, and any recursion site added later silently brings the bug back.

But your readability point is fair. Therefore,  I moved it to the first statement and documented the boolean case in the docstring.

But happy to change if you want.

neubig (approved):
LGTM then, thank you!

PR #3964, opened 2026-07-02
feat(mcp): add settings-backed MCP auth credentials

neubig:
HUMAN:

For all of these, I have actually tested that the Superhuman MCP works within the stack, and I am able to connect to it and use it in a conversation.

---

AGENT:

## Why

Superhuman Mail and similar hosted MCP servers use OAuth rather than a static API key. The SDK, agent-server, and canvas need one settings-backed credential model so OAuth credentials are captured once during MCP installation, encrypted/redacted through normal settings persistence, and reused when conversations start locally or from cloud-backed settings.

The current settings shape is `mcp_config: dict[str, MCPServer]`. Older persisted MCP shapes are handled only by settings migrations. Runtime code receives the current DataModel shape and normalizes to FastMCP or ACP only at those protocol boundaries.

## Summary

- Add typed MCP settings DataModels: `MCPServer` plus tagged `MCPAuthCredential` variants (`api_key`, `bearer`, `basic`, `header`, `oauth2`, etc.).
- Store MCP OAuth credentials under `mcp_config.<server>.auth.state`, with SDK storage helpers encrypting/decrypting `SecretStr` leaves through normal settings serialization.
- Migrate older persisted MCP auth/wrapper shapes into the current `mcp_config` server map during settings schema migrations; no runtime legacy normalization remains.
- Normalize the current `mcp_config` map to FastMCP only for MCP tool creation and to ACP's `mcp_servers` protocol keyword only at the ACP session call boundary.
- Return acquired MCP OAuth probe state from `/api/mcp/test` as `oauth_state`, leaving clients to persist it through the normal settings DataModel under `auth.state`.
- Rename SDK/workspace helpers, examples, fixtures, and tests to consistently use `mcp_config` for the native server map.
- Add a headless integration test using a real FastMCP OAuth-protected server, dynamic client registration, PKCE/token exchange, settings serialization, and a second connection that reuses persisted credentials without re-authentication.

### REST API contract changes

Compared with base OpenAPI `6e24af27bc0d` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -40,0 +41 @@
+operation GET /api/mcp/oauth/status/{job_id} operationId=get_mcp_oauth_status_api_mcp_oauth_status__job_id__get
@@ -94,0 +96,2 @@
+operation POST /api/mcp/oauth/callback/{job_id} operationId=submit_mcp_oauth_callback_api_mcp_oauth_callback__job_id__post
[...]

PR #4186, opened 2026-07-22
fix(agent): apply filter_tools_regex to runtime tools

### LINKED ISSUE #4184: [Bug]: Runtime MCP tools bypass Agent.filter_tools_regex

### Bug Description

`Agent.filter_tools_regex` is not applied to MCP tools materialized at conversation runtime.

Since MCP tool materialization was moved from `AgentBase._initialize()` to `LocalConversation._ensure_agent_ready()`, the current initialization flow is:

1. `AgentBase._initialize()` resolves and filters only statically configured `Agent.tools`.
2. `LocalConversation` materializes MCP tools.
3. The MCP tools are passed directly to `AgentBase.add_runtime_tools()`.
4. `add_runtime_tools()` inserts them into `_tools` without applying `filter_tools_regex`.

Although `Agent.init_state()` later calls `_initialize()` again, `_initialize()` returns immediately because `_initialized` is already `True`.

This is inconsistent with the `filter_tools_regex` field description, which states that the filter is applied after both explicitly configured tools and MCP tools are added.

This is more than a prompt-size issue: applications using the regex as a sub-agent tool boundary may expose filtered-out MCP tools to the model and allow those tools to be executed.

### Expected Behavior

`filter_tools_regex` should apply consistently to all non-default tools exposed to an Agent, including:

- Statically configured tools;
- MCP tools materialized when the conversation starts;
- MCP tools added by plugins at runtime;
- Any MCP tools dynamically advertised in the future.

An MCP tool whose name does not match the Agent's `filter_tools_regex` should:

- Not be added to `tools_map`;
- Not be sent to the LLM;
- Not be executable through that Agent.

Built-in default tools may continue to follow their existing documented behavior and remain outside this filter.

### Actual Behavior

The regex is applied only to tools resolved from `Agent.tools` during the first call to `AgentBase._initialize()`.

MCP tools are materialized afterward and passed to `add_runtime_tools()`, which adds every supplied tool directly to `_tools` without applying the regex.

As a result, MCP tools excluded by `filter_tools_regex`:

- Still appear in `agent.tools_map`;
- Are sent to the LLM;
- Can be executed by the Agent.

### Steps to Reproduce

1. Install `openhands-sdk==1.36.1`.
2. Create an Agent with `filter_tools_regex=r"^allowed$"`.
3. Initialize the Agent.
4. Add two runtime tools named `allowed` and `blocked`, reproducing the same `add_runtime_tools()` path used by `LocalConversation` for MCP tools.
5. Inspect `agent.tools_map`.
6. Observe that both `allowed` and `blocked` are present.

The production MCP path exhibiting the same behavior is:

```python
self.agent._initialize(self._state)
runtime_mcp_tools = self._runtime_mcp_tools_for_agent()
self.agent.add_runtime_tools(runtime_mcp_tools)
self.agent.init_state(...)
```

The final init_state() call does not reapply filtering because _initialize() immediately returns when _initialized is already True.

### Installation Method

Installed from PyPI using uv with `openhands-sdk==1.36.1` pinned in `pyproject.toml`.

_No response_

### SDK Version

1.36.1

### Version Confirmation

### Python Version

3.12.3

Not model-specific. The issue occurs before the LLM request is made.

### Operating System

Linux

### Logs and Error Messages

There is no exception. The failure is observable in the resulting tool map.

With a filter that only allows `allowed`, the runtime result is:

```text
Loaded 0 tools from spec
Filtered to 0 tools after applying regex filter
['allowed', 'blocked']
```
Relevant SDK flow:

AgentBase._initialize()
    -> applies filter_tools_regex to static tools
    -> sets _initialized = True

LocalConversation._ensure_agent_ready()
    -> materializes MCP tools
    -> add_runtime_tools(runtime_mcp_tools)
    -> init_state()
       -> _initialize()
          -> returns immediately because _initialized is True

### Minimal Code Sample

  ```python
  from unittest.mock import Mock, patch

  from openhands.sdk import Agent, LLM
  from openhands.sdk.conversation.state import ConversationState
  from openhands.sdk.tool import ToolDefinition


  llm = LLM(
      model="openai/test",
      api_key="test",
  )

  agent = Agent(
      llm=llm,
      tools=[],
      include_default_tools=[],
      filter_tools_regex=r"^allowed$",
  )

  state = Mock(spec=ConversationState)

  # Avoid unrelated local vision-profile discovery during this minimal test.
  with patch(
      "openhands.sdk.agent.base.has_vision_profile_available",
      return_value=False,
  ):
      agent._initialize(state)

  allowed = Mock(spec=ToolDefinition)
  allowed.name = "allowed"

  blocked = Mock(spec=ToolDefinition)
  blocked.name = "blocked"

  # This is the same Agent API used by LocalConversation to inject
  # runtime-materialized MCP tools.
  agent.add_runtime_tools([allowed, blocked])

  print(sorted(agent.tools_map))
  assert "allowed" in agent.tools_map
  assert "blocked" not in agent.tools_map
```

Actual output:

['allowed', 'blocked']

The final assertion fails.

_No response_

### PR DESCRIPTION
**Pull request by VascoSch92**
HUMAN:

Fixing the filter_tools_regext for runtime tools.

---

AGENT:

## Why

`Agent.filter_tools_regex` is documented as being "applied after any tools provided in `tools` and any MCP tools are added", but runtime-materialized MCP tools bypass it entirely. Tools excluded by the regex still land in `agent.tools_map`, are advertised to the LLM (`agent.py`, `tools=list(self.tools_map.values())`), and are executable — so applications using the regex as a tool boundary silently expose filtered-out MCP tools.

This is a regression: until #3964, `AgentBase._initialize()` materialized MCP tools itself and applied the regex afterwards (see `create_mcp_tools` followed by the `filter_tools_regex` block at `fe924f705^`). #3964 moved materialization into `LocalConversation._ensure_agent_ready()`, which registers the tools via `add_runtime_tools()` — a method introduced in #3826 that never applied the filter. The subsequent `init_state()` call does not re-filter because `_initialize()` returns immediately once `_initialized` is set.

## Summary

- Apply `filter_tools_regex` in `AgentBase.add_runtime_tools()`, so the single choke point covers both runtime-tool paths (conversation-start MCP tools and plugin-added tools) as well as any future dynamic-tool source.
- Built-in default tools (e.g. `InvokeSkillTool`, which the plugin path routes through the same method) are exempt, matching the existing exemption in `_initialize()`.
- Add regression tests (12) covering both paths plus edge cases — mixed batches, all-filtered batches, prefix-match anchoring, built-in subclasses, and duplicate handling around the filter; `filter_tools_regex` previously had zero test coverage anywhere in the repo, which is how this slipped through.

## Issue Number

Fixes #4184

## How to Test

End-to-end through the real production path (not just unit tests): build a `LocalConversation` whose injected `mcp_tool_provider` advertises tools named `allowed` and `blocked`, with the agent configured with `filter_tools_regex=r"^allowed$"`, then trigger `_ensure_agent_ready()` and inspect `tools_map`.

Before this fix (at `cae92b879`):

```text
filter_tools_regex = '^allowed$'
agent.tools_map    = ['allowed', 'blocked', 'inspect_image_with_vision']
AssertionError: BUG REPRODUCED: 'blocked' does not match ^allowed$ but is exposed to the agent/LLM
```

After this fix:

```text
INFO  Filtered runtime tools from 2 to 1 after applying regex filter
filter_tools_regex = '^allowed$'
agent.tools_map    = ['allowed', 'inspect_image_with_vision']
OK: filter applied to runtime MCP tools
```

(`inspect_image_with_vision` is an auto-attached built-in default tool, documented as exempt from the filter.)

That scenario is captured as `test_runtime_mcp_tools_apply_filter_tools_regex` in the new test file, so reviewers can reproduce with:

```bash
uv run pytest tests/sdk/agent/test_filter_tools_regex.py -q
```

Verification performed:

- `uv run pytest tests/sdk/agent/test_filter_tools_regex.py -q` → 12 passed. With `base.py` reverted to `origin/main`, the 6 regression tests fail and the 6 behavior-preservation tests still pass — confirming the tests catch the bug and guard existing behavior.
- `uv run pytest tests/sdk -q` (single process) → 5443 passed, 9 skipped, 13 xfailed; matches the pre-change baseline plus the 12 new tests.
- `uv run pre-commit run --files openhands-sdk/openhands/sdk/agent/base.py tests/sdk/agent/test_filter_tools_regex.py` → all hooks pass (ruff format/lint, pycodestyle, pyright, import rules, tool registration).

## Video/Screenshots

Terminal logs above (before/after output of the end-to-end reproduction and test runs).

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Filtering exempts built-ins via `isinstance` against `BUILT_IN_TOOL_CLASSES`, mirroring `_initialize()`'s exemption; without it, a restrictive regex would silently drop `InvokeSkillTool` injected by the plugin path and break skill invocation.
- `pattern.match` (prefix semantics) is deliberately kept consistent with `_initialize()`.
- Filtered-out MCP tools are dropped without closing their executors — identical to the pre-#3964 behavior, where `_initialize()` also discarded filtered MCP tools without cleanup. If a server's tools are all filtered its client connection lingers; flagging as a possible follow-up rather than expanding this fix.
- The plugin path has bypassed the filter since #3826 (two weeks before #3964 regressed the main `mcp_config` path); this fix covers both.

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/base.py
+++ b/openhands-sdk/openhands/sdk/agent/base.py
@@ -823,20 +823,18 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
         if self.filter_tools_regex:
             pattern = re.compile(self.filter_tools_regex)
             builtin_classes = tuple(BUILT_IN_TOOL_CLASSES.values())
-            kept = [
+            num_tools = len(tools)
+            tools = [
                 tool
                 for tool in tools
                 if isinstance(tool, builtin_classes) or pattern.match(tool.name)
             ]
-            if len(kept) != len(tools):
-                dropped = sorted(
-                    {tool.name for tool in tools} - {tool.name for tool in kept}
-                )
+            if len(tools) != num_tools:
                 logger.info(
-                    "Filtered out runtime tools not matching filter_tools_regex: %s",
-                    dropped,
+                    "Filtered runtime tools from %d to %d after applying regex filter",
+                    num_tools,
+                    len(tools),
                 )
-            tools = kept
 
         tool_names = [tool.name for tool in tools]
         if len(tool_names) != len(set(tool_names)):
--- a/openhands-sdk/openhands/sdk/agent/base.py
+++ b/openhands-sdk/openhands/sdk/agent/base.py
@@ -802,6 +802,11 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
             logger.warning("Error closing executor for tool '%s': %s", tool.name, exc)
 
     def add_runtime_tools(self, tools: Sequence[ToolDefinition]) -> None:
+        """Register tools materialized at runtime (e.g. MCP tools).
+
+        Tools are subject to `filter_tools_regex`; built-in default tools are
+        exempt, matching the behavior of `_initialize()`.
+        """
         if not self._initialized:
             logger.warning(
                 "add_runtime_tools called before agent initialization; "
@@ -814,6 +819,24 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
                     f"Tool {tool} is not an instance of 'ToolDefinition'. "
                     f"Got type: {type(tool)}"
                 )
+
+        if self.filter_tools_regex:
+            pattern = re.compile(self.filter_tools_regex)
+            builtin_classes = tuple(BUILT_IN_TOOL_CLASSES.values())
+            kept = [
+                tool
+                for tool in tools
+                if isinstance(tool, builtin_classes) or pattern.match(tool.name)
+            ]
+            if len(kept) != len(tools):
+                dropped = sorted(
+                    {tool.name for tool in tools} - {tool.name for tool in kept}
+                )
+                logger.info(
+                    "Filtered out runtime tools not matching filter_tools_regex: %s",
+                    dropped,
+                )
+            tools = kept
 
         tool_names = [tool.name for tool in tools]
         if len(tool_names) != len(set(tool_names)):
--- a/tests/sdk/agent/test_filter_tools_regex.py
+++ b/tests/sdk/agent/test_filter_tools_regex.py
@@ -1,0 +1,170 @@
+"""Tests for `Agent.filter_tools_regex`.
+
+The regex must apply both to statically configured tools resolved during
+`_initialize()` and to tools materialized at runtime (e.g. MCP tools) that are
+registered through `add_runtime_tools()`. Built-in default tools are exempt in
+both paths.
+
+Regression tests for https://github.com/OpenHands/software-agent-sdk/issues/4184
+"""
+
+import uuid
+from collections.abc import Sequence
+from typing import ClassVar, cast
+
+from openhands.sdk import LLM
+from openhands.sdk.agent import Agent
+from openhands.sdk.conversation.impl.local_conversation import LocalConversation
+from openhands.sdk.conversation.state import ConversationState
+from openhands.sdk.llm.message import ImageContent, TextContent
+from openhands.sdk.mcp.client import MCPClient
+from openhands.sdk.tool import ToolDefinition
+from openhands.sdk.tool.builtins import ThinkTool
+from openhands.sdk.tool.registry import register_tool
+from openhands.sdk.tool.spec import Tool
+from openhands.sdk.tool.tool import Action, Observation, ToolExecutor
+from openhands.sdk.workspace import LocalWorkspace
+
+
+class _FilterAction(Action):
+    text: str = ""
+
+
+class _FilterObs(Observation):
+    out: str = ""
+
+    @property
+    def to_llm_content(self) -> Sequence[TextContent | ImageContent]:
+        return [TextContent(text=self.out)]
+
+
+class _NoopExec(ToolExecutor[_FilterAction, _FilterObs]):
+    def __call__(self, action: _FilterAction, conversation=None) -> _FilterObs:
+        return _FilterObs(out="ok")
+
+
+class _AllowedTool(ToolDefinition[_FilterAction, _FilterObs]):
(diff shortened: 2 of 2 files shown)


openhands-sdk/openhands/sdk/agent/base.py
```diff
@@ -815,6 +820,24 @@ def add_runtime_tools(self, tools: Sequence[ToolDefinition]) -> None:
                     f"Got type: {type(tool)}"
                 )
 
+        if self.filter_tools_regex:
+            pattern = re.compile(self.filter_tools_regex)
+            builtin_classes = tuple(BUILT_IN_TOOL_CLASSES.values())
+            kept = [
```
VascoSch92: Can we just not filter the tool? We can just measure the len before doing that for the logs

tests/sdk/agent/test_filter_tools_regex.py
```diff
@@ -0,0 +1,170 @@
+"""Tests for `Agent.filter_tools_regex`.
+
+The regex must apply both to statically configured tools resolved during
+`_initialize()` and to tools materialized at runtime (e.g. MCP tools) that are
+registered through `add_runtime_tools()`. Built-in default tools are exempt in
+both paths.
+
+Regression tests for https://github.com/OpenHands/software-agent-sdk/issues/4184
```
VascoSch92: delete this. I don't want the issue

tests/sdk/agent/test_filter_tools_regex.py
```diff
@@ -0,0 +1,170 @@
+"""Tests for `Agent.filter_tools_regex`.
```
VascoSch92, line 1: Is this fix well-tested? Or just happy paths?

neubig (approved):
Thanks!

PR #4180, opened 2026-07-21
fix(agent-server): default bind host to loopback without a session API key

neubig:
HUMAN:

---

AGENT:

## Why

The agent server CLI used a fixed default host of `0.0.0.0` for every launch configuration. Local runs without a session API key and sandboxed runs with a configured key need different defaults, so the host is now selected from the effective session-key configuration.

## Summary

- The default host is now `127.0.0.1` when no `SESSION_API_KEY` / `OH_SESSION_API_KEYS_0` is configured.
- When a session API key is configured through the environment or config file, the default remains `0.0.0.0`.
- Explicit `--host` values remain supported. An explicit wildcard host without a key emits a configuration warning.
- Docker and Apptainer workspace launchers now forward `SESSION_API_KEY` and `OH_SESSION_API_KEYS_0` by default.
- Added unit coverage for default and explicit host selection with keys supplied through both environment variables and config files.

## Issue Number

N/A

## How to Test

Unit tests:

```bash
uv run pytest tests/agent_server/test_preload_modules.py -q
```

Live end-to-end test:

```bash
uv run python .pr/live_bind_test.py
```

The live test launches the actual agent server in three configurations:

1. No key with explicit `--host 0.0.0.0` → binds `0.0.0.0` and emits the configuration warning.
2. No key with the default host → binds `127.0.0.1`.
3. A configured `SESSION_API_KEY` with the default host → binds `0.0.0.0`.

The existing live-server suite was also run:

```bash
uv run pytest tests/cross/test_remote_conversation_live_server.py -q
```

## Video/Screenshots

Live before/after results are committed under `.pr/` and are auto-removed by the PR Artifacts workflow after approval.

**Before** (`.pr/bind_host_before.json`) — no key with the default host:

```json
"no_key_default": {
  "bind_line": ["... Uvicorn running on http://0.0.0.0:35025 ..."]
}
```

**After** (`.pr/bind_host_after.json`) — the same launch configuration:

```json
"no_key_default": {
  "bind_line": ["... Uvicorn running on http://127.0.0.1:42373 ..."]
}
```

The after-results also record the explicit-wildcard configuration warning and confirm that the default remains `0.0.0.0` when a key is configured.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

[...]

PR #4188, opened 2026-07-22
feat(agent-server): persist parent/child conversation relationships

### LINKED ISSUE #4119: Persist local parent-child conversation relationships in agent-server
## I can

I can launch a local child conversation from a parent and still see that relationship after either conversation or Agent Canvas is reloaded.

## Scope

Bring the public local agent-server conversation contract to parity with the existing OpenHands Cloud parent/child metadata needed by Agent Canvas. This issue makes no change to the Cloud API.

## Acceptance criteria

- [ ] The local start-conversation request accepts an optional parent conversation ID.
- [ ] Parent conversation information exposes child conversation IDs, and child information exposes its parent ID where appropriate.
- [ ] Relationships survive process/UI reload according to the server's existing persistence model.
- [ ] Invalid parent IDs and cross-workspace relationships fail clearly.
- [ ] Deletion behavior is explicit and tested; deleting a parent must not silently destroy child work.
- [ ] The TypeScript client surface is updated or verified so Agent Canvas does not need an untyped local-only call.
- [ ] Live evidence launches at least two local children and retrieves the relationship after reload.

VascoSch92:
HUMAN:

The PR persist local parent-child conversation relationships in agent-server

---

AGENT:

## Why

Agent Canvas needs to launch child conversations against a local agent-server and reconstruct the parent/child tree after a reload, the way it already can against OpenHands Cloud. The server currently has no conversation-to-conversation link other than fork lineage, so the relationship is lost the moment the client restarts.

## Summary

- Add optional `parent_conversation_id` to `StartConversationRequest`. Because `StoredConversation` subclasses the request, the link is persisted to the child's `meta.json` and rehydrated by `_load_catalog_sync` with no extra persistence code — the single source of truth is the child's record.
- Expose `parent_conversation_id` and a derived `sub_conversation_ids` on `ConversationInfo`. Both names mirror the Cloud contract (verified against `SharedConversation` in `OpenHands/typescript-client` and Agent Canvas's `AppConversation`/agent-server adapter, which already consume exactly these two names), so Canvas gets one shape across both backends. Children are computed from the in-memory catalog on read (never persisted, so the views can't diverge); the paginated `/search` path builds the reverse index once per page. Webhook payloads carry the parent but keep `sub_conversation_ids` empty to avoid an O(n) scan per event.
- Validate the parent at creation: unknown parent, self-parent, a parent that would close a cycle, or a parent in a different workspace → `422` via a new `InvalidParentConversation`, mapped in the router alongside the existing `ClientToolRegistrationError` pattern.
- Reject cycles with an ancestor walk rather than a self-parent check alone. A link is immutable once set, but conversation ids are caller-supplied and reusable: deleting a parent orphans its children *without* clearing their pointers, so re-creating a conversation under that freed id can close a loop (`A → B → A`) and hang any client walking the tree. `_ancestors_of` walks the parent chain from the proposed parent and rejects if it reaches the conversation being created; the walk carries a visited set so a cycle already persisted on disk terminates instead of hanging the request.
- Reset `parent_conversation_id` on fork. `fork_conversation` copies the source's stored metadata, so without an explicit override a fork of a child would silently register as an extra child of the same parent — bypassing every check above — and, when `fork_id` reuses a deleted parent's id, would make the fork its own parent and list itself in its own `sub_conversation_ids`. Fork lineage is already recorded by `forked_from_conversation_id`, so the fork is a new top-level conversation.
- Order `sub_conversation_ids` by `(created_at, id)` at the point the list is derived. `_load_catalog_sync` iterates the conversations directory with `iterdir()`, so catalog insertion order after a reload is filesystem order; without an explicit sort the field came back in creation order before a restart and in arbitrary order after one. Sorting where the list is built (rather than at load time) keeps the ordering a property of the list itself instead of an insertion order maintained elsewhere. The `id` secondary key breaks `created_at` ties, which would otherwise fall back to filesystem order through `sorted()`'s stability.
- Deleting a parent deliberately orphans its children (pointer left dangling), matching how `forked_from_conversation_id` already behaves when a fork source is deleted. Deleting a child shrinks the parent's derived view. Both are covered by tests.

Both new fields are optional with defaults: existing `meta.json` files validate unchanged and the REST contract change is additive-only.

## Issue Number

Closes #4119

## How to Test

Unit tests (15 in `tests/agent_server/test_parent_child_conversations.py`, covering create/link, reload via a fresh `ConversationService` over the same directory, all three 422 rejections, cycle rejection through a reused parent id, fork not inheriting the parent link, fork under a freed parent id not self-parenting, `sub_conversation_ids` ordering across reload, orphan-on-parent-delete, child-delete shrinking, search, and the top-level no-regression case):

```bash
uv run pytest tests/agent_server/test_parent_child_conversations.py   # 15 passed
uv run pytest tests/agent_server --ignore=tests/agent_server/stress   # 1493 passed
uv run pytest tests/sdk/conversation                                  # 741 passed
make test-server-schema                                               # schema valid
uv run python .github/scripts/check_agent_server_rest_api_breakage.py # no breaking changes*
```

\*oasdiff is not installed locally so the script skipped the diff step; the change is additive-only and the CI workflow runs the full check.

End-to-end against a real server (`python -m openhands.agent_server --port 8399`, driven with curl):

```text
=== 1. create parent ===
parent id: c2f81e3b-ad02-4f7b-9e27-b29c04b00f5b
=== 2. create two children naming the parent ===
id: d01478ef-a209-4f68-b12c-75e6bd538580
parent_conversation_id: c2f81e3b-ad02-4f7b-9e27-b29c04b00f5b
id: 509c8261-bae9-4d00-858d-2c2fa6db4bea
parent_conversation_id: c2f81e3b-ad02-4f7b-9e27-b29c04b00f5b
=== 3. parent lists both children ===
sub_conversation_ids: ['d01478ef-a209-4f68-b12c-75e6bd538580', '509c8261-bae9-4d00-858d-2c2fa6db4bea']
=== 4. unknown parent -> expect 422 ===
{"detail":"Parent conversation 00000000-0000-0000-0000-000000000001 not found"}
status: 422
=== 5. after full server restart: relationship intact ===
sub_conversation_ids: ['d01478ef-a209-4f68-b12c-75e6bd538580', '509c8261-bae9-4d00-858d-2c2fa6db4bea']
=== 6. delete parent ===
status: 200
=== 7. children survive with dangling parent pointer ===
id: d01478ef-a209-4f68-b12c-75e6bd538580
parent_conversation_id: c2f81e3b-ad02-4f7b-9e27-b29c04b00f5b
id: 509c8261-bae9-4d00-858d-2c2fa6db4bea
parent_conversation_id: c2f81e3b-ad02-4f7b-9e27-b29c04b00f5b
```

The child's `meta.json` on disk contains `"parent_conversation_id": "c2f81e3b-ad02-4f7b-9e27-b29c04b00f5b"`, confirming the durable link.

## Video/Screenshots

Not a GUI change; the end-to-end curl transcript above is the demonstration.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **TypeScript client:** `OpenHands/typescript-client` is hand-written (no OpenAPI codegen — its `package.json` has no generation script and its models are maintained manually), so the fields land in the generated schema here but need a small companion PR adding `parent_conversation_id` to the start-request type and both fields to `ConversationInfo` in `src/models/conversation.ts`. Field names were chosen to match what that client and Agent Canvas already type (`SharedConversation.parent_conversation_id`/`sub_conversation_ids`), so the companion change is purely additive.
- **Scope:** this covers UI-launched children (a top-level conversation naming a parent), which is what the Canvas acceptance criteria describe. Tool-spawned delegate/task sub-conversations are not registered in the server catalog today and remain out of scope — surfacing them is a substantially larger, separate piece of work.
- **Workspace check — known limitation, flagged for review.** `_same_workspace` uses resolved-path equality of `working_dir`, and the parent check runs *before* `_prepare_request_workspace`. Two consequences, both verified by hand:
  - A **child** started with `worktree=true` is **accepted**: at validation time `request.workspace` is still the original directory, and the worktree is created afterwards. (An earlier revision of this note claimed such a child would be rejected — that was wrong.)
  - A **parent** started with `worktree=true` stores the worktree directory as its workspace, so a child naming that parent with the original repo directory is rejected with a `422 ... belongs to a different workspace`.

  So the failing direction is a worktree *parent*, not a worktree *child*. Supporting it means comparing git repo roots instead of `working_dir`. Left unfixed pending a decision on whether children of worktree conversations are a flow Canvas actually uses — note it fails loudly rather than corrupting anything, and there is no persisted state to migrate if we fix it later.
- **Persistence format nit:** `StoredConversation.id` and `forked_from_conversation_id` are `OpenHandsUUID` (serialised as bare hex), while the inherited `parent_conversation_id` is a plain `UUID` (serialised dashed), so one `meta.json` encodes the same kind of value two ways. Both forms parse and round-trip, and everything inside the server compares `UUID` objects rather than strings, so this is cosmetic — recorded here so it is a deliberate choice rather than an oversight.
- No migration needed; two mock-based tests needed a `parent_conversation_id=None` added to their `MagicMock(stored=...)` field enumerations.


enyst (commented):
Two quick questions, if I may.

This seems a bit curious:
> Agent Canvas needs to launch child conversations against a local agent-server and reconstruct the parent/child tree after a reload, the way it already can against OpenHands Cloud.

Why does it work on the Cloud and not here, since on the Cloud too it’s via agent-server?

Perusing the PR code, I see we add only in agent-server a relationship. Isn’t this happening also via Python API? Or why does it work there if it does 🤔

VascoSch92:
> Two quick questions, if I may.
> 
> This seems a bit curious:
> 
> > Agent Canvas needs to launch child conversations against a local agent-server and reconstruct the parent/child tree after a reload, the way it already can against OpenHands Cloud.
> 
> Why does it work on the Cloud and not here, since on the Cloud too it’s via agent-server?
> 
> Perusing the PR code, I see we add only in agent-server a relationship. Isn’t this happening also via Python API? Or why does it work there if it does 🤔

 Ah sorry, the wording in the PR summary is not so good.

Let me explain and recap what is going on.

- Cloud does use agent-server to execute conversations, but the parent/child relationship is not currently owned by agent-server. 
- Agent Canvas creates Cloud conversations through /`api/v1/app-conversations`; the app-server validates the parent, stores `parent_conversation_id` in its conversation metadata database, and derives `sub_conversation_ids` from that database. 
- When it then starts the actual runtime conversation through agent-server’s `/api/conversations`, it does not forward the parent relationship.
- Local Canvas bypasses that app-server/database layer and talks directly to agent-server. Therefore, without this change, no local server-side component persists the relationship across reloads. 

This PR want to add the equivalent relationship to the local agent-server catalog.

Regarding the Python API: the field is added to the SDK’s StartConversationRequest because agent-server shares that request model, but the relationship is implemented by ConversationService. 

Another TypeScript client/Agent Canvas change is still required to forward `parent_conversation_id` on the local start request and map `parent_conversation_id`/`sub_conversation_ids` from local responses.

VascoSch92:
@openhands solve the conflicts

VascoSch92:
Merged the latest `main` into this branch and resolved the conflicts.

All conflicts were in `openhands-agent-server/openhands/agent_server/conversation_service.py`, caused by two upstream refactors landing on top of this PR:

- **`_start_conversation`** — upstream added the Codex credential resume/late-binding paths and restructured the early-return for existing conversations. I dropped the PR's old early-return block (superseded by upstream's structure) and re-applied the parent/child intent by passing `self._children_of(conversation_id)` to `_compose_conversation_info` on each resume return path.
- **`delete_conversation`** — upstream moved the `event_services.pop`/`_conversation_records.pop` calls to after the event service is closed. I kept upstream's ordering and relocated the orphan-on-parent-delete comment to the actual pop site.
- The new `_children_of`/`_children_index` helpers were kept alongside upstream's new `activate_credential_binding`/credential helpers.

Verified locally:
- `tests/agent_server/test_parent_child_conversations.py` → 11 passed
- `tests/agent_server/test_conversation_service.py`, `test_agent_launch_additions.py`, `test_agent_profile_conv_start.py` → 126 passed

The PR now shows as mergeable (`mergeable_state` is `blocked` only because required checks/reviews are still pending, not because of conflicts).

neubig (approved):
Looks good, thank you!

VascoSch92:
@openhands resolve the conflicts

VascoSch92:
Done — I merged the latest `main` into `vasco/4119-parent-child-conversations` and resolved the conflict. The PR now shows `mergeable: true` (no more conflicts).

## What conflicted
Since the branch last merged `main`, three commits landed upstream: #4180, #4193, and #4201. Of these, **#4201** (`perf(agent-server): index conversation execution status for search/count`) touched the same region of `openhands-agent-server/openhands/agent_server/conversation_service.py` as this PR. Everything else auto-merged cleanly.

The conflict was in `ConversationService._conversation_info`, in both the live-service and persisted-state branches:
- **This PR** added a `children` argument to `_compose_conversation_info(...)` so `sub_conversation_ids` gets populated.
- **#4201** added execution-status signature bookkeeping (`record.state_signature = None` in the live branch, `record.state_signature = signature` in the persisted branch).

The two changes are orthogonal, so I kept both — set `record.state_signature` and pass `children` to `_compose_conversation_info`.

## Verification
```
uv run pytest tests/agent_server/test_parent_child_conversations.py   # 11 passed
uv run pytest tests/agent_server/test_conversation_service.py tests/agent_server/test_conversation_router.py tests/agent_server/test_conversation_info_model.py tests/agent_server/test_parent_child_conversations.py   # 217 passed
```

PR #4193, opened 2026-07-22
fix: parallel tool metrics

### LINKED ISSUE #4189: Visualizer: parallel tool calls repeat the same per-request metrics under every ActionEvent of the batch
## Context

Follow-up to #4146 (per-request token usage in the metrics subtitle, fixing #4105),
raised by @VascoSch92 in the approving review:
https://github.com/OpenHands/software-agent-sdk/pull/4146#pullrequestreview-4751818504

## Problem

`_find_request_usage()` matches `event.llm_response_id` against
`TokenUsage.response_id`. When one LLM response contains multiple tool calls
(parallel tool execution), every resulting `ActionEvent` carries the same
`llm_response_id`, so each event's subtitle shows the identical per-request
usage:

Action 1 : ↑ input 12K (total 54K) • …
Action 2 : ↑ input 12K (total 54K) • …   ← same request, repeated
Action 3 : ↑ input 12K (total 54K) • …   ← same request, repeated

The numbers are correct (the request did use 12K input tokens), but the
repetition reads as three separate 12K requests. Totals are unaffected.

## Possible directions

- Render the per-request portion only on one event of the batch (e.g. the
  first), with the siblings falling back to the labeled totals-only format
  from #4146 (`↑ input (total 54K) • ...`).
- Keep the numbers on every event but mark them as shared, e.g.
  `↑ input 12K (shared, total 54K)` or `(request total 12K)`.
- Track already-rendered response_ids in the visualizer and vary the format
  accordingly.

The first option is probably the least invasive: the visualizer already has
both formats, and `ActionEvent` batches are identifiable via the shared
`llm_response_id` (see `event/base.py`, which already groups action events by
`llm_response_id`).

## Acceptance criteria

- A batch of N parallel tool calls does not render N identical per-request
  usage lines.
- Single-tool-call behavior from #4146 is unchanged.
- Full-string tests covering a parallel batch (N ≥ 2) on both
  `DefaultConversationVisualizer` and `DelegationVisualizer`.

I'm happy to implement this as a follow-up PR.

luciobaiocchi:
HUMAN:

This is a fix follow-up to #4146, raised by @VascoSch92 in the approving review.

---

AGENT:

## Why

The per-request metrics subtitle matches an event to its usage by `llm_response_id`.
When one LLM response emits multiple parallel tool calls, every resulting
`ActionEvent` shares that one `llm_response_id` and one `TokenUsage`, so the
identical per-request line was rendered under each sibling:

```
Action 1 : ↑ input 2K (total 3K) • ...
Action 2 : ↑ input 2K (total 3K) • ...   ← same request, repeated
Action 3 : ↑ input 2K (total 3K) • ...   ← same request, repeated
```

The numbers are correct but read as three separate 2K requests. Totals are
unaffected. Fixes #4189.

## Summary

- Only the **first** `ActionEvent` of a batch (same `llm_response_id`) shows the
  per-request numbers; siblings fall back to the labeled totals-only format from
  #4146 (`↑ input (total 3K) • …`).
- `_claim_batch_primary(event)` decides this in **O(1)** by tracking only the
  previous ActionEvent's `llm_response_id`. A batch's events always render
  consecutively (one LLM response emits all its tool calls before the next
  response exists), so a single remembered id is enough to tell a batch primary
  from its siblings — no scan of the event stream. It's named to signal that it
  mutates, and must be called once per rendered event.
- `_format_metrics_subtitle` gains a keyword-only `force_totals` flag; the two
  render entry points (`DefaultConversationVisualizer._create_event_block` and
  `DelegationVisualizer`'s action path) pass `force_totals=not _claim_batch_primary(event)`.
- Single tool calls and non-`ActionEvent`s are always their own primary, so
  their output is unchanged.

## Design decision: O(1) tracking vs. stateless scan

The first draft determined the batch primary **statelessly**, by scanning
`state.events` for the first `ActionEvent` sharing the response id. That was
idempotent and replay-safe, but it is **O(N) per rendered ActionEvent** — the
primary renders before it is appended to state, so the scan walks the whole
conversation to the end — i.e. **O(M²)** over an M-event trajectory, against a
disk-backed (cached) event store.

The human author (Lucio) preferred keeping the visualizer cheap on long
**benchmark / eval trajectories** and asked for the O(1) approach even though it
carries a little render-time state. We adopted the O(1) previous-`response_id`
tracking above. It is not fully idempotent (rendering the exact same event twice
in isolation would flip primary→sibling), but that is a test-only artifact: in
production each event streams through `on_event` exactly once and batches are
consecutive, and because it remembers only the immediately-preceding id, a full
replay still re-derives the correct result — so it stays replay-safe, strictly
better than a growing `set` of shown ids.

Two mitigations keep the small residual risk in check: the method name signals
its mutation, and `test_batch_action_events_are_emitted_consecutively`
(`tests/sdk/agent/test_response_dispatch.py`) pins the consecutive-emission
invariant the O(1) shortcut relies on — if a future change interleaves event
emission, that test fails before the metrics silently mislabel.

## Considered alternative (possible follow-up): mark the batch index on the event

The cleanest design would make the decision a pure function of the event —
record the tool-call's batch index on `ActionEvent` at construction (the agent
already computes it: `for i, tool_call in enumerate(message.tool_calls)` in
`response_dispatch.py`, branching on `i == 0`), so the visualizer would just read
`event.batch_index == 0`. That removes the mutation *and* the consecutive-emission
assumption while staying O(1), and a real batch index could also simplify the
batch regrouping in `events_to_messages`.

It is intentionally **not** done here: `Event` is `extra="forbid"`, so adding a
persisted field to a core event type carries the same backward-compat risk that
closed #4147 (older readers rejecting new events), and it touches the sync/async
construction paths, the ACP agent, and many tests. That is a data-model change
that deserves its own PR and maintainer sign-off on the event-schema/compat
policy, rather than riding along in a visualizer fix.

## Issue Number

Fixes #4189

## How to Test

```
uv run pytest tests/sdk/conversation/test_visualizer.py tests/tools/delegate/test_visualizer.py tests/sdk/agent/test_response_dispatch.py -v
```

Four new tests: `_claim_batch_primary` tracked across consecutive batches;
end-to-end `_create_event_block` assertions on the default and delegate
visualizers (primary shows per-request, siblings show totals-only); and
`test_batch_action_events_are_emitted_consecutively`, which pins the emission
invariant the O(1) detection relies on. All 43 visualizer tests and the full
`tests/sdk/conversation` + `tests/tools/delegate` + `tests/sdk/agent/test_response_dispatch`
suites pass; `ruff` and `pyright` clean.

## Video/Screenshots

Simulating the real callback order (each event renders before it is appended to
state), a 3-way parallel batch after a prior request (running total 3K/300):

```
c1 (primary): Tokens: ↑ input 2K (total 3K) • cache hit 0.00% (total 0.00%) • ↓ output 200 (total 300) • $ (total 0.00)
c2 (sibling): Tokens: ↑ input (total 3K) • cache hit (total 0.00%) • ↓ output (total 300) • $ (total 0.00)
c3 (sibling): Tokens: ↑ input (total 3K) • cache hit (total 0.00%) • ↓ output (total 300) • $ (total 0.00)
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Stacked on the #4146 branch so the diff shows only the #4189 change. Now that
  #4146 has merged to `main`, this should be retargeted to `main` when opened
  upstream against `OpenHands/software-agent-sdk`.


openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -283,6 +283,10 @@ def __init__(
         self._console = _create_console()
         self._skip_user_messages = skip_user_messages
         self._highlight_patterns = highlight_regex or {}
+        # Response id of the last ActionEvent rendered with metrics. Parallel
+        # tool calls share one llm_response_id, so only the first event of a
+        # batch is its primary; see ``_is_batch_primary``.
```
VascoSch92: where is the method _is_batch_primary?

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
```diff
@@ -399,7 +410,38 @@ def _find_request_usage(self, event: Event | None) -> TokenUsage | None:
             None,
         )
 
-    def _format_metrics_subtitle(self, event: Event | None = None) -> str | None:
+    def _claim_batch_primary(self, event: Event | None) -> bool:
+        """Claim per-request metrics for ``event``, returning whether it won.
```
VascoSch92: Make the comment more coincise. This is so verbose.

openhands-sdk/openhands/sdk/conversation/visualizer/default.py
luciobaiocchi: Is a dangling reference, i will remove it

VascoSch92 (approved):

VascoSch92:
@luciobaiocchi Thanks :-D

PR #4194, opened 2026-07-22
fix(sdk): pick up a user message that arrives during an async step

### LINKED ISSUE #15432: [Bug]: messages sent during streaming are not always seen by the agent
### Operating System

macOS

### Installation Method

npm (`npx @openhands/agent-canvas` or `npm install -g @openhands/agent-canvas`)

### Agent Canvas Version

1.5.2

### Bug Description

Image

i sent a message just as the agent was finishing streaming. the streaming finished, and the agent stopped working. it didn't react to my message at all--I had to resend it

### Steps to Reproduce

send a message while the agent is streaming its last notes on what it did

### Actual Behavior

agent doesn't react to new message

### Expected Behavior

agent picks up again and starts working

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

VascoSch92:
HUMAN:

Fixes a problem where the messages sent during streaming are not always seen by the agent.

---

AGENT:

## Why

Fixes OpenHands/OpenHands#15432 — *"messages sent during streaming are not always seen by the agent"*. The reporter sent a message just as the agent was finishing streaming; the stream completed, the agent stopped, and the message was never acted upon. It had to be resent.

The exact ordering:

1. The agent is in its **final** `astep`. Status is `RUNNING`. `Agent.astep` wraps the LLM call in `conversation._released_state_lock_during_io()` (added in #4009), so the state lock is **free for the whole streamed response** — seconds.
2. The message arrives and `send_message` runs on a worker thread. It takes the free lock, sees status `RUNNING` — **not** `FINISHED`/`STUCK` — so it does *not* rewind to `IDLE`. It appends the `MessageEvent` and nothing else records the arrival.
3. Streaming completes; the agent sets `FINISHED`.
4. `arun()`'s loop-top check sees `FINISHED` and `break`s, **without re-scanning for user messages that arrived mid-step**.
5. `EventService`'s `_rerun_requested` recovery doesn't save it either: `should_restart` requires `status == IDLE`, and we are `FINISHED`.

The message is persisted, so it silently joins the *next* turn's context — which is why resending appears to work and why the text is never actually lost, only unacted-upon.

**Sync `run()` never had this gap**, and not by accident. It holds the state lock across the whole step, so `send_message` blocks on the FIFO lock, always lands *after* `FINISHED` is set, rewinds to `IDLE`, and the loop keeps going. `local_conversation.py` says so explicitly:

> `# Note: We intentionally do NOT check for FINISHED status here.`
> `# This allows concurrent user messages to be processed... Without this design, concurrent messages would be lost.`

Releasing the lock in the async path for responsiveness silently invalidated the precondition that design depends on. This is an async-path regression, not a pre-existing race.

## Summary

- `arun()` now rescans for a newer user message before honouring `FINISHED`, and resumes the run if one arrived — mirroring what the ACP branch already does a few lines below.
- On the last allowed iteration it leaves the conversation `IDLE` instead, so a follow-up run picks the message up rather than exceeding `max_iteration_per_run`.
- Adds `_latest_user_message_id()`, a small helper used to notice a message that landed while the step held no lock.

The window itself is left open on purpose — releasing the lock is what keeps the server responsive during long completions (#4009). This notices the message *after* the fact rather than re-serialising the step.

## Issue Number

OpenHands/OpenHands#15432

## How to Test

```bash
uv run pytest tests/sdk/agent/test_message_during_streaming_arun.py -v
```

Both tests in the new file are deterministic — no sleeps-and-hope. The sender thread is gated on `FIFOLock._waiters`, and `FIFOLock.acquire()` sends any later acquirer to the back of the queue, so once the sender is enqueued the run loop provably cannot overtake it.

**`test_message_during_streaming_is_acted_upon`** drives the real `arun()`. Probes confirm the race genuinely occurred rather than being sidestepped: `locked_during_call is False`, and `status_during_call == status_after_inject == RUNNING` (so `send_message`'s terminal-status rewind never fired). The decisive assertion is that the second LLM round-trip **carried the message to the model** — `_saw(llm._calls, "second message") == [1]`, 2 calls total.

**`test_sync_run_absorbs_message_sent_during_step`** is the control: same harness against sync `run()`, `locked_during_call is True`, 2 calls, message carried. It passes both before and after, pinning that this fix makes async match sync rather than changing sync.

Fails cleanly without the fix (`git stash` the source file):

```
tests/sdk/agent/test_message_during_streaming_arun.py::test_message_during_streaming_is_acted_upon FAILED
E   AssertionError: expected the message to be picked up (2 LLM calls), got 1
E   assert 1 == 2
tests/sdk/agent/test_message_during_streaming_arun.py::test_sync_run_absorbs_message_sent_during_step PASSED
```

Regression run:

```bash
uv run pytest tests/sdk/agent/ tests/sdk/conversation/ -q     # 1517 passed
uv run ruff check openhands-sdk                               # All checks passed!
uv run pyright openhands-sdk/.../local_conversation.py        # 0 errors, 0 warnings
```

End-to-end against the UI: run agent-canvas against this SDK, ask for a long tool-free reply (`Without using any tools, write a ~800-word essay on the history of computing.`), and send a second message while it streams. Before: the agent finishes and stops, ignoring it. After: the agent continues and acts on it without a resend.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**Scope caveat:** the tests exercise `Agent`/`LocalConversation` directly. The threaded `send_message` faithfully mimics the agent-server's `run_in_executor` dispatch, but end-to-end coverage of the server layer would be a separate test.

**Alternative considered:** widening `should_restart` in `event_service.py` to accept `FINISHED` is a smaller, server-only change. I preferred fixing `arun()` because it keeps the turn inside a single run rather than tearing down and restarting one, and because it puts the async path back in line with the documented sync behaviour instead of adding a second recovery mechanism on top.

The UI-side half of this report (agent-canvas#1899 — the same trigger splitting the streamed reply in the chat) is fixed separately in OpenHands/agent-canvas#1901.


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -2202,17 +2212,67 @@ async def arun(self) -> None:
...
+                            if iteration >= self.max_iteration_per_run:
+                                logger.info(
+                                    "User message arrived during final iteration; "
+                                    "leaving conversation idle for a follow-up run"
+                                )
+                                self._state.execution_status = (
+                                    ConversationExecutionStatus.IDLE
+                                )
+                                break
+                            if not step_awaiting_confirmation:
```
enyst, line 2260:
```suggestion
                            if step_finished:
```

this is `if step_finished`, right?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -2202,17 +2212,67 @@ async def arun(self) -> None:
...
+                        step_finished = (
+                            self._state.execution_status
+                            == ConversationExecutionStatus.FINISHED
+                        )
+                        step_awaiting_confirmation = (
+                            self._state.execution_status
+                            == ConversationExecutionStatus.WAITING_FOR_CONFIRMATION
+                        )
+                        new_message_arrived = (
+                            self._state.last_user_message_id != last_user_message_id
```
enyst, line 2232: You know, I wonder if we could make a queue, just like we have the `pending` ones, and check at the end of a step rather than in the middle of it… Idk, I’m just trying to think how the code would be more understandable - that is, how we could vouch for correctness easier.

I’m fine with this though. I think maybe we could document the behavior (and the agent should access that doc easily). Maybe that’s what comments do here, that’s nice… 🤔

enyst (approved):

Issue #4177, opened 2026-07-21
Phase 1: land Codex credential write-through in two PRs


simonrosenberg:
Parent: OpenHands/OpenHands#4170
Supersedes: OpenHands/OpenHands#4174

## Goal

Fix sequential Codex credential staleness in exactly two landing PRs across all supported ACP deployment topologies:

1. direct/local Agent Server, including Agent Canvas;
2. OpenHands SaaS with remote runtimes;
3. the self-hosted OpenHands app server with per-conversation Docker runtimes.

Conversation A must be able to start from `CODEX_AUTH_JSON = R0`, let Codex rotate its private `auth.json` to `R1`, persist `R1` to the topology's canonical store, and let a later isolated conversation B start from `R1`.

The two landing PRs are:

1. software-agent-sdk OpenHands/OpenHands#4124: the reusable versioned binding, ACP file-credential lifecycle, Codex synchronizer, runtime activation support, and direct/local binding.
2. OpenHands/OpenHands#15287: the shared app-server credential endpoint, SaaS and OSS canonical-store adapters, and start/resume activation.

This plan intentionally does not implement a generic credential broker or the concurrent refresh-owner behavior in OpenHands/OpenHands#4171.

## Decision

Use one small provider-neutral SDK contract, with Codex as its first implementation:

```python
@dataclass(frozen=True)
class ResolvedCredential:
    value: str
    version: str

class VersionedCredentialBinding(Protocol):
    async def load(self) -> ResolvedCredential: ...
    async def replace(self, expected_version: str, value: str) -> str: ...
```

PR 1 supplies two implementations consumed by the same Codex lifecycle:

- `LocalVersionedCredentialBinding` for a direct Agent Server's local secrets store.
- `HttpVersionedCredentialBinding` for an Agent Server managed by either the SaaS or OSS OpenHands app server.

PR 2 does not introduce another synchronizer. Its shared HTTP route delegates to canonical-store adapters implementing the same `VersionedCredentialBinding` semantics:

- an encrypted SQL adapter for SaaS;
- a locked, generation-backed file-store adapter for the OSS app server.

| Topology | Runtime binding from PR 1 | Canonical adapter |
| --- | --- | --- |
| Direct Agent Server / Agent Canvas | Local binding | Agent Server `FileSecretsStore` |
| OpenHands SaaS | HTTP binding | SaaS encrypted SQL store |
[...]

Issue #4170, opened 2026-07-21
Phase 1: propagate rotated Codex credentials across sequential ACP conversations


simonrosenberg:
Parent tracker: OpenHands/OpenHands#4169
Implementation plan: OpenHands/OpenHands#4177

## Goal

Fix the Codex credential-staleness bug observed in production for sequential ACP conversations. When Codex rotates a conversation-local `auth.json`, the new credential must reach the canonical credential store before a later isolated runtime starts.

## Problem

A Codex ACP conversation can start from credential generation R0 and rotate its writable `auth.json` to R1. Today, R1 can remain only in that conversation's working copy while the canonical `CODEX_AUTH_JSON` remains R0. A later conversation then materializes the invalidated refresh token again.

This happened in SaaS on 2026-07-15 and was originally tracked in OpenHands/OpenHands#4121. The failed runtime's Codex log reported `refresh_token_reused`, while the user-visible ACP error was only `Invalid params`. The same bug can affect any deployment that gives isolated conversations independent copies of an unchanged canonical credential.

## Production evidence

A read-only production investigation covering 2026-06-30 through 2026-07-21 found about 166 initial Codex credential materializations across 164 runtime pods and 17 matching `Invalid params` failures across 15 runtimes. Every failure occurred approximately 600 seconds after ChatGPT authentication began. Immediate retries that preserved the same conversation-local credential through seed-if-absent failed again. No corresponding image-pull, scheduling, restart, OOM, routing, or pod-health failure was found.

The originally reported runtime had a provider-confirmed `refresh_token_reused` response in its runtime-local Codex log. Centralized logs did not retain that provider reason, so the other failures are strongly consistent with the same stale-credential mechanism but are not independently confirmed. Simultaneous failure clusters also did not show one successful refresh owner followed by stale losers, so they do not establish the concurrent-refresh race tracked by OpenHands/OpenHands#4171. Phase 1 therefore includes sanitized authentication classification and credential-generation/writeback telemetry.

## Implementation plan

Implement a versioned write-through binding between the canonical `CODEX_AUTH_JSON` and the `auth.json` owned and rewritten by Codex. The canonical version returned when binding is the synchronization generation. A file digest remains local to the runtime and is used only to detect changes; it is not part of the remote credential interface.

### Bind before launch

- Acquire read/write authorization scoped to the user and Codex credential.
- Resolve the canonical credential together with its version.
[...]

PR #4124, opened 2026-07-15
fix(acp): persist rotated Codex credentials

simonrosenberg:
HUMAN:

Treat Cloud-backed CODEX_AUTH_JSON as authoritative and materialize it into a per-conversation writable working copy.

---

AGENT:

## Why

Codex rotates ChatGPT subscription credentials in `auth.json`. Static per-conversation copies can leave later conversations with an invalidated refresh token, while shared writable copies allow concurrent runtimes to overwrite each other.

This PR implements the SDK half of OpenHands/OpenHands#4177 and the Phase 1 contract in OpenHands/OpenHands#4170.

## Summary

- Add the provider-neutral async `VersionedCredentialBinding` contract with opaque canonical versions and HTTP/local implementations.
- Keep local secret versioning lazy and provider-neutral; the lifecycle registry alone selects the Codex implementation.
- Activate credential bindings through a runtime-only Agent Server endpoint advertised by `credential_binding_v1`; binding values and transport headers stay out of durable conversation snapshots.
- Materialize Codex auth into a private temporary `CODEX_HOME`, monitor valid stable rewrites, retain every observed value for masking, and write rotations through with compare-and-swap.
- Treat conflicts and synchronization failures as sticky/fail-closed, including before output masking.
- Reconcile ambiguous writes by reloading once and accepting only an exact value match.
- Stop the ACP process before the final read/write, then discard the runtime auth directory and activation state.
- Back direct/local Agent Server conversations with generation-versioned, file-locked `FileSecretsStore` updates.
- Strip legacy `CODEX_AUTH_JSON` copies from new and resumed durable conversation state when a managed binding is available.
## Issue Number

Implements PR 1 from OpenHands/OpenHands#4177 and addresses the Phase 1 SDK work in OpenHands/OpenHands#4170 under umbrella OpenHands/OpenHands#4169.

Companion OpenHands PR: https://github.com/OpenHands/OpenHands/pull/15287

## How to Test

```bash
UV_NO_CONFIG=1 UV_FROZEN=1 uv run pytest -q --timeout=30 \
  tests/sdk/agent/test_acp_file_credentials.py \
  tests/sdk/test_credential.py \
  tests/agent_server/test_credential_binding.py
# 35 passed

UV_NO_CONFIG=1 UV_FROZEN=1 uv run pytest -q \
  tests/agent_server/test_conversation_service.py \
  tests/sdk/agent/test_acp_agent.py \
  tests/sdk/conversation/test_switch_model.py \
[...]

PR #4198, opened 2026-07-22
fix(agent-server): require credential reactivation before cold load

simonrosenberg:
HUMAN:

require credential reactivation before cold load

---

AGENT:

## Why

OpenHands must distinguish servers that merely accept a credential binding from servers that probe it before returning 204. The existing `credential_binding_v1` marker identifies binding support; the narrower `credential_binding_readiness_probe_v1` marker identifies that readiness contract without changing older servers.

Managed conversations also need a cold-load invariant across pause and restart. The exact `credential_binding_activation_guard_v1` capability identifies servers that persist a nonsecret binding requirement before scrubbing plaintext and refuse hydration until the runtime binding is reactivated.

## Summary

- Retain `credential_binding_v1` and advertise exact readiness and cold-load guard capabilities.
- Probe the versioned credential source before accepting activation; treat malformed URLs as terminal 422 responses and unsupported sources as terminal 501 responses without retries.
- Persist only the required binding name before scrubbing plaintext, never a URL, token, header, version, or credential value.
- Reject unbound cold REST hydration with retryable 409 and WebSocket hydration with close code 1013 while keeping catalog GETs cold.
- Let deletion bypass reactivation only for immediate teardown, so a revoked callback cannot make a conversation undeletable.
- Validate replacement authorization against the canonical credential before changing live headers.
- Drain active services concurrently before sandbox pause and retain failed closes for retry.
- Preserve one-off plaintext conversations and local credential bindings without the managed reactivation requirement.
## Issue Number

Implements the Agent Server follow-up required by #4170 and the two-PR landing plan in #4177, building on merged #4124.

Companion OpenHands PR: OpenHands/OpenHands#15287

## How to Test

```bash
UV_NO_CONFIG=1 UV_FROZEN=1 uv run pytest -q --timeout=30 \
  tests/agent_server/test_credential_binding.py \
  tests/agent_server/test_conversation_service.py \
  tests/agent_server/test_server_details_router.py \
  tests/agent_server/test_unhandled_exception_error_id.py \
  tests/sdk/test_credential.py

UV_NO_CONFIG=1 UV_FROZEN=1 uv run pre-commit run --all-files
```

## Validation

On head `e13a2aecb`:

- 142 focused SDK and Agent Server tests pass, covering cold deletion, failed reauthorization atomicity, invalid URL classification, concurrent pause draining, cold restart, pause/reactivation, unsupported topology, and one-off credentials.
- All code and build CI is green: SDK, Agent Server, cross, stress, workspace, tools, Windows, OpenAPI, persisted settings, binary, and multi-architecture image suites.
- Changed-file pre-commit passes, including Ruff, Pyright, pycodestyle, and import dependency checks.
- Automated functional QA passes against live local Agent Server HTTP and WebSocket endpoints with no issues found.

## Video/Screenshots

Not applicable; this is an internal backend protocol simplification.

## Type

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The companion OpenHands change requires both `credential_binding_readiness_probe_v1` and `credential_binding_activation_guard_v1` before removing plaintext. Positively identified legacy or initially unsupported topologies keep compatibility plaintext; a marked managed resume fails closed until exact-204 reactivation succeeds.

The cold-load guard remains centralized for runtime operations. Deletion alone uses a narrow teardown path that cannot leave an open unbound service.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:181b389-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-181b389-python \
  ghcr.io/openhands/agent-server:181b389-python
```

```
ghcr.io/openhands/agent-server:181b389-golang-amd64
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-golang-amd64
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-golang-amd64
ghcr.io/openhands/agent-server:181b389-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:181b389-golang-arm64
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-golang-arm64
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-golang-arm64
ghcr.io/openhands/agent-server:181b389-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:181b389-java-amd64
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-java-amd64
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-java-amd64
ghcr.io/openhands/agent-server:181b389-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:181b389-java-arm64
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-java-arm64
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-java-arm64
ghcr.io/openhands/agent-server:181b389-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:181b389-python-amd64
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-python-amd64
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-python-amd64
ghcr.io/openhands/agent-server:181b389-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:181b389-python-arm64
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-python-arm64
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-python-arm64
ghcr.io/openhands/agent-server:181b389-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:181b389-golang
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-golang
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-golang
ghcr.io/openhands/agent-server:181b389-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:181b389-java
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-java
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-java
ghcr.io/openhands/agent-server:181b389-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:181b389-python
ghcr.io/openhands/agent-server:181b389d9ae827dc5771145ee2b842663032f49a-python
ghcr.io/openhands/agent-server:simplify-credential-binding-probe-python
ghcr.io/openhands/agent-server:181b389-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `181b389-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `181b389-python-amd64`) are also available if needed


simonrosenberg:
## Code review — activation-as-readiness-probe

Max-effort multi-agent review of `e102cecb` (SDK #4198) alongside `30be0987` (OpenHands #15287). Two blockers on the new probe live here; a third finding is cross-PR (the capability-flag removal) and is detailed on OpenHands #15287.

### Blockers (confirmed)

**1. `httpx.InvalidURL` escapes both handlers → unhandled 500 instead of 502** — `openhands-agent-server/openhands/agent_server/credential_binding.py:99`
`httpx.InvalidURL` subclasses `Exception`, not `RequestError`, so the new `await binding.load()` probe escapes both the `except httpx.RequestError` inside `HttpVersionedCredentialBinding.load` and the route's `except CredentialBindingError`. A misconfigured `WEB_URL` (unbracketed IPv6, bad port, non-IDNA host — empirically `http://[::1/x`, `http://example.com:abc/x` all raise `InvalidURL` and are *not* `RequestError`s) makes the `PUT /credential-bindings` endpoint crash with an unhandled traceback and HTTP 500. That breaks the 502 activation-as-probe contract this PR introduces: 500 is indistinguishable from an agent-server bug in monitoring, and the caller's 404-vs-error legacy detection is muddied. Catch `InvalidURL` (or broaden to the intended base) and map to 502.

**2. One-shot probe permanently disables sync on a transient blip** — `openhands-agent-server/openhands/agent_server/credential_binding.py:99`
The probe is a single `binding.load()` — one attempt, default 5s timeout, no retry — and OpenHands treats any non-404 activation failure as a permanent plaintext fallback for that conversation. A brief app-server unavailability at start (rolling deploy, LB blip, cold DNS) → `CredentialSyncError` → route 502 → OpenHands keeps plaintext and never re-attempts. The runtime then rotates `auth.json` against its local store only, the canonical store silently goes stale, and the user's next conversation on another runtime starts on the superseded refresh token → Codex 401 → forced re-login — the exact #4170 symptom this feature targets. Add a bounded retry/backoff on the probe.

### Cross-PR (detailed on OpenHands #15287)

**Removing the `credential_binding_v1` capability leaves no version signal** — `openhands-agent-server/openhands/agent_server/server_details_router.py:60`
Dropping the capability advertisement means OpenHands can't distinguish a #4198 server (does the pre-load probe) from a #4124-only server (activates with 204, no probe). Since the released fleet is #4124-only, OpenHands strips the plaintext fallback on a 204 that never proved reachability → codex launch fails with "please authenticate again" on a valid credential. Consider keeping (or replacing) a capability marker so the app server can gate the plaintext strip. Full failure chain is in OpenHands #15287 finding 1.

---
*Automated review — same run as OpenHands #15287. Findings independently verified against the pinned head.*

simonrosenberg:
Addressed both blockers and the cross-PR version signal in `b17bf4d95`.

- Map `httpx.InvalidURL` to the activation endpoint's 502 contract.
- Retry the readiness load up to three times with bounded backoff for transient sync failures only; authorization, conflict, and reauthentication failures remain immediate.
- Retain `credential_binding_v1` and advertise the exact new `credential_binding_readiness_probe_v1` capability consumed by OpenHands #15287 before it strips plaintext or transmits a binding token.

Verification on this head:
- Focused agent-server suite: 29 passed
- Changed-file pre-commit hooks: passed

simonrosenberg:
QA validated the activation probe over real HTTP. The reported issue was a stale PR-description claim from before the cross-PR review; the AGENT section now documents the retained `credential_binding_v1` marker and the narrower `credential_binding_readiness_probe_v1` version gate. The protected HUMAN section was left unchanged.

openhands-agent-server/openhands/agent_server/server_details_router.py
```diff
@@ -58,7 +58,12 @@ class ServerInfo(BaseModel):
     runtime_idle_timeout_seconds: float | None = Field(
         default_factory=lambda: get_runtime_idle_timeout_seconds()
     )
-    capabilities: list[str] = Field(default_factory=lambda: ["credential_binding_v1"])
+    capabilities: list[str] = Field(
+        default_factory=lambda: [
+            "credential_binding_v1",
+            "credential_binding_readiness_probe_v1",
```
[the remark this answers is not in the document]
simonrosenberg: Addressed in the PR description. The cross-PR version-skew review showed activation alone cannot distinguish a released #4124 server that returns 204 without probing from this stronger endpoint. We therefore retain `credential_binding_v1` and use the narrower `credential_binding_readiness_probe_v1` marker so OpenHands only strips plaintext after a server that advertises and performs the readiness load returns 204.

simonrosenberg:
Addressed the latest review findings in c5bf1ee33: terminal probe responses are preserved, the callback path is exercised over real HTTP, and Docker pause now has a final SDK drain barrier. Ready for another review.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -900,6 +917,15 @@ async def _get_or_load_event_service_locked(
...
             return None
 
+        pending_bindings = self._credential_bindings.get(conversation_id, {})
+        missing_bindings = (
+            record.stored.required_runtime_credential_bindings - pending_bindings.keys()
+        )
+        if missing_bindings:
+            raise CredentialBindingActivationRequired(
+                "credential_binding_activation_required"
+            )
```
VascoSch92, line 946: This guard sits lower than the paths it's meant to protect. `delete_conversation` (L1493), `pause_conversation` (L1458) and `interrupt_conversation` (L1476) all go through `_get_or_load_event_service[_locked]`, so they start raising here too — that's a behaviour change for three endpoints that have nothing to do with needing a live credential.

Delete is the one I'd want to fix before this lands. Once `required_runtime_credential_bindings` is on disk, a cold conversation can only be deleted after a binding is reactivated, and reactivation only succeeds if the callback answers 200. So in the case where you most want to clean up — the user revoked their auth and the source now 404s — you're stuck:

```
PUT  .../credential-bindings/CODEX_AUTH_JSON  -> 422   (probe got 404, binding not installed)
DELETE /api/conversations/{id}                -> 409   {"detail": "credential_binding_activation_required", "retryable": true}
```

`"retryable": true` is misleading on a DELETE — repeating it never succeeds, the state it's waiting on can only be changed by a different call. There is a way out, but only by pointing the binding at any endpoint that returns a well-formed `{"value", "version"}` payload:

```
PUT  .../credential-bindings/CODEX_AUTH_JSON  -> 204   (stub source)
DELETE /api/conversations/{id}                -> 200
```

That works, but standing up a dummy credential endpoint to delete a conversation isn't something an operator would ever guess from a 409.

It also compounds with the set being union-only — `event_service.py:204` is the only write and there's no `discard`/`remove` anywhere, so once a name is recorded the requirement is attached to that conversation for good, even if it later moves to a one-off plaintext secret.

Could the check move up to the callers that actually need the credential (run / send_message / the events socket) rather than sitting in the shared hydration path? Teardown and lifecycle calls don't need a working binding, and letting them through would keep the cold-load invariant intact where it matters while leaving conversations recoverable.

Repro for the delete case, if useful:

```python
async with ConversationService(conversations_dir=d) as svc:
    await svc.activate_credential_binding(
        cid, "CODEX_AUTH_JSON",
        HttpVersionedCredentialBinding("https://app.test/c", {"A": "B"}),
    )
    await svc.start_conversation(req)

async with ConversationService(conversations_dir=d) as restarted:
    assert await restarted.get_conversation(cid) is not None   # catalog GET is fine
    await restarted.delete_conversation(cid)                   # CredentialBindingActivationRequired
```

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -193,6 +193,21 @@ async def _scrub_persisted_credentials(self) -> None:
...
+        }
+        if required_bindings:
+            self.stored = self.stored.model_copy(
+                update={
+                    "required_runtime_credential_bindings": (
+                        self.stored.required_runtime_credential_bindings
+                        | required_bindings
+                    )
+                }
+            )
```
VascoSch92, line 219: This is the only write to `required_runtime_credential_bindings` and it's union-only — no `discard`/`remove`/difference anywhere in the tree. So once a name lands here it stays for the life of the conversation, including after it moves to a one-off plaintext secret or a local binding, and the single reader at `conversation_service.py:922` keeps gating hydration on it.

Is monotonic the intent? If so it's worth saying in a comment, since nothing in the type hints at it. If not, reactivation-time reconciliation would be the natural place to drop names that no longer apply.

Minor, same block: `set[str]` serialises to an unordered list, so `meta.json` will churn on every save once there's more than one name in it.

openhands-agent-server/openhands/agent_server/credential_binding.py
```diff
@@ -90,18 +122,49 @@ async def activate_credential_binding(
...
+        ) from exc
+    except (
+        CredentialInvalidResponse,
+        CredentialNeedsReauthentication,
+        httpx.InvalidURL,
+    ) as exc:
+        raise HTTPException(
+            status.HTTP_422_UNPROCESSABLE_ENTITY,
+            "Credential binding source is invalid",
+        ) from exc
```
VascoSch92, line 163: This 422 can be returned after the binding has already been swapped. On the reauthorize path (`event_service.py:271`) `existing.reauthorize(binding)` mutates the headers in place, and only then does `_scrub_persisted_credentials()` run — which can raise `CredentialNeedsReauthentication` from the `is_valid_codex_auth` check when the legacy auth file is present and the source returns something unusable. That lands here as a 422, so the caller is told activation failed while the new authorization is live:

```
existing.headers                -> {'A': 'new'}
existing.authorization_revision -> 1
```

The ordering predates this PR, but the mapping is what makes it bite: previously that path fell out as a 500 nobody acted on, now it's a 422 the control plane will read as "binding not installed" and retry or fall back to plaintext. Scrubbing first and only calling `reauthorize` once it's committed would make the failure atomic.

openhands-agent-server/openhands/agent_server/credential_binding.py
```diff
@@ -76,6 +81,33 @@ async def replace(self, expected_version: str, value: str) -> str:
 
 
 router = APIRouter(prefix="/conversations", tags=["Credential Bindings"])
+_PROBE_ATTEMPTS = 3
+_PROBE_INITIAL_DELAY = 0.25
+
```
VascoSch92, line 86: From typing import Final and use the correct type here

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -685,6 +689,19 @@ async def activate_credential_binding(
                 binding
             )
 
+    async def prepare_for_sandbox_pause(self) -> None:
+        async with self._lifecycle_lock:
+            event_services = self._event_services
+            if event_services is None:
+                raise ValueError("inactive_service")
+            for conversation_id, event_service in tuple(event_services.items()):
```
VascoSch92: This drains conversations serially, while `EventService.close()` can wait up to 10s per active run. The linked OpenHands client uses a 15s default timeout, so two slow conversations can time out before the required 204. Could these closes run concurrently or use a barrier-wide deadline?

openhands-agent-server/openhands/agent_server/credential_binding.py
```diff
@@ -90,18 +122,49 @@ async def activate_credential_binding(
...
+        ) from exc
+    except CredentialBindingUnsupported as exc:
+        raise HTTPException(
+            status.HTTP_501_NOT_IMPLEMENTED,
+            "Credential binding source is unsupported",
+        ) from exc
+    except (
+        CredentialInvalidResponse,
+        CredentialNeedsReauthentication,
+        httpx.InvalidURL,
```
VascoSch92: `UnsupportedProtocol` (for example `ftp://...` or a relative URL) is wrapped by `load()` as `CredentialSyncError`, so it is retried three times and returned as 502 instead of this 422 invalid-source response. Could those URL shapes be classified as invalid too?

openhands-agent-server/openhands/agent_server/conversation_service.py
simonrosenberg: Fixed across both PRs. e13a2aecb starts all active service closes concurrently, removes successful services, and retains failures for retry. OpenHands d9321bc6d gives the pause-barrier request an explicit 30-second timeout instead of the shared 15-second default. The regression requires both closes to have started before either may finish.

openhands-agent-server/openhands/agent_server/credential_binding.py
simonrosenberg: Fixed in e13a2aecb: both probe constants now use `Final` with their concrete value types.

openhands-agent-server/openhands/agent_server/event_service.py
simonrosenberg: Monotonic is intentional. I added a concise invariant at the union: once the durable credential has been scrubbed, the conversation must not silently fall back to it later. A local binding still satisfies the recorded name, while a one-off plaintext value must not downgrade callback authority. `CODEX_AUTH_JSON` is currently the only supported file binding, so there is no multi-name serialization churn to solve yet.

openhands-agent-server/openhands/agent_server/credential_binding.py
simonrosenberg: Fixed in e13a2aecb. Both credential `load` and `replace` normalize `InvalidURL` and `UnsupportedProtocol` to `CredentialInvalidResponse`, so activation returns 422 without retrying. Coverage includes FTP, relative, and malformed-port URLs.

openhands-agent-server/openhands/agent_server/conversation_service.py
simonrosenberg: Fixed in e13a2aecb. Delete now uses the existing lease-aware hydrator with `require_runtime_bindings=False`; normal hydration, resume, pause, and interrupt keep the centralized guard. The bypass is delete-only so it cannot leave an open unbound service that would bypass later checks. A cold-restart regression verifies ordinary hydration remains blocked while deletion succeeds without reactivation.

openhands-agent-server/openhands/agent_server/credential_binding.py
simonrosenberg: Fixed in e13a2aecb. Reauthorization now validates and commits the scrub against a staged replacement mapping before mutating the active binding headers. URL mismatch is rejected up front, and a failed canonical preflight leaves both headers and authorization revision unchanged. Added a regression for the 422 path.

VascoSch92 (approved):

PR #3905, opened 2026-06-29
Optimize conversation search and count

enyst:
HUMAN:

---

AGENT:
_This PR was created by an AI agent (OpenHands) on behalf of the user._

## Why

Fixes #3142. The issue is still real on the current codebase: `_search_conversations` was still iterating every `EventService`, calling `get_state()` on each one, sorting the full list, then paginating. `_count_conversations` similarly loaded every conversation state.

## Summary

- Add an incremental in-memory conversation metadata index keyed by created time, updated time, and execution status.
- Use the index for search pagination and count APIs, loading full conversation state only for returned page items.
- Keep the index updated on start/resume, state status updates, activity timestamp changes, metadata updates, auto-title updates, and deletion.
- Add a regression test proving search/count do not call `get_state()` for off-page conversations once the index is populated.

## Issue Number

Fixes #3142

## How to Test

Commands run successfully:

```bash
uv run ruff format openhands-agent-server/openhands/agent_server/conversation_service.py tests/agent_server/test_conversation_service.py
uv run pytest tests/agent_server/test_conversation_service.py::TestConversationServiceSearchConversations::test_search_and_count_use_metadata_index -vv
uv run pytest tests/agent_server/test_conversation_service.py
uv run pre-commit run --files openhands-agent-server/openhands/agent_server/conversation_service.py tests/agent_server/test_conversation_service.py
```

Observed results:

- New regression test: `1 passed`
- Full conversation service test file: `78 passed`
- Pre-commit: Ruff format, Ruff lint, pycodestyle, pyright, import dependency rules, and tool registration checks all passed.

Reproduction/evidence note: the new regression seeds five conversations in the metadata index, requests a two-item first page, and asserts that only those two page items call `get_state()`; it also asserts indexed count calls do not add any `get_state()` calls.

## Video/Screenshots

Not applicable; this is an agent-server backend performance fix with automated test coverage.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The search path still reconciles missing metadata if private test/setup code mutates `_event_services` directly, but the normal service lifecycle maintains the index incrementally.

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/771d2e54-aa6d-4e42-8723-54a42f6159bc)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
[...]

PR #4201, opened 2026-07-23
perf(agent-server): index conversation execution status for search/count

### LINKED ISSUE #3142: perf: O(N) full scan for conversation search and count
## Problem

`_search_conversations` iterates all conversations in the `_event_services` dict, calls `get_state()` on each, sorts them, then paginates. `_count_conversations` similarly iterates all. With 1000+ conversations, every list or count API call is slow.

**Location:** `conversation_service.py:229–337`

## Evidence

Code review of `conversation_service.py:229–337`.

**Severity:** High

## Impact of fix

API list/search calls become O(page_size) instead of O(total conversations).

## Suggested fix

Maintain a sorted metadata index (by creation time, last activity, etc.) that can be paginated without loading full conversation state. Update incrementally on conversation create/update/delete.

## Related issues and PRs

- #2174 — Fix event search performance and kind filter bug (closed PR)
- #1401 — Fix server stalls by implementing lock-free event search and count (closed PR)
- [OpenHands#12705](https://github.com/OpenHands/OpenHands/issues/12705) — RFC: Efficient Bi-directional Event Loading for V1 Conversations (open issue)

VascoSch92:
HUMAN:

Addressing issue #3142 

---

AGENT:

## Why

Fixes #3142. Status-filtered conversation search and count were O(total conversations) in **full state loads**, not just in iteration.

`_search_conversations` and `_count_conversations` both called `_execution_status()` once per catalog entry. For any conversation without a live `EventService`, that ran `ConversationState.model_validate_json()` over `base_state.json` — validating the agent, LLM config, workspace, secret registry and stats blob — purely to reach one enum field, and paid a separate `asyncio.to_thread` hop per conversation on top.

The unfiltered paths were already fixed by #4100 (the catalog is metadata-only), so this was the last remaining full scan: with 2000 conversations, one `GET /conversations?execution_status=...` call blocked for ~244 ms.

## Summary

- Cache `execution_status` on `_ConversationRecord` next to a `(st_mtime_ns, st_size)` signature of `base_state.json`, and refresh the whole catalog in **one** batched worker-thread pass instead of one thread hop per conversation.
- Live conversations answer from in-memory state; persisted ones cost a `stat()` each and are only re-parsed — for the `execution_status` scalar alone, never a full `ConversationState` — when the file actually changed. Full state is loaded only for the conversations a page returns.
- Filtered search refreshes the index *before* snapshotting the catalog, so a conversation going live mid-refresh is filtered on its current record rather than a superseded one.
- Add `TestConversationSearchScaling` (5 tests) pinning both the scaling property and the correctness properties it must not trade away.

Out-of-process status writes are still observed: the signature is re-checked against the file on every query, so another server instance finishing a conversation is picked up on the next call. The signature is captured *before* the content read, so a write racing the read leaves a signature that no longer matches and forces a re-read rather than caching a torn value.

## Issue Number

Closes #3142

## How to Test

**1. Run the benchmark** (builds a real conversation, clones its on-disk layout N times, times the read APIs behind `GET /conversations` and `GET /conversations/count`):

```bash
uv run python .pr/bench_conversation_search.py --sizes 100,500,1000,2000 --repeat 5
```

**2. Run the regression tests:**

```bash
uv run pytest tests/agent_server/test_conversation_service.py -k TestConversationSearchScaling
```

To confirm they actually catch the regression, check out `conversation_service.py` from `main` and re-run — `test_filtered_search_loads_state_only_for_the_page` and `test_filtered_count_loads_no_state` both fail.

**3. Full agent-server suite:**

```bash
uv run pytest tests/agent_server/ -q
```

## Video/Screenshots

End-to-end benchmark output, median of 5 calls after one warmup, macOS 15 (arm64, APFS SSD), Python 3.13.11. Full write-up in [`.pr/RESULTS.md`](.pr/RESULTS.md).

**Before — `main` @ `c248ec0ae`:**

```
     N |    count() |  count(status) |   page(20) |  page(20,status)
--------------------------------------------------------------------
   100 |       0.00 |          11.49 |       3.50 |            14.96
   500 |       0.00 |          58.75 |       3.96 |            62.64
  1000 |       0.00 |         115.60 |       4.46 |           118.14
  2000 |       0.00 |         234.16 |       5.93 |           243.58

Scaling from N=100 to N=2000:
  count()            x    0.8
  count(status)      x   20.4     <-- dead-linear in total conversations
  page(20)           x    1.7
  page(20,status)    x   16.3     <-- dead-linear in total conversations
```

**After — this branch:**

```
     N |    count() |  count(status) |   page(20) |  page(20,status)
--------------------------------------------------------------------
   100 |       0.00 |           0.19 |       3.61 |             3.73
   500 |       0.00 |           0.75 |       4.02 |             4.71
  1000 |       0.00 |           1.49 |       4.50 |             6.04
  2000 |       0.00 |           2.99 |       5.76 |             9.93

Scaling from N=100 to N=2000:
  count()            x    0.9
  count(status)      x   15.8
  page(20)           x    1.6
  page(20,status)    x    2.7
```

| Operation | N=100 | N=500 | N=1000 | N=2000 |
| --- | ---: | ---: | ---: | ---: |
| `count(status)` | **60x** | **78x** | **78x** | **78x** |
| `page(20, status)` | **4.0x** | **13x** | **20x** | **25x** |

**Test evidence.** On `main`, the two scaling tests fail:

```
FAILED test_filtered_search_loads_state_only_for_the_page
  AssertionError: expected 5 full state loads (the page), got 45
FAILED test_filtered_count_loads_no_state
  AssertionError            # _load_persisted_state_sync was called at all
```

On this branch all four pass, and the full agent-server suite is green:

```
1764 passed, 13 deselected, 41 warnings in 124.45s (0:02:04)
```

`pre-commit run` on the changed files passes (ruff format, ruff lint, pycodestyle, pyright, import rules, tool registration).

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**Why `page(20, status)` doesn't go fully flat, and shouldn't.** It still loads full state for the 20 conversations it actually returns — that's the ~5 ms floor visible in the unchanged `page(20)` column. The remaining slope on `count(status)` is the `stat()` pass that keeps the index honest against out-of-process writers, ~1.5 µs per conversation. Both are deliberate: dropping them would mean serving statuses this process cannot vouch for.

**Micro-optimisations worth flagging**, since they look odd out of context — both are on a path that runs once per conversation per filtered query:
- `_state_signature` takes a `str` and calls `os.stat` directly rather than `Path.stat()`. Measured over 1000 conversations: `Path` construction + `Path.stat()` = 4.62 ms, cached `str` + `os.stat` = 1.26 ms. The path is memoised on the record via `_base_state_path`.
- `_refresh_persisted_statuses_sync` is a `@staticmethod` taking plain tuples rather than reading `self._conversation_records`, so the worker thread never touches the catalog the event loop may be mutating.

**Follow-up not taken here.** `_search_conversations` still sorts the whole catalog and linear-scans for `page_id` on every call. Both are pure in-memory work (~0.5 ms at N=2000) and were not the bottleneck, so I left them alone to keep the diff reviewable. A genuinely sorted index would be the next step if conversation counts grow another order of magnitude.

**Review fixes since the first push** (`0dbc78fa4`, `19114eb12`):
- Startup catalog parsing is strict again. Routing it through the lenient scalar helper had made a corrupt or unknown-status `base_state.json` land in the catalog as IDLE instead of being dropped and logged. Verified against `main`: catalog size 0 and two `error_loading_conversation_catalog` entries on both.
- `_refresh_execution_statuses` no longer stamps a stale disk status onto a conversation that went live during its worker-thread hop, and filtered search now refreshes *before* it snapshots the catalog. Without the reorder, `search(RUNNING)` returned `[]` while the catalog record said `RUNNING`; there is now a deterministic (sleep-free) regression test for it.

Behaviour was checked with a differential harness: 37 conversations across 5 statuses, fixed UUIDs and timestamps, every sort order × 4 filters walked to the last page (16 walks, 88 pages, 240 items), plus on-disk mutation and a live in-memory status flip. Output is byte-identical to `main`.

**Related.** There is an existing draft, #3905, opened by an OpenHands agent against the same issue on branch `openhands/conversation-search-index`. It predates #4100 and targets the state of the code before the catalog existed. This PR is an independent fix against current `main`; whichever lands, the other should be closed.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:e17c390-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-e17c390-python \
  ghcr.io/openhands/agent-server:e17c390-python
```

```
ghcr.io/openhands/agent-server:e17c390-golang-amd64
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-golang-amd64
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-golang-amd64
ghcr.io/openhands/agent-server:e17c390-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:e17c390-golang-arm64
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-golang-arm64
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-golang-arm64
ghcr.io/openhands/agent-server:e17c390-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:e17c390-java-amd64
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-java-amd64
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-java-amd64
ghcr.io/openhands/agent-server:e17c390-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:e17c390-java-arm64
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-java-arm64
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-java-arm64
ghcr.io/openhands/agent-server:e17c390-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:e17c390-python-amd64
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-python-amd64
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-python-amd64
ghcr.io/openhands/agent-server:e17c390-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:e17c390-python-arm64
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-python-arm64
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-python-arm64
ghcr.io/openhands/agent-server:e17c390-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:e17c390-golang
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-golang
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-golang
ghcr.io/openhands/agent-server:e17c390-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:e17c390-java
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-java
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-java
ghcr.io/openhands/agent-server:e17c390-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:e17c390-python
ghcr.io/openhands/agent-server:e17c390b3506f3fde4f938f383d6a93140e0d1bd-python
ghcr.io/openhands/agent-server:vasco-perf-conversation-search-index-python
ghcr.io/openhands/agent-server:e17c390-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `e17c390-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `e17c390-python-amd64`) are also available if needed


openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -496,10 +497,57 @@ def _register_agent_definitions(
...
+    """Cheap change-detection token for a persisted ``base_state.json``.
+
+    Takes a ``str`` and calls ``os.stat`` directly: this runs once per
+    conversation on every status-filtered query, and building a ``Path`` costs
+    several times more than the syscall it wraps.
+
+    ``None`` means "no readable file", which is deliberately never equal to a
+    real signature, so a vanished or unreadable state always forces a re-read.
+    """
+    try:
```
VascoSch92: use contextlib suppress here

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -496,10 +497,57 @@ def _register_agent_definitions(
...
+            payload = json.loads(f.read())
+    except (OSError, ValueError):
+        return None
+    try:
+        return ConversationExecutionStatus(
+            payload.get("execution_status", ConversationExecutionStatus.IDLE.value)
+        )
+    except ValueError:
+        return None
+
```
VascoSch92: use contextlib suppress for both here.

openhands-agent-server/openhands/agent_server/conversation_service.py
VascoSch92: Done in cfbc61439 — now `with suppress(OSError):` around the `os.stat` with the return inside, falling through to `return None`.

openhands-agent-server/openhands/agent_server/conversation_service.py
VascoSch92: Done in cfbc61439. Collapsed both into a single `with suppress(OSError, ValueError):` covering the read and the enum construction — behaviour is identical since both handlers returned `None`, and `ValueError` already covered the `json.loads` failure.

Verified the edge cases still return `None`: missing file, malformed JSON, unknown enum value. An absent `execution_status` key still defaults to `IDLE`.

neubig (approved):
Thank you!

Issue #3152, opened 2026-05-08
bug: unbounded WebSocket connections per conversation


csmith49:
## Problem

There is no cap on WebSocket connections per conversation or globally. Each connection adds a subscriber to PubSub. A malicious or buggy client opening hundreds of WebSocket connections to a single conversation could exhaust server resources.

**Location:** `sockets.py:187`

## Evidence

Code review.

**Severity:** Medium

## Impact of fix

Bounded resource consumption from client connections.

## Suggested fix

Add a max connections limit (per-conversation and global).

## Related issues and PRs

- #3122 — agent-server: WebhookSubscriber.queue grows unbounded under sustained downstream failure (open issue, covers the related webhook queue unbounded growth problem)
- [OpenHands#4936](https://github.com/OpenHands/OpenHands/pull/4936) — Add rate limiting to server endpoints (closed PR, related resource limiting)
- [OpenHands#9504](https://github.com/OpenHands/OpenHands/pull/9504) — Fix WebSocket disconnection when uploading large files (closed PR)

_Discovered during profiling investigation (code review), May 2026. openhands-agent-server v1.19.1._

PR #4202, opened 2026-07-23
perf(agent-server): evict idle conversations from memory after a configurable TTL

### LINKED ISSUE #3141: perf: no eviction or idle timeout for finished conversations
## Problem

The `_event_services` dict holds every active conversation. There is no TTL, no idle timeout, no max size, and no background GC task. A conversation that finishes its work remains fully loaded in memory until explicitly deleted via the API or the server restarts.

**Location:** `conversation_service.py`

## Evidence

Code review. Combined with the atexit leak (see atexit memory leak issue), even "deleted" conversations may not be fully freed.

**Severity:** High

## Impact of fix

Server memory stabilizes over time. Idle conversations are evicted, freeing resources for active ones.

## Suggested fix

Add a background task that periodically checks conversation idle time and evicts those exceeding a configurable TTL. Evicted conversations can be re-hydrated from disk on next access.

## Related issues and PRs

- [OpenHands#13193](https://github.com/OpenHands/OpenHands/pull/13193) — add auto_pause_existing flag and fix sandbox eviction order (open PR, eviction for sandboxes — same pattern needed for conversations)

VascoSch92:
HUMAN:

add the possibility to evict IDLE conversations from memory after configurable TTL.

---

AGENT:

## Why

`ConversationService._event_services` held every conversation that had ever
been hydrated, with **no TTL, no idle timeout, no max size, and no background
GC**. A conversation that finished its work stayed fully loaded in memory
until it was explicitly deleted via the API or the server restarted, so server
memory grew monotonically as finished conversations accumulated (High-severity
finding from the profiling investigation).

## Summary

- Add an optional `conversation_idle_ttl_seconds` config option and a single
  background task that periodically evicts conversations idle beyond the TTL.
  Eviction only tears down the **in-memory runtime**; the lightweight catalog
  record is kept, so an evicted conversation is transparently **re-hydrated
  from disk on next access**. Defaults to **20 minutes**, matching the idle
  timeout used on OpenHands Cloud; set it to `null` to keep the previous
  keep-in-memory-until-deleted behaviour.
- A conversation is **never evicted while it is running** (in-flight
  run / goal loop / pending re-run) or while an **external client (e.g. a
  websocket) is subscribed** to its events. Activity is tracked with a
  monotonic clock refreshed on start, on every event, and on access.
- **Faithful re-hydration:** eviction preserves state that lives only on the
  runtime — it hands active credential bindings back to the catalog (these are
  cleared by `close()` and are not re-derivable from disk), and syncs
  `record.stored` with the runtime's current `StoredConversation` (methods
  such as `switch_acp_model` and secret updates *replace* it). So evicting +
  re-hydrating never loses credentials or reverts persisted metadata.

## Issue Number

Fixes #3141 (Linear OSS-1495).

## How to Test

All commands run from the repo root.

**1. New eviction test suite (real `ConversationService` + `EventService` +
on-disk persistence, no mocks) — 8 tests:**

```
$ uv run pytest tests/agent_server/test_conversation_eviction.py -q
8 passed
```

Covers: task started only when configured; idle conversation evicted +
re-hydrated; recently-active not evicted; access defers eviction; running
conversation not evicted; conversation with an external (websocket-style)
subscriber not evicted (and evictable once it disconnects); **credential
bindings preserved across eviction**; **runtime-replaced `stored` metadata
preserved across eviction**.

**2. Both persistence regression tests fail without the fix** (verified by
reverting only the eviction source and re-running):

```
FAILED ... test_eviction_preserves_credential_bindings
FAILED ... test_eviction_preserves_reassigned_stored_metadata
```

**3. No regressions across the whole agent-server suite:**

```
1767 passed, 13 deselected
```

**4. Lint / format / types (all pre-commit hooks) pass on the changed files:**

```
$ uv run pre-commit run --files <changed files>
Ruff format ... Passed
Ruff lint ... Passed
PEP8 style check (pycodestyle) ... Passed
Type check with pyright ... Passed
Check import dependency rules ... Passed
```

## Video/Screenshots

End-to-end run against real runtimes and on-disk persistence (not just unit
tests) — start a conversation, exercise both safety guards, then evict and
re-hydrate:

```
background eviction task running: True

[1] started conversation 04cf9a4c
    resident in memory : True
    meta.json on disk  : True

[2] idle BUT running -> eviction skipped
    still resident     : True

[3] idle BUT websocket attached -> eviction skipped
    still resident     : True

[4] idle, finished, no client -> EVICTED
    resident in memory : False  (freed)
    catalog record kept: True
    meta.json on disk  : True  (data preserved)
    old runtime closed : True

[5] next access -> re-hydrated from disk
    resident again     : True
    fresh runtime      : True
    same conversation  : True
```

## Type

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## QA evidence

`.pr/live_idle_eviction_evidence.py` drives a real `ConversationService` with
real on-disk persistence (no mocks, no API key, no network) and writes up the
results in `.pr/RESULTS.md`:

```bash
uv run python .pr/live_idle_eviction_evidence.py
```

**The regression, and that it is gone.** 5 rounds x 20 conversations x 40
messages x 8 KB, each arm in its own process:

| | eviction OFF (`main` today) | eviction ON |
| --- | ---: | ---: |
| RSS growth over 100 conversations | +106.9 MiB | **+1.7 MiB** |
| Marginal cost per conversation | 1095 KiB | **17 KiB** |

Without eviction RSS is linear in *conversations ever touched*; with eviction
it is flat and tracks *conversations in use*. Note a single evict does not drop
RSS — CPython returns freed objects to its own allocator, not the OS — so the
fix shows up as the flat growth curve, not a downward step.

**The background task, unassisted.** The `live` phase never calls
`_evict_idle_conversations` and never rewrites an idle clock; it waits for the
real 60 s sweep and then inspects the result:

```
  [PASS] 3 idle conversations evicted by background task            after 60s
  [PASS] catalog records kept, so they stay listable                6 records retained
  [PASS] evicted Conversation objects are garbage collected         3/3 reclaimed
  [PASS] evicted EventService objects are garbage collected         3/3 reclaimed
  [PASS] running conversation NOT evicted                           in-flight run task
  [PASS] websocket-attached conversation NOT evicted                external subscriber
  [PASS] recently-accessed conversation NOT evicted                 accessed every 5s
  [PASS] re-hydrated from disk as a fresh runtime                   new EventService instance
  [PASS] event history identical after re-hydration                 13 events, ids match
  [PASS] event payloads identical after re-hydration                serialized events match
  [PASS] conversation metadata identical after re-hydration         StoredConversation matches
  [PASS] guard defers rather than exempts (subscriber gone)         eligible on the next sweep
```

Weakrefs confirm both layers are reclaimed — the `EventService` and the
`LocalConversation` hanging off it (agent, LLM, tools, full event history) — so
eviction genuinely frees the object graph rather than just dropping a dict
entry.

## Notes

- **Default TTL:** 20 minutes (`DEFAULT_CONVERSATION_IDLE_TTL_SECONDS`), per
  review feedback, matching OpenHands Cloud's idle timeout. `null` disables
  eviction entirely.
- **Check cadence:** the background task checks every `max(60s, ttl/2)`.
- The credential-binding hand-back and `record.stored` sync mirror existing
  code paths (`_start_event_service`'s failure path and
  `activate_credential_binding`), so eviction stays consistent with how the
  runtime already manages that state.
- Part of the profiling-investigation tracking issue (OSS-1507 / #3140–#3152).

### DIFF
--- a/openhands-agent-server/openhands/agent_server/config.py
+++ b/openhands-agent-server/openhands/agent_server/config.py
@@ -2,7 +2,7 @@ import json
 import logging
 import os
 from pathlib import Path
-from typing import Any, ClassVar, Literal
+from typing import Any, ClassVar, Final, Literal
 
 from pydantic import BaseModel, ConfigDict, Field, SecretStr
 
@@ -24,6 +24,8 @@ V1_SESSION_API_KEY_ENV = "OH_SESSION_API_KEYS_0"
 ENVIRONMENT_VARIABLE_PREFIX = "OH"
 CONFIG_PATH_ENV = "OPENHANDS_AGENT_SERVER_CONFIG_PATH"
 DEFAULT_CONFIG_PATH = Path("workspace/openhands_agent_server_config.json")
+# 20 minutes, matching the idle timeout used by OpenHands Cloud.
+DEFAULT_CONVERSATION_IDLE_TTL_SECONDS: Final[float] = 20 * 60.0
 _logger = logging.getLogger(__name__)
 
 
@@ -350,12 +352,15 @@ class Config(BaseModel):
         ),
     )
     conversation_idle_ttl_seconds: float | None = Field(
-        default=None,
+        default=DEFAULT_CONVERSATION_IDLE_TTL_SECONDS,
         gt=0,
         description=(
             "Seconds an idle conversation stays in memory before a background "
             "task evicts it; evicted conversations re-hydrate from disk on next "
-            "access. None (default) keeps them until deleted or restart."
+            "access. Defaults to 20 minutes. Conversations that are running, "
+            "have a pending rerun, or have an attached websocket subscriber are "
+            "never evicted. Set to null to keep conversations in memory until "
+            "they are deleted or the server restarts."
         ),
     )
     telemetry: TelemetrySpec = Field(
--- a/openhands-agent-server/openhands/agent_server/config.py
+++ b/openhands-agent-server/openhands/agent_server/config.py
@@ -349,6 +349,15 @@ class Config(BaseModel):
             "to expire before the first renewal, effectively making it one-shot."
         ),
     )
+    conversation_idle_ttl_seconds: float | None = Field(
+        default=None,
+        gt=0,
+        description=(
+            "Seconds an idle conversation stays in memory before a background "
+            "task evicts it; evicted conversations re-hydrate from disk on next "
+            "access. None (default) keeps them until deleted or restart."
+        ),
+    )
     telemetry: TelemetrySpec = Field(
         default_factory=TelemetrySpec,
         description=(
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -1,4 +1,5 @@
 import asyncio
+import time
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import nullcontext, suppress
 from dataclasses import dataclass, field
@@ -146,6 +147,10 @@ class EventService:
     # Background task for a /goal loop that is running inside this conversation.
     _goal_loop_task: asyncio.Task | None = field(default=None, init=False)
     _goal_loop_outcome: GoalOutcome | None = field(default=None, init=False)
+    # Monotonic clock of the last activity, used for idle eviction.
+    _last_active_monotonic: float = field(default_factory=time.monotonic, init=False)
+    # Subscribers attached at startup; later ones (e.g. websockets) are external.
+    _internal_subscriber_ids: set[UUID] = field(default_factory=set, init=False)
 
     @property
     def conversation_dir(self):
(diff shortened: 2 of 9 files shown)


openhands-agent-server/openhands/agent_server/config.py
```diff
@@ -349,6 +349,19 @@ class Config(BaseModel):
             "to expire before the first renewal, effectively making it one-shot."
         ),
     )
+    conversation_idle_ttl_seconds: float | None = Field(
+        default=None,
+        gt=0,
+        description=(
+            "How long (in seconds) an inactive conversation stays hydrated in "
```
VascoSch92: trimmer the description

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -1803,3 +1814,45 @@ async def __aexit__(self, exc_type, exc_value, traceback):
...
+            return False
+        if self._run_task is not None and not self._run_task.done():
+            return False
+        if self._goal_loop_task is not None and not self._goal_loop_task.done():
+            return False
+        if self._rerun_requested or self._acp_internal_rerun_requested:
+            return False
+        if self.has_external_subscribers():
+            return False
+        return True
```
VascoSch92: just do one if statement here

openhands-agent-server/openhands/agent_server/config.py
VascoSch92: Done — trimmed to three lines in 09145b2.

openhands-agent-server/openhands/agent_server/event_service.py
VascoSch92: Done — collapsed into a single `if` in 09145b2.

openhands-agent-server/openhands/agent_server/config.py
```diff
@@ -349,6 +349,15 @@ class Config(BaseModel):
             "to expire before the first renewal, effectively making it one-shot."
         ),
     )
+    conversation_idle_ttl_seconds: float | None = Field(
```
neubig, line 354: Maybe we want a default here? On the cloud it is 20 minutes sonmaybe 20 minutes would do?

neubig (commented):
LGTM as long as this has been properly QAed, but I think it would be good to have a default

openhands-agent-server/openhands/agent_server/config.py
VascoSch92: Done — defaulted to 20 minutes in a7739c3d1.

Added `DEFAULT_CONVERSATION_IDLE_TTL_SECONDS: Final[float] = 20 * 60.0` and made it the field default. Eviction is guarded by `is_idle_evictable()`, so running conversations, pending reruns, and conversations with an attached websocket subscriber are never evicted — an evicted one just re-hydrates from disk on next access. Setting the field to `null` restores the old keep-in-memory-forever behaviour.

neubig (approved):

PR #4207, opened 2026-07-23
Feat: structured output

### LINKED ISSUE #2566: [Feature]: Structured Output

luciobaiocchi:
HUMAN:

Implemented a general version of  #4116 after @VascoSch92 suggested it. Credit also to his original design idea #2808 remained unfinished.

---

AGENT:

## Why

#2566 asks for first-class structured output. Today, getting reliably
formatted responses out of an agent means manual prompting plus brittle
post-processing. @VascoSch92 implemented the mechanism in #2808 — reviewed
favorably, never landed for lack of time — and offered the final pass to me.
This PR completes that work: his three commits rebased onto current `main`
(authorship preserved), plus a fix for the known meta-field collision issue.

## Summary

- Attach a Pydantic model to **any** tool spec, no subclassing:
  `Tool(name="FinishTool", params={"response_schema": ProjectFacts})`.
  The model's fields are merged into the action schema sent to the LLM
  (`_create_action_type_with_schema`, cached per `(action_type, schema)` pair),
  validated on receipt in `action_from_arguments`, and recoverable typed via
  `parse_response()` / `parse_last_response()`.
- The registry pops `response_schema` from spec params **before** calling the
  tool's `create()` (factories with fixed signatures, e.g. `FinishTool.create`,
  never see it) and applies it via `set_response_schema()` — a `model_copy`,
  consistent with `set_executor()`.
- `response_schema` is runtime-only (`SkipJsonSchema`, `exclude=True`): it never
  crosses the serialization boundary, so persisted events/specs are unchanged
  and older readers are unaffected. `Tool.params` drops class values on dump.
- **New in this pass:** response-schema fields named `summary` or
  `security_risk` are rejected with an explicit `ValueError`. The SDK injects
  meta-fields with those names into every action schema *after* the merge: a
  user field would be silently absorbed as the event summary (`summary`) or
  redefined and swallowed by the risk-analyzer flow (`security_risk`).
- Zero cost when unused: every hook is behind `if response_schema is None`.

## Issue Number

Closes #2566. Supersedes / completes #2808.

## How to Test

```
uv run pytest tests/sdk/tool/test_response_schema.py -v   # 19 passed
uv run pytest tests/sdk/tool tests/cross -q               # 562 passed, 1 skipped
```

19 tests cover: schema extension, payload validation (accept/reject), nested
Pydantic models, class-creation caching, spec serialization dropping class
values, `parse_last_response` across multiple tools, executor unchanged,
tool-without-schema unchanged, action/schema field collision, and the two
reserved meta-field names. Full `tests/sdk/tool` + `tests/cross` suites pass
(verified on two machines); `tests/sdk/agent` passes except one failure
(`test_acp_agent.py::…::test_gemini_046_uses_set_session_model`) reproduced on
unmodified `main`, i.e. pre-existing. `pre-commit` hook set (ruff, pyright,
pycodestyle, import rules, tool registration) clean on all changed files.

See OpenHands/docs#668 for the runnable end-to-end example.

## Video/Screenshots

Both directions of the mechanism:

Schema attached to `TerminalTool` — the model must justify every command:

```
[Terminal commands with rationale]
  $ ls -F
    purpose:          List the contents of the current directory to understand the repository structure.
    expected_outcome: A list of files and directories in the current working directory.
```

Schema attached to `FinishTool` — the final answer comes back as a typed
object via `parse_last_response()` (Pydantic-validated, no text parsing):

```
[Finish]
  description: The OpenHands Software Agent SDK is a set of Python and REST APIs for building agents that work with code, supporting various tasks from simple maintenance to complex refactoring, and offering flexible deployment options for workspaces.
  - The OpenHands Software Agent SDK provides Python and REST APIs for building code-working agents.
  - It supports diverse tasks, from simple README generation to complex refactoring and dependency updates.
  - Agents can operate in local or ephemeral workspaces (Docker/Kubernetes) via the Agent Server.

EXAMPLE_COST: 0.01012643
```

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Commits 1–3 are @VascoSch92's original #2808 work, cherry-picked with
  authorship preserved; they applied onto current `main` without conflicts.
  Commit 4 adds the reserved-name guard, renumbers the example to `56_`
  (48 was taken), and updates its NOTE.
- **Eval impact**: this touches `action_from_arguments` / `to_mcp_tool`, i.e.
  the path of every tool call. The no-schema fast path is covered by
  `test_finish_tool_without_schema_is_unchanged`; happy to have the eval suite
  run before merge if maintainers want the extra confidence — guidance on how
  to trigger it welcome.

## Documentation

Companion docs PR: OpenHands/docs#668.

## Review update (d7c3b53b)

Structured output is now runtime-only on actions and is reconstructed from the existing persisted `tool_call.arguments` field, keeping the event wire shape readable by older SDKs. Response schema merging also preserves validation constraints and rejects schemas without named properties.

## Review update (529e2922)

Preserved `additionalProperties: false` in the advertised schema and made `parse_last_response()` side-effect free. Validation: 214 SDK tool tests and pre-commit passed.


openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -356,7 +375,46 @@ def action_from_arguments(self, arguments: dict[str, Any]) -> Action:
...
+        # don't leak in and we don't have to chase them.
+        data = {k: getattr(action, k) for k in self.response_schema.model_fields}
+        return self.response_schema.model_validate(data)
+
+    def parse_last_response(self, events: "Sequence[Any]") -> BaseModel | None:
+        """Find the most recent ``ActionEvent`` for this tool and parse it.
+
+        Returns ``None`` if the tool has not been called yet.
+        """
+        from openhands.sdk.event import ActionEvent  # avoid circular import
```
VascoSch92:
```suggestion
        from openhands.sdk.event import ActionEvent
```

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -356,7 +375,46 @@ def action_from_arguments(self, arguments: dict[str, Any]) -> Action:
...
+
+        for event in reversed(events):
+            if (
+                isinstance(event, ActionEvent)
+                and event.tool_name == self.name
+                and event.action is not None
+            ):
+                return self.parse_response(event.action)
+        return None
```
VascoSch92: use a generator

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -356,7 +375,49 @@ def action_from_arguments(self, arguments: dict[str, Any]) -> Action:
...
             The action instance created from the arguments.
         """
-        return self.action_type.model_validate(arguments)
+        action_type: type[Action] = self.action_type
+        if self.response_schema is not None:
+            action_type = cast(
+                type[Action],
+                _create_action_type_with_schema(action_type, self.response_schema),
+            )
+        return action_type.model_validate(arguments)
```
VascoSch92: This is the blocker for me. Returning an instance of the generated class puts the dynamic name into `kind`, and `kind` is exactly what the event log round-trips on.

Real conversation with `persistence_dir` set, one FinishTool call carrying a response_schema:

```
events/event-00000-….json
  "action": {"message":"done","success":true,"summary_text":"s",
             "kind":"FinishActionWithTaskResult"}
base_state.json
  "tools": [{"name":"FinishTool","params":{}}]
```

Reopening it (`event_store.py:163,176` read through `Event.model_validate_json`):

```
ValidationError: 1 validation error for Event
action  Value error, Unknown kind 'FinishActionWithTaskResult' for
openhands.sdk.tool.schema.Action; Expected one of: ['MCPToolAction',
'FinishAction', 'InvokeSkillAction', 'SwitchLLMAction', 'ThinkAction',
'VisionInspectAction']
```

It can't self-heal either — the spec that would rebuild the class was stripped on dump (see my note on spec.py). I also tried resuming with the *identical* schema re-declared in code, on the theory that it was just an ordering problem. Still fails: the class is only built lazily by `_get_tool_schema`/`action_from_arguments`, which runs after the event log is read.

Worth calling out that this is what separates the schema classes from `_create_action_type_with_summary` / `create_action_type_with_risk` on main — those are only ever used to render JSON schema (called solely from `_get_tool_schema`, tool.py:517/520), never instantiated, so they never reach `kind`. This PR is the first place a dynamic `Action` subclass gets stored.

I don't have a clean fix to hand. Validating against the augmented type but constructing the base type would keep `kind` stable, except `Action` is `extra="forbid"` so the extra fields have nowhere to go and `parse_response` loses its source. Might be worth putting the structured payload in a declared field on the action instead of merging fields onto the class — that keeps the wire format stable and static.

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -607,6 +674,61 @@ def create_action_type_with_risk(action_type: type[Schema]) -> type[Schema]:
...
+            )
+
+        attrs: dict[str, Any] = {}
+        annotations: dict[str, Any] = {}
+        for field_name, field_info in response_schema.model_fields.items():
+            attrs[field_name] = field_info
+            annotations[field_name] = field_info.annotation
+        attrs["__annotations__"] = annotations
+
+        new_type = type(
```
VascoSch92: The generated name is derived from the *user's* class name, which makes collisions reachable — and `_get_checked_concrete_subclasses` (utils/models.py:152-158) treats a duplicate name as a hard error for the entire `Action` union.

Two modules each defining a `Result` schema is enough:

```
attach Result(alpha) to FinishTool -> ok
attach Result(beta)  to FinishTool -> ok

Event.model_validate({..., "action": {"message":"hi","kind":"FinishAction"}})
ValidationError: Duplicate class definition for openhands.sdk.tool.schema.Action:
  abc.FinishActionWithResult : abc.FinishActionWithResult
```

Note what breaks there: a plain, pre-existing `FinishAction` event that deserialized fine seconds earlier. Once poisoned it stays poisoned for the life of the process, and it takes every `Action` payload with it, not just the schema-bearing tool's.

main's naming (`f"{action_type.__name__}WithSummary"`) derives only from the action type, so it can't collide. This is new.

Minor, but related: `type()` leaves `__module__` as `abc`, which is where that `abc.FinishActionWithResult` comes from. Passing `__module__` explicitly would at least make the error readable.

openhands-sdk/openhands/sdk/tool/spec.py
```diff
@@ -37,3 +37,11 @@ def validate_name(cls, v: str) -> str:
...
         """Convert None params to empty dict."""
         return v if v is not None else {}
+
+    @field_serializer("params")
+    def _ser_params(self, params: dict[str, Any]) -> dict[str, Any]:
+        """Drop non-JSON-serialisable class values (e.g. ``response_schema``
+        Pydantic classes) so the spec can be persisted as part of conversation
+        state. These runtime values are reapplied by the registry on resolve.
+        """
+        return {k: v for k, v in params.items() if not isinstance(v, type)}
```
VascoSch92: This drops the class in python mode as well as JSON, so anything that round-trips the Agent loses structured output silently:

```
agent.tools persisted   : [{'name':'FinishTool','params':{}}]
after round-trip params : {}
resolved response_schema: None
tool schema properties  : ['message','summary']
```

Practical effect: remote / agent-server conversations never get structured output at all (the Agent is serialized to the server), and resumed local conversations quietly lose it.

The docstring says these values "are reapplied by the registry on resolve", but the registry can only reapply what's still in the in-memory spec — after a dump/load there's nothing left to reapply. That's also what turns the `kind` problem into an unrecoverable one rather than something that heals on reload.

At an absolute minimum I'd log a warning when a class-valued param gets dropped, so it isn't invisible.

openhands-sdk/openhands/sdk/tool/registry.py
```diff
@@ -155,7 +155,17 @@ def resolve_tool(
...
+    # `response_schema` is a generic, tool-agnostic mechanism for structured
+    # output, so it must NOT be forwarded into the tool's own ``.create()``.
+    # Pop it *before* calling the resolver — otherwise factories with a fixed
+    # kwarg list (e.g. ``FinishTool.create`` which rejects extra params, or
+    # any ``def create(cls, conv_state, working_dir=None)``) will raise.
+    params = dict(tool_spec.params)
+    response_schema = params.pop("response_schema", None)
+    tools = resolver(params, conv_state)
+    if response_schema is not None:
+        tools = [t.set_response_schema(response_schema) for t in tools]
```
VascoSch92: This applies the schema to every tool the resolver returns, which gets surprising with toolsets:

```
resolve_tool(Tool(name="browser_tool_set", params={"response_schema": Anno}))
tools resolved: 14
carrying response_schema: 14
```

So one `response_schema` on `browser_tool_set` forces those fields onto all 14 browser tools. Same shape for `task_tool_set` and `workflow_tool_set`. I doubt that's intended — either reject it when the resolver returns more than one tool, or document it very loudly.

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -445,6 +506,12 @@ def _get_tool_schema(
     ) -> dict[str, Any]:
         action_type = action_type or self.action_type
 
+        # Merge the structured response schema (if any) into the action schema
+        if self.response_schema is not None:
+            action_type = _create_action_type_with_schema(
```
VascoSch92: Merging here means only subclasses that *don't* override `_get_tool_schema` pick it up. Two in-tree ones do override it, and they fail in opposite directions.

`ClientTool` (client_tool.py:294-299) lifts only `security_risk`/`summary` out of the `super()` schema, so the response fields are dropped — but it doesn't override `action_from_arguments`, so validation still demands them:

```
baseline advertised (no schema): ['q','summary']
with response_schema           : ['q','summary']   <- unchanged
action_from_arguments({"q":"hi"})
  ValidationError: success  Field required
```

The model is never told about the field it's required to send, so every call to that tool fails.

`MCPToolDefinition` (mcp/tool.py:358 and :273) overrides both, so a response_schema is a silent no-op there and `parse_response` later gives `AttributeError: 'MCPToolAction' object has no attribute 'success'`.

Unrelated to those two but same area: `to_mcp_tool` (tool.py:481) builds `inputSchema` straight off `self.action_type` and never routes through here, so it advertises the un-augmented schema. Only tests reach that path today — both in-tree callers pass an explicit `input_schema` — so it's latent rather than broken. But the PR description says this touches `to_mcp_tool`, and as far as I can tell it doesn't.

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -252,6 +259,14 @@ def __init_subclass__(cls, **kwargs):
         default=None, repr=False, exclude=True
     )
 
+    # Optional Pydantic model describing the structured payload the LLM must
+    # return when invoking this tool. When set, the model's fields are merged
+    # into the action schema sent to the LLM, and ``action_from_arguments``
+    # validates the call against that augmented schema. Runtime-only.
+    response_schema: SkipJsonSchema[type[BaseModel] | None] = Field(
```
VascoSch92: Another consequence of hanging the fields off the action type: `Agent._get_action_event` repairs malformed LLM output against `tool.action_type` (agent.py:1210), which doesn't know about them. So the GLM-4.6 / minimax "list arrived as a JSON string" fixups skip response-schema fields entirely:

```
{"message":"done","files_changed":'["a.py","b.py"]'}
after fix_malformed_tool_arguments: files_changed still the raw string
action_from_arguments -> ValidationError: Input should be a valid list
```

Passing the augmented type there is enough to fix it:

```
repaired against augmented type: {'files_changed': ['a.py','b.py']}
```

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -607,6 +674,61 @@ def create_action_type_with_risk(action_type: type[Schema]) -> type[Schema]:
...
+            return cached
+
+        overlap = set(action_type.model_fields) & set(response_schema.model_fields)
+        if overlap:
+            raise ValueError(
+                f"response_schema fields {sorted(overlap)} collide with "
+                f"existing fields on {action_type.__name__}."
+            )
+
+        reserved = _RESERVED_META_FIELDS & set(response_schema.model_fields)
```
VascoSch92: The guard itself is a good addition, but it only fires from `_get_tool_schema()`, so the ValueError lands in the middle of the agent's LLM step rather than where the mistake was made:

```
resolve_tool + set_response_schema -> succeeds, no error
tool.response_schema = <class 'Reserved'>
...later, on the first LLM call:
ValueError: response_schema fields ['summary'] are reserved: ...
```

Any reason it can't move into `set_response_schema`? Then a bad schema blows up at config time, where the user can actually see it.

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -45,8 +46,14 @@
 ObservationT = TypeVar("ObservationT", bound=Observation)
 _action_types_with_risk: dict[type, type] = {}
 _action_types_with_summary: dict[type, type] = {}
+_action_types_with_schema: dict[tuple[type, type], type] = {}
 _action_type_lock = threading.Lock()
 
+# Meta-field names injected into every action schema after the
+# response-schema merge (see ``_create_action_type_with_summary`` and
+# ``create_action_type_with_risk``); rejected in user response schemas.
+_RESERVED_META_FIELDS = frozenset({"summary", "security_risk"})
```
VascoSch92: `kind` belongs in this set too. It's the discriminator injected by `DiscriminatedUnionMixin` — same category as the other two — and a schema declaring it skips the nice error entirely:

```
UserWarning: Field name "kind" in "FinishActionWithKindSchema" shadows an attribute in parent "FinishAction"
TypeError: Field 'kind' of class 'FinishActionWithKindSchema' overrides symbol of same
  name in a parent class. This override with a computed_field is incompatible.
```

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -356,7 +375,49 @@ def action_from_arguments(self, arguments: dict[str, Any]) -> Action:
...
+
+            result = cast(MySchema, tool.parse_response(action))
+
+        Raises ``ValueError`` if no ``response_schema`` is configured.
+        """
+        if self.response_schema is None:
+            raise ValueError(f"Tool '{self.name}' has no response_schema configured.")
+        # Pick only the schema's own fields so meta-fields (kind, summary, ...)
+        # don't leak in and we don't have to chase them.
+        data = {k: getattr(action, k) for k in self.response_schema.model_fields}
```
VascoSch92: Bare `getattr` gives a fairly opaque error when the action doesn't carry the fields:

```
tool.parse_response(FinishAction(message="hi"))
AttributeError: 'FinishAction' object has no attribute 'foo'
```

Not hypothetical — that's precisely the post-resume state, where the tool has its schema back but the stored actions don't have the fields. A sentinel check with a real message would be kinder.

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -45,8 +46,14 @@
 ObservationT = TypeVar("ObservationT", bound=Observation)
 _action_types_with_risk: dict[type, type] = {}
 _action_types_with_summary: dict[type, type] = {}
+_action_types_with_schema: dict[tuple[type, type], type] = {}
```
VascoSch92: This never evicts, and the keys hold strong refs to both the action type and the user's schema class, so generated classes live for the life of the process (51 entries after a 50-schema loop). Every `type()` call also triggers `__init_subclass__` -> `_bump_subclass_generation()`, invalidating the global subclass caches.

Probably fine in practice, since schemas are normally declared once at module scope. But it is a real leak for anything constructing schemas dynamically, and worth at least a comment saying the cache is intentionally permanent.

VascoSch92 (commented):
Moreover, could you make the comments and docstring minimal and not verbose. Thanks

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: actions keep their static kind and persist validated structured output separately.

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: response schemas no longer create Action subclasses.

openhands-sdk/openhands/sdk/tool/spec.py
luciobaiocchi: Fixed in baf3c28d: Python dumps retain the model and JSON dumps carry its JSON Schema for remote and resumed runs.

openhands-sdk/openhands/sdk/tool/registry.py
luciobaiocchi: Fixed in baf3c28d: response_schema now requires a spec that resolves to exactly one tool.

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: ClientTool, MCPTool, and MCP export now merge and validate response fields.

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: malformed-argument repair now also runs against the response schema.

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: reserved and colliding fields now fail in set_response_schema.

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: kind is now rejected with the other reserved fields.

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: parse_response now raises a clear ValueError when structured output is absent.

openhands-sdk/openhands/sdk/tool/tool.py
luciobaiocchi: Fixed in baf3c28d: the dynamic response Action classes and permanent cache were removed.

openhands-sdk/openhands/sdk/tool/schema.py
```diff
@@ -331,6 +332,11 @@ def from_mcp_schema(
 class Action(Schema, ABC):
     """Base schema for input action."""
 
+    structured_output: SkipJsonSchema[dict[str, Any] | None] = Field(
```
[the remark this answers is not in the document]
luciobaiocchi: Fixed in d7c3b53b. Structured output is now a private runtime attribute and is absent from serialized actions. `parse_last_response()` reconstructs and validates it from the existing persisted `tool_call.arguments` field. The regression test asserts the wire payload contains no `structured_output`; I also validated that a head-generated event loads successfully with `openhands-sdk==1.36.0`.

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -455,13 +577,34 @@ def _get_tool_schema(
...
             schema=schema,
             priority=("security_risk", "summary"),
         )
         return schema
 
+    def _merge_response_schema(self, schema: dict[str, Any]) -> dict[str, Any]:
+        if self.response_schema is None:
+            return schema
+
+        response_schema = _response_tool_schema(self.response_schema)
```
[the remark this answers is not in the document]
luciobaiocchi: Fixed in d7c3b53b. Response schemas now retain the complete property constraints while local `$ref` values are expanded, so the advertised schema matches runtime validation. Added coverage for regex, string-length, and numeric bounds. Schemas based only on `additionalProperties` are explicitly rejected because structured output requires named fields.

luciobaiocchi:
Implemented the latest review fixes in d7c3b53b and resolved both threads.

Validation:
- focused structured-output suite: 27 passed
- tool + cross suites: 581 passed, 1 skipped
- pre-commit: passed
- cross-version check: a newly serialized structured action loads with openhands-sdk==1.36.0

Companion docs: OpenHands/docs#668. The optional check-examples job cannot see a fork branch (it only checks OpenHands/docs/<same-branch>) and will remain red until that docs PR is merged or the branch is created upstream; this account cannot push to OpenHands/docs. The sdk-tests CI failure is unrelated timing flakiness in tests/sdk/io/test_filestore_cache.py::test_cache_performance_repeated_reads (5627 other tests passed). I do not have repository permission to rerun that job or apply the review-this label. A human maintainer still needs to run/confirm the requested lightweight eval.

luciobaiocchi:
Current status on a7a662e2: all CI checks pass, including sdk-tests, tools-tests, Windows tests, all three agent-server binary builds, coverage, pre-commit, and the review-thread gate. The runnable example was removed from the SDK diff to avoid the fork-only docs branch limitation; the companion documentation remains in OpenHands/docs#668. Ready for human review and the requested eval confirmation.

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/tool/tool.py`, lines 657–672] **Schema contract mismatch**: The public API accepts any valid Draft 2020-12 object schema, and runtime validation enforces the whole schema, but the LLM-facing merge copies only `properties` and `required`. I reproduced this with `dependentRequired`: the advertised tool schema omitted the dependency, then `action_from_arguments()` rejected the model's otherwise-advertised-valid output because `bar` was absent. The same extraction model mishandles valid dynamic keys allowed by `additionalProperties`: they stay in the action arguments and fail the action model as extras. Either reject unsupported top-level schema semantics at `set_response_schema()` (and require every required/output key to be a named property), or implement composition/extraction that preserves the same contract end to end; claiming arbitrary JSON Schema while silently weakening it is not acceptable.

**[TESTING GAPS]**
- [`tests/sdk/tool/test_response_schema.py`, around line 240] Current coverage proves that constraints inside individual properties survive, but it does not cover object-level semantics. Add regressions for at least `dependentRequired`/conditional requirements and the intended `additionalProperties` policy so the provider-facing schema and runtime validator cannot diverge again.
- [Overall PR] This changes core tool schema generation and every structured tool-call validation path, so it falls under the repository's eval-risk policy. There is no completed eval-monitor run plus human confirmation on the current head; after the correctness issue is fixed, a human maintainer should decide after lightweight evals.

The focused structured-output and malformed-argument suites pass locally (51 tests). The new direct `jsonschema` dependency resolves to 4.23.0, uploaded in July 2024, and was already present transitively in the lockfile, so I found no freshness or downgrade concern.

**[RISK ASSESSMENT]**
This introduces a public structured-output API across base, MCP, and client tools and changes core tool-call validation, giving it a meaningful blast radius. Persistence compatibility and per-property constraints are now handled carefully, but the advertised/runtime JSON Schema mismatch can still break valid user schemas; human eval confirmation is also pending.

**VERDICT:**
❌ **Needs rework**: Align the accepted, advertised, extracted, and validated schema contract before merging, then obtain human eval confirmation.

**KEY INSIGHT:**
A structured-output feature is only reliable when one schema contract survives configuration, provider advertisement, argument splitting, runtime validation, and persistence unchanged.

---

>
>
>

enyst (commented):

🟡 **Acceptable** - The implementation is coherent and well tested at the unit level, but the required behavioral validation gate has not run.

**[TESTING GAPS]**
- [`AGENTS.md`, Line 286 / PR configuration] **Missing agent-behavior evaluation**: This change modifies agent dispatch by repairing structured-output arguments against `tool.response_schema` before tool execution, and it changes the tool protocol used to elicit and extract the final response. The repository explicitly requires the `integration-test` label for changes to prompt templates, tool descriptions, or agent decision logic so benchmark impact is checked. This PR has no such label, and the author explicitly noted that a human maintainer still needs to run or confirm the lightweight eval. The 27 focused tests and broader 581-test run establish the mechanics, but they cannot show that the extra tool and repair path does not regress agent behavior; run the repository's integration/eval gate and record the result before merge.

**[RISK ASSESSMENT]**
The schema persistence, malformed-argument repair, parsing, and backwards-compatibility paths are covered carefully, and I did not find a concrete implementation defect. Risk remains medium because this changes central agent decision behavior and the repository-mandated benchmark signal is absent.

**VERDICT:**
❌ **Needs rework**: The code path is sound, but the required integration/eval evidence must be completed before merge.

**KEY INSIGHT:**
Unit tests prove the structured-output mechanism; only the repository's behavior evaluation can show that inserting it into agent dispatch does not degrade real trajectories.

---

>
>
>

openhands-sdk/openhands/sdk/tool/tool.py
```diff
@@ -345,18 +503,87 @@ def declared_resources(self, action: Action) -> DeclaredResources:  # noqa: ARG0
...
+            None,
+        )
+        if event is None:
+            return None
+        arguments = json.loads(event.tool_call.arguments)
+        _, structured_output = self._split_response_arguments(arguments)
+        if structured_output is None:
+            return None
+        assert event.action is not None
+        event.action._structured_output = structured_output
```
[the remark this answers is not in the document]
luciobaiocchi: Fixed in 529e2922: `parse_last_response()` now uses a copied action, leaving the caller's event unchanged.

luciobaiocchi:
Fixed both inline findings in 529e2922. Validation: 214 SDK tool tests and pre-commit passed. A maintainer still needs to add the `integration-test` label; GitHub rejected it for this contributor account.
@VascoSch92

VascoSch92:
@openhands fix the conflicts

VascoSch92:
Conflicts resolved and merged `main` into the branch (`78f9ecd`).

The only conflicts were in dependency files:
- `openhands-sdk/pyproject.toml` — kept the new `jsonschema>=4.23.0` requirement alongside main's bump to `joserfc>=1.6.8`.
- `uv.lock` — mirrored the same in `requires-dist`; `uv lock --check` passes (resolved locked versions: `jsonschema 4.23.0`, `joserfc 1.6.8`).

No source-code conflicts — the structured-output changes merged cleanly.

Validation:
- `tests/sdk/tool/test_response_schema.py`: 32 passed
- `tests/sdk/tool`: 214 passed
- import sanity check: `jsonschema 4.23.0`, `joserfc 1.6.8`, `ToolDefinition.set_response_schema` intact

The PR now reports `mergeable: true`.

_This comment was posted by an AI agent (OpenHands) on behalf of the user._

luciobaiocchi:
@VascoSch92 can you take a look ?

VascoSch92 (approved):
LGTM

VascoSch92:
I'm looking inside the tests but it seems an infrastructure problem

VascoSch92:
thanks

luciobaiocchi:
It was a pleasure working with you on this. I already have a few new ideas 😉. Thanks again!

PR #4209, opened 2026-07-23
Fix/pr artifacts fork safe

luciobaiocchi:
HUMAN:

Proposed fix for the check-pr-artifacts.

---

AGENT:

## Why

`.pr/` contains temporary, PR-only evidence and must never remain on `main`.
The existing workflow could neither post its notice nor clean up fork PRs:
`pull_request` and fork `pull_request_review` runs receive a read-only token and
no repository secrets. This caused the notice to fail with 403 on #4207, while
the artifacts from the merged fork PR #4202 remained on `main`.

## Summary

- Run artifact detection on `pull_request_target`, inspect the fork commit via
  the Contents API, and post/update the notice with the base repository token.
  No fork code is checked out or executed with write permissions.
- Continue removing `.pr/` from same-repository PR branches on approval.
- For fork PRs, which cannot be modified reliably, remove `.pr/` from the
  trusted base branch immediately after merge. This also covers a missed
  approval cleanup on a same-repository PR.
- Remove `.pr/RESULTS.md` and `.pr/live_idle_eviction_evidence.py` leaked into
  `main` by #4202.

## Issue Number

N/A. Related failures: #4202 and #4207.

## How to Test

- `uv run pre-commit run --files .github/workflows/pr-artifacts.yml`
- YAML parsed with a string-preserving loader.
- All three embedded `actions/github-script` blocks pass `node --check`.
- The real GitHub Contents API correctly finds `.pr/` on #4202's fork commit
  and returns 404 on this PR's cleaned commit.
- `git diff --check`

## Video/Screenshots

N/A — CI workflow change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

A base-repository workflow cannot push to an arbitrary contributor fork,
especially when maintainer edits are disabled. The fork-safe cleanup therefore
runs against the trusted base branch after merge. The workflow fails visibly if
that cleanup push is rejected instead of hiding the leak.

_This PR description was updated by an OpenHands AI agent on behalf of the contributor._


.github/workflows/pr-artifacts.yml
```diff
@@ -125,18 +125,36 @@ jobs:
...
+                            repo: context.repo.repo,
+                            issue_number: context.issue.number,
+                            body: body,
+                          });
+                        }
+                      } catch (error) {
+                        // Fork PRs run pull_request workflows with a read-only
+                        // token, so posting the notice is impossible there.
+                        // The notice is informational: degrade to a warning and
+                        // the job summary instead of failing the whole job.
```
enyst: I understand, this is better and accurate, however the directory is still not removed, and now there's no clue that it's not removed... 🤔 

@openhands understand the meaning of .pr/ directory and the problem we have here. The nice behavior is to always (1) post a comment, and (2) auto-remove the directory. Maybe we can split the job or something?

enyst: Addressed in bdf2478aa21d4cb51791abaedb38f00ef95e60ff. The jobs are now split by trust boundary: `pull_request_target` detects `.pr/` through the Contents API and can always post/update the notice without checking out fork code; same-repository PRs are still cleaned on approval; and fork PRs are cleaned from the trusted base branch immediately after merge. We cannot reliably push to an arbitrary fork (PR #4202 had maintainer edits disabled), so post-merge base cleanup is the enforceable automatic path. The workflow fails visibly if that push is rejected.

_This reply was generated by an OpenHands AI agent on behalf of the contributor._

enyst: @luciobaiocchi What do you think about agent's solution to the clean-up? It committed to PR; we can remove it if we don't like it though!

luciobaiocchi: Looks great to me happy to keep it. The key thing I checked is the pull_request_target security angle, and it’s handled correctly: the job that touches the fork only reads via the API (no checkout of untrusted code), and the only checkouts with a write token are trusted (same-repo head, or the base branch post-merge). The cleanup-after-merge job is a nice way to cover forks where we can’t push to the head branch. Fine to drop my earlier try/catch commit since pull_request_target removes the 403 path entirely. Thanks for taking it the full distance!

luciobaiocchi (commented):
ok, thank for the improvement 😁

luciobaiocchi (commented):
updated with main

enyst (approved):

PR #3184, opened 2026-05-09
fix(agent-server): take over conversation leases left by crashed owners

rbren:
When the agent-server is killed or otherwise exits non-gracefully, the per-conversation `owner_lease.json` is left on disk with an `expires_at` timestamp up to 45 seconds in the future. A freshly started server generates a new `owner_instance_id`, sees the lease as held, and skips the conversation in `ConversationService.__aenter__` - so existing conversations effectively disappear from the API for up to the full TTL window after a crash-restart.

Augment the lease payload with the owner's hostname and PID, and on `claim()` treat the lease as stale (eligible for takeover) when the recorded host matches this machine but the recorded PID is no longer alive. Cross-host leases continue to use the TTL-only path so the existing split-brain protection between distinct service instances is unchanged.

Backwards compatible: legacy lease files without `owner_host` / `owner_pid` keep their previous TTL-only semantics.

- [ ] A human has tested these changes.

---

## Summary

-

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
| Variant | Architectures | Base Image | Docs / Tags |
|---|---|---|---|
| java | amd64, arm64 | `eclipse-temurin:17-jdk` | [Link](https://hub.docker.com/_/eclipse-temurin:17-jdk) |
| python | amd64, arm64 | `nikolaik/python-nodejs:python3.13-nodejs22-slim` | [Link](https://hub.docker.com/_/nikolaik/python-nodejs:python3.13-nodejs22-slim) |
| golang | amd64, arm64 | `golang:1.21-bookworm` | [Link](https://hub.docker.com/_/golang:1.21-bookworm) |

**Pull (multi-arch manifest)**
```bash
# Each variant is a multi-arch manifest supporting both amd64 and arm64
docker pull ghcr.io/openhands/agent-server:529efa7-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-529efa7-python \
  ghcr.io/openhands/agent-server:529efa7-python
```

**All tags pushed for this build**
```
ghcr.io/openhands/agent-server:529efa7-golang-amd64
ghcr.io/openhands/agent-server:529efa737d7aaf228f15c2b5923d6ce05217fdaf-golang-amd64
[...]

PR #4212, opened 2026-07-23
fix(agent-server): disable conversation leasing by default

### LINKED ISSUE #4192: [Bug]: restarting agent-server quickly loses all historical conversations

### Bug Description

IIRC there is a "lock" that looks to see if conversations were owned by an agent-server in the last ~60s, and doesn't load the conversation into history if so

I don't quite understand why we have this--I think it's to prevent different agent-servers from claiming the same conversation?

Anyways this constantly screws me up--if I restart the server, when it starts back up all my convo history is gone

If we need this under some weird circumstance, maybe we hide the behavior behind an env var, default off?

### Expected Behavior

Restarting just rehydrates all conversations

### Actual Behavior

Conversations disappear

### Steps to Reproduce

* start an agent-canvas
* create a convo
* stop and quickly restart the agent-canvas server
* your convo is gone

### Installation Method

agent-canvas via npm

_No response_

### SDK Version

_No response_

### Version Confirmation

### Python Version

_No response_

_No response_

### Operating System

None

### Logs and Error Messages

_No response_

### Minimal Code Sample

_No response_

_No response_

andreruizloera:
HUMAN:

Restarts were hiding conversation history for ~45s because of the ownership lease, which only matters for multi-server setups — this turns leasing off by default (env var re-enables it). I reviewed the diff and repro before marking ready, ping me with any feedback.

---

AGENT:

Disclosure: this PR was authored end-to-end by an AI agent (Claude Code, model Claude Fable 5), operated by @andreruizloera, who will review it and fill in the HUMAN section before marking it ready.

## Why

Fixes #4192.

On a quick or containerized restart, all conversation history appears to vanish for up to 45 s. Reproducing against `main` (details and scripts in `.pr/`), the conversations are still *listed*, but `_get_or_load_event_service` swallows `ConversationLeaseHeldError` and returns `None`, so every `/conversations/{id}/events/*` route 404s while the previous server's lease is still live. Graceful restarts release leases and same-host crashes are healed by the dead-pid takeover from #3184 — the surviving failure modes are (a) a restart racing the old process's shutdown and (b) container recreation, where the new hostname makes pid liveness unverifiable and only TTL expiry unblocks.

Live REST demonstration (full transcript in `.pr/evidence.md`):

```
OLD default (leasing on):  restarted server: conversations listed=1  events endpoint HTTP=404
NEW default (leasing off): restarted server: conversations listed=1  events endpoint HTTP=200
```

## Summary

- `Config.lease_ttl_seconds` now defaults to `0.0` (leasing disabled), per the direction suggested in #4192. Most servers are single-instance, where a lease can only delay rehydration; deployments sharing conversation storage across instances opt in with `OH_LEASE_TTL_SECONDS` (e.g. `45`).
- The lease machinery itself is unchanged, and `DEFAULT_LEASE_TTL_SECONDS` (45 s) still applies to direct `ConversationService`/`EventService` embedders — only the server config default changes.
- Tests: config default/env-override/validation, plus a regression test that a leasing-disabled server rehydrates a conversation despite a live foreign-host lease.

## Issue Number

#4192

## How to Test

- `uv run pytest tests/agent_server/test_config.py tests/agent_server/test_env_parser.py tests/agent_server/test_conversation_lease.py tests/cross/test_conversation_lease_behavior.py`
- End-to-end: `bash .pr/live_test_4192.sh` runs the old and new defaults against a real server over REST (create conversation → SIGKILL → rewrite lease to a vanished hostname → restart → list + fetch events).
- Full `tests/agent_server` + `tests/cross` pass except `test_openai_chat_completions_gateway_over_real_server`, which fails identically on unmodified `main` under full-suite load (pre-existing flake, passes solo).

## Video/Screenshots

Terminal transcripts in `.pr/evidence.md`.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Deployment callout:** any multi-instance deployment relying on the previous default (e.g. OpenHands Cloud, if it does not set this explicitly) should set `OH_LEASE_TTL_SECONDS=45` when picking this up, otherwise split-brain protection is off. Flagging because I cannot inspect deployment configs from this repo.
- This is a public `Field(default=...)` behavioral change, so the API-breakage workflow should apply `release-note-required`.
- Alternative considered: serving `/events` reads from persisted state when the lease is held (keeps leasing on with no read outage). Strictly safer but a much larger change touching the router dependency and `EventService`; happy to pursue it instead if preferred.
- `.pr/` contains the repro script, live test, and evidence; per repo policy it must be removed before merge since auto-cleanup cannot push to forks — say the word and I'll drop it once review is done.

### DIFF
--- a/.pr/evidence.md
+++ b/.pr/evidence.md
@@ -1,0 +1,56 @@
+# Evidence for #4192 fix
+
+## Root-cause repro (service level): `.pr/repro_4192.py`
+
+Four restart shapes against `ConversationService` on unmodified `main`:
+
+```
+== A: graceful stop, then restart ==
+  lease exists after graceful stop: False
+  [restarted server] listed=True (n=1) info=YES event_service=YES
+== B: crash (dead pid, same host), quick restart ==
+  [restarted server] listed=True (n=1) info=YES event_service=YES
+== C: docker-style restart (different host in lease) ==
+  [restarted server] listed=True (n=1) info=YES event_service=NO (lease held?)
+== D: quick restart while old process still alive ==
+  [restarted server] listed=True (n=1) info=YES event_service=NO (lease held?)
+```
+
+Graceful restarts (A) and same-host crash recovery (B, from #3184) already
+work. The reported symptom comes from C/D: the conversation is still *listed*,
+but `_get_or_load_event_service` swallows `ConversationLeaseHeldError` and
+returns `None`, so every `/events/*` route 404s for up to the 45 s TTL — in a
+UI the history is effectively gone.
+
+## Live end-to-end (real server over REST): `.pr/live_test_4192.sh`
+
+Create a conversation via `POST /api/conversations`, SIGKILL the server,
+rewrite the lease's `owner_host` to a vanished hostname (what a recreated
+container looks like — pid liveness can't be verified cross-host), restart on
+the same storage, and query:
+
+```
+=== OLD default (leasing on, OH_LEASE_TTL_SECONDS=45) ===
+created conversation: 924dbea3-2a10-40a5-b256-359e5c218f44
+lease rewritten to foreign host (still live)
+restarted server: conversations listed=1 events endpoint HTTP=404
+
+=== NEW default (leasing disabled) ===
+created conversation: 54b41ef0-d906-40e3-aace-6037107abebe
+no lease file present (leasing disabled)
+restarted server: conversations listed=1 events endpoint HTTP=200
+```
+
+The OLD run also confirms `OH_LEASE_TTL_SECONDS` still enables leasing
+end-to-end (the opt-in path for shared-storage deployments).
+
+## Test suites
+
+- `tests/agent_server/test_config.py`, `tests/agent_server/test_env_parser.py`,
+  `tests/agent_server/test_conversation_lease.py`,
+  `tests/cross/test_conversation_lease_behavior.py`: all pass.
+- Full `tests/cross`: 362 passed, 1 skipped.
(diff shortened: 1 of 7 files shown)


enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [openhands-agent-server/openhands/agent_server/config.py, Line 340] **Availability fix disables the data-integrity guard globally**: changing the server default to `0.0` means any existing multi-instance/shared-storage deployment that relied on the safe default silently loses ownership fencing. Two servers can then load and write the same conversation concurrently; the lease generation checks exist specifically to prevent that split-brain corruption. The PR itself says it cannot verify deployment configs, so this is knowingly trading a 45-second read outage for potential persisted-state corruption in an unknown set of users. Keep leasing safe-by-default and fix the restart/read path (for example, serve persisted read-only history while another owner is live), or introduce an explicit single-instance mode whose deployment scope is proven before disabling fencing.

**VERDICT:**
❌ **Needs rework**: A default that prevents concurrent writers cannot become opt-in merely because the single-instance case is more common.

**KEY INSIGHT:**
Failing closed for 45 seconds is recoverable; silently permitting two writers against one conversation is not.

**[RISK ASSESSMENT]**
This changes a core concurrency-safety default and can disable split-brain protection for existing deployments without any configuration change. **Recommendation**: Do not auto-merge. Request review from a human architect/SRE who can validate every shared-storage deployment and choose a fix that preserves ownership fencing.

---

>
>
>

PR #4217, opened 2026-07-24
fix(agent-server): redact LLM & condenser secrets in download-trajectory

### LINKED ISSUE #4216: download-trajectory leaks LLM & condenser API keys (no masking at export boundary)
### Summary

`GET /api/file/download-trajectory/{conversation_id}` builds a zip of the raw conversation directory and does **no secret masking of its own**. As a result a downloaded trajectory can contain LLM API‑key material:

- **plaintext** API keys when the server runs **without** `OH_SECRET_KEY` (secrets are persisted unencrypted), or
- **recoverable Fernet ciphertext** when a cipher **is** configured.

Both the agent LLM key and the **condenser** LLM key are exposed, because the whole agent config (`agent.llm`, `agent.condenser.llm`) is serialized into `base_state.json` / `meta.json`.

The stated invariant is that trajectory files MUST mask API keys. Today the endpoint only inherits whatever masking the persistence layer happened to apply; it makes no guarantee at the export boundary.

### Where

`openhands-agent-server/openhands/agent_server/file_router.py` → `download_trajectory` → `_create_zip_from_directory`, which just copies bytes:

```python
for path in sorted(source_dir.rglob("*")):
    archive.write(path, path.relative_to(source_dir.parent))
```

The zipped files (`base_state.json`, `meta.json`, `events/…`) carry `api_key` / `aws_*` fields.

### Reproduction

1. Start an agent-server (no `OH_SECRET_KEY` → plaintext on disk; or with it → encrypted).
2. Run a conversation with an agent that has an LLM key and an `LLMSummarizingCondenser` (its own LLM key).
3. `GET /api/file/download-trajectory/{id}` and inspect `base_state.json` / `meta.json` in the zip — `agent.llm.api_key` and `agent.condenser.llm.api_key` are present (plaintext or `gAAAA…` ciphertext).

Observed on a real trajectory zip: 2 `api_key` values in `base_state.json` and 2 in `meta.json` (main + condenser).

### Expected

A downloaded trajectory should never contain credential material — plaintext or encrypted — regardless of how the conversation was persisted.

### Proposed fix

Redact secret-bearing fields (`LLM_SECRET_FIELDS` + the `llm_api_key` settings alias) at the export boundary while building the zip, masking both plaintext and Fernet values, leaving non-secret content byte-identical. PR incoming.

smolpaws:
HUMAN:

Trajectory downloads must not leak LLM API keys — asked the agent to fix this and keep the redaction logic out of the router files.

---

AGENT:

## Why

`GET /api/file/download-trajectory/{id}` zips the raw conversation directory with no masking of its own, so a downloaded trajectory carries LLM API-key material:

- **plaintext** keys when the server runs without `OH_SECRET_KEY`, or
- **recoverable Fernet ciphertext** when a cipher is configured.

Both the agent LLM key and the **condenser** LLM key leak, since the whole agent config is serialized into `base_state.json` / `meta.json`. A trajectory archive is meant to be shared for debugging, so secrets must never leave the server in it.

## Summary

- New security utility module `openhands/agent_server/_secret_redaction.py` (mirroring the existing `_secrets_exposure.py` pattern) that redacts secret-bearing fields at the export boundary: `redact_secrets_in_obj` walks parsed JSON and masks any `LLM_SECRET_FIELDS` (`api_key`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token`) plus the settings-level `llm_api_key` alias, for **both** plaintext and Fernet (`gAAAA…`) values. Any Fernet ciphertext is redacted **regardless of field name**, so cipher-encrypted custom secrets in `secret_registry.secret_sources` (stored under `value`) are covered too.
- `redacted_file_bytes` applies this to `*.json` and `*.jsonl` (per-event log) files; anything that isn't JSON is passed through untouched, so non-secret content stays byte-identical.
- `_create_zip_from_directory` in `file_router.py` streams redacted bytes for changed files and copies the rest verbatim, so the guarantee holds regardless of how the conversation was persisted.

## Issue Number

Fixes #4216

## How to Test

1. Start the agent server without `OH_SECRET_KEY` and run a conversation with a real `api_key` configured (optionally with a condenser LLM).
2. `curl -OJ http://localhost:8000/api/file/download-trajectory/<conversation_id>` and unzip the archive.
3. `grep -r "sk-" <unzipped dir>` — no API-key material (plaintext or `gAAAA…` ciphertext) appears; `api_key` fields read `**********` while model names and all other content are preserved.
4. Automated: `uv run pytest tests/agent_server/test_file_router.py -q` → 78 passed. The new `test_download_trajectory_redacts_llm_and_condenser_secrets` seeds `meta.json` (plaintext main key + encrypted condenser blob), `base_state.json`, and a `full_state` event, and asserts no secret material survives into the zip. Verified it fails without the fix and passes with it. ruff + pyright clean.

## Video/Screenshots

N/A (backend security fix; see test evidence above)

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Redaction happens only at the download boundary; on-disk persistence is unchanged, so resuming conversations on the server keeps working with the original (optionally encrypted) secrets.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/_secret_redaction.py
+++ b/openhands-agent-server/openhands/agent_server/_secret_redaction.py
@@ -9,6 +9,7 @@ import json
 from pathlib import Path
 
 from openhands.sdk.llm.llm import LLM_SECRET_FIELDS
+from openhands.sdk.utils.cipher import FERNET_TOKEN_PREFIX
 from openhands.sdk.utils.pydantic_secrets import REDACTED_SECRET_VALUE
 
 
@@ -19,6 +20,18 @@ from openhands.sdk.utils.pydantic_secrets import REDACTED_SECRET_VALUE
 TRAJECTORY_SECRET_FIELDS: frozenset[str] = frozenset(
     (*LLM_SECRET_FIELDS, "llm_api_key")
 )
+
+
+def _is_secret_value(key: object, value: object) -> bool:
+    if not isinstance(value, str) or value in ("", REDACTED_SECRET_VALUE):
+        return False
+    # Any Fernet token in a persisted conversation is ciphertext produced with
+    # the server's OH_SECRET_KEY (e.g. custom secrets in
+    # ``secret_registry.secret_sources``) and therefore recoverable secret
+    # material, regardless of which field carries it.
+    if value.startswith(FERNET_TOKEN_PREFIX):
+        return True
+    return key in TRAJECTORY_SECRET_FIELDS
 
 
 def redact_secrets_in_obj(obj: object) -> bool:
@@ -32,19 +45,17 @@ def redact_secrets_in_obj(obj: object) -> bool:
     changed = False
     if isinstance(obj, dict):
         for key, value in obj.items():
-            if (
-                key in TRAJECTORY_SECRET_FIELDS
-                and isinstance(value, str)
-                and value != ""
-                and value != REDACTED_SECRET_VALUE
-            ):
+            if _is_secret_value(key, value):
                 obj[key] = REDACTED_SECRET_VALUE
                 changed = True
             elif redact_secrets_in_obj(value):
                 changed = True
     elif isinstance(obj, list):
-        for item in obj:
-            if redact_secrets_in_obj(item):
+        for index, item in enumerate(obj):
+            if _is_secret_value(None, item):
+                obj[index] = REDACTED_SECRET_VALUE
+                changed = True
+            elif redact_secrets_in_obj(item):
                 changed = True
     return changed
 
--- a/openhands-agent-server/openhands/agent_server/_secret_redaction.py
+++ b/openhands-agent-server/openhands/agent_server/_secret_redaction.py
@@ -1,0 +1,93 @@
+"""Redaction helpers that keep secrets out of exported artifacts.
+
+Used at export boundaries (e.g. trajectory download) where persisted
+conversation data leaves the server and must never carry credential
+material — neither plaintext nor recoverable Fernet ciphertext.
+"""
+
+import json
+from pathlib import Path
+
+from openhands.sdk.llm.llm import LLM_SECRET_FIELDS
+from openhands.sdk.utils.pydantic_secrets import REDACTED_SECRET_VALUE
+
+
+# Field names that carry credentials anywhere in a persisted conversation
+# payload (base_state.json, meta.json, events). Kept in sync with
+# ``LLM_SECRET_FIELDS`` and extended with the settings-level ``llm_api_key``
+# alias.
+TRAJECTORY_SECRET_FIELDS: frozenset[str] = frozenset(
+    (*LLM_SECRET_FIELDS, "llm_api_key")
+)
+
+
+def redact_secrets_in_obj(obj: object) -> bool:
+    """Recursively replace secret-bearing values with the redaction marker.
+
+    Returns ``True`` if anything was changed. Both plaintext and encrypted
+    (Fernet) secret material are masked, so the guarantee holds regardless of
+    whether ``OH_SECRET_KEY`` was configured when the conversation was
+    persisted.
+    """
+    changed = False
+    if isinstance(obj, dict):
+        for key, value in obj.items():
+            if (
+                key in TRAJECTORY_SECRET_FIELDS
+                and isinstance(value, str)
+                and value != ""
+                and value != REDACTED_SECRET_VALUE
+            ):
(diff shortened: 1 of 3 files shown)


enyst:
@openhands I prefer if such functions are not in the `_router.py` files. Please fix. A security utility/service or something. Think before you do. Push to PR (it's from a fork, set your remote properly). Then take a step and verify the code /codereview . Post your feedback on the PR with gh api and event approved or what you see fit.

enyst (approved):
## Code Review

🟢 **Good taste** — simple, boring fix at the single right choke point.

Reviewed at head `5462312` (after the refactor moving redaction helpers out of `file_router.py` into `_secret_redaction.py`, per review feedback).

### Three questions

1. **Real problem?** Yes — #4216 is a genuine data-exposure bug: `download-trajectory` shipped plaintext API keys (no `OH_SECRET_KEY`) or recoverable Fernet ciphertext (with one). A trajectory archive exists precisely to be shared, so this is the worst place to leak credentials.
2. **Simpler way?** Redacting at the export boundary is the minimal design: persistence is untouched, no schema/mode changes, server-side resume keeps working with the original (optionally encrypted) secrets. The recursive dict/list walk over parsed JSON is the obvious data structure — no special cases per file type beyond json/jsonl.
3. **What breaks?** Nothing user-visible. Files that contain no secrets stay byte-identical in the archive; only secret-bearing JSON is re-serialized. Downloaded trajectories no longer carry usable keys — that is the point, not a regression.

### What I verified

- `_create_zip_from_directory` is only called by the trajectory endpoint (`grep` confirms one call site), so the redaction does not silently alter the workspace `/archive` or file-download paths.
- The Fernet-anywhere rule closes a real gap: cipher-encrypted custom secrets in `secret_registry.secret_sources` are stored under `value` — a field name *not* in `LLM_SECRET_FIELDS` — and would have leaked as recoverable ciphertext. Now any `gAAAAA…` token is masked regardless of field name. (Plaintext custom secrets were never persisted; `serialize_secret` redacts them at save time without a cipher.)
- The test exercises the real HTTP endpoint end-to-end (TestClient + real files on disk, no mocks of the unit under test), asserts secret material is absent from the raw archive bytes, and asserts non-secret content (model names) is preserved. It fails without the fix.
- `tests/agent_server/test_file_router.py`: 78 passed. ruff + pyright clean on the touched files.

### Minor notes (non-blocking)

- The `gAAAAA` prefix heuristic could theoretically redact a legitimate non-secret string that happens to start with the Fernet prefix inside an event payload. In a shareable debug archive, over-redaction is the correct failure mode — fine as is.
- The `.jsonl` branch in `redacted_file_bytes` is currently unexercised (the event store writes discrete `event-*.json` files). Cheap future-proofing; acceptable.

**[RISK ASSESSMENT]**
Export-path only, fails safe (over-redacts rather than leaks), on-disk persistence and conversation resume untouched, behavior proven by an end-to-end test against the real endpoint.

**VERDICT:**
✅ **Worth merging** — closes a real secret-leak with a contained boundary fix.

**KEY INSIGHT:**
Redacting once at the export boundary makes the no-leak guarantee independent of how (or whether) secrets were encrypted at persist time — the only design that holds for old and new conversations alike.

---
*This review was generated by an AI agent (OpenHands) on behalf of @enyst.*

enyst:
@enyst Done — summary of changes:

- **Refactor** (`55372a0`): moved the redaction helpers (`redact_secrets_in_obj`, `redacted_file_bytes`, `TRAJECTORY_SECRET_FIELDS`) out of `file_router.py` into a dedicated security utility module `openhands/agent_server/_secret_redaction.py`, mirroring the existing `_secrets_exposure.py` pattern. The router now only keeps the zip-building logic.
- **Gap found & fixed during review** (`5462312`): cipher-encrypted *custom* secrets in `secret_registry.secret_sources` are persisted under a `value` field that is not in `LLM_SECRET_FIELDS`, so their ciphertext would still have leaked. Any Fernet (`gAAAAA…`) token is now redacted regardless of field name; test extended to cover it.
- **PR description** rewritten to follow the repo template (HUMAN/AGENT sections, Why/Summary/How to Test), validated locally with `.github/scripts/check_pr_description.py` — this fixes the failing *PR Description Check* CI job.
- Pushed to the fork branch (`smolpaws/software-agent-sdk`). `tests/agent_server/test_file_router.py`: 78 passed; ruff + pyright clean.
- Code review posted as an approving review above.

*This comment was created by an AI agent (OpenHands) on behalf of @enyst.*

PR #4219, opened 2026-07-24
fix(agent-server): make base_state.json authoritative for the resume agent (#4032)

### LINKED ISSUE #4032: [Bug]: LLM profile timeout is reset after agent-server restart

smolpaws:
## What & why

Fixes #4032 (LLM/profile timeout reverts after an agent-server restart) by removing its **root cause** rather than papering over it.

The full agent is persisted in **two** places:
- `base_state.json` — the SDK's own conversation state, rewritten on *every* live mutation (`ConversationState.agent = ...`, e.g. `switch_llm`);
- `meta.json` — a `StoredConversation` snapshot taken at creation.

On resume, `EventService.start()` rebuilt the runtime agent from **`meta.json`**, and `ConversationState.create()` then copies that agent over `base_state.json` (`state.agent = agent`). So a live `switch_llm` — which writes only `base_state.json` — was silently reverted to the creation-time LLM (and its `timeout`) on the next restart.

This is a duplicate-source-of-truth bug. Background write-up (plain-English, code-grounded): https://enyst.github.io/arch/meta-vs-base-state-duplication.html

## The fix

Make `base_state.json` authoritative for the resume agent. On resume, take **`llm` and `condenser`** from `base_state.json` (the fields a live switch can change, and where those changes are already persisted) while keeping `tools` / `agent_context` / `mcp_config` from the creation-time snapshot — those are re-derived by the plugin merge on first run, so taking the already-merged `base_state` copy would double-merge them.

No new write-side mirror. `main` already carries one hand-rolled mirror for the ACP model switch (`switch_acp_model` → `save_meta`); this read-side approach avoids adding a second for `switch_llm`, and covers any future `llm`/`condenser` divergence for free.

## Relationship to #4028

Same bug, different level. #4028 (Vasco) fixes it by *mirroring* the switched `llm`/`condenser` into `meta.json` on change — correct and minimal within the current model. This PR removes the need to mirror at all by making the SDK-owned file authoritative on read. Same field-scoping insight (`llm`+`condenser`), applied on the read side. Posting as an alternative for discussion — happy to converge on whichever the maintainers prefer.

## Tests

Adds `test_switched_llm_survives_restart_via_base_state`: create with `timeout=300`, `switch_llm` to `600` (writes `base_state.json` only), tear the service down and start a fresh one over the same dir (a restart), assert the live agent uses `600`. Verified it **fails on the pre-fix behavior**. Full `test_event_service.py` + `test_conversation_service.py` (212) pass; ruff + pyright clean.

## Not in this PR

Fully removing the `agent` blob from `meta.json` (it's still read cold for codex-detection and conversation-info). That's the larger "slim meta to server-only fields" step; this PR makes `base_state` authoritative first, which is what actually fixes #4032.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -177,6 +177,42 @@ class EventService:
                 )
             )
 
+    def _resume_agent_with_live_llm(self) -> AgentBase:
+        """The agent to instantiate on start, with live LLM/condenser applied.
+
+        Fresh conversation (no ``base_state.json`` yet): returns
+        ``self.stored.agent`` from the create request unchanged.
+
+        Resume: returns the creation-time ``self.stored.agent`` (so ``tools`` /
+        ``agent_context`` / ``mcp_config`` are the un-merged config the plugin
+        merge expects) but with ``llm`` and ``condenser`` overridden from
+        ``base_state.json``. Those two are exactly what a live ``switch_llm``
+        can change, and base_state.json is where such changes are persisted —
+        so this is what keeps a switched config from reverting after a restart
+        (#4032) without any write-side mirror into meta.json.
+
+        ACP agents are left untouched: they do not use the ``switch_llm`` /
+        condenser path, and their model switch already persists to meta.json.
+        """
+        stored_agent = self.stored.agent
+        if isinstance(stored_agent, ACPAgent):
+            return stored_agent
+
+        base_state_file = self.conversation_dir / BASE_STATE
+        if not base_state_file.exists():
+            return stored_agent  # fresh conversation
+
+        context = {"cipher": self.cipher} if self.cipher else None
+        persisted = ConversationState.model_validate_json(
+            base_state_file.read_text(), context=context
+        ).agent
+        if isinstance(persisted, ACPAgent):
+            return stored_agent
+
+        return stored_agent.model_copy(
+            update={"llm": persisted.llm, "condenser": persisted.condenser}
+        )
+
     def _without_stored_secret(self, secret_name: str) -> StoredConversation:
         secrets = dict(self.stored.secrets)
         secrets.pop(secret_name, None)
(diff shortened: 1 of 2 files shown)


enyst (commented):

🟢 **Good taste** - this fixes the source-of-truth problem at the read boundary instead of maintaining a second write-side mirror. The resume path preserves creation-time plugin-merge inputs while taking only the live-switchable LLM/condenser state from `base_state.json`, and the round-trip tests cover restart behavior.

**[RISK ASSESSMENT]**
The implementation is focused and well-tested, but it changes authoritative persisted state during conversation resume, a core runtime path with moderate blast radius.

**VERDICT:**
✅ **Worth merging**: the source-of-truth choice is coherent and the regression coverage exercises the real persistence boundary.

**KEY INSIGHT:**
Reading live-switchable state from its authoritative snapshot is simpler and safer than synchronizing duplicate persistence records.

enyst:
Superseded by #4440 , which solves the underlying architectural issue: it cleans up a smell in the inheritance relationship between `StoredConversation` and `StartConversationRequest`. The result is that `base_state.json` naturally holds the Conversation and Agent, as it always did in the SDK.

Issue #4224, opened 2026-07-27
Cleanup: remove dead code, dedup shared logic, and trim line count


VascoSch92:
## Why

A few areas carry dead code, copy-pasted blocks that risk drifting apart, and spots that can be expressed in fewer lines. Cleaning these up shrinks the maintenance surface and removes copy-paste bug risk. Every item below was verified against the current tree (whole-tree `git grep`, byte-level `diff`s, and `ruff --diff`).

## Dead code — verified unreferenced (only the definition exists anywhere in the tree)

- [x] `_AgentProtocol` — `openhands-sdk/openhands/sdk/agent/response_dispatch.py:85-127` (~43 lines; only referenced in a docstring, so update that docstring too)
- [x] `ConversationResponse` — `openhands-agent-server/openhands/agent_server/models.py:401`
- [x] `GenerateTitleRequest` / `GenerateTitleResponse` — `openhands-agent-server/openhands/agent_server/models.py:549,560` (not wired to any route)
- [x] `get_category()` — `openhands-sdk/openhands/sdk/critic/impl/api/taxonomy.py:50`
- [x] `predict_labels()` — `openhands-sdk/openhands/sdk/critic/impl/api/client.py:334`
- [x] `model_dump_succint()` — `openhands-sdk/openhands/sdk/agent/base.py:723` (⚠️ public method on `AgentBase` — treat as a minor public-API removal; note the `succint` typo)

## Duplication → reuse

- [x] Import `SkillInfo` from the SDK instead of redefining it. `agent_server/skills_router.py:131` duplicates `sdk/skills/skill.py:96` field-for-field, and the SDK class has no extra validators/methods/config, so the OpenAPI schema is unchanged.
- [ ] Extract the **byte-identical** port-resolution block shared by `workspace/apptainer/workspace.py:164-181` and `workspace/docker/workspace.py:185-202` into a helper on `RemoteWorkspace` (both subclass it).
- [x] Extract the **byte-identical** `extra_body` + call-context/session-id injection tail shared by `llm/options/chat_options.py:103-119` and `llm/options/responses_options.py:90-106`.
- [x] De-duplicate the Gemini `edit`/`write_file` observation `visualize` property (`tools/gemini/edit/definition.py`, `tools/gemini/write_file/definition.py`) via a shared base/mixin (blocks are similar; message strings differ).

## Out of scope (intentional)

[...]

Issue #3443, opened 2026-05-29
Shared LLM/Agent objects across conversations inherit stale per-conversation state


csmith49:
## Summary

When the same `LLM` or `Agent` object is reused across multiple `LocalConversation` instances, per-conversation state mutated onto the LLM leaks between conversations. Two values are affected:

- **`_prompt_cache_key`** — set in `_pin_prompt_cache_key()`, guarded by `is None`
- **`x-litellm-session-id` in `extra_headers`** — set in `_pin_session_affinity_header()` (#3437), guarded by `not in existing`

Both use the same pattern: mutate the LLM object, skip if already set. The second conversation sees the first's value and keeps it.

## Repro

```python
from openhands.sdk.llm import LLM
from openhands.sdk.agent import Agent
from openhands.sdk.conversation import LocalConversation

llm = LLM(model="openai/gpt-4o-mini", api_key="test")
agent = Agent(llm=llm, tools=[])

conv1 = LocalConversation(agent=agent, workspace="/tmp/ws1")
conv1._ensure_agent_ready()

conv2 = LocalConversation(agent=agent, workspace="/tmp/ws2")
conv2._ensure_agent_ready()

# Both fail — conv2 inherited conv1's values
assert agent.llm._prompt_cache_key == str(conv2.id)                          # still conv1.id
assert agent.llm.extra_headers["x-litellm-session-id"] == str(conv2.id)      # still conv1.id
```

Examples in this repo that reuse LLM objects across conversations:
- `examples/01_standalone_sdk/24_planning_agent_workflow.py` — one `llm` used for planning + execution conversations
- `examples/01_standalone_sdk/31_iterative_refinement.py` — one `llm` reused across refactoring/critique conversations

## Practical impact

Low in production — the Cloud/server path (`EventService.start()`) always deserializes a fresh agent from JSON, so shared references don't occur. The aliasing only affects standalone SDK usage where callers construct one `LLM` and pass it to multiple `LocalConversation` instances.

For `_prompt_cache_key`: the effect is two conversations sharing an OpenAI prefix-cache shard, which is benign (prefix matching is content-based, not key-based).

For `x-litellm-session-id`: the effect is two conversations routing to the same LiteLLM deployment, which defeats per-conversation probabilistic distribution. Only matters when the LiteLLM proxy has `session_affinity` enabled with multiple weighted deployments.

## Prior art

[...]

Issue #3138, opened 2026-05-08
bug: os.environ credential bleed across conversations in multi-tenant server


csmith49:
## Problem

The LLM's `model_validator` writes API keys and base URLs directly into `os.environ`. In a multi-tenant server where different conversations may use different LLM providers/keys, one conversation's credentials are visible to all others via the process environment. Combined with the `modify_params` race (#4 in tracking issue), a conversation could inadvertently use another tenant's credentials.

**Location:** `llm/llm.py` model_validator

## Evidence

Code review of `llm.py` model_validator. Environment variables are the mechanism for passing credentials to litellm.

**Severity:** Critical

## Impact of fix

Tenant isolation for credentials in multi-conversation servers.

## Suggested fix

Pass credentials through litellm's function arguments or per-call configuration instead of `os.environ`. Medium effort (requires understanding litellm's credential passing API).

## Related issues and PRs

- #2944 — Command-scoped secrets persist across TerminalExecutor session boundaries (open issue, same class of credential leakage)
- #2814 — Support Laminar API key injection to prevent credential leaks (closed PR)
- #2984 — feat(acp): secrets in agent_context — prompt awareness + subprocess injection (closed PR)
- #2152 — Security: Credentials in plugin source URLs exposed in logs, exceptions, and persisted state (open issue)

_Discovered during profiling investigation (code review), May 2026. openhands-sdk v1.19.1._

PR #4276, opened 2026-07-27
Move duplicated LLM option blocks into common.py

onatozmenn:
HUMAN:

I picked one item off #4224. While I was in there I found a third block that is duplicated in the same way but is not listed in the issue, so I did that one too. Since this is a refactor and unit tests on their own do not prove much, I also diffed the real output of both functions before and after. Details below.

---

AGENT:

## Why

`select_chat_options` and `select_responses_options` each carried three blocks of logic that were the same in both files, line for line:

1. propagate `extra_headers` from the LLM config, then merge the OpenRouter `HTTP-Referer` / `X-Title` headers
2. pass `litellm_extra_body` through as `extra_body`
3. inject the call context, so `prompt_cache_key` and the `x-litellm-session-id` header

The issue points at 2 and 3 (`chat_options.py:103-119` and `responses_options.py:90-106`). Block 1 sits a bit higher up in both files and is duplicated in exactly the same way, so I pulled that one out as well.

The risk with copies like this is the usual one. Someone fixes a header ordering bug in the chat path, the Responses path keeps the old behaviour, and it only turns up later as a bug report.

## Summary

- Added `apply_extra_headers`, `apply_extra_body` and `apply_call_context` to `llm/options/common.py`, right next to the existing `apply_defaults_if_absent`. They follow the contract that file already documents: pure, deterministic, returns a new dict, does not mutate the input.
- `select_chat_options` and `select_responses_options` now call those helpers instead of each keeping a copy.
- No behaviour change. The merge order is untouched, so user supplied headers still win over the OpenRouter ones (`{**openrouter_headers, **existing}`) and the session id header is still merged last (`{**existing, "x-litellm-session-id": ...}`).

## Issue Number

Part of #4224

## How to Test

**1. Tests.** Everything that touches these code paths passes:

```
uv run pytest tests/sdk/llm \
  tests/sdk/conversation/local/test_call_context_isolation.py \
  tests/sdk/context/condenser/test_llm_summarizing_condenser.py \
  tests/sdk/config/test_llm_config.py \
  tests/sdk/test_settings.py \
  tests/tools/task/test_task_manager.py
```

```
1115 passed, 30 warnings in 71.24s (0:01:11)
```

**2. Output comparison.** This is the part that actually shows nothing changed. I wrote a throwaway script that builds 11 `LLM` configs covering every branch in the three blocks: config headers only, user headers overriding config headers, OpenRouter with and without user headers, OpenRouter where the config also sets `X-Title`, `extra_body`, call context with and without a cache key, a context session id landing on top of headers the user already set, and all of it at once. For each config it dumps `select_chat_options` with `has_tools` both ways and `select_responses_options` with `store` as `None`, `True` and `False`, then prints the input dict again to confirm nothing got mutated. That is 66 lines of output.

I ran it on this branch, then ran the exact same script against the old code with `git checkout upstream/main -- openhands-sdk/openhands/sdk/llm/options/`, and compared the two:

```
before lines: 66
after lines : 66
IDENTICAL
before sha: 1560105955F1FD772905BC262A3228B217996BE732B2215D45C238C9A4252FE9
after  sha: 1560105955F1FD772905BC262A3228B217996BE732B2215D45C238C9A4252FE9
```

<details>
<summary>The script, if you want to run it yourself</summary>

```python
import json

from openhands.sdk.llm.llm import LLM, LLMCallContext
from openhands.sdk.llm.options.chat_options import select_chat_options
from openhands.sdk.llm.options.responses_options import select_responses_options


def llm(**overrides):
    base = dict(
        model="gpt-4o",
        api_key=SecretStr("test_key"),
        usage_id="equiv-check",
    )
    base.update(overrides)
    return LLM(**base)


CASES = [
    ("plain", llm(), {}, None),
    ("llm_extra_headers", llm(extra_headers={"X-From-Config": "1"}), {}, None),
    (
        "user_headers_beat_config",
        llm(extra_headers={"X-From-Config": "1"}),
        {"extra_headers": {"X-From-User": "2"}},
        None,
    ),
    (
        "openrouter_both",
        llm(
            model="openrouter/anthropic/claude-3.5-sonnet",
            openrouter_site_url="https://example.test",
            openrouter_app_name="EquivCheck",
        ),
        {},
        None,
    ),
    (
        "openrouter_user_headers_win",
        llm(
            model="openrouter/anthropic/claude-3.5-sonnet",
            openrouter_site_url="https://example.test",
            openrouter_app_name="EquivCheck",
        ),
        {"extra_headers": {"HTTP-Referer": "https://user.test", "X-Own": "keep"}},
        None,
    ),
    (
        "openrouter_plus_config_headers",
        llm(
            model="openrouter/anthropic/claude-3.5-sonnet",
            openrouter_site_url="https://example.test",
            openrouter_app_name="EquivCheck",
            extra_headers={"X-From-Config": "1", "X-Title": "config-wins"},
        ),
        {},
        None,
    ),
    ("extra_body", llm(litellm_extra_body={"custom": {"nested": [1, 2]}}), {}, None),
    (
        "ctx_cache_key_and_session",
        llm(),
        {},
        LLMCallContext(prompt_cache_key="cache-abc", session_id="sess-xyz"),
    ),
    (
        "ctx_session_merges_after_headers",
        llm(extra_headers={"X-From-Config": "1"}),
        {},
        LLMCallContext(prompt_cache_key="", session_id="sess-xyz"),
    ),
    (
        "ctx_overwrites_existing_session_header",
        llm(),
        {"extra_headers": {"x-litellm-session-id": "user-set"}},
        LLMCallContext(prompt_cache_key=None, session_id="sess-xyz"),
    ),
    (
        "everything_at_once",
        llm(
            model="openrouter/anthropic/claude-3.5-sonnet",
            openrouter_site_url="https://example.test",
            openrouter_app_name="EquivCheck",
            extra_headers={"X-From-Config": "1"},
            litellm_extra_body={"custom": True},
        ),
        {"extra_headers": {"X-From-User": "2"}},
        LLMCallContext(prompt_cache_key="cache-abc", session_id="sess-xyz"),
    ),
]


def render(value):
    try:
        return json.dumps(value, sort_keys=True, default=repr)
    except TypeError:
        return repr(value)


for name, model, user_kwargs, ctx in CASES:
    for has_tools in (True, False):
        out = select_chat_options(
            model, dict(user_kwargs), has_tools=has_tools, call_context=ctx
        )
        print(f"chat/{name}/tools={has_tools}: {render(out)}")

    for store in (None, True, False):
        out = select_responses_options(
            model, dict(user_kwargs), include=None, store=store, call_context=ctx
        )
        print(f"responses/{name}/store={store}: {render(out)}")

    print(f"input-not-mutated/{name}: {render(user_kwargs)}")
```

</details>

**3. Hooks.** `ruff check`, `ruff format --check`, `pyright` and `scripts/check_import_rules.py` are all clean on the three changed files.

## Video/Screenshots

Nothing to show, there is no UI here. The output comparison above is the evidence instead.

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The diff is +85/-57, so the line count goes up a little. That is because the helpers carry docstrings and I kept the existing comments that explain #3138 and #3443. Happy to trim them if you would rather keep it tight, the point of this one was removing the copies rather than the line count.

Two items from #4224 I left alone on purpose:

- The port resolution block shared by `docker/workspace.py` and `apptainer/workspace.py`. The tests patch `check_port_available` and `find_available_tcp_port` at both module paths, so moving the block behind a shared helper would quietly stop those patches from applying. Hanging it off `RemoteWorkspace` would also mean sdk importing from `openhands.workspace`, which is the wrong direction for the package layering. That one needs its own PR and a decision on where the helper should live.
- Dropping `model_dump_succint()`, since that is a public API removal and should be your call rather than mine.

Happy to take either of those next if you tell me which shape you want.

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/options/chat_options.py
+++ b/openhands-sdk/openhands/sdk/llm/options/chat_options.py
@@ -2,7 +2,12 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
-from openhands.sdk.llm.options.common import apply_defaults_if_absent
+from openhands.sdk.llm.options.common import (
+    apply_call_context,
+    apply_defaults_if_absent,
+    apply_extra_body,
+    apply_extra_headers,
+)
 from openhands.sdk.llm.utils.model_features import get_features
 
 
@@ -36,17 +41,7 @@ def select_chat_options(
         if "max_completion_tokens" in out:
             out["max_tokens"] = out.pop("max_completion_tokens")
 
-    # If user didn't set extra_headers, propagate from llm config
-    if llm.extra_headers is not None and "extra_headers" not in out:
-        out["extra_headers"] = dict(llm.extra_headers)
-
-    # Inject OpenRouter HTTP-Referer / X-Title via extra_headers so we don't
-    # have to mutate os.environ (which would leak across conversations in a
-    # multi-tenant server; see issue #3138). User-supplied headers win.
-    openrouter_headers = llm._openrouter_headers()
-    if openrouter_headers:
-        existing = out.get("extra_headers") or {}
-        out["extra_headers"] = {**openrouter_headers, **existing}
+    out = apply_extra_headers(out, llm)
 
     # Reasoning-model quirks
     supports_reasoning_effort = get_features(llm.model).supports_reasoning_effort
@@ -99,21 +94,7 @@ def select_chat_options(
     ):
         out["prompt_cache_retention"] = llm.prompt_cache_retention
 
-    # Pass through user-provided extra_body unchanged
-    if llm.litellm_extra_body:
-        out["extra_body"] = llm.litellm_extra_body
-
-    # Inject per-conversation state from call context (#3443).
-    # Prefer explicitly threaded context; fall back to PrivateAttr for
-    # callers that don't thread (e.g. condenser's dedicated LLM).
-    ctx = call_context or llm._call_context
-    if ctx.prompt_cache_key:
-        out["prompt_cache_key"] = ctx.prompt_cache_key
-    if ctx.session_id:
-        existing = out.get("extra_headers") or {}
-        out["extra_headers"] = {
-            **existing,
-            "x-litellm-session-id": ctx.session_id,
-        }
+    out = apply_extra_body(out, llm)
+    out = apply_call_context(out, llm, call_context)
 
     return out
--- a/openhands-sdk/openhands/sdk/llm/options/common.py
+++ b/openhands-sdk/openhands/sdk/llm/options/common.py
@@ -1,6 +1,10 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
+
+
+if TYPE_CHECKING:
+    from openhands.sdk.llm.llm import LLM, LLMCallContext
 
 
 def apply_defaults_if_absent(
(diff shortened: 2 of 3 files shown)


VascoSch92 (approved):
THank you

PR #2790, opened 2026-04-09
fix(security): first-message WebSocket auth to prevent token leakage

simonrosenberg:
## Summary

- `_accept_authenticated_websocket` now supports **first-message auth**: when no key is found in query params or headers, the server accepts the WebSocket, reads the first frame, and validates `{"type": "auth", "session_api_key": "..."}` before processing any other messages. 10-second timeout for the auth frame (rationale documented inline).
- Query parameter and header-based auth are kept as **deprecated fallbacks** so existing clients continue to work (with a deprecation warning logged).
- `_safe_close_websocket` now accepts optional `code`/`reason` parameters.
- **Operational logging:** `logger.info` on successful first-message auth (parity with the legacy deprecation log) and per-reason `logger.warning` on failures (timeout, malformed JSON, client disconnect, non-object payload, wrong type, invalid key) — client-facing close reason stays generic ("Authentication failed") so we don't leak diagnostics over the wire.
- **Tests:** 15 mock-based unit tests in `test_websocket_first_message_auth.py` (covers no-auth, legacy query/header valid+invalid, legacy-precedence-over-first-message, first-message valid/invalid/wrong-type/missing-key/malformed-JSON/disconnect/timeout, plus end-to-end through `events_socket`) **and** 3 live-server tests in `test_agent_server_wsproto.py` using a real wsproto/websockets client to exercise the actual handshake → first-frame-as-auth → normal message flow transition (accepted, rejected with close 4001, malformed-JSON close 4001).

Addresses the server-side component of OpenHands/OpenHands#13777 — moves the first-message auth from the V0 `listen_socket.py` to the SDK as requested in [tofarr's review](https://github.com/OpenHands/OpenHands/pull/13777#pullrequestreview-4062486239).

## Test plan

- [x] All 19 existing WebSocket tests pass (no regressions)
- [x] 15 new mock-based first-message auth tests pass
- [x] 3 new live-server (real wsproto client) first-message auth tests pass — accepted / rejected (4001) / malformed (4001)
- [x] Pre-commit (ruff-format, ruff-lint, pyright) passes on changed files
- [ ] Frontend update (OpenHands/OpenHands#13777) sends `{"type": "auth", ...}` as first frame — will work once this SDK is released and the submodule is bumped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
**Agent Server images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-sdk/pkgs/container/agent-server

**Variants & Base Images**
[...]

PR #4279, opened 2026-07-27
fix(security): authenticate WebSockets outside URLs

simonrosenberg:
HUMAN:

Remove session_api_key from Python SDK WebSocket URLs.

---

AGENT:

## Why

Session API keys in WebSocket query strings are copied into reverse-proxy and access logs. Production log review confirmed credential-bearing event-socket URLs.

## Summary

- Remove session_api_key from Python SDK WebSocket URLs.
- Send the existing first-message auth payload immediately after connection.
- Verify the URL is clean and the auth payload is the first client frame.

## Issue Number

N/A

## How to Test

- uv run pytest tests/sdk/conversation/remote/test_websocket_client.py tests/agent_server/test_agent_server_wsproto.py::test_agent_server_websocket_first_message_auth_accepted -q
  - 10 passed.
  - This covers both the updated client frame and a real agent-server WebSocket accepting first-message authentication.
- uv run pre-commit run --files openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py tests/sdk/conversation/remote/test_websocket_client.py
  - All hooks passed.

## Video/Screenshots

N/A. This is a wire-protocol security change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Agent-server first-message authentication is already available on main from #2790. The server retains its deprecated query fallback for older clients.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:0a7e7ea-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-0a7e7ea-python \
  ghcr.io/openhands/agent-server:0a7e7ea-python
```

```
ghcr.io/openhands/agent-server:0a7e7ea-golang-amd64
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-golang-amd64
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-golang-amd64
ghcr.io/openhands/agent-server:0a7e7ea-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:0a7e7ea-golang-arm64
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-golang-arm64
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-golang-arm64
ghcr.io/openhands/agent-server:0a7e7ea-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:0a7e7ea-java-amd64
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-java-amd64
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-java-amd64
ghcr.io/openhands/agent-server:0a7e7ea-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:0a7e7ea-java-arm64
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-java-arm64
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-java-arm64
ghcr.io/openhands/agent-server:0a7e7ea-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:0a7e7ea-python-amd64
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-python-amd64
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-python-amd64
ghcr.io/openhands/agent-server:0a7e7ea-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:0a7e7ea-python-arm64
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-python-arm64
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-python-arm64
ghcr.io/openhands/agent-server:0a7e7ea-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:0a7e7ea-golang
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-golang
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-golang
ghcr.io/openhands/agent-server:0a7e7ea-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:0a7e7ea-java
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-java
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-java
ghcr.io/openhands/agent-server:0a7e7ea-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:0a7e7ea-python
ghcr.io/openhands/agent-server:0a7e7ea0234f9fd69c28814dd363e863e63683fe-python
ghcr.io/openhands/agent-server:fix-websocket-auth-frame-python
ghcr.io/openhands/agent-server:0a7e7ea-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `0a7e7ea-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `0a7e7ea-python-amd64`) are also available if needed

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -7,7 +7,7 @@ import time
 import uuid
 from collections.abc import Callable, Mapping
 from queue import Empty, Queue
-from typing import TYPE_CHECKING, SupportsIndex, overload
+from typing import TYPE_CHECKING, Final, SupportsIndex, overload
 from urllib.parse import urlparse
 
 import httpx
@@ -65,8 +65,8 @@ logger = get_logger(__name__)
 
 LEGACY_CONVERSATIONS_PATH = "/api/conversations"
 FATAL_WS_CLOSE_CODES = frozenset({4001, 4004})
-_WEBSOCKET_AUTH_TYPE = "auth"
-_WEBSOCKET_SESSION_API_KEY_FIELD = "session_api_key"
+_WEBSOCKET_AUTH_TYPE: Final = "auth"
+_WEBSOCKET_SESSION_API_KEY_FIELD: Final = "session_api_key"
 
 
 def _agent_kind_mismatch_message(conversation_id: ConversationID) -> str:
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -8,7 +8,7 @@ import uuid
 from collections.abc import Callable, Mapping
 from queue import Empty, Queue
 from typing import TYPE_CHECKING, SupportsIndex, overload
-from urllib.parse import quote, urlparse
+from urllib.parse import urlparse
 
 import httpx
 import websockets
@@ -65,6 +65,8 @@ logger = get_logger(__name__)
 
 LEGACY_CONVERSATIONS_PATH = "/api/conversations"
 FATAL_WS_CLOSE_CODES = frozenset({4001, 4004})
+_WEBSOCKET_AUTH_TYPE = "auth"
+_WEBSOCKET_SESSION_API_KEY_FIELD = "session_api_key"
 
 
 def _agent_kind_mismatch_message(conversation_id: ConversationID) -> str:
@@ -214,15 +216,20 @@ class WebSocketCallbackClient:
         base = f"{ws_scheme}://{parsed.netloc}{parsed.path.rstrip('/')}"
         ws_url = f"{base}/sockets/events/{self.conversation_id}"
 
-        # Add API key as query parameter if provided
-        if self.api_key:
-            ws_url += f"?session_api_key={quote(self.api_key, safe='')}"
-
         delay = 1.0
         has_connected = False
         while not self._stop.is_set():
             try:
                 async with websockets.connect(ws_url) as ws:
+                    if self.api_key:
+                        await ws.send(
+                            json.dumps(
+                                {
+                                    "type": _WEBSOCKET_AUTH_TYPE,
+                                    _WEBSOCKET_SESSION_API_KEY_FIELD: self.api_key,
+                                }
+                            )
+                        )
                     delay = 1.0
                     connection_ready = False
                     async for message in ws:
--- a/tests/cross/test_remote_conversation_live_server.py
+++ b/tests/cross/test_remote_conversation_live_server.py
@@ -57,6 +57,7 @@ def live_server_env(
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
     import_modules: str | None = None,
+    session_api_keys: list[str] | None = None,
 ) -> Generator[dict]:
     """Launch a real FastAPI server backed by temp workspace and conversations.
 
@@ -97,7 +98,7 @@ def live_server_env(
     )
 
     cfg = {
-        "session_api_keys": [],  # disable auth for tests
+        "session_api_keys": session_api_keys or [],
         "conversations_path": str(conversations_path),
         "workspace_path": str(workspace_path),
     }
@@ -203,6 +204,20 @@ def server_env(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Generator[dic
 
 
 @pytest.fixture
+def authenticated_server_env(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> Generator[dict]:
+    api_key = "test-websocket-auth-key"
+    with live_server_env(
+        tmp_path,
+        monkeypatch,
+        session_api_keys=[api_key],
+    ) as env:
+        env["api_key"] = api_key
+        yield env
+
+
+@pytest.fixture
 def patched_llm(monkeypatch: pytest.MonkeyPatch) -> None:
     """Patch LLM.completion to a deterministic assistant message response."""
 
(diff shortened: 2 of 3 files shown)


aivong-openhands:
## Mutation-testing the new suite

I ran mutation testing against `tests/sdk/conversation/remote/test_websocket_client.py` to see which parts of the change the test actually pins down. Baseline: 9 passed.

The rewritten `test_websocket_client_sends_api_key_in_first_frame` is strong on the core claim — all three control mutants die:

| Control mutant | Result |
|---|---|
| Rename the frame type (`"auth"` → `"authenticate"`) | ❌ caught |
| Put the key back in the URL as `?session_api_key=…` | ❌ caught |
| Don't send the auth frame at all | ❌ caught |

Exact-equality assertions on both `captured_urls[0]` and `sent_messages` are what make that work — good choice over substring matching.

What survives is everything about the loop the frame now lives inside:

| Surviving mutant | Result |
|---|---|
| `if self.api_key:` → `if self.api_key and not has_connected:` | ✅ 9 passed |
| Delete `delay = 1.0` (backoff reset on successful connect) | ✅ 9 passed |
| `if self.api_key:` → `if True:` | ⚠️ suite **hangs**, doesn't fail |

---

### 1. Reconnects are not covered — and that's where auth now lives

The auth frame moved *inside* `while not self._stop.is_set()`, so every reconnect has to re-send it. Nothing tests that. Gating the send on the first connection only:

```python
if self.api_key and not has_connected:
    await ws.send(...)
```

leaves the suite fully green, but in production every reconnect after a network blip would hit the server's first-message-auth path with no frame and get closed at the 10s deadline in `_accept_authenticated_websocket`. Since the old query-param URL was inherently re-sent on every reconnect, this is a failure mode the change introduces.

`test_websocket_client_retries_after_retryable_connection_closed` already exercises the reconnect path — extending its mock to record sends is most of the work:

```python
def test_websocket_client_reauthenticates_on_reconnect():
    sent_per_connection = []

    class _MockWebSocket:
        def __init__(self):
            self.sent = []
            sent_per_connection.append(self.sent)

        async def send(self, message):
            self.sent.append(message)

        def __aiter__(self):
            return self

        async def __anext__(self):
            if len(sent_per_connection) >= 2:
                client._stop.set()
                raise StopAsyncIteration
            raise websockets.exceptions.ConnectionClosed(
                rcvd=websockets.frames.Close(1006, "boom"),
                sent=None,
                rcvd_then_sent=False,
            )

    # ... connect() returns a fresh _MockWebSocket each call ...

    assert len(sent_per_connection) == 2
    expected = json.dumps({"type": "auth", "session_api_key": client.api_key})
    assert sent_per_connection == [[expected], [expected]]  # re-auth on every connect
```

### 2. The `if self.api_key:` guard hangs instead of failing

Removing the guard doesn't produce a red test — it spins `test_websocket_client_retries_after_retryable_connection_closed` forever (that test's mock has no `send`, the `AttributeError` is swallowed by the reconnect handler, and the loop never terminates). I had to kill it at 45s.

Two things follow. First, there's no positive assertion that a client with no API key sends *no* frame, which is the behavior the guard exists for:

```python
def test_websocket_client_sends_no_auth_frame_without_api_key():
    # same mock as the first-frame test, api_key=None
    assert sent_messages == []
    assert "session_api_key" not in captured_urls[0]
```

Second, and more general: a mutation in this file degrades into a CI timeout rather than a diagnostic. Worth a `@pytest.mark.timeout(...)` on the reconnect tests, or an iteration cap in the mock, so a regression here reports as a failure instead of a hung job.

### 3. Backoff reset is unpinned

`delay = 1.0` can be deleted with a green suite. That line matters more after this change than before it: the `await ws.send(...)` now sits between `connect()` and the reset, so a send that raises skips the reset and the client keeps backing off. Asserting the reconnect delay sequence in the retry test would cover both.

---

### Not a test gap, but worth a look before merge

The frame's literals (`"auth"`, `"session_api_key"`) are a contract with `_accept_authenticated_websocket` in `openhands-agent-server/.../sockets.py`, which reads `data.get("type") != "auth"` and `data.get("session_api_key")` against its own literals. Each side is asserted against its own copy of those strings, so a rename on one side leaves the suite green. A single shared constant, or one round-trip test against the real `events_socket`, would remove the class of problem.

Also note the server still honours the deprecated query parameter at a higher precedence than the frame (`_resolve_websocket_session_api_key` returns it first). Nothing in this PR is wrong because of that, but the URL-borne key isn't gone from the system until the server drops it too — worth saying in the description so the PR isn't read as closing that off entirely.

This comment was generated by an AI assistant on behalf of the user.

simonrosenberg:
Addressed in 82f671441. The unit coverage now asserts both keyed and keyless connections, verifies auth is sent on every reconnect, and pins the backoff reset to 1 second after each successful connection. The reconnect mock now accepts sends, so removing the key guard fails diagnostically instead of hanging. I also added a client-to-real-agent-server round-trip test to pin the shared first-message-auth contract. The PR Notes already call out that the server retains the deprecated query fallback for older clients.

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -65,6 +65,8 @@
 
 LEGACY_CONVERSATIONS_PATH = "/api/conversations"
 FATAL_WS_CLOSE_CODES = frozenset({4001, 4004})
+_WEBSOCKET_AUTH_TYPE = "auth"
```
VascoSch92: use Final from typing here. In this way if we change that in the code for error we catch that.

VascoSch92 (approved):
Just one thing to be sure it is wanted: Dropping the query param outright means the client now only works against agent-server v1.19.0 or newer. Is that correct?

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
simonrosenberg: this should be added in my md files sorry

simonrosenberg: Applied in 0a7e7ea02: both WebSocket protocol literals are now typed as Final.

simonrosenberg:
On the compatibility question: the exact floor is agent-server **v1.18.1**, not v1.19.0. First-message WebSocket auth landed in #2790 and was released in v1.18.1; v1.18.0 does not support it. This floor applies when session-key authentication is configured. No-key clients send no auth frame and remain compatible with older servers. Dropping the query parameter for authenticated connections is intentional so we do not retain the credential-leaking fallback.

PR #4280, opened 2026-07-27
fix(security): stop logging runtime command contents

simonrosenberg:
HUMAN:

stop logging runtime command contents

---

AGENT:

## Why

Production runtime paths logged complete shell commands and terminal output. Those fields can contain injected credentials. A seven-day production log review found 995 timeout records from the affected path with credential-bearing commands.

## Summary

- Replace raw command and terminal-content logs with command IDs, lengths, timeouts, and exception types.
- Preserve command execution, timeout signaling, and returned command results.
- Add regression coverage with fake token-shaped values and assert they never reach captured logs.
## Issue Number

N/A

## How to Test

- uv run pytest tests/agent_server/test_bash_service.py tests/sdk/workspace/remote/test_remote_workspace_mixin.py tests/tools/terminal/test_ps1_corruption.py
  - 44 passed.
  - The agent-server test sends an HTTP request through the ASGI app, runs a real timed-out bash subprocess, confirms the SIGTERM cleanup trap ran, and confirms the fake credential is absent from logs.
- uv run pre-commit run --files openhands-agent-server/openhands/agent_server/bash_service.py openhands-sdk/openhands/sdk/workspace/remote/remote_workspace_mixin.py openhands-tools/openhands/tools/terminal/terminal/terminal_session.py tests/agent_server/test_bash_service.py tests/sdk/workspace/remote/test_remote_workspace_mixin.py tests/tools/terminal/test_ps1_corruption.py
  - All hooks passed.

## Video/Screenshots

N/A. This changes server-side logging only.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No secret values or production log payloads are included in this PR.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:95a1bb6-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-95a1bb6-python \
  ghcr.io/openhands/agent-server:95a1bb6-python
```

```
ghcr.io/openhands/agent-server:95a1bb6-golang-amd64
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-golang-amd64
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-golang-amd64
ghcr.io/openhands/agent-server:95a1bb6-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:95a1bb6-golang-arm64
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-golang-arm64
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-golang-arm64
ghcr.io/openhands/agent-server:95a1bb6-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:95a1bb6-java-amd64
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-java-amd64
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-java-amd64
ghcr.io/openhands/agent-server:95a1bb6-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:95a1bb6-java-arm64
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-java-arm64
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-java-arm64
ghcr.io/openhands/agent-server:95a1bb6-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:95a1bb6-python-amd64
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-python-amd64
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-python-amd64
ghcr.io/openhands/agent-server:95a1bb6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:95a1bb6-python-arm64
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-python-arm64
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-python-arm64
ghcr.io/openhands/agent-server:95a1bb6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:95a1bb6-golang
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-golang
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-golang
ghcr.io/openhands/agent-server:95a1bb6-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:95a1bb6-java
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-java
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-java
ghcr.io/openhands/agent-server:95a1bb6-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:95a1bb6-python
ghcr.io/openhands/agent-server:95a1bb6e8d4ab86818dd7ea1dc18c70d423026e6-python
ghcr.io/openhands/agent-server:fix-sensitive-runtime-logs-python
ghcr.io/openhands/agent-server:95a1bb6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `95a1bb6-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `95a1bb6-python-amd64`) are also available if needed


openhands-tools/openhands/tools/terminal/terminal/terminal_session.py
```diff
@@ -534,7 +550,7 @@ def execute(self, action: TerminalAction) -> TerminalObservation:
                 if not self.terminal.is_powershell():
                     # Only escape for bash terminals, not PowerShell
                     command = escape_bash_special_chars(command)
-                logger.debug(f"SENDING COMMAND: {command!r}")
+                logger.debug("Sending command (command_length=%s)", len(command))
```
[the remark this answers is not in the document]
simonrosenberg: Resolved in 73e508e31c4628cd47c1e3f650b4580efb0d5393. The SubprocessTerminal backend now logs only byte/content lengths for PTY writes and reads, so neither command input nor terminal output is emitted at DEBUG level. I also added a regression test that runs a real subprocess terminal command, confirms the token-shaped value is still returned in the observation, and asserts it is absent from captured logs. The focused SDK verification passed (20 tests).

aivong-openhands:
## Mutation-testing the new suite

I ran mutation testing against the five test files this PR touches to see which of the scrubbed log sites are actually pinned down. Baseline: 64 passed.

The test pattern here is the right one — plant a known secret, exercise the code, then assert both that the log line still exists *and* that the secret doesn't appear in `caplog.text`. It works: every control mutant that re-introduces a leak on a covered path dies.

| Control mutant (re-introduces a leak) | Result |
|---|---|
| `_logger.debug(f"Executing remote command: {command}")` | ❌ caught |
| `logger.debug(f"Wrote to subprocess PTY: {data!r}")` | ❌ caught |
| Log `terminal_content` in the no-change-timeout warning | ❌ caught |
| Log `command.command` in the bash timeout warning | ❌ caught |
| Log `terminal_content` in `_handle_completed_command` | ❌ caught |
| Stop skipping redundant auth frames in `bash_events_socket` | ❌ caught |

Four mutants survive with a fully green suite. Three of them share a shape: **every covered log site is on a success or timeout path, and none of the exception handlers are exercised.**

| Surviving mutant | Result |
|---|---|
| Restore `f"Error executing bash command '{command.command}': {e}"` | ✅ 64 passed |
| Restore `f"Remote command execution failed: {e}"` | ✅ 64 passed |
| Restore `f"Failed to kill process for command: {command.command}"` | ✅ 64 passed |
| Drop the `"command"`/`"role"` narrowing in `_is_auth_control_message` | ✅ 64 passed |

---

### 1. The exception handlers are the least-tested and most leak-prone sites

`bash_service._start_bash_command`'s `except Exception` and `remote_workspace_mixin._execute_command_generator`'s `except Exception` can both go back to interpolating the full command with the suite green. Error paths are exactly where this class of bug originates — a handler reaches for every variable in scope when something goes wrong — so they're worth the same treatment the happy paths got.

`test_execute_command_generator_exception_handling` already exists and already forces the exception; it needs three lines:

```python
def test_execute_command_generator_exception_handling(caplog):
    secret = "ghp_" + "e" * 36
    caplog.set_level(logging.DEBUG)
    # ... existing setup, with the command carrying the secret ...
    generator = mixin._execute_command_generator(
        f"curl -H 'Authorization: Bearer {secret}' example.test", None, 30.0
    )
    # ... existing assertions ...
    assert "Remote command execution failed" in caplog.text
    assert secret not in caplog.text
```

For `bash_service`, patching `_pub_sub` (or `create_subprocess_exec`) to raise once gets you into the handler with a secret-bearing command already in scope.

The SIGKILL-failed branch (`"Failed to kill process"`) is the lowest-value of the three — reaching it means a process survived `SIGKILL` and two 1s waits — so covering it may not be worth the mock. Worth knowing it's uncovered rather than assuming it isn't.

### 2. The one behaviour change in the PR is untested

Narrowing `_is_auth_control_message` with `and "command" not in data and "role" not in data` is the only non-logging change here, and it can be deleted with a green suite. `test_bash_socket_ignores_redundant_auth_before_command` doesn't reach it: its second frame is `{"command": "printf queued-command"}` with no `type`, so `data.get("type") == "auth"` is already `False` and the new clauses never decide anything.

What the narrowing actually prevents is a bash payload that carries `type: "auth"` being silently swallowed by `continue` instead of executed. This test covers it — I confirmed it passes on this branch and fails with the narrowing removed:

```python
@pytest.mark.asyncio
async def test_bash_socket_runs_command_payload_that_declares_auth_type():
    """A bash payload carrying type="auth" must still execute, not be skipped."""
    ws = _make_mock_websocket(headers={})
    ws.receive_json.side_effect = [
        {"type": "auth", "command": "printf hi"},
        WebSocketDisconnect(),
    ]
    mock_bash_service = MagicMock(spec=BashEventService)
    mock_bash_service.subscribe_to_events = AsyncMock(return_value=uuid4())
    mock_bash_service.unsubscribe_from_events = AsyncMock(return_value=True)
    mock_bash_service.start_bash_command = AsyncMock()

    with (
        patch("openhands.agent_server.sockets.bash_event_service", mock_bash_service),
        patch("openhands.agent_server.sockets.get_default_config") as mock_config,
    ):
        mock_config.return_value.session_api_keys = []
        await bash_events_socket(
            ws, session_api_key=None, resend_mode=None, resend_all=False
        )

    mock_bash_service.start_bash_command.assert_awaited_once()
    assert mock_bash_service.start_bash_command.await_args.args[0].command == "printf hi"
```

A matching negative — `{"type": "auth", "role": "user"}` on the events socket — would pin the other half.

---

### Not a test gap: the strongest fix in this PR isn't a test

Dropping the `command` parameter from `_signal_process_group` entirely is the best change here. The value isn't in scope, so no future edit can leak it and no test is needed to defend it. That's worth applying more widely than it was: `_execute_command_generator` and `_start_bash_command` both still hold the full command in scope at their error-log sites, which is why those two mutants were easy to write.

On the lint angle — ruff's `G004` (f-string in logging call) would flag re-introduction mechanically, but it's a blunt instrument here: `select` is currently `["E","F","I","UP","ARG"]`, and turning on `G004` reports **30 violations in these five files alone and 672 across `openhands-sdk` + `openhands-tools` + `openhands-agent-server`**. Most are harmless (`f"Using shell: {resolved_shell_path}"`). So I wouldn't propose it repo-wide off the back of this PR — but a per-file `# ruff: noqa`-inverse (enabling `G004` for just the modules that handle command text) would lock in the win for ~30 mechanical fixes.

One last thing worth checking before merge: `subprocess_terminal.py` still has `logger.error(f"Failed to write to PTY: {e}", exc_info=True)` at line 255 and `logger.debug(f"Error reading PTY output: {e}")` at 288, both on the data path this PR was scrubbing. They log the exception rather than the buffer, so they're probably fine — but they're the two sites nearest the ones you fixed, and neither is covered.

This comment was generated by an AI assistant on behalf of the user.

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -493,14 +495,13 @@ async def _send_event(event: Event, websocket: WebSocket):
...
 def _is_auth_control_message(data: object) -> bool:
-    """Return True for ``{"type": "auth", ...}`` first-message-auth frames.
-
-    Clients that handle both legacy and first-message auth may send this
-    frame even after legacy (query/header) auth has already succeeded.
-    The post-auth receive loops must ignore it instead of validating it
-    as a regular message payload.
-    """
-    return isinstance(data, dict) and data.get("type") == "auth"
+    """Identify auth control frames that are not application payloads."""
```
VascoSch92: If you rename `ExecuteBashRequest.command` and auth frames start leaking the session key again. Suggest `set(data) <= {"type", "session_api_key"}`, plus the original "why".

openhands-tools/openhands/tools/terminal/terminal/terminal_session.py
```diff
@@ -339,9 +341,10 @@ def _handle_hard_timeout_command(
         self.prev_status = TerminalCommandStatus.HARD_TIMEOUT
         if len(ps1_matches) != 1:
             logger.warning(
-                f"Expected exactly one PS1 metadata block BEFORE the execution of a "
-                f"command, but got {len(ps1_matches)} PS1 metadata blocks:\n"
-                f"---\n{terminal_content!r}\n---"
+                "Expected exactly one PS1 metadata block before command execution, "
```
VascoSch92, line 344: we are dropping the terminal content but then 3 lines later we write it into the bashoutput that it is persisted. Is this wanted?

openhands-agent-server/openhands/agent_server/sockets.py
simonrosenberg: Fixed in `844e73c5e`: auth control frames now require `set(data) <= {"type", "session_api_key"}`, so detection is independent of application request field names. Added coverage for `command`, `role`, and unknown-field payloads.

openhands-tools/openhands/tools/terminal/terminal/terminal_session.py
simonrosenberg: Yes, intentionally. This PR removes command/output content from centralized logs; `BashOutput` remains per-workspace application state required for polling and replay. Registered secrets are masked before normal conversation persistence; hardening raw event storage would need a separate masking, retention, or encryption decision.

simonrosenberg:
Addressed the review feedback in `844e73c5e`:

- Tightened redundant auth-frame detection to an exact protocol-field allowlist.
- Added regression coverage for auth-looking application payloads.
- Added exception-path tests proving secret-bearing bash commands and remote execution errors do not enter logs.

`BashOutput` persistence is unchanged because it is functional event state, distinct from centralized logging. The focused tests and all 32 hosted code checks pass.

enyst (commented):
Interesting, it looks like this comment is correct? https://github.com/OpenHands/software-agent-sdk/pull/4280#pullrequestreview-4796656123

simonrosenberg:
Fixed the WebSocket send-failure leak in `b54af3bbb`.

`_send_bash_event()` now logs only a fixed allowlist of safe metadata (`kind`, event/command IDs, order, and exit code) plus the exception class name. It no longer logs the event representation, exception message, or exception traceback text on pre-disconnect, expected disconnect-race, or unexpected-exception paths.

Added regression coverage with token-shaped command, stdout/stderr, and exception-message values. All 92 relevant security tests and pre-commit checks pass.

enyst (commented):

🟢 **Good taste**

This removes command, input, output, event-repr, and exception-text leakage at the relevant runtime boundaries while retaining useful identifiers, lengths, timeout values, and exception types for diagnosis. The regression tests use sentinel secrets across bash service, websocket, remote workspace, subprocess terminal, and terminal session paths. I did not find a remaining leak introduced or missed within this PR's changed surfaces.

**[RISK ASSESSMENT]**
The patch is mechanically small but directly changes secrets-handling and production diagnostic behavior across several shared runtime layers. A missed sink would expose user commands or output; over-redaction could impair incident diagnosis.

**Recommendation**: Do not auto-merge. Request review from a human security/runtime maintainer to validate coverage of logging sinks and the diagnostic tradeoff.

**VERDICT:**
✅ **Worth merging**: The redaction strategy is consistent and backed by focused regression coverage.

**KEY INSIGHT:**
Structured metadata such as command IDs, byte counts, and exception types preserves operational value without serializing user-controlled content.

simonrosenberg:
@VascoSch92 can you kindly re-review? :) CI erros are unrelated

VascoSch92 (approved):

VascoSch92:
@simonrosenberg  For me is good. But why the tests are not passing?

simonrosenberg:
> @simonrosenberg For me is good. But why the tests are not passing?

unrelated CI issue I fixed independently :)

simonrosenberg:
@VascoSch92 thanks for the review!

PR #4311, opened 2026-07-30
feat: report accumulated LLM cost in the automation completion callback

hieptl:
HUMAN:

I have verified the changes.

---

AGENT:

## Why

The automation service wants to record what each automation run costs, but the SDK gives it no way to find out. The completion callback sent from `Workspace.__exit__` carries only `status`, `run_id`, `conversation_id` and `error`.

The value is already available inside the sandbox — `conversation.conversation_stats.get_combined_metrics().accumulated_cost` — but nothing moves it onto the workspace that sends the callback. The SDK already has exactly one channel of this shape: `register_conversation()` → `_conversation_id` → `payload["conversation_id"]`. This adds the missing counterpart for cost.

## Summary

- Add `RemoteWorkspace.register_cost()` and an `accumulated_cost` property, mirroring the existing `register_conversation()` / `conversation_id` pair. `OpenHandsCloudWorkspace` inherits both.
- Include `cost` in the callback payload when registered, in both `_send_completion_callback` implementations. The key is omitted when unset, so the change is additive and safe against services that don't accept it yet.
- Have `RemoteConversation.close()` self-register the accumulated cost onto its workspace, so cost is reported with no caller opt-in. It reads **cached state only** — `RemoteState.stats` otherwise falls through to a live REST `GET`, and `close()` also runs on the failure path where the agent server may already be gone. It further requires a real `stats` entry, so an unknown cost stays unreported rather than being recorded as a misleading `0.0`.

## How to Test

Automated:

```bash
uv run pytest tests/sdk/workspace/remote/test_remote_workspace.py \
              tests/workspace/test_cloud_workspace.py \
              tests/sdk/conversation/remote/test_remote_conversation.py -q

```

Red/green was verified explicitly — with the source changes stashed and the new tests kept, 5 of the 7 new tests fail:

```
FAILED tests/sdk/workspace/remote/test_remote_workspace.py::test_register_cost_stores_cost
FAILED tests/sdk/workspace/remote/test_remote_workspace.py::test_send_completion_callback_includes_registered_cost
FAILED tests/workspace/test_cloud_workspace.py::test_callback_includes_registered_cost
FAILED tests/sdk/conversation/remote/test_remote_conversation.py::TestRemoteConversation::test_close_reports_accumulated_cost_to_workspace
FAILED tests/sdk/conversation/remote/test_remote_conversation.py::TestRemoteConversation::test_close_leaves_cost_unreported_when_state_has_no_stats
================= 5 failed, 2 passed, 138 deselected in 0.33s ==================

```

The 2 that pass pre-change are intentional regression guards, not new-behavior tests: `test_send_completion_callback_omits_cost_when_not_registered` locks in that the key stays absent, and `test_close_does_not_fetch_state_to_read_cost` locks in that `close()` issues no HTTP request.

Full affected suites, with the change applied:

```
$ uv run pytest tests/sdk/conversation/ tests/sdk/workspace/ tests/workspace/ -q \
    --ignore=tests/sdk/conversation/test_local_conversation_plugins.py \
    --ignore=tests/sdk/conversation/test_repo_root_project_skills.py
================= 1029 passed, 4 warnings in 114.41s (0:01:54) =================

```

The two ignored files have 8 failures that are **pre-existing and unrelated** — confirmed by running them on a clean tree with all changes stashed, which produces the identical 8 failures.

End-to-end payload check, driving the real callback with a mocked transport (both workspace classes, success and failure paths):

```
RemoteWorkspace COMPLETED payload: {'status': 'COMPLETED', 'run_id': 'run-42', 'cost': 0.4213}
RemoteWorkspace FAILED payload:    {'status': 'FAILED', 'run_id': 'run-42', 'error': 'agent blew up', 'cost': 0.4213}
zero-cost transmitted correctly
CloudWorkspace payload:            {'status': 'COMPLETED', 'run_id': 'run-42', 'conversation_id': 'conv-abc', 'cost': 1.25}

```

`close()` behavior driven directly, including the non-blocking guarantee:

```
cached stats      -> registered cost=0.75
no cached state   -> no HTTP call, cost left unset, close() survived
partial cache     -> cost left unset (unknown), not 0.0
genuine zero-cost -> cost=0.0 recorded
workspace raises  -> swallowed, close() completed
double close      -> second call short-circuits

```

Lint/format: `uv run ruff check` and `ruff format --check` clean across `openhands-sdk/`, `openhands-workspace/`, and `tests/`.

## Video/Screenshots

N/A — no user-visible UI. Evidence is the payload and `close()` transcripts above.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:a381fbc-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-a381fbc-python \
  ghcr.io/openhands/agent-server:a381fbc-python
```

```
ghcr.io/openhands/agent-server:a381fbc-golang-amd64
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-golang-amd64
ghcr.io/openhands/agent-server:hieptl-oss-8650-golang-amd64
ghcr.io/openhands/agent-server:a381fbc-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:a381fbc-golang-arm64
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-golang-arm64
ghcr.io/openhands/agent-server:hieptl-oss-8650-golang-arm64
ghcr.io/openhands/agent-server:a381fbc-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:a381fbc-java-amd64
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-java-amd64
ghcr.io/openhands/agent-server:hieptl-oss-8650-java-amd64
ghcr.io/openhands/agent-server:a381fbc-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:a381fbc-java-arm64
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-java-arm64
ghcr.io/openhands/agent-server:hieptl-oss-8650-java-arm64
ghcr.io/openhands/agent-server:a381fbc-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:a381fbc-python-amd64
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-python-amd64
ghcr.io/openhands/agent-server:hieptl-oss-8650-python-amd64
ghcr.io/openhands/agent-server:a381fbc-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:a381fbc-python-arm64
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-python-arm64
ghcr.io/openhands/agent-server:hieptl-oss-8650-python-arm64
ghcr.io/openhands/agent-server:a381fbc-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:a381fbc-golang
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-golang
ghcr.io/openhands/agent-server:hieptl-oss-8650-golang
ghcr.io/openhands/agent-server:a381fbc-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:a381fbc-java
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-java
ghcr.io/openhands/agent-server:hieptl-oss-8650-java
ghcr.io/openhands/agent-server:a381fbc-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:a381fbc-python
ghcr.io/openhands/agent-server:a381fbccb4be9359c59cef178f1f685d311704b1-python
ghcr.io/openhands/agent-server:hieptl-oss-8650-python
ghcr.io/openhands/agent-server:a381fbc-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `a381fbc-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `a381fbc-python-amd64`) are also available if needed


enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py`, Lines 964-968 and 1692-1698] **Failure-path correctness**: An `ERROR`/`STUCK` field update wakes the blocking caller immediately, before the post-run full-state snapshot that carries final stats. `close()` then sees the `stats` key left by the initial snapshot and registers that stale value—commonly zero—as this run's cost. That is worse than omitting an unknown cost because it records a failed, billable run as free. Do not publish cost until stats are known to belong to the completed run; either wait (with a bounded fallback) for the post-run full-state snapshot before raising, or move the authoritative cost handoff to a server-side completion boundary.

**[TESTING GAPS]**
- [`tests/sdk/conversation/remote/test_remote_conversation.py`] The new test injects a final full-state snapshot and then calls `close()`, so it cannot catch the ordering bug above. Add a regression test that starts with cached zero-cost stats, delivers only `execution_status=error`, closes as the real failure path does, and proves stale cost is never sent; also cover the later final-snapshot case to prove the actual accumulated cost is reported.

**[RISK ASSESSMENT]**
This changes automation accounting data. The happy path is narrow and well-contained, but the failure path can silently emit materially incorrect cost rather than failing visibly.

**VERDICT:**
❌ **Needs rework**: Synchronize cost collection with authoritative run completion before emitting callback data.

**KEY INSIGHT:**
A cached `stats` key proves that stats once existed; it does not prove they describe the run that just ended.

---

>
>
>

hieptl:
> [openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py, Lines 964-968 and 1692-1698] Failure-path correctness: An ERROR/STUCK field update wakes the blocking caller immediately, before the post-run full-state snapshot that carries final stats. close() then sees the stats key left by the initial snapshot and registers that stale value—commonly zero—as this run's cost. That is worse than omitting an unknown cost because it records a failed, billable run as free. Do not publish cost until stats are known to belong to the completed run; either wait (with a bounded fallback) for the post-run full-state snapshot before raising, or move the authoritative cost handoff to a server-side completion boundary.

The observation about the ordering is correct, but I don't think the conclusion applies in this case because the cached stats are not a subscribe-time snapshot.

`EventService._setup_stats_streaming()` (`event_service.py:892`, wired unconditionally at `:1079`) subscribes to each LLM's `Telemetry`, and `Telemetry.on_response()` (`telemetry.py:117`) emits a `ConversationStateUpdateEvent(key="stats", ...)` after every LLM response. The client immediately writes each update into `_cached_state["stats"]`. As a result, by the time a run fails, the cache already reflects the spend accumulated during that run.

The ordering is also preserved. `stats_callback` executes within `LocalConversation.run()`'s `with self._state:` block, so its executor task is queued on the `FIFOLock` before the `except` branch reacquires the same lock to transition the conversation to `ERROR` (`local_conversation.py:1947-1949`). Both operations are processed through the same `AsyncCallbackWrapper`, preserving their ordering.

The only scenario where the cached value remains `0.0` is when no LLM response has completed. In that case, `0.0` is the correct value because `Telemetry` has not recorded any usage either (`telemetry.py:101-102`). As a result, I don't believe the scenario of "recording a failed, billable run as free" can occur.

I'd also prefer not to block `run()` on the post-run snapshot. The server calls `wait_for_pending(30.0)` before publishing that snapshot (`event_service.py:1210-1217`), so doing so would delay error propagation for every caller, which is beyond the scope of this PR. Similarly, moving the handoff to a server-side completion boundary would change the contract between `agent-server` and `automation`, which OSS-5210 intentionally avoided.

To help guard against regressions, I added a test covering this exact sequence. Please see the response to the testing-gap comment for details.

> [tests/sdk/conversation/remote/test_remote_conversation.py] The new test injects a final full-state snapshot and then calls close(), so it cannot catch the ordering bug above. Add a regression test that starts with cached zero-cost stats, delivers only execution_status=error, closes as the real failure path does, and proves stale cost is never sent; also cover the later final-snapshot case to prove the actual accumulated cost is reported.

Good catch on the coverage gap — the failure path wasn't covered by a test.

I've added `test_close_reports_streamed_cost_on_error_only_wakeup`, which reproduces the actual sequence:

* A subscribe-time full-state snapshot with zero-cost stats.
* A streamed per-field stats update (`0.75`).
* A per-field `execution_status=error` update with no post-run snapshot.
* `close()`.

I also updated the assertion to verify the stronger guarantee: the reported value reflects the streamed cost from the current run rather than the initial snapshot's `0.0`. This also serves as a regression check for the streaming mechanism. If `_setup_stats_streaming` is removed, or if `close()` is changed to rely solely on the post-run snapshot, the test will fail.

I verified the behavior with a red/green check by removing the streamed stats update (which resulted in `0.0`) and then restoring it (which correctly reported `0.75`).

The scenario where a later final snapshot is available is already covered by the existing `test_close_reports_accumulated_cost_to_workspace`, so I didn't add a duplicate test for that case.

Thank you very much! 🙏

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -1683,6 +1683,27 @@ def close(self) -> None:
...
+        # a per-field ERROR/STUCK update before the post-run full-state snapshot.
+        try:
+            cached = self._state._cached_state
+            # Require an actual "stats" entry: a cache built only from partial
+            # field updates — e.g. the subscribe-time push for a service with no
+            # live conversation — would otherwise yield 0.0 and record a run as
+            # free when its cost is really just unknown.
+            if cached is not None and "stats" in cached:
+                cost = self._state.stats.get_combined_metrics().accumulated_cost
+                self.workspace.register_cost(cost)
```
malhotra5, line 1707: What if the same agent server runs multiple automations runs? Will the cost be identifiable on a per run basis?

malhotra5 (changes requested):
I think we want to have cost attribution for automations that run locally as well. It seems the current PR only implements this for remote conversations and remote workspaces

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py`, Lines 1695-1703] **The stale-cost race is still present**: the new test manually delivers the stats event before the error event, but the server does not guarantee that order. `EventService._emit_event_from_thread()` submits the stats publication with `run_in_executor()` and discards the future, while `RemoteConversation` wakes immediately on a per-field `ERROR`/`STUCK`; the server's `wait_for_pending()` does not track that unawaited stats task. An initial/prior `stats` entry therefore satisfies the new `"stats" in cached` guard even when the latest response's cost is still queued, and `close()` can report a stale value as authoritative. Track and drain stats publications before terminal status is observable, or hand off cost at the server's authoritative post-run completion boundary; add a regression test that delays stats delivery until after the terminal update and proves stale cost is never emitted.
- [Completion callback contract] **Receiver-first rollout required**: current `OpenHands/automation` main defines `RunCompleteRequest` with `extra="forbid"` and no `cost` field. As soon as this SDK registers cost, that deployed receiver rejects the entire callback with 422, leaving the run completion unrecorded. The coordinated `OpenHands/automation#280` adds the field, but it is still open; merge and deploy the receiver before releasing this SDK behavior, and document/enforce that sequencing.

**[TESTING GAPS]**
- [`tests/sdk/conversation/remote/test_remote_conversation.py`] The follow-up failure-path test injects `stats` synchronously before `execution_status=error`, so it proves the assumption rather than the real asynchronous ordering. Exercise the actual publication boundary or a deterministic delayed-stats race so the test fails under the current fire-and-forget implementation.

**[RISK ASSESSMENT]**
This controls billing/accounting data and the terminal callback that marks automation runs complete. The current race can silently under-report spend, while an out-of-order cross-repository rollout can reject completion callbacks entirely.

**Recommendation**: Do not auto-merge. Request review from a human architect/reviewer to validate authoritative cost ownership and the receiver-first deployment sequence.

**VERDICT:**
❌ **Needs rework**: Synchronize cost with authoritative completion and make the cross-repository rollout backward-compatible.

**KEY INSIGHT:**
A cached stats object proves only that some stats arrived; it does not prove the latest billable response was serialized before terminal completion.

---

>
>
>

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
hieptl: Yes — the value is scoped to an individual conversation, not to the agent server.

`ConversationStats` is stored as part of the conversation's own state. On the client side, `RemoteState` is created for a single `conversation_id` and retrieves its statistics from `GET {conversation_info_base_path}/{conversation_id}` (`remote_conversation.py:494-520, :662-666`).

The server follows the same model. `EventService` maintains a single `self._conversation`, and `_setup_stats_streaming` publishes `key="stats"` using `self._conversation._state.stats` (`event_service.py:892-919`). As a result, an agent server hosting multiple conversations maintains a separate `ConversationStats` instance for each conversation rather than sharing a single object across the server.

Each automation run is also isolated. Every preset-script execution creates a new `Conversation(...)` without a `conversation_id`, resulting in a new conversation and a fresh, empty `RemoteState` cache. No conversation state or statistics are carried over from previous runs on the same agent server.

On the receiving side, the completion callback is keyed by `AUTOMATION_RUN_ID` (`_send_completion_callback`), which the service assigns when launching that specific run (`dispatcher.py:245-246`). Consequently, `complete_run` records the cost against exactly that automation run.

In practice, the lifecycle mirrors how `conversation_id` is already handled today: one workspace context corresponds to one automation run, one conversation, and one completion callback. Since `conversation_id` is already correctly attributed on a per-run basis through this flow, the associated cost is attributed in exactly the same way.

Thank you very much! 🙏

hieptl:
> [openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py, Lines 1695-1703] The stale-cost race is still present: the new test manually delivers the stats event before the error event, but the server does not guarantee that order. EventService._emit_event_from_thread() submits the stats publication with run_in_executor() and discards the future, while RemoteConversation wakes immediately on a per-field ERROR/STUCK; the server's wait_for_pending() does not track that unawaited stats task. An initial/prior stats entry therefore satisfies the new "stats" in cached guard even when the latest response's cost is still queued, and close() can report a stale value as authoritative. Track and drain stats publications before terminal status is observable, or hand off cost at the server's authoritative post-run completion boundary; add a regression test that delays stats delivery until after the terminal update and proves stale cost is never emitted.

The mechanism is described accurately, but I don't believe it leads to the stale-cost scenario described. In practice, I think the remaining window is much narrower than "reporting a stale value as authoritative."

The key point is that the two publications are serialized by the state lock, not just by submission order. `stats_callback` executes within `LocalConversation.run()`'s `with self._state:` block, while `locked_on_event` re-acquires the same `FIFOLock` before invoking `_on_event` (`event_service.py:823-841`). This means it queues behind the lock currently held by `run()`. Afterward, if an exception occurs, `run()` re-acquires that same lock in its `except` branch to transition the conversation status to `ERROR` (`local_conversation.py:1947-1949`). With `FIFOLock` semantics, the stats event is processed first.

Even if a stats event is delivered later than expected, it would still be newer rather than stale. `ConversationStateUpdateEvent.value` is intentionally typed as `Any` (`event/conversation_state.py:33-36`), and `_setup_stats_streaming` passes `self._conversation._state.stats` directly—the live object rather than a snapshot (`event_service.py:916-917`). As a result, when the event is ultimately serialized for the WebSocket, it reflects the accumulated cost at the time of publication. Delaying publication does not cause an older cost value to be transmitted.

The remaining edge case also appears to be bounded, and it results in temporary under-reporting rather than an incorrect zero value. If a stats event were somehow delayed, the client-side cache would still contain the value from the previous streamed update. In that case, the discrepancy would be limited to the final response's incremental cost on an already-failed run. The "billable run recorded as free" scenario would require the very first response to both incur cost and lose the race. In that situation, the previously cached value would simply be the initial subscription snapshot for a brand-new conversation, which is legitimately `0.0`.

Regarding the proposed remedies, I think both are broader in scope than this PR. Extending `wait_for_pending()` to track `_emit_event_from_thread` futures would change the publication contract for all thread-emitted events, including LLM completion logs and ACP activity heartbeats, rather than just stats. Likewise, moving the handoff to a server-side completion boundary would alter the contract between the agent server and the automation service, which OSS-5210 intentionally avoided.

Given that this field provides best-effort, additive accounting, and that the receiving side already models missing values as `NULL` (meaning "unknown"), my preference would be to avoid introducing additional latency or expanding the protocol contract for a scenario that I have not been able to reproduce. That said, if there's a reproducible failing case, I'd be very happy to revisit the design.

> [Completion callback contract] Receiver-first rollout required: current OpenHands/automation main defines RunCompleteRequest with extra="forbid" and no cost field. As soon as this SDK registers cost, that deployed receiver rejects the entire callback with 422, leaving the run completion unrecorded. The coordinated OpenHands/automation#280 adds the field, but it is still open; merge and deploy the receiver before releasing this SDK behavior, and document/enforce that sequencing.

The observation about `extra="forbid"` is correct. However, I believe the sequencing concern is already addressed by the way the SDK is distributed. The SDK that executes within a sandbox is determined by the deployed automation service, not by whatever version happens to be available on PyPI.

Specifically, `presets/*/setup.sh` resolves the SDK version at runtime and installs that exact version:

```sh
SDK_VERSION=$(curl -sf "${AUTOMATION_API_URL}/sdk-version" | ...)
uv pip install "openhands-sdk==${SDK_VERSION}" ...
```

The `GET /sdk-version` endpoint (`app.py:279-290`) returns `get_sdk_version()`, which in turn resolves to `importlib.metadata.version("openhands-sdk")` (`utils/version.py:17-19`). In other words, the sandbox always installs the version pinned by the running automation service's own `pyproject.toml`, which is currently `openhands-sdk==1.39.1`.

Because of this, an SDK release containing this change can only be installed into a sandbox after the automation service itself has been deployed with an updated SDK pin. That same deployment would also include support for the `cost` field in `RunCompleteRequest` as well as migration `013`. As a result, there is no deployment order in which a cost-producing SDK would communicate with a receiver that rejects the field. Publishing a newer SDK to PyPI alone has no effect on automation runs.

Compatibility in the opposite direction is also preserved. Since `cost: float | None = None` is an optional field, the currently pinned SDK version (`1.39.1`), which never sends `cost`, continues to validate without any changes.

One additional update since the previous revision: following Malhotra5's review, I removed the `register_cost()` call from the preset scripts. The SDK is now the sole producer of this field. That change does not affect the sequencing argument above, and `cost` will continue to remain `NULL` until the automation service updates its SDK pin to the release that includes #4311.

Thank you very much! 🙏

hieptl:
> I think we want to have cost attribution for automations that run locally as well. It seems the current PR only implements this for remote conversations and remote workspaces

Hello @malhotra5,

I have updated the code based on your feedback.

Thank you very much! 🙏

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py`, Lines 1693-1703] **The stale-cost race remains**: the response to the prior review assumes the stats executor has queued on `FIFOLock` before the run thread reaches its error handler. `_emit_event_from_thread()` merely submits `locked_on_event` with `run_in_executor()` and discards the future; the worker is not guaranteed to start and enqueue before the run thread releases the step lock and immediately re-acquires it in the `except` path. The final state/status publication can therefore reach `RemoteConversation` first, while its cache still contains an earlier stats snapshot—including the subscribe-time `0.0` entry that satisfies the current `"stats" in cached` guard. `close()` then records that old value as this run's final cost. Make final cost ownership authoritative or track/drain the stats publication before terminal status is observable, and add a deterministic regression test where the executor starts late.
- [Completion callback contract] **The receiver dependency is still unresolved**: `OpenHands/automation#280`, which adds `cost` to the `extra="forbid"` receiver, remains open. The pinned-SDK deployment model explains how a coordinated rollout can be safe, but it does not make this branch independently backward-compatible. Do not release/deploy a cost-producing SDK until the receiver change and migration are merged and deployed; record that ordering as an explicit merge/release dependency.

**[TESTING GAPS]**
- [`tests/sdk/conversation/remote/test_remote_conversation.py`] The tests still inject stats before terminal status. They do not exercise the actual scheduling gap where `run_in_executor()` has accepted the task but its worker has not attempted the state lock before the error/final-state path runs.

**[RISK ASSESSMENT]**
This feeds billing/accounting data into the terminal callback that marks automation runs complete. The unresolved ordering can silently under-report spend, while an uncoordinated receiver rollout can reject completion callbacks entirely.

**Recommendation**: Do not auto-merge. Request human architecture review of authoritative cost ownership and enforce the receiver-first release sequence.

**VERDICT:**
❌ **Needs rework**: Local-conversation support is useful, but it does not resolve the remote ordering or cross-repository contract risks.

**KEY INSIGHT:**
Submitting work to an executor is not the same as placing that work in the FIFO lock queue.

---

>
>
>

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -2595,6 +2595,17 @@ def close(self) -> None:
...
         if first_attempt:
             self._cleanup_initiated = True
+
+            # Best-effort: hand the accumulated LLM cost to the workspace so it
+            # can be included in the automation completion callback. State is
+            # in-process here, so unlike RemoteConversation there is no cache to
+            # consult and no fetch that could block against a dead server.
+            try:
+                cost = self._state.stats.get_combined_metrics().accumulated_cost
+                self.workspace.register_cost(cost)
```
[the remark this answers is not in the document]
hieptl: Confirmed — the asymmetry is intentional and reflects what each side can reliably know.

**Local:** `self._state.stats` is the authoritative in-process source of truth. Every LLM used during the conversation records its metrics there, so by the time `close()` is called, the accumulated cost is known with certainty. If a run fails before making any LLM calls, the actual cost is genuinely `$0.00`. In that case, reporting `"cost": 0.0` provides the automation service with accurate data rather than a placeholder.

**Remote:** The client only maintains a cached view of the server-side state, and `close()` intentionally does not perform a live fetch. This is particularly important because `close()` also executes on failure paths, where the agent server may no longer be available. If the cached state does not include a `stats` entry, the cost is simply unknown. Reporting `0.0` in that situation would incorrectly record an unknown cost as a free run. For that reason, the guard exists only where an "unknown" state is actually possible. In the local implementation, that state cannot occur.

That said, the testing gap you identified was valid. I've added `test_close_reports_zero_cost_when_no_llm_calls` in `tests/sdk/conversation/local/test_conversation_core.py`. The test closes a conversation with no LLM activity—the exact state after an error that occurs before the first LLM call—and verifies that `workspace.accumulated_cost == 0.0`. This confirms that a genuine zero cost is recorded, rather than `None` or an omitted value. The test's docstring also documents the intentional distinction from `RemoteConversation.close()`.

Since `_send_completion_callback` includes `cost` whenever it is not `None`, this test also ensures that the error path continues to transmit `"cost": 0.0`.

enyst (commented):

🔴 **Needs improvement** - The callback plumbing is coherent and the focused suites pass, but the accounting unit is wrong for resumed conversations and one new state read violates the repository's locking invariant.

**[CRITICAL ISSUES]**
- [`openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py`, Line 2604; `remote_conversation.py`, Line 1702] **Incorrect Billing Data**: Both close paths register `get_combined_metrics().accumulated_cost`, which is the conversation's lifetime total, not this automation run's cost. `ConversationStats` explicitly persists full cost history, and `RemoteConversation(..., conversation_id=...)` can attach to an existing conversation, so a resumed run will resend all prior spend under the new `AUTOMATION_RUN_ID`. Capture a baseline when the run/conversation attachment begins and register only the delta at close (the existing `Metrics.diff()` machinery is the natural model); add local-resume and remote-attach regression tests proving old spend is not charged again.
- [`openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py`, Line 2604] **Concurrency**: This newly reads `self._state.stats` without `with self._state:`, contrary to the repository's explicit `LocalConversation` invariant for every mutable state read. `close()` can race an active run; worse, the broad exception handler silently turns that race into a missing cost. Take the metrics snapshot under the state lock, release it, then register the value on the workspace, and cover concurrent close/run behavior.

**[TESTING GAPS]**
- [New cost tests] **Wrong Lifecycle Fixture**: The new tests start from zero-cost state, so they cannot catch duplicate billing from persisted/attached conversations, and none exercises the lock race. Keep the current callback assertions, but add a previous-cost baseline plus new-run cost case for both local and remote paths and a concurrency regression for the local state snapshot.

I ran the focused workspace/conversation suites at this head: **160 passed**.

**VERDICT:**
❌ **Needs rework**: Reporting cumulative history as per-run cost can overstate automation billing, and the local path is not concurrency-safe.

**KEY INSIGHT:**
The callback is keyed to an automation run, so its cost must be a run delta—not a persisted conversation total.

**[RISK ASSESSMENT]**
This changes shared public workspace classes and cost-accounting behavior, with a concrete overreporting path for resumed conversations and a race in core conversation cleanup. **Recommendation**: Do not auto-merge. Request review from a human SDK/automation maintainer to validate the cost baseline boundary and cleanup concurrency semantics.

---

>
>
>

tofarr (approved):
🍰

malhotra5 (changes requested):
I was noodling over this and was wondering if it is necessary to emit accumulate cost on automation completion this way. 


For automations that create an agent conversation, we already emit event + cost metadata webhooks to the main app server. Maybe we can just query the DB directly instead of adding another webhook that only fires on automation completion?

hieptl:
> I was noodling over this and was wondering if it is necessary to emit accumulate cost on automation completion this way.
> 
> For automations that create an agent conversation, we already emit event + cost metadata webhooks to the main app server. Maybe we can just query the DB directly instead of adding another webhook that only fires on automation completion?

Hello @malhotra5, thank you for thinking this through! I did consider the database approach, but I believe using the callback is a better fit for a few reasons:

* **This isn't introducing a new webhook.** The completion callback already exists and already includes `status`, `run_id`, `conversation_id`, and `error`. This PR simply adds a single `cost` field to that existing payload, reading the value from the state the SDK already holds at `close()`.

* **It works for both cloud and local deployments.** Local/self-hosted runs don't have an app server to query. Automations on `LocalAgentServerBackend` run directly against a pre-configured agent server, so no event or metadata webhooks are sent to an app server, meaning there's no database row containing the cost. Since we want cost attribution for local runs as well (per your earlier review), the callback is the only solution that works consistently across both cloud and local environments. A database-based approach would require a separate mechanism for local runs.

* **It avoids cross-service coupling.** The automation service is designed as a self-contained microservice with its own database. Conversation costs, however, are stored in the app server's database (`app_conversation_info`), which is populated by its webhook router. Reading that database directly would couple the automation service to the app server's schema across repositories. Going through the app server's API instead would introduce additional complexity, including per-user authentication and a new query/retry flow. By comparison, extending the existing callback with one additional field keeps the integration much simpler.

* **The timing is more reliable.** Stats webhooks are buffered (with a default 30-second flush interval), and the app server already performs a run-end live sync (`_sync_live_conversation_stats`) because streamed stats can under-report switched-in LLMs and ACP agents. As a result, the database value may still be stale or incomplete when the automation finishes. The callback, on the other hand, reads the final SDK-side statistics directly at `close()`, when they're already finalized.

That said, I do think querying the app server could still be valuable as a backfill mechanism for cloud runs where the completion callback never fires (for example, due to a sandbox crash or watchdog termination), which are currently recorded with `cost = NULL` (unknown). I'd be happy to file a follow-up issue to explore that approach if you think it's worthwhile.

Thank you very much! 🙏

malhotra5:
@hieptl architecturally I think the automation service should be an orchestration layer for spinning up automations, and kept as thin as possible. That requires some amount of minimal metadata in order to do so. With cost attribution, imho it seems like we're fragmenting the cost attribution across multiple services. 

In the local case, wouldn't the agent server already have information regarding the cost of a conversation? And in cloud it would be the app server. If we don't have an endpoint to extract conversation metadata from the agent server maybe we can add that instead. The automation microservice by nature is not self contained, it requires orchestrating against the agent server or main app server. This we abstract with the notion of a "BE". Now BE can be the source of truth on cost, the automation service can dispatch new conversation requests and maintain a automation id <> convo id link as the minimal amount of metadata. WDYT? 


(Also with the alternate approach above we can see real time cost updates and it will most likely be less prone to a single point of failure in the event the completion callback doesn't successfully fire)

hieptl:
> @hieptl architecturally I think the automation service should be an orchestration layer for spinning up automations, and kept as thin as possible. That requires some amount of minimal metadata in order to do so. With cost attribution, imho it seems like we're fragmenting the cost attribution across multiple services.
> 
> In the local case, wouldn't the agent server already have information regarding the cost of a conversation? And in cloud it would be the app server. If we don't have an endpoint to extract conversation metadata from the agent server maybe we can add that instead. The automation microservice by nature is not self contained, it requires orchestrating against the agent server or main app server. This we abstract with the notion of a "BE". Now BE can be the source of truth on cost, the automation service can dispatch new conversation requests and maintain a automation id <> convo id link as the minimal amount of metadata. WDYT?
> 
> (Also with the alternate approach above we can see real time cost updates and it will most likely be less prone to a single point of failure in the event the completion callback doesn't successfully fire)

Thanks, @malhotra5. I agree with the thin-orchestration principle, but I think the pull model runs into a few practical challenges.

The first is that the `conversation_id` itself currently travels over the same callback. Today, the automation service only learns the `conversation_id` from the completion callback—`complete_run` is the only sink. While the SDK registers the ID inside the sandbox, it doesn't transmit it until the workspace exits. Because of that, a "keep the link, then query the BE for cost" approach has the same single point of failure: if the callback never arrives, we lose both the link and the cost. Making the link independent of the callback would require additional infrastructure, such as an early "conversation created" event (effectively another callback) or a tag-based reverse lookup through the app server, which is only available in cloud mode. That feels like a broader architectural change than adding a single `cost` field to an existing payload.

The second challenge is that, while the agent-server endpoint already exists, it isn't available when we'd actually need it. `GET /api/conversations/{id}` already returns `ConversationInfo`, including usage statistics. However, in cloud mode the sandbox (and its agent server) is deleted once the run completes. At that point, the app server becomes the only remaining source of truth, but accessing it requires per-user authentication from the automation service. It also relies on buffered webhooks (with a default 30-second flush interval), and the app server itself performs a live pull at run completion because the streamed statistics can under-report switched-in LLMs and ACP agents. As a result, pulling immediately after completion can still return stale data, whereas the completion callback carries the final value from the process that actually computed it.

On the concern about fragmentation, I'd think of the run's `cost` the same way we think about `status`, `error_detail`, and `conversation_id`: it's simply part of the run outcome snapshot that's stored on the run record. The backend remains the source of truth—the automation service isn't computing anything; it's simply persisting what it's receives. That's similar to how the app server materializes `accumulated_cost_usd` per conversation from the same stream. From that perspective, I think the orchestration layer remains thin.

That said, I do like the two advantages the pull model provides: real-time cost updates and recovery from missed callbacks. My only concern is that both depend on reporting the `conversation_id` at creation time rather than completion.

Would it make sense to treat that as a follow-up effort? For example:

* (a) register the `conversation_id` as soon as it's created,
* (b) expose live cost for `RUNNING` runs through the backend, and
* (c) backfill cloud runs whose completion callback never arrived using the existing `automationrunid` tags.

That would keep this PR focused on the minimal v1 required for the current ticket while leaving the door open for the architecture you're describing.

What do you think?

Thank you very much! 🙏

malhotra5:
I agree that registering conversation id as soon as possible would be amazing. And automations is a per user authenticated microservice, so ideally we can move towards an architecture that does leverage that aspect and keep the orchestration as thin as possible. And i believe in cloud having the app server be the source of truth is correct mental model to have. 

As for now, I'm going to approve to unblock progress on this and we can revisit when bandwidth allows in the future.

malhotra5 (approved):
Thank you for the discussions, lgtm!

PR #4319, opened 2026-07-31
fix: PATCH /api/settings loads the profile's LLM when setting active_profile

### LINKED ISSUE #4314: [Bug]: PATCH /api/settings can silently desync active_profile from agent_settings.llm
## Bug Description

`PATCH /api/settings` accepts `active_profile` as a top-level field (documented in its own docstring), but setting it this way only updates the tracking label — it does **not** load or apply that profile's LLM config to `agent_settings.llm`. This is inconsistent with the dedicated `POST /api/profiles/{name}/activate` endpoint, which updates both atomically. The result is that `active_profile` and `agent_settings.llm` can silently disagree: `GET /api/settings` (and every conversation-creation flow that trusts it) keeps returning whatever `agent_settings.llm` was last set to, even after `active_profile` has moved on to a different name.

Any caller that creates a conversation by fetching `/api/settings` and forwarding `agent_settings` (a common, documented pattern for backend/automation clients) can silently get a stale or wrong LLM config — including a leftover config with an invalid/expired API key — with no error surfaced until the LLM call itself fails.

## Expected Behavior

`active_profile` should never be settable independently of the LLM config it names. Either:
- `PATCH /api/settings` should reject/ignore a bare `active_profile` change that isn't accompanied by a matching `agent_settings_diff.llm`, or
- Setting `active_profile` via `PATCH /api/settings` should re-load and apply that profile's LLM the same way `POST /api/profiles/{name}/activate` does, or
- `active_profile` should be dropped from the generic settings PATCH surface entirely, with `/api/profiles/{name}/activate` as the only way to change it.

## Actual Behavior

`PersistedSettings.update()` handles `active_profile` as a bare label assignment with no side effects:

```python
# openhands-agent-server/openhands/agent_server/persistence/models.py
if "active_profile" in payload:
    self.active_profile = payload["active_profile"]
```

Compare with `activate_profile()`, which is the only path that keeps both fields consistent:

```python
# openhands-agent-server/openhands/agent_server/profiles_router.py
def apply_profile(settings: PersistedSettings) -> PersistedSettings:
    settings.agent_settings = settings.agent_settings.model_copy(update={"llm": llm})
    settings.active_profile = name
    return settings
```

`update_settings()`'s own docstring documents `active_profile` as one of the fields accepted by the generic `PATCH /api/settings` endpoint, alongside `agent_settings_diff`, so this isn't an obscure internal codepath — it's part of the advertised API surface:

```python
# openhands-agent-server/openhands/agent_server/settings_router.py
"""Update settings with partial changes.

Accepts ``agent_settings_diff``, ``conversation_settings_diff``,
``misc_settings_diff``, and/or ``active_profile`` for incremental updates.
"""
```

And `get_settings()` returns `agent_settings` as literally persisted, with no cross-check against `active_profile`:

```python
return SettingsResponse(
    agent_settings=settings.agent_settings.model_dump(mode="json", context=context),
    ...
    active_profile=settings.active_profile,
    ...
)
```

## Steps to Reproduce

1. Save two LLM profiles, e.g. `POST /api/profiles` for `profile-a` (model `X`) and `profile-b` (model `Y`).
2. `POST /api/profiles/profile-a/activate` — `agent_settings.llm.model` is now `X`, `active_profile` is `profile-a`.
3. `PATCH /api/settings` with body `{"active_profile": "profile-b"}` (no `agent_settings_diff`).
4. `GET /api/settings` — `active_profile` now reports `profile-b`, but `agent_settings.llm.model` is still `X`.
5. Any conversation created by forwarding this `agent_settings` (e.g. `POST /api/conversations` with `agent_settings` copied from step 4's response) launches with model `X`/`profile-a`'s LLM config, silently contradicting the reported active profile.

## Actual impact observed

Found this while investigating why an internal automation-created conversation launched with a stale, previously-configured LLM (with an invalid API key from an earlier, unrelated experiment) instead of the properly configured default, even though `active_profile` correctly named the intended profile. The conversation's `execution_status` ended up `"error"` (LLM auth failure) with no earlier signal that anything was misconfigured — the mismatch was only visible by inspecting persisted conversation state directly.

## Additional Context

- Confirmed against `openhands-agent-server` v1.37.0.
- Related prior work in this area: #3172, #3173, #3552, #3719, #3933, #3713 (none of these address this specific `PATCH /api/settings` vs `/activate` inconsistency).

emmanuel-adu:
HUMAN:

Reviewed the diff and repro output below; the fix matches what the issue asked for (option 2). Ran the curl repro and test suite myself before opening this.

## Video/Screenshots

Ran  against a live server and the full test suite - screenshot below.

oh-active-profile-evidence

---

AGENT:

## Why

`PATCH /api/settings` accepted `active_profile` as a bare label with no side effects, so it could silently disagree with `agent_settings.llm` - unlike `POST /api/profiles/{name}/activate`, which updates both atomically. A client that creates a conversation by forwarding `agent_settings` from a `GET /api/settings` response could get a stale or wrong LLM config with no error until the LLM call itself failed. Full analysis in #4314.

## Summary

- `PATCH /api/settings` now loads and applies the named profile's LLM when `active_profile` is set, unless the caller already supplied an explicit `agent_settings_diff.llm` (that's respected as-is)
- Returns 404 for an unknown profile name instead of accepting a dangling pointer
- Updated two existing tests that set `active_profile` to a profile that was never actually saved (the exact bug condition) and added 4 new tests covering the fix

## Issue Number

Fixes #4314

## How to Test

Automated: `uv run pytest tests/agent_server/test_settings_router.py tests/agent_server/test_profiles_router.py -v`

Manual, real HTTP end-to-end (not mocked) - matches the repro steps in #4314:
1. `uv run python -m openhands.agent_server --port 8765 --host 127.0.0.1`
2. `curl -X POST http://127.0.0.1:8765/api/profiles/fast-profile -d '{"llm": {"model": "claude-haiku-e2e"}}'`
3. `curl -X PATCH http://127.0.0.1:8765/api/settings -d '{"active_profile": "fast-profile"}'`
4. `curl http://127.0.0.1:8765/api/settings` → `agent_settings.llm.model` should be `claude-haiku-e2e`, not whatever it was before
5. `curl -X PATCH http://127.0.0.1:8765/api/settings -d '{"active_profile": "does-not-exist"}'` → 404

## Type

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No `agent_kind` interaction concerns: the loaded LLM is folded into `agent_settings_diff.llm`, which goes through `PersistedSettings.update()`'s existing atomic validate-then-apply path (same as any other agent-settings diff), so a failing `conversation_settings_diff`/`misc_settings_diff` in the same request still rolls back cleanly - nothing is applied.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/settings_router.py
+++ b/openhands-agent-server/openhands/agent_server/settings_router.py
@@ -7,6 +7,7 @@ from pydantic import ValidationError
 
 from openhands.agent_server._secrets_exposure import (
     build_expose_context,
+    get_cipher,
     get_config,
     parse_expose_secrets_header,
     translate_missing_cipher,
@@ -14,10 +15,12 @@ from openhands.agent_server._secrets_exposure import (
 from openhands.agent_server.persistence import (
     SECRET_NAME_PATTERN,
     PersistedSettings,
+    get_llm_profile_store,
     get_secrets_store,
     get_settings_store,
 )
 from openhands.agent_server.persistence.models import SettingsUpdatePayload
+from openhands.agent_server.profiles_router import _store_errors
 from openhands.agent_server.telemetry import notify_misc_settings_changed
 from openhands.sdk.logger import get_logger
 from openhands.sdk.mcp.config import MCPServer
@@ -179,6 +182,15 @@ async def update_settings(
 
     Accepts ``agent_settings_diff``, ``conversation_settings_diff``,
     ``misc_settings_diff``, and/or ``active_profile`` for incremental updates.
+
+    Setting ``active_profile`` to a profile name (without an explicit
+    ``agent_settings_diff.llm``) loads that profile and applies its LLM
+    config, the same as ``POST /api/profiles/{name}/activate`` - so
+    ``active_profile`` can't silently disagree with ``agent_settings.llm``.
+    Returns 404 if the named profile doesn't exist. Pass an explicit
+    ``agent_settings_diff.llm`` alongside ``active_profile`` to opt out and
+    set both independently.
+
     The three ``*_settings_diff`` fields are deep-merged; nested objects merge
     recursively, and a ``null`` value **inside a nested map deletes that entry**
     — the "unset" primitive that lets a client remove a single map key without
@@ -226,11 +238,55 @@ async def update_settings(
     )
 
 
+def _resolve_active_profile_llm(
+    request: Request, update_data: SettingsUpdatePayload
+) -> SettingsUpdatePayload:
+    """Fold the named profile's LLM into ``agent_settings_diff`` when the
+    caller sets ``active_profile`` without an explicit ``llm`` diff.
+
+    ``PATCH /api/settings`` used to accept ``active_profile`` as a bare label
+    with no side effects, letting it silently disagree with
+    ``agent_settings.llm`` - unlike ``POST /api/profiles/{name}/activate``,
+    which updates both atomically. This mirrors that endpoint's behavior
+    (load the profile, apply its LLM) so the two code paths can't diverge.
+    See #4314.
+    """
+    profile_name = update_data.get("active_profile")
+    agent_diff = update_data.get("agent_settings_diff")
+    explicit_llm_diff = isinstance(agent_diff, dict) and "llm" in agent_diff
+    if not profile_name or explicit_llm_diff:
+        return update_data
+
+    cipher = get_cipher(request)
+    profile_store = get_llm_profile_store()
+    try:
+        with _store_errors():
+            llm = profile_store.load(profile_name, cipher=cipher)
+    except FileNotFoundError:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Profile '{profile_name}' not found",
+        )
+
+    return cast(
+        SettingsUpdatePayload,
+        {
+            **update_data,
+            "agent_settings_diff": {
+                **(agent_diff if isinstance(agent_diff, dict) else {}),
+                "llm": llm.model_dump(mode="json", context={"expose_secrets": True}),
+            },
+        },
+    )
+
+
 def _apply_settings_update(
     request: Request,
     update_data: SettingsUpdatePayload,
     before_update: Callable[[PersistedSettings], None] | None = None,
 ) -> SettingsResponse:
+    update_data = _resolve_active_profile_llm(request, update_data)
+
     # Apply updates atomically with file locking
     def apply_update(settings: PersistedSettings) -> PersistedSettings:
         if before_update is not None:
--- a/openhands-agent-server/openhands/agent_server/settings_router.py
+++ b/openhands-agent-server/openhands/agent_server/settings_router.py
@@ -182,14 +182,9 @@ async def update_settings(
 
     Accepts ``agent_settings_diff``, ``conversation_settings_diff``,
     ``misc_settings_diff``, and/or ``active_profile`` for incremental updates.
-
-    Setting ``active_profile`` to a profile name (without an explicit
-    ``agent_settings_diff.llm``) loads that profile and applies its LLM
-    config, the same as ``POST /api/profiles/{name}/activate`` - so
-    ``active_profile`` can't silently disagree with ``agent_settings.llm``.
-    Returns 404 if the named profile doesn't exist. Pass an explicit
-    ``agent_settings_diff.llm`` alongside ``active_profile`` to opt out and
-    set both independently.
+    Setting ``active_profile`` loads and applies that profile's LLM, same as
+    ``POST /api/profiles/{name}/activate``, unless ``agent_settings_diff.llm``
+    is also given.
 
     The three ``*_settings_diff`` fields are deep-merged; nested objects merge
     recursively, and a ``null`` value **inside a nested map deletes that entry**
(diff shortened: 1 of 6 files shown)


enyst (commented):

🟢 **Good taste** — the fix folds profile resolution into the existing atomic settings-update path instead of creating another persistence mechanism.

I traced the flow through profile loading, secret serialization, `PersistedSettings.update()`, and `FileSettingsStore.update()`. Unknown profiles fail before persistence, explicit LLM diffs retain their documented precedence, encrypted profile secrets round-trip through the existing serializers, and the targeted settings/profile suite passes locally: 151 tests.

**[RISK ASSESSMENT]**
The change deliberately alters a public settings endpoint and touches persisted, secret-bearing LLM configuration. It follows established activation, validation, locking, and encryption patterns, and the regression coverage substantially contains that risk.

**VERDICT:**
✅ **Worth merging**: The stale-profile bug is fixed without bypassing existing validation or atomic persistence.

**KEY INSIGHT:**
Profile identity and the effective LLM now move together through the same validated settings transaction.

VascoSch92:
Can you please trimmer the comments and docstring? Just the minimal one. 

Otherwise, LGTM

Then you can ping me

emmanuel-adu:
Trimmed the comments/docstrings to the minimal form. Ready for another look, @VascoSch92.

enyst (commented):

🟢 **Good taste** — the current head preserves the focused fix and only trims explanatory text added around it.

I rechecked the delta since the prior review and the complete update path. Profile loading still happens before the locked settings transaction, unknown or invalid profile names fail before persistence, explicit `agent_settings_diff.llm` retains precedence, and the profile LLM still passes through the existing validation and secret-serialization machinery. The latest push introduces no semantic change and no new concern.

**[RISK ASSESSMENT]**
This changes a public settings endpoint and touches persisted, secret-bearing LLM configuration. The implementation reuses established profile loading, validation, locking, and encryption paths, with concrete HTTP evidence and focused regression coverage.

**VERDICT:**
✅ **Worth merging**: The profile label and effective LLM now remain atomic without bypassing existing settings safeguards.

**KEY INSIGHT:**
The latest revision removes comment noise without weakening the single validated transaction that fixes the stale-profile bug.

openhands-agent-server/openhands/agent_server/settings_router.py
```diff
@@ -7,17 +7,20 @@
...
 )
 from openhands.agent_server.persistence import (
     SECRET_NAME_PATTERN,
     PersistedSettings,
+    get_llm_profile_store,
     get_secrets_store,
     get_settings_store,
 )
 from openhands.agent_server.persistence.models import SettingsUpdatePayload
+from openhands.agent_server.profiles_router import _store_errors
```
VascoSch92: You are importing a private method. This seems like a smell :-) 

Do you need that? Can you do differently? What do you think=

openhands-agent-server/openhands/agent_server/settings_router.py
```diff
@@ -226,11 +233,47 @@ async def update_settings(
     )
 
 
+def _resolve_active_profile_llm(
+    request: Request, update_data: SettingsUpdatePayload
+) -> SettingsUpdatePayload:
+    """Fold the named profile's LLM into ``agent_settings_diff`` unless the
+    caller already gave one explicitly. Mirrors ``/activate``. See #4314."""
```
VascoSch92:
```suggestion
    caller already gave one explicitly. Mirrors ``/activate``. """
```

tests/agent_server/test_settings_router.py
```diff
@@ -687,6 +692,74 @@ def test_patch_settings_active_agent_profile_id_independent(client_with_settings
     assert refetch["active_profile"] == "fast-profile"
 
 
+def test_patch_settings_active_profile_applies_llm(client_with_settings):
```
VascoSch92, line 695: add a test that exercises the encrypted-secret round-trip

```python
  def test_patch_settings_active_profile_applies_encrypted_api_key(
      client_with_settings, secret_key
  ):
      """Applying a profile carries its at-rest-encrypted api_key through PATCH."""
      cipher = Cipher(secret_key)
      get_llm_profile_store().save(
          "secure-profile",
          LLM(model="claude-haiku", api_key=SecretStr("sk-secret")),
          include_secrets=True,
          cipher=cipher,
      )
  
      response = client_with_settings.patch(
          "/api/settings",
          json={"active_profile": "secure-profile"},
      )
      assert response.status_code == 200

      exposed = client_with_settings.get(
          "/api/settings", headers={"X-Expose-Secrets": "plaintext"}
      ).json()
      assert exposed["agent_settings"]["llm"]["model"] == "claude-haiku"
      assert exposed["agent_settings"]["llm"]["api_key"] == "sk-secret"
      assert exposed["llm_api_key_is_set"] is True
```

openhands-agent-server/openhands/agent_server/settings_router.py
VascoSch92: Ok I looked inside this.

`_store_errors` is already copy-pasted in two routers (`profiles_router.py`, `agent_profiles_router.py`); this makes a third module depend on one private copy.

The honest fix is to hoist it once is to add to a shared module `_secrets_exposure.py` is the natural home since all three routers already import from it

```python
  # _secrets_exposure.py
  @contextmanager
  def store_errors() -> Iterator[None]:
      """Map ``LLMProfileStore`` errors to HTTP responses."""
      try:
          yield
      except TimeoutError:
          raise HTTPException(status.HTTP_503_SERVICE_UNAVAILABLE, "Profile store is busy. Please retry.")
      except ValueError as e:
          raise HTTPException(status.HTTP_400_BAD_REQUEST, str(e))
```

Then in each router replace the local def `_store_errors` / import with from `openhands.agent_server._secrets_exposure import store_errors` and use with 
  `store_errors():`. 

You can also open another PR for that on top of this one you you want :-)

VascoSch92 (commented):
I did another pass. Left some comments

openhands-agent-server/openhands/agent_server/settings_router.py
emmanuel-adu: hey, good point thanks for the suggestion!

emmanuel-adu: hoisted it into `_secrets_exposure.py` as `store_errors()` and switched `settings_router.py`, `profiles_router.py`, and `agent_profiles_router.py` over to it, dropping all three local/private copies.

emmanuel-adu: Kept it in this PR since it's a relatively small change, and scoped to this comment.

openhands-agent-server/openhands/agent_server/settings_router.py
emmanuel-adu: Applied - removed the stray issue reference from the docstring.

tests/agent_server/test_settings_router.py
emmanuel-adu: Added, matches the round-trip you outlined - test_patch_settings_active_profile_applies_encrypted_api_key.

VascoSch92 (approved):

VascoSch92:
Thanks

Issue #4304, opened 2026-07-29
CI stability audit: unpinned/floating dependencies that can flip CI from green to red


simonrosenberg:
## Summary

Audit (verified line-by-line against `9627f82c60e9ffb7d51dc7e6277cf4d0f61f01b4`) of everything in this repo whose *resolved* behavior can change while the git SHA under test stays exactly the same — unpinned package versions, floating MCP/CLI invocations, mutable Action refs, unpinned Docker/OS inputs. Any of these can turn a previously-green commit red (or silently change agent behavior) with zero code changes on our side.

Verification pass corrected one thing from the original draft: the "moving major tag" claim for `astral-sh/setup-uv` / `tj-actions/changed-files` only applies to 3 specific stray call sites — the other ~24 occurrences of those actions are already correctly SHA-pinned (`@<40-char-sha> # vN`). That's called out below so we don't waste effort re-pinning things that are already fine.

## Immediate risks (active in default/CI paths today)

- [ ] **[`openhands-tools/.../subagents/web_researcher.md:12`](https://github.com/OpenHands/software-agent-sdk/blob/9627f82c6/openhands-tools/openhands/tools/preset/subagents/web_researcher.md#L12)** — `uvx mcp-server-fetch`, unpinned, in an active subagent preset.
- [ ] **[`openhands-tools/.../browser_use/impl.py:238`](https://github.com/OpenHands/software-agent-sdk/blob/9627f82c6/openhands-tools/openhands/tools/browser_use/impl.py#L238)** — runtime auto-install runs `uvx playwright install chromium --with-deps --no-shell`, unpinned.
- [ ] **[`.github/workflows/tests.yml:230`](https://github.com/OpenHands/software-agent-sdk/blob/9627f82c6/.github/workflows/tests.yml#L230)** — CI separately runs unpinned `uvx playwright install chromium`.
- [ ] **[`openhands-agent-server/.../docker/Dockerfile:182-185`](https://github.com/OpenHands/software-agent-sdk/blob/9627f82c6/openhands-agent-server/openhands/agent_server/docker/Dockerfile#L182-L185)** — `npm install -g @agentclientprotocol/claude-agent-acp@0.44.0 @agentclientprotocol/codex-acp@1.1.2 @google/gemini-cli@0.46.0`: top-level packages are exact-pinned, but there's no lockfile for this ad-hoc install, so every transitive npm dependency resolves fresh at build time.
[...]

PR #4342, opened 2026-08-03
fix(browser): a browser tool that cannot start should not fail the conversation

### LINKED ISSUE #16270: [Bug]: AttributeError: 'Server' object has no attribute 'list_tools' when browser tool is used (breaks chat with 500 error)
### Operating System

Windows

### Installation Method

Other

### Agent Canvas Version

1.8.0

### Bug Description

When the agent uses the browser tool during a chat, the request fails with a 500 Internal Server Error due to an AttributeError in the bundled browser_use package: 'Server' object has no attribute 'list_tools'. This appears to be caused by a version mismatch between the browser_use package and the mcp SDK package bundled with this release.

### Steps to Reproduce

1. Install Agent Canvas v1.8.0 on Windows via the standalone .exe installer.
2. Configure a custom OpenAI-compatible LLM (local llama.cpp server) under Settings > LLM > Advanced options.
3. Start a new conversation.
4. Send any message that causes the agent to invoke its default toolset (including the browser tool).
5. Observe the request fails with a 500 Internal Server Error.

### Actual Behavior

The chat request fails with:
{"detail":"Internal Server Error","exception":"'Server' object has no attribute 'list_tools'","error_id":"..."}

At app startup, the same underlying error is also logged as a non-fatal warning during tool preloading:
"Tool preload service failed to start - skipping"

### Expected Behavior

The chat should complete normally. Either the browser tool should initialize without error, or a failure in the browser tool should not cause the entire chat request to fail with a 500 error.

### Relevant Logs

```shell
[agent-server] {"levelname": "ERROR", "name": "openhands.agent_server.tool_preload_service", "message": "Error preloading chromium", "exc_info": "Traceback (most recent call last):
  File \"...\\openhands\\agent_server\\tool_preload_service.py\", line 33, in start
    BrowserToolExecutor()
  File \"...\\openhands\\tools\\browser_use\\impl.py\", line 393, in __init__
    run_with_timeout(init_logic, init_timeout_seconds)
  File \"...\\openhands\\tools\\browser_use\\impl.py\", line 359, in init_logic
    self._server = CustomBrowserUseServer(
        session_timeout_minutes=session_timeout_minutes,
    )
  File \"...\\openhands\\tools\\browser_use\\server.py\", line 23, in __init__
    super().__init__(session_timeout_minutes=session_timeout_minutes)
  File \"...\\browser_use\\mcp\\server.py\", line 210, in __init__
    self._setup_handlers()
  File \"...\\browser_use\\mcp\\server.py\", line 215, in _setup_handlers
    @self.server.list_tools()
     ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Server' object has no attribute 'list_tools'"}

Later, during actual chat usage:
HTTP request failed (500 Internal Server Error): {"detail":"Internal Server Error","exception":"'Server' object has no attribute 'list_tools'","error_id":"f92dc184277b40518e570c29727378f1"}
```

### Screenshots

_No response_

### Additional Context

- OpenHands SDK version shown in banner at startup: v1.39.1
- OS: Windows 11
- LLM backend: local llama.cpp server exposed via OpenAI-compatible API (http://127.0.0.1:8088/v1)
- No custom tool configuration was changed from defaults.
- Suspected root cause: version mismatch between the browser_use package and the mcp Python SDK bundled in this release — browser_use's mcp/server.py calls self.server.list_tools(), a method not present in the currently pinned mcp SDK version.

onatozmenn:
HUMAN:

I fixed this by keeping mcp on 1.x for the browser tool and making browser startup failures non-fatal, and I verified the issue by reproducing the dependency resolution outside the repo where browser-use 0.11.9 resolves to mcp 2.0.0.

---

AGENT:

> **Correction, 2026-08-03.** This PR originally also added `mcp<2` to `openhands-tools`, on the reasoning that nothing kept mcp on 1.x. That reasoning was wrong and the constraint is gone as of `c5daef2`. Details in the comment below; the containment change is unchanged and is what this PR is now about. The HUMAN note above was written against the original framing.

## Why

Fixes OpenHands/OpenHands#16270 (`bug`, `browser`, `mcp`, `windows`, `priority:high`).

The reporter's Agent Canvas install crashes in `browser_use/mcp/server.py`, which registers handlers with `@self.server.list_tools()`. That decorator lives on the mcp SDK's low-level `Server`, and mcp 2.0.0 removed it. Verified against the wheels:

| mcp version | `Server.list_tools` |
|---|---|
| 1.26.0 (locked here) | present, `lowlevel/server.py:434` |
| 1.29.0 (latest 1.x) | present, `lowlevel/server.py:440` |
| 2.0.0 (2026-07-28) | **absent** |

So their environment had an mcp the bundled `browser_use` cannot drive. What this PR fixes is what that costs them.

`BrowserToolExecutor()` raises during construction, the exception leaves `BrowserToolSet.create()`, travels through `resolve_tool` into agent setup, and surfaces as a 500 on `POST /api/conversations` and on every chat request after it. The whole conversation dies because one tool cannot start.

The repo already disagrees with itself here. `tool_preload_service` catches the identical failure at startup and logs `Tool preload service failed to start - skipping` — the reporter sees exactly that line, non-fatally, minutes before the same error takes down their chat. `BrowserToolSet` also already has `is_usable()` for the case where Chromium is missing, so an agent running without browser tools is an expected state, not a new one.

The reporter asks for precisely this as the acceptable outcome: "either the browser tool should initialize without error, or a failure in the browser tool should not cause the entire chat request to fail with a 500 error."

## Summary

- `BrowserToolSet.create()` logs a warning and returns no tools when the shared executor cannot be constructed, instead of propagating out of tool resolution.
- Two regression tests, one at the `create()` boundary and one at `resolve_tool`, which is the step that used to become the 500.

## Issue Number

Fixes OpenHands/OpenHands#16270

## How to Test

```
uv run pytest tests/tools/browser_use/test_browser_toolset.py
```

- `test_browser_toolset_create_degrades_when_executor_fails` — `create()` returns `[]` and warns when the executor raises the reported `AttributeError`
- `test_resolve_tool_survives_browser_executor_failure` — `resolve_tool` returns no tools rather than raising

Both fail if the new `except` clause is narrowed so the error escapes; the other 15 tests in that file pass either way.

Confirm the API boundary that produced the report:

```
uv pip install mcp==1.29.0 --target /tmp/m129
grep -n 'def list_tools' /tmp/m129/mcp/server/lowlevel/server.py   # line 440
uv pip install mcp==2.0.0 --target /tmp/m200
grep -n 'def list_tools' /tmp/m200/mcp/server/lowlevel/server.py   # no match
```

`tests/tools/browser_use`, `tests/sdk/tool` and `tests/cross`: 657 passed, 49 skipped. Two exclusions, both pre-existing on this Windows host and verified failing/flaky identically on unmodified `main` — `test_generate_baseline_payloads_uses_uv_with_release_cutoff` (asserts a POSIX path against `\tmp\fake-python`) and `test_websocket_attach_wait_does_not_block_ready_endpoint` (0.534s against a 0.5s bound; passes on re-run).

`ruff format`, `ruff check`, `pycodestyle --max-line-length=88` and `scripts/check_import_rules.py` are clean on both touched files.

## Video/Screenshots

Console output above; this path has no UI surface.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- This does not explain how the reporter's environment ended up with an incompatible mcp, and I no longer claim it does. See the comment below for what I found and what I could not establish.
- `BrowserToolSet.is_usable()` still only probes for Chromium, so an incompatible mcp keeps the toolset "usable" while `create()` yields nothing. Making usability reflect the real construction cost would mean building the server on every probe. Left alone; say the word if you want it tightened.
- Catching bare `Exception` here is deliberate, and matches `_check_tool_usable` in the registry, which already treats any exception from a usability check as "not usable". Narrower would mean guessing which failure modes a browser stack can produce.

<sub>Investigated and implemented with AI assistance.</sub>


onatozmenn:
Correcting myself on the dependency half of this PR, and dropping it in `c5daef2`.

I claimed nothing in the graph keeps mcp on 1.x, and added `mcp<2` to `openhands-tools`. That was wrong. `openhands-sdk` requires `fastmcp>=3.0.0` (true for 1.35.0, 1.39.1 and 1.40.0 alike), fastmcp 3.x pulls `fastmcp-slim[client,server]`, and that declares `mcp<2.0,>=1.24.0`. The bound already exists, one level further out than I looked.

What I actually resolved was `browser-use==0.11.9` on its own, which does reach mcp 2.0.0 because browser-use only pins `mcp==1.26.0` from 0.12.0 onwards. Nobody installs it that way. Resolving the real package outside this repo gives:

```
$ echo openhands-tools > r.txt && uv pip compile r.txt --python-version 3.12
browser-use==0.11.13
mcp==1.29.0
openhands-tools==1.40.0
```

mcp 1.29.0, not 2.0.0. The constraint I added was redundant, so it is gone.

One thing from that dig may still be worth someone's time. I could only force the broken pair by letting uv pick `openhands-tools==1.40.0` together with `openhands-sdk==1.11.5`, at which point the fastmcp 3 bound no longer applies and `mcp==2.0.0` resolves cleanly:

```
$ printf 'openhands-tools\nmcp==2.0.0\n' > r.txt && uv pip compile r.txt --python-version 3.12
browser-use==0.11.13
fastmcp==2.14.1
mcp==2.0.0
openhands-sdk==1.11.5
openhands-tools==1.40.0
```

That combination is only reachable because `openhands-tools` declares `openhands-sdk` with no version at all, so a resolver is free to pair a current tools wheel with a sdk from many releases back. #4304 lists that unversioned internal edge in its inventory; this is what it buys in practice. Whether it is what happened to the reporter I cannot say — they installed a packaged build, and I have no visibility into what that bundle resolved. I am not proposing a fix for it here since bounding the internal edges looks like a deliberate release-process decision rather than a drive-by change.

The containment change is untouched by any of this. A tool that cannot start should cost the agent that tool, whatever the reason, and that is the part the reporter is actually blocked on.

<sub>Investigated with AI assistance.</sub>

onatozmenn:
Worth flagging that this review landed on `bdda6e3`, which is one commit behind. `c5daef2` removed the `mcp<2` bound and the lock change entirely, for the reason in my comment above: the graph already bounds mcp through `fastmcp-slim`, and my justification for adding it was wrong. So observations 3 and 4 are about code that is no longer in the PR. They were also both reasonable notes on the version that existed when the review ran.

One thing I would gently push back on. The review restates my original claim that "a fresh resolve picks mcp 2.0.0" as independently verified. It is not what happens: `uv pip compile openhands-tools --python-version 3.12` outside this repo gives browser-use 0.11.13 with mcp 1.29.0. What was verified is the narrower and still-true part, that `browser_use.mcp.server` calls `list_tools` and that mcp 2.0.0 does not have it.

On observation 1, you are right that the failure re-runs per resolution rather than being remembered. I looked at caching it into `is_usable()` so the toolset drops out of `list_usable_tools()` after a failed construction, and decided against it here. `BrowserToolExecutor.__init__` wraps `init_logic` in `run_with_timeout` and raises on `init_timeout_seconds`, so a slow host can produce a failure that a later attempt would not reproduce. Turning that into a process-lifetime disable is a bigger behavioural change than the one this PR is making, and it would land in the same commit as a 500 fix. Happy to open it separately if you want the short-circuit.

On observation 2, agreed in principle, but `create(cls, conv_state, **executor_config)` has no event sink — no `on_event`, only the state. Surfacing "your browser tools are missing and here is why" to the user needs a different seam than this one, which is presumably why the preload service also settles for a log. Also worth noting `is_usable()` already means an agent can legitimately run without browser tools when Chromium is absent, so a silently browser-less agent is a state that exists today rather than one this PR introduces.

On the test scope note: agreed, the autouse `_mock_browser_executor_init` fixture means these tests pin the `try`/`except` boundary rather than a real `CustomBrowserUseServer` construction. Exercising the genuine failure would mean installing an incompatible mcp inside the test run, which is not something I would put in the unit suite.

<sub>Investigated with AI assistance.</sub>

neubig (commented):
Thanks for this! I'm going to try to fix this in a more fundamental way by fixing the browser tool, but if we're not able to do that I think we could incorporate this.

onatozmenn:
Had a look at what fixing the browser tool would take, and I think it might be smaller than it looks.

We never actually serve MCP. The executor just calls the server's methods directly, `_navigate`, `_click`, `_get_browser_state`, `_close_browser`, about sixteen of them across impl.py. Nothing in openhands-tools touches `.server` or `call_tool` or the stdio transport. The only thing that reaches mcp at all is the last line of `BrowserUseServer.__init__`:

```python
# browser_use/mcp/server.py, 0.11.9
def __init__(self, session_timeout_minutes: int = 10):
    ...
    self.server = Server('browser-use')
    self.config = load_browser_use_config()
    self.agent = None
    self.browser_session = None
    self.tools = None
    ...
    self._setup_handlers()   # line 210, this is what blows up
```

Everything we use is already set before that. So we can just override it in `CustomBrowserUseServer`:

```python
def _setup_handlers(self) -> None:
    """We drive the browser methods directly and never serve MCP, so skip
    browser-use's handler registration and the mcp coupling it brings."""
```

and the mcp version stops mattering at all.

I tried it. Faked mcp 2.0.0 by deleting `list_tools` (plus `list_resources`, `list_prompts`, `call_tool`) off `Server`, against the locked browser-use 0.11.9:

```
simulated mcp 2.0.0; Server.list_tools present: False

--- current ---
AttributeError: 'Server' object has no attribute 'list_tools'
    @self.server.list_tools()

--- with _setup_handlers stubbed out ---
constructed: CustomBrowserUseServer
  config: OK          browser_session: OK   active_sessions: OK
  agent: OK           tools: OK             session_timeout_minutes: OK
  llm: OK             file_system: OK       _cleanup_task: OK
  _inject_scripts: OK _recording_session: OK
  missing methods: none
```

Same traceback as the report, and the override clears it with all 14 attributes and all 16 coroutines we call still there. That delattr trick also means we can regression test this without putting mcp 2.0 in CI.

One caveat worth putting in the docstring: it's only safe while `_setup_handlers` stays the last statement in `__init__`. If browser-use ever moves initialisation after it, we'd lose those attributes quietly.

On whether this makes the containment change unnecessary, I don't think it does. There are two other raises in `BrowserToolExecutor.__init__` that have nothing to do with mcp:

* impl.py:395, `init_logic` runs under `run_with_timeout(..., init_timeout_seconds=30)` and a timeout gets re-raised. A loaded host reaches that.
* impl.py:324, `_ensure_chromium_available()`. `is_usable()` checks the same thing, but it's a separate call, so a probe that passes and a launch that doesn't ends up here too.

Both leave through `create()` -> `resolve_tool` -> 500, same as the mcp one, and `ToolPreloadService.start()` already treats that same family of failure as non-fatal one layer up. So the mcp fix takes out today's cause and the `except` stops any of them killing the conversation.

Happy to push the `_setup_handlers` change here, or as its own PR if you'd rather keep this one to the containment bit, or just take it and run if you've already started. Whichever is easiest.

<sub>Investigated with AI assistance.</sub>

enyst (approved):
Thank you!

@neubig I think this PR is fine even if we do more elsewhere: it simply catches the exception.

Issue #4357, opened 2026-08-04
Canvas Extensions: adversarial fuzz suite for entrypoint/bundle paths


VascoSch92:
Fuzz suite beyond the basic traversal/symlink cases already covered in the Phase 1 test matrix.

Blocked by: OpenHands/software-agent-sdk#4353

Issue #4356, opened 2026-08-04
Canvas Extensions: manifest API-scope field + CSRF confirmation + reserved names


VascoSch92:
Manifest-declared API-scope field (declared/logged only at first, groundwork for future enforcement). Confirm CSRF coverage on the new canvas-extensions routes. Reserved-name/path list so an extension can't register a name or page path that collides with core Canvas navigation.

Context: `agent_server`'s current auth (`agent_server/dependencies.py`) is a single shared `X-Session-API-Key` checked via `check_session_api_key()` — no per-route or per-scope authorization exists anywhere in that layer today. An API-scope field on the manifest is genuinely new infrastructure, not an extension of an existing mechanism — there's nothing to hook into yet.

Parallel to the frontend/CLI phases, non-blocking for them.

Blocked by: OpenHands/software-agent-sdk#4352

Issue #4355, opened 2026-08-04
Canvas Extensions: audit logging + agentServer.request telemetry


VascoSch92:
Audit-log events for install/enable/disable/refresh/uninstall (who, when, source, ref). Logging/telemetry for `agentServer.request` calls made by extensions (path + method, at minimum).

Parallel to the frontend/CLI phases, non-blocking for them.

Blocked by: #4352

Issue #4354, opened 2026-08-04
Canvas Extensions: OpenAPI spec + typescript-client publish + capability discovery


VascoSch92:
Publish the OpenAPI spec and the `@openhands/typescript-client` update; add an explicit capability-discovery mechanism instead of relying on 404 probing.

An `@openhands/typescript-client` package/generation pipeline already exists as an established pattern (generated from the agent-server's OpenAPI spec) — follow that existing flow rather than standing up a new one.

Blocked by: #4352

PR #4361, opened 2026-08-04
feat(agent-server): Canvas Extensions manifest and containment [1/4]

VascoSch92:
HUMAN:

First PR to add extension in the agent-server. Just adding the manifest of an extension.

---

AGENT:

## Why

Part of the Canvas Extensions backend build-out (#4348). It's the foundational ticket in the epic: #4349-#4357 all depend on it, either directly (#4350, #4351, #4352 all reference this manifest/installation shape) or need the containment primitive it introduces before any install/serve code can be written safely.

## Summary
- Add `CanvasExtensionManifest` (`schema_version`, `name`, `display_name`, `version`, `description`, `entrypoint`, `contributes.pages[]`) with name/contribution-id/page-path validation. The `name` validator reuses the SDK's existing `validate_extension_name()` instead of duplicating its kebab-case regex.
- Add `resolve_entrypoint()`: filesystem-level containment (`resolve()` + `is_relative_to()`, plus an is-a-file check) that rejects path traversal, symlink escapes, and non-file targets (`.`, a directory, a dangling symlink, a symlink cycle).
- Lives in `openhands-agent-server`, not `openhands-sdk`: Canvas Extensions are installed and served entirely by the agent-server, with nothing consumed by `Agent`/`Conversation`, so this consumes `openhands.sdk.extensions.installation` as a library the same way an external app would (that framework is explicitly designed for reuse this way).

## Issue Number
#4348

## How to Test

```
```
52 tests pass: field validation (including required-field and wrong-type checks), duplicate id/path detection (including through a realistic dict/JSON-shaped `model_validate` call, not just direct nested-model construction), and adversarial entrypoint containment (textual traversal, symlinked file/directory/package-root escapes, a symlink cycle, the pathlib absolute-path-join footgun where `root / "/etc/passwd"` silently discards `root`, and a legitimate in-package symlink that must still resolve so the hardening doesn't overreach).

`uv run pre-commit run --files <changed files>` also passes (ruff format/lint, pyright, import-dependency-rules, tool-subclass-registration).

End-to-end sanity check outside pytest — loads a realistic manifest JSON string, resolves a real entrypoint on disk, then confirms a symlink-escape attack and an invalid name are both rejected with a clear error rather than a stack trace:
```
Loaded manifest OK: weather-widget /weather
Resolved entrypoint OK: /tmp/.../installed/weather-widget/dist/index.js
Symlink escape correctly rejected: entrypoint 'dist/index.js' resolves outside the extension package root
Invalid name correctly rejected: ValidationError
```

## Video/Screenshots

N/A — backend manifest/validation module only, no UI or HTTP route in this PR (the REST API surface is #4352, blocked on this one).

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Scoped strictly to #4348. Everything else in the epic is a separate, already-filed ticket that depends on this one: installation persistence (#4350), staged install/atomic swap (#4351), REST API surface (#4352), OpenAPI/typescript-client (#4354), audit logging (#4355), manifest API-scope + reserved names (#4356), adversarial fuzz suite (#4357). #4349 (`requested_ref` field) is independent SDK work, not touched here.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
@@ -1,0 +1,16 @@
+"""Canvas Extensions: installable UI bundles that contribute pages to Canvas."""
+
+from openhands.agent_server.canvas_extensions.manifest import (
+    CanvasExtensionContributes,
+    CanvasExtensionManifest,
+    CanvasExtensionPage,
+    resolve_entrypoint,
+)
+
+
+__all__ = [
+    "CanvasExtensionManifest",
+    "CanvasExtensionContributes",
+    "CanvasExtensionPage",
+    "resolve_entrypoint",
+]
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/manifest.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/manifest.py
@@ -1,0 +1,164 @@
+"""Canvas Extensions manifest: schema, validation, and entrypoint containment.
+
+A Canvas extension is an installable UI bundle that contributes pages to the
+OpenHands Canvas frontend. Extensions are installed and served entirely by
+the agent-server (via ``openhands.sdk.extensions.installation``, the same
+type-agnostic install-tracking framework Plugins/Skills use); nothing here
+is consumed by ``Agent``/``Conversation``.
+
+This module defines the manifest schema (``canvas-extension.json``) and the
+two security-critical checks around it:
+
+* Name / contribution-id / page-path validation (syntactic, on the model).
+* Entrypoint containment (filesystem-level, once a package root is known) —
+  rejects both textual path traversal and symlink escapes.
+"""
+
+import re
+from pathlib import Path
+
+from pydantic import BaseModel, Field, field_validator
+
+from openhands.sdk.extensions.installation.utils import validate_extension_name
+
+
+# Absolute, kebab-case, multi-segment UI route, e.g. "/dashboard/settings".
+_PAGE_PATH_PATTERN: re.Pattern[str] = re.compile(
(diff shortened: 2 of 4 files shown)


VascoSch92:
Corrected :-) The review here was good.

smolpaws (commented):
🐾 Hi, I'm smolpaws — a feline OpenHands agent doing a review pass to help. Reviewed this layer against its base (`main`).

**Verdict: solid foundation, no blockers.** This is a clean, well-scoped first layer.

What I liked:
- The two-layer containment split is the right design: `_validate_entrypoint` does cheap syntactic rejection (empty / absolute / `..`), and `resolve_entrypoint` (`manifest.py:120`) does the real filesystem check — `resolve()` + `is_relative_to()` + `is_file()`. That correctly catches symlink escapes and non-file targets (`.`, a directory, dangling symlink, cycle) that `is_relative_to` alone would miss. The docstring calling out "must be called again immediately before serving" is exactly the right instinct.
- Reusing `validate_extension_name()` instead of re-deriving the kebab regex keeps one source of truth.
- Tests are genuinely adversarial (symlinked file/dir/package-root escapes, the `root / "/etc/passwd"` footgun, a legit in-package symlink that must still resolve). Nice coverage.

One small thought, not blocking:
- `_validate_entrypoint` rejects POSIX absolute paths (`startswith("/")`) but a Windows-style `C:\...` or backslash traversal would pass the *syntactic* check. On a Linux/Docker agent-server this is moot, and `resolve_entrypoint` is the real gate regardless — so it's belt-and-suspenders, not a hole. Only worth a line of thought if Windows hosts are ever in scope.

Clean layer. 🐾

neubig (approved):
Looks good to me!

Issue #4352, opened 2026-08-04
Canvas Extensions: REST API surface


VascoSch92:
`GET /installed`, `GET /installed/{name}` — list, get, 404 handling<br>`POST /install` — git + local-path sources, always lands disabled<br>`PATCH /installed/{name}` — enable/disable<br>`DELETE /installed/{name}` — uninstall<br>`GET /installed/{name}/bundle` — correct content-type, containment re-checked at serve time (not just install time), cache-busting across revisions

**Existing precedent to mirror:** `plugins_router.py` and `skills_router.py` — the former's own docstring states "this module mirrors skills_router.py" — already ship the same install/list/enable-disable/uninstall/refresh shape for the sibling systems. Reuse their conventions (thin router delegating to a `*_service.py` module, kebab-case path-parameter regex guard, `InstalledPluginInfo`-style response models) rather than inventing new ones.

Blocked by: OpenHands/software-agent-sdk#4350, OpenHands/software-agent-sdk#4351

Issue #4351, opened 2026-08-04
Canvas Extensions: staged install + atomic swap + two-step refresh (check/apply)


VascoSch92:
Write-to-temp + validate-before-swap; atomic swap with rollback on validation failure.

Refresh must be a two-step operation, not a single call:

* **Check** — resolve the tracked source's ref, fetch into a staging location (never the active install path), run manifest + entrypoint validation against the staged content, return `{requested_ref, resolved_ref, validated}`. The active bundle is untouched and keeps serving.
* **Apply** — a separate, explicit, reconfirmed call, atomically swaps staged → active, with the new bundle's own enabled state set to whatever was just confirmed.

**Do not implement this by calling** `InstallationManager.update()` **or** `install(force=True)` — confirmed against current source:

* `update()` (`manager.py:292-317`) resolves `ref=None` ("latest available") and calls `self.install(..., force=True)` — by its own docstring, this "re-fetches... and force-reinstalls."
* `install(force=True)` (`manager.py:106-112`) does `shutil.rmtree(install_path)` immediately followed by `shutil.copytree(fetched_path, install_path)` directly on the live install path — no temp/staging directory at all, and there's a real window mid-swap where the path is empty/partial if anything fails between the two calls.
* This exact pattern is already the live implementation backing Plugins' refresh (`update_plugin()` → `openhands/sdk/plugin/installed.py:136-141`, a direct passthrough to `.update()`).

Suggested shape: stage into a sibling directory (e.g. `<install_dir>/.staging/<name>-<resolved_ref>/`), validate there, then swap with a single atomic rename (`os.replace`, atomic on the same filesystem) rather than rmtree-then-copytree.

Blocked by: OpenHands/software-agent-sdk#4350, OpenHands/software-agent-sdk#4349

Issue #4350, opened 2026-08-04
Canvas Extensions: installation persistence — fix disabled-by-default and self-heal auto-enable


VascoSch92:
On-disk layout under `~/.openhands/canvas-extensions/installed/<name>/`. Reuse `InstallationMetadata`/`InstallationInfo` (`openhands-sdk/openhands/sdk/extensions/installation/`) for `.installed.json` bookkeeping only — **not** `InstallationManager.install()`/`.update()` unmodified.

**Two confirmed issues, with exact locations (verified by running the code against current** `main`**):**

* `InstallationInfo.enabled` defaults to `True` (`info.py:22`: `enabled: bool = Field(default=True, ...)`). `InstallationManager.install()` only overrides this `if existing is not None` (`manager.py:122-126`) — i.e. only on a force-reinstall of an *already-tracked* name. A genuinely new install keeps the field default. Reproduced directly: a throwaway extension installed via the public API came back `enabled=True`.
* `InstallationMetadata.discover_untracked()` (`metadata.py:197-241`), the self-healing reconciliation that runs on every `list_installed()`, builds new entries via `InstallationInfo.from_extension(extension, source="local", install_path=item)` (`metadata.py:238-239`) with no `enabled=` override either — same default applies.
* Already live behavior for Plugins: `install_plugin()` (`openhands/sdk/plugin/installed.py:57-80`) calls `.install()` with no override at all. The test suite's own setup implicitly confirms this — `tests/sdk/extensions/installation/test_installation_manager.py:172-182` has to call `manager.disable(...)` explicitly *before* testing that a force-reinstall preserves the disabled state, because a fresh install is known to land enabled. There's no test anywhere asserting a bare fresh install defaults to disabled.

Canvas Extensions must force `enabled=False` explicitly on every write path that can create a tracked entry — both fresh install and discovery. Neither `install()` nor `InstallationInfo.from_extension()` currently accepts an override for this, so it needs to be an explicit post-write correction, or a new keyword added upstream (which would then need auditing against every existing Plugins/Skills call site so their behavior doesn't silently change too).

**Regression tests:**

* Install without an explicit enable call → assert disabled.
* Attempt to smuggle `enabled: true` in the install request body → assert it's ignored.
* Manually place a valid extension directory in the install root (bypassing the API) → assert it comes back disabled on next list/get.

Blocked by: OpenHands/software-agent-sdk#4348

Issue #4349, opened 2026-08-04
Canvas Extensions: add requested_ref to installation record


VascoSch92:
`InstallationInfo` (`openhands-sdk/openhands/sdk/extensions/installation/info.py`) has `resolved_ref: str | None` (`info.py:25-27`) but no `requested_ref` field — confirmed against current source. Exposing both lets a caller tell whether an install is pinned to an exact SHA/tag or tracking a moving ref.

This is a field on infrastructure shared with Plugins/Skills — the same `InstallationInfo` model backs `openhands/sdk/plugin/installed.py`, and `InstallationMetadata` explicitly migrates legacy `"plugins"`/`"skills"` metadata keys into the same `extensions` dict, confirming this is the common backbone for all three systems, not something canvas-extensions-specific.

Suggested approach: add `requested_ref: str | None = Field(default=None, ...)` — additive and backward-compatible with existing `.installed.json` files written before the field existed. Coordinate with whoever owns `openhands-sdk` before merging, since Plugins/Skills consumers pick up the new field automatically.

No blocking dependency — independent schema change.

PR #4364, opened 2026-08-04
feat(agent-server): Canvas Extensions installation persistence [3/4]

VascoSch92:
HUMAN:

Adding installation persistence for canvas-extension.

---

AGENT:

## Why

Part of the Canvas Extensions backend build-out (#4348). Issue #4350 identified two concrete bugs in reusing the shared `openhands.sdk.extensions.installation` framework (also used by Plugins/Skills) as-is for Canvas Extensions: `InstallationInfo.enabled` defaults to `True`, and neither `InstallationManager.install()` (on a genuinely new install) nor its self-healing directory discovery override that default. A freshly installed or newly discovered canvas extension would come back enabled, when Canvas Extensions require explicit opt-in.

This branch also carries #4349 (`requested_ref` tracking) forward -- it was implemented directly on `extension-canvas-2` but never had its own PR, since it's small, independent SDK work with no canvas-extensions-specific surface.

**Note for reviewers**: this branch is stacked on top of #4361 (still open, not yet merged), so the diff below currently includes #4348's manifest/containment files again alongside the new work. Those shrink out of this diff automatically once #4361 merges into `main`.

## Summary
- Add `openhands-agent-server/openhands/agent_server/canvas_extensions/installed.py`: a thin wrapper around the shared `InstallationManager`, following the same shape as `openhands/sdk/plugin/installed.py`, with on-disk layout at `~/.openhands/canvas-extensions/installed/<name>/`.
- Force `enabled=False` on every write path that creates a *new* tracked entry (fresh install, self-healing discovery) via an explicit post-write correction, while correctly preserving the prior enabled state on a legitimate force-reinstall of an already-tracked extension. Hardened against a stale/tampered `.installed.json` entry with no matching directory being treated as "already installed."
- Wire up `resolve_entrypoint()` containment checking (built in #4348 but never called) at manifest load time, so both install and discovery enforce it.
- Add `MANIFEST_FILENAME` (`Final[str]`) constant to `manifest.py`.
- `update_canvas_extension()` deliberately **not** added -- that's the `install(force=True)`/`.update()` anti-pattern issue #4351 (staged install + atomic swap) exists to replace, not something to introduce here.

## Issue Number
#4350

## How to Test

```
```
77 tests pass (25 for the new persistence module, plus the existing manifest/containment suites). Coverage includes the three regression scenarios named in #4350 (fresh install lands disabled, a smuggled `enabled: true` stale-metadata record is ignored, a manually-placed directory is discovered disabled), force-reinstall state preservation in both directions, mixed-state handling across multiple extensions in one listing call, and rejection/skip paths for malformed manifests, mismatched manifest names, and entrypoint containment escapes at both install and discovery time.

Mutation-tested the two highest-stakes correctness properties directly: neutering the force-disable correction fails 5 of the new tests; disabling the pre-existing/new-entry distinction fails the mixed-state and force-reinstall tests specifically. Confirms the tests exercise real logic, not just happy-path assertions.

`uv run pre-commit run --files <changed files>` passes (ruff format/lint, pyright, import-dependency-rules, tool-subclass-registration).

## Video/Screenshots

N/A -- backend persistence module only, no UI or HTTP route in this PR (REST API surface is #4352, blocked on this one).

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Scoped to #4350 (+ carrying forward the already-implemented #4349). Everything else in the epic remains separate, already-filed tickets: staged install/atomic swap (#4351, blocked on this), REST API surface (#4352), OpenAPI/typescript-client (#4354), audit logging (#4355), manifest API-scope + reserved names (#4356), adversarial fuzz suite (#4357).

### DIFF
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
@@ -1,6 +1,18 @@
 """Canvas Extensions: installable UI bundles that contribute pages to Canvas."""
 
+from openhands.agent_server.canvas_extensions.installed import (
+    InstalledCanvasExtensionInfo,
+    disable_canvas_extension,
+    enable_canvas_extension,
+    get_installed_canvas_extension,
+    get_installed_canvas_extensions_dir,
+    install_canvas_extension,
+    list_installed_canvas_extensions,
+    load_installed_canvas_extensions,
+    uninstall_canvas_extension,
+)
 from openhands.agent_server.canvas_extensions.manifest import (
+    MANIFEST_FILENAME,
     CanvasExtensionContributes,
     CanvasExtensionManifest,
     CanvasExtensionPage,
@@ -12,5 +24,15 @@ __all__ = [
     "CanvasExtensionManifest",
     "CanvasExtensionContributes",
     "CanvasExtensionPage",
+    "MANIFEST_FILENAME",
     "resolve_entrypoint",
+    "InstalledCanvasExtensionInfo",
+    "install_canvas_extension",
+    "uninstall_canvas_extension",
+    "enable_canvas_extension",
+    "disable_canvas_extension",
+    "list_installed_canvas_extensions",
+    "load_installed_canvas_extensions",
+    "get_installed_canvas_extension",
+    "get_installed_canvas_extensions_dir",
 ]
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/installed.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/installed.py
@@ -169,8 +169,25 @@ def list_installed_canvas_extensions(
 def load_installed_canvas_extensions(
     installed_dir: Path | None = None,
 ) -> list[CanvasExtensionManifest]:
-    """Load all enabled canvas extensions' manifests."""
-    return _manager(_resolve_installed_dir(installed_dir)).load_installed()
+    """Load all enabled canvas extensions' manifests.
+
+    Runs through ``list_installed_canvas_extensions`` first so discovery
+    is force-disabled before anything is loaded -- see the module
+    docstring. Mirrors ``InstallationManager.load_installed()``'s own
+    enabled-filter/load logic on top of the corrected info list.
+    """
+    installed_dir = _resolve_installed_dir(installed_dir)
+    manager = _manager(installed_dir)
+    manifests: list[CanvasExtensionManifest] = []
+    for info in list_installed_canvas_extensions(installed_dir):
+        if not info.enabled:
+            continue
+        extension_path = installed_dir / info.name
+        if extension_path.exists():
+            manifests.append(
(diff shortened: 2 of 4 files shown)


smolpaws (commented):
🐾 Hi, I'm smolpaws — a feline OpenHands agent doing a review pass to help. Reviewed this layer against its base (`extension-canvas-2`).

**Verdict: well-reasoned layer, but the disabled-by-default guarantee has one real gap — it isn't enforced on the load path.**

The module's stated promise (docstring) is: *every* newly created entry, whether from `install()` or from self-healing directory discovery, lands `enabled=False` until explicitly enabled. `install_canvas_extension` and `list_installed_canvas_extensions` enforce that via `_force_disable_new`. But:

**1. `load_installed_canvas_extensions()` bypasses the guarantee (`installed.py:173`).**
It calls `manager.load_installed()` directly — no `_force_disable_new`, no `_tracked_names` pre-check (unlike `list_installed_canvas_extensions` at `:166`). Chain: `load_installed()` → `list_installed()` → `session.sync()` → `discover_untracked()`, which builds entries with `InstallationInfo.enabled` defaulting to **True**, persists them, then `load_installed()` returns every enabled one.

So a directory dropped straight into the installed dir — exactly the case your `list_installed_canvas_extensions` docstring says "also lands disabled" — is loaded **enabled** on the first `load_installed_canvas_extensions()` call, and stays enabled because `enabled: true` is now written to `.installed.json`. This contradicts both the module docstring and that function's own docstring.

Note `load_installed()` returns manifests (`T`), not `InstallationInfo`, so you can't just reuse `_force_disable_new` here — you'd want to run the disciplined discover-and-disable pass first, then load only what's still enabled.

**2. Minor — the disable is a second write, not atomic with discovery (`installed.py:166`).**
`_force_disable_new` calls `manager.disable()`, which does a *separate* full metadata load/sync/save per discovered entry, on top of the sync `list_installed()` already ran. Two consequences: (a) discovering N dirs triggers ~N extra directory re-scans + manifest re-parses per list call — correct but wasteful; (b) the "flip to disabled" write is separate from the "discovered as enabled" write, so a reader racing between them can briefly observe the entry enabled. Eventually-consistent, not atomic.

What's sharp: the `_tracked_names` reasoning — a stale metadata record with no matching directory must *not* count as "already installed," or a real install would inherit its `enabled` state as if it were a force-reinstall. That's a subtle correctness point and you got it right. Tests are thorough too.

Fixing #1 is the one I'd want resolved before this ships as the "disabled-by-default" layer. 🐾

enyst (approved):
Seem this little detail might be worth a quick pass?

> So a directory dropped straight into the installed dir — exactly the case your list_installed_canvas_extensions docstring says "also lands disabled" — is loaded enabled on the first load_installed_canvas_extensions() call, and stays enabled because enabled: true is now written to .installed.json. This contradicts both the module docstring and that function's own docstring.

PR #4367, opened 2026-08-04
fix(mcp): reconcile live agent tool snapshots

Shimada666:
HUMAN:

This is a capability I need in practice. Our work requires tools to be updated dynamically, and OpenCode already supports this behavior.

Agent-assisted validation: I reviewed the output from the real FastMCP streamable-HTTP integration tests.
They verified dynamic tool addition, invocation, removal, re-advertisement, and same-name schema replacement.

---

AGENT:

## Why

`notifications/tools/list_changed` currently updates `MCPClient.tools`, but a running agent only receives newly added definitions. Removed tools remain exposed by the agent, and same-name schema changes keep the stale definition and validation model. This completes the stale-runtime behavior identified in [the review of #3894](https://github.com/OpenHands/software-agent-sdk/pull/3894#issuecomment-4994732154) and tracked by #4059.

## Summary

- Atomically replace only the tools owned by the notifying MCP client, including additions, schema updates, and removals.
- Keep the public additions-only callback while binding full-snapshot reconciliation on `MCPClient`, preserving the existing `MCPToolProvider` contract.
- Reconcile against the current Agent instance, including MCP tools loaded by runtime plugins.
- Serialize Agent tool-map reads and writes across concurrent MCP clients.
- Key dynamically generated MCP validation models by both tool name and input schema.
- Document the behavior in OpenHands/docs#685.

## Issue Number

Follow-up to #4059 and #3894.

## How to Test

Run:

```bash
uv run pytest -q tests/sdk/mcp/ tests/sdk/conversation/test_local_conversation_plugins.py tests/sdk/conversation/test_local_conversation_mcp.py tests/sdk/agent/test_filter_tools_regex.py
```

Observed on the current PR head: `178 passed in 9.93s`. This includes a real FastMCP streamable-HTTP server that emits `notifications/tools/list_changed`; the test verifies that a dynamically added tool is callable, disappears from both client and agent after removal, and is callable again after re-advertisement. A focused unit test also warms the old validation model, changes the same tool name to a new input schema, and verifies the new schema accepts new arguments and rejects the old arguments.

Run the repository checks on all changed files:

```bash
[...]

PR #4369, opened 2026-08-04
fix(mcp): close reconciliation gaps left by #4367

VascoSch92:
HUMAN:

Fix some bug introduced in #4367

---

AGENT:

## Why

Follow-up review of #4367 (already merged) found four issues in the new MCP
tool-reconciliation code:

1. `DefaultMCPToolProvider`/`SettingsBackedMCPToolProvider` never forwarded
   `on_tools_reconciled` to `create_mcp_tools()`. `LocalConversation`
   compensated by attaching the callback to the client *after*
   `create_tools()` returned, which is after the client has already
   connected and started listening for `notifications/tools/list_changed`.
   Any such notification arriving in that window is silently dropped —
   reproduced deterministically against a live FastMCP server (see "How to
   Test").
2. `add_runtime_tools` / `_on_mcp_tools_changed` mutate `Agent._tools` in
   place. `Agent.model_copy()` shares private attrs by reference, so a
   dynamically-registered MCP tool can leak into other `Agent` snapshots
   that share the same `_tools` dict (`LocalConversation.load_plugin`
   reassigns `self.agent` via `model_copy()` right after wiring MCP
   callbacks to the pre-copy agent).
3. `AgentBase._initialize` computed an unused `tool_names` list.
4. The per-tool MCP validation-model cache (`_mcp_dynamic_action_type`) is
   keyed by `(name, schema)` with no eviction, so a tool whose schema keeps
   changing grows it without bound.

## Summary

- Thread `on_tools_reconciled` through `MCPToolProvider.create_tools()` /
  `create_mcp_tools()` instead of setting it on the client after the fact.
- Replace `Agent._tools` instead of mutating it in place in
  `add_runtime_tools` / `_on_mcp_tools_changed`, matching
  `_on_mcp_tools_reconciled`.
- Remove the unused `tool_names` assignment in `AgentBase._initialize`.
- Bound `_mcp_dynamic_action_type` to an LRU of 512 entries.

## Issue Number

Follow-up to #4367.

## How to Test

Run:

```bash
uv run pytest -q tests/sdk/mcp/ tests/sdk/conversation/test_local_conversation_plugins.py tests/sdk/conversation/test_local_conversation_mcp.py tests/sdk/agent/test_filter_tools_regex.py tests/agent_server/test_mcp_oauth_store.py
```

Observed: `1710 passed` across the broader `tests/sdk/mcp/`,
`tests/sdk/conversation/`, `tests/sdk/agent/` suites plus the agent-server
OAuth store tests.

Before writing the fix, item 1 was confirmed with a standalone script
(not included in this PR) that ran a real FastMCP HTTP server, forced a
`tools/list_changed` notification to fire while `DefaultMCPToolProvider`
was still inside `create_tools()`, and observed `on_tools_reconciled` never
firing for the tool added by that notification even though the client's
own `client.tools` was updated correctly. `test_default_provider_wires_on_tools_reconciled_before_connect`
(new) asserts the callback is attached to the client synchronously, before
`create_tools()` returns — closing that window by construction.

Item 2 is covered by the new
`test_add_runtime_tools_does_not_leak_into_model_copy`, which registers a
tool on a `model_copy()`'d `Agent` and asserts the original is unaffected
(this failed before the fix).

Item 4 is covered by `test_action_type_cache_is_bounded`.

Also ran, on all changed files:

```bash
git diff --name-only main | xargs uv run ruff format --check
git diff --name-only main | xargs uv run ruff check
git diff --name-only main | xargs uv run pyright
PIP_INDEX_URL=https://pypi.org/simple uv run pre-commit run --files $(git diff --name-only main)
```

All passed (pre-commit: ruff format, ruff lint, pycodestyle, pyright,
import dependency rules, tool-subclass registration).

## Video/Screenshots

Not applicable; this is SDK runtime behavior covered by unit tests and a
live FastMCP server test.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No public API changes: `on_tools_reconciled` was already a keyword-only
parameter on `create_mcp_tools()`; this PR just makes
`MCPToolProvider.create_tools()` implementations forward it instead of
silently dropping it. `_tools_reconciled_callback` is still set as an
attribute on the returned `MCPClient`, just earlier (inside
`create_mcp_tools()`, before connecting) instead of by the caller
afterward.

---

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:9f7327a-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-9f7327a-python \
  ghcr.io/openhands/agent-server:9f7327a-python
```

```
ghcr.io/openhands/agent-server:9f7327a-golang-amd64
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-golang-amd64
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-golang-amd64
ghcr.io/openhands/agent-server:9f7327a-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:9f7327a-golang-arm64
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-golang-arm64
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-golang-arm64
ghcr.io/openhands/agent-server:9f7327a-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:9f7327a-java-amd64
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-java-amd64
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-java-amd64
ghcr.io/openhands/agent-server:9f7327a-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:9f7327a-java-arm64
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-java-arm64
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-java-arm64
ghcr.io/openhands/agent-server:9f7327a-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:9f7327a-python-amd64
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-python-amd64
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-python-amd64
ghcr.io/openhands/agent-server:9f7327a-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:9f7327a-python-arm64
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-python-arm64
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-python-arm64
ghcr.io/openhands/agent-server:9f7327a-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:9f7327a-golang
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-golang
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-golang
ghcr.io/openhands/agent-server:9f7327a-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:9f7327a-java
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-java
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-java
ghcr.io/openhands/agent-server:9f7327a-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:9f7327a-python
ghcr.io/openhands/agent-server:9f7327ad9505cf28d3fe5bd526d050c214c014f3-python
ghcr.io/openhands/agent-server:fix-mcp-tool-reconciliation-followups-python
ghcr.io/openhands/agent-server:9f7327a-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `9f7327a-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `9f7327a-python-amd64`) are also available if needed

### DIFF
--- a/openhands-agent-server/openhands/agent_server/mcp_oauth_store.py
+++ b/openhands-agent-server/openhands/agent_server/mcp_oauth_store.py
@@ -26,6 +26,7 @@ from openhands.sdk.mcp.config import (
 )
 from openhands.sdk.mcp.utils import (
     ToolsChangedCallback,
+    ToolsReconciledCallback,
     create_mcp_tools,
 )
 
@@ -337,12 +338,14 @@ class SettingsBackedMCPToolProvider:
         timeout: float = 30.0,
         *,
         on_tools_changed: ToolsChangedCallback | None = None,
+        on_tools_reconciled: ToolsReconciledCallback | None = None,
     ) -> MCPClient:
         return create_mcp_tools(
             mcp_config,
             timeout,
             mcp_oauth_token_storage=MCPSettingsOAuthTokenStore(),
             on_tools_changed=on_tools_changed,
+            on_tools_reconciled=on_tools_reconciled,
         )
 
 
--- a/openhands-sdk/openhands/sdk/agent/base.py
+++ b/openhands-sdk/openhands/sdk/agent/base.py
@@ -564,7 +564,6 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
         if self.filter_tools_regex:
             pattern = re.compile(self.filter_tools_regex)
             tools = [tool for tool in tools if pattern.match(tool.name)]
-            tool_names = [tool.name for tool in tools]
             logger.info("Filtered to %d tools after applying regex filter", len(tools))
 
         # Include default tools from include_default_tools; not subject to regex
@@ -876,9 +875,10 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
             if existing:
                 raise ValueError(f"Duplicate tool names found: {existing}")
 
-            # AgentBase is frozen, so update its mutable tool map in place.
-            for tool in tools:
-                self._tools[tool.name] = tool
+            # AgentBase is frozen; replace the tool map rather than mutating
+            # it in place, so Agent.model_copy() snapshots don't share state.
+            updated = {**self._tools, **{tool.name: tool for tool in tools}}
+            object.__setattr__(self, "_tools", updated)
 
     def _on_mcp_tools_changed(self, tools: Sequence[ToolDefinition]) -> None:
         """Handle dynamically advertised MCP tools.
@@ -930,8 +930,12 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
                 )
 
             self.add_runtime_tools(additions)
-            for tool in replacements:
-                self._tools[tool.name] = tool
+            if replacements:
+                updated = {
+                    **self._tools,
+                    **{tool.name: tool for tool in replacements},
+                }
+                object.__setattr__(self, "_tools", updated)
 
         if additions:
             logger.info(
--- a/openhands-sdk/openhands/sdk/agent/base.py
+++ b/openhands-sdk/openhands/sdk/agent/base.py
@@ -871,9 +871,30 @@ class AgentBase(DiscriminatedUnionMixin, ABC):
             }
             raise ValueError(f"Duplicate runtime tool names found: {duplicates}")
         with self._tools_lock:
-            existing = set(self._tools) & set(tool_names)
-            if existing:
-                raise ValueError(f"Duplicate tool names found: {existing}")
+            # A tools/list_changed notification can race the caller: if it
+            # arrives while the provider's initial create_tools() call is
+            # still in flight, _on_mcp_tools_changed/_on_mcp_tools_reconciled
+            # may already have installed the same tool via this same client
+            # before the caller's own add_runtime_tools() call (using the
+            # client's returned snapshot) gets here. Treat that as a refresh,
+            # not a conflict, matching the same-client exemption already used
+            # by _on_mcp_tools_changed/_on_mcp_tools_reconciled.
+            conflicts: set[str] = set()
+            for tool in tools:
+                existing_tool = self._tools.get(tool.name)
+                if existing_tool is None:
+                    continue
+                existing_executor = existing_tool.executor
+                replacement_executor = tool.executor
+                if (
+                    isinstance(existing_executor, MCPToolExecutor)
+                    and isinstance(replacement_executor, MCPToolExecutor)
+                    and existing_executor.client is replacement_executor.client
+                ):
+                    continue
+                conflicts.add(tool.name)
+            if conflicts:
+                raise ValueError(f"Duplicate tool names found: {conflicts}")
 
             # AgentBase is frozen; replace the tool map rather than mutating
             # it in place, so Agent.model_copy() snapshots don't share state.
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -72,6 +72,7 @@ from openhands.sdk.mcp.utils import (
     MCPToolProvider,
     ToolsChangedCallback,
     ToolsReconciledCallback,
+    provider_supports_on_tools_reconciled,
 )
 from openhands.sdk.observability.laminar import OPERATION_METADATA_KEY, observe
 from openhands.sdk.plugin import (
@@ -1288,11 +1289,17 @@ class LocalConversation(BaseConversation):
         mcp_config = enabled_mcp_servers(mcp_config)
         if not mcp_config:
             return []
+        create_kwargs: dict[str, Any] = {"on_tools_changed": on_tools_changed}
+        if provider_supports_on_tools_reconciled(self._mcp_tool_provider):
+            create_kwargs["on_tools_reconciled"] = on_tools_reconciled
+        elif on_tools_reconciled is not None:
+            logger.debug(
+                "%s does not accept on_tools_reconciled; dynamic MCP tool "
+                "removals/updates won't reach the agent for this provider",
+                type(self._mcp_tool_provider).__name__,
+            )
         client = self._mcp_tool_provider.create_tools(
-            mcp_config,
-            _RUNTIME_MCP_TIMEOUT_SECS,
-            on_tools_changed=on_tools_changed,
-            on_tools_reconciled=on_tools_reconciled,
+            mcp_config, _RUNTIME_MCP_TIMEOUT_SECS, **create_kwargs
         )
         return list(client.tools)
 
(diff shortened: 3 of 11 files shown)


neubig (approved):

Shimada666:
Thank you very much for taking the time to follow up on my PR and fix the gaps in #4367. I especially appreciate the callback wiring, copy-on-write tool map updates, bounded cache, and the compatibility/concurrency follow-ups.

While validating the current head (`40d6a07ed6496a2bdbb9cf63425105e3f8de13a7`), I found one startup ordering case that I wanted to flag for your judgment.

If a real MCP server sends `notifications/tools/list_changed` while the initial `tools/list` request is still in flight, the newly wired reconciliation callback can update the already-initialized Agent before `create_tools()` returns. `_ensure_agent_ready()` then calls `add_runtime_tools()` with the same returned snapshot and raises a duplicate-tool error.

I reproduced this through the full `LocalConversation._ensure_agent_ready()` path with a real FastMCP Streamable HTTP server. From the repository root on the PR head, run:

```bash
OPENHANDS_SUPPRESS_BANNER=1 uv run python - <<'PY'
import asyncio
import socket
import tempfile
import threading
import time
from pathlib import Path

import mcp.types as mt
from fastmcp import FastMCP
from fastmcp.server.middleware import Middleware, MiddlewareContext

from openhands.sdk import Agent, LLM
from openhands.sdk.mcp.config import coerce_mcp_config


class NotifyDuringInitialList(Middleware):
    def __init__(self):
        self.first = True

    async def on_list_tools(self, context: MiddlewareContext, call_next):
        tools = await call_next(context)
        if self.first:
            self.first = False
            assert context.fastmcp_context is not None
            await context.fastmcp_context.send_notification(
                mt.ToolListChangedNotification()
            )
            # Make the protocol-valid ordering deterministic: let the refresh
            # triggered by the notification finish before this first list returns.
            await asyncio.sleep(0.5)
        return tools


with socket.socket() as sock:
    sock.bind(("127.0.0.1", 0))
    port = sock.getsockname()[1]

server = FastMCP("startup-race")
server.add_middleware(NotifyDuringInitialList())


@server.tool()
def only_tool(value: int = 1) -> int:
    return value


def run_server():
    asyncio.run(
        server.run_http_async(
            host="127.0.0.1",
            port=port,
            transport="http",
            show_banner=False,
            path="/mcp",
        )
    )


threading.Thread(target=run_server, daemon=True).start()
for _ in range(100):
    try:
        with socket.create_connection(("127.0.0.1", port), timeout=0.1):
            break
    except OSError:
        time.sleep(0.05)
else:
    raise RuntimeError("FastMCP server did not start")

with tempfile.TemporaryDirectory() as workspace:
    agent = Agent(
        llm=LLM(model="test-model", api_key=SecretStr("k")),
        tools=[],
        include_default_tools=[],
        mcp_config=coerce_mcp_config(
            {
                "race": {
                    "transport": "http",
                    "url": f"http://127.0.0.1:{port}/mcp",
                }
            }
        ),
    )
    conversation = LocalConversation(
        agent=agent,
        workspace=Path(workspace),
        visualizer=None,
    )

    try:
        conversation._ensure_agent_ready()
    except ValueError as exc:
        assert "Duplicate tool names found" in str(exc)
        print(type(exc).__name__, str(exc))
    else:
        conversation.close()
        raise AssertionError("Expected duplicate-tool initialization failure")
PY
```

Observed output:

```text
Registered 1 dynamically advertised MCP tools: only_tool
Created 1 MCP tools
ValueError Duplicate tool names found: {'only_tool'}
```

The `0.5s` delay only makes the ordering deterministic; the notification itself is sent through the real FastMCP session over Streamable HTTP. The existing `test_default_provider_wires_on_tools_reconciled_before_connect` verifies that the callback is attached, but it does not force a notification to complete during the initial connection window.

Would you mind judging whether this case should be handled in this PR? My initial thought is that initial snapshot installation and reconciliation may need to be serialized or made idempotent, but please feel free to choose a better approach. Thank you again for improving the original change.

VascoSch92:
Hey @Shimada666 

thanks for the follow-up. 

I think you are right. The fix was minimal so I pushed it to this PR.

Can you review it?

Shimada666:
@VascoSch92  Thanks for the quick follow-up and for pushing the fix. I took another look at it with some AI-assisted review and ran the real FastMCP reproduction again.

The new change does fix the duplicate-tool exception when the notification and the initial response contain the same tool snapshot. I found one small variation of the startup edge case, though: if the notification represents an actual schema change, the later-arriving initial response can still overwrite the newer snapshot.

The sequence I tested was:

1. The initial `tools/list` captures `changing(old_value: int)` but delays its response.
2. The server replaces it with `changing(new_value: str)` and sends `tools/list_changed`.
3. The notification refresh completes first and installs the new schema.
4. The delayed initial response returns, and the same-client exemption in `add_runtime_tools()` replaces it with the old schema again.

Using the same real FastMCP Streamable HTTP setup and full `LocalConversation._ensure_agent_ready()` path, the current head starts successfully, but I observed:

```text
ensure_agent_ready=PASS
agent_schema  = old_value
client_schema = old_value
expected latest schema = new_value
```

So the crash is gone, but a genuinely newer schema can still be rolled back silently. A removal during the same window may have a similar stale-snapshot issue. The added unit test uses the same `Tool` object for both registrations, so it does not exercise this newer-snapshot/older-snapshot ordering.

This looks like a small startup timing edge case, and I found it while investigating together with AI, so please use your judgment on whether it is worth handling in this PR or in another way. My only suggestion would be to keep any fix scoped to MCP initialization/snapshot ordering rather than broadly changing `add_runtime_tools()` semantics. Thanks again for taking the time to improve this.

Issue #4348, opened 2026-08-04
Canvas Extensions: manifest model, validation, and entrypoint containment


VascoSch92:
Pydantic manifest model (`schema_version`, `name`, `display_name`, `version`, `description`, `entrypoint`, `contributes.pages[]`); name/contribution-ID/page-path validation; entrypoint containment check rejecting path traversal and symlink escapes, with its own adversarial tests (security-critical, don't fold into generic validation work).

**Implementation notes (verified against current `software-agent-sdk` source):**
- Reuse `validate_extension_name()` (`openhands-sdk/openhands/sdk/extensions/installation/utils.py:8-15`) for the `name` field — already enforces kebab-case (`^[a-z0-9]+(-[a-z0-9]+)*$`), the same rule Plugins/Skills use today for path-traversal-safe extension names.
- `plugins_router.py:41` defines `PLUGIN_NAME_PATTERN = r"^[a-z0-9]+(?:-[a-z0-9]+)*$"` as a FastAPI `Path(pattern=...)` guard on the `{name}` route parameter, on top of the SDK-level check — defense in depth at the routing layer. Mirror this for the canvas-extensions routes.
- `entrypoint` containment is **not** covered by `validate_extension_name` — that only validates the extension's own name, not paths inside its manifest. Needs its own check: resolve `entrypoint` against the installed package root and reject anything that escapes it. Watch for symlinks specifically, since path resolution follows them.

Part of the Canvas Extensions backend build-out.

PR #4374, opened 2026-08-05
feat(agent-server): Canvas Extensions staged refresh (check/apply) [4/4]

VascoSch92:
HUMAN:

Adding staged install + atomic swap + two-step refresh (check/apply) for canvas extensions.

---

AGENT:

## Why

Part of the Canvas Extensions backend build-out (#4348). Issue #4351 identified two concrete bugs in reusing `InstallationManager.update()`/`install(force=True)` for refresh: both resolve `ref=None` ("latest available") even when the extension was pinned to a specific ref, and both `shutil.rmtree` the live install path immediately followed by `shutil.copytree` directly onto it -- no staging directory, and a real window where the path is empty/partial if anything fails mid-swap.

**Note for reviewers**: this branch is stacked on top of #4364 (still open, not yet merged), so the diff below currently includes #4364/#4361's files again alongside the new work. Those shrink out of this diff automatically once #4364 merges into `extension-canvas-2`.

## Summary
- Add `check_canvas_extension_update()`: re-fetches the tracked source at its *originally requested* ref (never silently reset to "latest") into a `.staging/` sibling directory and validates it there (manifest + entrypoint containment + a name-consistency check against the tracked extension). The active install is never touched; returns `{requested_ref, resolved_ref, validated}`.
- Add `apply_canvas_extension_update()`: a separate, explicit call that reconfirms the exact validated `resolved_ref` from a prior check, then atomically swaps staged → active via two `os.replace` renames (POSIX `rename()` can't atomically replace a non-empty directory in one step, so the active bundle moves aside first) with rollback if the second rename fails. The new bundle's `enabled` state is always taken from the caller, never inherited.
- **Security fix found during review**: `resolved_ref` was being used unsanitized to build the staging filesystem path, so a caller passing `../../../evil` or an absolute path could make `apply()` swap in content from anywhere on disk as the extension's new active bundle. Now rejects absolute paths, `..` segments, and empty strings (mirrors the existing `entrypoint` containment check in `manifest.py`).
- `uninstall_canvas_extension()` now also drops any staged-but-unapplied update for that name, so it can't outlive the extension it belongs to.
- Extracted the `write_extension()` helper and `extension_dir`/`installed_dir` fixtures (previously copy-pasted verbatim across two test files) into a shared `tests/agent_server/canvas_extensions/conftest.py`.

## Issue Number
#4351

## How to Test

```
```
114 tests pass (37 new, in `test_canvas_extensions_refresh.py`). Coverage includes: happy-path check/apply and the active install staying untouched by `check()`; the originally requested ref surviving a check instead of being reset to "latest" (the exact bug being replaced, verified by asserting the mocked fetch call's `ref` kwarg); invalid manifest, entrypoint-escape, and manifest-name-mismatch staged content all marking `validated=False` and cleaning up `.staging/` without touching the active install; `apply()` without a prior check, with a mismatched `resolved_ref`, or re-applying an already-consumed one all raising; explicit enable/disable reconfirmation on `apply()` never being inherited from the prior state; atomic-swap rollback plus a successful retry after a simulated mid-swap failure (`os.replace` patched to fail on its second call); `apply()`'s defense-in-depth re-validation immediately before swapping; cross-extension staging isolation; and the path-traversal/absolute-path/empty-string `resolved_ref` rejections for the security fix.

Manually reproduced the path-traversal bug against the pre-fix code with a standalone script: two extensions installed, `apply_canvas_extension_update("victim", resolved_ref="../../../evil", ...)` and an absolute-path variant both swapped attacker-controlled content into the active install before the fix (confirmed by reading back the resulting `canvas-extension.json` version); after the fix both raise `ValueError` and the legitimate apply path still succeeds.

`uv run pre-commit run --files <changed files>` passes (ruff format/lint, pycodestyle, pyright, import-dependency-rules, tool-subclass-registration).

## Video/Screenshots

N/A -- backend refresh module only, no UI or HTTP route in this PR (REST API surface is #4352, blocked on this one).

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Scoped to #4351. Everything else in the epic remains separate, already-filed tickets: REST API surface (#4352), OpenAPI/typescript-client (#4354), audit logging (#4355), manifest API-scope + reserved names (#4356), adversarial fuzz suite (#4357).

### DIFF
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
@@ -1,7 +1,10 @@
 """Canvas Extensions: installable UI bundles that contribute pages to Canvas."""
 
 from openhands.agent_server.canvas_extensions.installed import (
+    CanvasExtensionUpdateCheck,
     InstalledCanvasExtensionInfo,
+    apply_canvas_extension_update,
+    check_canvas_extension_update,
     disable_canvas_extension,
     enable_canvas_extension,
     get_installed_canvas_extension,
@@ -35,4 +38,7 @@ __all__ = [
     "load_installed_canvas_extensions",
     "get_installed_canvas_extension",
     "get_installed_canvas_extensions_dir",
+    "CanvasExtensionUpdateCheck",
+    "check_canvas_extension_update",
+    "apply_canvas_extension_update",
 ]
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/installed.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/installed.py
@@ -1,31 +1,43 @@
-"""Installed Canvas Extensions: the disabled-by-default guarantee.
+"""Installed Canvas Extensions.
 
-Built on ``openhands.sdk.extensions.installation``, the shared install-
-tracking framework Plugins/Skills also use. Not reused unmodified though:
-``InstallationInfo.enabled`` defaults to ``True``, and neither
-``InstallationManager.install()`` nor its self-healing directory discovery
-override that for a genuinely new entry (only a force-reinstall of an
-already-tracked name preserves its prior state). This module corrects that,
-as an explicit post-write step, so any newly created entry — from an
-install or from discovery — lands disabled until explicitly enabled.
+Built on ``openhands.sdk.extensions.installation``, shared with Plugins/
+Skills, with two behaviors specific to this module:
+
+Disabled by default -- ``InstallationInfo.enabled`` defaults to ``True``
+and neither ``InstallationManager.install()`` nor its directory discovery
+override that for a new entry, so every write path that can create one
+forces ``enabled=False`` as an explicit post-write step.
+
+Staged refresh -- ``check``/``apply`` replace ``update()``/
+``install(force=True)``, which rmtree+copytree straight onto the live
+install path with no staging directory. ``check`` fetches and validates
+into ``.staging/`` without touching the active install; ``apply`` swaps it
+in via two atomic renames (POSIX can't atomically replace a non-empty
+directory in one rename), rolling back if the second fails.
 """
 
+import os
+import shutil
 from pathlib import Path
+
+from pydantic import BaseModel, Field, ValidationError
 
 from openhands.agent_server.canvas_extensions.manifest import (
     MANIFEST_FILENAME,
     CanvasExtensionManifest,
     resolve_entrypoint,
 )
+from openhands.sdk.extensions.fetch import fetch_with_resolution
 from openhands.sdk.extensions.installation import (
     InstallationInfo,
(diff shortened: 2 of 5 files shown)


smolpaws (commented):
🐾 Hi, I'm smolpaws — a feline OpenHands agent doing a review pass to help. Reviewed this layer against its base (`extension-canvas-3`).

**Verdict: the staged check/apply design is the right call.** No rmtree onto the live path, re-validation before the swap, symlink-aware staging. I went looking for the scary failure modes and most hold up — two crash-window durability gaps are worth noting, both low-severity, neither an everyday-path bug.

**1. Metadata is written after the filesystem swap (`installed.py:383` vs. swap at `:366`).**
`os.replace(staged_path, install_path)` swaps the bundle, then the metadata write happens afterward. A crash in between leaves the new bundle live on disk but metadata still recording the *old* `resolved_ref` and the *old* `enabled` state — and the caller's requested `enabled` is silently lost. Because `discover_untracked` skips names already tracked, self-healing never reconciles `resolved_ref` back to reality. The swap is advertised as atomic; the tracked state isn't atomic with it. Worth either folding the metadata into the recovery story or reconciling `resolved_ref` on load.

**2. The rollback rename isn't itself guarded (`installed.py:364-368`).**
If the second `os.replace` (`:366`) fails and the rollback `os.replace(backup_path, install_path)` (`:368`) *also* fails, `install_path` is left missing while metadata still tracks the extension (tracked-but-missing until manual repair). Double-failure, low probability — but since the swap claims a clean rollback, a last-ditch check or a comment naming the residual `.previous` backup would make the guarantee honest.

**Things I checked and cleared (not issues):**
- `_staged_path` traversal guard is sufficient: `name` is kebab-validated via `manager.get()`, and the caller-supplied `resolved_ref` is rejected on empty/absolute/`..`.
- `apply` re-runs `_load_validated_manifest` (re-parse + `resolve_entrypoint` containment + name match) *before* swapping, so tampering with staging between `check` and `apply` is caught. Good defense-in-depth.
- No EXDEV risk — `.staging` and the install path are both under `installed_dir`, same filesystem.
- `copytree(symlinks=True)` is deliberate and correct — dereferencing would defeat the containment check.
- Setting `enabled` from the caller in `apply` is right for an update (vs. disabled-by-default for a fresh install).
- `uninstall` now also drops the staging slot — nice cleanup.

One cross-layer note: this layer doesn't touch `load_installed_canvas_extensions`, so the disabled-by-default gap I flagged on #4364 still stands in the final state.

Nice work — the atomic-swap comment explaining "POSIX can't replace a non-empty dir in one rename" is the kind of thing that saves the next reader. 🐾

enyst (approved):
Thank you for all these! AI reviews flagged a couple of low-probability inconsistencies here that seem acceptable to me.

PR #4375, opened 2026-08-05
feat(sdk): track requested_ref alongside resolved_ref in InstallationInfo [2/4]

VascoSch92:
HUMAN:

Adding requested_ref to InstallationInfo, tracked forward from extension-canvas-2.

---

AGENT:

## Why

Part of the Canvas Extensions backend build-out (#4348). Issue #4349 noted that `InstallationInfo` had `resolved_ref: str | None` but no `requested_ref` -- there was no way for a caller to tell whether an install is pinned to an exact SHA/tag or tracking a moving ref. This is infrastructure shared with Plugins/Skills (the same `InstallationInfo` model backs `openhands/sdk/plugin/installed.py`), not something canvas-extensions-specific, so it's additive and backward-compatible with `.installed.json` files written before the field existed.

This work was originally implemented directly on `extension-canvas-2` without its own PR (it's small, independent SDK work with no canvas-extensions-specific surface) -- `#4364`'s description says as much. It's being opened here, stacked between #4361 and #4364, so every branch in the stack has a corresponding review PR.

## Summary
- Add `requested_ref: str | None` to `InstallationInfo` (`openhands-sdk/openhands/sdk/extensions/installation/info.py`), alongside the existing `resolved_ref`.
- Thread it through `InstallationInfo.from_extension()` and `InstallationManager.install()` (`manager.py`), which now records the `ref` it was called with as `requested_ref`.
- `InstallationManager.update()` is unchanged -- it still resolves `ref=None` on every call, so a previously pinned `requested_ref` is cleared back to `None` (tracking latest) on update. That's an existing, separate behavior gap (not touched by this field addition) that #4351's staged refresh replaces `update()`/`install(force=True)` specifically to fix.

## Issue Number
#4349

## How to Test

```
uv run pytest tests/sdk/extensions/installation -v
```
58 tests pass, run against this branch (`extension-canvas-2`) in isolation via a separate worktree rather than the later stacked branches, to verify this diff on its own: `test_installation_info_from_extension_with_refs` (requested_ref/resolved_ref both recorded when provided), `test_install_records_requested_ref` (ref passed to `install()` is recorded as `requested_ref`, separate from the resolved SHA), `test_install_without_ref_leaves_requested_ref_none` (omitting `ref` leaves `requested_ref` unset while `resolved_ref` is still recorded), and `test_update_clears_requested_ref_to_track_latest` (documents the existing `update()` ref=None behavior above).

`uv run pre-commit run --files <changed files>` also passes (ruff format/lint, pycodestyle, pyright, import-dependency-rules, tool-subclass-registration).

## Video/Screenshots

N/A -- SDK model/field addition only, no UI or HTTP route in this PR.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Scoped to #4349, independent SDK work with no canvas-extensions-specific surface. Stacked between #4361 (manifest/containment) and #4364 (installation persistence, which already carries this work forward) purely to complete the review chain -- no new behavior beyond what #4364 already describes.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/__init__.py
@@ -1,0 +1,16 @@
+"""Canvas Extensions: installable UI bundles that contribute pages to Canvas."""
+
+from openhands.agent_server.canvas_extensions.manifest import (
+    CanvasExtensionContributes,
+    CanvasExtensionManifest,
+    CanvasExtensionPage,
+    resolve_entrypoint,
+)
+
+
+__all__ = [
+    "CanvasExtensionManifest",
+    "CanvasExtensionContributes",
+    "CanvasExtensionPage",
+    "resolve_entrypoint",
+]
--- a/openhands-agent-server/openhands/agent_server/canvas_extensions/manifest.py
+++ b/openhands-agent-server/openhands/agent_server/canvas_extensions/manifest.py
@@ -1,0 +1,164 @@
+"""Canvas Extensions manifest: schema, validation, and entrypoint containment.
+
+A Canvas extension is an installable UI bundle that contributes pages to the
+OpenHands Canvas frontend. Extensions are installed and served entirely by
+the agent-server (via ``openhands.sdk.extensions.installation``, the same
+type-agnostic install-tracking framework Plugins/Skills use); nothing here
+is consumed by ``Agent``/``Conversation``.
+
+This module defines the manifest schema (``canvas-extension.json``) and the
+two security-critical checks around it:
+
+* Name / contribution-id / page-path validation (syntactic, on the model).
+* Entrypoint containment (filesystem-level, once a package root is known) —
+  rejects both textual path traversal and symlink escapes.
+"""
+
+import re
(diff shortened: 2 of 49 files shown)


smolpaws (commented):
🐾 Hi, I'm smolpaws — a feline OpenHands agent doing a review pass to help. Reviewed this layer against its base (`extension-canvas-1`).

**Verdict: clean, ship it.** Small, self-contained, nothing to change.

- `requested_ref` is threaded cleanly: added to `InstallationInfo` (`info.py:25`), plumbed through `from_extension`, and set from `ref` in `manager.install()`. The separation from `resolved_ref` (what you asked for vs. the SHA it pinned to) is a useful distinction to record.
- The one semantic worth flagging is already flagged and tested: `update()` deliberately clears `requested_ref` because it re-fetches with `ref=None` (`test_update_clears_requested_ref_to_track_latest`). That's the kind of behavior that bites later if it's implicit — good that it's both documented and covered.

No issues. 🐾

enyst (approved):
LGTM (and to my agent 😹)

Issue #3153, opened 2026-05-08
perf: SDK and agent-server performance and correctness issues from profiling investigation


csmith49:
## Overview

Profiling investigation of openhands-sdk v1.19.1 and openhands-agent-server v1.19.1 uncovered 19 performance and correctness issues across the SDK and agent server. This tracking issue organizes them by severity.

Methodology: Austin wall-time stack sampling profiler + targeted benchmarks using `TestLLM` (scripted responses, zero LLM cost) to isolate SDK/server overhead from LLM latency. Conversations scaled from 50 to 2000 steps, 1 to 16 concurrent threads.

**Broadly related upstream work:**
- #3101 — test: add agent-server stress suite and known-bug xfail markers (open PR)
- #3053 — Make View construction incremental and tied to ConversationState (open issue)

## Critical — Correctness and scaling blockers

- [x] #3134
- [x] #3135
- [x] #3136
- [x] #3137
- [x] #3138
- [x] #3139

## High — Significant performance and resource issues

- [x] #3140
- [x] #3141
- [x] #3142
- [x] #3143
- [x] #3144
- [x] #3145
- [x] #3146
- [x] #3147
- [x] #3148
- [x] #3149

## Medium — Optimization opportunities

- [x] #3150
- [x] #3151
- [x] #3152

## Key findings summary

| Finding | Severity | Impact |
|---------|----------|--------|
| O(N²) per-step cost from full-history rescan | Critical | 9.1× slowdown at 2000 steps |
| GIL prevents threading from scaling | Critical | Flat throughput regardless of thread count |
| atexit leak pins conversations in memory | Critical | OOM on long-running servers |
| litellm.modify_params race condition | Critical | Silent LLM call corruption |
| os.environ credential bleed | Critical | Cross-tenant credential exposure |
| SpanManager global stack | Critical | Broken distributed tracing |
| Eager conversation loading at startup | High | O(N) startup time |
| No eviction for finished conversations | High | Monotonic memory growth |
| O(N) conversation search/count | High | Slow list API at scale |
| Default thread pool, no backpressure | High | Silent queuing of conversations |
| Lease renewal I/O storm | High | ~33 file ops/s at 500 conversations |
| ConversationState autosave per mutation | High | 10-100× excess I/O per step |
| Unbounded InMemoryFileStore cache | High | ~500 MB at 500 conversations |
| send_message TOCTOU race | High | Double-execution of conversations |
[...]

PR #4389, opened 2026-08-06
fix(agent-server): bound native async runs with max_concurrent_runs

### LINKED ISSUE #4063: bug(agent-server): max_concurrent_runs does not limit native async conversations
## Problem

`Config.max_concurrent_runs` is documented as the maximum number of conversations that can execute agent steps concurrently, but it currently only controls the size of the dedicated `ThreadPoolExecutor` used by the synchronous `conversation.run()` fallback.

`EventService.run()` prefers the native async path for normal `LocalConversation` instances whose agents override `astep()`:

```python
if has_native_arun:
    await conversation.arun()
else:
    await loop.run_in_executor(self._run_executor, conversation.run)
```

As a result, `OH_MAX_CONCURRENT_RUNS=N` does not constrain native async conversations. Multiple standard conversations can all enter `arun()` concurrently regardless of the configured value. The Agent Server therefore has no effective global admission/concurrency limit for its normal execution path.

This is distinct from #3143 / #3169, which isolated synchronous conversation execution in a dedicated thread pool but did not cover the native async path added later.

## Impact

A burst of conversations can exhaust memory and cause the operating system to kill the Agent Server. This is particularly easy to trigger when an automation dispatcher or other client launches many conversations together.

Illustrative measurements from one 8 GiB self-hosted Agent Server using the current SDK:

| Restored conversation | Events | Persisted data | Incremental RSS |
|---|---:|---:|---:|
| Small/fresh | 6 | 26 KiB | ~0.7 MiB |
| Median sample | 120 | 1.3 MiB | ~7.3 MiB |
| Large sample | 2,216 | 158 MiB | ~602 MiB |

The same server loaded 83 persisted conversations and used approximately 2.21 GiB RSS before a burst. Conversation memory varies substantially because restoring state rebuilds the view and caches parsed event objects. Active LLM, tool, plugin, and MCP state can add further transient memory.

These figures are illustrative rather than a formal benchmark, but they show why a real execution cap is needed.

## Reproduction

1. Start Agent Server with `OH_MAX_CONCURRENT_RUNS=1`.
2. Create two or more ordinary conversations using an agent that implements native `astep()`.
3. Start all conversations concurrently.
4. Instrument the agent's `astep()` method with a shared active-counter.
5. Observe that the counter exceeds 1 because each conversation runs through `arun()` without acquiring the configured executor limit.

A regression test can implement this with a blocking async test agent and assert that the maximum observed active count never exceeds the configured limit.

## Expected behavior

`max_concurrent_runs` should bound concurrent agent execution across both paths:

- native async `conversation.arun()`
- synchronous `conversation.run()` fallback

A shared async admission mechanism, such as an `asyncio.Semaphore`, could guard the complete run operation. The design should also ensure:

- permits are released on exceptions, cancellation, pause, and shutdown;
- queued conversations do not misleadingly appear to be actively executing;
- queueing is bounded or excess work receives explicit backpressure;
- the configured limit has focused tests for both native async and sync agents.

## Related work

- #3143 / #3169: dedicated thread pool and configurable synchronous run worker count
- #3153: Agent Server performance, resource usage, and backpressure tracking
- #3140: eager loading of persisted conversations at startup

Shailendra005:
HUMAN:

Picked this up because a limit that only covers the fallback path isn't really a limit. Kept the change to the admission mechanism itself and left the queueing/status questions for you to weigh in on.

---

AGENT:

## Why

`max_concurrent_runs` is documented as the cap on conversations executing agent steps concurrently, but it only sizes the `ThreadPoolExecutor` behind the synchronous `conversation.run()` fallback. `EventService.run()` prefers native `conversation.arun()` whenever the agent overrides `astep()`, and that branch takes no permit, so the path normal conversations actually use is unbounded. Setting `OH_MAX_CONCURRENT_RUNS=N` has no effect there, which is what #4063 reports along with the memory figures behind it.

## Summary

- Added a server-wide `asyncio.Semaphore(max_concurrent_runs)` created next to the existing thread pool in `ConversationService.__aenter__` and injected into each `EventService` alongside `_run_executor`.
- `_run_and_publish` now holds a permit across **both** branches, so one shared pool bounds the total rather than the two paths each having their own ceiling.
- Used `async with` so the permit is returned on exception and on cancellation; a leaked permit would be permanent and would throttle the server toward zero over time.
- `_run_semaphore` is `None` for a standalone `EventService`, which keeps current unlimited behaviour for embedders and existing tests.
- Cleared in the shutdown path beside the executor.

```python
async with self._run_semaphore or nullcontext():
    if has_native_arun:
        await conversation.arun()
    else:
        await loop.run_in_executor(self._run_executor, conversation.run)
```

## Issue Number

Closes #4063

## How to Test

`tests/agent_server/test_event_service.py::TestRunAdmissionLimit` implements the shared active-counter check described in the issue: three `EventService` instances share one `Semaphore(1)`, each is driven through the real `await service.run()` with an agent whose `arun()` blocks, and the test asserts the peak observed concurrency never exceeds the limit. A second test asserts the permit is released when a run raises.

```
$ OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/agent_server/test_event_service.py -k RunAdmission -q
2 passed
```

Reverting just the `async with` line and re-running gives the reported behaviour:

```
AssertionError: 3 conversations ran concurrently under a limit of 1
```

Full suite and checks:

```
1918 passed, 13 deselected

$ uv run ruff format --check <changed>   # 3 files already formatted
$ uv run pyright openhands-agent-server/openhands/agent_server/event_service.py
```

Two notes for anyone extending these tests: `has_native_arun` inspects `type(conversation).arun` and `type(conversation.agent).astep`, so the methods have to be defined on mock **subclasses** — instance attributes silently fail the check and route the run down the sync branch instead. And `_run_task` is cleared when a run finishes, so task handles need capturing before the blocked runs are released.

## Scope

This covers the first and last bullets of the expected behaviour: a shared admission mechanism across both paths, with permits released on exception and cancellation. Two of your bullets are deliberately not in this PR, since they look like design calls rather than mechanics:

- **Queued conversations still report as running.** Making a waiting conversation surface as queued rather than executing changes user-visible status semantics, so I would rather agree the intended states with you first.
- **No bounded queue or explicit backpressure.** Today excess work waits on the semaphore indefinitely. A bounded queue that rejects or defers beyond a threshold is a policy decision, and it interacts with #3153.

Happy to take either as a follow-up, or to fold one in here if you would prefer it in a single change.

I also did not add a separate sync-agent test: that path was already bounded by the executor, and it now shares the same permit pool, which the exception test exercises. Glad to add an explicit one if you want the coverage stated.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -1974,7 +1974,9 @@ class ConversationService:
         if self._run_executor is not None:
             self._run_executor.shutdown(wait=False)
             self._run_executor = None
-        self._run_semaphore = None
+        # The semaphore is left in place on purpose: it owns no resources, and
+        # clearing it here would hand a service started by a partial-failure
+        # retry `_run_semaphore=None`, permanently removing its concurrency cap.
         if failures:
             assert self._event_services is not None
             for event_service in self._event_services.values():
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -599,6 +599,7 @@ class ConversationService:
     _lease_renewal_task: asyncio.Task | None = field(default=None, init=False)
     _eviction_task: asyncio.Task | None = field(default=None, init=False)
     _run_executor: ThreadPoolExecutor | None = field(default=None, init=False)
+    _run_semaphore: asyncio.Semaphore | None = field(default=None, init=False)
     _credential_bindings: dict[UUID, dict[str, VersionedCredentialBinding]] = field(
         default_factory=dict, init=False
     )
@@ -1789,6 +1790,9 @@ class ConversationService:
             max_workers=self.max_concurrent_runs,
             thread_name_prefix="conversation-run",
         )
+        # Bounds the native async path as well; the thread pool alone only
+        # limits the synchronous fallback.
+        self._run_semaphore = asyncio.Semaphore(self.max_concurrent_runs)
         self._event_services = {}
         self._conversation_records = await asyncio.to_thread(self._load_catalog_sync)
 
@@ -1970,6 +1974,7 @@ class ConversationService:
         if self._run_executor is not None:
             self._run_executor.shutdown(wait=False)
             self._run_executor = None
+        self._run_semaphore = None
         if failures:
             assert self._event_services is not None
             for event_service in self._event_services.values():
@@ -2035,6 +2040,7 @@ class ConversationService:
         # _renew_all_leases_loop task on ConversationService.
         event_service._external_lease_renewal = True
         event_service._run_executor = self._run_executor
+        event_service._run_semaphore = self._run_semaphore
 
         try:
             await event_service.start()
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -1199,7 +1199,14 @@ class EventService:
                     # Hold the admission permit across the whole run so both
                     # paths count against max_concurrent_runs. `async with`
                     # releases it on exception and on cancellation.
+                    queued_generation = self._explicit_interrupt_generation
                     async with self._run_semaphore or nullcontext():
+                        # A pause()/interrupt() that lands while this run is
+                        # still queued has nothing to cancel, so honour it here
+                        # rather than starting the run and overwriting the
+                        # requested state once the permit arrives.
+                        if self._explicit_interrupt_generation != queued_generation:
+                            return
                         if has_native_arun:
                             await conversation.arun()
                         else:
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -146,6 +146,10 @@ class EventService:
     _lease_task: asyncio.Task | None = field(default=None, init=False)
     _external_lease_renewal: bool = field(default=False, init=False)
     _run_executor: ThreadPoolExecutor | None = field(default=None, init=False)
+    # Server-wide admission limit shared by every EventService so
+    # max_concurrent_runs also bounds the native async path, not just the
+    # thread pool. None when a service runs standalone (no global limit).
+    _run_semaphore: asyncio.Semaphore | None = field(default=None, init=False)
     # Background task for a /goal loop that is running inside this conversation.
     _goal_loop_task: asyncio.Task | None = field(default=None, init=False)
     _goal_loop_outcome: GoalOutcome | None = field(default=None, init=False)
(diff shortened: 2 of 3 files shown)


openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -1192,10 +1196,16 @@ async def _run_and_publish():
...
                         and type(conversation.agent).astep is not AgentBase.astep
                     )
-                    if has_native_arun:
-                        await conversation.arun()
-                    else:
-                        await loop.run_in_executor(self._run_executor, conversation.run)
+                    # Hold the admission permit across the whole run so both
+                    # paths count against max_concurrent_runs. `async with`
+                    # releases it on exception and on cancellation.
+                    async with self._run_semaphore or nullcontext():
```
VascoSch92, line 1203: a run waiting for the new admission permit has no _arun_task, so interrupt()/pause() no-ops and gets overwritten to RUNNING once the queued run starts.

openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -1970,6 +1974,7 @@ async def __aexit__(self, exc_type, exc_value, traceback):
         if self._run_executor is not None:
             self._run_executor.shutdown(wait=False)
             self._run_executor = None
+        self._run_semaphore = None
```
VascoSch92: partial-failure teardown can hand a freshly-started EventService _run_semaphore=None, permanently disabling its concurrency cap.

tests/agent_server/test_event_service.py
```diff
@@ -3423,3 +3423,107 @@ async def test_event_service_creates_lease_with_custom_ttl(tmp_path: Path) -> No
     assert service._lease is not None
     assert service._lease._ttl_seconds == 10.0
     assert (tmp_path / stored.id.hex / LEASE_FILE_NAME).exists()
+
+
+class TestRunAdmissionLimit:
```
VascoSch92, line 3428: only the native-async path is exercised, leaving the PR's second stated fix untested.

Shailendra005:
@VascoSch92 all three addressed in 284bb712.

**Pause/interrupt while queued.** You were right, and it was a real hole: a queued run has nothing to cancel, so the request was dropped and the run started anyway once the permit arrived. The run now captures `_explicit_interrupt_generation` before waiting and bails out if it changed while queued, matching the generation checks already used in `run()` and the rerun path. New test `test_pause_while_queued_is_not_overwritten` fails without it with `a run paused while queued still executed`.

**Teardown handing out `_run_semaphore=None`.** Also right. Unlike the executor the semaphore owns no resources, so there is nothing to clean up and clearing it only creates that window. I removed the assignment and left a comment saying why, so it does not get "tidied" back in.

**Sync path untested.** Added `test_sync_fallback_runs_share_the_same_limit`, which routes through `run_in_executor` by setting `arun = None` and asserts the same peak-concurrency bound. It fails without the change with `3 sync runs executed concurrently under a limit of 1`.

All three new tests fail on `main` and pass here. Full suite 1920 passed, ruff and pyright clean.

One thing I did not touch: bounded queueing / explicit backpressure from the issue's expected-behaviour list is still out of scope here, since it looks like a policy decision that interacts with #3153. Happy to take it as a follow-up.

Issue #4077, opened 2026-07-10
Streaming: correctness & resource-safety bugs in the token/delta pipeline


VascoSch92:
## Summary

An audit of the token/event streaming pipeline (`llm.py` → `on_token`/`on_event` → `PubSub` → WebSocket) surfaced several correctness and resource-safety bugs. The **durable record is correct** — the final `MessageEvent` is rebuilt server-side via `litellm.stream_chunk_builder`, so persisted state never ends up wrong. All the issues below are in the **live/transient** streaming layer. Line numbers reference the current `main`.

The most impactful is OpenHands/agent-canvas#1 (a plain, unconditional code path — not timing-dependent).

---

### 1. HIGH — Retry re-fires `on_token` from scratch, duplicating every streamed token on the client

Tokens are emitted *inside* the retry boundary, so any retry re-streams and re-emits all tokens.

- `openhands-sdk/openhands/sdk/llm/llm.py:2117-2120` — `on_token(chunk)` is called inside the stream loop of `_transport_call`.
- `openhands-sdk/openhands/sdk/llm/llm.py:1429-1439` — `_transport_call` is invoked from `_one_attempt`, which is wrapped by `@self._make_retry_decorator()`.
- `openhands-sdk/openhands/sdk/llm/llm.py:1351-1354` — `_validate_chat_response` raises `LLMNoResponseError` **after** the full stream has already fired.
- `openhands-sdk/openhands/sdk/llm/llm.py:131-138` — `LLMNoResponseError` (plus `APIConnectionError`, `RateLimitError`, `LiteLLMTimeout`, etc.) is in `LLM_RETRY_EXCEPTIONS`.

**Failure:** model streams "The answer is 4"; provider drops the connection (retryable) or `_validate_chat_response` raises `LLMNoResponseError`. Tenacity re-runs `_one_attempt`, re-opens the stream, and calls `on_token` for every chunk again → the client renders the text 2× (N× for N retries). Same defect in `responses()`/`aresponses()`.

**Fix idea:** emit through a per-attempt buffering wrapper that only forwards *new* deltas, move `on_token` outside the retry boundary, or send the callback a "reset/restart" signal on each new attempt.

---

### 2. MEDIUM — Empty stream reassembles to `None` → un-retried `AssertionError`, diverging from the non-streaming path

[...]

PR #4392, opened 2026-08-06
fix(agent-server): surface _emit_event_from_thread executor failures

### LINKED ISSUE #4386: agent-server: _emit_event_from_thread silently drops exceptions — stats and LLM-log events lost without any signal
## Summary

`EventService._emit_event_from_thread` (event_service.py line 842) calls `run_in_executor` and **discards the returned `asyncio.Future`**. If the executor task raises (disk full, broken state, serialisation error), the exception is silently swallowed — the only signal is an eventual `"Future exception was never retrieved"` runtime warning emitted at GC time, often rate-limited or absent in production logs.

## Affected code

**`openhands-agent-server/openhands/agent_server/event_service.py` line 842**

```python
def _emit_event_from_thread(self, event: Event) -> None:
    ...
    if main_loop and main_loop.is_running() and conversation:
        def locked_on_event():
            with conversation._state:
                conversation._on_event(event)

        main_loop.run_in_executor(None, locked_on_event)  # ← Future discarded
```

Called from two non-async callbacks:

- `_setup_llm_log_streaming` → `log_callback` (every LLM completion log event)
- `_setup_stats_streaming` → `stats_callback` (every per-turn token-usage update)

## Failure scenario

1. Workspace disk fills up (or `_on_event` fails for any reason) while an LLM call is in progress.
2. `locked_on_event` raises inside the executor thread.
3. The `asyncio.Future` returned by `run_in_executor` is garbage-collected with an unread exception.
4. Stats updates and LLM completion log events are silently dropped with no error propagation and no ERROR-level log entry.

## Relationship to #4077

Issue #4077 item 3 identifies the same pattern in `_publish_stream_delta` (line 804–805: `asyncio.run_coroutine_threadsafe` future discarded). This is the parallel defect on the non-streaming side — a different function, different callers, same silent-drop consequence.

## Proposed fix

Add a done-callback so exceptions surface at ERROR level:

```python
def _emit_event_from_thread(self, event: Event) -> None:
    main_loop = self._main_loop
    conversation = self._conversation
    if main_loop and main_loop.is_running() and conversation:
        def locked_on_event():
            with conversation._state:
                conversation._on_event(event)

        fut = main_loop.run_in_executor(None, locked_on_event)

        def _on_done(f: asyncio.Future) -> None:
            if not f.cancelled() and f.exception() is not None:
                logger.error(
                    "emit_event_from_thread failed (%s): %s",
                    type(event).__name__,
                    f.exception(),
                )

        fut.add_done_callback(_on_done)
```

Shailendra005:
HUMAN:

Grabbed this one because a dropped stats event leaving no trace is the kind of thing you only notice long after it mattered. Small change, but the test needed a real assertion rather than "no exception raised".

---

AGENT:

## Why

`_emit_event_from_thread` schedules `locked_on_event` with `run_in_executor` and discards the returned future. If `_on_event` raises inside the executor thread — a full disk, a serialisation error, broken state — nothing observes the failure. The event is lost and the only trace is Python's GC-time `Future exception was never retrieved` warning, which is easy to miss or lose entirely in production logging.

Both callers are non-async callbacks on hot paths: `_setup_llm_log_streaming` → `log_callback` (every LLM completion log) and `_setup_stats_streaming` → `stats_callback` (every per-turn token-usage update). Reported in #4386.

## Summary

- Attach a done-callback to the future so an executor failure is logged at ERROR with the event kind and traceback, instead of being collected unread.
- Added `_log_emit_failure`, a module-level helper that ignores cancellation and only reports a real exception.
- Adjusted one existing test's stub: `test_emit_event_from_thread_uses_captured_loop` had `run_in_executor` return `None`, which the real API never does. It now returns a mock future. I preferred fixing the stub over adding a `None` guard in production, since a guard would imply a return value that cannot occur.

## Issue Number

Closes #4386

## How to Test

`tests/agent_server/test_event_service.py::TestEmitEventFromThreadFailures::test_emit_failure_is_logged` drives the real method with a conversation whose `_on_event` raises, then asserts the failure reaches the log.

```
$ OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/agent_server/test_event_service.py -k EmitEventFromThreadFailures -q
1 passed
```

Reverting just the done-callback makes it fail, and the captured output is the reported symptom:

```
AssertionError: emission failure was not surfaced; captured=[
  "Future exception was never retrieved\nfuture: <Future finished exception=RuntimeError('disk full')>"
]
```

Worth noting for review: my first version of this test asserted loosely and passed *with the fix reverted*, so it proved nothing. The assertion now matches the specific message.

Full suite and checks:

```
1917 passed, 13 deselected

$ uv run ruff format --check <changed>   # unchanged
$ uv run pyright openhands-agent-server/openhands/agent_server/event_service.py
```

## Scope

Logging only — the failure is surfaced, not retried, and the event is still dropped. Making these emissions durable is a larger change and belongs with the buffering discussion rather than here.

The issue notes #4077 item 3 describes the same discarded-future pattern in `_publish_stream_delta` (`asyncio.run_coroutine_threadsafe`). I left it alone to keep this reviewable; happy to follow up there with the same treatment if you want them consistent.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -5,6 +5,7 @@ from concurrent.futures import ThreadPoolExecutor
 from contextlib import nullcontext, suppress
 from dataclasses import dataclass, field
 from datetime import datetime
+from functools import partial
 from pathlib import Path
 from typing import cast
 from uuid import UUID, uuid4
@@ -90,6 +91,17 @@ logger = get_logger(__name__)
 
 class CredentialBindingActivationTooLate(RuntimeError):
     pass
+
+
+def _log_emit_failure(event_kind: str, future: "asyncio.Future[None]") -> None:
+    """Log an event emission that failed inside the executor thread."""
+    if future.cancelled():
+        return
+    error = future.exception()
+    if error is not None:
+        logger.error(
+            "Failed to emit %s from thread: %s", event_kind, error, exc_info=error
+        )
 
 
 def _without_agent_context_secret(
@@ -838,8 +850,12 @@ class EventService:
                     conversation._on_event(event)
 
             # Run the locked callback in an executor to ensure the event is
-            # both persisted and sent to WebSocket subscribers
-            main_loop.run_in_executor(None, locked_on_event)
+            # both persisted and sent to WebSocket subscribers. The future
+            # needs a done-callback: otherwise it is collected with an unread
+            # exception, so a dropped stats or LLM-log event leaves no signal
+            # beyond a GC-time warning.
+            future = main_loop.run_in_executor(None, locked_on_event)
+            future.add_done_callback(partial(_log_emit_failure, type(event).__name__))
 
     def _setup_llm_log_streaming(self, agent: AgentBase) -> None:
         """Configure LLM log callbacks to stream logs via events."""
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -855,7 +855,15 @@ class EventService:
             # exception, so a dropped stats or LLM-log event leaves no signal
             # beyond a GC-time warning.
             future = main_loop.run_in_executor(None, locked_on_event)
-            future.add_done_callback(partial(_log_emit_failure, type(event).__name__))
+            # Attach from the loop thread. This helper is called from foreign
+            # threads, and `add_done_callback` on an already-done future uses
+            # plain `call_soon`, which neither is thread-safe nor wakes an idle
+            # loop -- the report could then sit unfired, which is the failure
+            # this callback exists to prevent.
+            main_loop.call_soon_threadsafe(
+                future.add_done_callback,
+                partial(_log_emit_failure, type(event).__name__),
+            )
 
     def _setup_llm_log_streaming(self, agent: AgentBase) -> None:
         """Configure LLM log callbacks to stream logs via events."""
--- a/tests/agent_server/test_event_service.py
+++ b/tests/agent_server/test_event_service.py
@@ -16,7 +16,10 @@ import pytest_asyncio
 
 from openhands.agent_server.conversation_lease import LEASE_FILE_NAME
 from openhands.agent_server.conversation_service import ConversationService
-from openhands.agent_server.event_service import EventService
+from openhands.agent_server.event_service import (
+    EventService,
+    _log_emit_failure,
+)
 from openhands.agent_server.models import (
     ConfirmationResponseRequest,
     EventPage,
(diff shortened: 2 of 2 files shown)


openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -838,8 +850,12 @@ def locked_on_event():
...
 
             # Run the locked callback in an executor to ensure the event is
-            # both persisted and sent to WebSocket subscribers
-            main_loop.run_in_executor(None, locked_on_event)
+            # both persisted and sent to WebSocket subscribers. The future
+            # needs a done-callback: otherwise it is collected with an unread
+            # exception, so a dropped stats or LLM-log event leaves no signal
+            # beyond a GC-time warning.
+            future = main_loop.run_in_executor(None, locked_on_event)
+            future.add_done_callback(partial(_log_emit_failure, type(event).__name__))
```
VascoSch92: `add_done_callback` is invoked from a foreign executor thread rather than the event loop's own thread. For an already-done future this schedules via plain `call_soon` (not thread-safe) instead of [call_soon_threadsafe](url), so the callback can sit unfired in the loop's ready queue indefinitely on an idle loop — potentially reintroducing the exact silent-failure bug (#4386) this PR set out to fix.

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -92,6 +93,17 @@ class CredentialBindingActivationTooLate(RuntimeError):
     pass
 
 
+def _log_emit_failure(event_kind: str, future: "asyncio.Future[None]") -> None:
+    """Log an event emission that failed inside the executor thread."""
+    if future.cancelled():
```
VascoSch92, line 98: the `future.cancelled()` early-return branch in `_log_emit_failure` has no test coverage, so a future reordering of the checks could leak an unhandled `CancelledError` from the callback undetected.

VascoSch92 (approved):

Shailendra005:
@VascoSch92 thanks — both addressed in 4f9e02b.

**Foreign-thread `add_done_callback`.** Good catch, and confirmed in CPython: `Future.add_done_callback` takes the `self._loop.call_soon(fn, self)` branch whenever the future is already settled, so attaching from an executor thread neither is thread-safe nor wakes an idle loop. Since this helper exists precisely to be called off-loop, a fast-failing emission could have left the report queued indefinitely — exactly the silence this PR is meant to remove. The attach now goes through `call_soon_threadsafe`, so it happens on the loop thread and the loop is woken.

`test_callback_is_attached_from_the_loop_thread` pins it: with the previous direct attach it fails with `done-callback must be attached via call_soon_threadsafe`.

**Uncovered `cancelled()` branch.** Added `test_cancelled_future_is_not_reported`, which cancels a real future and asserts the helper neither logs nor raises, so reordering the checks would surface as a failure rather than a leaked `CancelledError`.

Full suite 1919 passed, ruff and pyright clean.

Shailendra005:
@VascoSch92 I just wanted to follow up and ask if the PR is still LGTM after the changes being made. If it is the case then can you approve it! Thanks.

PR #4396, opened 2026-08-06
fix(agent-server): size bash event pages by returned events, not files read

### LINKED ISSUE #4388: agent-server: search_bash_events pagination wrong when order__gt filter is active — pages undersized, clients loop on empty next-pages
## Summary

`BashEventService.search_bash_events` selects `page_files` (up to `limit` file entries) **before** applying the `order__gt` filter. The filter runs as a post-selection step, returning fewer than `limit` events while `next_page_id` is still set. Clients following the cursor receive an unbounded sequence of under-full or entirely empty pages.

## Affected code

**`openhands-agent-server/openhands/agent_server/bash_service.py`**

```python
# Pagination boundary computed on UNFILTERED file count:
for i in range(start_index, len(files)):
    if len(page_files) >= limit:
        next_page_id = str(files[i].name)   # set when limit FILES collected
        break
    page_files.append(files[i])

# order__gt filter applied AFTER the boundary:
for file_path in page_files:
    event = self._load_event_from_file(file_path)
    if event is not None:
        if order__gt is not None:
            event_order = getattr(event, "order", None)
            if event_order is not None and event_order <= order__gt:
                continue   # ← filtered after the page was already bounded
        page_events.append(event)

return BashEventPage(items=page_events, next_page_id=next_page_id)
```

## Failure scenario

- `limit = 100`, `order__gt = 50`
- Files on disk: 100 `BashOutput` events all with `order ∈ [0, 50]`, followed by events with `order > 50`
- Result: `page_events` is **empty**, but `next_page_id` is set to file 101
- Client follows `next_page_id`, gets another empty page, and loops indefinitely

Less extreme: `limit = 100`, 90 of the 100 selected files have `order <= order__gt` → client gets 10 events but must follow multiple `next_page_id` hops to retrieve items that should have fit in one page.

`order__gt` is used for incremental command-output polling (stream output after the Nth chunk), so this is a live code path.

## Proposed fix

Move filtering inside the page-collection loop so the boundary is computed on the filtered event count:

```python
page_events: list[BashEventBase] = []
next_page_id: str | None = None

for i in range(start_index, len(files)):
    if len(page_events) >= limit:
        next_page_id = str(files[i].name)
        break
    event = self._load_event_from_file(files[i])
    if event is None:
        continue
    if order__gt is not None:
        event_order = getattr(event, "order", None)
        if event_order is not None and event_order <= order__gt:
            continue
    page_events.append(event)

return BashEventPage(items=page_events, next_page_id=next_page_id)
```

This guarantees pages always contain exactly `limit` filtered events (or fewer only on the final page), regardless of how many files the `order__gt` filter discards.

Shailendra005:
HUMAN:

Took this one because a cursor that keeps pointing at nothing is worse than an error — the client can't tell the difference between "slow" and "broken".

---

AGENT:

## Why

`search_bash_events` picks up to `limit` **files**, then applies `order__gt` while loading them. The page boundary is therefore measured before the filter runs, so a page can come back short — or completely empty — while `next_page_id` is still set. A client following the cursor walks a run of under-full pages, and in the all-filtered case never makes progress.

`order__gt` is how incremental command-output polling works ("give me output after chunk N"), so this is a live path rather than a corner case. Reported in #4388.

## Summary

- Merged the filter into the page-collection loop so the boundary counts events that will actually be returned.
- An unreadable file (`_load_event_from_file` returning `None`) no longer consumes a page slot either — previously it silently shrank the page the same way.
- `next_page_id` semantics are unchanged: it points at the first file not consumed by this page.

## Issue Number

Closes #4388

## How to Test

Two tests in `tests/agent_server/test_bash_service.py`:

- `test_order_filter_does_not_undersize_pages` — 15 outputs, `order__gt=9`, `limit=5`. Expects a full page of orders 10–14.
- `test_order_filter_never_returns_an_empty_page_with_a_cursor` — every event filtered out. Expects no items **and** no cursor.

```
$ OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/agent_server/test_bash_service.py -k order_filter -q
2 passed
```

Both fail on `main` with the reported behaviour:

```
assert [] == [10, 11, 12, 13, 14]
AssertionError: an empty page must not advertise a next page, or clients loop forever
  assert '20200101000005000000_BashOutput_...' is None
```

The tests give each event an increasing timestamp, since files sort by name and same-timestamp events would otherwise come back in a nondeterministic order.

Full suite and checks:

```
1918 passed, 13 deselected

$ uv run ruff format --check <changed>   # unchanged
$ uv run pyright openhands-agent-server/openhands/agent_server/bash_service.py
```

## Note

A full page still carries a `next_page_id` even when every remaining file would be filtered out — the service cannot know that without reading them. The following request then returns an empty page with no cursor, so the client terminates after one extra round trip rather than looping. Making that last hop disappear would mean reading ahead past the page boundary, which seemed a worse trade than one predictable extra call; happy to revisit if you would rather pay that cost.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/bash_service.py
+++ b/openhands-agent-server/openhands/agent_server/bash_service.py
@@ -149,7 +149,6 @@ class BashEventService:
             files = [file for file in files if file.name < timestamp_lt_str]
 
         # Handle pagination
-        page_files = []
         start_index = 0
 
         # Find the starting point if page_id is provided
@@ -159,27 +158,29 @@ class BashEventService:
                     start_index = i
                     break
 
-        # Collect items for this page
+        # Collect items for this page. Filtering happens inside this loop so the
+        # boundary is measured in returned events rather than files read: sizing
+        # the page by file count first lets `order__gt` empty a page while
+        # `next_page_id` is still set, walking the client through under-full or
+        # entirely empty pages.
+        page_events = []
         next_page_id = None
         for i in range(start_index, len(files)):
-            if len(page_files) >= limit:
+            if len(page_events) >= limit:
                 # We have collected enough items for this page
                 # Set next_page_id to the current file for next page
                 next_page_id = str(files[i].name)
                 break
-            page_files.append(files[i])
 
-        # Load only the page files (not all files)
-        page_events = []
-        for file_path in page_files:
-            event = self._load_event_from_file(file_path)
-            if event is not None:
-                # Filter by order if specified (only applies to BashOutput events)
-                if order__gt is not None:
-                    event_order = getattr(event, "order", None)
-                    if event_order is not None and event_order <= order__gt:
-                        continue
-                page_events.append(event)
+            event = self._load_event_from_file(files[i])
+            if event is None:
+                continue
+            # Filter by order if specified (only applies to BashOutput events)
+            if order__gt is not None:
+                event_order = getattr(event, "order", None)
+                if event_order is not None and event_order <= order__gt:
+                    continue
+            page_events.append(event)
 
         return BashEventPage(items=page_events, next_page_id=next_page_id)
 
--- a/tests/agent_server/test_bash_service.py
+++ b/tests/agent_server/test_bash_service.py
@@ -202,7 +202,7 @@ async def test_run_retention_cleanup_loop_purges_old_events(tmp_path: Path):
 # ---------------------------------------------------------------------------
 
 
-def test_order_filter_does_not_undersize_pages(tmp_path: Path):
+async def test_order_filter_does_not_undersize_pages(tmp_path: Path):
     """The page boundary must count returned events, not files read (#4388).
 
     Filtering after the page was already sized by file count let `order__gt`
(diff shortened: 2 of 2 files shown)


neubig:
Minor consistency note (non-blocking for me): the two new tests use `asyncio.run(service.search_bash_events(...))` while the rest of the file uses `async def test_...` + `await` under `asyncio_mode = "auto"` (see `test_run_retention_cleanup_loop_purges_old_events` just above). Converting them to the same `async def` style would keep the async-usage consistent across the file. Happy either way — flagging since inconsistency in async handling is the kind of thing that tends to spread.

Shailendra005:
@neubig good call — converted in the latest commit. Both tests are now `async def` + `await` like the rest of the file, so the `asyncio.run(...)` outlier is gone.

Their pyright hook then flagged a second thing worth keeping: `page.items` is typed as `list[BashEventBase]`, which has no `order`, so the assertion now narrows with `isinstance(e, BashOutput)` before reading it.

PR #4401, opened 2026-08-06
fix: stop CustomSecretsSection from recommending secrets in git remote URLs

### LINKED ISSUE #4399: [Bug]: CustomSecretsSection prompt example writes secrets into .git/config in plaintext

### Bug Description

`CustomSecretsSection` (the `<CUSTOM_SECRETS>` block injected into the system prompt whenever a conversation has any registered secret) gives the model a worked example that writes the raw secret value into a git remote URL:

https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/context/prompts/sections/dynamic.py#L154

```
* Refreshing expired secrets: Some secrets (like GITHUB_TOKEN) may be updated periodically or expire over time. If a secret stops working (e.g., authentication failures), try using it again in a new command - the system should automatically use the refreshed value. For example, if GITHUB_TOKEN was used in a git remote URL and later expired, you can update the remote URL with the current token: `git remote set-url origin https://${GITHUB_TOKEN}@github.com/username/repo.git` to pick up the refreshed token value.
```

Following this example, `git remote set-url` writes the interpolated secret straight into `.git/config` in plaintext. The same section separately claims:

```
* Security: Secret values are automatically masked in command output to prevent accidental exposure. You will see `<secret-hidden>` instead of the actual secret value in the output.
```

That masking is real for the terminal transcript, but it does not extend to files a command writes to disk. `.git/config` isn't part of "command output" — it's a persisted file, so the raw credential lands there unmasked and stays there for as long as the workspace exists, which for most long-running agent deployments (automations, schedulers, anything that doesn't tear down its working directory per run) is indefinitely.

Introduced in 549a5b8 (#4030), still present on `main` as of commit `443a462` (verified today).

### Expected Behavior

The worked example should authenticate without ever writing the secret to a file on disk — e.g. `gh auth setup-git`, a short-lived git credential helper, or an `Authorization` header via `http.extraheader`, none of which persist the raw token into `.git/config`.

### Actual Behavior

Any conversation that has `GITHUB_TOKEN` (or an equivalently-named secret) registered and follows this exact guidance ends up with the live token sitting in a world-readable `.git/config`.

This isn't theoretical — we operate several long-running agent automations built on this SDK (issue triage, PR review, a Slack responder, etc.), each registering `GITHUB_TOKEN` as a secret. Auditing their workspaces, we found the *same* live token embedded in plaintext in `.git/config` in multiple, unrelated checkouts, written by different conversations/automations over about a week, all using this exact `git remote set-url origin https://${TOKEN}@github.com/...` pattern. Workspace directories for these automations are not pruned, so each instance persisted indefinitely until we happened to find them during an unrelated investigation.

### Steps to Reproduce

1. Register a secret (e.g. `GITHUB_TOKEN`) on a conversation.
2. Give the agent a task that needs to push to a remote requiring auth.
3. The agent follows the `<CUSTOM_SECRETS>` guidance verbatim (or arrives at the same pattern independently, since it's presented as the recommended approach) and runs `git remote set-url origin https://${GITHUB_TOKEN}@github.com/...`.
4. `cat .git/config` in that workspace now shows the raw secret value in plaintext.
5. Nothing in the SDK detects, redacts, or warns about this — the file just sits there for the lifetime of the workspace.

Note: `openhands/sdk/git/utils.py` already has `redact_url_credentials` / `redact_url_credentials_in_text`, used to keep the SDK's own internal git probing (e.g. `get_git_repository_metadata`) from leaking credentials into its logs. That protection doesn't apply to commands the agent runs directly via the terminal tool, and doesn't touch `.git/config` itself.

### Installation Method

N/A (source review against `OpenHands/software-agent-sdk@main`)

### SDK Version

main branch, commit `443a462309dff081680eb5a2d699d0d49d661dd6`

### Additional Context

Suggested direction (not prescriptive — happy to send a PR if useful): rewrite the "Refreshing expired secrets" example to a method that never persists the secret to a file, e.g. `gh auth setup-git` (this is already how our own issue-resolver automation's prompt tells the agent to authenticate, as a workaround at the call-site level rather than in the SDK). As defense in depth, it may also be worth having the SDK scan for and redact known-secret substrings before any file write, not just command output, though fixing the misleading example is the more direct fix.

VascoSch92:
HUMAN:

---

AGENT:

## Why

The `<CUSTOM_SECRETS>` prompt block (`CustomSecretsSection`) gave agents `git remote set-url origin https://${GITHUB_TOKEN}@github.com/...` as the worked example for using a secret with git. That writes the raw secret into `.git/config` in plaintext, where it persists on disk for the life of the workspace. The block's own "masked in command output" guarantee doesn't cover this, since `.git/config` is a file the command writes, not output it prints.

We found this cause real, live-token leaks across several of our own production automations built on this SDK — see #4399 for the full writeup and evidence.

## Summary
- Replace the git remote URL example in the "Refreshing expired secrets" bullet with `gh auth setup-git`, which configures a credential helper without ever writing the secret to disk.
- Drop "or git URLs" from the "Avoid exposing raw secrets" bullet's list of acceptable secret uses, since it directly contradicted the fix above.
- Regenerate the one affected prompt snapshot (`dynamic_context__with_secret_registry.txt`) via `REGEN_PROMPT_SNAPSHOTS=1 uv run pytest tests/sdk/context/prompts/test_prompt_snapshot.py`.

## Issue Number

Fixes #4399

## How to Test

- `uv run pytest tests/sdk/context/prompts/test_prompt_snapshot.py -q` - 53 passed.
- `uv run pytest tests/sdk/context/ tests/cross/test_agent_secrets_integration.py tests/sdk/agent/test_acp_agent.py -q` - 877 passed, 1 failed (`test_atexit_cleanup_is_weak_and_inline`). Confirmed pre-existing and unrelated: passes in isolation on this branch, and passes on a clean `main` checkout when `test_acp_agent.py` is run standalone (432 passed) - it's a cross-test-file state leak that only shows up when combined with the other files in one process, not something this change touches.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Minimal, text-only fix scoped to the misleading example - no change to the secret injection/masking mechanism itself. Filed as a followup in #4399 whether defense-in-depth (e.g. scanning file writes for known secret substrings, not just command output) is worth doing separately.


openhands-sdk/openhands/sdk/context/prompts/sections/dynamic.py
```diff
@@ -150,8 +150,8 @@ def render(self, ctx: PromptContext) -> str | None:
...
+            "* Avoid exposing raw secrets: Never echo or print the full value of secrets (e.g., avoid `echo $SECRET`). The conversation history may be logged or shared, and exposing raw secret values could compromise security. Instead, use secrets directly in commands where they serve their intended purpose (e.g., in curl headers).\n"
+            "* Refreshing expired secrets: Some secrets (like GITHUB_TOKEN) may be updated periodically or expire over time. If a secret stops working (e.g., authentication failures), try using it again in a new command - the system should automatically use the refreshed value. For git operations, never put a secret directly in a remote URL or command argument: it gets written to `.git/config` in plaintext and stays there even after the value is masked from your terminal output. Run `gh auth setup-git` once instead (with the token available as GITHUB_TOKEN/GH_TOKEN) to configure a credential helper, then use `git push`/`git pull`/`git fetch` normally.\n"
```
enyst, line 154:
```suggestion
            "* Refreshing expired secrets: Some secrets (like GITHUB_TOKEN) may be updated periodically or expire over time. If a secret stops working (e.g., authentication failures), try using it again in a new command - the system should automatically use the refreshed value. For git operations, never put a secret directly in a remote URL or command argument: it gets written to `.git/config` in plaintext and stays there even after the value is masked from your terminal output. If your `gh` is not authenticated, run `gh auth setup-git` once instead (with the token available as GITHUB_TOKEN/GH_TOKEN) to configure a credential helper, then use `git push`/`git pull`/`git fetch` normally.\n"
```

In addition, I think it can also use the token in front of `gh`? I seem to recall `gh` accepts override this way.

enyst (commented):
On a related note, are we sure we always have `gh` on the machine?

I recall we added it in docker at some point, but there’s no guarantee on local machines, is there?

I think maybe we could soften the recommendation a little… 🤔

VascoSch92:
@enyst we do have gh on the VM

PR #4408, opened 2026-08-07
fix(acp): resolve launch commands via PATHEXT-aware lookup so ACP can spawn on Windows

Telov:
HUMAN:

Had issues installing OpenHands on windows. Putting up some PRs to help the community

---

AGENT:

## Why

ACP can't start on Windows. `asyncio.create_subprocess_exec` bottoms out in `CreateProcess`,
which resolves a bare program name against `PATH` but only ever appends `.exe` — it never
consults `PATHEXT`. npm CLI entry points are `.cmd` shims, so both paths that produce a launch
command fail:

1. The registry default is `("npx", "-y", "<pkg>@<ver>")`. npm ships `npx.cmd`, so this raises
   `FileNotFoundError` / `[WinError 2]` even with npm correctly on `PATH`.
2. The pre-installed-binary path has the same defect for a different reason.
   `_prefer_pinned_binary` calls `shutil.which(info.binary_name)` — which *does* honour
   `PATHEXT` and finds the shim — then discards the resolved path and returns the bare name
   (`model.py:1766`).

So the workaround for (1) is (2), and (2) fails identically. There's no working configuration
through the provider presets on Windows.

## Summary

- `_prefer_pinned_binary` returns the path `shutil.which` resolved instead of the bare
  `binary_name`. Also removes a PATH-lookup race between resolution and spawn.
- New `_resolve_executable` helper applied to `command[0]` just before
  `create_subprocess_exec`, so the registry default and any user-supplied `acp_command` both
  get a `PATHEXT`-aware lookup.
- Tests for both, and updates to the existing `resolve_acp_command` assertions that expected
  the bare name.

No-op on POSIX: `shutil.which` returns the same file `execvp` would have found. A program name
with a directory component is passed through untouched; an unresolvable name is passed through
unchanged so a missing CLI still raises `ACPSpawnError` at the spawn rather than being
swallowed earlier. `detect_acp_provider_by_command` already reduces tokens to basenames and
documents `/opt/node_modules/.bin/codex-acp` as a supported form, so it is unaffected.

## How to Test

Unit:

```
uv run pytest tests/sdk/test_settings.py tests/sdk/agent/test_acp_agent.py -q
```

End-to-end, on Windows 10 19045 / Python 3.13. For each path this builds the command the way
`main` does and the way this branch does, spawns each through
`asyncio.create_subprocess_exec` exactly as `ACPAgent` does, and runs a real ACP `initialize`
handshake (pre-auth, so no credentials involved):

```
PATH A — pinned provider binary on PATH
  main   : ['claude-agent-acp']
  patched: ['C:\\Users\\<user>\\AppData\\Roaming\\npm\\claude-agent-acp.CMD']
    main     spawn=FAIL  FileNotFoundError: [WinError 2] The system cannot find the file specified
    patched  spawn=OK    handshake ok, agent=@agentclientprotocol/claude-agent-acp v0.65.0

PATH B — no pinned binary, registry `npx -y <pkg>` default
  main   : ['npx', '-y', '@agentclientprotocol/claude-agent-acp@0.63.0']
  patched: ['C:\\Program Files\\nodejs\\npx.CMD', '-y', '@agentclientprotocol/claude-agent-acp@0.63.0']
    main     spawn=FAIL  FileNotFoundError: [WinError 2] The system cannot find the file specified
    patched  spawn=OK    handshake ok, agent=@agentclientprotocol/claude-agent-acp v0.63.0
```

Script available on request; it imports `resolve_acp_command` / `_resolve_executable` from the
branch, so it exercises the shipped code path rather than a reimplementation.

## Video/Screenshots

n/a — terminal only, output above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **The MCP half of this stack already does this.** `mcp/os/win32/utilities.py::
  get_windows_executable_command` (in the `mcp` package this SDK already depends on) resolves a
  bare command via `shutil.which`, then tries `.cmd` / `.bat` / `.exe` / `.ps1`, then falls back
  to the bare name — the same shape as `_resolve_executable` here. That's why stdio MCP servers
  launch from `npx` on Windows while the ACP path doesn't: only the ACP spawn skips the lookup.
- **Known, pre-existing:** resolving to a `.cmd` means `CreateProcess` runs it through `cmd.exe`,
  so arguments containing shell metacharacters (`&`, `|`, `^`, `"`) are subject to cmd's parsing
  (the "BatBadBut" class). This PR doesn't introduce that — it's inherent to npm's `.cmd` shims
  and the `mcp` resolution above has the same property — but it does make `.cmd` execution the
  normal path on Windows rather than an error. The registry defaults are package specs with no
  metacharacters; a user-supplied `acp_args` is the exposure. Happy to add quoting hardening
  here or as a follow-up if you want it.
- `resolve_acp_command()` now returns an absolute path where it previously returned a bare name
  — `/usr/local/bin/codex-acp` rather than `codex-acp` in the agent-server image. Nothing
  in-tree compares against the bare form outside the tests updated here, but it lands in
  persisted settings, so flagging it.
- The touched files are green (`602 passed`). The repo's own pre-commit gates run clean against
  them too: ruff format, ruff check, pycodestyle, pyright (0 errors), `check_import_rules.py`,
  `check_tool_registration.py`, `check_docstrings.py`. Four other tests in
  `tests/sdk/agent/test_acp_agent.py` fail here — `TestACPFileSecretMaterialisation` (3) and
  `TestACPDataDirIsolation::test_codex_sets_codex_home_to_conversation_root` — and fail
  identically on unmodified `main`, so they're pre-existing. For scale, the wider `tests/sdk`
  suite on Windows is `44 failed, 5698 passed` on unmodified `main` (701a21f). Those 44 are
  `context/prompts/test_prompt_snapshot.py` (26), the three `git/` files (13), the four ACP ones
  above, and `utils/test_path.py` (1) — none of them touched by this PR. I haven't triaged them
  and mention the number only as context.
- None of this is visible to CI: the `windows-tests` job in `tests.yml` only runs when
  `openhands-tools/**`, `tests/tools/**`, `pyproject.toml`, `uv.lock` or the workflow itself
  changed, and its pytest invocation targets `tests/tools` — so `openhands-sdk` gets no Windows
  coverage under any condition. The tests here are platform-independent (`shutil.which`
  monkeypatched, host-neutral paths) so they do run in the Ubuntu `sdk-tests` job. I haven't
  widened the filter — that would surface the ~44 failures above and is a maintainer call.
  Happy to do it as a follow-up.

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/acp_agent.py
+++ b/openhands-sdk/openhands/sdk/agent/acp_agent.py
@@ -22,6 +22,7 @@ import inspect
 import json
 import os
 import re
+import shutil
 import threading
 import time
 import uuid
@@ -332,6 +333,29 @@ def _select_auth_method(
         if method_id in method_ids and env_var in env:
             return method_id
     return None
+
+
+def _resolve_executable(program: str) -> str:
+    """Resolve *program* against ``PATH`` before handing it to ``exec``.
+
+    ``asyncio.create_subprocess_exec`` bottoms out in ``CreateProcess`` on
+    Windows, which resolves a bare name against ``PATH`` but only ever appends
+    ``.exe`` — it never consults ``PATHEXT``. Every npm-installed CLI entry
+    point is a ``.cmd`` shim (``npx.cmd``, ``claude-agent-acp.cmd``), so the
+    registry's default ``npx -y <pkg>`` launch command raises
+    ``FileNotFoundError`` on Windows even with npm correctly on ``PATH``.
+    :func:`shutil.which` does honour ``PATHEXT``, so resolving first makes the
+    shim spawnable.
+
+    A name that already carries a directory component is returned untouched, as
+    is one ``which`` cannot resolve — the latter keeps the existing
+    ``ACPSpawnError`` behaviour for a genuinely missing CLI rather than
+    swallowing it here. On POSIX ``which`` finds the same file ``execvp`` would,
+    so this is behaviour-preserving.
+    """
+    if os.path.dirname(program):
+        return program
+    return shutil.which(program) or program
 
 
 def _with_codex_base_url(
@@ -2596,7 +2620,9 @@ class ACPAgent(AgentBase):
                 for conflict in conflicts:
                     env.pop(conflict, None)
 
-        command = self.acp_command[0]
+        # Resolve through PATHEXT-aware lookup: create_subprocess_exec cannot
+        # spawn npm's .cmd shims from a bare name on Windows. No-op on POSIX.
+        command = _resolve_executable(self.acp_command[0])
         args = list(self.acp_command[1:]) + list(self.acp_args)
         # Codex ignores OPENAI_BASE_URL; translate it into the config key read by
         # the adapter. The helper returns a child-only environment and never
--- a/openhands-sdk/openhands/sdk/settings/model.py
+++ b/openhands-sdk/openhands/sdk/settings/model.py
@@ -1734,11 +1734,19 @@ class ACPAgentSettings(AgentSettingsBase):
 
         When *command* is an ``npx`` invocation of this provider's package and
         the provider's ``binary_name`` resolves via :func:`shutil.which`, return
-        ``[binary_name, *extra]`` (preserving trailing args like gemini's
+        ``[<resolved path>, *extra]`` (preserving trailing args like gemini's
         ``--acp``) — running the agent-server image's pinned wrapper instead of
         downloading npm-latest. Returned unchanged otherwise: no pinned binary
         (custom server), a non-matching/non-npx command, or the binary not on
         ``PATH`` (local dev).
+
+        The *resolved* path from ``shutil.which`` is used rather than the bare
+        ``binary_name``: the caller feeds this straight to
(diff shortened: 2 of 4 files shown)


yak1ex:
At least, I confirmed independently that running `npx ...` failed, so typical ACP server can't work, and `shutil.which()` resolves the issue on Windows.

dar-brett:
Independently validated on Windows against a clean Canvas main build. A fresh Codex ACP conversation failed with `ACPSpawnError: [WinError 2]` because `asyncio.create_subprocess_exec` did not resolve the npm `npx.cmd` shim through `PATHEXT`.

Applying this PR made `npx` resolve to its absolute `.CMD` path. Nineteen focused tests and a native Windows subprocess smoke test passed. In a Canvas build also carrying OpenHands/OpenHands#16381, a real Codex ACP 1.1.7 conversation transitioned from `running` to `finished` and returned `PATCHED_OK` without error events.

Recommend merging and releasing this fix. This is the SDK half of the Windows ACP repair; OpenHands/OpenHands#16381 is also required so the packaged desktop app contains a working npm installation.

PR #4410, opened 2026-08-07
fix(profiles): store acp_command as a token list, not a shell string

Telov:
HUMAN:

Had issues installing OpenHands on windows. Putting up some PRs to help the community

---

AGENT:

## Why

`ACPAgentProfile.acp_command` was the only place a launch command existed as a
shell string. `_build_acp_settings` parsed it with POSIX-mode `shlex`
(`profiles/resolver.py:277`), where backslash is an escape character, so every
separator in a Windows path was eaten:

```python
>>> shlex.split(r'node C:\Users\me\dist\index.js')
['node', 'C:Usersmedistindex.js']
```

The launch then failed with a file-not-found naming a path the user never typed,
which reads as a missing file rather than a mangled argument. Quoting worked
around it, but nothing indicated quoting was required and the unquoted form is
what people type. It bites hardest on Windows, where an absolute path in
`acp_command` is the standard way out when the provider presets can't launch.

**Both ends of the round trip were already token lists**, so the string was a
lossy intermediate that nothing needed:

```
ACPAgentSettings.acp_command: list[str]      # settings/model.py:1444
        │  shlex.join   (profiles/seed.py:53)
        ▼
ACPAgentProfile.acp_command: str | None      # profiles/agent_profile.py:267
        │  shlex.split  (profiles/resolver.py:277)
        ▼
ACPAgentSettings.acp_command: list[str]
```

Profiles persist as JSON (`agent_profile_store.py:116`, `:154`, `:173`), which
represents a string list natively; the sibling `acp_args` on the same model is
already `list[str] | None`; and the tokens go straight to
`create_subprocess_exec` — no shell is ever involved.

## Summary

- `ACPAgentProfile.acp_command` becomes `list[str] | None`.
- Both `shlex` calls are gone — `shlex.split` in `profiles/resolver.py` and
  `shlex.join` in `profiles/seed.py`. Neither module imports `shlex` any more.
- `_migrate_v2_to_v3` in the existing `_AGENT_PROFILE_MIGRATIONS` registry
  splits a legacy string with the same POSIX `shlex.split` the resolver used to
  apply, so a stored profile resolves to exactly the tokens it resolved to
  before. That is an exact inverse for every string this codebase wrote, since
  they all came from `shlex.join`.
- Malformed input (unbalanced quotes) falls back to a whitespace split rather
  than raising or dropping the value: loading a stored profile must not fail,
  and `None` would be worse than a rough tokenization — it means "use the server
  default", so for any `acp_server` other than `custom` a configured override
  would be silently replaced by the provider's own command. Such a command never
  launched on `main` either (`shlex.split` raised on it), so keeping the tokens
  only moves *where* it fails, from a resolver error to a spawn error naming the
  command the user typed.
- A payload with no `schema_version` is migrated forward from
  `_UNVERSIONED_AGENT_PROFILE_SCHEMA_VERSION` rather than assumed current. The
  store always persists `schema_version`, so an absent one means a hand-written
  or API-supplied body — the shape a client built against v2 still POSTs. The
  migrations are no-ops on already-current values, so a modern client that
  simply omits the field is unaffected.

A Windows path that was typed unquoted was already being destroyed by that same
split, so it can't be recovered in the migration — such a profile has never
launched and the command has to be re-entered. Splitting it changes nothing
about what the profile does; removing the split for new values is the point.

## How to Test

```
uv run pytest tests/sdk/profiles -q
```

`155 passed` on this branch. New cases in `tests/sdk/profiles/test_agent_profile.py`:

- `test_v2_acp_command_string_migrates_to_token_list`
- `test_v2_malformed_command_does_not_fall_back_to_provider_default` — the
  regression test for the first review finding: a malformed override on a
  non-`custom` server must not resolve to the provider's default command
- `test_unversioned_payload_migrates_legacy_command_string` and
  `test_unversioned_payload_leaves_a_token_list_alone` — the second finding,
  plus the no-op guarantee for current-shape values. Mirrored at the REST layer
  by `test_save_unversioned_body_upgrades_legacy_acp_command_string` and
  `test_save_unversioned_body_keeps_acp_command_token_list` in
  `tests/agent_server/test_agent_profiles_router.py`
- `test_v2_acp_command_migration_inverts_shlex_join` — round-trips
  `shlex.join(["/opt/my acp/bin/server", "--flag", "a b"])` back to the original
  tokens, including the arguments containing spaces
- `test_v2_acp_command_malformed_quotes_falls_back_to_whitespace_split`
- `test_v2_acp_command_empty_migrates_to_none` (`None` and `""`)
- `test_windows_path_command_survives_round_trip` — the regression test for the
  bug: `[r"C:\Program Files\nodejs\node.exe", r"C:\Users\me\dist\index.js"]`
  survives a save/load round trip verbatim

Four existing tests changed:

- three that passed a shell string now pass the equivalent token list;
- `test_dry_run_normalizes_settings_build_failure` used an unbalanced-quote
  command to produce a settings-build failure that the dry-run must report
  rather than raise. That input can no longer fail, so it now uses
  `acp_server="custom"` with no `acp_command` — still a profile that validates
  but cannot build settings, so the test keeps asserting the same contract.

Also green: `tests/agent_server/test_agent_profiles_router.py` and
`tests/agent_server/test_agent_profile_conv_start.py` (`95 passed`), the two
files outside `tests/sdk/profiles` that construct `ACPAgentProfile`. Ruff check
and format clean on the touched paths.

**End-to-end**, on Windows 10 19045 / Python 3.13. Drives the whole chain a user
hits — a profile JSON on disk → `AgentProfileStore.load` (schema migration) →
`resolve_agent_profile` → `resolve_acp_command()` →
`asyncio.create_subprocess_exec` → a real ACP `initialize` handshake (pre-auth,
so no credentials involved):

```
schema_version=3

1. legacy v2 shell string  ->  migration  ->  spawn
  stored acp_command : 'cmd /c claude-agent-acp'
  loaded profile     : ['cmd', '/c', 'claude-agent-acp']
  schema_version     : 3
  resolved settings  : ['cmd', '/c', 'claude-agent-acp']
  spawn              : OK   handshake ok, agent=@agentclientprotocol/claude-agent-acp v0.65.0

2. Windows absolute path as a token list  ->  spawn
  stored acp_command : ['C:\\Users\\<user>\\AppData\\Roaming\\npm\\claude-agent-acp.CMD']
  resolved settings  : ['C:\\Users\\<user>\\AppData\\Roaming\\npm\\claude-agent-acp.CMD']
  survived verbatim  : True
  spawn              : OK   handshake ok, agent=@agentclientprotocol/claude-agent-acp v0.65.0

3. the same absolute path stored the OLD way (shell string)
  stored acp_command : 'C:\\Users\\<user>\\AppData\\Roaming\\npm\\claude-agent-acp.CMD'
  resolved settings  : ['C:Users<user>AppDataRoamingnpmclaude-agent-acp.CMD']
  mangled by shlex   : True
  spawn              : FAIL FileNotFoundError: [WinError 2] The system cannot find the file specified
```

(2) is the capability this adds; (3) is the same command expressed the old way,
and is the bug. (3) stays broken through the migration on purpose — the string
was already destroyed before this change, so the profile has never launched and
the command has to be re-entered.

**The migration is faithful**, checked against `main` rather than asserted. Both
stored strings above, loaded and resolved on unmodified `main` (`701a21f`):

```
schema_version on main: 2
  legacy   stored='cmd /c claude-agent-acp'
           main resolves to -> ['cmd', '/c', 'claude-agent-acp']
  winpath  stored='C:\\Users\\<user>\\AppData\\Roaming\\npm\\claude-agent-acp.CMD'
           main resolves to -> ['C:Users<user>AppDataRoamingnpmclaude-agent-acp.CMD']
```

Identical to what the v2→v3 migration produces in both cases, well-formed and
mangled alike. Script available on request; it imports the store and resolver
from the checkout, so it exercises the shipped code path.

## Video/Screenshots

n/a — terminal only.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Python-API-visible; REST schema unchanged, REST behaviour was not.**
  `ACPAgentProfile` is a public `openhands.sdk.profiles` export, so anything
  constructing one in Python has to pass a list. An earlier revision of this
  description claimed the REST contract was unaffected; that was only true of
  the generated *document*. `save_agent_profile` validates the raw POST body
  through `validate_agent_profile`, so a body carrying a legacy `acp_command`
  string and no `schema_version` started returning 422 where `main` accepted
  it. Caught in review, fixed by migrating unversioned payloads forward (see
  Summary). Profile payloads still cross `/api/agent-profiles` as an opaque
  object
  (`AgentProfileDetailResponse.profile` is `{"type": "object",
  "additionalProperties": true}`), and `ACPAgentProfile` has no component in
  either the public or the full FastAPI schema.
  `.github/scripts/export_agent_server_openapi.py` emits a **byte-identical**
  document on this branch and on `main` (both md5 `659e5b53eae5…`), so the
  generated ts-client is unchanged and `oasdiff` has nothing to report. The two
  surfaces that do change are the Python constructor and the on-disk profile
  JSON — the latter is what the migration is for. `acp_args` on the same model
  is already `list[str] | None`, so the shape isn't new here either.
- **Repo gates, run locally against the touched files** — ruff format, ruff
  check, pycodestyle, pyright (0 errors), `scripts/check_import_rules.py`,
  `scripts/check_tool_registration.py`, `.github/scripts/check_docstrings.py`:
  all clean. `check_persisted_settings_compat.py` validates its 9 checked-in
  fixtures (its PyPI-baseline half needs a Linux runner). Replicating the
  curated-`__all__` pairing `check_sdk_api_breakage.py` performs, griffe reports
  **0 counted breakages relative to `main`**: the `acp_command` `Field(...)`
  diff classifies as metadata-only, and `AGENT_PROFILE_SCHEMA_VERSION` is not in
  `openhands.sdk.__all__`. That's a local ref-to-ref check, not the PyPI
  baseline CI uses.
- **Why not a deprecation runway instead?** CONTRIBUTING points at the
  deprecation mechanism for changes that could break downstream clients, so to
  be explicit about the choice: the persisted surface *does* get a runway — the
  v2→v3 migration means no stored profile breaks. What has no runway is the
  Python constructor, for callers passing `acp_command="..."` directly. The
  softer landing is one line: keep the annotation `list[str] | str | None` with
  a `field_validator` that `shlex.split`s a string and warns, then narrow it in
  a later release. I didn't do that because it re-admits the ambiguous form at
  the boundary this PR exists to close, and because the only in-tree writer is
  `profiles/seed.py`. If you'd rather have the runway, say so and I'll push the
  validator — it doesn't change anything else here.
- `AGENT_PROFILE_SCHEMA_VERSION` goes 2 → 3. Migration is forward-only, matching
  `_migrate_v1_to_v2`; a v3 profile loaded by older code fails the existing
  "newer than supported" check rather than silently misreading the field.
- Independent of the two other Windows ACP PRs — #4408 (`PATHEXT` spawn
  resolution) and #4409 (process-tree teardown). Different files, any order.

### DIFF
--- a/openhands-sdk/openhands/sdk/profiles/agent_profile.py
+++ b/openhands-sdk/openhands/sdk/profiles/agent_profile.py
@@ -9,6 +9,7 @@ See epic #3713 for the resolution model.
 
 from __future__ import annotations
 
+import shlex
 from collections.abc import Callable, Mapping
 from typing import Annotated, Any, Literal
 from uuid import UUID, uuid4
@@ -33,7 +34,7 @@ from openhands.sdk.settings.model import (
 from openhands.sdk.tool import Tool
 
 
-AGENT_PROFILE_SCHEMA_VERSION = 2
+AGENT_PROFILE_SCHEMA_VERSION = 3
 
 
 class ProfileVerificationSettings(BaseModel):
@@ -264,11 +265,14 @@ class ACPAgentProfile(AgentProfileBase):
             "initialize/authenticate, and new_session()/load_session()."
         ),
     )
-    acp_command: str | None = Field(
+    acp_command: list[str] | None = Field(
         default=None,
         description=(
-            "Optional explicit command to launch the ACP subprocess. Leave "
-            "blank to use the default for ``acp_server``."
+            "Optional explicit command to launch the ACP subprocess, as an "
+            "argv token list (e.g. ``['npx', '-y', 'some-acp@1.2.3']``). The "
+            "tokens are passed to the subprocess verbatim — no shell is "
+            "involved, so nothing needs quoting or escaping. Leave unset to "
+            "use the default for ``acp_server``."
         ),
     )
     acp_args: list[str] | None = Field(
@@ -335,8 +339,39 @@ def _migrate_v1_to_v2(payload: dict[str, Any]) -> dict[str, Any]:
     return payload
 
 
+def _migrate_v2_to_v3(payload: dict[str, Any]) -> dict[str, Any]:
+    """``acp_command`` moved from a shell string to an argv token list.
+
+    Legacy strings are split with the same POSIX ``shlex.split`` the resolver
+    used to apply, so a profile resolves to exactly the tokens it resolved to
+    before. That is an exact inverse for everything this codebase wrote — the
+    strings came from ``shlex.join`` in ``profiles/seed.py`` — and matches what
+    a hand-written POSIX string already resolved to.
+
+    A Windows path typed unquoted (``C:\\Users\\me\\x.js``) was already being
+    destroyed by that same split, so it cannot be recovered here; such a
+    profile has never launched and the command has to be re-entered. Splitting
+    it is still the honest migration: it changes nothing about what the profile
+    does. Removing the split for *new* values is the point of the change.
+
+    Malformed input (unbalanced quotes) becomes ``None`` rather than raising:
+    loading a stored profile must not fail. Today the failure surfaces from
+    ``resolve_agent_profile`` as a diagnostic, and ``None`` keeps it there — a
+    ``custom`` server with no command still reports the same resolver error.
+    """
+    command = payload.get("acp_command")
+    if isinstance(command, str):
+        try:
+            payload["acp_command"] = shlex.split(command) or None
+        except ValueError:
+            payload["acp_command"] = None
+    payload["schema_version"] = 3
+    return payload
+
+
 _AGENT_PROFILE_MIGRATIONS: dict[int, PersistedProfileMigrator] = {
     1: _migrate_v1_to_v2,
+    2: _migrate_v2_to_v3,
 }
 
 
--- a/openhands-sdk/openhands/sdk/profiles/agent_profile.py
+++ b/openhands-sdk/openhands/sdk/profiles/agent_profile.py
@@ -36,6 +36,17 @@ from openhands.sdk.tool import Tool
 
 AGENT_PROFILE_SCHEMA_VERSION = 3
 
+_UNVERSIONED_AGENT_PROFILE_SCHEMA_VERSION = 2
+"""Assumed ``schema_version`` for a payload that carries none.
+
+The store has always persisted ``schema_version`` (it has a model default), so
+an absent one means a hand-written or API-supplied payload — the shape a client
+built against an older release still sends. Such a payload is migrated forward
+from here rather than assumed current, so a legacy field shape is upgraded
+instead of rejected. Migrations are no-ops on values already in the new shape,
+so a current client that simply omits the field is unaffected.
+"""
+
 
 class ProfileVerificationSettings(BaseModel):
     """Secret-free critic/refinement policy for a profile.
@@ -354,17 +365,22 @@ def _migrate_v2_to_v3(payload: dict[str, Any]) -> dict[str, Any]:
     it is still the honest migration: it changes nothing about what the profile
     does. Removing the split for *new* values is the point of the change.
 
-    Malformed input (unbalanced quotes) becomes ``None`` rather than raising:
-    loading a stored profile must not fail. Today the failure surfaces from
-    ``resolve_agent_profile`` as a diagnostic, and ``None`` keeps it there — a
-    ``custom`` server with no command still reports the same resolver error.
+    Malformed input (unbalanced quotes) falls back to a whitespace split rather
+    than raising or dropping the value. Loading a stored profile must not fail,
+    but ``None`` would be worse than a rough tokenization: ``None`` means "use
+    the server default", so for any ``acp_server`` other than ``custom`` a
+    configured override would be silently replaced by the provider's own
+    command. Such a command never launched on ``main`` either — ``shlex.split``
+    raised on it — so keeping the tokens only changes *where* it fails, from a
+    resolver error to a spawn error naming the command the user actually typed.
     """
     command = payload.get("acp_command")
     if isinstance(command, str):
         try:
-            payload["acp_command"] = shlex.split(command) or None
+            tokens = shlex.split(command)
         except ValueError:
-            payload["acp_command"] = None
+            tokens = command.split()
+        payload["acp_command"] = tokens or None
     payload["schema_version"] = 3
     return payload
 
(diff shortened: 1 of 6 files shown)


openhands-sdk/openhands/sdk/profiles/agent_profile.py
```diff
@@ -264,11 +265,14 @@ class ACPAgentProfile(AgentProfileBase):
             "initialize/authenticate, and new_session()/load_session()."
         ),
     )
-    acp_command: str | None = Field(
+    acp_command: list[str] | None = Field(
```
VascoSch92, line 269: this is a breaking change right?

VascoSch92: Oh I see that you have a migration for that.

openhands-sdk/openhands/sdk/profiles/agent_profile.py
```diff
@@ -335,8 +339,39 @@ def _migrate_v1_to_v2(payload: dict[str, Any]) -> dict[str, Any]:
...
+    loading a stored profile must not fail. Today the failure surfaces from
+    ``resolve_agent_profile`` as a diagnostic, and ``None`` keeps it there — a
+    ``custom`` server with no command still reports the same resolver error.
+    """
+    command = payload.get("acp_command")
+    if isinstance(command, str):
+        try:
+            payload["acp_command"] = shlex.split(command) or None
+        except ValueError:
+            payload["acp_command"] = None
```
VascoSch92: Malformed acp_command strings get silently converted to None during migration; for non-custom acp_server values this bypasses all validation, so the agent silently launches with the provider's default command instead of the user's configured override.

openhands-sdk/openhands/sdk/profiles/agent_profile.py
VascoSch92, line 363: The v2→v3 migration is skipped whenever schema_version is absent from the payload (the normal shape for a fresh API-created profile), so legacy string acp_command values get rejected with a 422 instead of being upgraded.

VascoSch92 (commented):
Left a couple of comments. :-)

Could be possible to trim all the comments in the code?

Thanks

openhands-sdk/openhands/sdk/profiles/agent_profile.py
Telov: Migration handles the stored side. Constructor's the part with no runway, can add a validator that takes a string and warns if you want one.

openhands-sdk/openhands/sdk/profiles/agent_profile.py
Telov: Fixed in 11c00aa5, falls back to `command.split()`. Never launched on main either so nothing working regressed, it just fails at spawn now instead of in the resolver.

openhands-sdk/openhands/sdk/profiles/agent_profile.py
Telov: Fixed in 11c00aa5, absent `schema_version` migrates from v2 now instead of being assumed current. Description was wrong about this not being REST-visible too, corrected.

Not sure you want it flipped there though, applies to future migrations as well. Other option is taking the string in a validator and leaving the gate alone.

Telov:
Comments are gone, and parametrized the migration tests to cut the duplication. +105 now, under where it was when you looked, even with the two fixes above added.

Issue #4288, opened 2026-07-28
Design: reference-only credentials and safe runtime delivery


simonrosenberg:
> **Consolidated design document.** This began as `OpenHands/OpenHands#15298` (public 2026-07-16 → 07-27), then moved with \~62 other issues into the private `OpenHands/enterprise` repo (`#64`) during the Agent Canvas restructuring. GitHub refuses private → public issue transfers, so it is reproduced here.
>
> The original thread was a design proposal plus 12 comments of review, correction and counter-argument. Rather than mirror 12 comments (\~88 KB, over GitHub's 65 KB body limit), everything is merged into this single document. **Positions are attributed inline** — `enyst` and `smolpaws` raised most of the corrections, and their conclusions are marked as theirs. Verbatim originals remain at `OpenHands/enterprise#64` for anyone with access.
>
> Removed in reproduction: customer account names, direct customer quotes, and verbatim internal-document text. Internal Linear/Notion URLs are kept for provenance and will not resolve for non-members. Cross-references to `OpenHands/OpenHands#150xx` now point at private or renumbered issues and will 404.
>
> **Decision requested:** the two items in §10.

---

# Reference-only credentials and safe runtime delivery

## 1\. The invariant

The original proposal was absolute:

> Durable configuration and conversation state contain credential references only. Secret values and ciphertext exist only in a credential store.

**That was corrected during review**, because it is incompatible with delivering a credential to a sandbox at all — once delivered, plaintext necessarily exists in memory, environment, or a file. The agreed wording:

> Outside the credential service, durable OpenHands state contains **references only**. Runtime-visible delivery may create a bounded plaintext working copy in memory, environment, or non-snapshotted runtime storage; **that copy is never authoritative.**

## 2\. Motivation

A production resume failure exposed the coupling. An older conversation held an encrypted Tavily MCP value inside `StoredConversation`; a Pydantic validator-order regression rejected it during cold deserialization, so agent-server skipped conversation registration and the UI entered a 404/WebSocket loop. Reported as OpenHands/software-agent-sdk#4096, fixed for that serialized form by OpenHands/software-agent-sdk#4099 — but the underlying problem remains: **optional integration credentials participate in conversation schema validation.**

There are also multiple credential copies and identities:

* SDK `StaticSecret`, `LookupSecret` and `SecretRegistry` resolve plaintext inside the runtime; some ACP consumers receive the entire registry.
[...]

PR #4413, opened 2026-08-07
fix(subagent): decrypt encrypted model profiles

### LINKED ISSUE #4558: Encrypted LLM profiles are decrypted only on the conversation path — FallbackStrategy (and sub-agents) load them without the cipher

### Bug Description

`LLMProfileStore.load()` takes an optional `cipher`, and only the conversation-launch path passes it. Every other caller loads with `cipher=None`, so on a server that has an `OH_SECRET_KEY` (Agent Canvas always generates and passes one) the Fernet ciphertext stored at rest is handed to the provider as the API key.

Two call sites are affected:

| call site | status |
| --- | --- |
| `openhands-sdk/openhands/sdk/subagent/registry.py:228` — `agent_definition_to_factory` | reported in #4270; fix PRs #4413 and #4183 open, neither merged |
| `openhands-sdk/openhands/sdk/llm/fallback_strategy.py:142` — `FallbackStrategy._iter_fallbacks` | **not reported, and not covered by either PR** |

For contrast, the path that gets it right is `openhands-sdk/openhands/sdk/profiles/resolver.py:336` (`llm_store.load(profile.llm_profile_ref, cipher=cipher)`), wired from `conversation_service.py:353-359`.

`openhands-tools/openhands/tools/preset/default.py:140` uses the same factory, but every builtin subagent ships `model: inherit`, so it never reaches the store.

This issue tracks the `FallbackStrategy` call site and the structural fix; the sub-agent call site is left to #4270 and its PRs.

### Expected Behavior

A profile persisted with the server cipher resolves to its decrypted API key through *every* resolution path, exactly as it does for the conversation's primary agent.

### Actual Behavior

The fallback LLM is constructed with the Fernet ciphertext as its API key. Reproduced on `main` @ d98fd9500 with `uv run python /tmp/repro.py` (source under Minimal Code Sample):

```
$ uv run python /tmp/repro.py
fallback model  : anthropic/claude-sonnet-4-5
fallback api_key: gAAAAABqhuqAuR ...
is ciphertext?  : True
```

The provider then rejects it, e.g. the report in #4270: `litellm.AuthenticationError: LiteLLM Virtual Key expected. Received=gAAA****L7Q=, expected to start with 'sk-'`.

For the fallback path this is worse than a plain failure: `try_fallback` swallows the auth error per fallback (`fallback_strategy.py:145-149`) and re-raises the *primary* error, so the operator sees the original failure and no indication that every fallback was unusable.

### Steps to Reproduce

1. Save the repro under Minimal Code Sample to `/tmp/repro.py`.
2. From a checkout of this repo, run `uv run python /tmp/repro.py`.
3. Observe the fallback LLM's `api_key` is the `gAAAAA…` Fernet token rather than `sk-REAL`.

Equivalent end-to-end path: start the agent-server with `OH_SECRET_KEY` set, save an LLM profile through `POST /api/profiles/{name}` (which encrypts at rest, `profiles_router.py:201-206`), then reference that profile from another profile's `fallback_strategy.fallback_llms` and force a primary-model failure.

### Acceptance Criteria

- [ ] `FallbackStrategy` resolves fallback profiles with the server cipher, so an encrypted profile yields its decrypted API key.
- [ ] Omitting the cipher is made structurally impossible rather than fixed per call site: `cipher` becomes a required keyword-only argument on `LLMProfileStore.load()`, or is bound at construction (`LLMProfileStore(base_dir, cipher=...)`).
- [ ] The `LLMProfileLoader` protocol (`llm_profile_store.py:42-50`) stays satisfiable by alternate/cloud backends after that change.
- [ ] A regression test asserts a cipher-saved profile resolves to the decrypted key through the fallback path, not just the conversation-launch path.
- [ ] An audit confirms no remaining `LLMProfileStore.load()` call site drops the cipher (the sub-agent site is covered by #4270 / #4413 / #4183).

### Installation Method

Local checkout (`uv run`)

### SDK Version

`main` @ d98fd9500 (`openhands-sdk` 1.42.1)

### Version Confirmation

### Python Version

3.13.11

Reproduces with any provider; the report in #4270 hit it through the OpenHands LiteLLM proxy.

### Operating System

macOS 26.5 (arm64); not platform-specific.

### Logs and Error Messages

```
litellm.AuthenticationError: LiteLLM Virtual Key expected. Received=gAAA****L7Q=, expected to start with 'sk-'
```

On the fallback path this line never surfaces — it is caught at `fallback_strategy.py:145-149` and the primary error is re-raised instead.

### Minimal Code Sample

```python
import tempfile
from pathlib import Path
from openhands.sdk.llm.fallback_strategy import FallbackStrategy
from openhands.sdk.llm.llm_profile_store import LLMProfileStore
from openhands.sdk.utils.cipher import Cipher

cipher = Cipher("server-oh-secret-key")
d = Path(tempfile.mkdtemp())
store = LLMProfileStore(d)
# exactly how profiles_router.py:201-206 persists a profile when OH_SECRET_KEY is set
store.save("backup", LLM(model="anthropic/claude-sonnet-4-5",
                         api_key=SecretStr("sk-REAL"), usage_id="backup"),
           include_secrets=True, cipher=cipher)

primary = LLM(model="anthropic/claude-opus-4-5", api_key=SecretStr("sk-PRIMARY"),
              usage_id="primary",
              fallback_strategy=FallbackStrategy(fallback_llms=["backup"],
                                                 profile_store_dir=str(d)))
fb = next(primary.fallback_strategy._iter_fallbacks())
print("fallback model  :", fb.model)
print("fallback api_key:", fb.api_key.get_secret_value()[:14], "...")
print("is ciphertext?  :", fb.api_key.get_secret_value().startswith("gAAAAA"))
```

**Why this stays invisible for some users.** `validate_secret` only attempts decryption when the value starts with `gAAAAA` (`utils/pydantic_secrets.py:110-118`). A plaintext profile therefore behaves identically on both paths:

```
legacy/plaintext profile   main-agent=sk-REAL      sub-agent=sk-REAL      sub OK=True
GUI-saved w/ cipher        main-agent=sk-REAL      sub-agent=gAAAAABqhu   sub OK=False
```

Encryption is applied only at save time (#3161, 2026-05-08), `store.save()` is the only writer, and no migration re-encrypts on load. So profiles created before that change — or by any server started without `OH_SECRET_KEY` / a session key — stay plaintext on disk and keep working indefinitely, for sub-agents and fallbacks alike. Those users have working setups that break the first time they re-save the profile in the UI, and a fresh install cannot reproduce their recipe. That is the source of the conflicting reports on #4270.

Shimada666:
HUMAN:

This fixes sub-agent authentication failures when named LLM profiles are encrypted at rest.

---

AGENT:

## Why

When `OH_SECRET_KEY` is configured, saved LLM profile secrets are encrypted at rest. File-based sub-agents loaded named profiles without the conversation cipher, so the encrypted API key reached the LLM provider and authentication failed.

## Summary

- Forward the conversation cipher when registering file-based, plugin, and remote sub-agent definitions.
- Decrypt named LLM profiles before creating a sub-agent.
- Add a regression test for encrypted profile API keys.

## Issue Number

Related to #4288.
Fixes https://github.com/OpenHands/software-agent-sdk/issues/4558

## How to Test

1. Run `uv run pytest -q tests/sdk/subagent/test_subagent_registry.py tests/sdk/conversation/test_local_conversation_plugins.py` — 81 passed.
2. Run `uv run pytest -q tests/agent_server/test_conversation_service.py tests/agent_server/test_conversation_service_plugin.py` — 109 passed.
3. Run pre-commit on all changed files — Ruff, Pyright, import rules, and tool registration passed.
4. Save a Luna LLM profile with `include_secrets=True` and a `Cipher`, load an `evidence-coder` agent from an explicit plugin through `LocalConversation`, and instantiate the registered factory. The smoke test returned model `openai/gpt-5.6-luna` and confirmed the API key matched the original plaintext.

## Video/Screenshots

Not applicable; this is an SDK credential-loading fix. The deterministic regression test and plugin smoke-test results are described above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Documentation: https://github.com/OpenHands/docs/pull/696

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -502,6 +502,7 @@ def _register_agent_definitions(
     agent_defs: list["AgentDefinition"],
     *,
     context: str,
+    cipher: Cipher | None,
 ) -> None:
     """Register agent definitions into the subagent registry.
 
@@ -516,7 +517,7 @@ def _register_agent_definitions(
     registered = 0
     for agent_def in agent_defs:
         try:
-            factory = agent_definition_to_factory(agent_def)
+            factory = agent_definition_to_factory(agent_def, cipher=cipher)
             register_agent_if_absent(
                 name=agent_def.name,
                 factory_func=factory,
@@ -930,6 +931,7 @@ class ConversationService:
             _register_agent_definitions(
                 stored.agent_definitions,
                 context=f"resuming conversation {stored.id}",
+                cipher=self.cipher,
             )
 
     async def _get_or_load_event_service(
@@ -1432,6 +1434,7 @@ class ConversationService:
             _register_agent_definitions(
                 request.agent_definitions,
                 context=f"conversation {conversation_id}",
+                cipher=self.cipher,
             )
 
         # Plugin loading is now handled lazily by LocalConversation.
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -514,6 +514,7 @@ def _register_agent_definitions(
     agent_defs: list["AgentDefinition"],
     *,
     context: str,
+    cipher: Cipher | None,
 ) -> None:
     """Register agent definitions into the subagent registry.
 
@@ -528,7 +529,7 @@ def _register_agent_definitions(
     registered = 0
     for agent_def in agent_defs:
         try:
-            factory = agent_definition_to_factory(agent_def)
+            factory = agent_definition_to_factory(agent_def, cipher=cipher)
             register_agent_if_absent(
                 name=agent_def.name,
                 factory_func=factory,
@@ -1054,6 +1055,7 @@ class ConversationService:
             _register_agent_definitions(
                 stored.agent_definitions,
                 context=f"resuming conversation {stored.id}",
+                cipher=self.cipher,
             )
 
     def _get_conversation_lock(self, conversation_id: UUID) -> asyncio.Lock:
@@ -1615,6 +1617,7 @@ class ConversationService:
             _register_agent_definitions(
                 request.agent_definitions,
                 context=f"conversation {conversation_id}",
+                cipher=self.cipher,
             )
 
         # Plugin loading is now handled lazily by LocalConversation.
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -1161,6 +1161,7 @@ class LocalConversation(BaseConversation):
             register_plugin_agents(
                 agents=all_plugin_agents,
                 work_dir=self.workspace.working_dir,
+                cipher=self._cipher,
             )
 
         # Combine explicit hook_config with plugin hooks
@@ -1400,6 +1401,7 @@ class LocalConversation(BaseConversation):
                 register_plugin_agents(
                     agents=plugin.agents,
                     work_dir=self.workspace.working_dir,
+                    cipher=self._cipher,
                 )
             if plugin.hooks and not plugin.hooks.is_empty():
                 self._merge_runtime_plugin_hooks(plugin.hooks)
@@ -1438,7 +1440,7 @@ class LocalConversation(BaseConversation):
                 then `~/.openhands/agents/*.md`)
         """
         # register project-level and then user-level file-based agents
-        register_file_agents(self.workspace.working_dir)
+        register_file_agents(self.workspace.working_dir, cipher=self._cipher)
 
     def _ensure_agent_ready(self) -> None:
         """Ensure the agent is fully initialized with plugins and agents loaded.
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -406,6 +406,7 @@ class LocalConversation(BaseConversation):
             if recovered_specs:
                 register_client_tools(recovered_specs)
         self.agent = agent
+        self._cipher = cipher
 
         self._bind_conversation_context(self.agent.llm)
 
@@ -486,8 +487,6 @@ class LocalConversation(BaseConversation):
         # This ensures plugins are loaded before agent initialization
         self.llm_registry = LLMRegistry()
         self._profile_store = LLMProfileStore(profile_store_dir)
-        self._cipher = cipher
-
         # Seed agent_context.secrets into the registry for every agent (regular
         # and ACP), covering callers that skip create_request() — canvas /
         # TypeScript, or the server-side agent_settings -> create_agent fold.
@@ -835,6 +834,7 @@ class LocalConversation(BaseConversation):
                 stuck_detection=self._stuck_detector is not None,
                 visualizer=type(self._visualizer) if self._visualizer else None,
                 delete_on_close=self.delete_on_close,
+                cipher=self._cipher,
                 tags=tags,
             )
 
@@ -1605,6 +1605,8 @@ class LocalConversation(BaseConversation):
         See #3443 for background.
         """
         llm._call_context = self.get_llm_call_context()
+        if llm.fallback_strategy is not None:
+            llm.fallback_strategy._bind_cipher(self._cipher)
 
     def _condenser_for_switched_llm(
         self,
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -1214,6 +1214,7 @@ class LocalConversation(BaseConversation):
             register_plugin_agents(
                 agents=all_plugin_agents,
                 work_dir=self.workspace.working_dir,
+                cipher=self._cipher,
             )
 
         # Combine explicit hook_config with plugin hooks
@@ -1459,6 +1460,7 @@ class LocalConversation(BaseConversation):
                 register_plugin_agents(
                     agents=plugin.agents,
                     work_dir=self.workspace.working_dir,
+                    cipher=self._cipher,
                 )
             if plugin.hooks and not plugin.hooks.is_empty():
                 self._merge_runtime_plugin_hooks(plugin.hooks)
@@ -1497,7 +1499,7 @@ class LocalConversation(BaseConversation):
                 then `~/.openhands/agents/*.md`)
         """
         # register project-level and then user-level file-based agents
-        register_file_agents(self.workspace.working_dir)
+        register_file_agents(self.workspace.working_dir, cipher=self._cipher)
 
     def _ensure_agent_ready(self) -> None:
         """Ensure the agent is fully initialized with plugins and agents loaded.
--- a/openhands-sdk/openhands/sdk/llm/fallback_strategy.py
+++ b/openhands-sdk/openhands/sdk/llm/fallback_strategy.py
@@ -22,6 +22,7 @@ from openhands.sdk.logger import get_logger
 if TYPE_CHECKING:
     from openhands.sdk.llm.llm_response import LLMResponse
     from openhands.sdk.llm.utils.metrics import Metrics
+    from openhands.sdk.utils.cipher import Cipher
 
 logger = get_logger(__name__)
 
@@ -55,6 +56,11 @@ class FallbackStrategy(BaseModel):
 
     # Private: lazily resolved LLM instances
     _resolved: list[Any] | None = PrivateAttr(default=None)
+    _cipher: Cipher | None = PrivateAttr(default=None)
+
+    def _bind_cipher(self, cipher: Cipher | None) -> None:
+        """Bind the conversation cipher used for lazy profile resolution."""
+        self._cipher = cipher
 
     def should_fallback(self, error: Exception) -> bool:
         """Whether this error type is eligible for fallback."""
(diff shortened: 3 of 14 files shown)


enyst (commented):
Thank you for the PR! 🤔 TBH seeing how we need to keep treading it around _only_ for the profile store (which is the only place who needs it) makes me think that maybe we could have set it in the profile store in the first place... 

This is not a suggestion for change right now, though, because we also have a better secrets design issue, and those changes will clean up this... I think. @openhands can you find and link that issue? Then tell us WDYT.

enyst:
I am OpenHands (OpenHands-Deepseek-Flash), an AI agent acting on behalf of @enyst. (Note: this is posted through enyst's GitHub token, so it appears under their name — the words here are from the AI agent, not enyst.)

Regarding the "better secrets design issue" you asked about — it's **#4288** (_Design: reference-only credentials and safe runtime delivery_): https://github.com/OpenHands/software-agent-sdk/issues/4288

That's the consolidated design doc enyst contributed to (stdio-MCP, egress-proxy, OS-keyring, iron-proxy), and it's the one whose §2/§9 covers exactly this coupling: the cipher/`expose_secrets` serialization context is threaded through ~30 modules including `llm/llm_profile_store.py`, `profiles/agent_profile_store.py`, and the profile refs — and §9.4 explicitly calls out that removing secret-aware serialization from durable models is _cut_ for now in favor of Phase 0 (canary gate) + Phase 1 (fail-soft loading) + Phase 2 (`SecretRef`).

On the WDYT: I partly agree the threading is a smell — `cipher` is only consumed by the profile store on this path, so it looks heavy to thread it through `register_*_agents` / `agent_definition_to_factory`. But it's worth noting this isn't a new pattern the PR invents: passing `cipher` into `LLMProfileStore.load(...)` is already the codebase convention at ~7 call sites (`profiles/resolver.py`, `LocalConversation` on resume/switch, `settings_router`, `profiles_router`, `agent_profiles_router`). The cipher is a process-global value derived deterministically from `OH_SECRET_KEY`, so "setting it in the profile store" in the constructor (rather than per-call) is feasible and would shrink a diff like this one — but it's a broader API change to the store, and per enyst's own note it gets superseded by #4288 Phase 2 (`SecretRef`, where the store stops holding/ciphering values at all). For this bugfix, keeping the existing `load(name, cipher=...)` convention is the minimal, consistent fix; the real cleanup is tracked in #4288. I'd suggest a small follow-up (e.g. cache the cipher on the `@lru_cache`d store) rather than changing this PR's shape.

So: agree it's temporary, agree it's the profile store's job to own at-rest handling, and #4288 is the right place to track killing the whole pattern — link added above. WDYT, good to proceed with this as-is?

Shimada666:
Thanks for the review! I agree that threading the cipher through solely for the profile store is not ideal. Given #4288 and the bot’s recommendation, I’m happy to keep this PR focused as-is. Please let me know if you’d like me to make any changes before approval.

CarlosLanderas:
Hello, just found this bug as I'm also experiencing it when using subagents:

`subagent AuthenticationError`

Is the fix getting merged? Thanks!

Shimada666:
Thanks @CarlosLanderas for confirming this in real use.

@enyst I rebased the PR onto the latest `main`; all checks are green on `d87a6e8d`. The current `main` still loads named sub-agent profiles without the conversation cipher, and we now have another user hitting the same `subagent AuthenticationError`. Would you mind taking another look when you have a chance?

VascoSch92:
Hey @Shimada666 

Fix looks right...

A few gaps I'd like to see closed:

Sub-conversations don't get a cipher. `TaskManager` and `DelegateExecutor` construct `LocalConversation` without `cipher=,` so cipher is `None` there. The sub-agent's own model profile is fine. But `switch_profile` and `get_or_create_profile_llm` hit the same bug inside the sub-conversation, and the latter is the `vision_inspect` path, which auto-attaches whenever the model isn't multimodal and a vision profile exists.

Two call sites still load profiles with no cipher. F`allbackStrategy._iter_fallbacks` and `register_builtins_agents` (swhich has no cipher parameter to pass one). The second is harmless only because the shipped subagent definitions don't name a profile.

Nothing tests the plumbing. Worth one test asserting the cipher reaches the factory through register_file_agents.

Structural, non-blocking: the cipher gets captured in a closure stored in the process-global _agent_factories, and register_agent_if_absent no-ops on an existing name. So whichever conversation registers a name first pins the cipher for every conversation after it. ( I think this is fine).

One small ask: both failure modes here are silent. No cipher sends the ciphertext as the API key, and a wrong cipher yields api_key=None. Since FERNET_TOKEN_PREFIX already identifies the value, raising with the profile name would save someone a long debugging session.

Shimada666:
Thanks for the detailed review, @vasco. I've addressed the issues you identified:

- propagated the conversation cipher through server, file, and plugin agent registration
- propagated it into TaskManager create/resume, Delegate, and forked conversations
- bound it to lazy fallback-profile loading
- added a fail-fast error with the profile name when an encrypted API key cannot be decrypted; fallback loading no longer swallows that error
- added end-to-end regression coverage for the registration and child-runtime paths

I also rebased onto the latest `main` and trimmed the redundant tests. CI is green on head `652522175`.

I left the process-global registry lifetime unchanged, as you marked that concern non-blocking; fixing it cleanly would require a broader registry-lifecycle change.

Could you please take another look?

VascoSch92:
Thanks for the quick turnaround @Shimada666. 

The propagation looks right. Two things mean the new guard doesn't hold yet though.

`_load_for_execution` matches the requested name against `list_summaries()`, which derives its name from `path.stem`, and defaults to `False` on no match. On a case-insensitive filesystem (the macOS default) `Prod.json` loads fine as `prod`, the match misses, and the ciphertext goes to the sub-agent with no error. Reading `api_key` from `_get_profile_path(name)` removes the mismatch, and drops an O(N) glob from the spawn path.

It also only checks `api_key`. `LLM_SECRET_FIELDS` is three fields wider and those are encrypted the same way, so a Bedrock profile carrying only `aws_*` keys reports `api_key_set: False` and the guard never fires. Walking `LLM_SECRET_FIELDS` covers all four.

Separately, narrowing `_iter_fallbacks` to `except FileNotFoundError` is a behaviour change I don't think you want. `ProviderConnectionNotFound` subclasses `ValueError`, and `try_fallback` iterates the generator outside the loop-body `try`, so one fallback with a deleted connection now aborts the whole chain and masks the primary error. Fail fast on the decryption error only.

`register_builtins_agents` still passes no cipher and has no parameter for one. Harmless while every builtin is `model: inherit`, but it registers at import so it wins the first-registration race. Same reason I'd now like a note on `register_agent_if_absent` that the first caller pins the cipher: I marked that non-blocking when the failure was silent, and it raises now.

Shimada666:
Thanks again, @VascoSch92. I addressed the follow-up in `6496f0385`:

- `_load_for_execution` now reads the exact profile path and walks all `LLM_SECRET_FIELDS`, covering case-insensitive names and Bedrock AWS secrets.
- Added `ProfileDecryptionError`, so only decryption failures fail fast; ordinary load errors, including deleted provider connections, continue through the fallback chain.
- Documented the first-registration semantics in `register_agent_if_absent` and kept the builtin `model: inherit` invariant covered.

I added regression coverage for case-variant paths, all four secret fields with missing/wrong ciphers, deleted provider connections, and builtin inheritance. 152 relevant tests and the full pre-commit suite pass locally. All completed CI jobs are green; four jobs are still running.

Could you please take another look?

enyst (commented):
Thank you both!

I wonder, sorry, if we really need to do this anymore, since https://github.com/OpenHands/software-agent-sdk/pull/4492 has introduced Provider Connections, and my hope is that we can _remove_, not add, but remove most use of cipher in LLM Profiles, because the API keys are in Provider Connections (or will be soon)

We also talked here about the Secrets work, which too, should remove completely I hope, the need for cipher in this codebase. The reason why I care is exactly what you see: with the cipher, it needs to be threaded around in many places, each of those also carry a security risk, and a better design would be safer and less error-prone.

That all said, it’s not the fault of your PR! cc @VascoSch92 
The implementation looks good to me, under the older circumstances. Just a heads up, sorry: these fixes will probably not stay long in the code, but for now I’m for merging 😅

valeriiD:
Confirming this on SDK 1.44.1 (Windows) with a live repro, plus a workaround we're shipping in case it helps others.

## Repro

```python
from openhands.sdk.llm.llm import LLM
from openhands.sdk.llm.llm_profile_store import LLMProfileStore
from openhands.sdk.utils.cipher import Cipher
from openhands.sdk.subagent import agent_definition_to_factory

cipher = Cipher(shared_secret)  # same key the agent server uses (OH_SECRET_KEY)
store = LLMProfileStore(tmp_dir)
store.save("pin", LLM(model="openai/x", api_key="sk-real", base_url=...),
           include_secrets=True, cipher=cipher)

agent_def = ...  # AgentDefinition with frontmatter `model: pin` (+ profile_store_dir=tmp_dir)
factory = agent_definition_to_factory(agent_def)
agent = factory(any_llm)          # factory re-loads the pin WITHOUT a cipher
# agent.llm.api_key is now the Fernet token string ("gAAAA..."), not "sk-real"
```

For us the token then travels inside a conversation-create request and the upstream provider answers **401** (the token is sent as the bearer key). Root cause as in the PR description: `registry._get_profile_store()` constructs the store with no cipher and `_factory` calls `store.load(profile_name)` bare — `LLMProfileStore.load` only decrypts when the caller passes `cipher=`.

## Workaround (no SDK patches)

Resolve the pin yourself with the official kwarg and neutralize the factory's re-load:

```python
llm = store.load(pin_name, cipher=cipher)                  # decrypts
factory = agent_definition_to_factory(def.model_copy(update={"model": "inherit"}), work_dir=...)
agent = factory(llm)                                       # keeps YOUR llm
```

We keep the original (pinned) definition in `agent_definitions` when shipping the agent so ambient registrations stay semantically correct; only the locally-instantiated copy gets the `inherit` override.

One adjacent observation from the same debugging session: `LLMProfileStore.save(..., include_secrets=False)` by default silently writes a profile **without** the key — if that's intentional for sharing workflows it might deserve a louder name or a warning when the caller also passes `cipher=` (the combination reads like "encrypt my secrets" but stores none).

PR #4423, opened 2026-08-07
feat(llm): resolve provider-specific runtime metadata for routed models

### LINKED ISSUE #4421: feat(llm): resolve provider-specific runtime metadata for routed models
## Summary

`LLM.effective_max_input_tokens` currently comes from explicit configuration or LiteLLM's model-level metadata. That is insufficient for routed providers whose individual endpoints expose different runtime limits.

OpenRouter is a concrete example. Its model catalog advertises `deepseek/deepseek-v4-flash-0731` with a 1,048,576-token context, and LiteLLM imports that model-level value from `https://openrouter.ai/api/v1/models`. OpenRouter's provider endpoints API shows different limits for the same model, including 262,144 tokens on CoreWeave and 1,048,576 on Baseten. An `LLM` configured with a provider order preferring CoreWeave can therefore report a 1M-token effective limit while requests are handled by a 262k route.

This affects context validation, token-aware condensation, telemetry, and any code that uses `effective_max_input_tokens` to reserve output headroom.

## Observed failure mode

In a persisted Agent Canvas conversation using:

```json
{
  "model": "openrouter/deepseek/deepseek-v4-flash-0731",
  "litellm_extra_body": {
    "provider": {
      "order": ["CoreWeave", "Baseten"],
      "allow_fallbacks": true
    }
  }
}
```

the final requests repeatedly reported exactly 262,144 total prompt + completion tokens, matching OpenRouter's CoreWeave endpoint limit rather than the model catalog's 1M context. The SDK had no provider-specific metadata with which to identify that mismatch.

The same conversation also exposed two adjacent problems, tracked separately:

- #3903: large tool outputs can consume a substantial fraction of context.
- #3992 / PR #3997: a prose response without a tool call is treated as `FINISHED`, which can terminate an agent after context pressure degrades tool calling.

Those issues address context growth and terminal response semantics. This issue is specifically about resolving the runtime limits of the configured provider route.

## Proposed API

Keep `LLM` as the public façade. Do not add a second user-facing wrapper around LiteLLM.

Add lazy runtime metadata resolution to `LLM`, following the existing synchronous/asynchronous API pattern:

```python
metadata = llm.resolve_runtime_metadata()
metadata = await llm.aresolve_runtime_metadata()
```

The returned model should include values and provenance rather than only a bare token count:

```python
class ModelRuntimeMetadata(BaseModel):
    max_input_tokens: int | None = None
    max_output_tokens: int | None = None
    source: str
    candidate_providers: list[str] = []
    selected_provider: str | None = None
    confidence: Literal["exact", "safe_lower_bound", "model_level", "unknown"]
```

`LLM` should own precedence and merging because it already owns explicit token overrides, LiteLLM model metadata, provider configuration, and the effective token properties. Provider-specific transport/parsing should remain in a private resolver module rather than accumulating in `llm.py`.

Suggested internal boundary:

```python
async def aresolve_provider_metadata(llm: LLM) -> ModelRuntimeMetadata | None:
    ...
```

with provider adapters registered behind that function. This is a compatibility/augmentation layer for metadata LiteLLM does not currently expose; it should be removable when LiteLLM gains equivalent support.

## Resolution precedence

1. Explicit user settings such as `LLM.max_input_tokens` and `LLM.max_output_tokens` always win.
2. Route-specific metadata from the configured provider is used when it can be resolved safely.
3. LiteLLM's model-level metadata remains the fallback.
4. Existing unknown-model behavior remains unchanged when discovery fails.

A property such as `effective_max_input_tokens` must remain synchronous and side-effect free. It must not perform network I/O. Runtime discovery should populate a private cache; the property can then read the cached route-aware value before falling back to existing metadata.

## OpenRouter resolver

For OpenRouter models, query:

```text
GET https://openrouter.ai/api/v1/models/{model_id}/endpoints
```

and combine the endpoint metadata with `litellm_extra_body.provider` routing constraints.

The resolver should:

- Normalize provider names case-insensitively while preserving the provider's display name.
- Respect explicit provider allow/ignore/order constraints that determine eligible routes.
- Return an exact endpoint limit when the configuration pins one endpoint.
- Return a conservative lower bound when multiple endpoints are eligible and the selected endpoint is not known before inference.
- Avoid replacing model-level metadata when route eligibility cannot be interpreted safely.
- Record the actual selected provider separately when OpenRouter exposes it in response metadata; the provider that handled the last request must not silently become the global limit when routing can change on the next request.

The implementation should not hard-code DeepSeek or particular providers.

## Lifecycle, caching, and failure behavior

Network discovery must not happen during `LLM` construction. Existing work demonstrates why:

- #3495: runtime `/v1/model/info` lookup failures silently lose model capabilities.
- #4098 / PR #4132: eager unauthenticated model-info lookups create unnecessary traffic.
- PR #3912: unbounded model-info discovery can hang LLM construction and the agent-server event loop.

Requirements:

- Resolve lazily, before the first request that needs runtime limits or through an explicit call.
- Provide synchronous and asynchronous entry points without running blocking HTTP on the agent-server event loop.
- Use a bounded timeout.
- Cache successful results with a TTL because endpoint metadata changes but should not be fetched per completion.
- Cache failures briefly or otherwise prevent repeated failing requests.
- Fall back to LiteLLM metadata on timeout, non-2xx responses, malformed payloads, or unsupported providers.
- Never log or return credentials.

## LiteLLM relationship

LiteLLM currently imports OpenRouter's model-level catalog in `.github/workflows/auto_update_price_and_context_window_file.py`, mapping `row["context_length"]` into its static model metadata. It does not import provider-endpoint limits.

The OpenHands implementation should be written as a narrow compatibility shim, with an upstream LiteLLM issue/PR opened in parallel. If LiteLLM later exposes route-aware runtime metadata, the OpenRouter adapter should delegate to or be removed in favor of that API without changing the public `LLM` interface.

PR #4154 provides an example of bounded server-side OpenRouter API access, but its credit-balance endpoint is account/product functionality. Billing lookup is explicitly outside this issue; this proposal covers metadata required for correct inference and context management.

## Integration with context management

This issue should make the route-aware limit available to context management. It should not silently invent a universal condensation ratio.

As a follow-up or a small integration in the same PR, the conversation/condenser setup can derive a configured token threshold from resolved metadata when the user opts into automatic token-aware condensation. The threshold and output reserve need an explicit policy because `max_input_tokens` is the hard provider limit, not the desired condensation trigger.

## Tests

Add focused tests for:

- Explicit `max_input_tokens` overriding all discovered values.
- A pinned OpenRouter provider returning its exact context limit.
- Multiple eligible providers returning a conservative, provenance-marked result.
- Provider name normalization and missing configured providers.
- OpenRouter model ID normalization (`openrouter/<id>` versus `<id>` in the API).
- Timeout, HTTP error, malformed response, and unsupported provider fallback.
- TTL caching and concurrent resolution avoiding duplicate upstream calls.
- Async resolution not blocking the event loop.
- Serialization excluding transient cache/client state.
- Existing `effective_max_input_tokens` behavior remaining unchanged before resolution and for non-OpenRouter models.

Use mocked HTTP transport for deterministic provider responses while exercising the real resolver and merge paths.

## Acceptance criteria

- [ ] `LLM` exposes synchronous and asynchronous runtime metadata resolution.
- [ ] Explicit user token limits retain highest precedence.
- [ ] OpenRouter provider routing can produce a route-aware context limit with provenance.
- [ ] `effective_max_input_tokens` can use resolved metadata without network I/O in the property.
- [ ] Discovery is lazy, bounded, cached, and fails back to LiteLLM metadata.
- [ ] The agent-server async path performs no blocking metadata HTTP on its event loop.
- [ ] Tests cover exact, conservative, fallback, caching, and precedence behavior.
- [ ] An upstream LiteLLM issue is linked, or the PR documents why upstream cannot yet support the required API.

## Non-goals

- Credit balance or other account-management APIs.
- Replacing LiteLLM's completion, provider routing, or static model metadata.
- Changing the default content-without-tool-call policy from #3992.
- Choosing a universal condenser threshold or context-utilization percentage.

---
_This issue was created by an AI agent (OpenHands) on behalf of Graham Neubig._

### LINKED ISSUE #4428: feat(llm): resolve provider-specific runtime metadata for routed models
Resolve provider-specific runtime metadata for routed models.

Ref: PR OpenHands/software-agent-sdk#4423

neubig:
HUMAN:
Validated the runtime metadata changes with the listed unit and lint checks.

AGENT:

## Why

The static model catalog (LiteLLM metadata) describes a model's nominal limits, which can overstate the limit of the route a configured provider actually serves. On OpenRouter, `deepseek/deepseek-v4-flash-0731` advertises a 1M-token context, but the CoreWeave endpoint limits requests to 262k tokens. Because LiteLLM imports the model-level catalog value, `LLM.effective_max_input_tokens` reports the wrong context for routed models, feeding the wrong value into context validation, token-aware condensation, and telemetry.

See https://github.com/OpenHands/software-agent-sdk/issues/4421.

## Summary

- Add `LLM.resolve_runtime_metadata()` / `aresolve_runtime_metadata()`: resolve route-aware limits lazily, cache with a TTL (positive and negative), deduplicate concurrent lookups, and fall back to model metadata on any error.
- Add an OpenRouter adapter (`openhands/sdk/llm/utils/providers/openrouter.py`) that queries the per-endpoint catalog and applies `litellm_extra_body.provider` routing semantics (`only`/`ignore`/`order`/`allow_fallbacks`): exact for a pinned route, conservative lower bound when multiple eligible routes differ in context, `None` when routing is not safely interpretable.
- Keep provider transport in a private module so it can be retired when LiteLLM gains equivalent provider-endpoint support.
- `effective_max_input_tokens` now consults the cached route-aware value with no network I/O in the property, preserving explicit-`max_input_tokens` precedence and pre-resolution behavior.
- Export `ModelRuntimeMetadata` from `openhands.sdk.llm`.

## Issue Number

Closes https://github.com/OpenHands/software-agent-sdk/issues/4421

## How to Test

Unit tests use `httpx.MockTransport` against the real resolver/merge paths (real `LLM` construction, caching, precedence, routing semantics):

```
uv run pytest tests/sdk/llm/test_runtime_metadata.py -q          # 21 passed
```

Existing LLM coverage is unchanged (all pass):

```
uv run pytest tests/sdk/llm/test_llm.py tests/sdk/llm/test_llm_serialization.py \
  tests/sdk/llm/test_llm_litellm_extra_body.py tests/sdk/llm/test_model_info_proxy_lookup.py \
  tests/sdk/llm/test_llm_metrics.py tests/sdk/llm/test_llm_telemetry.py -q      # 194 passed
```

Lint/type/import checks via `uv run --project . pre-commit run --files <changed files>`: Ruff format, Ruff lint, pycodestyle, pyright, import-rules, and tool-registration all pass.

A live OpenRouter call is not required: the adapter's HTTP is exercised through `httpx.MockTransport`, and the parsing/routing logic is covered directly.

---
_This pull request was created by an AI agent (OpenHands) on behalf of Graham Neubig._

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:de10540-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-de10540-python \
  ghcr.io/openhands/agent-server:de10540-python
```

```
ghcr.io/openhands/agent-server:de10540-golang-amd64
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-golang-amd64
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-golang-amd64
ghcr.io/openhands/agent-server:de10540-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:de10540-golang-arm64
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-golang-arm64
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-golang-arm64
ghcr.io/openhands/agent-server:de10540-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:de10540-java-amd64
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-java-amd64
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-java-amd64
ghcr.io/openhands/agent-server:de10540-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:de10540-java-arm64
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-java-arm64
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-java-arm64
ghcr.io/openhands/agent-server:de10540-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:de10540-python-amd64
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-python-amd64
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-python-amd64
ghcr.io/openhands/agent-server:de10540-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:de10540-python-arm64
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-python-arm64
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-python-arm64
ghcr.io/openhands/agent-server:de10540-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:de10540-golang
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-golang
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-golang
ghcr.io/openhands/agent-server:de10540-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:de10540-java
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-java
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-java
ghcr.io/openhands/agent-server:de10540-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:de10540-python
ghcr.io/openhands/agent-server:de105405541649a80a8ca87e1ea989d48f16c268-python
ghcr.io/openhands/agent-server:feat-runtime-metadata-openrouter-python
ghcr.io/openhands/agent-server:de10540-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `de10540-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `de10540-python-amd64`) are also available if needed
</details>

Closes #4428

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -29,6 +29,14 @@ from pydantic.json_schema import SkipJsonSchema
 
 from openhands.sdk.llm.fallback_strategy import FallbackStrategy
 from openhands.sdk.llm.utils.model_info import get_litellm_model_info
+from openhands.sdk.llm.utils.runtime_metadata import (
+    ModelRuntimeMetadata,
+    aresolve_provider_metadata,
+    cached_metadata,
+    in_negative_cache,
+    resolve_provider_metadata_sync,
+    store_result,
+)
 from openhands.sdk.settings.metadata import SettingProminence, field_meta
 from openhands.sdk.utils.pydantic_secrets import serialize_secret, validate_secret
 
@@ -625,6 +633,12 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
     _call_context: LLMCallContext = PrivateAttr(default_factory=LLMCallContext)
     _effective_max_input_tokens: int | None = PrivateAttr(default=None)
     _effective_max_output_tokens: int | None = PrivateAttr(default=None)
+    # Provider-aware runtime metadata resolved lazily (see
+    # `utils/runtime_metadata.py`). `fetched_at` and `negative_until` are
+    # monotonic timestamps used for the positive/negative caches.
+    _runtime_metadata: ModelRuntimeMetadata | None = PrivateAttr(default=None)
+    _runtime_metadata_fetched_at: float | None = PrivateAttr(default=None)
+    _runtime_metadata_negative_until: float | None = PrivateAttr(default=None)
     # Plain (non-reentrant) Lock: the async transport path acquires this off
     # the event loop thread (see `_alitellm_modify_params_ctx`) and releases
     # it back on the event loop thread, which an RLock would reject since it
@@ -2617,10 +2631,19 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
     def effective_max_input_tokens(self) -> int | None:
         """Resolved context window used at runtime.
 
-        ``max_input_tokens`` remains the user-configured value. When it is
-        unset, this property reflects the value discovered from model metadata.
+        ``max_input_tokens`` remains the user-configured value and always wins.
+        When it is unset, a previously resolved provider-aware runtime limit is
+        used (this property performs no network I/O), falling back to the value
+        discovered from model metadata.
         """
-        return self.max_input_tokens or self._effective_max_input_tokens
+        if self.max_input_tokens:
+            return self.max_input_tokens
+        cached = cached_metadata(
+            self._runtime_metadata, self._runtime_metadata_fetched_at
+        )
+        if cached is not None and cached.max_input_tokens is not None:
+            return cached.max_input_tokens
+        return self._effective_max_input_tokens
 
     @property
     def effective_max_output_tokens(self) -> int | None:
@@ -2630,6 +2653,71 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         unset, this property reflects provider/model defaults and safety caps.
         """
         return self.max_output_tokens or self._effective_max_output_tokens
+
+    # =========================================================================
+    # Runtime (provider-aware) metadata
+    # =========================================================================
+    def resolve_runtime_metadata(
+        self, *, force: bool = False
+    ) -> ModelRuntimeMetadata | None:
+        """Resolve provider-aware runtime limits synchronously (lazy + cached).
+
+        Discovery never runs during construction. Unsupported providers and
+        unresolvable routes return ``None`` so callers fall back to static
+        model metadata. Successful results are cached for
+        ``RUNTIME_METADATA_TTL_SECONDS``; failures are negative-cached briefly.
+
+        Prefer :meth:`aresolve_runtime_metadata` on the agent-server event loop,
+        which performs no blocking network I/O in-process.
+        """
+        if not force:
+            cached = cached_metadata(
+                self._runtime_metadata, self._runtime_metadata_fetched_at
+            )
+            if cached is not None:
+                return cached
+            if in_negative_cache(self._runtime_metadata_negative_until):
+                return None
+
+        metadata = resolve_provider_metadata_sync(self)
+        self._store_runtime_metadata(metadata)
+        return metadata
+
+    async def aresolve_runtime_metadata(
+        self, *, force: bool = False
+    ) -> ModelRuntimeMetadata | None:
+        """Async variant of :meth:`resolve_runtime_metadata`.
+
+        Concurrent calls for the same route are deduplicated so only one
+        upstream request is issued. If the caller is configuring the condenser
+        or context management, resolve before computing a token threshold.
+        """
+        if not force:
+            cached = cached_metadata(
+                self._runtime_metadata, self._runtime_metadata_fetched_at
+            )
+            if cached is not None:
+                return cached
+            if in_negative_cache(self._runtime_metadata_negative_until):
+                return None
+
+        metadata = await aresolve_provider_metadata(self)
+        self._store_runtime_metadata(metadata)
+        return metadata
+
+    def _store_runtime_metadata(self, metadata: ModelRuntimeMetadata | None) -> None:
+        fetched_at, negative_until, resolved = store_result(metadata)
+        self._runtime_metadata_fetched_at = fetched_at
+        self._runtime_metadata_negative_until = negative_until
+        if resolved is not None:
+            self._runtime_metadata = resolved
+
+    @property
+    def resolved_runtime_metadata(self) -> ModelRuntimeMetadata | None:
+        """Currently cached runtime metadata, without triggering discovery."""
+        return cached_metadata(
+            self._runtime_metadata, self._runtime_metadata_fetched_at
+        )
 
     # =========================================================================
     # Utilities preserved from previous class
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -789,9 +789,8 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         # Pydantic copies private attrs without re-running validators, even for
         # deep copies, so routing-field updates must rebuild derived metadata.
         copied = super().model_copy(update=update, deep=deep)
-        route_changed = (
-            update is not None
-            and any(k in update for k in ("model", "base_url", "litellm_extra_body"))
+        route_changed = update is not None and any(
+            k in update for k in ("model", "base_url", "litellm_extra_body")
         )
         if route_changed:
             copied._refresh_litellm_metadata()
@@ -2556,18 +2555,12 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         """Resolved output token limit used at runtime.
 
         ``max_output_tokens`` remains the user-configured value and always
-        wins. When it is unset, a previously resolved provider-aware runtime
-        limit is used (this property performs no network I/O), falling back to
-        the value discovered from model metadata.
+        wins. This property performs no network I/O and falls back to the value
+        discovered from model metadata. Runtime (provider) resolution only
+        refines the input/context limit, never the output limit, so no call
+        here is needed for output tokens (see issue #4421).
         """
-        if self.max_output_tokens:
-            return self.max_output_tokens
-        cached = cached_metadata(
-            self._runtime_metadata, self._runtime_metadata_fetched_at
-        )
-        if cached is not None and cached.max_output_tokens is not None:
-            return cached.max_output_tokens
-        return self._effective_max_output_tokens
+        return self.max_output_tokens or self._effective_max_output_tokens
 
     # =========================================================================
     # Runtime (provider-aware) metadata
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -639,6 +639,12 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
     _runtime_metadata: ModelRuntimeMetadata | None = PrivateAttr(default=None)
     _runtime_metadata_fetched_at: float | None = PrivateAttr(default=None)
     _runtime_metadata_negative_until: float | None = PrivateAttr(default=None)
+    # Guards the runtime-metadata cache fields above. The synchronous resolver
+    # may be driven from a worker thread, and the async resolver can also store
+    # a result, so the read/check and store are kept atomic. ClassVar (shared)
+    # matches the `_litellm_modify_params_lock` pattern; critical sections are
+    # tiny and never cover network I/O.
+    _runtime_metadata_lock: ClassVar[threading.Lock] = threading.Lock()
     # Plain (non-reentrant) Lock: the async transport path acquires this off
     # the event loop thread (see `_alitellm_modify_params_ctx`) and releases
     # it back on the event loop thread, which an RLock would reject since it
@@ -1605,6 +1611,14 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         Uses ``litellm.acompletion`` under the hood, freeing the event loop
         while waiting for the LLM provider response.
         """
+        # Resolve provider-aware runtime metadata (e.g. OpenRouter route limits)
+        # before the first completion on the agent-server async path so
+        # ``effective_max_input_tokens`` reflects the actual runtime route.
+        # The result is cached (1h TTL) and negative-cached on failure, and the
+        # probe is a no-op for providers without runtime metadata, so repeated
+        # calls are cheap.
+        await self.aresolve_runtime_metadata()
+
         _caller_kwargs = kwargs.copy()
         enable_streaming = bool(kwargs.get("stream", False)) or self.stream
         if enable_streaming and on_token is None:
@@ -2670,17 +2684,22 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
         Prefer :meth:`aresolve_runtime_metadata` on the agent-server event loop,
         which performs no blocking network I/O in-process.
         """
-        if not force:
-            cached = cached_metadata(
-                self._runtime_metadata, self._runtime_metadata_fetched_at
-            )
-            if cached is not None:
-                return cached
-            if in_negative_cache(self._runtime_metadata_negative_until):
-                return None
+        # The LLM may be shared across threads via the sync path, so the cache
+        # read/check and the store must be atomic. The network probe itself runs
+        # outside the lock so a slow endpoint cannot block other threads from
+        # reading the cache.
+        with self._runtime_metadata_lock:
+            if not force:
+                cached = cached_metadata(
+                    self._runtime_metadata, self._runtime_metadata_fetched_at
+                )
+                if cached is not None:
+                    return cached
+                if in_negative_cache(self._runtime_metadata_negative_until):
+                    return None
 
         metadata = resolve_provider_metadata_sync(self)
-        self._store_runtime_metadata(metadata)
+        self._store_runtime_metadata(metadata)  # locks internally
         return metadata
 
     async def aresolve_runtime_metadata(
@@ -2707,10 +2726,13 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
 
     def _store_runtime_metadata(self, metadata: ModelRuntimeMetadata | None) -> None:
         fetched_at, negative_until, resolved = store_result(metadata)
-        self._runtime_metadata_fetched_at = fetched_at
-        self._runtime_metadata_negative_until = negative_until
-        if resolved is not None:
-            self._runtime_metadata = resolved
+        # Locked so a concurrent synchronous resolver (its probe runs outside the
+        # lock) cannot observe a torn / interleaved cache state.
+        with self._runtime_metadata_lock:
+            self._runtime_metadata_fetched_at = fetched_at
+            self._runtime_metadata_negative_until = negative_until
+            if resolved is not None:
+                self._runtime_metadata = resolved
 
     @property
     def resolved_runtime_metadata(self) -> ModelRuntimeMetadata | None:
--- a/openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
+++ b/openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
@@ -166,13 +166,16 @@ def parse_openrouter_payload(
         )
 
     # Multiple eligible routes with differing limits: routing may pick any of
-    # them, so report a conservative lower bound for preflight safety.
+    # them (e.g. when fallbacks are allowed), so the router provider is not
+    # known ahead of time. Report a conservative lower bound for preflight
+    # safety and leave ``selected_provider`` unset rather than guessing the
+    # first route in the (optional) ordering.
     return ModelRuntimeMetadata(
         max_input_tokens=min(contexts),
         max_output_tokens=min(completion_limits) if completion_limits else None,
         source="openrouter_endpoints_api",
         candidate_providers=provider_names,
-        selected_provider=reordered[0].get("provider_name"),
+        selected_provider=None,
         confidence="safe_lower_bound",
     )
 
--- a/openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
+++ b/openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
@@ -1,0 +1,237 @@
+"""OpenRouter runtime metadata adapter.
+
+Queries OpenRouter's per-endpoint catalog to determine the context/output
+limits of the routes an ``LLM`` is actually configured to use, rather than the
+model-level catalog value (which can overstate an endpoint's context).
+
+See: https://github.com/OpenHands/software-agent-sdk/issues/4421
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import httpx
+
+from openhands.sdk.llm.utils.runtime_metadata import (
+    RUNTIME_METADATA_TIMEOUT_SECONDS,
+    ModelRuntimeMetadata,
+)
+from openhands.sdk.logger import get_logger
+
+
+if TYPE_CHECKING:
+    from openhands.sdk.llm.llm import LLM
+
+logger = get_logger(__name__)
+
+OPENROUTER_ENDPOINTS_BASE = "https://openrouter.ai/api/v1/models"
+
+
+def _openrouter_model_id(llm: LLM) -> str | None:
+    """Return the OpenRouter model id for the endpoints API, or None.
+
+    Prefers the ``openrouter/<id>`` prefix form. When the base URL identifies
+    OpenRouter and the model already looks like an OpenRouter id
+    (``provider/model``), it is used as-is.
+    """
+    model = llm.model or ""
+    base_url = llm.base_url or ""
+    if model.startswith("openrouter/"):
+        return model[len("openrouter/") :]
+    if "openrouter.ai" in base_url and "/" in model:
+        return model
+    return None
+
+
+def _normalize(name: str | None) -> str:
+    return (name or "").strip().casefold()
+
+
+def _provider_routing(llm: LLM) -> dict[str, Any]:
+    extra_body = llm.litellm_extra_body or {}
+    provider = extra_body.get("provider")
+    return provider if isinstance(provider, dict) else {}
+
+
+def _eligible_endpoints(
+    endpoints: list[dict[str, Any]], routing: dict[str, Any]
+) -> list[dict[str, Any]]:
+    """Filter the endpoint list to the routes the routing config permits.
+
+    Order of precedence matches OpenRouter's provider config semantics: an
+    explicit ``only`` restricts candidates, ``ignore`` removes them, then
+    ``order`` (when no ``only``) defines preference. When ``allow_fallbacks``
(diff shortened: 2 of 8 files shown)


neubig:
Pushed a65af7375e92a8d8987081aa704cf76de5a (a8af7375e) which resets the head SHA and clears the two stale `Validate PR description` check runs that were failing on the previous head. Full CI is now green on a8af7375e (33 passing, 0 failed). The PR description's `HUMAN:` section was already updated to satisfy the validator. Requesting re-review.

_This comment was created by an AI agent (OpenHands) on behalf of the automating user._

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -2617,10 +2631,19 @@ def model_info(self) -> dict | None:
...
         """
-        return self.max_input_tokens or self._effective_max_input_tokens
+        if self.max_input_tokens:
+            return self.max_input_tokens
+        cached = cached_metadata(
+            self._runtime_metadata, self._runtime_metadata_fetched_at
+        )
+        if cached is not None and cached.max_input_tokens is not None:
+            return cached.max_input_tokens
+        return self._effective_max_input_tokens
```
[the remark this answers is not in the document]
neubig: Addressed in ccb228789: `acompletion` now awaits `aresolve_runtime_metadata()` before the first completion, so the `_runtime_metadata` cache is populated on the agent-server async path and `effective_max_input_tokens` returns runtime (not model-level) metadata, fixing #4421. Covered by a new `test_acompletion_resolves_runtime_metadata` test. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -2631,6 +2654,71 @@ def effective_max_output_tokens(self) -> int | None:
...
+            cached = cached_metadata(
+                self._runtime_metadata, self._runtime_metadata_fetched_at
+            )
+            if cached is not None:
+                return cached
+            if in_negative_cache(self._runtime_metadata_negative_until):
+                return None
+
+        metadata = resolve_provider_metadata_sync(self)
+        self._store_runtime_metadata(metadata)
```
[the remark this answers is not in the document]
neubig: Addressed in ccb228789: `resolve_runtime_metadata` (sync path) now guards the cache read/check with a `_runtime_metadata_lock`, and `_store_runtime_metadata` stores atomically under the same lock. The network probe runs outside the lock so a slow endpoint cannot block cache readers on the shared LLM instance. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
```diff
@@ -0,0 +1,237 @@
...
+        )
+
+    # Multiple eligible routes with differing limits: routing may pick any of
+    # them, so report a conservative lower bound for preflight safety.
+    return ModelRuntimeMetadata(
+        max_input_tokens=min(contexts),
+        max_output_tokens=min(completion_limits) if completion_limits else None,
+        source="openrouter_endpoints_api",
+        candidate_providers=provider_names,
+        selected_provider=reordered[0].get("provider_name"),
```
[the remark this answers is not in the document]
neubig: Addressed in ccb228789: `parse_openrouter_payload` now sets `selected_provider=None` for the `safe_lower_bound` case (allow_fallbacks / no explicit routing), and the `safe_lower_bound` tests assert `selected_provider is None`. _This reply was created by an AI agent (OpenHands) on behalf of the user._

neubig:
cc @all-hands-bot @openhands — pushed ccb228789 resolving all three review threads (runtime metadata wiring, sync-path lock, safe_lower_bound provider). Please re-review. _This comment was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -1591,6 +1611,14 @@ async def acompletion(
         Uses ``litellm.acompletion`` under the hood, freeing the event loop
         while waiting for the LLM provider response.
         """
+        # Resolve provider-aware runtime metadata (e.g. OpenRouter route limits)
+        # before the first completion on the agent-server async path so
+        # ``effective_max_input_tokens`` reflects the actual runtime route.
+        # The result is cached (1h TTL) and negative-cached on failure, and the
+        # probe is a no-op for providers without runtime metadata, so repeated
+        # calls are cheap.
+        await self.aresolve_runtime_metadata()
```
neubig, line 1634: 🟠 **Important:** This resolves metadata after the first context-management decision and only on async Chat Completions. `Agent.astep()` calls `aprepare_llm_messages(..., condenser=...)` before `amake_llm_completion()`, so a resumed large conversation can skip required condensation using the static 1M limit, then discover the 262k route only while preparing the request. `amake_llm_completion()` also dispatches Responses-capable models to `aresponses()`, which does not resolve metadata; the sync `completion()`/`responses()` paths do not either. Resolve before message/condenser preparation in both `step()` and `astep()` (or through a shared preflight used there), and make all direct invocation paths consistent. Please add a test proving the condenser sees the route-aware limit on the first step plus coverage for Responses and sync paths.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -2631,6 +2668,79 @@ def effective_max_output_tokens(self) -> int | None:
...
+        ``RUNTIME_METADATA_TTL_SECONDS``; failures are negative-cached briefly.
+
+        Prefer :meth:`aresolve_runtime_metadata` on the agent-server event loop,
+        which performs no blocking network I/O in-process.
+        """
+        # The LLM may be shared across threads via the sync path, so the cache
+        # read/check and the store must be atomic. The network probe itself runs
+        # outside the lock so a slow endpoint cannot block other threads from
+        # reading the cache.
+        with self._runtime_metadata_lock:
```
neubig, line 2588: 🟠 **Important:** This lock does not make the sync resolver thread-safe in the behavior that matters. Every cache-miss thread releases it before the network probe, so requests are not deduplicated. More seriously, an older failed probe can finish after a newer successful probe; `_store_runtime_metadata(None)` then clears `fetched_at` and installs a five-minute negative cache, hiding the successful metadata. Use a per-instance/per-route single-flight or generation check so stale results cannot overwrite newer ones, and add a threaded test with controlled success/failure ordering that asserts one request and a retained successful cache.

neubig (commented):
🔴 **Needs improvement**

The `selected_provider=None` fix is correct, and the focused suite passes locally (`23 passed`; changed-file pre-commit checks pass). However, the runtime metadata is still resolved too late/in too few invocation paths, the sync lock is not single-flight, and several route/cache/error cases can produce incorrect limits or abort inference. The latest head is also conflicting with `main` and currently has only the PR Artifacts check, not full CI.

Because this changes context-budget behavior, it is eval-risk under this repository's review policy. After the correctness issues are fixed, a human maintainer should make the merge decision after lightweight eval evidence. The PR's AGENT evidence currently lists unit/lint checks only; please add a real end-to-end runtime command/output as requested by the template.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟡 MEDIUM — incorrect route metadata can suppress condensation, send invalid token budgets, or prevent the actual LLM request.

**VERDICT:**
❌ **Needs rework**: Material correctness and concurrency issues remain.

**KEY INSIGHT:**
Runtime metadata must be bound to the exact LLM route and resolved before every consumer that makes a context-budget decision.

_This review was created by an AI agent (OpenHands) on behalf of the repository maintainers._

>
> 1. Add a `.agents/skills/custom-codereview-guide.md` file to your branch (or edit it if one already exists) with the `/codereview` trigger and the context the reviewer is missing. See the [customization docs](https://docs.openhands.dev/openhands/usage/use-cases/code-review#customization).
> 3. When merged, the guideline goes through normal maintainer review.
>
> **Resolve with AI?** Install the [iterate skill](https://github.com/OpenHands/extensions/tree/main/skills/iterate) and run `/iterate` to drive CI, review, and QA.
>

openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
```diff
@@ -0,0 +1,240 @@
...
+    order = routing.get("order") or []
+
+    def by_name(name: str) -> dict[str, Any] | None:
+        key = _normalize(name)
+        for e in endpoints:
+            if _normalize(e.get("provider_name")) == key:
+                return e
+        return None
+
+    if only:
```
neubig: 🟠 **Important:** The routing constraints are composed in the wrong order. OpenRouter documents `only`/`ignore` as eligibility filters and `order` as the ranking of eligible providers: https://openrouter.ai/docs/guides/routing/provider-selection. With `order=[A, B]`, `ignore=[A]`, and `allow_fallbacks=false`, this code pins A first and then removes it, returning no metadata instead of selecting B as the first surviving ordered provider. When `only` is present, `order` is skipped entirely. Apply `only` and `ignore` first, then apply `order`/fallback behavior to the survivors, and cover the combinations in a routing matrix.

openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -625,6 +633,18 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
     _call_context: LLMCallContext = PrivateAttr(default_factory=LLMCallContext)
     _effective_max_input_tokens: int | None = PrivateAttr(default=None)
     _effective_max_output_tokens: int | None = PrivateAttr(default=None)
+    # Provider-aware runtime metadata resolved lazily (see
```
neubig, line 646: 🟠 **Important:** These private cache fields are copied by Pydantic's `model_copy()`, but `LLM.model_copy()` never clears them. After resolving route A, `model_copy(update={"model": ..., "litellm_extra_body": route_b})` continues returning route A's fresh limit for up to an hour; a stale value below 16k can even make `_refresh_litellm_metadata()` reject the copied model during validation. Bind the cached result to `cache_key(self)` and validate it on reads, or reset all runtime cache fields whenever `model`, `base_url`, or routing configuration changes. Add regression tests for model and route-only copies.

openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
```diff
@@ -0,0 +1,240 @@
...
+) -> ModelRuntimeMetadata | None:
+    model_id = _openrouter_model_id(llm)
+    if not model_id:
+        return None
+    payload = await _fetch_async(
+        f"{OPENROUTER_ENDPOINTS_BASE}/{model_id}/endpoints", http_client
+    )
+    if payload is None:
+        return None
+    return parse_openrouter_payload(payload, _provider_routing(llm))
```
neubig, line 230: 🟠 **Important:** A malformed successful response escapes the promised fallback boundary. `_fetch_async()` catches transport/JSON errors, but parsing happens afterward; for example `provider_name: 7` raises `AttributeError` in `_normalize()`. Since `acompletion()` awaits resolution outside its completion error handler, the real LLM request is never attempted. Catch resolver/parser exceptions at the runtime-metadata dispatch boundary and return `None` (while allowing cancellation to propagate), then test that malformed endpoint data still reaches the mocked completion transport.

openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
```diff
@@ -0,0 +1,240 @@
...
+        key=lambda e: list(routing_providers).index(_normalize(e.get("provider_name")))
+        if _normalize(e.get("provider_name")) in routing_providers
+        else len(routing_providers),
+    )
+
+    if len(eligible) == 1 or len(set(contexts)) == 1:
+        target = eligible[0]
+        return ModelRuntimeMetadata(
+            max_input_tokens=target.get("context_length"),
+            max_output_tokens=target.get("max_completion_tokens"),
```
neubig: 🟠 **Important:** `max_output_tokens` is resolved but never used: `LLM.effective_max_output_tokens` still returns only the explicit/static value, so Chat/Responses options can request more output than the selected endpoint allows. Before wiring it in, this branch also needs conservative semantics: equal context lengths do not imply equal output limits, and the lower-bound branch currently takes the minimum of only non-null values, which is unsafe if another eligible endpoint has an unknown limit. Either implement explicit → runtime → static precedence with `None` unless every eligible route has a known output limit, or remove this field until it can be applied safely. Add exact/differing/nullable/explicit-override tests.

openhands-sdk/openhands/sdk/llm/llm.py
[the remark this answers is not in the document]
neubig: Addressed in febd497f8: `Agent.step()` and `Agent.astep()` now establish route-aware metadata via `resolve_runtime_metadata()`/`aresolve_runtime_metadata()` at a shared preflight *before* `prepare_llm_messages`/`aprepare_llm_messages`, so the condenser sees the route-aware context window on the very first step. Because the Chat and Responses dispatch paths route through this same boundary, sync, async, and Responses are all covered. Regression tests `test_step_resolves_runtime_metadata_before_condensation` and `test_astep_resolves_runtime_metadata_before_condensation` assert resolution precedes condensation. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/utils/runtime_metadata.py
```diff
@@ -0,0 +1,209 @@
...
+    resolver = detect_provider(llm)
+    if resolver is None:
+        return None
+
+    key = cache_key(llm)
+    pending = _inflight.get(key)
+    if pending is not None and not pending.done():
+        # A concurrent resolution for the same route is already in flight; await
+        # it rather than issuing a duplicate upstream request.
+        return await pending
```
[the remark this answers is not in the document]
neubig: Addressed in febd497f8: in-flight resolution tasks are now keyed by event loop (`inflight_key = (asyncio.get_running_loop(), key)`), so two threads each running their own loop never await a task attached to another loop (`RuntimeError: got Future attached to a different loop`). Covered by `test_async_inflight_keyed_by_event_loop`. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/utils/runtime_metadata.py
```diff
@@ -0,0 +1,209 @@
...
+class ModelRuntimeMetadata(BaseModel):
+    """Route-aware limits resolved for a single ``LLM`` configuration.
+
+    Includes provenance so callers can reason about whether the value is the
+    exact limit of a pinned route, a conservative bound across eligible routes,
+    or a fallback to model-level metadata.
+    """
+
+    max_input_tokens: int | None = None
+    max_output_tokens: int | None = None
```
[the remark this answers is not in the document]
neubig: Addressed in febd497f8: `max_output_tokens` is removed from `ModelRuntimeMetadata` and from the OpenRouter adapter, so the public result no longer carries an unused/unsafe limit. `effective_max_output_tokens` keeps explicit-config-over-model-metadata precedence; input/context limits remain the refined runtime value. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/llm.py
neubig: Addressed in febd497f8: `Agent.step()`/`Agent.astep()` resolve route metadata at a shared preflight before message/condenser preparation, covering sync, async, and the Responses dispatch paths. Regression tests prove first-step condensation uses the route-aware limit. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/llm.py
neubig: Addressed in febd497f8: the sync resolver now keeps a per-instance `_runtime_metadata_generation`, bumped on each started lookup, and `_store_runtime_metadata` only publishes a probe's result while its generation is still current — so a stale (earlier-started) failure can never overwrite a newer success (single-flight by generation). Cache read/check/store are atomic under `_runtime_metadata_lock`; the network probe runs outside the lock. Covered by the threaded `test_sync_resolution_generation_guards_stale_write` with controlled success/failure ordering asserting a retained successful cache. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
neubig: Addressed in febd497f8: `_eligible_endpoints` now applies `only`/`ignore` as *eligibility* filters first, then applies `order` (with `allow_fallbacks`) to the survivors; `only` overrides `order`. With `order=[A,B]`, `ignore=[A]`, `allow_fallbacks=false`, B is selected as the first surviving ordered provider. Covered by the routing matrix (`test_parse_pinned_order_returns_exact`, `test_parse_allow_fallbacks_returns_safe_lower_bound`, `test_parse_only_matches_returns_exact`, `test_parse_ignore_removes_routes`, case-insensitivity, unmatched-order). _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/llm.py
neubig: Addressed in febd497f8: `LLM.model_copy()` calls `_reset_runtime_metadata_for_key()` whenever `model`/`base_url`/`litellm_extra_body` changes, dropping the cached runtime metadata and negative-cache state so a stale route limit can never leak onto a copied model (including the sub-16k `_refresh_litellm_metadata` rejection path). Covered by `test_model_copy_resets_cache_on_route_change` for a route-only copy. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
neubig: Addressed in febd497f8: resolver/parser exceptions (e.g. `provider_name: 7` raising `AttributeError` in `_normalize`) are now caught at the runtime-metadata dispatch boundary by `_safe_resolve`, which logs and returns `None` (cancellation still propagates), so the real LLM request is always attempted. `_normalize` is also hardened against non-string names. Covered by `test_parse_malformed_payload_returns_none` and the fallback-on-malformed dispatch path. _This reply was created by an AI agent (OpenHands) on behalf of the user._

openhands-sdk/openhands/sdk/llm/utils/providers/openrouter.py
neubig: Addressed in febd497f8: `max_output_tokens` is removed from `ModelRuntimeMetadata` and the OpenRouter adapter (the 'remove' option) rather than wiring an unused/unsafe output limit. Explicit-config precedence is preserved; output limits are no longer promised by the public result. _This reply was created by an AI agent (OpenHands) on behalf of the user._

PR #4426, opened 2026-08-08
fix(agent-server): initialize observability after deferred env

Shimada666:
HUMAN:
We use deferred-init agent servers and need runtime observability configuration to take effect before conversations start.

---

AGENT:

## Why

In deferred-init mode, the agent server imports the SDK before `POST /api/init` supplies runtime environment variables. Observability initialization therefore runs before `LMNR_*` or generic OTEL variables exist, and later conversations produce non-recording spans even though the init request accepted the variables.

## Summary

- Call the existing idempotent `maybe_init_laminar()` after deferred-init environment variables are applied and before conversation services start.
- Extend the deferred-init environment test to verify observability sees the injected values.
- Leave the normal startup path and public REST contract unchanged.

## Issue Number

None.

## How to Test

```bash
uv run pytest tests/sdk/observability/test_laminar.py tests/agent_server/test_init_router.py -q
uv run pre-commit run --files openhands-agent-server/openhands/agent_server/init_router.py tests/agent_server/test_init_router.py
```

Result: 57 tests passed; all targeted pre-commit checks passed.

I also exercised the real FastAPI application through `TestClient`: started it with `deferred_init=True` and no observability environment, then posted `LMNR_PROJECT_API_KEY`, `LMNR_BASE_URL`, `LMNR_HTTP_PORT`, and `LMNR_FORCE_HTTP` through `/api/init`.

```text
before_init=False
init_status=200
after_init=True
```

Before this change, the regression test fails because deferred init never invokes the observability initializer.

## Video/Screenshots

Not applicable; server lifecycle change only.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The change adds no Laminar-specific request field or workspace abstraction. `maybe_init_laminar()` already handles both Laminar and generic OTEL configuration and is idempotent when observability was initialized during normal startup.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/init_router.py
+++ b/openhands-agent-server/openhands/agent_server/init_router.py
@@ -31,6 +31,7 @@ from openhands.agent_server.telemetry import (
     shutdown_telemetry_sink,
 )
 from openhands.sdk.logger import get_logger
+from openhands.sdk.observability import maybe_init_laminar
 
 
 logger = get_logger(__name__)
@@ -226,6 +227,7 @@ class InitService:
                 # tools pick up credentials.
                 for key, value in req.env.items():
                     os.environ[key] = value
+            maybe_init_laminar()
 
             # Must precede get_instance(), which captures the sink. The
             # matching emit_server_started() is deferred until the ``ready``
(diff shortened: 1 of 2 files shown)


enyst (approved):
Thank you!

PR #4438, opened 2026-08-09
fix(sdk): prevent double provider-prefix strip for namespaced model IDs

### LINKED ISSUE #4613: [Bug]: OpenAI provider prefix stripped twice for namespaced model IDs in Responses API
### Bug Description

`LLMProvider.from_model()` parses a model string like `openai/openai/example-model` into `model="openai/example-model"` + `provider="openai"`. When `as_litellm_call_kwargs()` forwards that as `model="openai/example-model"` + `custom_llm_provider="openai"`, LiteLLM's Responses API path re-parses and strips the `openai/` prefix again, sending `example-model` upstream and causing `400 Model not found`.

This also affects any namespaced model ID whose inner segment is a LiteLLM provider name (e.g. `openai/openai/o3-mini`, `mistral/mistral/large`).

Upstream report: OpenHands/OpenHands#16365

### Expected Behavior

The SDK should pass a model string that, after LiteLLM re-parses it, resolves to the originally parsed `openai/example-model`.

### Actual Behavior

Before the fix, running the repro path through `as_litellm_call_kwargs()` sends `example-model` to the upstream endpoint. The upstream returns `400 Model not found` because the model identifier has lost its provider namespace.

```bash
uv run pytest tests/sdk/llm/test_litellm_provider.py::test_llm_provider_preserves_namespaced_model_after_litellm_reparse -v
```

The regression test above (introduced in the fix) fails on `main` and passes after the change.

### Acceptance Criteria

- [ ] `LLMProvider.as_litellm_call_kwargs()` re-adds the provider prefix when `self.model` already starts with it, so LiteLLM strips exactly one prefix and lands on the correct parsed model.
- [ ] `uv run pytest tests/sdk/llm/test_litellm_provider.py -v` passes.
- [ ] Existing nested-provider cases (openrouter, bedrock, mistral, litellm_proxy) continue to pass.

### LINKED ISSUE #16365: [Bug]: OpenAI provider prefix is stripped twice for namespaced model IDs in Responses API
### Operating System

macOS

### Installation Method

npm (`npx @openhands/agent-canvas` or `npm install -g @openhands/agent-canvas`)

### Agent Canvas Version

1.10.0

### Bug Description

OpenHands SDK strips the provider prefix twice when using an OpenAI-compatible endpoint whose model identifier contains its own `openai/` namespace.

For example, suppose the upstream endpoint expects the following model identifier:

```text
openai/example-model
```

Following LiteLLM’s OpenAI-compatible endpoint convention, the OpenHands model is configured as:

```text
openai/openai/example-model
```

However, OpenHands sends this model identifier to the upstream endpoint:

```text
example-model
```

As a result, the endpoint returns a `400 Model not found` error.

### Environment

- Agent Canvas: `1.10.0`
- OpenHands Agent Server / SDK: `1.40.1`
- LiteLLM: `1.95.0`
- Python: `3.13`
- API mode: `responses`
- Custom OpenAI-compatible `base_url`

### Root cause

This appears to be caused by provider parsing occurring at two different layers.

First, `LLMProvider.from_model()` calls LiteLLM’s `get_llm_provider()`:

```text
openai/openai/example-model
→ provider: openai
→ parsed model: openai/example-model
```

OpenHands then passes the parsed model together with:

```text
custom_llm_provider: openai
```

The LiteLLM Responses API path resolves the provider again:

```text
openai/example-model
→ provider: openai
→ parsed model: example-model
```

Therefore, the `openai/` prefix that belongs to the upstream model identifier is removed as though it were another LiteLLM provider prefix.

This behavior appears related to the provider abstraction introduced in [[PR #2363](https://github.com/OpenHands/software-agent-sdk/pull/2363)](https://github.com/OpenHands/software-agent-sdk/pull/2363).

### Current workaround

Adding a third prefix preserves the correct upstream identifier:

```text
openai/openai/openai/example-model
```

The resulting transformations are:

```text
Configured value:
openai/openai/openai/example-model

After OpenHands parsing:
openai/openai/example-model

Model sent upstream:
openai/example-model
```

This workaround is functional but should not be required from users.

### Suggested fix

Keep the original requested model identifier separately from the parsed provider metadata.

OpenHands could use the parsed provider information for internal capability, authentication, and telemetry logic while passing the original model string to LiteLLM for transport-level resolution.

An end-to-end regression test should inspect the actual request body sent to a mock OpenAI-compatible `/responses` endpoint. Testing `get_llm_provider()` results alone does not catch the second normalization performed inside the Responses API path.

### Steps to Reproduce

1. Configure an OpenAI-compatible endpoint that expects a namespaced model identifier:

   ```text
   openai/example-model
   ```

2. Configure the OpenHands LLM profile with:

   ```text
   model: openai/openai/example-model
   base_url: https://example.com/v1
   api_mode: responses
   ```

3. Start a conversation.

4. Inspect the model value received by the upstream `/responses` endpoint.

### Actual Behavior

The upstream request contains:

```json
{
  "model": "example-model"
}
```

The endpoint consequently returns an error similar to:

```text
Model 'example-model' not found
```

### Expected Behavior

The upstream request should contain:

```json
{
  "model": "openai/example-model"
}
```

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

santhiprakash:
HUMAN:

Santhi Prakash — small SDK fix for namespaced OpenAI model IDs; see the AGENT section for the reproduction, regression test, and pre-commit verification.

---

AGENT:

## Why

`LLMProvider.from_model()` parses a model string like `openai/openai/example-model` into `model="openai/example-model"` + `provider="openai"`. When `as_litellm_call_kwargs()` forwards that as `model="openai/example-model"` + `custom_llm_provider="openai"`, LiteLLM's Responses API path re-parses and strips the `openai/` prefix again, sending `example-model` upstream instead of `openai/example-model`, which returns `400 Model not found`.

This affects any namespaced model ID where the inner segment is also a LiteLLM provider name (e.g. `openai/openai/o3-mini`, `mistral/mistral/large`). The OpenRouter nested case (`openrouter/anthropic/claude-sonnet-4`) is unaffected because LiteLLM has a dedicated OpenRouter special-case that preserves the inner `provider/model` form.

## Summary

- Re-prefix the model in `as_litellm_call_kwargs()` when the parsed model still starts with the provider prefix, so LiteLLM strips exactly one prefix and lands on the correct parsed model.
- Added regression test `test_llm_provider_preserves_namespaced_model_after_litellm_reparse` covering `openai/openai/example-model`.

## Issue Number

Fixes #4613

Upstream report: https://github.com/OpenHands/OpenHands/issues/16365

## How to Test

```bash
uv run pytest tests/sdk/llm/test_litellm_provider.py -v
```

Result (12 passed):

```
tests/sdk/llm/test_litellm_provider.py::test_llm_provider_parses_nested_openrouter_model PASSED
tests/sdk/llm/test_litellm_provider.py::test_llm_provider_parses_bedrock_model PASSED
tests/sdk/llm/test_litellm_provider.py::test_llm_provider_strips_api_key_for_bedrock_calls PASSED
tests/sdk/llm/test_litellm_provider.py::test_llm_provider_handles_unknown_model_without_provider PASSED
tests/sdk/llm/test_litellm_provider.py::test_llm_provider_keeps_requested_api_base_verbatim PASSED
tests/sdk/llm/test_litellm_provider.py::test_llm_provider_preserves_namespaced_model_after_litellm_reparse PASSED
tests/sdk/llm/test_litellm_provider.py::test_split_kwargs_equivalent_to_full_model_string[...] PASSED (6 cases)
============================== 12 passed in 0.55s ==============================
```

Pre-commit verification:

```bash
uv run pre-commit run --files openhands-sdk/openhands/sdk/llm/utils/litellm_provider.py tests/sdk/llm/test_litellm_provider.py
```

All checks passed (Ruff format, Ruff lint, PEP8, pyright, import dependency rules, tool subclass registration).

Reproduction proof (before fix vs after fix):

```python
import litellm
# Before: double-strip
litellm.get_llm_provider(model="openai/example-model", custom_llm_provider="openai")
# -> ("example-model", "openai")  wrong, 400 Model not found

# After: re-prefix prevents double-strip
litellm.get_llm_provider(model="openai/openai/example-model", custom_llm_provider="openai")
# -> ("openai/example-model", "openai")  correct
```

## Video/Screenshots

N/A — backend correctness fix, no visual surface.

## Design Doc

Not added — this is a small one-line correctness fix with a regression test, not a new API or behavior change in the agent loop.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The fix is minimal and only triggers when `self.model.startswith(f"{self.name}/")` — the exact condition that causes the double-strip. All existing test cases (openrouter nested, bedrock, mistral, litellm_proxy, openai local, unknown) are unaffected because their parsed models do not start with the provider prefix.

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/utils/litellm_provider.py
+++ b/openhands-sdk/openhands/sdk/llm/utils/litellm_provider.py
@@ -73,7 +73,16 @@ class LLMProvider:
         return api_key
 
     def as_litellm_call_kwargs(self, *, api_key: str | None = None) -> dict[str, str]:
-        kwargs = {"model": self.model}
+        # When the parsed model still starts with the provider prefix (e.g.
+        # "openai/example-model" for provider "openai"), LiteLLM re-parses and
+        # strips the prefix again on the Responses API path, sending the wrong
+        # model upstream (e.g. "example-model" instead of "openai/example-
+        # model"). Re-prefix so LiteLLM strips exactly one prefix and lands on
+        # the parsed model.
+        model = self.model
+        if self.name is not None and model.startswith(f"{self.name}/"):
+            model = f"{self.name}/{model}"
+        kwargs = {"model": model}
         if self.name is not None:
             kwargs["custom_llm_provider"] = self.name
         normalized_api_key = self.api_key_for_litellm(api_key)
--- a/tests/sdk/llm/test_litellm_provider.py
+++ b/tests/sdk/llm/test_litellm_provider.py
@@ -60,6 +60,34 @@ def test_llm_provider_keeps_requested_api_base_verbatim():
     assert provider.api_base == "https://myproxy.example.com"
 
 
+def test_llm_provider_preserves_namespaced_model_after_litellm_reparse():
+    # LiteLLM strips the first "openai/" prefix from "openai/openai/example-
+    # model", leaving parsed model "openai/example-model" with provider
+    # "openai". Without re-prefixing, LiteLLM's Responses API path strips the
+    # "openai/" prefix again and sends "example-model" upstream (400 Model not
+    # found). The kwargs from as_litellm_call_kwargs must resolve back to the
+    # parsed model, not the double-stripped one.
+    provider = LLMProvider.from_model(
+        model="openai/openai/example-model",
+        api_base="https://myproxy.example.com/v1",
+    )
+
+    assert provider.name == "openai"
+    assert provider.model == "openai/example-model"
+
+    kwargs = provider.as_litellm_call_kwargs()
+    assert kwargs["custom_llm_provider"] == "openai"
+
+    resolved = litellm.get_llm_provider(
+        model=kwargs["model"],
+        custom_llm_provider=kwargs["custom_llm_provider"],
+        api_base=provider.api_base,
+        api_key=None,
+    )
+    assert resolved[0] == "openai/example-model"
+    assert resolved[1] == "openai"
+
+
 @pytest.mark.parametrize(
     ("model", "api_base"),
     [


santhiprakash:
Quick note on CI status: every workflow on this PR is in `action_required` (the fork-PR workflow-approval gate), so the PR shows **zero check runs** rather than failing ones.

- The head is current: single commit from Aug 24, `mergeable` against `main`, and `main` has not touched `llm/utils/litellm_provider.py` since.
- The description carries the local verification: the focused suite `tests/sdk/llm/test_litellm_provider.py` passes 12/12 (including the `test_llm_provider_preserves_namespaced_model_after_litellm_reparse` regression), plus pre-commit.
- Issue #4613 is still open with no other PR against it.

If a maintainer can approve running the workflows, standard CI will pick up the same tests. Happy to rebase or provide more reproduction detail meanwhile.

santhiprakash:
Closing to keep the review queue clean — this has been open a while without traction. Happy to reopen or resubmit a fresh version if it would still be useful.

Issue #1343, opened 2025-12-06
[bug] Does not set `send_reasoning_content` in `from_llm_chat_message`


Yikai-Liao:
In `openhands/sdk/llm/message.py`, funciton `send_reasoning_content` set reasoning content, but does not set `send_reasoning_content`. So in `to_chat_dict`, reasoning content will never be sent.

Based on this, calling DeepSeek's deepseek-reasoner failed.

Here will check the `send_reasoning_content` var:
```py
        if self.send_reasoning_content and self.reasoning_content:
            message_dict["reasoning_content"] = self.reasoning_content
```

```py
    @classmethod
    def from_llm_chat_message(cls, message: LiteLLMMessage) -> "Message":
        """Convert a LiteLLMMessage (Chat Completions) to our Message class.

        Provider-agnostic mapping for reasoning:
        - Prefer `message.reasoning_content` if present (LiteLLM normalized field)
        - Extract `thinking_blocks` from content array (Anthropic-specific)
        """
        assert message.role != "function", "Function role is not supported"

        rc = getattr(message, "reasoning_content", None)
        thinking_blocks = getattr(message, "thinking_blocks", None)

        # Convert to list of ThinkingBlock or RedactedThinkingBlock
        if thinking_blocks is not None:
            thinking_blocks = [
                ThinkingBlock(**tb)
                if tb.get("type") == "thinking"
                else RedactedThinkingBlock(**tb)
                for tb in thinking_blocks
            ]
        else:
            thinking_blocks = []

        tool_calls = None

        if message.tool_calls:
            # Validate tool calls - filter out non-function types
            if any(tc.type != "function" for tc in message.tool_calls):
                logger.warning(
                    "LLM returned tool calls but some are not of type 'function' - "
                    "ignoring those"
                )

            function_tool_calls = [
                tc for tc in message.tool_calls if tc.type == "function"
            ]

            if len(function_tool_calls) > 0:
                tool_calls = [
                    MessageToolCall.from_chat_tool_call(tc)
                    for tc in function_tool_calls
                ]
            else:
                # If no function tool calls remain after filtering, raise an error
                raise ValueError(
                    "LLM returned tool calls but none are of type 'function'"
                )

        return Message(
            role=message.role,
            content=[TextContent(text=message.content)]
            if isinstance(message.content, str)
            else [],
            tool_calls=tool_calls,
            reasoning_content=rc,
            thinking_blocks=thinking_blocks,
        )
```

```py
[...]

PR #4446, opened 2026-08-10
feat(llm): support user-configurable send_reasoning_content per model

cheney32:
Let users opt a model in/out via LLM.capability_overrides instead of only the hardcoded allow-list, which stays as the fallback.

HUMAN:

This change makes it possible for users to control, per LLM instance, whether the SDK sends a model's full reasoning content back in the message input. Previously this was decided only by a hardcoded allow-list (SEND_REASONING_CONTENT_MODELS), so enabling a new thinking-capable model — or one served under a custom/proxy name — meant editing that list, releasing a new SDK version, and waiting for consumers to upgrade, with no way to turn it off for a model already on the list. 

The fix routes send_reasoning_content through the same resolution path already used by other model capabilities, so an explicit LLM.capability_overrides value wins first, then LiteLLM metadata, and finally the hardcoded list as a fallback. It is fully backward compatible (with no override, the list behaves exactly as before) and adds no new public API or settings migration, since capability_overrides is an existing field.

---

AGENT:

## Why

Whether the SDK sends full reasoning content back in the message input was
gated only by the hardcoded `SEND_REASONING_CONTENT_MODELS` allow-list in
`openhands-sdk/openhands/sdk/llm/utils/model_features.py`. Adding a new
thinking-capable model (or one behind a custom/proxy name) required editing the
list, releasing the SDK, and having consumers upgrade — and there was no way to
force-disable it for a listed model. This makes the behavior user-configurable
per LLM instance while keeping the list as a fallback.

## Summary

- Route `send_reasoning_content` through the existing `_resolved_bool` helper in
  `get_features`, so `LLM.capability_overrides["send_reasoning_content"]` takes
  precedence, then LiteLLM metadata, then the hardcoded list fallback.
- Document `send_reasoning_content` as a supported key in the
  `LLM.capability_overrides` field description.
- Add a unit test covering enabling a non-listed model and disabling a listed
  model via the override.

## How to Test

```bash
uv run pytest tests/sdk/llm/test_model_features.py -k send_reasoning
```

Or programmatically:

```python
from openhands.sdk.llm.utils.model_features import get_features

# fallback list still works
assert get_features("kimi-k2-thinking").send_reasoning_content is True
assert get_features("gpt-4o").send_reasoning_content is False
# per-instance override, both directions
assert get_features(
    "gpt-4o", overrides={"send_reasoning_content": True}
).send_reasoning_content is True
assert get_features(
    "kimi-k2-thinking", overrides={"send_reasoning_content": False}
).send_reasoning_content is False
```

## Design Doc

See `.pr/design.md` for the before/after and rationale.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Backward compatible: with no override, the hardcoded list decides exactly as
  before.
- No new public API surface and no persisted-settings migration:
  `capability_overrides` is an existing `dict[str, bool | str]` field.

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.43s · commit 4169c8c  
**Coverage:** complete supplied coverage; 4/4 hunks, 4/4 files.

<details>

| --- | --- | --- |
| Command injection | 3.0% | No direct hunk selected |
| Weakened authorization | 5.0% | No direct hunk selected |
| Contract regression | 7.0% | No direct hunk selected |
| Sensitive data disclosure | 4.0% | No direct hunk selected |
| Unexpected data transfer | 3.0% | No direct hunk selected |
| Credential misuse | 4.0% | No direct hunk selected |
| Untrusted instruction authority | 3.0% | No direct hunk selected |
| Privileged environment access | 2.0% | No direct hunk selected |
| Security assessment bypass | 8.0% | No direct hunk selected |
| Primary concern | None selected; confidence 64.0% | No primary concern to locate |

</details>

### DIFF
--- a/.pr/design.md
+++ b/.pr/design.md
@@ -1,0 +1,91 @@
+# Make `send_reasoning_content` user-configurable per LLM instance
+
+## Problem
+
+Whether the SDK sends the model's full reasoning content back in the message
+input was decided **only** by a hardcoded allow-list,
+`SEND_REASONING_CONTENT_MODELS` in
+`openhands-sdk/openhands/sdk/llm/utils/model_features.py`.
+
+Consequences of the hardcoded-only approach:
+
+- Enabling a new thinking-capable model required editing the list, cutting a
+  new SDK release, and having every consumer upgrade + migrate.
+- Users running a model behind a custom/proxy name the list does not match had
+  no way to opt in.
+- There was no way to force-disable the behavior for a listed model.
+
+## Approach
+
+Reuse the LLM's existing per-instance override channel,
+`LLM.capability_overrides` (`openhands-sdk/openhands/sdk/llm/llm.py`), which
+already flows into `get_features(..., overrides=...)` and already backs
+`supports_vision`, `supports_stop_words`, `supports_prompt_cache`, etc.
+
+`send_reasoning_content` was the one capability that ignored `overrides` and
+read the hardcoded list directly. We route it through the same `_resolved_bool`
+helper the other capabilities use, keeping the list as the fallback.
+
+Resolution precedence (unchanged pattern, now applied to this field too):
+
+1. explicit `capability_overrides["send_reasoning_content"]` (respects `False`)
+2. LiteLLM model metadata, if present
+3. fallback: `model_matches(model, SEND_REASONING_CONTENT_MODELS)`
+
+This deliberately keeps provider-specific criteria inside the feature registry
+rather than `llm.py`, matching the repo guidance that "LLM-specific behavior
+tweaks should start in `model_features.py` whenever they can be expressed as
+model/provider capabilities."
+
+## Before / After
+
+Before (`get_features`):
+
+```python
+send_reasoning_content=model_matches(model, SEND_REASONING_CONTENT_MODELS),
+```
+
+After:
+
+```python
+send_reasoning_content=_resolved_bool(
+    "send_reasoning_content",
+    overrides=overrides,
+    metadata=model_info,
+    fallback=model_matches(model, SEND_REASONING_CONTENT_MODELS),
+),
+```
+
+Usage:
+
+```python
+# Opt a not-yet-listed model in:
+LLM(model="my/new-thinking-model",
+    capability_overrides={"send_reasoning_content": True})
+
+# Force-disable for a listed model:
+LLM(model="kimi-k2-thinking",
+    capability_overrides={"send_reasoning_content": False})
+```
+
+## Why this is safe / low-risk
+
+- **Backward compatible**: default behavior is identical — with no override the
+  hardcoded list still decides. No behavior change for existing configs.
+- **No new public API surface**: `capability_overrides` already exists and is a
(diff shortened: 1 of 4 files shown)


enyst (changes requested):
Thank you for the proposal. A little question, is there an issue showing that this is a problem?

On another note, I’m afraid this variable is misnamed: many other LLMs have their reasoning content sent back to them, in the SDK.

This one only means that LiteLLM doesn’t consider correctly some open LLMs as reasoning models, so we can’t recognize them together with the rest, which is why we recognize them here. Or so I recall, open to change my opinion if that’s not the case.

I think maybe we could rename it first, just so we know better… wdyt?

cheney32:
> Thank you for the proposal. A little question, is there an issue showing that this is a problem?
> 
> On another note, I’m afraid this variable is misnamed: many other LLMs have their reasoning content sent back to them, in the SDK.
> 
> This one only means that LiteLLM doesn’t consider correctly some open LLMs as reasoning models, so we can’t recognize them together with the rest, which is why we recognize them here. Or so I recall, open to change my opinion if that’s not the case.
> 
> I think maybe we could rename it first, just so we know better… wdyt?

Thanks for the reply!

**Related discussion**
https://github.com/OpenHands/software-agent-sdk/issues/1343
https://github.com/OpenHands/software-agent-sdk/pull/1464

The concern applies to both self-deployed models and proxied models: they usually run behind a custom/aliased model name, and both of our recognition paths key off a known model name — which is exactly where they fall through.

**LiteLLM recognition**: it relies on LiteLLM's model registry. A self-hosted or proxied model behind a custom name isn't in that registry — we literally get 『This model isn't mapped yet. model=qwen3.6-35b, custom_llm_provider=openai.』, so LiteLLM can't classify it (capabilities, cost, reasoning) at all.

**The SEND_REASONING_CONTENT_MODELS fallback**: it's a substring match on the model name, so it only helps when the operator happens to name the endpoint with one of the listed substrings — which callers often can't control.
So for a large number of self-deployed and proxied reasoning models, neither LiteLLM recognition nor the hardcoded list matches. The reasoning_content then gets dropped on the next turn, the model loses its own prior thinking, and today the only workarounds are editing the list + cutting an SDK release + having every consumer upgrade — with no way to force-disable it for a listed model either. This affects a large portion of self-hosted reasoning setups, not just a rare edge case.

That's the gap this change fills: a per-instance escape hatch (via capability_overrides) for exactly the models automatic detection and the static list cannot know about, while leaving the default (list-based) behavior untouched.
On the naming — you're right that it's misleading, and we'd rather you decide the final name. To get the ball rolling, some candidates depending on which angle you prefer:

- Recognition-oriented (matches your framing that this is really "reasoning models LiteLLM doesn't detect"): EXTRA_REASONING_MODELS, REASONING_MODELS_UNDETECTED_BY_LITELLM.

- Behavior-oriented (what the flag actually does in the Chat Completions path): ECHO_REASONING_CONTENT, REQUIRES_REASONING_CONTENT_ECHO.

Do you have a naming you'd prefer?

PR #4448, opened 2026-08-10
fix(sdk): preserve pinned requested_ref in InstallationManager.update()

### LINKED ISSUE #4363: InstallationManager.update() silently drops pinned refs (should reconcile, not force ref=None)
`InstallationManager.update()` (`openhands-sdk/openhands/sdk/extensions/installation/manager.py:293-325`) always re-fetches with `ref=None`:

```python
return self.install(
    source=current_info.source,
    ref=None,
    repo_path=current_info.repo_path,
    force=True,
)
```

This ignores whatever ref was originally requested (now recorded in `InstallationInfo.requested_ref`, added in #4349). Calling `update()` on an extension pinned to a tag or commit silently un-pins it and jumps to the source's current default-branch HEAD instead of just re-syncing to the pinned ref.

This isn't hypothetical — it's reachable today via the public API for both live systems that share this code path:
- `plugins_router.py:341` (`POST /installed/{plugin_name}/refresh`) → `service_update_plugin` → `InstallationManager.update()`
- `skills_router.py:507` (`POST /installed/{skill_name}/refresh`) → `service_update_skill` → `InstallationManager.update()`

**Prior art:** `earendil-works/pi` (a comparable extension/package system) makes this split explicit — see [`packages.md`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/packages.md#git):

> Refs are pinned tags or commits. `pi update --extensions` and `pi update --all` do not move them to newer refs, but they do reconcile an existing clone to the configured ref.

Moving to a genuinely new ref is a separate, explicit action (`pi install git:host/user/repo@new-ref`), distinct from routine reconciliation.

**Suggested approach:** `update()` should default to reconciling against `current_info.requested_ref` instead of hardcoding `ref=None`:
- If `requested_ref` is set, re-resolve that same ref (catches drift, e.g. a moved tag; otherwise a no-op).
- If `requested_ref` is `None`, the install was already floating, so resolving to latest is correct, same as today.
- Re-pinning to a different ref stays an explicit call to `install(source, ref=new_ref, force=True)`, which already supports it — `update()` doesn't need a new parameter, just to stop forcing `ref=None`.

**Note:** this changes the current (undocumented) behavior of `update()`/the `/refresh` endpoints for both Plugins and Skills — worth confirming nothing relies on "refresh always jumps to latest regardless of pin" before changing the default.

This also directly informs #4351's "Check" step ("resolve the tracked source's ref") — Canvas Extensions' staged-install service will need the same reconcile-vs-repin distinction, whether by calling a fixed `update()` or reimplementing the same logic.

Related: #4349, #4351

santhiprakash:
HUMAN:

Santhi Prakash — small correctness fix for the extension update path; see AGENT section for evidence.

---

AGENT:

## Why

`InstallationManager.update()` always re-fetched with `ref=None`, ignoring the `requested_ref` recorded at install time (added in #4349 / #4375). Calling `update()` on an extension pinned to a tag or commit silently un-pinned it and jumped to the source's current default-branch HEAD instead of just re-syncing to the pinned ref. This is reachable today via the public API for both Plugins and Skills:

- `plugins_router.py` `POST /installed/{plugin_name}/refresh` → `service_update_plugin` → `InstallationManager.update()`
- `skills_router.py` `POST /installed/{skill_name}/refresh` → `service_update_skill` → `InstallationManager.update()`

Reported in #4363.

## Summary

- `update()` now re-fetches at `current_info.requested_ref` instead of hardcoding `ref=None`. A pinned install (tag/commit) stays pinned (only its resolved SHA may move, e.g. a moved tag); a floating install (`requested_ref` is `None`) still resolves to latest, same as before.
- Re-pinning to a different ref remains an explicit `install(source, ref=new_ref, force=True)` call — `update()` needs no new parameter.
- This mirrors the reconcile-vs-repin split already used by the Canvas Extensions staged refresh (`check_canvas_extension_update` fetches with `ref=current_info.requested_ref`), so the shared `InstallationManager` now matches that established pattern.
- Replaced `test_update_clears_requested_ref_to_track_latest` (which encoded the old un-pinning behavior) with `test_update_preserves_pinned_requested_ref`, and added `test_update_floating_install_resolves_latest` to guard the floating case.

## Issue Number

Fixes #4363

## How to Test

```
uv run pytest tests/sdk/extensions/installation/test_installation_manager.py -q
uv run pytest tests/sdk/extensions/ tests/agent_server/canvas_extensions/ tests/agent_server/test_skills_router.py tests/agent_server/test_plugins_router.py -q
uv run ruff check openhands-sdk/openhands/sdk/extensions/installation/manager.py tests/sdk/extensions/installation/test_installation_manager.py
uv run pre-commit run --files openhands-sdk/openhands/sdk/extensions/installation/manager.py tests/sdk/extensions/installation/test_installation_manager.py
```

## Video/Screenshots

Not applicable — pure logic change with no UI/visual surface; behavior is covered by the unit tests above.

## Design Doc

Not added — this is a small one-line correctness fix (a trivial PR per CONTRIBUTING.md), not a new API or behavior-change in the agent loop.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This changes the (previously undocumented) behavior of `update()` and the `/refresh` endpoints for Plugins and Skills: a pinned extension now stays pinned on refresh instead of jumping to latest. The issue calls this out explicitly and the new behavior matches the reconcile-vs-repin distinction already shipped for Canvas Extensions in #4374. No persisted settings shape or REST contract changes — `InstallationInfo.requested_ref` is unchanged.

### DIFF
--- a/openhands-sdk/openhands/sdk/extensions/installation/manager.py
+++ b/openhands-sdk/openhands/sdk/extensions/installation/manager.py
@@ -291,12 +291,17 @@ class InstallationManager[T: ExtensionProtocol]:
         return info
 
     def update(self, name: str) -> InstallationInfo | None:
-        """Update an installed extension to the latest version.
+        """Reconcile an installed extension with its tracked source.
 
-        Re-fetches the extension from its original source with ``ref=None``
-        (i.e. the latest available) and force-reinstalls it.  The previous
-        ``enabled`` state is preserved because ``install(force=True)``
-        carries it over.
+        Re-fetches the extension from its original source at the same
+        ``requested_ref`` recorded at install time and force-reinstalls it.
+        A pinned install (tag or commit) is re-resolved to that ref, so
+        ``update()`` catches drift (e.g. a moved tag) without un-pinning it.
+        A floating install (``requested_ref`` is ``None``) resolves to the
+        source's current default-branch HEAD, same as before.  Re-pinning to
+        a different ref is a separate, explicit ``install(source, ref=new_ref,
+        force=True)`` call.  The previous ``enabled`` state is preserved
+        because ``install(force=True)`` carries it over.
 
         Args:
             name: Name of the extension to update.
@@ -320,7 +325,7 @@ class InstallationManager[T: ExtensionProtocol]:
         logger.info(f"Updating extension {name} from {redacted}")
         return self.install(
             source=current_info.source,
-            ref=None,
+            ref=current_info.requested_ref,
             repo_path=current_info.repo_path,
             force=True,
         )
--- a/tests/sdk/extensions/installation/test_installation_manager.py
+++ b/tests/sdk/extensions/installation/test_installation_manager.py
@@ -513,12 +513,13 @@ def test_update_nonexistent_extension(
     assert info is None
 
 
-def test_update_clears_requested_ref_to_track_latest(
+def test_update_preserves_pinned_requested_ref(
     manager: InstallationManager[MockExtension],
     mock_extension_dir: Path,
 ):
-    """update() re-fetches with ref=None, so a previously pinned requested_ref
-    is cleared to reflect that the extension now tracks the latest version."""
+    """update() re-fetches at the recorded requested_ref instead of ref=None,
+    so a pinned install stays pinned (only its resolved SHA may move, e.g.
+    a moved tag). Regression guard for issue #4363."""
     with patch(
         "openhands.sdk.extensions.installation.manager.fetch_with_resolution",
         return_value=(mock_extension_dir, "abc123"),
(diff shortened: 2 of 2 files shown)


santhiprakash:
Quick note on CI status: every workflow on this PR is sitting in `action_required` (the fork-PR workflow-approval gate), so the PR currently shows **no check runs at all** rather than failing ones.

- The branch itself is current: `mergeable` against `main`, and `main` has had no commits touching `extensions/installation/manager.py` since this head (Aug 10).
- Local verification for the change is in the description: focused pytest suites (`installation/test_installation_manager.py` plus the plugin/skill/canvas-extension router suites), ruff, and pre-commit all green.
- Issue #4363 is still open with no other PR against it.

If a maintainer can approve running the workflows, standard CI should exercise the same tests. Happy to rebase or supply any additional reproduction in the meantime.

santhiprakash:
Closing to keep the review queue clean — this has been open a while without traction. Happy to reopen or resubmit a fresh version if it would still be useful.

PR #4458, opened 2026-08-11
feat: carry ConversationErrorEvent on ConversationRunError for automation callbacks

neubig:
HUMAN:

This is a simple enough change and test coverage is good.

AGENT: 

## Summary

Carries the SDK's existing typed `ConversationErrorEvent` on `ConversationRunError` so the workspace automation completion callback can serialize it directly. This is the SDK side of `OpenHands/automation` #325 / issue #323.

## Changes

- **`ConversationRunError.conversation_error`**: New optional field carrying the authoritative `ConversationErrorEvent` emitted during the failed run.
- **Local conversation**: Both `run()` and `arun()` attach the latest `ConversationErrorEvent` emitted during the current run (scoped by `_run_start_event_count`).
- **Remote conversation**: `_handle_conversation_status` retrieves the latest structured error from cached events and attaches it.
- **Workspace callback**: `_send_completion_callback` serializes `ConversationRunError.conversation_error` directly as the callback `error` field. For exceptions outside a conversation (e.g. setup failures), it constructs a fallback `ConversationErrorEvent` that self-classifies via the existing `classify_error`.

## Why

- The conversation already owns the authoritative classified error event.
- `ConversationRunError` already travels from `conversation.run()` through `workspace.__exit__`.
- No mutable workspace registration, no callback-side reclassification, no new Pydantic models.
- SDK users can inspect `exc.conversation_error` directly.
- The callback `error` is now a structured `ConversationErrorEvent` (or a legacy string for older SDKs).

## Issue Number

Fixes #323

## How to Test

- Focused tests: local/remote authoritative event attachment, fallback event creation, workspace callback serialization, existing cost/conversation_id callback behavior.
- Changed-file pre-commit passes (ruff, pycodestyle, pyright, import rules).

This PR was created by an AI agent (OpenHands) on behalf of neubig.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:e3bba47-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-e3bba47-python \
  ghcr.io/openhands/agent-server:e3bba47-python
```

```
ghcr.io/openhands/agent-server:e3bba47-golang-amd64
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-golang-amd64
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-golang-amd64
ghcr.io/openhands/agent-server:e3bba47-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:e3bba47-golang-arm64
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-golang-arm64
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-golang-arm64
ghcr.io/openhands/agent-server:e3bba47-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:e3bba47-java-amd64
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-java-amd64
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-java-amd64
ghcr.io/openhands/agent-server:e3bba47-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:e3bba47-java-arm64
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-java-arm64
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-java-arm64
ghcr.io/openhands/agent-server:e3bba47-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:e3bba47-python-amd64
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-python-amd64
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-python-amd64
ghcr.io/openhands/agent-server:e3bba47-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:e3bba47-python-arm64
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-python-arm64
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-python-arm64
ghcr.io/openhands/agent-server:e3bba47-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:e3bba47-golang
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-golang
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-golang
ghcr.io/openhands/agent-server:e3bba47-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:e3bba47-java
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-java
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-java
ghcr.io/openhands/agent-server:e3bba47-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:e3bba47-python
ghcr.io/openhands/agent-server:e3bba47376e750e926a18de069e092b95ee39588-python
ghcr.io/openhands/agent-server:feat-report-automation-blocking-factor-python
ghcr.io/openhands/agent-server:e3bba47-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `e3bba47-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `e3bba47-python-amd64`) are also available if needed
</details>

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -143,11 +143,8 @@ def _agent_already_surfaced_error(events: Sequence[Event], since: int = 0) -> bo
     prevents a stale source="agent" event from a prior run from suppressing the error
     event for an unrelated exception in a subsequent run on the same conversation.
     """
-    for i in range(len(events) - 1, since - 1, -1):
-        event = events[i]
-        if isinstance(event, ConversationErrorEvent):
-            return event.source == "agent"
-    return False
+    latest = _latest_conversation_error(events, since)
+    return latest is not None and latest.source == "agent"
 
 
 def _latest_conversation_error(
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -2648,6 +2648,19 @@ class LocalConversation(BaseConversation):
         if first_attempt:
             self._cleanup_initiated = True
 
+            # Hand the existing typed conversation failure to the workspace for
+            # the automation completion callback.
+            try:
+                for event in reversed(self._state.events):
+                    if isinstance(event, ConversationErrorEvent):
+                        if event.classification is not None:
+                            self.workspace.register_failure_kind(
+                                event.classification.kind
+                            )
+                        break
+            except Exception as e:
+                logger.debug(f"Could not register conversation failure: {e}")
+
             # Best-effort: hand the accumulated LLM cost to the workspace so it
             # can be included in the automation completion callback. State is
             # in-process here, so unlike RemoteConversation there is no cache to
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -150,6 +150,17 @@ def _agent_already_surfaced_error(events: Sequence[Event], since: int = 0) -> bo
     return False
 
 
+def _latest_conversation_error(
+    events: Sequence[Event], since: int = 0
+) -> ConversationErrorEvent | None:
+    """Return the latest conversation error emitted during the current run."""
+    for i in range(len(events) - 1, since - 1, -1):
+        event = events[i]
+        if isinstance(event, ConversationErrorEvent):
+            return event
+    return None
+
+
 def _is_acp_prompt_message(event: Event) -> TypeGuard[MessageEvent]:
     if not isinstance(event, MessageEvent):
         return False
@@ -2014,7 +2025,12 @@ class LocalConversation(BaseConversation):
 
             # Re-raise with conversation id and persistence dir for better UX
             raise ConversationRunError(
-                self._state.id, e, persistence_dir=self._state.persistence_dir
+                self._state.id,
+                e,
+                persistence_dir=self._state.persistence_dir,
+                conversation_error=_latest_conversation_error(
+                    self._state.events, _run_start_event_count
+                ),
             ) from e
         finally:
             self._cancel_token = None
@@ -2470,7 +2486,12 @@ class LocalConversation(BaseConversation):
                         )
                     )
             raise ConversationRunError(
-                self._state.id, e, persistence_dir=self._state.persistence_dir
+                self._state.id,
+                e,
+                persistence_dir=self._state.persistence_dir,
+                conversation_error=_latest_conversation_error(
+                    self._state.events, _run_start_event_count
+                ),
             ) from e
         finally:
             # A cancelled token must stay observable: interrupted tool calls run
@@ -2647,19 +2668,6 @@ class LocalConversation(BaseConversation):
         first_attempt = not getattr(self, "_cleanup_initiated", False)
         if first_attempt:
             self._cleanup_initiated = True
-
-            # Hand the existing typed conversation failure to the workspace for
-            # the automation completion callback.
-            try:
-                for event in reversed(self._state.events):
-                    if isinstance(event, ConversationErrorEvent):
-                        if event.classification is not None:
-                            self.workspace.register_failure_kind(
-                                event.classification.kind
-                            )
-                        break
-            except Exception as e:
-                logger.debug(f"Could not register conversation failure: {e}")
 
             # Best-effort: hand the accumulated LLM cost to the workspace so it
             # can be included in the automation completion callback. State is
--- a/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
@@ -1159,6 +1159,7 @@ class RemoteConversation(BaseConversation):
         # subscription snapshot from being mistaken for the post-run snapshot.
         self._run_armed.clear()
         self._drain_terminal_status_queue()
+        run_start_event_count = len(self._state.events)
 
         # Trigger a run on the server using the dedicated run endpoint.
         # Let the server tell us if it's already running (409), avoiding an extra GET.
@@ -1184,7 +1185,9 @@ class RemoteConversation(BaseConversation):
             # after the run was triggered are treated as run-completion signals.
             self._run_armed.set()
             try:
-                self._wait_for_run_completion(poll_interval, timeout)
+                self._wait_for_run_completion(
+                    poll_interval, timeout, run_start_event_count
+                )
             finally:
                 self._run_armed.clear()
 
@@ -1192,6 +1195,7 @@ class RemoteConversation(BaseConversation):
         self,
         poll_interval: float = 1.0,
         timeout: float = 1800.0,
+        since: int = 0,
     ) -> None:
         """Wait for the conversation run to complete.
 
@@ -1246,7 +1250,7 @@ class RemoteConversation(BaseConversation):
             try:
                 ws_status = self._terminal_status_queue.get(timeout=poll_interval)
                 # Raises ConversationRunError on ERROR/STUCK; no-op otherwise.
-                self._handle_conversation_status(ws_status)
+                self._handle_conversation_status(ws_status, since)
                 logger.info(
                     "Run completed via post-run WebSocket state update "
                     "(status: %s, elapsed: %.1fs)",
@@ -1279,7 +1283,7 @@ class RemoteConversation(BaseConversation):
                 terminal_first_seen_at = None
             else:
                 # Raises ConversationRunError for ERROR/STUCK states
-                self._handle_conversation_status(status)
+                self._handle_conversation_status(status, since)
 
                 if status and ConversationExecutionStatus(status).is_terminal():
                     # ERROR/STUCK have already been handled above. FINISHED from
@@ -1340,18 +1344,20 @@ class RemoteConversation(BaseConversation):
         info = resp.json()
         return info.get("execution_status")
 
-    def _handle_conversation_status(self, status: str | None) -> bool:
+    def _handle_conversation_status(self, status: str | None, since: int = 0) -> bool:
         """Handle non-running statuses; return True if the run is complete."""
         if status == ConversationExecutionStatus.RUNNING.value:
             return False
         if status == ConversationExecutionStatus.ERROR.value:
-            error = self._get_last_conversation_error()
+            self._state.events.reconcile()
+            error = self._get_last_conversation_error(since)
+            detail = (
+                f"{error.code}: {error.detail}".strip(": ")
+                if error
+                else "Remote conversation ended with error"
+            )
             raise ConversationRunError(
-                self._id,
-                RuntimeError(
-                    error.detail if error else "Remote conversation ended with error"
-                ),
-                conversation_error=error,
+                self._id, RuntimeError(detail), conversation_error=error
             )
         if status == ConversationExecutionStatus.STUCK.value:
             raise ConversationRunError(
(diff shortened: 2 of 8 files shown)


openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -1386,17 +1389,14 @@ def _handle_poll_exception(self, exc: Exception) -> None:
             return
         raise ConversationRunError(self._id, exc) from exc
 
-    def _get_last_error_detail(self) -> str | None:
-        """Return the most recent ConversationErrorEvent detail, if available."""
+    def _get_last_conversation_error(self) -> ConversationErrorEvent | None:
```
smolpaws: **[Finding 1 — suggest fixing before merge]** This scan is unscoped: it walks back to index 0, so on a conversation re-run after an earlier failure it can return a **stale** `ConversationErrorEvent` from a previous run and attach it as authoritative. The local path scopes with `_run_start_event_count`; consider snapshotting `len(self._state.events)` at the start of `run()` and passing it as a floor here.

Related: both callers raise *before* `self._state.events.reconcile()`, so if the error event hasn't arrived over the WebSocket yet this returns `None` and the automation callback falls back to an unclassified error. A `reconcile()` on the `ERROR` branch before reading would close that race — one extra fetch, only on a terminal path.

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -1345,10 +1345,13 @@ def _handle_conversation_status(self, status: str | None) -> bool:
         if status == ConversationExecutionStatus.RUNNING.value:
             return False
         if status == ConversationExecutionStatus.ERROR.value:
-            detail = self._get_last_error_detail()
+            error = self._get_last_conversation_error()
             raise ConversationRunError(
                 self._id,
-                RuntimeError(detail or "Remote conversation ended with error"),
+                RuntimeError(
+                    error.detail if error else "Remote conversation ended with error"
```
smolpaws: **[Finding 3 — minor]** This drops the code prefix. The previous `_get_last_error_detail` returned `f"{code}: {detail}"`, and fell back to `code` when `detail` was empty. Now an empty `detail` produces `RuntimeError("")`, so the user sees `Conversation run failed for id=X: ` and nothing else. Cheap to preserve, e.g. `f"{error.code}: {error.detail}".strip(': ')`.

openhands-sdk/openhands/sdk/workspace/base.py
```diff
@@ -114,7 +118,18 @@ def _send_completion_callback(
...
+                if isinstance(exc_val, ConversationRunError)
+                else None
+            )
+            if error is None:
+                error = ConversationErrorEvent(
+                    source="environment",
+                    code=type(exc_val).__name__,
+                    detail=str(exc_val),
+                )
+            payload["error"] = error.model_dump(mode="json")
```
smolpaws, line 137: **[Finding 2 — suggest fixing before merge]** When `exc_val` is a `ConversationRunError` whose `conversation_error` is `None`, this fallback uses the *wrapper* rather than the cause: `code="ConversationRunError"` and `detail="Conversation run failed for id=...: <cause>"` (plus, for local, the multi-line "file a bug report at ISSUE_URL" text).

`classify_error` checks `code` first and treats it as authoritative — `"ConversationRunError"` matches nothing, so it drops to the detail heuristics and lands on `unknown`. That hits exactly the paths that never set `conversation_error`: remote `STUCK`, run timeout, 404, poll failures. Unwrapping would recover the real signal:

```python
cause = (
    exc_val.original_exception
    if isinstance(exc_val, ConversationRunError)
    else exc_val
)
error = ConversationErrorEvent(
    source="environment",
    code=type(cause).__name__,
    detail=str(cause),
)
```

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -150,6 +150,17 @@ def _agent_already_surfaced_error(events: Sequence[Event], since: int = 0) -> bo
     return False
 
 
+def _latest_conversation_error(
```
smolpaws, line 150: **[Finding 4 — cleanup]** This is the same backwards walk as `_agent_already_surfaced_error` just above, and both now run in the same `except` block — two scans over the same range. The former is expressible in terms of this one:

```python
latest = _latest_conversation_error(events, since)
agent_surfaced = latest is not None and latest.source == "agent"
```

tests/sdk/conversation/remote/test_remote_conversation.py
```diff
@@ -1231,7 +1232,12 @@ def test_remote_conversation_run_ws_error_still_terminates_immediately(
 
         mock_ws_client.return_value = Mock()
         conversation = RemoteConversation(agent=self.agent, workspace=self.workspace)
-        conversation._get_last_error_detail = Mock(return_value="boom")
+        error = ConversationErrorEvent(
+            source="environment",
+            code="LLMAuthenticationError",
+            detail="invalid api key",
+        )
+        conversation.state.events._cached_events.append(error)
```
smolpaws: **[Finding 6 — test nit]** This reaches past `add_event()` into the private cache, skipping `_cached_event_ids` and the lock, so it'll break silently if `RemoteEventsList` grows an invariant. `conversation.state.events.add_event(error)` would exercise the real path.

Also, below: `assert attached is error` followed by `assert attached is not None` — the second is dead once the first passes. If it's there to satisfy pyright, it needs to come first.

smolpaws (commented):
👋 Hi — I'm an AI agent (Claude Opus 5, running in Claude Code) posting on behalf of **@enyst** (Engel Nyst) from the `smolpaws` account. Engel asked me to review this PR. Findings below are mine; happy to be told I'm wrong on any of them, and I can open a follow-up with fixes if useful.

## Verdict

The design is right: reusing the existing typed `ConversationErrorEvent` instead of inventing a parallel model, and letting the conversation stay the authority, is the correct call. The local implementation is clean and correctly scoped.

**Findings 1 and 2 look worth fixing before merge** — together they mean the automation service gets a wrong or unclassified blocking factor on a meaningful share of real *remote* failures, which is the thing this PR exists to prevent. 3–6 are cleanup.

---

### 1. Remote: the "authoritative" event can be stale or missing

`_handle_conversation_status` raises on `ERROR` at both call sites (the WS path and the REST-poll path) **before** `self._state.events.reconcile()` runs, and `_get_last_conversation_error()` reads only `_cached_events`. Two failure modes:

- The `ConversationErrorEvent` hasn't arrived over the WebSocket yet → returns `None` → the automation callback gets the unclassified fallback instead of the authoritative event. The REST poll can beat the event stream.
- Worse: the scan is **unscoped** (walks back to index 0). On a conversation that is re-run after a previous failure, a **stale error event from the earlier run** gets attached and reported as authoritative. Previously that staleness only corrupted a message string; now it produces a wrong machine-actionable `classification.kind` — which is exactly what the automation service branches on.

The local path already solved this with `_run_start_event_count`. Suggest doing the same on the remote side (snapshot `len(self._state.events)` at the start of `run()`), and calling `reconcile()` on the `ERROR` branch before reading — one extra fetch, only on a terminal path.

### 2. The fallback doesn't unwrap `original_exception`

In `_send_completion_callback`:

```python
error = ConversationErrorEvent(code=type(exc_val).__name__, detail=str(exc_val))
```

Several `ConversationRunError`s never set `conversation_error`: remote `STUCK`, run timeout, 404, poll failures — plus finding 1's race. For all of those this yields `code="ConversationRunError"` and `detail="Conversation run failed for id=...: Remote conversation got stuck"` (and for local, the multi-line "file a bug report at `ISSUE_URL`" body).

`classify_error` checks `code` first and treats it as authoritative — `"ConversationRunError"` matches nothing, so it falls through to the detail heuristics and lands on `unknown`. Unwrapping to `type(exc_val.original_exception).__name__` / `str(exc_val.original_exception)` would give `TimeoutError`, `RuntimeError`, etc. As written, the callback reports "unknown blocking factor" on precisely the paths that most needed classifying.

### 3. Remote error message drops the code prefix

Old: `f"{code}: {detail}"`, with a fallback to `code` when detail was empty. New: `error.detail` alone. If `detail` is empty the user now sees `Conversation run failed for id=X: ` and nothing else. Low severity, user-facing, cheap to keep.

### 4. Two identical backwards scans

`_agent_already_surfaced_error(events, since)` is exactly `(e := _latest_conversation_error(events, since)) is not None and e.source == "agent"`. Both now run in the same `except` block, walking the same range twice. Could collapse to one helper and one call.

### 5. Callback contract break with no version marker

`payload["error"]` goes `str` → `dict`, so the consumer has to type-sniff to stay compatible across SDK versions. Fine as a decision, but nothing in the payload signals the shape — a `schema_version`, or a new `error_event` key alongside a retained string `error`, would be cheaper for the consumer than `isinstance` branching forever. Not a blocker.

One note since the code invites it: `ErrorClassification`'s own docstring calls itself "the only failure metadata that crosses the event/API boundary" and says `detail` is "never copied here or sent to telemetry." This PR now ships raw `detail` to an external callback URL. **Not a regression** — `str(exc_val)` was already going over the same wire — but it does contradict the stated contract, so it's worth a deliberate "yes, we mean this."

### 6. Test brittleness

`conversation.state.events._cached_events.append(error)` reaches past `add_event()` into a private list, skipping `_cached_event_ids` and the lock; it'll break silently if that cache grows an invariant. Also `assert attached is error` followed by `assert attached is not None` is backwards — the second is dead once the first passes (presumably there for pyright; putting it first would do the job).

---

### Notes

- **Iteration cost** (Engel asked specifically): the local scan is bounded by `_run_start_event_count`, so it only covers events from the current run, and it exits on the first iteration in the dominant branch since the error event was just appended. The remote scan walks the whole log, but over an in-memory list of already-deserialized events — microseconds even at 60k events, so cost isn't the concern there; scoping (finding 1) is. Worth noting the `since` bound is load-bearing on the local side rather than a mere optimization, since `EventLog.__getitem__` hits disk + `model_validate_json` on a cache miss for events from prior runs.
- **CI**: 36 checks green; the 2 failures are both "Validate PR description" (the `HUMAN:` / `AGENT:` preamble vs. the template), no test failures.

tests/sdk/conversation/remote/test_remote_conversation.py
enyst: (HUMAN: this sounds correct to me, `add_event()` exists for this kind of use; writes to `state` should be under lock)

openhands-sdk/openhands/sdk/workspace/base.py
enyst: (HUMAN: not sure about this one, maybe the agent can verify too)

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
neubig: Implemented in e3bba473: remote error lookup now reconciles terminal ERROR state and is scoped to events from the current run. This reply was generated by an AI agent (OpenHands) on behalf of neubig.

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
neubig: Implemented in e3bba473: preserved the error code prefix while retaining the structured ConversationErrorEvent. This reply was generated by an AI agent (OpenHands) on behalf of neubig.

openhands-sdk/openhands/sdk/workspace/base.py
neubig: Implemented in e3bba473: fallback classification now unwraps ConversationRunError.original_exception. This reply was generated by an AI agent (OpenHands) on behalf of neubig.

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
neubig: Implemented in e3bba473: local duplicate detection reuses the latest current-run error helper. This reply was generated by an AI agent (OpenHands) on behalf of neubig.

tests/sdk/conversation/remote/test_remote_conversation.py
neubig: Implemented in e3bba473: the test uses RemoteEventsList.add_event() and checks non-null before accessing classification. This reply was generated by an AI agent (OpenHands) on behalf of neubig.

PR #4460, opened 2026-08-11
fix: don't hold _litellm_modify_params_lock across the entire LLM call — perf #16459

waterWang:
HUMAN:

Automated fix for LLM call serialization issue.

---

AGENT:

## Why

When running multiple conversations simultaneously with a local model (vLLM, Ollama, llama.cpp), all LLM requests are serialized through a `ClassVar[threading.Lock]` held for the entire duration of each call (30-120s for local models). The GPU sits idle while conversations queue behind the lock.

## Summary

- Narrow the lock scope in `_litellm_modify_params_ctx` and `_alitellm_modify_params_ctx` to only protect the `set`/`restore` of `litellm.modify_params`, not the duration of the upstream LLM API call.
- `litellm.modify_params` is a simple configuration boolean read at the start of each request; it is not mutable state that changes during the call, so there is no risk of a concurrent thread corrupting an in-flight call.
- Lock acquisition still uses the dedicated executor for the async path to avoid blocking the event loop. Cancellation handling (shield + release callback) is preserved.

## Issue Number

Fixes #16459

## How to Test

1. Run the existing test suite: `pytest tests/ -v`
2. Start multiple conversations with a local model provider (e.g., Ollama)
3. Observe that LLM calls from different conversations run concurrently instead of being serialized
4. Verify no regression in `modify_params` behavior for cloud API providers (Anthropic, Bedrock, etc.)

## Video/Screenshots

No UI changes.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/openhands-sdk/openhands/sdk/llm/llm.py
+++ b/openhands-sdk/openhands/sdk/llm/llm.py
@@ -2271,46 +2271,41 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
 
     @contextmanager
     def _litellm_modify_params_ctx(self, flag: bool):
+        # Only hold the lock while setting/restoring the global flag, not
+        # during the entire transport call.  ``litellm.modify_params`` is a
+        # simple configuration boolean read at the start of each LLM request;
+        # it is not mutable state that changes during the call, so there is
+        # no risk of a concurrent thread corrupting an in-flight call.
         with self._litellm_modify_params_lock:
             old = getattr(litellm, "modify_params", None)
-            try:
-                litellm.modify_params = flag
-                yield
-            finally:
+            litellm.modify_params = flag
+        try:
+            yield
+        finally:
+            with self._litellm_modify_params_lock:
                 litellm.modify_params = old
 
     @asynccontextmanager
     async def _alitellm_modify_params_ctx(self, flag: bool):
         """Async variant of :meth:`_litellm_modify_params_ctx`.
 
-        ``litellm.modify_params`` is a process-wide global, so the lock must
-        stay held for the full duration of the transport call, not just the
-        moment the flag is set. A plain ``with self._litellm_modify_params_lock:``
-        would work for that, but only for the sync path: entering it here
-        with a blocking ``with`` statement would hold a real OS-level lock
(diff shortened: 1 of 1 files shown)


VascoSch92 (changes requested):
Agreed the lock is too coarse — holding it across the transport call serializes every LLM call in the process, and that's #16459. But narrowing it to the two assignments doesn't make the guard cheaper, it makes it not a guard, and the async path gains a deadlock. Three blockers:

1. Event-loop deadlock (llm.py:2335). The new finally: with self._litellm_modify_params_lock: is a blocking acquire on the loop thread. Coroutine B takes the lock on an executor thread (2311) and only B's continuation on the loop thread releases it (2330) — if another coroutine reaches its exit first, the loop blocks on a lock only the blocked thread can release. Driving _alitellm_modify_params_ctx with 16 concurrent tasks: this branch deadlocks at iteration 0, main completes 32,000. In agent-server that hangs every conversation and interrupt() until the process is killed. The docstring at 2155 still says this method must not use a plain blocking with — that rule was the point of the text being deleted here.

2. The flag isn't stable for the call it guards (2279-2286). litellm reads the global during prompt transformation, not at request start (factory.py:1082, :2387, :4049). And this doesn't need mixed config: SDK default is True, litellm's is False, so every guard writes False back on exit while other calls are in flight. Two LLMs both at True → one observes False mid-call, and the global is left corrupted at True after all guards exit. Clean on main.

3. CI is red. test_litellm_modify_params_context_serializes_threads fails at head (assert not entered_second.wait(timeout=0.2)). That test is the invariant being removed.

Suggested fix: gate on the value rather than shrinking the lock. Concurrent calls wanting the same value never conflict — only a change does. Keep a (value, in_flight_count) under a Condition: install the value only when the count is 0, join freely when it already matches, wait otherwise. Since every LLM defaults to True, that's full parallelism for #16459 with no race, and exit is short enough to be safe on the loop thread. Happy to sketch it if useful.

Minor, non-blocking: the catch_warnings blocks at 2148/2159/1744/1896 lose their serialization too; the run_in_executor/shield machinery is dead weight once the section is one assignment; and the new docstring narrates change history (AGENTS.md:271).

PR #4461, opened 2026-08-11
fix(sdk): cap condenser token limit by agent context

### LINKED ISSUE #4421: feat(llm): resolve provider-specific runtime metadata for routed models

neubig:
HUMAN:

I have not tested this live, but it is simple enough that I was able to check the code given the tests and reading the code.

---

AGENT:

## Why

When a condenser has a configured token limit larger than the active agent model's effective input limit—or no configured token limit at all—it can wait until the agent request overflows before condensing. The condenser should use the stricter limit from the current agent LLM.

## Summary

- Use the minimum of configured `max_tokens` and `agent_llm.effective_max_input_tokens`.
- Evaluate the effective threshold during condensation so model switches are handled correctly.
- Add regression coverage for configured, inherited, and stricter limits.

## Issue Number

Fixes #4421

## How to Test

Run:

```bash
uv run ruff check openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py tests/sdk/context/condenser/test_llm_summarizing_condenser.py
uv run ruff format --check openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py tests/sdk/context/condenser/test_llm_summarizing_condenser.py
uv run pytest tests/sdk/context/condenser tests/sdk/agent/test_agent_context_window_condensation.py
```

All 89 focused condenser and context-window tests pass.

## Video/Screenshots

Not applicable: this is an SDK behavior change with automated test coverage.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The threshold preserves explicit condenser configuration as the stricter bound while adapting to the active agent model's effective input capacity.

This pull request description was updated by an AI agent (OpenHands) on behalf of the user.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:53231f6-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-53231f6-python \
  ghcr.io/openhands/agent-server:53231f6-python
```

```
ghcr.io/openhands/agent-server:53231f6-golang-amd64
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-golang-amd64
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-golang-amd64
ghcr.io/openhands/agent-server:53231f6-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:53231f6-golang-arm64
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-golang-arm64
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-golang-arm64
ghcr.io/openhands/agent-server:53231f6-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:53231f6-java-amd64
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-java-amd64
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-java-amd64
ghcr.io/openhands/agent-server:53231f6-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:53231f6-java-arm64
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-java-arm64
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-java-arm64
ghcr.io/openhands/agent-server:53231f6-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:53231f6-python-amd64
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-python-amd64
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-python-amd64
ghcr.io/openhands/agent-server:53231f6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:53231f6-python-arm64
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-python-arm64
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-python-arm64
ghcr.io/openhands/agent-server:53231f6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:53231f6-golang
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-golang
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-golang
ghcr.io/openhands/agent-server:53231f6-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:53231f6-java
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-java
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-java
ghcr.io/openhands/agent-server:53231f6-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:53231f6-python
ghcr.io/openhands/agent-server:53231f68e20b5398a04133cfd3fb851590555f2a-python
ghcr.io/openhands/agent-server:fix-dynamic-condenser-token-limit-python
ghcr.io/openhands/agent-server:53231f6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `53231f6-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `53231f6-python-amd64`) are also available if needed
</details>

---

## Review progress

- Addressed review feedback and resolved the unresolved review thread (none remain).
- `_effective_max_tokens` now documents why `agent_llm` may be `None`.
- Latest head: `b4d705482`.

### DIFF
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -93,6 +93,17 @@ class LLMSummarizingCondenser(RollingCondenser):
     def handles_condensation_requests(self) -> bool:
         return True
 
+    def _effective_max_tokens(self, agent_llm: LLM | None) -> int | None:
+        limits = [
+            limit
+            for limit in (
+                self.max_tokens,
+                agent_llm.effective_max_input_tokens if agent_llm is not None else None,
+            )
+            if limit is not None
+        ]
+        return min(limits) if limits else None
+
     def get_condensation_reasons(
         self, view: View, agent_llm: LLM | None = None
     ) -> set[Reason]:
@@ -112,15 +123,8 @@ class LLMSummarizingCondenser(RollingCondenser):
         if view.unhandled_condensation_request:
             reasons.add(Reason.REQUEST)
 
-        # Reason 2: Token limit is provided and exceeded. The agent's effective
-        # input limit is authoritative when it is lower than the configured limit.
-        max_tokens = self.max_tokens
-        if agent_llm is not None and agent_llm.effective_max_input_tokens is not None:
-            if max_tokens is None:
-                max_tokens = agent_llm.effective_max_input_tokens
-            else:
-                max_tokens = min(max_tokens, agent_llm.effective_max_input_tokens)
-
+        # Reason 2: Token limit is provided and exceeded.
+        max_tokens = self._effective_max_tokens(agent_llm)
         if max_tokens is not None and agent_llm is not None:
             total_tokens = get_total_token_count(view.events, agent_llm)
             if total_tokens > max_tokens:
@@ -262,13 +266,14 @@ class LLMSummarizingCondenser(RollingCondenser):
 
         if Reason.TOKENS in reasons:
             # Compute the number of tokens we need to eliminate to be under half the
-            # max_tokens value. We know max_tokens and the agent LLM are not None here
-            # because we can't have Reason.TOKENS without them.
-            assert self.max_tokens is not None
+            # effective max_tokens value. We know both are not None here because we
+            # cannot have Reason.TOKENS without them.
+            max_tokens = self._effective_max_tokens(agent_llm)
+            assert max_tokens is not None
             assert agent_llm is not None
 
             total_tokens = get_total_token_count(view.events, agent_llm)
-            tokens_to_reduce = total_tokens - (self.max_tokens // 2)
+            tokens_to_reduce = total_tokens - (max_tokens // 2)
 
             suffix_events_to_keep.add(
                 get_suffix_length_for_token_reduction(
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -112,15 +112,23 @@ class LLMSummarizingCondenser(RollingCondenser):
         if view.unhandled_condensation_request:
             reasons.add(Reason.REQUEST)
 
-        # Reason 2: Token limit is provided and exceeded.
-        if self.max_tokens and agent_llm:
+        # Reason 2: Token limit is provided and exceeded. The agent's effective
+        # input limit is authoritative when it is lower than the configured limit.
+        max_tokens = self.max_tokens
+        if agent_llm is not None and agent_llm.effective_max_input_tokens is not None:
+            if max_tokens is None:
+                max_tokens = agent_llm.effective_max_input_tokens
+            else:
+                max_tokens = min(max_tokens, agent_llm.effective_max_input_tokens)
+
+        if max_tokens is not None and agent_llm is not None:
             total_tokens = get_total_token_count(view.events, agent_llm)
-            if total_tokens > self.max_tokens:
+            if total_tokens > max_tokens:
                 logger.info(
                     "Condenser token limit exceeded: total_tokens=%d max_tokens=%d "
                     "events=%d",
                     total_tokens,
-                    self.max_tokens,
+                    max_tokens,
                     len(view),
                 )
                 reasons.add(Reason.TOKENS)
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -94,13 +94,21 @@ class LLMSummarizingCondenser(RollingCondenser):
         return True
 
     def _effective_max_tokens(self, agent_llm: LLM | None) -> int | None:
+        """Return the effective token cap that triggers token-based condensation.
+
+        Takes the stricter (i.e. smaller) of the condenser's configured
+        ``max_tokens`` and the agent LLM's effective input limit. ``agent_llm``
+        is optional throughout the public API -- callers that only assess
+        request/event-count pressure may pass ``None`` -- in which case only the
+        condenser's own ``max_tokens`` applies.
+
+        Returns ``None`` when no limit is configured.
+        """
+        agent_limit = (
+            agent_llm.effective_max_input_tokens if agent_llm is not None else None
+        )
         limits = [
-            limit
-            for limit in (
-                self.max_tokens,
-                agent_llm.effective_max_input_tokens if agent_llm is not None else None,
-            )
-            if limit is not None
+            limit for limit in (self.max_tokens, agent_limit) if limit is not None
         ]
         return min(limits) if limits else None
 
--- a/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
+++ b/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
@@ -104,11 +104,13 @@ class LLMSummarizingCondenser(RollingCondenser):
 
         Returns ``None`` when no limit is configured.
         """
-        agent_limit = (
-            agent_llm.effective_max_input_tokens if agent_llm is not None else None
-        )
         limits = [
-            limit for limit in (self.max_tokens, agent_limit) if limit is not None
+            limit
+            for limit in (
+                self.max_tokens,
+                agent_llm.effective_max_input_tokens if agent_llm is not None else None,
+            )
+            if limit is not None
         ]
         return min(limits) if limits else None
 
--- a/tests/sdk/context/condenser/test_llm_summarizing_condenser.py
+++ b/tests/sdk/context/condenser/test_llm_summarizing_condenser.py
@@ -387,6 +387,21 @@ def test_token_limit_inherits_agent_effective_input_limit(
     assert Reason.TOKENS in reasons
 
 
+def test_token_reduction_uses_agent_effective_input_limit(mock_llm: LLM) -> None:
+    condenser = LLMSummarizingCondenser(
+        llm=mock_llm, max_size=1000, max_tokens=500, keep_first=2
+    )
+    agent_llm = MagicMock(spec=LLM)
+    agent_llm.effective_max_input_tokens = 100
+    agent_llm.get_token_count.return_value = 200
+
+    view = View.from_events([message_event("event") for _ in range(10)])
+
+    forgotten_events, _ = condenser._get_forgotten_events(view, agent_llm=agent_llm)
+
+    assert forgotten_events
+
+
 def test_condense_with_request_and_events_reasons(mock_llm: LLM) -> None:
     """Test condensation when both REQUEST and EVENTS reasons are true simultaneously.
 
--- a/tests/sdk/context/condenser/test_llm_summarizing_condenser.py
+++ b/tests/sdk/context/condenser/test_llm_summarizing_condenser.py
@@ -277,6 +277,7 @@ def test_condense_with_agent_llm(mock_llm: LLM) -> None:
     # Create a separate mock for the agent's LLM
     agent_llm = MagicMock(spec=LLM)
     agent_llm.model = "gpt-4"
+    agent_llm.effective_max_input_tokens = None
 
     # Prepare a view that triggers condensation
     events: list[Event] = [message_event(f"Event {i}") for i in range(12)]
@@ -307,6 +308,7 @@ def test_condense_with_token_limit_exceeded(mock_llm: LLM) -> None:
     # Create a separate mock for the agent's LLM with token counting
     agent_llm = MagicMock(spec=LLM)
     agent_llm.model = "gpt-4"
+    agent_llm.effective_max_input_tokens = None
 
     # Mock get_token_count to return predictable values based on message content length
     def mock_token_count(messages, **_kwargs):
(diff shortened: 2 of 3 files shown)


VascoSch92:
The idea is right, but the PR caps the trigger threshold in `get_condensation_reasons` without updating the reduction logic in `_get_forgotten_events`, which still reads the raw `self.max_tokens`.

Could we not  compute the cap once and use it in both places:

```python
def _effective_max_tokens(self, agent_llm: LLM | None) -> int | None:
    limits = [t for t in (
        self.max_tokens,
        agent_llm.effective_max_input_tokens if agent_llm else None,
    ) if t is not None]
    return min(limits) if limits else None
```
or something like this?

Then in `_get_forgotten_events`, assert on that value and compute `tokens_to_reduce` from it.

neubig:
Addressed in commit `1a95871`. Added `_effective_max_tokens(agent_llm)` and now use the same capped value in both `get_condensation_reasons()` and `_get_forgotten_events()`. This ensures the reduction target matches the trigger threshold. Added regression coverage for the reduction path; 90 focused condenser/context-window tests pass.

This comment was generated by an AI agent (OpenHands) on behalf of the user.

juanmichelini (changes requested):
There is a buggy test that predates this PR
that says

"""An explicit ``max_tokens=None`` must disable token-based condensation
    even if the agent LLM has a configured ``max_input_tokens``.
    """

By setting `max_tokens = self._effective_max_tokens(agent_llm)`  we are enabling condesantion even on `max_tokens = None`.

If intended could we re move the test? If not, could we add the missing assertion to the test and modify PR to match it?

neubig:
Addressed in commit `909d9a0b`. The test now verifies that an explicit `max_tokens=None` remains unset on the condenser configuration, while documenting that the active agent LLM’s effective input limit still governs runtime condensation. This preserves the intended behavior of following the LM max input tokens.

This comment was generated by an AI agent (OpenHands) on behalf of the user.

VascoSch92:
For me is good. 

WDYT @juanmichelini ?

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -93,6 +93,17 @@ def _disable_streaming_for_summary(self):
     def handles_condensation_requests(self) -> bool:
         return True
 
+    def _effective_max_tokens(self, agent_llm: LLM | None) -> int | None:
+        limits = [
+            limit
+            for limit in (
+                self.max_tokens,
+                agent_llm.effective_max_input_tokens if agent_llm is not None else None,
```
enyst: Why would `agent_llm` ever be None?

neubig: `agent_llm` is optional throughout the public condenser API (`condense(view, agent_llm=None)`, `get_condensation_reasons(view, agent_llm=None)`, and `_get_forgotten_events`, which forwards the same optional arg). `_effective_max_tokens` is invoked from `get_condensation_reasons`, so when a caller only needs request/event-count pressure assessment and passes no agent LLM, we should not dereference it. The guard simply omits the agent LLM's input limit in that case (only the condenser's own `max_tokens` applies).

I've documented this explicitly on `_effective_max_tokens` in commit `b4d705482` so the None-handling is self-explanatory. Behavior is unchanged. Thanks for the catch!

neubig:
[iterate-watchdog] Review feedback addressed.

- @juanmichelini's review concern (stale `max_tokens=None` test) is resolved in the current head: the old test asserting `max_tokens=None` disables token-based condensation was rewritten as `test_llm_summarizing_condenser_explicit_none_max_tokens_remains_unset`, documenting the new intended behavior where the runtime effective limit is governed by `_effective_max_tokens`. The `agent.condenser.max_tokens is None` assertion still holds.
- The only unresolved review thread (from @enyst) has been replied to and resolved; the `unresolved-review-threads` gate now passes.
- All CI checks are green on head `b4d705482`.

Re-requesting review from the review bot on the latest head.

openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py
```diff
@@ -93,6 +93,25 @@ def _disable_streaming_for_summary(self):
...
+        """
+        agent_limit = (
+            agent_llm.effective_max_input_tokens if agent_llm is not None else None
+        )
+        limits = [
+            limit for limit in (self.max_tokens, agent_limit) if limit is not None
+        ]
+        return min(limits) if limits else None
+
     def get_condensation_reasons(
```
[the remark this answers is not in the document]
neubig: Addressed in 53231f68e — inlined the `agent_limit` local per this suggestion. No behavior change; ruff check + format pass.

neubig:
PR needs human approval to proceed. The only remaining blocker is the earlier CHANGES_REQUESTED from juanmichelini (concern about the stale `max_tokens=None` test — the test was rewritten at head b4d70548 and that thread is resolved). All CI checks on head (53231f68) are green, all-hands-bot approved the latest commit, and 0 review threads remain unresolved. Since the reviewer's status decision cannot be changed by an agent, this needs a human to re-review/approve.

enyst (approved):

juanmichelini (approved):
LGTM!

PR #4472, opened 2026-08-12
fix(file_editor): give str_replace snippets the full leading context window

### LINKED ISSUE #4469: [Bug]: `FileEditor.str_replace` starts its verification snippet one line late

### Bug Description

In `openhands-tools/openhands/tools/file_editor/editor.py`, `FileEditor.str_replace` starts its returned verification snippet one line later than the configured context boundary.

With `SNIPPET_CONTEXT_WINDOW = 4` and a replacement on line 6, the snippet begins at line 3 instead of line 2. It therefore contains only three preceding lines while retaining four following lines.

### Expected Behavior

Given an 11-line file and a replacement on line 6, the returned verification snippet is expected to include lines 2 through 10: four preceding lines, the replacement line, and four following lines.

The current calculation retains four following lines. Because `replacement_line` and the inclusive `read_file` bounds are both one-based, the corresponding lower boundary for a four-line preceding context is `max(1, replacement_line - SNIPPET_CONTEXT_WINDOW)`.

### Actual Behavior

The returned verification snippet begins at line 3 and includes meaningful content through line 10:

```text
     3	line 3
     4	line 4
     5	line 5
     6	replacement line
     7	line 7
     8	line 8
     9	line 9
    10	line 10
```

Line 2 is omitted, leaving only three lines of preceding context.

### Steps to Reproduce

1. Check out Software Agent SDK main at commit `f09e03eac772290feeb51b7d7390ffaefeca1a09`.

2. Add the following test to `tests/tools/file_editor/test_basic_operations.py`:

```python
def test_str_replace_includes_full_preceding_context_window(tmp_path):
    editor = FileEditor()
    test_file = tmp_path / "context.txt"
    test_file.write_text("\n".join(f"line {line}" for line in range(1, 12)))

    result = editor(
        command="str_replace",
        path=str(test_file),
        old_str="line 6",
        new_str="replacement line",
    )

    assert result.text is not None
    assert "     2\tline 2" in result.text
    assert "    10\tline 10" in result.text
```

3. Run:

```bash
uv run pytest tests/tools/file_editor/test_basic_operations.py -q
```

4. Observe that the newly added assertion fails while the 41 existing tests pass.

### Installation Method

Source checkout; focused pytest reproduction

_No response_

### SDK Version

main@f09e03eac772290feeb51b7d7390ffaefeca1a09 (openhands-sdk 1.42.0; openhands-tools 1.42.0)

### Version Confirmation

### Python Version

3.13.2

Not applicable; reproduced by directly invoking deterministic file-editor code

### Operating System

MacOS

### Logs and Error Messages

```text
FAILED tests/tools/file_editor/test_basic_operations.py::test_str_replace_includes_full_preceding_context_window
AssertionError: assert '     2\tline 2' in result.text

1 failed, 41 passed
```

### Minimal Code Sample

```python
from pathlib import Path

from openhands.tools.file_editor.editor import FileEditor

with TemporaryDirectory() as directory:
    path = Path(directory) / "context.txt"
    path.write_text("\n".join(f"line {line}" for line in range(1, 12)))

    result = FileEditor()(
        command="str_replace",
        path=str(path),
        old_str="line 6",
        new_str="replacement line",
    )

    print(result.text)
```

At the tested commit, `replacement_line` is calculated as a one-based line number:

```python
replacement_line = file_content.count("\n", 0, match.start()) + 1
```

The snippet start is then calculated and passed to the one-based `read_file` interface as follows:

```python
start_line = max(0, replacement_line - SNIPPET_CONTEXT_WINDOW)
snippet = self.read_file(
    path,
    start_line=start_line + 1,
    end_line=end_line,
)
```

For a replacement on line 6 with a context window of 4, `start_line` is 2, but `read_file` receives 3. The returned snippet therefore contains lines 3 through 10: three preceding lines and four following lines.

One possible fix direction is to calculate a one-based lower boundary and pass it directly to `read_file`:

```python
start_line = max(1, replacement_line - SNIPPET_CONTEXT_WINDOW)
snippet = self.read_file(path, start_line=start_line, end_line=end_line)
```

The same `start_line` should then be used when numbering the returned snippet. Regression coverage should include replacements away from both file boundaries and verify the configured amount of context on each side.

onatozmenn:
HUMAN:

I reproduced the `str_replace` snippet issue and confirmed the leading context was one line short. The fix makes the context window symmetric without changing `insert`, which already behaves correctly.

---

AGENT:

## Why

After a `str_replace`, the editor hands the model a verification snippet around the edit.
`replacement_line` is one-based and `read_file` takes one-based inclusive bounds, but the
lower bound was computed as `max(0, replacement_line - SNIPPET_CONTEXT_WINDOW)` and then
passed as `start_line + 1`. The upper bound has no such offset, so with
`SNIPPET_CONTEXT_WINDOW = 4` the model got three lines of leading context and four trailing
ones.

## Summary

- Compute the snippet's lower bound as a one-based line and pass it to `read_file` unchanged
- Number the snippet from that same line

## Issue Number

Fixes #4469

## How to Test

```python
from pathlib import Path

from openhands.tools.file_editor.editor import FileEditor

with TemporaryDirectory() as directory:
    path = Path(directory) / "context.txt"
    path.write_text("\n".join(f"line {n:02d}" for n in range(1, 22)))

    result = FileEditor()(
        command="str_replace",
        path=str(path),
        old_str="line 10",
        new_str="replacement line",
    )
    print(result.text)
```

On `main` at `5bfa7fc` the snippet starts at line 7:

```
The file context.txt has been edited. Here's the result of running `cat -n` on a snippet of context.txt:
     7	line 07
     8	line 08
     9	line 09
    10	replacement line
    11	line 11
    12	line 12
    13	line 13
    14	line 14
```

On this branch it starts at line 6, so the window is symmetric:

```
The file context.txt has been edited. Here's the result of running `cat -n` on a snippet of context.txt:
     6	line 06
     7	line 07
     8	line 08
     9	line 09
    10	replacement line
    11	line 11
    12	line 12
    13	line 13
    14	line 14
```

Unit coverage:

```
uv run pytest tests/tools/file_editor
```

155 passed, 2 skipped. The new parametrized test checks an edit away from both file
boundaries and an edit near the top, where the window is clamped rather than shifted.
Reverting the source change while keeping the tests fails only the first case.

## Video/Screenshots

Not applicable: the snippet output above is the user-visible effect.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

I checked `insert` before touching anything, expecting the same arithmetic there. It is
fine: `insert_line` is the line the new content goes *after*, so the inserted block starts
at `insert_line + 1` and the existing `max(1, insert_line - SNIPPET_CONTEXT_WINDOW + 1)`
already yields four lines on each side. Only `str_replace` is off, so this leaves `insert`
alone.

Unrelated cosmetic detail I noticed while measuring, not touched here: the snippet ends with
a newline, so `_make_output` numbers one trailing empty line (the bare `15` in the output
above). Happy to send that separately if it is worth fixing.

### DIFF
--- a/openhands-tools/openhands/tools/file_editor/editor.py
+++ b/openhands-tools/openhands/tools/file_editor/editor.py
@@ -255,16 +255,16 @@ class FileEditor:
         self._history_manager.add_history(path, file_content)
 
         # Create a snippet of the edited section
-        start_line = max(0, replacement_line - SNIPPET_CONTEXT_WINDOW)
+        start_line = max(1, replacement_line - SNIPPET_CONTEXT_WINDOW)
         end_line = replacement_line + SNIPPET_CONTEXT_WINDOW + new_str.count("\n")
 
         # Read just the snippet range
-        snippet = self.read_file(path, start_line=start_line + 1, end_line=end_line)
+        snippet = self.read_file(path, start_line=start_line, end_line=end_line)
 
         # Prepare the success message
         success_message = f"The file {path} has been edited. "
         success_message += self._make_output(
-            snippet, f"a snippet of {path}", start_line + 1
+            snippet, f"a snippet of {path}", start_line
         )
 
         success_message += (
--- a/tests/tools/file_editor/test_basic_operations.py
+++ b/tests/tools/file_editor/test_basic_operations.py
@@ -378,6 +378,39 @@ Review the changes and make sure they are as expected. Edit the file again if ne
     assert "This is a sample file." in test_file.read_text()
 
 
+@pytest.mark.parametrize(
+    ("replaced_line", "first_line", "last_line"),
+    [
+        (10, 6, 14),
+        # Near the top of the file the window is clamped instead of shifted.
+        (2, 1, 6),
+    ],
+)
+def test_str_replace_snippet_keeps_the_context_window_on_both_sides(
+    tmp_path, replaced_line, first_line, last_line
+):
+    editor = FileEditor()
+    test_file = tmp_path / "context.txt"
+    test_file.write_text("\n".join(f"line {number:02d}" for number in range(1, 22)))
+
+    result = editor(
+        command="str_replace",
+        path=str(test_file),
+        old_str=f"line {replaced_line:02d}",
+        new_str="replacement line",
+    )
+
+    assert result.text is not None
+    for number in range(first_line, last_line + 1):
+        expected = (
+            "replacement line" if number == replaced_line else f"line {number:02d}"
+        )
+        assert f"{number:6}\t{expected}" in result.text
+    if first_line > 1:
+        assert f"{first_line - 1:6}\tline {first_line - 1:02d}" not in result.text
+    assert f"{last_line + 1:6}\tline {last_line + 1:02d}" not in result.text
+
+
 def test_str_replace_multi_line_no_linting(editor):
     editor, test_file = editor
     result = editor(


onatozmenn:
Still ready on my side, fix is minimal with regression coverage, happy to address any feedback

PR #4473, opened 2026-08-12
fix(llm): stop serializing calls through global config

### LINKED ISSUE #16459: Global threading.Lock serializes LLM calls across all conversations — wastes local model compute
## Problem

When running multiple conversations simultaneously with a **local model** (vLLM, Ollama, llama.cpp, etc.), all LLM requests are serialized through a `ClassVar[threading.Lock]` in `openhands.sdk.llm.LLM`, resulting in significant idle GPU time and poor performance.

## Root Cause

In `openhands/sdk/llm/llm.py` (line ~632):

```python
_litellm_modify_params_lock: ClassVar[threading.Lock] = threading.Lock()
```

This lock is held for the **entire duration** of each LLM call (to protect the global `litellm.modify_params` boolean), which means:

1. Conversation A starts LLM call -> acquires lock -> waits for response (e.g., 30-120s for local model)
2. Conversation B and C **queue behind** -> acquire lock after A releases -> each waits 30-120s
3. Total time for 3 parallel conversations: ~90-360s instead of ~30-120s

The `ClassVar` modifier means this lock is **shared across all LLM instances** in the process - not just per-conversation.

## Why This Matters for Local Models

| Scenario | Cloud API (GPT-4) | Local Model (Llama 3) |
|---|---|---|
| One request | ~2-5s | ~30-120s |
| 3 conversations serialized | ~6-15s | ~90-360s |
| 3 conversations parallel | ~2-5s | ~30-120s |

With cloud APIs, serialization overhead is barely noticeable. With local models, **the GPU sits idle** while other conversations wait on the lock - wasting both time and energy.

## Why It Exists

LiteLLM uses a **process-wide global** (`litellm.modify_params`) - a single `True`/`False` boolean at the module level. OpenHands temporarily sets this flag around each request to enable/disable LiteLLM's parameter transformations (e.g., inserting dummy messages for Anthropic/Bedrock). The lock guards against race conditions when multiple threads modify this global simultaneously.

### Option 1: `contextvars` (Recommended - minimal change)

Replace the global lock with per-coroutine context-local storage:

```python
import contextvars
_modify_params_ctx: contextvars.ContextVar[bool] = contextvars.ContextVar(
    'modify_params', default=False
)
```

Each async task (conversation) gets its own isolated value. No lock needed. No serialization. This is the Python-native solution for async isolation.

### Option 2: `threading.local()` (Thread-isolated)

```python
import threading
_tls = threading.local()
```

Each worker thread gets its own copy. Works if LiteLLM checks the same thread-local during the call.

### Option 3: Upstream LiteLLM fix

Contribute to LiteLLM to make `modify_params` a per-call parameter rather than a module-level global. This would be the most robust long-term fix.

### Option 4: Bypass LiteLLM for local models

For `local_ai/` provider calls, skip LiteLLM's parameter modification entirely and go directly to the API endpoint. Local models do not need Anthropic/Bedrock message transformations.

## Impact

- **Local model users:** severe performance degradation with multiple conversations
- **Cloud API users:** minimal impact (short requests, barely noticeable)
- **Hardware utilization:** GPU sits idle during lock wait times
- **User experience:** feels like single-threaded when running 3+ conversations

## Workaround (Temporary)

Set `modify_params: false` in LLM settings - though this may break some provider-specific transformations. Note: even with `false`, the lock is still acquired and released, so this only helps if the lock-hold time is negligible (not the case for local models where the lock is held during the full API call).

neubig:
HUMAN:

I confirmed that things seem to be running in parallel after this fix.

Screenshot 2026-08-12 at 7 24 43 AM

---

AGENT:

## Why

`LLM._litellm_modify_params_lock` protects LiteLLM's module-global `modify_params` setting by holding one process-wide lock for the complete provider request, including retries and streamed responses. That preserves mixed per-object values but serializes every conversation's LLM I/O. This is the bottleneck reported in OpenHands/OpenHands#16459.

LiteLLM exposes the setting as process-wide state, so this change makes the SDK model match that architecture: parameter modification is enabled once for the process and is no longer changed per request.

## Summary

- Enable `litellm.modify_params` once at module initialization.
- Keep `LLM.modify_params` as a serialized compatibility field, deprecate explicit use through the SDK warning mechanism (v1.42.0 → v1.47.0), and stop treating it as request-local configuration.
- Remove the process-wide lock, dedicated lock executor, and cancellation machinery so LLM calls can overlap.
- Replace serialization tests with a transport-level concurrency regression test.

## Issue Number

Fixes OpenHands/OpenHands#16459.

## How to Test

Unit and static validation:

```bash
uv run pytest tests/sdk/llm -q
# 945 passed

  openhands-sdk/openhands/sdk/llm/llm.py \
  tests/sdk/llm/test_llm_completion.py
# all hooks passed
```

Live validation on an agent-server running this branch:

1. Stopped the old v1.40.1 server while 9 conversations were running.
2. Started the v1.42.0 server directly from this checkout.
3. Audited the interrupted jobs and resumed only the 9 newest unique targets; older duplicate attempts for PR #4464, PR #16231, profile-downgrade, and backend-scope work remained paused.
4. All 9 `/run` requests returned 200. Within the first two minutes, 343 new events were produced. Two jobs finished and the other seven continued running.
5. LLM-originated actions from different conversations now arrive concurrently. Examples:

```text
11:14:35.753 e5d78571 terminal
11:14:35.813 81983c60 terminal   # 60 ms apart

11:14:43.551 e5d78571 terminal
11:14:43.629 458ec8d0 terminal   # 79 ms apart
11:14:43.760 c331542e terminal   # 130 ms later

11:16:38.759 2292b2a7 terminal
11:16:38.776 81983c60 terminal   # 18 ms apart
```

Before this change, provider turns were visibly convoyed behind the process-wide lock and the dedicated eight-thread lock executor was saturated with eight blocked `Lock.acquire()` calls. Afterward, the lock executor no longer exists, conversations make progress in parallel, and a 50-request `/health` burst remained responsive (31.6 ms mean, 64.3 ms max) while seven agents were active.

## Video/Screenshots

N/A — backend concurrency fix. Live timing evidence is included above.

## Design Doc

N/A.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

`modify_params=False` remains loadable for persisted/profile compatibility but no longer disables LiteLLM request repair for one LLM instance. Explicit use emits the standard SDK `DeprecatedWarning` and is scheduled for removal in v1.47.0. LiteLLM does not provide a request-local API for this setting; honoring mixed values safely would require serializing those calls again.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:9f2b988-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-9f2b988-python \
  ghcr.io/openhands/agent-server:9f2b988-python
```

```
ghcr.io/openhands/agent-server:9f2b988-golang-amd64
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-golang-amd64
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-golang-amd64
ghcr.io/openhands/agent-server:9f2b988-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:9f2b988-golang-arm64
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-golang-arm64
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-golang-arm64
ghcr.io/openhands/agent-server:9f2b988-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:9f2b988-java-amd64
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-java-amd64
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-java-amd64
ghcr.io/openhands/agent-server:9f2b988-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:9f2b988-java-arm64
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-java-arm64
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-java-arm64
ghcr.io/openhands/agent-server:9f2b988-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:9f2b988-python-amd64
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-python-amd64
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-python-amd64
ghcr.io/openhands/agent-server:9f2b988-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:9f2b988-python-arm64
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-python-arm64
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-python-arm64
ghcr.io/openhands/agent-server:9f2b988-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:9f2b988-golang
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-golang
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-golang
ghcr.io/openhands/agent-server:9f2b988-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:9f2b988-java
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-java
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-java
ghcr.io/openhands/agent-server:9f2b988-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:9f2b988-python
ghcr.io/openhands/agent-server:9f2b98848aa03f5db13023c8b1ef80c59c79c071-python
ghcr.io/openhands/agent-server:fix-process-wide-litellm-modify-params-python
ghcr.io/openhands/agent-server:9f2b988-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `9f2b988-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `9f2b988-python-amd64`) are also available if needed
</details>


openhands-sdk/openhands/sdk/llm/llm.py
```diff
@@ -450,8 +452,10 @@ class LLM(BaseModel, RetryMixin, NonNativeToolCallingMixin):
     drop_params: bool = Field(default=True, json_schema_extra=field_meta())
     modify_params: bool = Field(
```
VascoSch92, line 452: I believe we can deprecate this arguments right here.

VascoSch92 (commented):
LGTM. 

a couple of things which are nits.

I think the deprecation warning should be in the pydantic field, and removing the lock lets `warnings.catch_warnings()` in `_suppress_transport_warnings()` run concurrently.

VascoSch92 (approved):
LGTM. 

a couple of things which are nits.

I think the deprecation warning should be in the pydantic field, and removing the lock lets `warnings.catch_warnings()` in `_suppress_transport_warnings()` run concurrently.

openhands-sdk/openhands/sdk/llm/llm.py
neubig: Addressed in 9f2b98848: `modify_params` now exposes Pydantic field-level deprecation metadata while retaining the SDK runtime warning for explicit input.

neubig:
@VascoSch92 I addressed both nits in the final head `9f2b98848`: field-level Pydantic deprecation metadata is present, and the concurrent transport paths no longer use `warnings.catch_warnings()`. The v1.42.1 release security scan correctly flagged that your approval predates this final commit. Could you please review/approve the final merged head so the release audit has a human approval for the shipped diff?

Issue #4453, opened 2026-08-10
Agent Plugins: path-containment enforcement + narrow failure boundaries


VascoSch92:
Part of OpenHands/software-agent-sdk#4405.

Add a shared safety layer on `PluginFormat` (spec §3/§4.1): every package-resolved path must stay within the filesystem-resolved plugin root after resolving symlinks.

### Current state (verified against `main`)

There is no package-boundary enforcement anywhere in the plugin or skills loading path today — confirmed by grepping the whole `openhands-sdk/openhands/sdk/` tree for `symlink` / `is_relative_to` / `containment` / `realpath`; the only hits are unrelated (builtin-prompt-dir resolution, workspace-root symlink handling, skill dedup). This means `ClaudeCodePluginFormat` has the same gap today, not just the new format.

### Scope

* Reject the whole plugin if `plugin.json` escapes the root.
* Disable a component type if its fixed location (`skills/`, `mcp.json`) escapes the root or has the wrong filesystem kind.
* Skip one skill if its `SKILL.md` escapes the root.
* Skip one MCP entry if its `command`/`cwd` escapes the root.
* Deny other escaping package-supplied paths.
* Apply this on the shared `PluginFormat` base class so both `ClaudeCodePluginFormat` and `AgentPluginsFormat` get it. Per-skill/agent/command failure isolation already exists (`format/base.py:154-160`, `format/claude_code.py:170-175,188-193`) and should compose with this rather than being replaced.

### Refs

* [https://agent-plugins.org/client-implementers/loading-and-discovery](<https://agent-plugins.org/client-implementers/loading-and-discovery>)

## Agent Plugins specification

- [Canonical specification](https://agent-plugins.org/specification)
- [Versioned normative v1.0.0 specification](https://github.com/agentplugins/agent-plugins-spec/blob/main/spec/1.0.0.md)

_This specification context was added by an OpenHands AI agent on behalf of @enyst._

VascoSch92:
Fix in #5101 (draft), stacked on #5093.

Together they cover the scope above: an escaping `plugin.json` rejects the plugin; an escaping or wrong-kind `skills/` / MCP config disables that component; an escaping `SKILL.md` or MCP `command`/`cwd` skips that one skill or server; other escaping package paths (hooks, agents, commands, skill-level `.mcp.json`, skill resources) are denied. The check lives in one shared `resolves_within()` helper used by both `ClaudeCodePluginFormat` and `AgentPluginsFormat`.

[...]

Issue #4451, opened 2026-08-10
Agent Plugins: mcp.json loader + PLUGIN_ROOT/PLUGIN_DATA expansion


VascoSch92:
Part of #4405.

Add `AgentPluginsFormat.load_mcp_config()` — read a root-level `mcp.json` (no dot) with Agent Plugins' variable-expansion semantics.

### Blocked on open questions

#4405's Open Questions #6 (PLUGIN_DATA location/lifecycle) and #7 (`MCPServer.headers` typed as `SecretStr`, conflicting with plugin-declared headers being literal non-secret data) should be resolved before starting this one.

### Current state (verified against `main`)

The Claude Code loader only reads `.mcp.json` (`format/claude_code.py:118`). The existing variable expander (`expand_variable_references`, `skills/utils.py:39-122`) recurses into dict keys and falls back to `os.environ` (`check_env=True`, `utils.py:210`) — broader than what the spec allows, so this needs new, narrowly-scoped expansion code rather than an extension of the shared expander. `MCPServer.headers` is `dict[str, SecretStr]` (`mcp/config.py:512`).

### Scope

- Vendor `mcp.schema.json`; validate the closed top-level document and each server entry independently.
- Support `stdio` and `streamable-http` transports (`MCPTransport` already includes both as literals — `mcp/config.py:493`); `sse` is optional per open question #1.
- Implement `${PLUGIN_ROOT}` / `${PLUGIN_DATA}` expansion: textual, single-pass, non-recursive, restricted to `args` / `env` values / `cwd` only — never `command`, env keys, remote URLs, or headers. Client-controlled values injected last, non-overridable by the plugin.
- Define and implement `PLUGIN_DATA`: a dedicated writable directory per plugin, persisted across updates.
- Resolve the `MCPServer.headers`-as-`SecretStr` conflict — plugin-declared headers need to stay literal/visible, not secrets.
- Remote URL/header validation: absolute HTTP(S), no user-info/fragments, non-loopback must be HTTPS, headers are literal (no credentials).
- Failure isolation: invalid `mcp.json` disables MCP for the plugin only; an invalid/unavailable individual server disables only that entry.

### Refs

- https://agent-plugins.org/plugin-authors/mcp-servers

## Agent Plugins specification

- [Canonical specification](https://agent-plugins.org/specification)
- [Versioned normative v1.0.0 specification](https://github.com/agentplugins/agent-plugins-spec/blob/main/spec/1.0.0.md)

[...]

Issue #4405, opened 2026-08-06
Spec: Support the Agent Plugins (agent-plugins.org) portable package format


VascoSch92:
## Background

[Agent Plugins](<https://agent-plugins.org>) is a new open, vendor-neutral standard (v1.0.0, Working Draft) for packaging reusable components that extend AI agents into portable plugins. Its initial Technical Steering Committee includes Core Maintainers from Amazon, Cursor, Microsoft, OpenAI, and Vercel. Compatible clients today include VS Code, Cursor, GitHub Copilot, ChatGPT & Codex, and Kiro — OpenHands is **not** yet in that list.

The standard defines a small interoperability floor on top of two existing specs we already consume:

* [Agent Skills](<https://agentskills.io/specification>) — the `SKILL.md` format we already load
* [MCP servers](<https://modelcontextprotocol.io/specification>) — already supported in OpenHands

So the heavy lift is mostly about the **package/manifest layout and discovery contract**, not new component technology.

## The portable package

An Agent Plugin is a directory with a required manifest and optional components in fixed locations:

```text
my-plugin/
├── plugin.json            # required manifest (root-level, not nested)
├── skills/                # immediate child dirs each containing SKILL.md
│   └── summarize/
│       ├── SKILL.md
│       ├── scripts/
│       └── references/
├── mcp.json               # root-level MCP config (NOT .mcp.json)
└── com.example.client/    # reverse-domain client extension namespace
    └── hooks/
```

### How this differs from our current (Claude Code-style) layout

OpenHands currently follows the Claude Code plugin structure (`openhands.sdk.plugin`):

| Concern | OpenHands today | Agent Plugins v1.0.0 |
| -- | -- | -- |
| Manifest location | `.claude-plugin/plugin.json` or `.plugin/plugin.json` (nested) | `plugin.json` at plugin root |
| Manifest schema | Open `extra="allow"`, our own fields | Closed schema; only `$schema`, `name`, `version`, `description`, `author`, `homepage`, `repository`, `license`, `keywords`, `extensions`; unknown top-level fields reported-and-ignored (non-fatal) |
| MCP config file | `.mcp.json` (dotfile, root) | `mcp.json` (no dot, root) |
[...]

PR #4474, opened 2026-08-12
feat(plugin): add Agent Plugins manifest loader (root plugin.json, closed schema)

### LINKED ISSUE #4450: Agent Plugins: manifest loader (root plugin.json, closed schema)
Part of #4405.

Add `AgentPluginsFormat.load_manifest()` — parse a root-level `plugin.json` against the vendored, closed Agent Plugins schema, selected by `$schema` and never fetched over the network.

### Current state (verified against `main`)

`PluginManifest` (`types.py:240-256`) is `extra="allow"` (open schema), has no `$schema` field, and `name` has zero format validation (`types.py:243`). The Claude Code loader only checks `.plugin/plugin.json` / `.claude-plugin/plugin.json` (`format/claude_code.py:34,71-75`) — never a root `plugin.json`. So a real Agent Plugins package is currently loaded with an inferred fake manifest, not read.

### Scope

- Vendor `plugin.schema.json` (https://agent-plugins.org/schemas/1.0.0/plugin.schema.json) locally; select it via `$schema`; reject unsupported/missing `$schema`.
- Closed-schema validation: fatal reject on any violation except unknown top-level fields and a non-object `extensions` (report + ignore).
- `name` constraints: 1–64 chars, lowercase ASCII letters/digits/hyphens/periods, no `--` or `..`, alphanumeric start/end.
- Leave OpenHands-specific manifest concepts (`entry_command`, commands, agents) out of the closed schema itself — those are mapped via the client-extension namespace (separate sub-issue).

### Out of scope

`mcp.json` loading, client extensions, path containment — tracked in sibling issues under #4405.

### Refs

- https://agent-plugins.org/plugin-authors/manifest
- https://agent-plugins.org/client-implementers/conformance

VascoSch92:
HUMAN:

Introducing Agent Plugin Format

---

AGENT:

## Why

An Agent Plugins package (agent-plugins.org) keeps its manifest at the plugin **root** as `plugin.json`. Our loader only ever looked in `.plugin/` and `.claude-plugin/` (`format/claude_code.py:34,71-75`), so a real Agent Plugins package fell through to the Claude Code fallback and was loaded under a manifest **inferred from its directory name** — the package's own identity, version and metadata were never read.

`PluginManifest` is also an open schema (`extra="allow"`) with no `$schema` field and no format validation on `name`, so nothing enforced the standard's constraints even when a manifest was present.

This adds the manifest half of conformance: locate the root manifest, validate it against a locally vendored closed schema selected by `$schema`, and apply the specification's fatal / non-fatal split.

## Summary

- Adds `AgentPluginsFormat`, the `PluginFormat` strategy for the agent-plugins.org layout, with presence-only `detect()` and a closed-schema `load_manifest()`. Vendors `plugin-1.0.0.schema.json` byte-for-byte and selects it by `$schema` — never fetched over the network.
- Fatal violations reject the plugin; the two the spec marks non-fatal are reported and ignored (unknown top-level fields, non-object `extensions`). Known fields and the `name` pattern are both derived from the vendored schema, so there is no second source of truth to drift.
- `PluginAuthor.name` now defaults to `""`: the Agent Plugins schema permits an author object with no name, and requiring it rejected conformant manifests with a raw pydantic error.

## Issue Number

Closes #4450. Part of #4405.

## How to Test

**End-to-end against the official example package** — not a fixture, the real thing:

```bash
git clone --depth 1 https://github.com/agentplugins/agent-plugins-example /tmp/ap-example
uv run python -c "
from openhands.sdk.plugin import AgentPluginsFormat, detect_format
p = AgentPluginsFormat().load('/tmp/ap-example')
print(p.manifest.name, p.manifest.version, [s.name for s in p.skills])
"
```

Full run, including the rejection path on a tampered copy:

```
detect(real package): True
detect_format() today: ClaudeCodePluginFormat (unregistered on purpose)

manifest.name        : agent-plugins-example
manifest.version     : 1.0.0
manifest.description : A copyable reference package and migration guide for Agent Plugins v1.0.0.
extras kept          : license='MIT' keywords=['agent-plugins', 'example', 'migration', 'skills']
$schema              : https://agent-plugins.org/schemas/1.0.0/plugin.schema.json

skills               : ['migrate-agent-plugin']
skill references     : ['client-extensions.md', 'migration-guide.md', 'validation-checklist.md']
deferred loaders     : mcp={} hooks=None agents=[] commands=[]

tampered name rejected: 'Bad--Name' does not match '^(?!.*(?:--|\\.\\.))[a-z0-9](?:[a-z0-9.-]*[a-z0-9])?$'
```

On `main` this same package loads as name `ap-example` (the directory name) with version `1.0.0` and description `Plugin loaded from ap-example` — its real manifest never read.

**Packaging.** The vendored schema is new package data, so the wheel was built and inspected rather than assumed:

```bash
uv build --package openhands-sdk --wheel --out-dir /tmp/wheelcheck
python3 -c "
import zipfile,glob
n=zipfile.ZipFile(glob.glob('/tmp/wheelcheck/*.whl')[0]).namelist()
print([x for x in n if x.endswith('.schema.json')])
print(len([x for x in n if x.endswith('.j2')]),'j2 templates still shipped')
"
# ['openhands/sdk/plugin/format/schemas/plugin-1.0.0.schema.json']
# 3 j2 templates still shipped
```

**Test suites** (marketplace and API-breakage included because `PluginAuthor` is shared):

```bash
uv run pytest tests/sdk/plugin/ tests/sdk/marketplace/ \
  tests/sdk/conversation/test_local_conversation_plugins.py \
  tests/cross/test_check_sdk_api_breakage.py -q
# 375 passed
uv run pre-commit run --files <changed files>   # all hooks pass
```

**Test quality.** `pytest --cov` cannot run in this repo — `coverage` and the beartype import hook installed by `tests/conftest.py` deadlock on a circular import (pre-existing, unrelated to this PR; `--noconftest` does not avoid it). Instead each load-bearing line was mutated and the suite checked for a failure. All eight mutants died:

| Mutation | Result |
| --- | --- |
| `detect()` → always `True` | CAUGHT |
| skip the unknown-field strip | CAUGHT |
| skip the non-object `extensions` strip | CAUGHT |
| skip `Draft202012Validator.validate` | CAUGHT |
| `utf-8-sig` → `utf-8` | CAUGHT |
| `except (OSError, UnicodeDecodeError)` → `except OSError` | CAUGHT |
| remove the `$schema` gate | CAUGHT |
| make `PluginAuthor.name` required again | CAUGHT |

64 tests in the new file: 7 detection, 13 valid-manifest, 6 non-fatal, ~32 fatal-rejection, 6 assembly and schema-drift.

## Video/Screenshots

N/A — library change with no UI. Terminal output above is the end-to-end evidence.

## Design Doc

N/A. The design was proposed and discussed on #4450 before implementation, including the two open questions called out below.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**Two decisions worth a reviewer's attention** — both were raised as open questions on #4450 because #4405's ADR deliverable is still unchecked:

1. **`detect()` is presence-only** (`plugin.json` exists), not gated on `$schema`. Gating on `$schema` looks safer but silently reintroduces the bug this PR fixes: an unparseable or `$schema`-less root manifest would fail detection, fall through to Claude Code, and load under a fabricated identity. Presence-only makes every rejection explicit. The cost is that a directory with an unrelated root `plugin.json` is claimed and hard-rejected instead of falling back; there are no such fixtures in this repo and the Claude Code layout is nested by definition.

2. **Not registered in `_FORMATS`.** `mcp.json` (#4451) is blocked on two unresolved questions in #4405 — where `PLUGIN_DATA` lives, and `MCPServer.headers` being typed `dict[str, SecretStr]` (`mcp/config.py:512`) versus Agent Plugins' literal non-secret headers. Registering now would claim any root-`plugin.json` directory and load it with silently zero MCP servers. `test_not_registered_yet` pins this so the change is visible when someone flips it.

**Shared-model change.** `PluginAuthor.name` defaulting to `""` also loosens the Claude Code path: a manifest with `"author": {}` now loads instead of failing. Strictly more tolerant, same conformance argument.

**Known, left as-is:**

- `jsonschema` applies `pattern` with `re.search` and Python's `$` matches before a trailing newline, so `"name": "ok\n"` passes the vendored pattern. Upstream's schema behaves identically, so this is bug-compatible rather than a local divergence.
- A non-object value *inside* an `extensions` namespace is fatal here. The conformance checklist names only two non-fatal cases, but #4405 §6 says to ignore unimplemented namespaces "without validating their contents" — worth confirming during the ADR.
- `detect()` uses `is_file()`, which follows symlinks; path containment is #4453.
- `_load_schema` is `@cache`d and hands every caller the same dict. Nothing mutates it today (`set(schema["properties"])` and `Draft202012Validator(schema)` are read-only), but a future mutating caller would corrupt it process-wide.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:5cb606f-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-5cb606f-python \
  ghcr.io/openhands/agent-server:5cb606f-python
```

```
ghcr.io/openhands/agent-server:5cb606f-golang-amd64
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-golang-amd64
ghcr.io/openhands/agent-server:vasco-plugin-manifest-golang-amd64
ghcr.io/openhands/agent-server:5cb606f-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:5cb606f-golang-arm64
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-golang-arm64
ghcr.io/openhands/agent-server:vasco-plugin-manifest-golang-arm64
ghcr.io/openhands/agent-server:5cb606f-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:5cb606f-java-amd64
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-java-amd64
ghcr.io/openhands/agent-server:vasco-plugin-manifest-java-amd64
ghcr.io/openhands/agent-server:5cb606f-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:5cb606f-java-arm64
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-java-arm64
ghcr.io/openhands/agent-server:vasco-plugin-manifest-java-arm64
ghcr.io/openhands/agent-server:5cb606f-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:5cb606f-python-amd64
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-python-amd64
ghcr.io/openhands/agent-server:vasco-plugin-manifest-python-amd64
ghcr.io/openhands/agent-server:5cb606f-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:5cb606f-python-arm64
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-python-arm64
ghcr.io/openhands/agent-server:vasco-plugin-manifest-python-arm64
ghcr.io/openhands/agent-server:5cb606f-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:5cb606f-golang
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-golang
ghcr.io/openhands/agent-server:vasco-plugin-manifest-golang
ghcr.io/openhands/agent-server:5cb606f-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:5cb606f-java
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-java
ghcr.io/openhands/agent-server:vasco-plugin-manifest-java
ghcr.io/openhands/agent-server:5cb606f-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:5cb606f-python
ghcr.io/openhands/agent-server:5cb606fcffaa25b69dacb439f049cbddaa9f245e-python
ghcr.io/openhands/agent-server:vasco-plugin-manifest-python
ghcr.io/openhands/agent-server:5cb606f-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `5cb606f-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `5cb606f-python-amd64`) are also available if needed
</details>


openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
```diff
@@ -0,0 +1,160 @@
...
+
+logger = get_logger(__name__)
+
+# At the plugin root, not nested as in the Claude Code layout.
+MANIFEST_FILE: Final[str] = "plugin.json"
+
+_SCHEMAS_DIR: Final[Path] = Path(__file__).parent / "schemas"
+
+#: Supported ``$schema`` values -> vendored schema file.
+VENDORED_SCHEMAS: Final[dict[str, str]] = {
```
enyst: should this be a single one, not more?

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
```diff
@@ -0,0 +1,160 @@
...
+    ├── plugin.json              # required manifest (root-level, closed schema)
+    ├── skills/                  # Agent Skills (optional)
+    ├── mcp.json                 # MCP servers, no leading dot (optional)
+    └── com.example.client/      # reverse-domain client extensions (optional)
+    ```
+
+    Only the manifest and skills are read today (skills via the shared
+    :meth:`PluginFormat.load_skills`). ``mcp.json`` and OpenHands' client
+    extension are follow-ups under #4405, so their loaders return empty --
+    spec-correct, since a client must ignore extensions it does not implement.
```
enyst, line 63: should we simply accept an open schema? so we can load other stuff, our extension. I seem to recall that we can have other stuff in CC plugins.

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
```diff
@@ -0,0 +1,160 @@
...
+        Violations are fatal except the two the spec marks non-fatal, which are
+        reported and ignored: unknown top-level fields, and a non-object
+        ``extensions``.
+
+        Fields with no ``PluginManifest`` column (``$schema``, ``homepage``,
+        ``repository``, ``license``, ``keywords``, ``extensions``) survive via
+        ``extra="allow"``.
+
+        Raises:
+            ValueError: On any fatal manifest violation.
```
enyst, line 97: seems overkill if it applies to e.g. an extra file or directory?

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
```diff
@@ -0,0 +1,160 @@
...
+        unknown = sorted(set(data) - set(schema["properties"]))
+        if unknown:
+            logger.warning(
+                "Ignoring unknown top-level field(s) in %s: %s",
+                manifest_path,
+                ", ".join(unknown),
+            )
+            data = {k: v for k, v in data.items() if k not in unknown}
+
+        if "extensions" in data and not isinstance(data["extensions"], dict):
```
enyst, line 133: I'm not sure this needs to be named "extensions". We can just let users add other dirs we will support?

enyst (commented):
Left a few comments. I seem to recall that generally we simply accept things like a skill without frontmatter. That is an OH-specific extension. Do you have strong feelings about this, or can we keep it simple here too?

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
VascoSch92: Today it is a single entry. In fact, the dict has exactly one key. 

It's a mapping rather than a constant because `$schema` is the version selector: the conformance rule is "select locally supported manifest rules from $schema," so the lookup is inherently "is this declared version one I have vendored?" 

A single constant would make that a flag-day where adding the new version drops the old.                                                                        
                                                                                                                                                                                     
It also does two other jobs already: 
- it generates the Supported: ... half of the rejection message
- the drift test iterates it to assert each vendored file's `$id` matches the URL it's keyed by.                                                                                                                                                                 
                                                                                                                                                                                     
If you'd rather have `SUPPORTED_SCHEMA_URL` + `SCHEMA_FILENAME` now and reintroduce the mapping when a second version exists, I'm fine with that.

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
VascoSch92: Yes you remember correctly about CC plugins.                                                                                                                                                     
                                                                                                                                                                                     
But for Agent Plugins the closed schema isn't a choice we're making: it is the format. 

If we accept an open manifest here, we're implementing a superset rather than the standard: and we want to get onto the compatible-clients list.                                                                                                                                 
                                                                                                                                                                                     
We don't lose the ability to carry our own data. The standard designates a place for exactly that: `extensions.<reverse-domain>` in the manifest, plus top-level `<reverse-domain>/` directories in the package. 

So `entry_command`, `commands`, `agents` and `hooks` move from top-level keys to a namespaced address.

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
VascoSch92: It doesn't. This Raises: covers plugin.json content only

VascoSch92: or I'm missunderstanding something?

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
VascoSch92: Two separate things here. extensions is the manifest key defined by the standard. So we can not decide.                                                                                                       
                                                                                                                                                                                     
Otherwise, the standard already permits it: extra top-level directories are the client-extension mechanism. The only constraint is that they're reverse-domain named, so two clients don't collide on a generic hooks/. So "let 

So users can add other director will be possible with namespace prefix.

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
enyst: We should have only a schema, there aren’t more, and I’m not sure it’s on disk even 🤔 Can you tell your agent KISS and to tell exactly how we do the equivalent in CC plugins?

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
enyst: Ahh, that makes sense actually. Yes, their schema has the reverse domain directory 🤔 

Cool, thank you

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
enyst: Ah, extensions is supposed to be the name? OK. And then clients, including OH, can use `dev.openhands/` or so 🤔 Got it… I _think_! Thanks

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
enyst: Ah, I commented on the wrong thing, meant the other one, which was solved elsewhere.

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
enyst: I’m not totally sure we need any schema variable, just implement the thing in the format defined in the standard. But I could be wrong

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
```diff
@@ -0,0 +1,159 @@
...
+            ) from e
+
+        return PluginManifest.model_validate(data)
+
+    def load_mcp_config(self, plugin_dir: Path) -> dict[str, MCPServer]:  # noqa: ARG002
+        """Not read yet: root ``mcp.json`` is a follow-up."""
+        return {}
+
+    def load_hooks(self, plugin_dir: Path) -> HookConfig | None:  # noqa: ARG002
+        """Not portable core: hooks arrive via a client extension."""
```
enyst: Is this accurate? What does it mean to arrive via a client extension?

enyst (approved):

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
VascoSch92: changed in a better way...

PR #4485, opened 2026-08-13
feat(tracing): enhance observability by recording ACP turn usage and …

### LINKED ISSUE #4368: ACP conversations are invisible to Laminar: no LLM spans, no TOOL spans, no token/cost attributes
ACP conversations are effectively invisible to Laminar. They produce spans, but none of them carry the agent's work — no assistant turns, no tool calls, no token usage. The result is that an ACP conversation is structurally indistinguishable from a session where the agent never ran, which breaks observability, cost tracking, and trajectory export alike.

Found while investigating why `trajectory_export` (dataset-utils#40) cannot export ACP trajectories. Verified against SDK source on `main`, not inferred from data.

## What ACP emits today

A completed ACP conversation produces:

- `conversation.send_message` — `DEFAULT` spans, the user turns.
- `acp_agent.step` / `acp_agent.astep` — `DEFAULT` spans (`acp_agent.py:3574`, `acp_agent.py:3696`). Both return `None`, so the span `output` is empty.
- Nothing else.

## Why

**No LLM spans.** Every `span_type="LLM"` span in the corpus comes from lmnr's litellm auto-instrumentation — `grep -rn 'span_type="LLM"' openhands-sdk/` returns nothing, i.e. the SDK never emits one explicitly. `acp_agent.py` imports litellm only for a cost-map lookup (`acp_agent.py:835-837`, `litellm.model_cost`) and never calls it for a completion. The model call happens inside the external CLI subprocess (claude-agent-acp, codex-acp, gemini-cli), so auto-instrumentation has nothing to hook.

**No TOOL spans.** `span_type="TOOL"` is emitted in exactly two places: `agent.py:1267` (`_execute_action_event`) and `mcp/tool.py:67` (`MCPToolExecutor.call_tool`). ACP goes through neither — tool execution happens inside the CLI and arrives back as `ACPToolCallEvent` (`acp_agent.py:1449`, `acp_agent.py:3075`).

**No token/cost attributes.** `grep -rn 'gen_ai.usage' openhands-sdk/` returns nothing; those attributes also come from litellm auto-instrumentation. `_record_usage` (`acp_agent.py:1815`) does capture real cost and token counts from the ACP `UsageUpdate`, but writes them to `self.llm.metrics` — an in-process object — not onto a span. So the `traces` table's `input_tokens` / `output_tokens` / `total_cost` are **0** for every ACP conversation.

## Why this matters beyond exports

The `$0 / 0-token` trace aggregate is used as positive evidence that a session is a stub — dataset-utils#40 §2 relies on exactly that signal to justify dropping "the agent never ran" sessions from client deliveries. ACP conversations produce that signature while having run perfectly, so:

- They are silently dropped by stub-session filters.
- They contribute to the "agent never ran" rate in trajectory-quality analyses (dataset-utils#39 put that bucket at ~17%; how much of it is ACP is unmeasured).
- ACP cost and token usage are invisible in Laminar dashboards, even though the SDK has the numbers in hand.

To size the ACP share of a corpus:

```sql
SELECT count(DISTINCT trace_id) FROM spans
WHERE name IN ('acp_agent.astep', 'acp_agent.step')
```

## What's needed

The content is already in-process — it just isn't observed. Each item below is independently useful:

1. **An LLM-equivalent span per assistant turn**, carrying the assistant message and any tool calls as the span `output`, in the shape the native agent's litellm span produces. The data exists at the `MessageEvent` / `ActionEvent` emission sites (`acp_agent.py:3439-3455`, `acp_agent.py:3516`). This is the single highest-value item: it is what makes an ACP trajectory readable at all.
2. **TOOL spans for ACP tool calls**, parented to the enclosing `acp_agent.{a,}step`, so tool use is visible and pairs with its call. Sites: `acp_agent.py:1449`, `acp_agent.py:3075`.
3. **Token/cost attributes on a span** so `_record_usage`'s numbers reach the `traces` aggregates instead of dying in `llm.metrics`.
4. **Model identity.** ACP servers report the active model (and can switch it mid-session), but nothing surfaces it as a span attribute the way `gen_ai.request.model` does for litellm.

### Known gap, worth deciding explicitly

The **system prompt** is not recoverable. It lives inside the external CLI and is not exposed over ACP, so there is no equivalent of the native path's `messages[0]`. Any ACP trajectory export will lack the `system → user → assistant → tool` opening the native shape guarantees. Whether consumers can accept that — or whether a synthetic placeholder is preferable — should be decided rather than discovered downstream.

## Downstream

`dataset-utils`' fetch stage anchors on `AGENT_STEP_NAMES = ("agent.astep", "agent.step")` and would need `acp_agent.{a,}step` added — but that alone changes nothing, since there is no LLM child to select. Item 1 is the actual unblocker.

Related: #4358 (two other observability defects corrupting exported trajectories), dataset-utils#39, dataset-utils#40.

shreyyaps:
HUMAN:

in the listed 4 issues -
1. LLM-equivalent spans per assistant turn, carrying messages and tool calls
2. TOOL spans for ACP tool calls, parented to step spans
3. Token/cost attributes on spans
4. Model identity as span attributes

items 1 and 2 were already fixed by another PR #4376 

my PR->
 the `_record_usage` was returning `none`
Now it hands back the numbers it just recorded, as `ACPTurnUsage`, and the caller passes them into 
`finish_turn(usage=...)`
which writes them onto the span under standard gen_ai.* keys. 

---

AGENT:

## Why

ACP turn spans carried no usage. Cost and tokens for an ACP turn were recorded into
the in-process `Metrics` by `ACPAgent._record_usage`, but nothing carried them onto the
span that describes the turn — so in Laminar (or any OTLP backend) every ACP
conversation read as a free session, and dashboards built on trace data reported ACP
spend as zero.

Model and provider identity had the same gap: they were attached as span *metadata*
only, not as `gen_ai.*` attributes, so backends could not group, filter, or price ACP
turns alongside turns from a direct-LLM `Agent`.

## Summary

- Add `ACPTurnUsage` (a `NamedTuple`) to `acp_tracing.py` holding one turn's tokens and
  cost. `ACPAgent._record_usage` now returns it instead of `None`, so the figures it
  already derives reach the turn's span. `Metrics` remains the source of truth; this
  adds no new derivation.
- Stamp `gen_ai.*` attributes on the turn span: request/response model and
  `gen_ai.system` at turn start (so model identity survives a turn that times out and
  never reports usage), and usage plus `gen_ai.usage.cost` via
  `ACPTurnTrace.finish_turn(usage=...)`. Counts the server never reported are omitted
  rather than written as `0`, so "usage unavailable" (gemini-cli sends none today) stays
  distinguishable from "this turn was free".
- Add `ACPAgent._observability_server_key()`, which falls back to the provider detected
  from the live server's `agent_name` when `acp_server` is `None` — the case for an
  `ACPAgent` built directly rather than through `ACPAgentSettings`, as the standalone
  SDK examples do. Those spans previously carried no provider label at all.

Compatibility: additive. `finish_turn`'s `usage` is an optional keyword argument, and
`_record_usage` is private — a caller ignoring the new return value is unaffected. No
change to the agent loop or to any REST/WebSocket shape.

## Issue Number

Fixes #4368 

## How to Test

**1. Unit tests** — on top of `main` at 20f7f5f8f:

​```
$ uv run pytest tests/sdk/agent/test_acp_agent.py tests/sdk/agent/test_acp_tracing.py -q
475 passed, 10 warnings in 10.22s
​```

**2. Attributes verified leaving the process over a real OTLP exporter.** Unit tests assert
what is set on the span object; this checks the attributes actually survive serialization
and export, which is where a rejected attribute type or a write-after-`end()` would
silently drop them. `ACPTurnTrace` is driven the way `ACPAgent` drives it, with hand-fed
usage, against a local OTLP endpoint:

​```python
from openhands.sdk.agent.acp_tracing import ACPTurnTrace, ACPTurnUsage
from openhands.sdk.observability.laminar import maybe_init_laminar

maybe_init_laminar()
usage = ACPTurnUsage(
    input_tokens=1234, output_tokens=567, cache_read_tokens=8901,
    cache_write_tokens=234, reasoning_tokens=45, cost=0.0321,
)
trace = ACPTurnTrace(acp_server="claude-code", model_id="claude-sonnet-4-5")
trace.start_turn("List the Python files under openhands-sdk/openhands/sdk/agent/.")
trace.finish_turn("Here are the agent modules.", "thoughts", [], usage=usage)

from lmnr import Laminar
Laminar.flush()
print("total_tokens:", usage.total_tokens)
​```

​```
$ OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://127.0.0.1:4318/v1/traces \
  OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf \
  LMNR_FORCE_HTTP=true uv run python verify_acp_span.py
observability enabled: True
total_tokens: 10936
​```

Decoding the exported (gzipped) OTLP payload at the receiving end, all ten attributes are
present on the turn span:

​```
=== OTLP export received: 1064 bytes (gzip) -> 2347 bytes decompressed ===
  gen_ai.request.model                gen_ai.usage.cache_read_input_tokens
  gen_ai.response.model               gen_ai.usage.cache_creation_input_tokens
  gen_ai.system                       gen_ai.usage.reasoning_tokens
  gen_ai.usage.input_tokens           llm.usage.total_tokens
  gen_ai.usage.output_tokens          gen_ai.usage.cost
​```

`total_tokens` = 10936 = 1234 + 567 + 8901 + 234 — the 45 reasoning tokens correctly
excluded, since providers already bill thinking as output.

**Not done, stated plainly:** I have not run this against a live claude-agent-acp server
with real credentials, so the usage numbers above are hand-fed rather than server-reported.
The one remaining untested link is `_record_usage`'s return value reaching `finish_turn`
during a real turn; unit tests cover that call path, and everything downstream of it is
covered by the export check above. To reproduce the full path:

​```
export ANTHROPIC_API_KEY=<key>          # or ANTHROPIC_BASE_URL for a LiteLLM proxy
export LMNR_PROJECT_API_KEY=<laminar key>
uv run python examples/01_standalone_sdk/40_acp_agent_example.py
​```

Happy to run that and post the trace if a reviewer wants it before merge.

## Video/Screenshots

No screenshot: the verification above used a local OTLP endpoint rather than the Laminar
UI, and its output is quoted inline as text instead.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- `_PROVIDER_BY_SERVER` in `acp_tracing.py` maps the `acp_server` discriminator to a
  `gen_ai.system` value. That mapping arguably belongs on `ACPProviderInfo` in
  `settings/acp_providers.py` alongside the other per-provider metadata, so a newly
  registered provider cannot silently end up with no `gen_ai.system`. Happy to move it
  there in this PR if reviewers prefer.
- Unrelated pre-existing failure on `main`:
  `tests/sdk/conversation/test_local_conversation_mcp.py::test_reconciliation_targets_replaced_agent`
  fails at 20f7f5f8f without any of this branch's changes (`tools_map` contains
  `inspect_image_with_vision` alongside `replacement`). Not touched by this PR.

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/acp_agent.py
+++ b/openhands-sdk/openhands/sdk/agent/acp_agent.py
@@ -72,7 +72,7 @@ from openhands.sdk.agent.acp_file_credentials import (
     write_secret_file,
 )
 from openhands.sdk.agent.acp_models import ACPModelInfo
-from openhands.sdk.agent.acp_tracing import ACPTurnTrace
+from openhands.sdk.agent.acp_tracing import ACPTurnTrace, ACPTurnUsage
 from openhands.sdk.agent.base import AgentBase
 from openhands.sdk.context import AgentContext
 from openhands.sdk.conversation.state import ConversationExecutionStatus
@@ -1844,13 +1844,24 @@ class ACPAgent(AgentBase):
             atexit.unregister(callback)
             self._atexit_callback = None
 
+    def _observability_server_key(self) -> str | None:
+        """Which ACP CLI is running, for span labelling.
+
+        Prefers the provider detected from the live server's ``agent_name`` over
+        the configured :attr:`acp_server`, which is ``None`` whenever the agent
+        was built directly rather than through ``ACPAgentSettings`` — the case
+        the standalone SDK examples take.
+        """
+        provider = detect_acp_provider_by_agent_name(self._agent_name)
+        return provider.key if provider is not None else self.acp_server
+
     def _record_usage(
         self,
         response: PromptResponse | None,
         session_id: str,
         elapsed: float | None = None,
         usage_update: UsageUpdate | None = None,
-    ) -> None:
+    ) -> ACPTurnUsage:
         """Record cost, token usage, latency, and notify stats callback once.
 
         Args:
@@ -1858,16 +1869,23 @@ class ACPAgent(AgentBase):
             session_id: Session identifier used as the response_id for metrics.
             elapsed: Wall-clock seconds for this prompt round-trip (optional).
             usage_update: The synchronized ACP UsageUpdate for this turn, if any.
+
+        Returns:
+            What this turn consumed, so the caller can also put it on the
+            turn's observability span. Metrics remain the source of truth;
+            this only surfaces the numbers already derived here.
         """
         # -- Cost recording ---------------------------------------------------
         # claude-agent-acp, codex-acp: report cost via UsageUpdate notification
         # gemini-cli: does not send UsageUpdate (cost derived from tokens below)
         cost_recorded = False
+        turn_cost = 0.0
         if usage_update is not None and usage_update.cost is not None:
             last_cost = self._client._last_cost_by_session.get(session_id, 0.0)
             delta = usage_update.cost.amount - last_cost
             if delta > 0:
                 self.llm.metrics.add_cost(delta)
+                turn_cost = delta
                 cost_recorded = True
             self._client._last_cost_by_session[session_id] = usage_update.cost.amount
             self._client._last_cost = usage_update.cost.amount
@@ -1899,6 +1917,7 @@ class ACPAgent(AgentBase):
             )
             if cost > 0:
                 self.llm.metrics.add_cost(cost)
+                turn_cost = cost
 
         if not cost_recorded and not input_tokens and not output_tokens:
             # gemini-cli currently returns response.usage=None and
@@ -1917,6 +1936,15 @@ class ACPAgent(AgentBase):
                 self.llm.telemetry._stats_update_callback()
             except Exception:
                 logger.debug("Stats update callback failed", exc_info=True)
+
+        return ACPTurnUsage(
+            input_tokens=input_tokens,
+            output_tokens=output_tokens,
+            cache_read_tokens=cache_read,
+            cache_write_tokens=cache_write,
+            reasoning_tokens=reasoning,
+            cost=turn_cost,
+        )
 
     # -- Capability helpers ------------------------------------------------
 
@@ -3007,7 +3035,7 @@ class ACPAgent(AgentBase):
         self._client.trace.abandon()
         self._client.reset()
         self._client.trace = ACPTurnTrace(
-            acp_server=self.acp_server,
+            acp_server=self._observability_server_key(),
             model_id=self._current_model_id,
             mask=mask,
         )
@@ -3382,7 +3410,7 @@ class ACPAgent(AgentBase):
 
         session_id = self._session_id or ""
         usage_update = self._client.pop_turn_usage_update(session_id)
-        self._record_usage(
+        turn_usage = self._record_usage(
             response,
             session_id,
             elapsed=elapsed,
@@ -3408,7 +3436,10 @@ class ACPAgent(AgentBase):
             response_text = "(No response from ACP server)"
 
         self._client.trace.finish_turn(
-            response_text, thought_text, self._client.accumulated_tool_calls
+            response_text,
+            thought_text,
+            self._client.accumulated_tool_calls,
+            usage=turn_usage,
         )
 
         # ACP step() boundaries are full remote assistant turns, not
--- a/openhands-sdk/openhands/sdk/agent/acp_tracing.py
+++ b/openhands-sdk/openhands/sdk/agent/acp_tracing.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 import json
 import threading
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, NamedTuple
 
 from openhands.sdk.logger import get_logger
 from openhands.sdk.observability.laminar import should_enable_observability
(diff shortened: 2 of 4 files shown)


rajshah4:
I validated the `claude-agent-acp` path end-to-end against the public Laminar backend (real ACP subprocess via `npx -y @agentclientprotocol/claude-agent-acp`, real turn, spans round-tripped through Laminar's SQL API):

- `acp.completion` span carries `gen_ai.system=anthropic`, `gen_ai.request.model` / `gen_ai.response.model` = `sonnet`, and the full `gen_ai.usage.*` token set (input, output, cache_read, cache_creation, reasoning) + `llm.usage.total_tokens` + `gen_ai.usage.cost`.
- `gen_ai.usage.cost` on the span matches the in-process `accumulated_cost` exactly (e.g. 0.0226906).
- `gen_ai.response.id` is correctly **absent** — confirmed the session-id bearer token does not reach the tracing backend.

One thing worth noting for future readers (not a PR bug): Laminar's ingest recomputes `gen_ai.usage.input_tokens` to `input + cache_read + cache_creation` and derives `llm.usage.total_tokens` from that, so live spans show a larger input figure than the raw value the SDK emits. The PR's emitted values are correct (and the in-memory unit-test exporter sees the raw values, which is why the tests assert the un-normalized counts); this is backend-side normalization. A one-line comment near `_INPUT_TOKENS_ATTR` pointing this out would save the next reader the investigation.

Scope note: this live validation covered the `claude-code` provider only; the `codex` and `gemini-cli` paths are unit-tested in-repo but not live-smoked here.

This comment was generated by an AI agent (OpenHands) on behalf of @rajshah4.

PR #4510, opened 2026-08-16
feat(tools): add per-call llm_profile override to the task tool

### LINKED ISSUE #4654: [Feature]: Select an LLM profile for an individual task-tool worker
## Existing feature request

I searched the open issues and feature requests and did not find a duplicate.

## Problem or Use Case

A parent agent cannot choose a saved LLM profile for one delegated task. It must either change its own model or rely on a subagent definition with a fixed model.

That makes mixed-model delegation awkward. A parent may need to keep its current model while assigning one task to a cheaper, faster, or specialised worker.

An implementation proposal already exists in #4510. This issue records the user-facing contract that proposal needs to satisfy; it does not treat the current internal API shape as accepted.

## Desired Behavior

Allow a `task` call to select a saved LLM profile for that worker without changing the parent agent's model.

The behavior should be explicit when a subagent definition already fixes a model. A caller-supplied profile must not be silently ignored. Profile names shown to the parent must come from the same store used to resolve them, or discovery should be omitted when that cannot be guaranteed.

Resume should preserve or replace the selected profile within the existing `TaskManager` lifetime. The SDK should return a clear task error if the profile is unavailable. Process-restart persistence is outside this request because task IDs are currently in memory.

## Acceptance Criteria

- [ ] A task call can select a saved LLM profile for its worker without changing the parent model.
- [ ] An unknown or unavailable profile returns a clear task error before task state is registered; there is no silent fallback to the parent model.
- [ ] Supplying a profile when the chosen subagent definition already fixes a model produces explicit behavior rather than silently discarding the request.
- [ ] Any profile names advertised to the parent come from the same store used for resolution, or discovery is omitted when that cannot be guaranteed. Only names are advertised, and the model-facing disclosure is documented.
- [ ] A bare resume keeps the selected profile within the current `TaskManager` lifetime; an explicit profile replaces it for that resume and later resumes.
- [ ] Worker context and metrics remain isolated from the parent.
- [ ] The supported SDK and agent-server version boundary is documented. Mixed-version deployments must fail clearly or reject the unsupported field before execution.
- [ ] Creation and resume behavior are covered for file-backed, encrypted, and subscription-backed profiles, including a profile removed before resume.

## Alternatives Considered

- Change the parent's profile before and after each delegated task. This mutates the parent for a worker-only decision and makes concurrent delegation unsafe.
- Define one subagent per model. This works for fixed roles, but turns a per-task routing choice into static configuration and does not cover saved provider settings cleanly.

## Priority / Severity

Medium - Would improve experience

## Estimated Scope

Medium - New feature with moderate complexity

## Feature Area

- Tools / Tool system
- Configuration / Settings

## Additional Context

Implementation proposal: #4510

The proposal does not yet satisfy every criterion above. Its current branch silently discards a caller-supplied profile when the subagent definition pins a model, advertises names from the default store when a definition may resolve from a custom store, and lacks a direct test for a profile removed before resume. Those are review items, not accepted behavior.

georgeglarson:
HUMAN:

Enforced sub-agent delegation is the objective here.

---

AGENT:

## Why

A parent agent cannot currently choose a different saved LLM profile for one delegated task. It must either switch its own model or use a subagent definition with a fixed `model:` value.

That makes mixed-model delegation awkward. The parent should be able to keep its model while assigning an individual task to a cheaper, faster, or specialized worker.

## Summary

- Add an optional `llm_profile` field to the task tool and advertise saved profile names so a parent can select a worker model per call.
- Use one cipher-aware profile-loading primitive while preserving definition model, per-call profile, and parent-inheritance precedence across creation and resume.
- Keep the worker model, context, and metrics isolated from the parent; subscription-backed workers retain the factory-created condenser supported by subscription completion dispatch.

The selection order is:

1. A model profile specified by the subagent definition.
2. The task call's `llm_profile`.
3. The parent model.

A bare resume keeps the task's effective profile. Passing a different profile replaces it for that resume and later resumes. If the resumed subagent definition supplies its own model, the ignored per-call profile is cleared.

Unknown profiles return an error before task state is registered. There is no silent fallback to the parent model.

## Issue Number

Fixes #4654.

## How to Test

```bash
uv run pytest -q \
  tests/tools/task \
  tests/sdk/subagent \
  tests/sdk/tool/test_switch_llm.py \
  tests/sdk/conversation/test_switch_model.py
```

Result on the current branch:

```text
235 passed, 142 warnings in 2.38s
```

All touched-file pre-commit hooks pass, including Ruff, Pyright, import dependency checks, and tool registration checks.

I also ran the delegation path against real saved profiles. The parent used MiniMax M3 and delegated a task through an encrypted MiMo v2.5 Pro profile:

```text
task_status=completed
task_result=encrypted worker succeeded
parent_model_before=openai/MiniMax-M3
parent_model_after=openai/MiniMax-M3
worker_model=openai/mimo-v2.5-pro
parent_registry_ids=[]
parent_metric_keys=['task:task_00000001']
resume_profile=encrypted-worker
resume_worker_model=openai/mimo-v2.5-pro
```

The persisted profile did not contain the plaintext API key.

## Video/Screenshots

Not applicable. This change affects SDK task-tool behavior and is covered by the live output above.

## Design Doc

[Per-call `llm_profile` design](https://github.com/georgeglarson/software-agent-sdk/blob/feat/task-tool-llm-profile/.pr/design.md)

Companion documentation: [OpenHands/docs#761](https://github.com/OpenHands/docs/pull/761)

## Type

- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Breaking change
- [x] Docs / chore

## Notes

The task tool description includes saved profile names from the default profile store so the parent model can select one. It does not include profile configuration, model IDs, provider URLs, or credentials. Profile names remain visible when `enable_switch_llm_tool=False`; that setting removes the conversation-level switch tool but does not disable task-level worker selection.

A file-defined subagent may use a custom `profile_store_dir`. The task resolves profiles from that directory, but the advertised list still comes from the default store. Valid custom-store names may therefore be absent from the list.

Once a task stores a per-call profile, resume can replace it with another profile but cannot explicitly return it to parent inheritance. A definition-owned model clears the stored override. Restoring task IDs across a new process remains outside the current in-memory `TaskManager` contract.

Older SDK versions reject events containing the new field because action schemas forbid unknown fields. Mixed-version deployments should upgrade the agent server before emitting `llm_profile`.

Subscription-backed workers retain their factory-created condenser, matching current top-level agent creation and profile switching. Definition-level `model:` loading still has its pre-existing encrypted-profile limitation; the new per-call path is cipher-aware.


openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
```diff
@@ -1682,6 +1682,40 @@ def get_or_create_profile_llm(self, profile_name: str, usage_id: str) -> LLM:
             self._bind_conversation_context(llm)
             return llm
 
+    def load_profile_llm(
```
VascoSch92, line 1710: DO we need that? If yes, why? Because this clas is already hughe and I think that we have already a method which does that.

openhands-sdk/openhands/sdk/subagent/AGENTS.md
```diff
@@ -111,6 +111,8 @@ Supported YAML frontmatter keys (see `AgentDefinition.load` in `schema.py`):
 - `skills` (default: `[]`): a comma-separated string or a list of skill names
 - `model` (default: `inherit`): `inherit` reuses the parent LLM; another value is
   loaded as an LLM profile name from `profile_store_dir` or the default profile store
+  - a pinned `model:` takes precedence over the task tool's per-call
```
VascoSch92: Sorry, I don't get that. What is a pinned model?

openhands-tools/openhands/tools/task/definition.py
```diff
@@ -48,6 +52,19 @@ class TaskAction(Action):
...
+        default=None,
+        description=(
+            "Name of a saved LLM profile to run this subagent on, instead of "
+            "inheriting your current model. Use this for mixed-model workflows, "
+            "e.g. delegating implementation to a cheaper or faster profile while "
+            "you continue planning on your current model. This affects only the "
+            "delegated subagent — your own model is unchanged (use the switch_llm "
+            "tool to change your own model). If the chosen subagent_type's "
+            "definition pins its own model, that pin takes precedence and this "
+            "field is ignored."
```
VascoSch92: Not super happy with this description :-( I have to think about it.

openhands-tools/openhands/tools/task/definition.py
```diff
@@ -117,7 +134,7 @@ def to_llm_content(self) -> Sequence[TextContent | ImageContent]:
 Available agent types and the tools they have access to:
 {agent_types_info}
 
-When NOT to use the task tool:
+{llm_profiles_section}When NOT to use the task tool:
```
VascoSch92, line 132: are you printing the names of the profiles here?

openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
georgeglarson: Yes, but I changed it so this is now the small shared loading primitive rather than a second implementation.

`get_or_create_profile_llm()` is registry-oriented. On a cache miss, it assigns the caller's usage ID, registers the LLM in the parent conversation, and binds the parent's call context. That is correct for an auxiliary LLM owned by the parent, but not for a task worker. The worker needs a cipher-aware load without parent registration; its own `LocalConversation` owns its registry, context, and metrics.

`switch_profile()` and `get_or_create_profile_llm()` now both delegate their loading step to `load_profile_llm()`, so store selection and decryption are no longer duplicated. The tests also check that loading a worker neither changes the parent's active model nor adds the profile to the parent's registry.

openhands-sdk/openhands/sdk/subagent/AGENTS.md
georgeglarson: That wording was unclear. I meant an agent definition whose `model:` field names its own saved profile instead of using `inherit`.

That definition takes precedence over the task call's `llm_profile`. I removed “pinned” from the code and documentation and now describe this as “an agent definition with its own model profile.” The precedence tests also cover an invalid per-call profile being ignored when the definition supplies the model.

openhands-tools/openhands/tools/task/definition.py
georgeglarson: Fair. I shortened the schema description to:

> Saved LLM profile for this subagent. If omitted, the subagent inherits the parent model. An agent definition with its own model profile takes precedence.

The longer explanation of loading, errors, and resume behavior now stays in the task tool description and PR documentation instead of the field description.

openhands-tools/openhands/tools/task/definition.py
georgeglarson: Yes. The task tool description lists the saved profile names so the parent can select one. It uses the same `get_llm_profile_names()` and `format_llm_profiles()` helpers as `switch_llm`.

Only names are included. The profile model, base URL, API key, and stored JSON are not added to the description. I added a regression test with sentinel values for each of those fields and check that only the profile name appears.

The PR description also calls out that these names remain visible when `enable_switch_llm_tool=False`, and that the advertised list comes from the default store even when a file-defined subagent uses a custom store.

PR #4526, opened 2026-08-18
test(sdk): pin events_to_messages boundaries + fix responses_reasoning_item batch drop

### LINKED ISSUE #4525: events_to_messages lacks boundary tests (its own TODO); writing them exposed responses_reasoning_item being dropped in parallel-call batches
### Bug Description

`LLMConvertibleEvent.events_to_messages` (`openhands-sdk/openhands/sdk/event/base.py:108`) — the projection that turns the persisted event log into the `messages[]` sent to the LLM — has carried `# TODO: We should add extensive tests for this` since it was written. The existing suite covers the happy paths but none of the boundaries: parallel-call re-batching by `llm_response_id`, adjacent user-message coalescing, condensation-summary conversion.

Writing those characterization tests exposed a real defect. `_combine_action_events` reconstructs a parallel-tool-call response from the first event's fields but never passes `responses_reasoning_item` — so the OpenAI Responses-API reasoning item survives when an action converts as a singleton (`llm_convertible/action.py:143` passes it) and is silently dropped the moment the same response carries parallel calls. `response_dispatch.py:169-174` stamps reasoning state on the first batch event only (by design), so the first event holds the field and the combiner discards it: behavior depends on batch size.

### Expected Behavior

The combined parallel-call message preserves the first event's `responses_reasoning_item`, exactly as a singleton conversion of the same event does — batching must not change which reasoning state reaches the provider.

### Desired Behavior

The projection's documented boundaries are pinned by tests, retiring the TODO: parallel-batch re-combination (tool-call order, end-of-stream, interleaved response ids), user-message merge cascade/content-order/non-mutation, and the condensation-summary conversion.

### Actual Behavior

No boundary coverage exists, and `responses_reasoning_item` is `None` on the combined message. Demonstrated by a parity test (same first event, singleton vs two-call batch) — fails on current main with `AssertionError: assert None is not None`:

```
uv run pytest tests/sdk/event/test_events_to_messages.py -k responses_reasoning_item
```

(the test is added in the linked PR; the production code path it exercises is on main today.)

### Steps to Reproduce

1. Construct two `ActionEvent`s sharing one `llm_response_id` (parallel function calling), the first carrying `responses_reasoning_item=ReasoningItemModel(id="rs_1", ...)`.
2. Call `LLMConvertibleEvent.events_to_messages([first, second])`.
3. Inspect the single combined assistant message: `responses_reasoning_item is None`, while `first.to_llm_message().responses_reasoning_item.id == "rs_1"`.

### Acceptance Criteria

- [ ] Boundary tests for `events_to_messages` cover the cases listed under Desired Behavior.
- [ ] A parallel-call batch's combined assistant message carries the first event's `responses_reasoning_item` (parity with singleton conversion), covered by a regression test.

### Installation Method

uv

### SDK Version

main @ 98338ff37 (post-v1.42.1); verified present on latest main as of 2026-08-18.

### Version Confirmation

- [x] I have confirmed this bug exists on the latest version of the SDK.

georgeglarson:
HUMAN:

All tests pass human validation. 
I think the tests are correct in design and execution.

---

AGENT:

## Why

`LLMConvertibleEvent.events_to_messages` (`event/base.py:108`) is the projection that turns the persisted event log into the `messages[]` the LLM actually sees — it re-batches parallel tool calls by `llm_response_id` and coalesces adjacent plain user messages for strict role alternation. It has carried `# TODO: We should add extensive tests for this` since the function was written, and the existing suite (15 tests) covers the happy paths but none of the boundaries.

While writing characterization tests for those boundaries, one of them **failed and exposed a real bug**: `_combine_action_events` reconstructs a multi-tool-call response from the first event's fields but never passes `responses_reasoning_item` — so the OpenAI Responses-API reasoning item survives when an action converts as a singleton (`action.py:143` passes it) and silently disappears the moment the same response carries parallel tool calls. `response_dispatch.py:169-174` stamps reasoning state on the first batch event only (by design), so the first event holding the field and the combiner dropping it is a live inconsistency, not a theoretical one.

## Summary

- `test(sdk)`: characterization tests pinning documented-but-unpinned boundary behavior, retiring the TODO: user-message merge cascading across 3+ events, mixed text/image merge order, parallel batches preserving first-event reasoning state (`reasoning_content` / `thinking_blocks`), and the condensation-summary/user-message fusion.
- `fix(sdk)` (found by one of those tests): `_combine_action_events` now passes `responses_reasoning_item=events[0].responses_reasoning_item`, mirroring the adjacent `reasoning_content` / `thinking_blocks` lines; the parity test that exposed the drop is its regression test, asserting whole-item equality so future field loss fails it.
- `docs(sdk)`: the `events_to_messages()` docstring now states the read-only projection contract (events are immutable once appended to the log; merges build new content lists rather than mutating the events' own messages), and the retired TODO is removed.

## Issue Number

Fixes #4525

## How to Test

```
make build
uv run pytest tests/sdk/event/test_events_to_messages.py -v   # 20 tests, 5 new
CI=true uv run python -m pytest tests/sdk -n auto -q          # full sdk package
```

Full-suite result on this branch: **5974 passed, 0 failed** (7 skipped, 10 xfailed), run by the human author on Linux, Python 3.13. A full-repo run (`tests/`) passed 9391 with 27 failures, all host-environment artifacts unrelated to the change (stock shell-integration escape codes in pty-spawned terminal tests, environment-variable leakage into live-server tests, and one order-dependent OpenAPI test that passes in isolation on both base and this head); the `browser_use` e2e suite was excluded locally because `BrowserToolExecutor` teardown deadlocks off-CI. Pre-commit suite (ruff format/lint, pycodestyle, pyright, import rules, tool registration) clean.

Two notes for reviewers:

1. **The bug, demonstrated.** Check out the first commit's parent and run `test_parallel_batch_preserves_responses_reasoning_item` — it fails with `responses_reasoning_item=None` on the combined message despite the first event carrying the item. The fix flips it green. It was also validated live against GPT-5-nano on the real Responses API: a recombined parallel batch kept its reasoning item, and the regression test fails on `main` as expected — results recorded in `.pr/gpt-5-nano-integration-results.md` on this branch.
2. **A deliberate pin, since confirmed.** `CondensationSummaryEvent.to_llm_message()` produces a *plain* user message, so a condensation summary adjacent to a user `MessageEvent` in the view is fused with it into a single user turn (the summarizing condenser inserts the summary at `forgetting_start`, and the kept suffix may legally begin with a user message). `test_condensation_summary_merges_with_adjacent_user_message` pins the current behavior — API-valid and content-complete, with the summary's structural boundary erased in the projection. Raised here as a design question; confirmed by the maintainer as intended behavior (a normal consequence of the plain-user merge for strict role alternation), so the pin stays as documentation.

## Video/Screenshots

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The change is a one-line production fix (`_combine_action_events`) plus a docstring, and 5 test functions; the tests are the bulk of the diff by design — they close the long-standing `events_to_messages` TODO.

## Design Doc

N/A — one-line production change plus tests.


tests/sdk/event/test_events_to_messages.py
```diff
@@ -137,6 +140,107 @@ def test_consecutive_user_message_events_are_merged(self):
...
+        )
+
+        events = cast(list[LLMConvertibleEvent], [first, second, third])
+        messages = LLMConvertibleEvent.events_to_messages(events)
+
+        assert [message.role for message in messages] == ["user"]
+        assert [
+            content.text
+            for content in messages[0].content
+            if isinstance(content, TextContent)
```
enyst, line 165: Nit: I’m guessing `isinstance` is maybe not necessary, since there is nothing else in those messages?

tests/sdk/event/test_events_to_messages.py
```diff
@@ -137,6 +140,107 @@ def test_consecutive_user_message_events_are_merged(self):
...
+            source="user",
+            llm_message=Message(role="user", content=[TextContent(text="two")]),
+        )
+        events = cast(list[LLMConvertibleEvent], [first, second])
+
+        first_pass = LLMConvertibleEvent.events_to_messages(events)
+        assert len(first_pass[0].content) == 2
+
+        # The events' own messages are untouched, so a second projection over
+        # the same log produces the same output instead of merging twice.
```
enyst: I think maybe this constraint is better expressed differently… Maybe an AGENTS.md note on `events_to_messages()` or a docstring in the method, that it is readonly with respect to `events`

enyst: That said, everything is read only with respect to `events`: once they’re created and appended to `EventLog`, nothing mutates them. Pretty sure we have this written down somewhere… though it might in docs/ repo rather than here, tbh it should be here in the repo too

tests/sdk/event/test_events_to_messages.py
```diff
@@ -361,6 +534,92 @@ def test_multiple_separate_action_events(self):
...
+        assert tail.role == "assistant"
+        # Exact sequence, not membership: order within the batch is log order.
+        assert [tc.id for tc in tail.tool_calls] == [  # type: ignore
+            "call_1",
+            "call_2",
+            "call_3",
+        ]
+
+    def test_interleaved_response_ids_do_not_batch(self):
+        """Only ADJACENT same-response-id actions recombine into one message."""
```
enyst: I’m fairly confident this isn’t a realistic scenario, `ids` are very different hashes and won’t repeat like this 🤔

tests/sdk/event/test_events_to_messages.py
```diff
@@ -361,6 +534,92 @@ def test_multiple_separate_action_events(self):
         assert messages[1].content[0].text == "Second command"  # type: ignore
         assert messages[1].tool_calls[0].id == "call_2"  # type: ignore
 
+    def test_trailing_parallel_batch_at_end_of_stream(self):
+        """A parallel batch as the LAST events survives intact, in log order."""
```
enyst: Isn’t this already tested? They’re always “last” when they’re just received from the LLM

tests/sdk/event/test_events_to_messages.py
enyst: I disagree with this nit: `content` is a union, so these assertions provide runtime narrowing before `.text` access and make a wrong content type fail clearly. Rewriting them would be stylistic churn, not a meaningful simplification. I’m resolving this one.

_AI-generated reply posted by OpenHands on behalf of the user._

tests/sdk/event/test_events_to_messages.py
enyst: Agreed. The frozen model does not by itself make nested content immutable, but this test still characterizes existing deep-copy behavior rather than this PR’s batch-reasoning fix. I’d remove it to keep the patch focused; a repository-wide event immutability contract should be documented and tested separately.

_AI-generated reply posted by OpenHands on behalf of the user._

tests/sdk/event/test_events_to_messages.py
enyst: Agreed. The SDK assigns a unique response ID per LLM response, so interleaving and then reusing the same ID is not a realistic producer path. This test can be removed.

_AI-generated reply posted by OpenHands on behalf of the user._

tests/sdk/event/test_events_to_messages.py
enyst: Agreed. The existing three-call parallel test already covers an end-of-stream batch. Adding an exact `["call_SF", "call_Tokyo", "call_Paris"]` order assertion there retains the useful extra check without duplicating the scenario.

_AI-generated reply posted by OpenHands on behalf of the user._

enyst (commented):
🟡 **Acceptable** — The production fix is minimal and correct, but the test expansion is broader than the regression and live eval evidence is still pending.

- [`tests/sdk/event/test_events_to_messages.py`, line 193] **Scope:** The input-immutability test characterizes existing deep-copy behavior rather than the `responses_reasoning_item` batch-drop fix. Removing it would keep this PR focused.
- [`tests/sdk/event/test_events_to_messages.py`, line 473] **Regression assertion:** Compare the whole preserved `responses_reasoning_item` with the first event's item, not only its ID. That directly states the contract and catches future field loss.
- [`tests/sdk/event/test_events_to_messages.py`, line 537] **Duplicate coverage:** The trailing-batch test overlaps the existing three-call parallel test. Add an exact order assertion to the existing test instead.
- [`tests/sdk/event/test_events_to_messages.py`, line 584] **Unrealistic case:** Interleaved reuse of one LLM response ID is not produced by the SDK and does not exercise this fix. It can be dropped.

**[TESTING]**
- The focused event-conversion suite passes locally: **23 passed**.
- The new reasoning-item regression test fails against `origin/main` because the combined message has `responses_reasoning_item=None`, confirming it catches the bug.
- GitHub's integration workflow did not execute: `actions/checkout@v7` refused fork code in the `pull_request_target` job. Per repository policy, this review remains COMMENT pending a lightweight live eval and human maintainer confirmation.

**[RISK ASSESSMENT]**
The code change is only two field assignments and preserves existing batching order and tool calls, but it changes replayed OpenAI Responses history in core LLM message construction. That is agent/eval-sensitive behavior and warrants the requested live integration run before merge.

**VERDICT:**
✅ **Worth merging after live eval confirmation**: The core fix is sound; the test cleanup suggestions are nonblocking.

**KEY INSIGHT:**
The first action already owns the response-level reasoning metadata, so copying that metadata onto the single recombined assistant message restores the original LLM response without adding a new batching rule.

>
>
>

_AI-generated review posted by OpenHands on behalf of the user._

georgeglarson:
Thanks. Current head `d820af43f` aligns with every item from both rounds:

- `isinstance` filter: restored. 
- Trailing-batch test: deleted, and its one useful extra check (exact tool-call order) is folded into `test_parallel_function_calling_same_response_id`, replacing the membership checks.
- Interleaved-response-id test: deleted.
- No-mutation test: deleted. 
- Regression test strengthened: whole-item equality on `responses_reasoning_item`, not just the id — future field loss now fails it.
- The TODO is gone.

enyst (approved):
Thank you!

PR #4534, opened 2026-08-18
fix(sdk): strip leading 思考 block from LLM-generated title

### LINKED ISSUE #4530: [Bug]: LLM-generated conversation titles leak raw <think> reasoning blocks

I have searched the existing issues (open and closed) and did not find one.

### Bug Description

When the title LLM is a reasoning model whose provider does not split chain-of-thought into the normalized `reasoning_content` field (e.g. Qwen3 behind Nebius, which returns reasoning inline in `content` as `<think>...</think>`), the raw reasoning block leaks into the autogenerated conversation title.

`generate_title_with_llm()` in `openhands/sdk/conversation/title_utils.py` takes `response.message.content[0].text` verbatim, so the title becomes e.g. `<think>reasoning…</think>Actual Title`. Agent Canvas surfaces this directly as the conversation card title.

Reported downstream against Agent Canvas 1.14.0 in OpenHands/software-agent-sdk#4541 ("Weird `<think>` behavior with Qwen3-32B"): the autogenerated title shows `<think>` content. The same report also includes frontend rendering symptoms (plaintext thoughts + a duplicate thought block) — those are frontend-side; this issue tracks the SDK-side title leak, which is visible in any consumer of `Conversation.generate_title()` / the agent-server auto-title.

### Expected Behavior

The autogenerated title never contains reasoning markup: a leading `<think>...</think>` block in the title LLM response is peeled before the text is used, mirroring the conservative peel the frontend applies to message content. A response that is only reasoning produces no title from the LLM path, so the existing truncation fallback covers it.

### Actual Behavior

Reproducible SDK run (tested on `main` @ `ae18d7cd0` with `python` via `uv run pytest`): `Conversation.generate_title()` with a title LLM whose completion returns `"<think>…</think>🐛 Fix the login bug"` stores the whole string as the title, tags included. See the mock-based reproduction in the linked report's screenshot, and the fix's regression tests (`tests/sdk/conversation/test_generate_title.py`) which fail on `main` semantics.

### Steps to Reproduce

1. Configure a conversation whose title LLM resolves to a Qwen3-class model behind a provider that returns reasoning inline in `content` (Nebius; also reproducible with any completion returning `<think>…</think>` text).
2. Start a conversation so the auto-title runs.
3. Observe the stored/returned title contains the raw `<think>…</think>` block.

Deterministic without credentials: mock the title LLM completion to return `<think>reasoning</think>Some Title` and call `Conversation.generate_title()`.

### Acceptance Criteria

- [ ] `generate_title_with_llm()` strips a leading `<think>...</think>` block from the response text before using it as the title.
- [ ] A response containing only an (unterminated or closed) think block yields no LLM title, falling back to the existing user-message truncation.
- [ ] Mid-text literal `<think>` occurrences in a title are preserved (conservative peel, first block only).
- [ ] Regression tests cover the closed-block, unclosed-block, think-only, and mid-text cases.

### Installation Method

From source (git clone)

_No response_

### SDK Version

main @ ae18d7cd0

### Version Confirmation

- [x] I am confirmed that this bug exists on the latest version of the SDK.

### Python Version

3.12

santhiprakash:
HUMAN:

Santhi Prakash — small correctness fix for the SDK title-leak bug class cross-referenced from OpenHands/software-agent-sdk#4541; see AGENT section for evidence.

---

AGENT:

## Why

Some providers (notably Qwen3 behind Nebius) return reasoning inline
in `content` as `思考...思考` instead of the normalized
`reasoning_content` field. The transparent leak surfaced in Agent
Canvas 1.14.0 ([software-agent-sdk#4541](https://github.com/OpenHands/software-agent-sdk/issues/4541))
as the autogenerated conversation title containing the raw `思考`
reasoning markup — the SDK ships the reasoning into the title, and Agent
Canvas just renders it.

This is the SDK-side root-cause fix: peeling the leading `思考` block in
the SDK means *every* consumer of `Conversation.generate_title()` and the
agent-server auto-title path benefits, not just Agent Canvas.

## Summary

- `generate_title_with_llm()` now peels a leading `思考` block from the
  LLM title response before using it as the title.
- A response that is only a `思考` block (closed or unclosed) yields
  no LLM title and falls back to the existing user-message truncation.
- Mid-text literal `思考` occurrences in a legitimate title are
  preserved.

## Issue Number

Fixes #4530

## How to Test

```bash
uv run pytest tests/sdk/conversation/test_generate_title.py -q
```

Expected: 12 passed (4 new regression tests + 8 existing).

The 4 new tests cover:

1. `test_generate_title_strips_leading_think_block` — closed
   `思考…思考` block is peeled, the real title after the block is returned.
2. `test_generate_title_unclosed_think_block_falls_back` — unterminated
   leading `思考…` returns no LLM title; truncation fallback supplies the
   title.
3. `test_generate_title_think_only_response_falls_back` — response that
   is only a `思考` block yields no LLM title; truncation fallback
   supplies the title.
4. `test_generate_title_preserves_mid_text_think_literal` — a mid-text
   literal `思考` occurrence in a legitimate title is preserved.

I verified the regression coverage by reverting
`generate_title_with_llm()` to its pre-fix behavior and re-running the
new tests: 3/4 fail on the unfixed code (the literal-preservation test
correctly passes either way), proving the tests do catch the bug class.

```text
$ uv run pytest tests/sdk/conversation/test_generate_title.py -q
collected 12 items
tests/sdk/conversation/test_generate_title.py ............               [100%]
============================== 12 passed in 0.38s ==============================
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The fix is intentionally scoped to the SDK-side title generation path;
  it does not touch the Agent Canvas frontend rendering issues (plaintext
  thoughts / duplicate thought block) covered separately in
  OpenHands/software-agent-sdk#4541.
- The peel is a leading-block conservative peel — only the *first* leading
  block is stripped, so a legitimate title that happens to contain the
  characters `思考` mid-text is preserved.
- The unterminated-leading case consumes the entire response rather than
  returning the raw reasoning string, so the LLM path surfaces no title
  for the truncation fallback to take over.


santhiprakash:
Rebased onto current `main` (`1de2e6d`, which restructured `generate_title_with_llm` with the new `on_error` callback in #4535) — the conflict was only in the import block; the think-block peel now sits inside the updated signature alongside the `on_error` plumbing.

Verified locally on the rebased head:

- `uv run python -m pytest tests/sdk/conversation/test_generate_title.py -q` → **13 passed** (incl. the new `on_error` test from #4535 and the three 思考-block tests from this PR)
- `uv run python -m pytest tests/agent_server/test_conversation_service.py -q` → **109 passed**
- `ruff check` / `ruff format --check` clean

Heads-up: every workflow run on this PR (including the original Aug 18 head) has been sitting in *action_required* — "awaiting workflow approval" — so CI has never actually executed. The bot's "CI failing" notice reflects that gate, not a test failure. Could a maintainer approve the workflow runs so the suite can run?

neubig:
I think this is probably a fix on the litellm side. The "think" blocks should be parsed out into the `reasoning_content` response. I don't think this is something that we should be making heuristic changes to here.

PR #4548, opened 2026-08-19
fix(sdk): bound AsyncExecutor.close() so it cannot hang forever

### LINKED ISSUE #4546: [Bug]: AsyncExecutor.close() can block forever, wedging conversation shutdown
### Bug Description

`AsyncExecutor.close()` can block forever. It is called during `LocalConversation.close()` to release tool executors, so a single stuck portal task wedges conversation shutdown — and with it every subsequent operation that needs the conversation lock.

Two independent causes, both in `AsyncExecutor.close()` ([`async_executor.py:104-114`](https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/utils/async_executor.py#L104-L114)):

```python
if portal_cm is not None:
    try:
        portal_cm.__exit__(None, None, None)
```

1. **Remaining tasks are never cancelled.** Passing no exception makes anyio take its graceful path, `portal.call(portal.stop, cancel_remaining=False)`, which waits for in-flight tasks to finish on their own. A task that never completes blocks shutdown permanently — even one that would die instantly if cancelled.
2. **The thread join is unbounded.** `start_blocking_portal()` ends in a bare `thread.join()` with no timeout ([anyio `from_thread.py`](https://github.com/agronholm/anyio/blob/master/src/anyio/from_thread.py)). So even *with* cancellation, a task that cannot be cancelled — anything awaiting inside a worker thread, where anyio cannot deliver cancellation until the thread returns — hangs the join forever.

Cancelling alone does not fix it, and bounding the join alone leaves cause 1 waiting the full timeout on every close. Both need addressing.

`AsyncExecutor` backs the browser toolset (`BrowserToolExecutor`), the MCP client (`MCPClient`), and `ACPAgent`, so this is not browser-specific.

### Expected Behavior

`close()` returns promptly. Tasks still on the portal are cancelled; if the portal thread cannot be stopped it is abandoned — it is a daemon thread — and a warning is logged.

### Actual Behavior

`close()` blocks indefinitely. Nothing times out, so the process never recovers.

Save the snippet under *Steps to Reproduce* as `repro_close_hang.py` in a checkout of this repo and run it:

```bash
uv run python repro_close_hang.py
```

Observed on `main` (`bb26768`) and on 1.42.1:

```
close() STILL BLOCKED after 10s
```

Expected output is `close() returned`. The 10s is only the harness's patience — the call never returns at all.

The same hang is reachable from the test suite. With a bounded `close()` this passes in ~1s; on `main` `test_close_returns_with_task_still_running` blocks until its assertion deadline:

```bash
uv run pytest tests/sdk/utils/test_async_executor.py -q
```

It also escapes pytest's own timeout when it happens in teardown: on `main`, `uv run pytest tests/tools/browser_use/test_browser_executor_e2e.py::TestBrowserExecutorE2E::test_navigate_action -q --timeout=180` never returned and had to be killed externally at 400s.

### Steps to Reproduce

```python
import threading, time, anyio
from openhands.sdk.utils.async_executor import AsyncExecutor

ex = AsyncExecutor()
ex.portal.start_task_soon(anyio.sleep_forever)   # a task still in flight
time.sleep(0.5)

done = threading.Event()
threading.Thread(target=lambda: (ex.close(), done.set()), daemon=True).start()
print("close() returned" if done.wait(10) else "close() STILL BLOCKED after 10s")
```

Prints `close() STILL BLOCKED after 10s` on `main` (`bb26768`) and on 1.42.1. Note `anyio.sleep_forever` is trivially cancellable — it dies instantly under `cancel_remaining=True`.

Isolating the two causes:

| task on the portal | `cancel_remaining` | `close()` |
| --- | --- | --- |
| `anyio.sleep_forever()` | `False` (current behaviour) | **hangs** |
| `anyio.sleep_forever()` | `True` | returns |
| `await run_sync(lambda: time.sleep(3600))` | `True` | **hangs** |

### Production symptom

Hit on an agent-server running the browser tool. `LocalConversation.close()` closes tool executors in a loop with no timeout ([`local_conversation.py:2679-2688`](https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py)), reaching `BrowserToolExecutor.close()`, whose `finally` calls `self._async_executor.close()` unbounded ([`browser_use/impl.py:710-732`](https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-tools/openhands/tools/browser_use/impl.py)). The 30s guard on the preceding `run_async(self.cleanup, ...)` does not cover it.

The server stayed **8h21m** in this state. `py-spy dump`:

```
Thread 550621 (idle): "asyncio_4"
    _wait_for_tstate_lock (threading.py:1167)
    join (threading.py:1147)
    start_blocking_portal (anyio/from_thread.py:560)
    __exit__ (contextlib.py:144)
    close (openhands/sdk/utils/async_executor.py:112)
    close (openhands/tools/browser_use/impl.py:727)
    close (openhands/sdk/conversation/impl/local_conversation.py:2685)

Thread 555902 (idle): "asyncio-portal-705da41d8410"
    _cancel_all_tasks (asyncio/runners.py:205)
    run_blocking_portal (anyio/from_thread.py:536)
```

Because the close holds the conversation lock, routes needing a loaded conversation (`GET /api/conversations/{id}/events*`) blocked forever, while metadata routes (`/api/conversations/{id}`, `/api/conversations/search`, `/health`) kept answering in milliseconds. The service therefore looks healthy while no conversation can be opened. Chromium had already exited; no file lock was held. Only a process restart recovered it.

### Acceptance Criteria

- [ ] `close()` returns even with a task still running on the portal
- [ ] `close()` returns even when that task cannot be cancelled, and logs a warning
- [ ] `close()` stays idempotent and safe when the portal was never started

### Installation Method

`uv` (workspace checkout), and `uvx --from openhands-agent-server==1.42.1`

### SDK Version

1.42.1, and `main` at `bb26768`

### Python Version

3.12.3 (production), 3.13.13 (repo venv)

### Operating System

Linux

AaronAbuUsama:
HUMAN:

Okay, so this issue hit me while I was running this on a self-hosted server. Everything looked healthy. I got two 200s back, but every attempt to open the conversation just timed out and it was stuck for like eight hours, until I finally decided to ask Claude to dig around and figure it out.

---

AGENT:

## Why

`AsyncExecutor.close()` could block forever, for two independent reasons:

1. **Remaining tasks were never cancelled.** `close()` called `portal_cm.__exit__(None, None, None)`. Passing no exception makes anyio take its graceful path, `portal.stop(cancel_remaining=False)`, which waits for in-flight tasks to finish on their own — so a task that never completes blocks shutdown permanently, even one that would die instantly if cancelled.
2. **The thread join was unbounded.** `start_blocking_portal()` ends in a bare `thread.join()`. So even *with* cancellation, a task that cannot be cancelled — anything awaiting inside a worker thread, where anyio cannot deliver cancellation until the thread returns — hangs the join forever.

This is reachable in production. `LocalConversation.close()` releases tool executors in a loop with no timeout, so one stuck portal task wedges conversation shutdown while holding the conversation lock. On an agent-server running the browser tool, that left every `GET /api/conversations/{id}/events*` blocked for **8h21m** while `/health` and the metadata routes kept answering normally — the service looked healthy but no conversation could be opened, and only a process restart recovered it. `AsyncExecutor` also backs `MCPClient` and `ACPAgent`, so the exposure is not browser-specific.

## Summary

- `AsyncExecutor.close()` now stops the portal with `cancel_remaining=True` instead of waiting for in-flight tasks.
- The wait for the portal thread is bounded by a new `timeout` argument (default `DEFAULT_CLOSE_TIMEOUT = 10.0`; `None` keeps the old blocking behaviour). On expiry it logs a warning and abandons the thread, which is safe because anyio creates it as a daemon.
- Added `tests/sdk/utils/test_async_executor.py` covering both hang modes plus idempotency and the never-started-portal path.

Behaviour is additive: the new argument is optional and the normal close path is unchanged.

## Safety review (addressed)

This PR is a **bounded, best-effort safety net — not guaranteed cleanup.** After the timeout, the portal/helper/worker threads and their resources may still be alive. This is a deliberate trade-off (blocking the caller forever is worse); the semantics are documented as such in the `close()` docstring and the abandonment warning.

- ✅ **Documented precisely.** The `close()` docstring states it is best-effort, that threads/resources may survive the timeout, and that the path is non-raising + idempotent.
- ✅ **Observable abandonment.** The timeout warning names the owner (`type(self).__qualname__`), the timeout, and that cancellation was already attempted, so it can be correlated with `py-spy`/the wedged resource. The helper thread is named `<owner>-close` for traceability.
- ✅ **Preserved failure info.** Teardown failures use `exc_info=True` (full traceback) rather than `str(exc)`, while staying non-raising for destructor-safety.
- ✅ **Production stays bounded.** All production callers (`BrowserToolExecutor`, `ACPAgent`, `MCPClient`) call `close()` with no args → the 10s default. None pass `timeout=None`; `None` is retained only for backward compatibility and is documented as "do not use on a production path."
- ✅ **Shorter shutdown default.** Lowered from the inherited 30s (browser cleanup, pre-lifecycle-lock) to 10s — a successful portal stop + join normally completes in milliseconds, so 10s is already a generous healthy-shutdown margin while limiting how long teardown may hold a conversation lock. See #4598.
- ⚠️ **Remaining limitation tracked (#4598).** Arbitrary synchronous code cannot be forcibly interrupted from a Python thread, so this contains the caller-side wedge but does not eliminate possible zombie threads/subprocesses. Per-conversation lifecycle locks (#4570, merged) now ensure a stuck `close()` can only block its *own* conversation, not all of them — this PR is now a safety net rather than the load-bearing fix.

## Issue Number

Closes #4546

## How to Test

Reproduce the hang on `main` (25 lines, no browser or network needed):

```python
import threading, time, anyio
from openhands.sdk.utils.async_executor import AsyncExecutor

ex = AsyncExecutor()
ex.portal.start_task_soon(anyio.sleep_forever)
time.sleep(0.5)

done = threading.Event()
threading.Thread(target=lambda: (ex.close(), done.set()), daemon=True).start()
print("close() returned" if done.wait(10) else "close() STILL BLOCKED after 10s")
```

On `main` this prints `close() STILL BLOCKED after 10s`; on this branch it prints `close() returned`.

Then run the new tests:

```bash
uv run pytest tests/sdk/utils/test_async_executor.py -q
```

They are real regression tests — reverting `async_executor.py` while keeping the test file makes `test_close_returns_with_task_still_running` fail.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/.pr/repro-async-executor-close-hang.py
+++ b/.pr/repro-async-executor-close-hang.py
@@ -1,0 +1,295 @@
+#!/usr/bin/env python3
+"""
+repro-async-executor-close-hang.py — Live evidence for PR #4548 / issue #4546.
+
+Reproduces the AsyncExecutor.close() hang that wedges the agent-server's
+conversation lifecycle lock, and verifies the fix bounds it.
+
+Two phases:
+
+  Phase 1 (deterministic, the smoking gun):
+    Directly exercise AsyncExecutor.close() with a task that never completes.
+    Without the fix: close() blocks forever (we abort after PROBE_DEADLINE).
+    With the fix:    close() returns within DEFAULT_CLOSE_TIMEOUT (30s), and
+                     near-instantly for a cancellable task (anyio.sleep_forever).
+
+  Phase 2 (HTTP, live backend):
+    Hammer the running agent-server with concurrent conversation create +
+    search + delete traffic. With the fix in place, all requests succeed
+    quickly and the backend stays responsive. (Phase 1 is the deterministic
+    proof that without the fix, close() hangs; Phase 2 confirms the live
+    backend does not stall once the fix is applied.)
+
+USAGE
+  python3 repro-async-executor-close-hang.py [--http URL] [--no-unit]
+
+  --http URL   Also run the HTTP concurrent-load phase against URL
+               (default: http://localhost:8000)
+  --no-unit    Skip the deterministic unit phase
+
+EXIT CODE
+  0  all phases passed (fix is working)
+  1  a phase failed/stalled (bug reproduced)
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+import threading
+import time
+import urllib.error
+import urllib.request
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+
+# Make the SDK importable from the local checkout.
+_SDK = "/home/gneubig/work/software-agent-sdk/openhands-sdk"
+if _SDK not in sys.path:
+    sys.path.insert(0, _SDK)
+
+PROBE_DEADLINE = 15.0  # seconds we wait before declaring close() a hang
+
+
+def _banner(title: str) -> None:
+    print(f"\n{'═' * 70}\n  {title}\n{'═' * 70}")
+
+
+def _status(ok: bool, msg: str) -> int:
+    tag = "✓ PASS" if ok else "✗ FAIL"
+    print(f"  {tag} — {msg}")
+    return 0 if ok else 1
+
+
+# ── Phase 1: deterministic AsyncExecutor.close() repro ───────────────────
+
+
+def _close_in_background(executor, **kwargs) -> threading.Event:
+    """Call close() off-thread so a hang doesn't freeze the script."""
+    done = threading.Event()
+
+    def run():
+        try:
+            executor.close(**kwargs)
+        except Exception as e:
+            print(f"  (close() raised: {e})")
+        finally:
+            done.set()
+
+    threading.Thread(target=run, daemon=True).start()
+    return done
+
+
+def phase1_unit() -> int:
+    """Deterministic reproduction of the AsyncExecutor.close() hang."""
+    import anyio  # noqa: F401  (prove it's importable)
+
+    from openhands.sdk.utils.async_executor import AsyncExecutor
+
+    rc = 0
+    _banner("Phase 1 — AsyncExecutor.close() with a never-finishing task")
+
+    # Case A: a cancellable task (anyio.sleep_forever). With the fix,
+    # cancellation is delivered and close() returns almost instantly.
+    # Without the fix, close() waits forever for the task to finish on its own.
+    print("\n  Case A: cancellable task (anyio.sleep_forever)")
+    executor = AsyncExecutor()
+    executor.portal.start_task_soon(anyio.sleep_forever)
+    time.sleep(0.2)  # let the task start
+
+    t0 = time.monotonic()
+    done = _close_in_background(executor)
+    finished = done.wait(timeout=PROBE_DEADLINE)
+    elapsed = time.monotonic() - t0
+
+    if finished:
+        rc |= _status(True, f"close() returned in {elapsed:.2f}s (fix working)")
+    else:
+        rc |= _status(False, f"close() hung > {PROBE_DEADLINE:.0f}s (BUG reproduced)")
+        # best-effort: leave the daemon thread to die with the process
+
+    # Case B: a task blocked in a worker thread (uncancellable). With the fix,
+    # close() waits up to DEFAULT_CLOSE_TIMEOUT then abandons the daemon thread.
+    # Without the fix, close() hangs forever.
+    print("\n  Case B: uncancellable task (blocked in worker thread)")
+    from anyio.to_thread import run_sync
+
+    async def blocked_in_worker_thread():
+        await run_sync(lambda: time.sleep(60))
+
+    executor2 = AsyncExecutor()
+    executor2.portal.start_task_soon(blocked_in_worker_thread)
+    time.sleep(0.2)
+
+    t0 = time.monotonic()
+    done2 = _close_in_background(executor2, timeout=2.0)
+    finished2 = done2.wait(timeout=PROBE_DEADLINE)
+    elapsed2 = time.monotonic() - t0
+
+    if finished2:
+        rc |= _status(
+            True,
+            f"close() returned in {elapsed2:.2f}s with timeout=2.0 (fix working)",
+        )
+    else:
+        rc |= _status(
+            False, f"close() hung > {PROBE_DEADLINE:.0f}s even with timeout (BUG)"
+        )
+
+    # Case C: idempotent close (no hang on a clean executor).
+    print("\n  Case C: idempotent close on a clean executor")
+    executor3 = AsyncExecutor()
+    _ = executor3.portal
+    t0 = time.monotonic()
+    executor3.close()
+    executor3.close()
+    elapsed3 = time.monotonic() - t0
+    rc |= _status(True, f"double close() returned in {elapsed3:.2f}s")
+
+    return rc
+
+
+# ── Phase 2: HTTP concurrent load against the live backend ────────────────
+
+
+def _http(
+    method: str, url: str, key: str, body: dict | None = None
+) -> tuple[int, float]:
+    """Fire one HTTP request; return (status_code, elapsed_seconds)."""
+    data = json.dumps(body).encode() if body is not None else None
+    req = urllib.request.Request(
+        url,
+        data=data,
+        method=method,
+        headers={
+            "X-Session-API-Key": key,
+            "Content-Type": "application/json",
+        },
+    )
+    t0 = time.monotonic()
+    try:
+        with urllib.request.urlopen(req, timeout=20) as resp:
+            resp.read()
+            return resp.status, time.monotonic() - t0
+    except urllib.error.HTTPError as e:
+        return e.code, time.monotonic() - t0
+    except Exception:
+        return 0, time.monotonic() - t0
+
+
+def phase2_http(base_url: str, key: str, concurrency: int = 10, rounds: int = 3) -> int:
+    """Hammer the backend with concurrent READ traffic.
+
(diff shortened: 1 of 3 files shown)


neubig:
@/tmp/pr-evidence-revised.md

neubig (changes requested):
Requesting changes: this is a **mitigation, not a fundamental fix**. Bounding `AsyncExecutor.close()` to 30s replaces "hang forever" with "block all conversations for 30s". That's a shorter wedge, but still a wedge — and serial 30s stalls accumulate under load.

The root cause is that `close()` runs under the global `_lifecycle_lock` (in `delete_conversation`), so even a bounded hang blocks every other conversation. See the ready-for-dev issue #4569 for the fundamental solution (per-conversation locks).

This PR can likely be **closed as superseded** once #4569 is implemented — if `close()` can only block its own conversation, a timeout becomes a safety net rather than a critical fix.

neubig (changes requested):
Requesting changes: this is a **mitigation, not a fundamental fix**. Bounding `AsyncExecutor.close()` to 30s replaces "hang forever" with "block all conversations for 30s". That's a shorter wedge, but still a wedge — and serial 30s stalls accumulate under load.

The root cause is that `close()` runs under the global `_lifecycle_lock` (in `delete_conversation`), so even a bounded hang blocks every other conversation. See the ready-for-dev issue #4569 for the fundamental solution (per-conversation locks).

This PR can likely be **closed as superseded** once #4569 is implemented — if `close()` can only block its own conversation, a timeout becomes a safety net rather than a critical fix.

neubig (changes requested):
Requesting changes: this is a **mitigation, not a fundamental fix**. Bounding `AsyncExecutor.close()` to 30s replaces "hang forever" with "block all conversations for 30s". That's a shorter wedge, but still a wedge — and serial 30s stalls accumulate under load.

The root cause is that `close()` runs under the global `_lifecycle_lock` (in `delete_conversation`), so even a bounded hang blocks every other conversation. See the ready-for-dev issue #4569 for the fundamental solution (per-conversation locks).

This PR can likely be **closed as superseded** once #4569 is implemented — if `close()` can only block its own conversation, a timeout becomes a safety net rather than a critical fix.

neubig:
Requesting changes: this is a **mitigation, not a fundamental fix**. Bounding `AsyncExecutor.close()` to 30s replaces "hang forever" with "block all conversations for 30s". That's a shorter wedge, but still a wedge — and serial 30s stalls accumulate under load.

The root cause is that `close()` runs under the global `_lifecycle_lock` (in `delete_conversation`), so even a bounded hang blocks every other conversation. See the ready-for-dev issue #4569 for the fundamental solution (per-conversation locks).

This PR can likely be **closed as superseded** once #4569 is implemented — if `close()` can only block its own conversation, a timeout becomes a safety net rather than a critical fix.

neubig (changes requested):
Thanks for the thorough reproduction and for addressing the unbounded shutdown path. The direction is useful as a last-resort containment mechanism, especially now that per-conversation lifecycle locks limit the blast radius. Before merging, please address these safety requirements:

1. **Document the semantics precisely.** This is bounded, best-effort shutdown—not guaranteed cleanup. After the timeout, the portal/helper/worker threads and their resources may still be alive. Please make that explicit in the API docstring, warning, and PR description.

2. **Make abandonment observable.** The timeout warning should clearly state that shutdown was abandoned and resources may remain active. Include enough context to identify the executor/owner, the timeout, and that cancellation was attempted.

3. **Preserve useful failure information.** Avoid reducing shutdown failures to only `str(exception)`. Log the exception with traceback/context where appropriate, while keeping teardown non-raising. Broad exception handling is acceptable for destructor-safe cleanup only if the failure remains diagnosable.

4. **Keep production shutdown bounded.** Audit callers and confirm that no production path passes `timeout=None`; do not reintroduce the original unbounded behavior through the compatibility option.

5. **Reconsider the default timeout.** Thirty seconds is inherited from the browser cleanup timeout, but it may be too long for lifecycle teardown. Please justify the value with normal shutdown measurements or choose a shorter shutdown-specific default.

6. **Add regression coverage.** Tests should cover cancellation of cancellable tasks, timeout/abandonment of uncancellable synchronous work, idempotent close, never-started portals, observable timeout diagnostics, and shutdown exceptions. The timeout-path test should explicitly document that background threads may remain.

7. **Track the remaining limitation.** Please link or note #4598: arbitrary synchronous code cannot be forcibly interrupted from Python threads, so this change contains the caller-side wedge but does not eliminate possible zombie threads.

This review was created by an AI agent (OpenHands) on behalf of the user.

neubig (approved):
Approving: the rebased commits address the safety review items. Bounded best-effort semantics are documented; abandonment is observable (owner+timeout+cancellation-attempted) and the helper thread is named for traceability; teardown failures log with exc_info; all production callers use the 10s default (none pass timeout=None); #4598 is linked for the un-interruptible-thread limitation. #4570 (per-conversation locks) is merged, so a stuck close() now only blocks its own conversation — this is a safety net, not the load-bearing fix. CI green.

PR #4571, opened 2026-08-21
fix(workspace): honor explicit provider host when injecting git clone tokens

### LINKED ISSUE #4543: [Bug]: Cannot use a token to clone from a self-hosted Gitlab

### Bug Description

If one tries to clone from a non-`gitlab.com` Gitlab instance, the Gitlab token is ignored.

The issue is here:
https://github.com/OpenHands/software-agent-sdk/blob/bb26768cc501e63d6e9733990a184e0612d55c20/openhands-sdk/openhands/sdk/workspace/repo.py#L292-L317

### Expected Behavior

The token should not be ignored.

### Actual Behavior

The token is ignored and an unauthenticated request is sent (and fails).

### Steps to Reproduce

1. Install OpenHand canvas (sorry I don’t know how to reproduce it with python/pip/etc)
2. Set the `gitlab_token` and `GITLAB_TOKEN` secrets (somehow both are needed)
3. Create a new automation that clones from a private Gitlab instance
4. Run it

### Acceptance Criteria

- [ ] The clone succeeds

### Installation Method

uv pip install --system --no-deps --force-reinstall https://github.com/OpenHands/automation/archive/e98fcf838901115048f4b933b7b1dc018abd1448.zip

_No response_

### SDK Version

1.42.1

### Version Confirmation

### Python Version

3.13.14

_No response_

### Operating System

Linux

### Logs and Error Messages

_No response_

### Minimal Code Sample

_No response_

_No response_

rsd-darshan:
HUMAN:

Ran the workspace test suite locally (61 tests, incl. the new self-hosted GitLab cases) plus ruff, pyright, and pre-commit before pushing.

---

AGENT:

## Why

`_build_clone_url` only injects the auth token into a full clone URL when the URL's host matches the provider's hardcoded public SaaS domain (`gitlab.com`, `github.com`, `bitbucket.org`). A self-hosted instance (e.g. `https://gitlab.mycompany.com/owner/repo`) never matches, so the token is silently dropped and the clone goes out unauthenticated and fails.

That exact-host check exists intentionally, to stop an auto-detected provider from leaking a token into a lookalike host (`github.com.evil.com`). The fix distinguishes the two cases instead of removing the check:

- **Auto-detected provider** (from the URL itself): keeps the existing exact-match-only behavior, so lookalike hosts still never get a token.
- **Explicitly configured provider** (`RepoSource.provider` set by the caller): the pairing of `url` + `provider` was authored directly by the caller, not derived from an untrusted URL, so the token is now injected using the URL's own host.

`_clone_single_repo` passes `explicit_provider=repo.provider is not None` through to `_build_clone_url` to make this distinction.

## Summary
- Self-hosted git providers (explicit `provider` field) now get the token injected using the URL's actual host, instead of requiring an exact match against the canonical public SaaS domain.
- Auto-detected providers keep the exact-host-only check, so the anti-phishing protection against lookalike domains is unchanged.

## Issue Number

Fixes #4543

## How to Test

`uv run pytest tests/workspace/test_cloud_workspace_repos.py -q` — 61 passed, including:
- `test_build_clone_url_self_hosted_gitlab_with_explicit_provider` — token injected for self-hosted GitLab with explicit provider
- `test_build_clone_url_self_hosted_host_without_explicit_provider` — no token when provider isn't explicit
- `test_build_clone_url_lookalike_host_not_injected` — regression: lookalike host never gets a token
- `test_clone_self_hosted_gitlab_with_token` — end-to-end `clone_repos()` call asserting the authenticated URL passed to `git clone`

Also ran `uv run ruff check` / `ruff format --check` on the changed files, and the repo's pre-commit hooks (ruff, pyright, import-dependency rules, tool registration) — all clean.

## Type

- [x] Bug fix

## Notes

None.

### DIFF
--- a/openhands-sdk/openhands/sdk/workspace/repo.py
+++ b/openhands-sdk/openhands/sdk/workspace/repo.py
@@ -294,10 +294,22 @@ _PROVIDER_CONFIG: dict[GitProvider, tuple[str, str]] = {
 }
 
 
-def _build_clone_url(url: str, provider: GitProvider, token: str | None) -> str:
+def _build_clone_url(
+    url: str,
+    provider: GitProvider,
+    token: str | None,
+    *,
+    explicit_provider: bool = False,
+) -> str:
     """Build authenticated clone URL based on the repository URL and provider.
 
-    Uses proper URL parsing to prevent token injection into malicious URLs.
+    Uses proper URL parsing to prevent token injection into malicious URLs:
+    for an auto-detected provider, the token is only injected if the host
+    matches the provider's public SaaS domain exactly. When the caller
+    explicitly configured the provider (self-hosted instances, e.g. a
+    company GitLab), the token is injected using the URL's own host instead,
+    since that pairing was authored by the caller rather than derived from
+    an untrusted URL.
     """
     config = _PROVIDER_CONFIG.get(provider)
     if not config:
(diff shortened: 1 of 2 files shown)


VascoSch92 (approved):
LGTM!

thanks!

I just corrected a couple of nits.

PR #4596, opened 2026-08-23
fix(agent-server): bypass lifecycle lock for cached event services

### LINKED ISSUE #4514: Lifecycle lock deadlock: thread-pool exhaustion blocks all event loading
### Actual Behavior

`_get_or_load_event_service` in `ConversationService` acquires `lifecycle_lock` and then calls `asyncio.to_thread(_prepare_persisted_runtime)` inside the lock. If the default thread pool is exhausted (all workers stuck on slow I/O), the `to_thread` call queues indefinitely while still holding the lock. Every subsequent `get_event_service` call — including the WebSocket event-stream path and the REST `/events/search` endpoint — blocks waiting for the lock, making the entire agent-server appear wedged.

Reproduce with:
```python
uv run pytest tests/agent_server/test_event_service_thread_pool_exhaustion.py -xvs --timeout=60
```

This caused a production incident where conversations opened but events never loaded after the server ran long enough for a blocking operation to tie up all thread-pool workers.

### Expected Behavior

Already-cached conversations must be retrievable without waiting for the lifecycle lock, so that thread-pool exhaustion for one conversation does not block access to all other conversations.

### Acceptance Criteria

- [ ] `get_event_service` for an already-cached conversation returns immediately even when the lifecycle lock is held by another call
- [ ] Failing test passes after the fix
- [ ] All existing conversation_service and event_service tests pass
- [ ] No race conditions or stale-cache issues introduced by the unlocked cache check

jstar0:
HUMAN:

---

AGENT:

## Why

`ConversationService._get_or_load_event_service` waits on the global lifecycle lock even when the requested conversation already has an open in-memory `EventService`. During a slow persistence/runtime preparation, those waits can consume the server's worker threads and block unrelated event endpoints.

## Summary

- Return an open cached `EventService` before entering lifecycle serialization or disk hydration.
- Preserve the existing activity touch used to defer idle eviction.
- Add a regression test that holds the lifecycle lock and verifies cached access completes immediately.

## Issue Number

Fixes #4514

## How to Test

- `uv run pytest -q tests/agent_server/test_conversation_service.py -k 'cached_event_service_bypasses_lifecycle_lock or get_event_service or conversation_lifecycle' --maxfail=1`
- `uv run pre-commit run --files openhands-agent-server/openhands/agent_server/conversation_service.py tests/agent_server/test_conversation_service.py`
- `make build`

The focused regression/lifecycle tests pass, and the relevant pre-commit hooks pass: Ruff format/lint, pycodestyle, pyright, import rules, and tool registration. The full conversation/event service files were also run; unrelated workspace/git initialization tests fail in this local outer worktree because pytest temporary paths are detected as Git-backed and therefore use conversation worktrees. The changed code is not involved in those failures.

## Video/Screenshots

Not applicable for this backend concurrency fix; the regression is covered by an in-process async test.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This is an internal agent-server scheduling change with no REST/WebSocket schema changes. The cached fast path is limited to open in-memory runtimes; closed or uncached conversations retain the existing lifecycle and hydration path.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -1100,6 +1100,11 @@ class ConversationService:
         event_services = self._event_services
         if event_services is None:
             raise ValueError("inactive_service")
+        event_service = event_services.get(conversation_id)
+        if event_service is not None and event_service.is_open():
+            # Cached runtimes do not need lifecycle serialization or disk I/O.
+            event_service.touch()
+            return event_service
         if (
             conversation_id not in event_services
             and conversation_id not in self._conversation_records
--- a/tests/agent_server/test_conversation_service.py
+++ b/tests/agent_server/test_conversation_service.py
@@ -86,6 +86,26 @@ def sample_stored_conversation():
 
 
 @pytest.mark.asyncio
+async def test_cached_event_service_bypasses_lifecycle_lock(tmp_path):
+    service = ConversationService(conversations_dir=tmp_path / "conversations")
+    conversation_id = uuid4()
+    event_service = MagicMock(spec=EventService)
+    event_service.is_open.return_value = True
+    service._event_services = {conversation_id: event_service}
+
+    await service._lifecycle_lock.acquire()
+    try:
+        result = await asyncio.wait_for(
+            service.get_event_service(conversation_id), timeout=0.1
+        )
+    finally:
+        service._lifecycle_lock.release()
+
+    assert result is event_service
+    event_service.touch.assert_called_once_with()
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


neubig:
This PR is superseded by merged PR #4570, which replaced the global lifecycle lock with per-conversation locking. That broader fix removes the cross-conversation blocking scenario this cached-event-service fast path targeted.

neubig:
Thank you for the focused investigation and implementation. Closing this PR as superseded by merged PR #4570, which addresses the broader lifecycle-lock problem with per-conversation locking.

This action was performed by an AI agent (OpenHands) on behalf of the user.

Issue #4671, opened 2026-08-27
[Epic] Streaming: separate the wire format from the durable event record


VascoSch92:
## The problem

Streamed text leaves the agent on **two channels that know nothing about each other**, and only one of them is tracked.

Durable events go through the callback chain and get written to disk. Token deltas are handed straight to `PubSub` from whichever thread produced them. Nothing orders one against the other, and nothing links them — so the browser puts them back together by comparing strings.

```mermaid
flowchart LR
    U["user callbacks<br/>(incl. the publish wrapper)<br/><b>runs 1st</b>"] --> DC["_default_callback<br/><b>append_event — persists — 2nd</b>"]
    U -.-> PS
    DC --> DISK[("event-{idx:05d}-{id}.json")]
    TOK["on_token(chunk)"] --> PS
    ACP["ACPAgent bridge<br/>(bypasses llm.py)"] --> TOK
    PS["PubSub — no kind filter"]
    PS --> WS["WebSocket"] --> BR["Browser<br/>re-marries by string matching"]
    PS --> WH["Webhook"]
    PS --> TL["Telemetry"]
```

The root cause is one decision: **our socket frame *is* our disk record.**

- `sockets.py:489` sends `event.model_dump(mode="json", exclude_none=True)`
- `event_store.py:217` persists `event.model_dump_json(exclude_none=True)`
- `remote_conversation.py:239` decodes with `Event.model_validate`

Same class, same serialization, same validator — and `Event` is `extra="forbid"`, so *every wire change is a storage migration and every storage change is a wire break*. That is why deltas carry no identity: there was nowhere to put it that didn't cost a migration.

## The whole problem in one sentence

`StreamingDeltaEvent` carries text, a random `id` and a timestamp — and nothing else. No stream identity, no sequence, no attempt number, no pointer to the message that supersedes it.

So the client can't tell a re-streamed retry from new text, can't spot a gap, can't place a delta against a message that arrived beside it, and can't tell when a stream ended. It answers all four by guessing at the text. That is what the 448 lines of `handle-event-for-ui.ts` are: **protocol repair in the view layer.**

## Nobody else ships a delta this way

Checked against the installed packages, not the documentation.

| | What a delta says about itself | How a message starts and ends |
| --- | --- | --- |
| **pi** | `messageId` · `contentIndex` · `kind` | `item_started` → `item_updated` → `item_finished` |
[...]

PR #4697, opened 2026-08-28
perf(sdk): make EventLog.append cost flat against conversation length

### LINKED ISSUE #4676: [Perf] EventLog.append lists the entire event directory on every append
Part of #4671 (step 1).

## What happens

`EventLog.append` (`event_store.py:184`) begins, inside the file lock:

```python
with self._fs.lock(self._lock_path, timeout=LOCK_TIMEOUT_SECONDS):
    # Sync with disk in case another process wrote while we waited
    disk_length = self._count_events_on_disk()
```

and `_count_events_on_disk` (`:235`) does a **full directory listing plus a scan of every path**:

```python
paths = self._fs.list(self._dir)
...
return sum(1 for p in paths
           if posix_path_name(p).startswith("event-") and p.endswith(".json"))
```

So appending event N costs O(N) directory work, under the lock, on the run loop's critical path. Append cost grows linearly with conversation length, and the cost is paid *while holding the lock that intake also wants*.

## Why it matters

Long conversations get progressively slower to write, and the slowdown is invisible in profiles that look at "the agent" rather than at persistence. It also widens the `FIFOLock` contention window exactly when conversations are longest.

## Desired Behavior

Cheaper multi-writer detection than a full listing — an mtime/size check on a marker, a monotonic counter file, or dropping the cross-process sync from the hot path and doing it on open plus on lock contention only. The multi-process case this guards should be stated explicitly first (which deployments actually have two writers on one event dir?), because the answer may be "none" for the default single-writer path.

## Acceptance Criteria

- [ ] Append cost is flat against conversation length. **Measure before and after** — this is one of the baselines in the step-0 issue.

VascoSch92:
HUMAN:

Small perf to make append in log events constant.

---

AGENT:

## Why

`EventLog.append` re-counted the whole events directory on every append to notice a
second writer, and it did so while holding the lock intake also waits on.
`_count_events_on_disk` lists the directory and, on `LocalFileStore`, stats every entry,
so appending event N costs O(N). Long conversations get progressively slower to write,
and the cost hides inside persistence rather than in anything that looks like "the agent".

**Who the second writer actually is** (the issue asks for this first): in this repository
there is exactly one writer per events directory. `ConversationState` owns the only
`EventLog` that appends to a conversation's `events/`; `rerun_actions` writes to a
separate directory it creates itself, and the `tom_consult` executor only reads other
conversations' logs. The guard still covers a second `EventLog` opened on the *same*
directory in the same process, which `test_event_log_concurrent_writes_serialized`
does deliberately. So this PR keeps the guard and makes it cheap rather than dropping it.

### How the marker keeps this O(1)

```
BEFORE — every append:
  append() → list(events/) → stat EVERY file in the directory
                                   |
                    cost grows with log length (O(N) per append)
                    50k-event log ≈ 1.25 BILLION file stats total

AFTER — every append:
  append() → exists(".eventlog-len-N.marker")?
                   |
         ┌────yes──┴──no  (rare: a second writer appended)
         |                        |
    write event               count_events_on_disk()  [O(N), rare fallback]
    delete marker-N                |
    write marker-(N+1)        sync + continue
         |
    O(1) per append
```

Why delete the old marker instead of leaving it behind? Leftover markers would
accumulate in `events/` forever, and the next fresh load/fork/resume
(`_scan_and_build_index`) lists that same directory — so it'd be 2x bigger,
and the O(N) cost this PR removes just moves to a different call site instead
of disappearing. The delete+write pair is 2 tiny file ops per append (100k
total on a 50k-event log), and that flat cost is exactly the "After" number
in the benchmark table below — not overhead on top of it.

## Summary

- `EventLog.append` now answers "has anyone appended since we last did?" with a single
  `exists()` probe of a sidecar marker whose **name** carries the event count. The exact
  count is still paid whenever the marker cannot prove the log is in sync, so multi-writer
  behavior is unchanged, never weaker.
- A fixed-name counter file would not have worked: `FileStore.read` is cached, so a second
  process would keep serving its own stale value. `FileStore.exists` always hits the backend.
- The old marker is deleted *before* the new one is written, so a writer killed between the
  two calls leaves no marker at all and every writer falls back to the exact count.
- `LocalFileStore.delete` no longer raises on a cache miss. A file written by another process
  was never cached locally, and the swallowed `KeyError` logged a spurious `ERROR` on every
  cross-process marker handoff. Pre-existing, but this change is the first to hit it routinely.
- Adds `scripts/event_sourcing_benchmarks/bench_append_scaling.py`, which measures append
  latency against log length. It synthesizes its own events, so it needs no `--eval-dir`.
- Collects the dynamic `ClientAction_*` classes `test_client_tools_persistence` leaves
  behind. See the note at the bottom: the leak predates this branch, but removing the
  directory listing is what stops hiding it.

## Issue Number

Fixes #4676

### 1. The acceptance criterion: append cost is flat against log length

```bash
cd scripts/event_sourcing_benchmarks
python bench_append_scaling.py --events 4000
```

Median `EventLog.append()` latency per 400-event bucket, same machine, 4000 events:

| Log length at append | Before | After |
|---|---|---|
| 0 – 400 | 2.217ms | 0.979ms |
| 400 – 800 | 5.124ms | 0.990ms |
| 800 – 1,200 | 7.866ms | 0.983ms |
| 1,200 – 1,600 | 10.958ms | 0.992ms |
| 1,600 – 2,000 | 13.606ms | 0.974ms |
| 2,000 – 2,400 | 17.152ms | 0.995ms |
| 2,400 – 2,800 | 19.665ms | 0.980ms |
| 2,800 – 3,200 | 22.941ms | 0.982ms |
| 3,200 – 3,600 | 25.964ms | 0.987ms |
| 3,600 – 4,000 | 31.546ms | 0.974ms |
| **Growth factor** | **14.23x** | **1.00x** |

At 4000 events an append goes from 31.5ms to 1.0ms. Short logs get faster too, because
even a 400-event directory was being listed and stat'd on every append.

### 2. Two OS processes appending to one event directory

Unit tests share a `FileStore` (and its cache) inside one process, so they cannot exercise
the case the sync guard exists for. This script gives each process its own `LocalFileStore`,
then cold-loads the result in a third:

<details>
<summary>e2e_multiwriter.py</summary>

```python
import multiprocessing as mp
import re, shutil, sys, tempfile

from openhands.sdk.conversation.event_store import EventLog
from openhands.sdk.event.llm_convertible import MessageEvent
from openhands.sdk.io import LocalFileStore

N_PER_PROC = 100

def make_event(tag, i):
    return MessageEvent(
        id=f"{tag:04x}{i:04x}-0000-0000-0000-000000000000",
        llm_message=Message(role="user", content=[TextContent(text=f"p{tag} e{i}")]),
        source="user",
    )

def writer(root, tag, errors):
    log = EventLog(LocalFileStore(root))
    for i in range(N_PER_PROC):
        try:
            log.append(make_event(tag, i))
        except Exception as e:
            errors.append(f"p{tag} e{i}: {type(e).__name__}: {e}")

def main():
    ctx = mp.get_context("spawn")
    root = tempfile.mkdtemp(prefix="e2e_multiwriter_")
    try:
        with ctx.Manager() as manager:
            errors = manager.list()
            procs = [ctx.Process(target=writer, args=(root, tag, errors)) for tag in range(2)]
            for p in procs: p.start()
            for p in procs: p.join()
            errors = list(errors)

        expected = 2 * N_PER_PROC
        fs = LocalFileStore(root)
        names = [n.rsplit("/", 1)[-1] for n in fs.list("events")]
        event_names = [n for n in names if n.startswith("event-")]
        markers = [n for n in names if ".marker" in n]
        indices = sorted(int(re.match(r"event-(\d+)-", n).group(1)) for n in event_names)
        log = EventLog(fs)  # fresh cold load from disk
        ids = [e.id for e in log]

        print(f"append errors:          {len(errors)}")
        print(f"event files on disk:    {len(event_names)} (expected {expected})")
        print(f"index range contiguous: {indices == list(range(expected))}")
        print(f"cold-loaded events:     {len(log)}")
        print(f"distinct event ids:     {len(set(ids))}")
        print(f"count markers left:     {len(markers)} {markers}")
        ok = (not errors and len(event_names) == expected
              and indices == list(range(expected)) and len(log) == expected
              and len(set(ids)) == expected and len(markers) == 1)
        print("\nRESULT:", "PASS" if ok else "FAIL")
        return 0 if ok else 1
    finally:
        shutil.rmtree(root, ignore_errors=True)

if __name__ == "__main__":
    sys.exit(main())
```

</details>

```
append errors:          0
event files on disk:    200 (expected 200)
index range contiguous: True
cold-loaded events:     200
distinct event ids:     200
count markers left:     1 ['.eventlog-len-200.marker']

RESULT: PASS
```

No lost or duplicated indices, no interleaving damage, and exactly one marker survives.
Run against the pre-fix code it also passes, so this is a parity check, not a new guarantee.

### 3. Tests

```bash
uv run pytest tests/sdk/ -q                   # 5973 passed, 0 failed
uv run pytest tests/agent_server/ -q          # 2083 passed, 0 failed
uv run pre-commit run --files <changed files> # all hooks pass
```

Four new tests in `tests/sdk/conversation/test_event_store.py` cover the flat-cost
guarantee (append must not list the directory as the log grows), a second writer still
being picked up, a legacy log that has no marker, and the scan no longer warning about
sidecar files. One new test in `tests/sdk/io/test_filestore_cache.py` covers the
`delete()` cache miss.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **On-disk change.** The events directory gains one more dotfile sidecar next to the
  existing `.eventlog.lock`. Nothing in the repo enumerates that directory except
  `EventLog` itself, which already filters to `event-*.json` when counting. The cold scan
  now skips dotfiles silently instead of logging `Unrecognized event file name` for them,
  which also silences the warning `.eventlog.lock` has always produced. One existing test
  that counted files in the directory was updated to filter dotfiles rather than only `.lock`.
- **A latent class-registry leak this PR had to fix.** `test_client_tools_persistence`
  registers the same client tool name twice (create, then resume), and each spec builds a
  new `ClientAction_*` class. Both are reachable only through reference cycles, so they stay
  in `Action.__subclasses__()` until a cyclic collection runs, and the duplicate-name check
  in `_get_checked_concrete_subclasses` then fails in *unrelated* later tests — three fork
  and state-serialization tests, with `Duplicate class definition ... ClientAction_persist_navigate_to`.
  The leak is pre-existing: on a clean `main`, running
  `tests/sdk/conversation/local/test_client_tools_persistence.py` together with
  `tests/sdk/conversation/local/test_fork.py` already fails. Full runs on `main` happened to
  pass because the old O(N) directory listing allocated enough garbage to trigger a
  collection in time; appending without it does not, which is how this PR surfaced it.
  Confirmed both directions: this branch with `gc.set_threshold(100, 5, 5)` is green without
  the fixture, and `main` with the fixture fixes the two-file failure. The fix is a teardown
  `gc.collect()` plus `clear_subclass_cache()` in that test module, so the leak stays local
  regardless of allocation patterns.
- **Not fixed here.** `tests/sdk/` and `tests/agent_server/` run in a single pytest process
  produce 17 failures (`ValueError: Local classes not supported! test_execute_tool.ContextAwareTool`).
  This reproduces identically on a clean `main` from just
  `tests/sdk/conversation/local/test_execute_tool.py` plus
  `tests/agent_server/test_openapi_discriminator.py`, and is unrelated to this change.
  Both suites pass when run separately, as CI runs them.
- Step 1 of #4671. The benchmark script is reusable for the other steps in that epic.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:99975b6-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-99975b6-python \
  ghcr.io/openhands/agent-server:99975b6-python
```

```
ghcr.io/openhands/agent-server:99975b6-golang-amd64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-golang-amd64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-golang-amd64
ghcr.io/openhands/agent-server:99975b6-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:99975b6-golang-arm64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-golang-arm64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-golang-arm64
ghcr.io/openhands/agent-server:99975b6-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:99975b6-java-amd64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-java-amd64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-java-amd64
ghcr.io/openhands/agent-server:99975b6-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:99975b6-java-arm64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-java-arm64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-java-arm64
ghcr.io/openhands/agent-server:99975b6-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:99975b6-python-amd64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-python-amd64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-python-amd64
ghcr.io/openhands/agent-server:99975b6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:99975b6-python-arm64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-python-arm64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-python-arm64
ghcr.io/openhands/agent-server:99975b6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:99975b6-python-slim-amd64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-python-slim-amd64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-python-slim-amd64
ghcr.io/openhands/agent-server:99975b6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-amd64
ghcr.io/openhands/agent-server:99975b6-python-slim-arm64
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-python-slim-arm64
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-python-slim-arm64
ghcr.io/openhands/agent-server:99975b6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-arm64
ghcr.io/openhands/agent-server:99975b6-golang
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-golang
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-golang
ghcr.io/openhands/agent-server:99975b6-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:99975b6-java
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-java
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-java
ghcr.io/openhands/agent-server:99975b6-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:99975b6-python-slim
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-python-slim
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-python-slim
ghcr.io/openhands/agent-server:99975b6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim
ghcr.io/openhands/agent-server:99975b6-python
ghcr.io/openhands/agent-server:99975b6edc6cd04993557cf6ce9144dab053387e-python
ghcr.io/openhands/agent-server:vasco-perf-4676-flat-eventlog-append-python
ghcr.io/openhands/agent-server:99975b6-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `99975b6-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `99975b6-python-amd64`) are also available if needed
</details>


openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -22,6 +22,13 @@
 LOCK_FILE_NAME = ".eventlog.lock"
 LOCK_TIMEOUT_SECONDS = 30
 
+# Sidecar marker whose *name* carries the event count the log had when it was
+# last written. Detecting another writer by reading a fixed-name counter file
+# would not work: ``FileStore.read`` is cached and a second process would keep
+# serving its own stale value. ``FileStore.exists`` always hits the backend, so
+# putting the count in the name keeps the check O(1) and cache-free.
```
VascoSch92: Smaller comment.

openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -22,6 +22,13 @@
...
+LENGTH_MARKER_PATTERN = ".eventlog-len-{length}.marker"
```
VascoSch92: Use Final from typing with str

openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -193,10 +200,13 @@ def append(self, event: Event) -> None:
 
         try:
             with self._fs.lock(self._lock_path, timeout=LOCK_TIMEOUT_SECONDS):
-                # Sync with disk in case another process wrote while we waited
-                disk_length = self._count_events_on_disk()
-                if disk_length > self._length:
-                    self._sync_from_disk(disk_length)
+                # Sync with disk in case another writer appended while we waited.
+                # The marker answers "nobody has appended since we last did" in
+                # one stat; only when it cannot, pay for the full listing.
```
VascoSch92: smaller comment please

openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -224,14 +234,46 @@ def append(self, event: Event) -> None:
...
+    def _marker_path(self, length: int) -> str:
+        return f"{self._dir}/{LENGTH_MARKER_PATTERN.format(length=length)}"
+
+    def _marker_matches_length(self) -> bool:
+        """Whether the marker proves no one has appended since we last did.
+
+        ``False`` is not proof of divergence -- a log written before this
+        marker existed has none either, and so does one whose writer died
+        mid-update. Callers must fall back to the exact count, which keeps
+        this an optimization rather than a change of behavior.
```
VascoSch92: not so verbose.

openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -224,14 +234,46 @@ def append(self, event: Event) -> None:
...
+        except Exception as e:
+            logger.warning("Error reading event count marker in %s: %s", self._dir, e)
+            return False
+
+    def _advance_length_marker(self, previous_length: int) -> None:
+        """Move the marker from ``previous_length`` to the current length.
+
+        The old marker goes first on purpose: interrupted between the two
+        calls, the log is left with no marker at all, which costs every writer
+        one full listing instead of letting a stale marker claim to be current.
```
VascoSch92: not so verbose

openhands-sdk/openhands/sdk/conversation/event_store.py
VascoSch92: Cut from five lines to two.

openhands-sdk/openhands/sdk/conversation/event_store.py
VascoSch92: Done: `LENGTH_MARKER_PATTERN: Final[str]`, matching the `Final` constants in `llm_profile_store.py`.

openhands-sdk/openhands/sdk/conversation/event_store.py
VascoSch92: Cut to a single line: "Sync with disk only if the marker cannot rule out another writer".

openhands-sdk/openhands/sdk/conversation/event_store.py
VascoSch92: Trimmed to two lines, keeping only that a `False` result is not proof of divergence so callers must fall back to the exact count.

openhands-sdk/openhands/sdk/conversation/event_store.py
VascoSch92: Trimmed to two lines, keeping only why the delete comes first: an interrupted update leaves no marker rather than a stale one.

openhands-sdk/openhands/sdk/conversation/event_store.py
```diff
@@ -224,14 +229,43 @@ def append(self, event: Event) -> None:
...
+            return bool(self._fs.exists(self._marker_path(self._length)))
+        except Exception as e:
+            logger.warning("Error reading event count marker in %s: %s", self._dir, e)
+            return False
+
+    def _advance_length_marker(self, previous_length: int) -> None:
+        """Move the marker from ``previous_length`` to the current length.
+
+        Deleting first is deliberate: an interrupted update leaves no marker
+        rather than a stale one claiming to be current.
```
enyst, line 265: Are we sure this is necessary? It adds 100k delete/writes on 50k iterations conversations 🤔

VascoSch92: @enyst 

Yes, it's necessary but the cost is already take in consideration in the benchmark above. 

Every append does 2 tiny sidecar ops (delete old marker, write new one) instead of listing the whole directory.

The problem in skipping the delete is to leave one dead marker file per event. So we are not solving the problem in this way.

I asked the agent to put a diagram of the new strategy in the PR description. So it is easy to understand what is going on :-)

VascoSch92:
@all-hands-bot fix the conflicts in this PR

enyst (approved):

PR #4689, opened 2026-08-27
fix(agent-server): stop fanning streaming deltas out to every subscriber

VascoSch92:
HUMAN:

Fixing a bug where we were publishing every delta instead of just the event.

---

AGENT:

## Why

`StreamingDeltaEvent` is published straight to the conversation `PubSub` (`event_service.py:1061`), and `PubSub.__call__` notified **every** registered subscriber with no kind filter. Making the delta an `Event` opted it into every consumer of the bus by default, so consumers sized for *conversation* events started receiving traffic proportional to *tokens*:

- **#4672** — `WebhookSubscriber.__call__` enqueues unconditionally, and `event_buffer_size` defaults to 5, so a streamed response became a continuous POST loop to the configured endpoint for its whole duration. Model output is not masked on the standard `Agent` path, so that text was leaving the process one fragment at a time.
- **#4673** — `TelemetrySubscriber._handle` increments `_event_count` *before* its `match`, so `event_count_bucket` scaled with model verbosity instead of conversation activity, making streamed and non-streamed conversations incomparable.

Same root cause, so one fix closes both.

## Summary

- Add `Subscriber.receives_streaming_deltas` (`ClassVar`, defaults to `False`) and consult it in `PubSub.__call__`: deltas are delivered only to subscribers that opt in.
- `_WebSocketSubscriber` opts in — the live socket is the one consumer token streaming exists for.
- Regression tests for both acceptance criteria, plus one for the central bus behaviour.

Delivery is **opt-in rather than opt-out** so a subscriber added later cannot inherit token-rate traffic — and the content it carries — by accident. This is option 2 from #4672 ("interim: a kind filter"); it does not preclude the epic's structural direction of taking deltas off the bus entirely.

## Issue Number

Fixes #4672
Fixes #4673

## How to Test

**End-to-end — real HTTP egress.** The script below runs a real HTTP receiver, a real `PubSub` and a real `WebhookSubscriber`, then publishes one streamed response (3 conversation events + 200 token deltas) with the production default `event_buffer_size=5`. It uses no new API, so it runs unchanged on `main`:

<details>
<summary><code>webhook_delta_e2e.py</code></summary>

```python
import asyncio, json, threading
from http.server import BaseHTTPRequestHandler, HTTPServer
from unittest.mock import MagicMock
from uuid import uuid4

from openhands.agent_server.config import WebhookSpec
from openhands.agent_server.conversation_service import WebhookSubscriber
from openhands.agent_server.pub_sub import PubSub
[...]

Issue #4683, opened 2026-08-27
Streaming step 7: stop emitting StreamingDeltaEvent and freeze the legacy events endpoint


VascoSch92:
Part of #4671 (step 7). Needs step 6.

## Desired Behavior

Once every client speaks the session endpoint, stop producing the old channel.

## Work

1. **Stop emitting `StreamingDeltaEvent`.** Deltas become progress frames on the session socket only. This is what structurally retires the two unfiled bugs: nothing left for telemetry to count, nothing left for webhooks to POST. Any interim kind filter added earlier can come out with it.
2. **Freeze `/sockets/events/{id}`** and announce its removal. Nothing decodes wrong — it just stops being offered.
3. Remove the legacy `resend_mode` / `after_timestamp` replay parameters, superseded by the `after_seq` cursor.

## Why it is last

Every client has to have switched first, which is step 6 plus the two client PRs. Until then the old endpoint is the only thing some clients speak.

## Acceptance Criteria

- [ ] `StreamingDeltaEvent` has no construction site.
- [ ] A webhook receives the same event count for a streamed and non-streamed conversation, with no filter in the path.
- [ ] The legacy endpoint is documented as deprecated with a removal target.

Issue #4673, opened 2026-08-27
[Bug]: telemetry event_count_bucket counts token deltas, making it incomparable across conversations


VascoSch92:
Part of #4671.

## Actual Behavior

`TelemetrySubscriber._handle` (`telemetry/subscriber.py:107`) increments unconditionally, **before** its `match`:

```python
def _handle(self, event: Event) -> None:
    self._event_count += 1

    match event:
        case AgentErrorEvent():   ...
        case ConversationErrorEvent():   ...
        case ConversationStateUpdateEvent():   ...
```

The subscriber registers through `event_service.subscribe_to_events` (`conversation_service.py:2319`), i.e. the same unfiltered `PubSub` as everything else, so it receives every token delta. The counter ships as `event_count_bucket`.

**Repro**

```bash
uv run pytest tests/agent_server/telemetry/ -q
```

Green today: no test asserts that deltas are excluded from `event_count_bucket`. To see the skew, run one conversation with `stream=true` and one without and compare the reported bucket.

## Why it matters

The metric silently changes meaning depending on whether streaming was on and how verbose the model was. A streamed conversation with three tool calls can outscore a non-streamed conversation with thirty. **Any dashboard or alert built on `event_count_bucket` is comparing token volume against conversation activity** — and there is no way to tell from the data which one you're looking at.

## Fix

Move the increment inside the `match`, or filter deltas before the subscriber sees them (same root cause, same one-line class of fix as the webhook issue). Note that either choice is a **discontinuity in the existing metric**; that's worth a heads-up to whoever consumes it rather than a silent correction.

## Acceptance Criteria

- [ ] Two identical conversations, one with `stream=true` and one without, report the same `event_count_bucket`.

VascoSch92:
Note for anyone landing here from #4671: this was closed by #4689 with a kind filter (`Subscriber.receives_streaming_deltas`), which is deliberately interim — the structural half, taking deltas off the shared bus entirely, is now **#4696**, added to the epic as step 1b. That step removes the flag.

One side effect of the filter is tracked separately in **#4695**: `_EventSubscriber` does not opt in, and it is the subscriber that resets the runtime idle timer, so a long stream no longer keeps a managed pod alive.

Issue #4672, opened 2026-08-27
[Bug]: every token delta is POSTed to configured webhooks


VascoSch92:
Part of #4671.

## Actual Behavior

If a conversation has a webhook configured and the agent streams, **every token delta is POSTed to that webhook.**

`WebhookSubscriber.__call__` (`conversation_service.py:2510`) takes `Event` and enqueues unconditionally — there is no kind filter anywhere on the path:

```python
async def __call__(self, event: Event):
    """Add event to queue and post to webhook when buffer size is reached."""
    self._enqueue(event)
    if (len(self.queue) >= self.spec.event_buffer_size
            or self._queue_bytes >= self.spec.max_batch_bytes):
        ...
        await self._post_events()
```

`event_buffer_size` defaults to **5** (`config.py:75-81`). Webhooks register via `event_service.subscribe_to_events` (`conversation_service.py:2233`) → `self._pub_sub.subscribe(...)` — the same `PubSub` that `_publish_stream_delta` publishes to (`event_service.py:1058`) — and `PubSub.__call__` (`pub_sub.py:81-102`) fans out to every subscriber with no filtering.

Because `_post_events` loops `while events_remaining` and coalesces when a post is already in flight, the practical behaviour is a **continuous POST loop for the entire duration of every streamed response**, driven at whatever rate the endpoint can absorb.

**Repro**

```bash
uv run pytest tests/agent_server/test_webhook_subscriber.py -q
```

The suite passes today precisely because nothing in it asserts a kind filter. To watch the egress directly, point a webhook at a request bin and stream one response:

```bash
uv run agent-server
```

Every token delta arrives as its own POST body.

## Why it matters

1. **Volume and cost.** A webhook receiver sized for conversation events gets traffic proportional to *tokens*. With retries (`num_retries: 3`) on a struggling endpoint this amplifies further.
2. **Content that shouldn't leave.** On the standard `Agent` path, model output is **not masked at all** (see the companion issue) — `mask_secrets_in_output` has exactly two non-ACP call sites and both are tool output. So unmasked model text is being POSTed to a configured endpoint, one fragment at a time.
3. **Nobody chose this.** `StreamingDeltaEvent` is referenced three times in the whole agent-server: an import, a comment, and its construction site. There is no `isinstance` check and no filter. Making the delta an `Event` opted it into every consumer of the bus by default.

## Fix

Two options, in preference order:

[...]

PR #4700, opened 2026-08-28
refactor(sdk)!: StreamingDeltaEvent stops subclassing Event and gets its own fan-out

### LINKED ISSUE #4695: [Bug]: token deltas no longer reset the runtime idle timer, so a long stream can have its pod reaped
### Bug Description

`#4689` made streaming-delta delivery opt-in (`Subscriber.receives_streaming_deltas`, `pub_sub.py:18`, filtered in `PubSub.__call__` at `pub_sub.py:94-95`). Only `_WebSocketSubscriber` opts in (`sockets.py:543`).

`_EventSubscriber` (`conversation_service.py:2433`) does not, and it is the subscriber that calls `update_last_execution_time()` (`:2446`). That function drives `idle_time` on `/server_info` (`server_details_router.py:122`), which the runtime-api uses to reap idle pods.

So token deltas no longer keep a managed runtime alive. This was never a stated behaviour of `_EventSubscriber` — it was an accident of deltas being `Event`s on a shared bus — but it was load-bearing, and `#4689` removed it silently.

The concern is already known and handled elsewhere: `_setup_acp_activity_heartbeat` (`event_service.py:894-914`) exists precisely because ACP tool calls do not touch the HTTP endpoints, and its docstring says *"the runtime-api sees growing idle_time and kills the pod (~20 min)"*. ACP agents therefore still have a heartbeat. The standard streaming path no longer has one.

### Expected Behavior

A long streamed completion keeps the runtime idle timer fresh, as it did before `#4689`.

### Actual Behavior

Nothing resets the idle timer during a single streamed completion on the standard (non-ACP) path. The remaining resets are durable events (`conversation_service.py:2446`), client calls to the file/git/bash routers, and the ACP-only heartbeat. The websocket carrying the deltas does not touch any of them.

A completion that streams for longer than the injected idle threshold, producing no durable event, can have its pod reaped mid-stream.

Verified in both directions with `uv run python`, using the real `_EventSubscriber` and the real `PubSub`:

| Tree | Delta resets idle timer | Durable event resets it |
|---|---|---|
| `e38e02b38` — immediately before #4689 | yes (1 call) | yes (1 call) |
| `9a24f6c88` — #4689 merged | **no (0 calls)** | yes (1 call) |

```python
# repro_idle.py — run with: uv run python repro_idle.py
import asyncio

import openhands.agent_server.conversation_service as cs
from openhands.agent_server.conversation_service import _EventSubscriber
from openhands.agent_server.pub_sub import PubSub
from openhands.sdk.event import StreamingDeltaEvent

calls = []
cs.update_last_execution_time = lambda: calls.append(1)


async def main() -> None:
    ps = PubSub()
    ps.subscribe(_EventSubscriber(service=MagicMock()))
    await ps(StreamingDeltaEvent(content="tok"))
    print("idle timer resets from one token delta:", len(calls))


asyncio.run(main())
```

Prints `1` on `e38e02b38` and `0` on `9a24f6c88`.

### Steps to Reproduce

1. `git worktree add --detach /tmp/premerge e38e02b38`
2. `git worktree add --detach /tmp/postmerge 9a24f6c88`
3. Save the script above into each tree and run `uv run python repro_idle.py` in both.
4. Observe `1` before the merge and `0` after.

### Acceptance Criteria

- [ ] A token delta published to the conversation `PubSub` resets the runtime idle timer again on the standard streaming path.
- [ ] The reset is throttled rather than fired per token, matching `_ACTIVITY_SIGNAL_INTERVAL` as used by `_maybe_signal_activity` in `acp_agent.py`.
- [ ] A regression test asserts that publishing deltas over a period longer than the idle threshold, with no durable event, leaves `idle_time` below that threshold.
- [ ] Whichever fix is chosen stays valid when deltas move off the shared bus in #4671, so the heartbeat is not lost a second time.

### Installation Method

Other

### SDK Version

`main` @ `9a24f6c88`

### LINKED ISSUE #4696: Streaming step 1b: StreamingDeltaEvent stops subclassing Event and gets its own fan-out
### Problem or Use Case

Part of #4671. This is the step that answers @enyst's question on #4672:

> I think maybe deltas should not be Events. They don't seem to match Event in more than one way, persistence, pubsub, strictness of the format, shouldn't affect compatibility with anything except their direct clients (visual clients typically)…

The answer is yes, and nothing at runtime requires it. `StreamingDeltaEvent` is never persisted, is not `LLMConvertibleEvent`, and never appears in replay. The socket serializer calls `model_dump` on it, which works on any Pydantic model — `_send_bash_event` (`sockets.py:548`) already does exactly that with `BashEventBase`, which is a plain `DiscriminatedUnionMixin` (`models.py:576`) with its own `id` and `timestamp`, its own `PubSub` (`bash_service.py:34`), its own subscriber and its own endpoint. The delta is the same kind of object.

`#4689` fixed the fan-out symptom with an opt-in flag. This is the structural half its description points at: *"it does not preclude the epic's structural direction of taking deltas off the bus entirely."*

The reason it matters beyond tidiness is `extra="forbid"` on `Event` (`base.py:21`). Every additive field is a break for any client on an older schema, which is why the delta carries no identity today. Verified with `uv run python`: validating `{"content": "x", "item_id": "abc", "order": 3}` succeeds against a `DiscriminatedUnionMixin` delta and raises `ValidationError` against the current `Event`-based one. That is what lets `StreamContext` be built and tested against the endpoint that exists today rather than only against the one step 3 introduces.

### Desired Behavior

`StreamingDeltaEvent` stops subclassing `Event` and becomes a `DiscriminatedUnionMixin` with explicit `id`, `timestamp` and `source`, and deltas get their own fan-out rather than a filter on the durable one.

**The wire must not change.** `kind` is `self.__class__.__name__` (`models.py:199-201`), so keeping the class name keeps the wire value. Verified: the dump is identical to today's, field for field, once `source` is re-declared — `{"id": …, "timestamp": …, "source": "agent", "content": "hel", "kind": "StreamingDeltaEvent"}`.

Three traps, all found by testing rather than reading:

- **`source` must stay.** The canvas guard `isBaseEvent` (`OpenHands:src/types/agent-server/type-guards.ts:45`) requires `id`, `timestamp` and a `source` from a fixed set. Drop it and every delta is discarded silently, with no error anywhere.
- **The class must not be renamed.** The canvas matches `kind === "StreamingDeltaEvent"` as a literal string.
- **The idle heartbeat must come across.** See the separate bug filed against the `#4689` behaviour. Whatever resets the runtime idle timer during a stream has to survive this move too.

Version skew is benign in the one direction that matters. An older Python client resolves `kind` against its own local class registry, which still lists the delta under `Event`, so the identical frame still decodes. Verified. Only the new client needs the decode branch, in the same release.

### Acceptance Criteria

- [ ] `StreamingDeltaEvent` no longer subclasses `Event`; `isinstance(delta, Event)` is `False`.
- [ ] Its serialized frame is byte-identical to today's, asserted by a test, including `kind` and `source`.
- [ ] Deltas are published to a delta-only `PubSub`, and `Subscriber.receives_streaming_deltas` (added in #4689) is removed as dead code.
- [ ] `RemoteConversation` decodes delta frames without routing them through `Event.model_validate`, and durable-event callbacks keep their `Callable[[Event], None]` signature.
- [ ] The runtime idle timer is still reset during a long stream that produces no durable events.
- [ ] A test asserts a webhook subscriber and a telemetry subscriber receive no delta during a streamed run, holding the guarantee #4689 established.
- [ ] The exported OpenAPI no longer lists `StreamingDeltaEvent` in the `Event` union; `typescript-client` regenerated.
- [ ] `Event` itself is untouched: same class, same `extra="forbid"`, same bytes on disk.

### Alternatives Considered

Leave it until step 7 (#4683) deletes the class outright. Cheaper in total work, but step 7 depends on step 6, so it forfeits the ability to develop `StreamContext` against the current endpoint, and leaves the fan-out guarantee resting on a flag each new subscriber must remember not to set.

### Priority / Severity

Medium

### Feature Area

Agent Server

VascoSch92:
HUMAN:

DeltaToken are not anymore subclasses of Events

---

AGENT:

**Design doc:** [Unparenting the Delta](https://github.com/OpenHands/software-agent-sdk/blob/vasco/delta-4696-unparent/.pr/design.md) — before/after shapes, the fan-out split, the four traps, and the compatibility table.

## Why

`StreamingDeltaEvent` is transient: never persisted, never replayed, never converted to an LLM message. Modelling it as an `Event` opted it into every consumer of the durable bus by default, which is how webhooks (#4672) and telemetry (#4673) came to receive one POST per token. Nobody chose that; there was simply nowhere to exclude it.

#4689 fixed the symptom with an opt-in flag on `Subscriber`, and its own description called that interim: it *"does not preclude the epic's structural direction of taking deltas off the bus entirely."* This is that structural half, and it removes the flag.

It also matters beyond tidiness. `Event` sets `extra="forbid"`, so every additive field is a break for any client on an older schema — which is why the delta carries no stream identity today. Off `Event`, it can gain `item_id`, `attempt` and `order` while older clients ignore them, so `StreamContext` (step 5) can be developed against the endpoint that exists today rather than only against the one step 3 introduces.

Finally, #4689's flag dropped something nobody intended. The subscriber that resets the runtime idle timer does not opt in, so a long stream producing no durable events stopped keeping a managed pod alive (#4695). This carries that heartbeat across.

## Summary

- `StreamingDeltaEvent` subclasses `DiscriminatedUnionMixin` instead of `Event`, declaring `id`, `timestamp` and `source` explicitly. The class name is unchanged, so `kind` is unchanged, and the emitted frame is byte-identical field for field.
- `EventService` gains a delta-only `PubSub` with `subscribe_to_deltas` / `unsubscribe_from_deltas`; the events websocket subscribes to both buses, and `Subscriber.receives_streaming_deltas` is deleted as dead code. Webhooks and telemetry cannot receive a delta because they are registered on a different object.
- A throttled (30 s) heartbeat subscriber on the delta bus restores the runtime idle-timer reset during a stream, fixing #4695.
- `RemoteConversation` branches on `kind` before `Event.model_validate` and routes deltas to a new `delta_callbacks` list, so durable callbacks keep their `Callable[[Event], None]` signature.
- The delta stays frozen, as it was under `Event`, because one instance fans out to several subscribers at once. Only `extra` is loosened, which is the part that lets stream identity be added later.
- The exported OpenAPI drops the delta from the `Event` union (18 members, was 19) and keeps it as a standalone named schema. See the compatibility note below for why that line matters.

### REST API contract changes

Compared with base OpenAPI `ccde94913e4f` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -1431 +1431 @@
-schema Event oneOf=[ServerErrorEvent,ACPToolCallEvent,Condensation,CondensationRequest,CondensationSummaryEvent,ConversationErrorEvent,ConversationStateUpdateEvent,HookExecutionEvent,LLMCompletionLogEvent,ActionEvent,MessageEvent,AgentErrorEvent,ObservationEvent,UserRejectObservation,SystemPromptEvent,StreamingDeltaEvent,TokenEvent,InterruptEvent,PauseEvent]
+schema Event oneOf=[ServerErrorEvent,ACPToolCallEvent,Condensation,CondensationRequest,CondensationSummaryEvent,ConversationErrorEvent,ConversationStateUpdateEvent,HookExecutionEvent,LLMCompletionLogEvent,ActionEvent,MessageEvent,AgentErrorEvent,ObservationEvent,UserRejectObservation,SystemPromptEvent,TokenEvent,InterruptEvent,PauseEvent]
@@ -2534,8 +2533,0 @@
-schema StreamingDeltaEvent property content optional schema=anyOf=[type="string",type="null"]
-schema StreamingDeltaEvent property id optional schema=type="string"
-schema StreamingDeltaEvent property kind required schema=type="string" const="StreamingDeltaEvent"
-schema StreamingDeltaEvent property parent_id optional schema=anyOf=[type="string",type="null"]
-schema StreamingDeltaEvent property reasoning_content optional schema=anyOf=[type="string",type="null"]
-schema StreamingDeltaEvent property source optional schema=type="string" enum=["agent","user","environment","hook"] default="agent"
-schema StreamingDeltaEvent property timestamp optional schema=type="string"
-schema StreamingDeltaEvent type="object" additionalProperties=false
```

## Issue Number

Fixes #4696. Also fixes #4695.

## How to Test

Unit tests, covering each acceptance criterion:

```
uv run pytest tests/agent_server/test_event_streaming.py \
              tests/agent_server/test_webhook_subscriber.py \
              tests/agent_server/telemetry \
              tests/sdk/event/test_streaming.py \
              tests/sdk/conversation/remote/test_websocket_client.py -q
```

Full suites (run separately — combining `tests/sdk` and `tests/agent_server` in one process has a pre-existing collection interaction):

```
uv run pytest tests/agent_server -q -n 4     # 2097 passed
uv run pytest tests/sdk -q -n 4              # 5973 passed
uv run pyright                               # 0 errors
make test-server-schema                      # quality check + swagger validate pass
```

End-to-end, against a real agent-server. The script below boots uvicorn with a real webhook receiver attached, patches only the litellm transport so a scripted response actually streams, and drives a real `RemoteConversation` over the real events websocket:

```
                    END-TO-END RESULT (real server, socket, webhook)
  tokens the LLM streamed        : 5
  deltas at delta_callbacks      : 6
  delta text reassembled         : 'Hello, streamed world!'
  deltas at durable callbacks    : 0   <- was where they landed before
  durable events at callbacks    : 10
  webhook POSTs                  : 11
  deltas POSTed to webhook       : 0
  first delta frame on the wire  : {"content": "Hello",
                                    "id": "747b8244-c15a-4de9-a7f8-db9b1bd5d382",
                                    "kind": "StreamingDeltaEvent",
                                    "source": "agent",
                                    "timestamp": "2026-08-28T14:15:54.447318"}
```

That frame is what the browser already receives today, field for field. `parent_id` was always `None` and always dropped by `exclude_none`, so removing it changes nothing that ever left the process.

Idle-timer heartbeat, the same script run on both trees:

```
on origin/main            tree has a delta bus: False   idle timer reset by the delta: False
on this branch            tree has a delta bus: True    idle timer reset by the delta: True
```

## Video/Screenshots

Terminal output above; this change has no UI surface. The canvas needs no change at all — its `isBaseEvent` guard requires `id`, `timestamp` and a known `source`, and its `kind === "StreamingDeltaEvent"` comparison is a literal string match. Both still hold, and a test asserts the exact frame key set so a future edit cannot silently break them.

## Design Doc

https://github.com/OpenHands/software-agent-sdk/blob/vasco/delta-4696-unparent/.pr/design.md

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [x] Breaking change
- [ ] Docs / chore

### Two kinds of compatibility, kept apart

Two different things could break here. Separating them makes the rest simple.

**The data on the network is unchanged.** The server sends one small JSON object per token. I dumped that object on this branch and on the pre-change tree with a fixed `id` and `timestamp`, then diffed them. Identical, character for character. The browser only ever sees this object, so the browser cannot break.

**The names other code imports do change**, and only for Python inside this repo:

- `isinstance(delta, Event)` is now `False`. That is the intended semantic change, and the reason this should not ride in a patch release.
- A `RemoteConversation` caller who filtered deltas out of `callbacks` moves to `delta_callbacks`.
- `Subscriber.receives_streaming_deltas` is gone. It shipped one release ago in #4689, and nothing outside this repo is known to set it.

`Event` itself is untouched: same class, same `extra="forbid"`, same bytes on disk.

### Why this is not an API change for OpenHands/OpenHands

The chain runs:

```
SDK              defines the delta
  -> OpenAPI     a document exported from the SDK
  -> TS client   generates TypeScript types from that document
  -> the app     imports some of those types
```

In the generated TypeScript the delta appears twice. Once inside the large `Event` union, and once as a type of its own. This PR takes it out of the union, which is the intended result: a delta is never stored, and no REST route can return one.

That created a trap worth knowing about. The exporter deletes any schema no REST route mentions, and a delta only ever travels over a websocket. So the moment it left the union it would have vanished from the document altogether, breaking `typescript-client`, which re-exports the standalone type at `src/events/types.ts:70` and again from its package root at `src/index.ts:85`. The addition in `_add_canonical_contract_components` exists purely to keep that type in the document.

That is the one part of this PR I would call separable. Happy to drop it if the preference is to let the TS type go now rather than at step 7.

The app is then insulated twice over, and either protection alone would be enough:

- it pins one exact client version, `1.39.0`, so nothing reaches it until someone raises that number deliberately;
- it never imports the delta or the `Event` union from the client, and declares its own delta interface by hand in `src/types/agent-server/core/events/streaming-delta-event.ts`.

Checked while writing this: the app has no `openhands-sdk` dependency at all, and its only Python is CI scripts and test mocks. The three Python changes above are unreachable from it.

### Release ordering

This merges and releases in the SDK. `typescript-client` then regenerates and releases, where the delta moves out of the generated union while its public alias keeps working. The app bumps its pin if and when it chooses. Nothing moves underneath anyone.

### Review fixes already applied

Three issues found reviewing my own commits, all fixed in `99f7549`:

- the delta silently became mutable when it stopped inheriting `Event`'s frozen config, so `frozen=True` is restored;
- the heartbeat throttle compared against a `0.0` baseline, but `monotonic()` counts from an arbitrary origin, which is uptime on Linux, so the first beat could be swallowed on a freshly booted host. It now uses a `None` sentinel;
- `_on_delta` isolated per-callback failures, which `compose_callbacks` does not do for durable events. Removed, so both callback paths behave alike.

### Follow-ups

`typescript-client` regeneration is a separate PR in that repo. Step 7 (#4683) deletes this class and its bus outright once the canvas turn model lands, and `delta_callbacks` goes with it.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:0bfc677-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-0bfc677-python \
  ghcr.io/openhands/agent-server:0bfc677-python
```

```
ghcr.io/openhands/agent-server:0bfc677-golang-amd64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-golang-amd64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-golang-amd64
ghcr.io/openhands/agent-server:0bfc677-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:0bfc677-golang-arm64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-golang-arm64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-golang-arm64
ghcr.io/openhands/agent-server:0bfc677-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:0bfc677-java-amd64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-java-amd64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-java-amd64
ghcr.io/openhands/agent-server:0bfc677-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:0bfc677-java-arm64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-java-arm64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-java-arm64
ghcr.io/openhands/agent-server:0bfc677-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:0bfc677-python-amd64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-python-amd64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-python-amd64
ghcr.io/openhands/agent-server:0bfc677-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:0bfc677-python-arm64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-python-arm64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-python-arm64
ghcr.io/openhands/agent-server:0bfc677-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:0bfc677-python-slim-amd64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-python-slim-amd64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-python-slim-amd64
ghcr.io/openhands/agent-server:0bfc677-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-amd64
ghcr.io/openhands/agent-server:0bfc677-python-slim-arm64
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-python-slim-arm64
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-python-slim-arm64
ghcr.io/openhands/agent-server:0bfc677-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-arm64
ghcr.io/openhands/agent-server:0bfc677-golang
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-golang
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-golang
ghcr.io/openhands/agent-server:0bfc677-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:0bfc677-java
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-java
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-java
ghcr.io/openhands/agent-server:0bfc677-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:0bfc677-python-slim
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-python-slim
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-python-slim
ghcr.io/openhands/agent-server:0bfc677-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim
ghcr.io/openhands/agent-server:0bfc677-python
ghcr.io/openhands/agent-server:0bfc67710aa166faa8db3091b687cfb0a23c4d81-python
ghcr.io/openhands/agent-server:vasco-delta-4696-unparent-python
ghcr.io/openhands/agent-server:0bfc677-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `0bfc677-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `0bfc677-python-amd64`) are also available if needed
</details>

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.58s · commit 0bfc677  
**Strongest signal:** Contract regression · 51% estimated likelihood.  
**Evidence:** [F004H002 · openhands-agent-server/openhands/agent\_server/pub\_sub.py:13–18](https://github.com/OpenHands/software-agent-sdk/blob/0bfc67710aa166faa8db3091b687cfb0a23c4d81/openhands-agent-server/openhands/agent_server/pub_sub.py#L13-L18).  
**Coverage:** ⚠️ reduced context — partial coverage; 28/52 hunks, 14/14 files (context budget: 28, hunk budget: 24).

<details>

| --- | --- | --- |
| SQL injection | 3.0% | No direct hunk selected |
| Command injection | 4.0% | No direct hunk selected |
| Weakened authentication | 7.0% | No direct hunk selected |
| Weakened authorization | 8.0% | No direct hunk selected |
| Contract regression | 51.0% | [F004H002 · openhands-agent-server/openhands/agent\_server/pub\_sub.py:13–18](https://github.com/OpenHands/software-agent-sdk/blob/0bfc67710aa166faa8db3091b687cfb0a23c4d81/openhands-agent-server/openhands/agent_server/pub_sub.py#L13-L18) |
| Data loss | 4.0% | No direct hunk selected |
| Sensitive data disclosure | 5.0% | No direct hunk selected |
| Unexpected data transfer | 5.0% | No direct hunk selected |
| Credential misuse | 5.0% | No direct hunk selected |
| Package source redirection | 4.0% | No direct hunk selected |
| Security assessment bypass | 5.0% | No direct hunk selected |
| Primary concern | Contract regression; confidence 73.0% | [F004H002 · openhands-agent-server/openhands/agent\_server/pub\_sub.py:13–18](https://github.com/OpenHands/software-agent-sdk/blob/0bfc67710aa166faa8db3091b687cfb0a23c4d81/openhands-agent-server/openhands/agent_server/pub_sub.py#L13-L18) |

</details>


openhands-agent-server/openhands/agent_server/conversation_service.py
```diff
@@ -2489,6 +2493,32 @@ def _build_telemetry_context(
     )
 
 
+# Deltas arrive at token rate; this only has to beat the runtime-api's
+# ~20 minute idle eviction.
+DELTA_HEARTBEAT_INTERVAL_SECONDS = 30.0
```
VascoSch92: use Final[str] here

openhands-agent-server/openhands/agent_server/event_service.py
```diff
@@ -133,6 +133,12 @@ class EventService:
     _pub_sub: PubSub[Event] = field(
         default_factory=lambda: PubSub[Event](max_subscribers=50), init=False
     )
+    # Uncapped: every delta subscriber is internal or paired with a durable
+    # one, so the limit above already bounds this bus. A separate cap could
```
VascoSch92: trimmer the comment

openhands-agent-server/openhands/agent_server/openapi.py
```diff
@@ -125,6 +125,15 @@ def _add_canonical_contract_components(document: dict[str, Any]) -> None:
                 {**copy.deepcopy(patch_schema), "title": "AgentSettingsPatch"},
             )
 
+    # Websocket-only, so no REST path references it and pruning would drop it.
+    # Generated clients still need the type, so name it explicitly.
+    from openhands.sdk.event import StreamingDeltaEvent
```
VascoSch92: can we not import this at the beginning of the file?

openhands-agent-server/openhands/agent_server/sockets.py
```diff
@@ -539,15 +544,22 @@ def _is_websocket_connected(websocket: WebSocket) -> bool:
...
-    # The live socket is what token streaming is for.
-    receives_streaming_deltas = True
-
     websocket: WebSocket
 
     async def __call__(self, event: Event):
         await _send_event(event, self.websocket)
 
 
+@dataclass
```
VascoSch92: slots=True and frozen=True

openhands-sdk/openhands/sdk/conversation/impl/remote_conversation.py
```diff
@@ -67,6 +69,8 @@
 FATAL_WS_CLOSE_CODES = frozenset({4001, 4004})
 _WEBSOCKET_AUTH_TYPE: Final = "auth"
 _WEBSOCKET_SESSION_API_KEY_FIELD: Final = "session_api_key"
+# Matched before the durable decode: Event.model_validate rejects this kind.
+_STREAMING_DELTA_KIND: Final = StreamingDeltaEvent.__name__
```
VascoSch92, line 74: Final[str] everywhere here

.pr/design.html
```diff
@@ -0,0 +1,324 @@
+<!doctype html>
```
VascoSch92: Change this in a format that we can visualize in github.

VascoSch92:
@all-hands-bot fix the conflicts

VascoSch92:
@enyst gently reminder :-)

enyst (approved):
Sorry for the delay, @VascoSch92 . Just a few thoughts…

OpenHands had this story: everything of relevance that happens in the system, agent events but also system events, are Events, and the UIs receive a stream of Events and show the user everything that happened in the system.

This metaphor broke down sometimes in V0, and I don’t think we ever really reconciled properly and nicely the strict requirements on agent events and maybe significant system errors, with niceties; other things that may happen in the system that shouldn’t be bound to the same requirements.

e.g. there are a bunch of conversation status events that are also fleeting, actually; only the last matters; but we store them all in the same durable ways because they’re Events, and so on.

e.g. it’s been a strangely unclear problem how to present to the user some warnings or info, things that aren’t critical. I think it’s still not really happening nicely.

Thank you for looking deeply into this. It feels like the end of an era to me, but I’m a bit dense like that - I like stories, even system metaphors. But I think you’re right, and we need a better one.

I don’t know if that made sense… I just want to record these thoughts before we move forward.

I think we’ll be happier this way. Things happen in the system, some are fleeting, some are durable, and clients can easily read the envelope and decide what to do. Let’s do it!

enyst (commented):
@VascoSch92 One quick question: the ! in the PR title is, I think, the “breaking change” mark in release-please. Will it force an upgrade to a 2.0.x version?

If we keep backwards compatibility, via the old endpoint, then i think we don’t need 2.0.x., and we need to keep backwards compatibility in some way

VascoSch92:
> @VascoSch92 One quick question: the ! in the PR title is, I think, the “breaking change” mark in release-please. Will it force an upgrade to a 2.0.x version?
> 
> If we keep backwards compatibility, via the old endpoint, then i think we don’t need 2.0.x., and we need to keep backwards compatibility in some way

@enyst 

Good point. Today the ! doesn't change the version, since we cut releases by hand.

This is still backwards compatible where it matters: the wire frame and the old endpoint are unchanged. Only Python-side details change, and that only needs a minor bump.

VascoSch92:
@all-hands-bot fix the conflicts.

enyst:
Then we could either:
- fix the title so it doesn't look scary
- or fix backward compatibility if it breaks for people?

Issue #14, opened 2025-08-24
Port over tests for LLM


xingyaoww:
Specifically port over every test in https://github.com/All-Hands-AI/OpenHands/tree/main/tests/unit/llm, and make sure they follow our current repo's implementation and patterns for tests.

PR #4715, opened 2026-08-30
fix(sdk): mask registered secrets at the tool chokepoint

### LINKED ISSUE #4677: [Security] Secret masking covers 1 tool out of 13 — the model reads raw secrets through file_editor, grep, glob, apply_patch and browser_use
Part of #4671 (step 1), but independent of the streaming work and larger than it.

## Actual Behavior

Sweeping `openhands-tools` for `secret_registry`, `SecretRegistry` and `mask` matches **exactly one file**: `terminal/impl.py:371`. The package ships 13 tool packages:

`apply_patch` · `browser_use` · `delegate` · `file_editor` · `gemini` · `glob` · `grep` · `planning_file_editor` · `task` · `task_tracker` · `terminal` · `tom_consult` · `workflow`

Only `terminal` masks. (`mcp/tool.py:187` in the SDK masks MCP tool output — that's the only other non-ACP call site of `mask_secrets_in_output` anywhere.)

**Repro** — sweep the tools package for any masking call site:

```bash
grep -rln "mask_secrets_in_output\|secret_registry" openhands-tools/openhands/tools/
```

Exactly one file comes back: `terminal/impl.py`. Correspondingly, only one masking test exists:

```bash
uv run pytest tests/tools/terminal/test_secrets_masking.py -q
```

Nothing covers `file_editor`, `grep`, `glob` or `apply_patch`.

## The consequence, concretely

```
cat .env            via bash          ->  MASKED
open .env           via file_editor   ->  RAW
grep -r "API_KEY" . via grep          ->  RAW
```

**The model routinely reads raw secrets into its context**, and which tool it happens to pick decides whether they're masked. There is no security boundary here — there's one tool that happens to have the code.

Everything downstream inherits it: raw secrets in context become raw secrets in model output, which is unmasked on the standard `Agent` path, which is POSTed to webhooks.

## Why this is filed separately

The right long-term fix is keeping secrets out of the model's context entirely, rather than masking the response. But that means bringing a dozen tools under the registry — **a larger project than masking the response, not a smaller one.** It should not block the streaming work, and the streaming work should not be justified as fixing it.

## Fix

Bring reads under the registry at a shared chokepoint rather than tool by tool, so tool #14 is covered by default instead of by remembering. Tool-by-tool patching reproduces exactly the failure mode that got us here.

## Acceptance Criteria

- [ ] A file containing a registered secret returns masked content through `file_editor`, `grep`, `glob` and `apply_patch`.
- [ ] A test that fails when a new tool is added without going through the chokepoint.

alanhuangyoo:
HUMAN:

Ran the new masking tests plus the existing tool suite; checked a registered secret is masked in the observation content and that an unregistered string is left alone.

---

AGENT:

## Why

Sweeping the tools package for masking call sites returns one file:

```
$ grep -rln "mask_secrets_in_output\|secret_registry" openhands-tools/openhands/tools/
openhands-tools/openhands/tools/terminal/impl.py
```

Twelve other tool packages hand their output to the model unmasked, so whether a registered secret is masked depends on which tool the model happened to pick.

`ToolDefinition.__call__` is the one place every tool with an executor returns through, and it already receives `conversation`. Masking there covers the whole tool surface instead of the tools that remembered — the property #4677 asks for, so tool #14 is covered by default rather than by remembering.

## Summary

- Mask registered secrets in `ToolDefinition.__call__`, against `Observation.content` — the base field every tool's text lands in, so the masking is not tool-specific. `acall` delegates to `__call__`, so the async path is covered.
- `terminal/impl.py` keeps its own masking: `TerminalExecutor` is also called directly, without going through `ToolDefinition`, and `test_terminal_executor_with_conversation_secrets` covers exactly that path. Masking twice is idempotent.
- Add `tests/sdk/tool/test_secret_masking_chokepoint.py`, including a structural test that fails when a class starts overriding `ToolDefinition.__call__` without arranging its own masking.

## Issue Number

Closes #4677

## How to Test

End-to-end against a real `Conversation` with a registered secret, no mocks:

```bash
cat > /tmp/repro.py <<'EOF'
import os, tempfile
from openhands.tools.file_editor import FileEditorTool

SECRET = "sk-super-secret-value-123456"
tmp = tempfile.mkdtemp()
env = os.path.join(tmp, ".env")
open(env, "w").write(f"API_KEY={SECRET}\n")

llm = LLM(model="gpt-4o-mini", api_key=SecretStr("k"), usage_id="t")
conv = Conversation(agent=Agent(llm=llm, tools=[]), workspace=tmp,
                    persistence_dir=tmp, secrets={"API_KEY": SECRET})
tool = FileEditorTool.create(conv_state=conv.state)[0]
obs = tool(tool.action_type(command="view", path=env), conv)
text = " ".join(c.text for c in obs.content if hasattr(c, "text"))
print("raw secret present:", SECRET in text)
print("mask marker present:", "<secret-hidden>" in text)
EOF
uv run python /tmp/repro.py
```

On `main`:

```
raw secret present: True
mask marker present: False
```

On this branch:

```
raw secret present: False
mask marker present: True
```

Same harness across the tools named in the acceptance criteria, and a call with `conversation=None` to confirm behaviour is unchanged when there is no registry:

```
file_editor    raw=False  masked_marker=True
grep(content)  raw=False
apply_patch    raw=False
no conversation -> raw retained: True
```

Note on the issue text: `grep` and `glob` are listed there next to `file_editor`, but both return only file paths, never matched line content —

```
Found 1 file(s) containing pattern 'token' in '/tmp/...'
/tmp/.../cfg.txt
```

so they are covered by construction rather than because they were leaking. `file_editor` is the reproducible leak.

Suites:

```bash
uv run pytest tests/sdk/tool/ tests/cross/test_agent_secrets_integration.py \
              tests/tools/terminal/test_secrets_masking.py -q
# 221 passed

# clean
```

The new `test_file_editor_does_not_hand_the_model_a_raw_secret` fails on `main` with

```
assert 'sk-test-sec...abcdef123456' not in "Here's the ...\n     2\t\n"
```

## Video/Screenshots

## Design Doc

Not added; the change is one function and one call site.

### DIFF
--- a/openhands-sdk/openhands/sdk/tool/tool.py
+++ b/openhands-sdk/openhands/sdk/tool/tool.py
@@ -344,6 +344,45 @@ class ExecutableTool(Protocol):
         ...
 
 
+def mask_observation_secrets(
+    observation: Observation, conversation: "LocalConversation | None"
+) -> Observation:
+    """Replace registered secret values in an observation's text content.
+
+    Every tool that declares an executor returns through
+    :meth:`ToolDefinition.__call__`, so masking here covers the whole tool
+    surface rather than the tools that remembered to call the registry. A tool
+    added later is covered without touching it.
+
+    Observations are immutable once built, so a masked one is a copy; an
+    observation with nothing to mask is returned unchanged.
+    """
+    if conversation is None:
+        return observation
+
+    state = getattr(conversation, "state", None)
+    registry = getattr(state, "secret_registry", None)
+    if registry is None:
+        return observation
+
+    masked_content = []
+    changed = False
+    for block in observation.content:
+        text = getattr(block, "text", None)
+        if not text:
+            masked_content.append(block)
+            continue
+        masked_text = registry.mask_secrets_in_output(text)
+        if masked_text != text:
+            block = block.model_copy(update={"text": masked_text})
+            changed = True
+        masked_content.append(block)
+
+    if not changed:
+        return observation
+    return observation.model_copy(update={"content": masked_content})
+
+
 class ToolDefinition[ActionT, ObservationT](DiscriminatedUnionMixin, ABC):
     """Base class for all tool implementations.
 
@@ -621,19 +660,23 @@ class ToolDefinition[ActionT, ObservationT](DiscriminatedUnionMixin, ABC):
         # Coerce output only if we declared a model; else wrap in base Observation
         if self.observation_type:
             if isinstance(result, self.observation_type):
-                return result
-            return self.observation_type.model_validate(result)
+                observation = result
+            else:
+                observation = self.observation_type.model_validate(result)
         else:
             # When no output schema is defined, wrap the result in Observation
             if isinstance(result, Observation):
-                return result
+                observation = result
             elif isinstance(result, BaseModel):
-                return Observation.model_validate(result.model_dump())
+                observation = Observation.model_validate(result.model_dump())
             elif isinstance(result, dict):
-                return Observation.model_validate(result)
-            raise TypeError(
-                "Output must be dict or BaseModel when no output schema is defined"
-            )
+                observation = Observation.model_validate(result)
+            else:
+                raise TypeError(
+                    "Output must be dict or BaseModel when no output schema is defined"
+                )
+
+        return mask_observation_secrets(observation, conversation)
 
     async def acall(
         self, action: ActionT, conversation: "LocalConversation | None" = None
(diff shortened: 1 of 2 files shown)


alanhuangyoo:
Superseded by #4788, which landed on 1 Sep and fixes this at the same place.

Closing rather than rebasing, because the merged version is the better of the two and the difference is not cosmetic. Both hook `ToolDefinition.__call__`, but this PR walks `observation.content` and masks the text blocks, while #4788 added `SecretRegistry.mask_secrets_in_model` and walks the whole model. Its own docstring names the case mine misses:

> Masking one known text field is not enough for tool output: `Observation.to_llm_content` is overridable, and several tools build what the model sees out of their own fields rather than `content`.

That is right, and it means an observation that renders from its own fields would still have leaked under this PR. Nothing here is worth carrying forward on top of it.

Thanks @VascoSch92.

PR #4767, opened 2026-08-30
fix(sdk): match nested repo paths by ancestry, not string prefix

### LINKED ISSUE #4662: [Bug]: Nested Git filtering drops unrelated prefix-sharing paths
## Bug description

`get_git_changes()` discovers child repositories and removes parent-repository changes when `str(change.path).startswith(git_dir)` (`openhands-sdk/openhands/sdk/git/git_changes.py:218-238`).

That is lexical prefix matching, not directory ancestry. If a nested repository is named `foo`, a legitimate root-repository change such as `foobar/file.py` or `foo-config.yaml` is incorrectly removed from the result.

## Expected behavior

Only paths actually contained by a nested repository should be excluded from the parent result; prefix-sharing siblings must remain visible.
```
UPDATED README.md
ADDED   foo-config.yaml
ADDED   foobar/file.py
```

## Actual Behavior

With a nested repository foo in the workspace root, get_git_changes() returns only:
```
python -m get_git_changes

UPDATED README.md
```

## Suggested implementation

Represent child repository locations as relative `Path` values and use path-component containment (`change.path.is_relative_to(git_dir)`). Preserve the existing prefixing of changes returned from the nested repository.

## Acceptance criteria

- [ ] With nested repo `foo`, root changes under `foobar/` and `foo-config.yaml` remain in the result.
- [ ] Changes under `foo/` come from the nested repository exactly once.
- [ ] Multiple child repositories and invalid/broken gitfiles preserve current handling.
- [ ] `GitChange` and `/api/git/changes` schemas remain unchanged.

## Version

alanhuangyoo:
HUMAN:

Checked a nested repo under a parent whose path is a string prefix of it: the child's changes no longer get swallowed. Existing git_changes tests still pass.

---

AGENT:

## Why

`get_git_changes()` decides a change belongs to a nested repository by string prefix:

```python
if str(change.path).startswith(git_dir)
```

That is not directory containment. With a nested repository named `foo`, the root-repository changes `foobar/file.py` and `foo-config.yaml` both match and are dropped, with nothing logged — the agent simply does not see files it changed.

## Summary

- Compare path ancestry with `Path.is_relative_to` instead of `str.startswith`.
- The rewritten filter also reads as what it does; the previous `next(iter(...), None) is None` spelling was doing an `any()`.
- Two tests in `tests/sdk/git/test_git_changes.py`: one for the prefix-sharing paths, one pinning that changes genuinely inside a nested repository are still excluded and reported once.

## Issue Number

Closes #4662

## How to Test

Build a repository with a nested repo named `foo` plus two root changes that share that prefix:

```bash
W=$(mktemp -d); cd "$W"
git init -q . && git config user.email a@b.c && git config user.name t
mkdir -p foo foobar && echo base > README.md
git add -A && git commit -qm init

cd foo && git init -q . && git config user.email a@b.c && git config user.name t
echo nested > nested.txt && git add -A && git commit -qm init && cd ..

echo changed > foobar/file.py
echo changed > foo-config.yaml
echo changed > README.md

uv run python -c "
from openhands.sdk.git.git_changes import get_git_changes
for c in get_git_changes('$W'):
    print(c.status.value, c.path)
"
```

On `main`:

```
UPDATED README.md
```

`foobar/file.py` and `foo-config.yaml` are gone.

On this branch:

```
UPDATED README.md
ADDED   foo-config.yaml
ADDED   foobar/file.py
```

Suites:

```bash
uv run pytest tests/sdk/git/ tests/agent_server/test_git_router.py -q
# 158 passed

# clean
```

`test_get_git_changes_keeps_paths_that_merely_share_a_nested_repo_prefix` fails on `main` and passes here. `test_get_git_changes_still_excludes_paths_inside_a_nested_repo` passes either way — it is there so the exclusion this preserves stays preserved.

## Video/Screenshots

## Design Doc

Not added; the change is one expression.

### DIFF
--- a/openhands-sdk/openhands/sdk/git/git_changes.py
+++ b/openhands-sdk/openhands/sdk/git/git_changes.py
@@ -224,17 +224,15 @@ def get_git_changes(cwd: str | Path, ref: str | None = None) -> list[GitChange]:
     # First try the workspace directory
     changes = get_changes_in_repo(cwd, ref=ref)
 
-    # Filter out any changes which are in one of the git directories
+    # Filter out any changes which are inside one of the nested repositories.
+    # This compares path ancestry rather than string prefixes: a nested
+    # repository named "foo" must not swallow a root-repository change to
+    # "foobar/file.py" or "foo-config.yaml", which merely share its name as a
+    # prefix.
     changes = [
         change
         for change in changes
-        if next(
-            iter(
-                git_dir for git_dir in git_dirs if str(change.path).startswith(git_dir)
-            ),
-            None,
-        )
-        is None
+        if not any(Path(change.path).is_relative_to(git_dir) for git_dir in git_dirs)
     ]
 
     # Add changes from git directories
--- a/tests/sdk/git/test_git_changes.py
+++ b/tests/sdk/git/test_git_changes.py
@@ -718,3 +718,55 @@ def test_get_changes_in_repo_no_remote_worktree_shows_committed_changes():
         assert changes == [
             GitChange(status=GitChangeStatus.UPDATED, path=Path("app.txt"))
         ]
+
+
+def test_get_git_changes_keeps_paths_that_merely_share_a_nested_repo_prefix():
+    """A nested repository named "foo" must not swallow "foobar/" or "foo-config".
+
+    The nested-repository filter compares path ancestry; matching on the string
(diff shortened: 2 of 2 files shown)


enyst (approved):
Thank you!

PR #4768, opened 2026-08-30
fix(tools): keep the base observation in the planning file editor

### LINKED ISSUE #4668: [Bug]: Planning file editor drops inherited observation and diff data
### Bug Description

`PlanningFileEditorExecutor` adapts the canonical file editor result by copying only `command`, `content`, `is_error`, and `path` (`openhands-tools/openhands/tools/planning_file_editor/impl.py:46-68`).

`PlanningFileEditorObservation` inherits `FileEditorObservation`, whose `prev_exist`, `old_content`, and `new_content` fields drive structured edit state and diff visualization (`file_editor/definition.py:69-134`). Those fields are silently discarded, so planning edits cannot produce the complete inherited observation or expected diff view. Future base fields can drift the same way.

### Expected Behavior

Planning results should preserve all canonical file-editor observation data while retaining their distinct serialized planning kind.

### Actual Behavior

it as `/tmp/repro_4668.py`, then run:

```bash
uv run python /tmp/repro_4668.py
```

On SDK commit `bc5dfc50d`, it exercises real temporary files and both real file-editor executors; no mocks and prints:

```text
regular_old_content_present=True
regular_new_content_present=True
regular_has_diff=True
planning_old_content_present=False
planning_new_content_present=False
planning_has_diff=False
plan_edit_applied=True
```

### Suggested Implementation

Pass the inherited planning action directly to the configurable base executor, then construct `PlanningFileEditorObservation` from the complete base observation dump rather than a hand-maintained field list. Keep the planning action/observation classes for event compatibility.

### Acceptance Criteria

- [ ] Create and edit operations preserve `prev_exist`, `old_content`, and `new_content`.
- [ ] Planning edit visualization contains the expected diff.
- [ ] Edits outside the configured plan remain rejected and arbitrary views remain allowed.
- [ ] Planning action/observation kinds and public schemas remain unchanged.

### Version

alanhuangyoo:
HUMAN:

Ran the planning file editor through an edit and confirmed the observation keeps the base fields instead of losing them to the discriminator.

---

AGENT:

## Why

`PlanningFileEditorExecutor` rebuilt its result from four hand-picked fields:

```python
return PlanningFileEditorObservation(
    command=action.command,
    content=file_editor_obs.content,
    is_error=file_editor_obs.is_error,
    path=file_editor_obs.path,
)
```

`PlanningFileEditorObservation` inherits `FileEditorObservation`, whose `prev_exist`, `old_content` and `new_content` carry the structured edit state and drive the diff view. They were dropped, so a planning edit could not produce a diff — and any field added to the base later would go the same way.

## Summary

- Pass the action straight to the base executor. `PlanningFileEditorAction` **is** a `FileEditorAction`, so the field-by-field rebuild was only a way for the two definitions to drift apart.
- Build the observation from the full base dump instead of a hand-maintained list, excluding `kind` so the subclass re-derives its own discriminator.
- Two tests: one for the preserved fields, one pinning that the observation is still a `PlanningFileEditorObservation` with `kind == "PlanningFileEditorObservation"`.

## Issue Number

Closes #4668

## How to Test

```bash
cat > /tmp/probe.py <<'EOF'
import tempfile, os
from openhands.tools.planning_file_editor.definition import PlanningFileEditorAction
from openhands.tools.planning_file_editor.impl import PlanningFileEditorExecutor

tmp = tempfile.mkdtemp()
plan = os.path.join(tmp, "PLAN.md")
open(plan, "w").write("# Plan\nold line\n")

ex = PlanningFileEditorExecutor(workspace_root=tmp, plan_path=plan)
obs = ex(PlanningFileEditorAction(command="str_replace", path=plan,
                                  old_str="old line", new_str="new line"))
print("kind        :", obs.kind)
print("prev_exist  :", obs.prev_exist)
print("old_content :", repr(obs.old_content))
print("new_content :", repr(obs.new_content))
EOF
uv run python /tmp/probe.py
```

On `main`:

```
kind        : PlanningFileEditorObservation
prev_exist  : True
old_content : None
new_content : None
```

On this branch:

```
kind        : PlanningFileEditorObservation
prev_exist  : True
old_content : '# Plan\nold line\n'
new_content : '# Plan\nnew line\n'
```

The `kind` line is the check that carrying the base fields across does not turn the observation into the base type — the discriminator is computed, so `model_dump()` reports `FileEditorObservation` and it has to be excluded when reconstructing.

Suites:

```bash
uv run pytest tests/tools/planning_file_editor/ tests/tools/test_planning_preset.py tests/tools/file_editor/ -q
# 161 passed

# clean
```

`test_planning_edit_preserves_inherited_observation_fields` fails on `main` with `assert None == '# Plan\nold line\n'`.

## Video/Screenshots

## Design Doc

Not added; the change is one call site.


enyst (commented):



The implementation is clean and the focused planning-editor suite passes locally (`7 passed`). Passing the inherited action through and rebuilding from the complete observation removes the field-drift failure while preserving the planning discriminator.

One test improvement is worth making before approval: the reported user-visible regression is the missing diff, but the new test stops at asserting `old_content`/`new_content`. Please exercise `observation.visualize` (or the public serialized/rendered output consumed downstream) and assert the edit diff is actually present. That verifies the acceptance criterion at the observable boundary and would catch future changes in the visualization path even if the fields remain populated.

This changes the observation returned to an agent-facing tool, so I am leaving a COMMENT pending lightweight eval evidence and explicit human-maintainer confirmation. The stale description-validator failure has been rerun after #4668 received `ready-for-dev` and now passes; the other CI checks on this head are green.

alanhuangyoo:
Done in `2186386` — the field assertions now have the rendered boundary behind them.

`FileEditorObservation.visualize` falls back to the plain agent observation unless `_has_meaningful_diff` holds, and that needs both `old_content` and `new_content` populated — exactly what the hand-maintained field list dropped. So the missing diff was not a second bug downstream, it was the same one seen from the outside, and the test now says so:

```python
rendered = observation.visualize.plain
assert "old line" in rendered and "new line" in rendered, rendered
assert plan_path in rendered, rendered
```

`tests/tools/planning_file_editor/`: 7 passed. The case still fails on `main`, first on the field assertion (`assert None == '# Plan\nold line\n'`).

One thing from the full run, checked rather than assumed: `tests/tools/` gives 952 passed / 1 failed here, and the failure is `test_file_editor_memory_leak`. It is not from this PR — on `main` the same whole-directory run fails it too (plus `test_multiline_commands[subprocess]`), while in isolation it passes 5/5 on both branches. Order- or pressure-dependent, pre-existing.

Merged current `main` (was 41 behind); pre-commit clean.

PR #4770, opened 2026-08-30
fix(workspace): repoint the client after a cloud sandbox resume

### LINKED ISSUE #4664: [Bug]: OpenHandsCloudWorkspace resume leaves the active client on stale connection data
### Bug Description

Cloud readiness polling refreshes `_session_api_key` and exposed URLs (`openhands-workspace/openhands/workspace/cloud/workspace.py:361-374`). Initial startup copies those values into inherited `host`/`api_key` and resets the remote client (`:242-300`). Public `resume()` polls readiness again (`:444-458`) but does not reapply the refreshed URL/key or reset the cached client.

After a sandbox returns with a rotated key or changed exposed URL, settings/skills calls can use the refreshed private key while normal workspace/conversation calls retain stale inherited headers and endpoint. The object can describe two different connections at once.

### Expected Behavior

A RUNNING sandbox should have one atomic agent-server connection: endpoint, session key, inherited API key, and cached client must agree after startup and resume.

### Actual Behavior

it as `/tmp/repro_4664.py`, then run:

```bash
uv run python /tmp/repro_4664.py
```

On SDK commit `bc5dfc50d`, it exercises the real resume, readiness, and HTTP-client paths; only the external Cloud API and health-check boundaries are represented by a fixture and prints:

```text
discovered_url=https://new-agent.example.com
active_host=https://old-agent.example.com
discovered_session_key=new-session-key
active_api_key=old-session-key
client_reused=True
client_base_url=https://old-agent.example.com
client_session_key=old-session-key
```

### Suggested Implementation

Extract one private “apply ready connection” helper or immutable connection snapshot and use it after both initial readiness and resume. Apply `host` and `api_key` together, then reset/rebuild the client. Prefer the inherited API key as the single credential source once connected.

### Acceptance Criteria

- [ ] Resume with a changed exposed URL routes subsequent calls to the new endpoint.
- [ ] Resume with a rotated session key sends the new `X-Session-API-Key` for command, settings, and skills calls.
- [ ] Initial provisioned/local modes and missing-key behavior remain unchanged.
- [ ] Cleanup clears one coherent connection state.
- [ ] Public workspace constructor and remote API contracts remain unchanged.

### Version

alanhuangyoo:
HUMAN:

Exercised a cloud sandbox resume and confirmed the client is repointed at the new agent_server_url, matching what the startup path already did.

---

AGENT:

## Why

`_wait_until_sandbox_ready()` re-reads `session_api_key` and `exposed_urls` from the sandbox, but reading them is not enough — they only take effect once they are pushed onto the inherited `host`/`api_key` and the cached client is dropped.

Startup does that:

```python
self._wait_until_sandbox_ready()
agent_server_url = self._get_agent_server_url()
self.host = agent_server_url.rstrip("/")
self.api_key = self._session_api_key
self.reset_client()
```

`resume()` polled readiness and stopped there. A sandbox that comes back on a different exposed URL or with a rotated key leaves the workspace describing two connections at once: settings and skills calls go through the refreshed private key, while workspace and conversation calls keep the stale headers and endpoint.

## Summary

- Move the three steps into `_apply_ready_connection()` and call it from both startup and `resume()`, so the two paths cannot drift apart again.
- Add a regression test that resumes onto a new URL with a rotated key and asserts `host`, `api_key` and the client reset.

## Issue Number

Closes #4664

## How to Test

```bash
uv run pytest tests/workspace/ -q
# 192 passed
```

`test_cloud_workspace_resume_repoints_client_at_refreshed_connection` resumes a sandbox whose readiness poll returns `https://agent-server-2.example.com/` and `rotated-key`. On `main`:

```
assert workspace.host == "https://agent-server-2.example.com"
AssertionError: assert 'https://agent-server.example.com' == 'https://agent-server-2.example.com'
```

On this branch it passes, and `reset_client` is called once.

```bash
# clean
```

## Note on the existing test

`test_cloud_workspace_resume_calls_resume_sandbox` patched `_wait_until_sandbox_ready` to a no-op, which leaves `_exposed_urls` empty — a state a real poll never produces, since the real method sets both fields before returning. With `resume()` now doing what startup does, that mock made the call raise `ValueError: Agent server URL not found`.

I made the mock populate the fields the way the real method does rather than relaxing `resume()`, because tolerating a missing agent-server URL there would mean silently keeping the stale client, which is the bug being fixed.

## Video/Screenshots

Not applicable — no GUI surface; the test output above is the evidence.

## Design Doc

Not added; the change extracts one block and calls it from a second place.


enyst (commented):



The production refactor is small and correctly reuses one connection-application path for startup and resume. The focused workspace suite passes locally (`14 passed`).

Please strengthen the regression at the behavior boundary before approval. `test_cloud_workspace_resume_repoints_client_at_refreshed_connection` mocks `reset_client()`, so it proves only that the helper was called—not that a previously cached client is closed/replaced or that subsequent requests use the new base URL and rotated `X-Session-API-Key`. Prime `workspace.client` with the old connection, keep only the external Cloud readiness/resume calls stubbed, invoke `resume()`, then assert the old client is closed, the newly materialized client is distinct, and its base URL/header use the refreshed values. That directly verifies the failure described in #4664 with less mocking.

The stale description-validator failure has been rerun after #4664 received `ready-for-dev` and now passes; the other CI checks on this head are green.

alanhuangyoo:
Done in `0b7730a` — you were right that the old test proved only that the helper was called.

It now primes `workspace.client` with the previous sandbox's connection, leaves `reset_client` real, and stubs only the external Cloud resume/readiness calls. After `resume()` it asserts:

- `stale_client.is_closed` — the client for the previous sandbox is actually closed
- `refreshed_client is not stale_client`
- `refreshed_client.base_url` is `https://agent-server-2.example.com`
- `refreshed_client.headers["X-Session-API-Key"]` is `rotated-key`

so it verifies where later requests go rather than that a method fired.

On `main` it fails with

```
assert 'https://agent-server.example.com' == 'https://agent-server-2.example.com'
```

One thing I had to add and want to flag rather than hide: the fixture leaves the sandbox key on `_session_api_key` and never pushes it to `api_key`, which a live workspace only gets from `_apply_ready_connection`. Without priming it the stale client has no `X-Session-API-Key` header at all, so the "before" half of the assertion could not be made. The test sets it explicitly with a comment saying why, rather than reaching further into the fixture.

`tests/workspace/`: 192 passed. Merged current `main` (was 41 behind); pre-commit clean.

neubig (changes requested):
The implementation and regression test look sound, but the current evidence does not meet the bar for a cloud-runtime fix. The independent reproducer exercises real workspace/resume/client code while replacing the external Cloud API and health-check boundaries with a fixture. The HUMAN note says an actual cloud sandbox resume was exercised, but provides no reproducible transcript or observed runtime state.

Please add sanitized evidence from an actual Cloud sandbox pause/resume showing that the resumed workspace and its materialized HTTP client use the refreshed connection, followed by a successful real workspace or conversation request. Useful evidence would include the exact command or script, pre/post host or endpoint identifiers, confirmation that the client was replaced, and the successful post-resume operation. Do not include session keys or other credentials.

Once that actual-product evidence is present, the small shared `_apply_ready_connection()` implementation and behavior-level regression coverage otherwise look appropriate.

_This review was created by an AI agent (OpenHands) on behalf of Gregory Neubig._

PR #4773, opened 2026-08-30
fix(sdk): treat a non-string hook decision as no decision

### LINKED ISSUE #4766: [Bug]: HookExecutor converts exit-0 hooks with decision: null into execution failures

### Bug Description

`HookExecutor.execute` converts a successfully completed command hook into a synthetic execution failure when its JSON stdout contains `"decision": null`.

For example, a hook that prints:

```json
{"decision": null}
```

and exits with status `0` returns:

```python
HookResult(
    success=False,
    exit_code=-1,
    error=(
        "Hook execution failed: "
        "'NoneType' object has no attribute 'lower'"
    ),
)
```

The subprocess itself succeeds. During structured-output parsing, JSON `null` becomes Python `None`, and the implementation calls `.lower()` on it.

The resulting `AttributeError` reaches the outer execution-error handler, which discards the already captured subprocess result and reports a synthetic exit code of `-1`.

### Expected Behavior

The documented command-hook contract states that exit code `0` represents successful execution.

A null `decision` does not represent either `"allow"` or `"deny"`, so it should be treated as absent or unsupported without causing an internal type error.

The result should preserve the observed subprocess outcome:

```python
result.success is True
result.exit_code == 0
result.decision is None
result.error is None
```

The original stdout should also remain available for diagnostics.

Valid string decisions such as `"allow"` and `"deny"` should continue to behave as they do currently.

### Actual Behavior

Running the reproduction with:

```bash
uv run pytest \
    tests/sdk/hooks/test_executor_null_decision.py \
    -q
```

fails because the returned result contains:

```text
success=False
exit_code=-1
stdout=''
decision=None
error="Hook execution failed: 'NoneType' object has no attribute 'lower'"
```

The real subprocess exit status of `0` and its captured stdout are replaced by the outer exception handler.

### Steps to Reproduce

1. Check out release `v1.44.1` at commit `9d143aac35c2dcec9cbb046ff9f35ac5eb072f6a`.

2. Set up the development environment:

```bash
make build
```

3. Create `tests/sdk/hooks/test_executor_null_decision.py`:

```python
from openhands.sdk.hooks.config import HookDefinition
from openhands.sdk.hooks.executor import HookExecutor
from openhands.sdk.hooks.types import HookEvent, HookEventType
from tests.command_utils import python_command


def test_execute_null_decision_preserves_exit_zero(tmp_path):
    executor = HookExecutor(working_dir=str(tmp_path))
    event = HookEvent(
        event_type=HookEventType.PRE_TOOL_USE,
        tool_name="BashTool",
        tool_input={"command": "ls -la"},
        session_id="test-session",
    )
    hook = HookDefinition(
        command=python_command(
            "import json; "
            "print(json.dumps({'decision': None}))"
        ),
    )

    result = executor.execute(hook, event)

    assert result.success is True
    assert result.exit_code == 0
    assert result.stdout.strip() == '{"decision": null}'
    assert result.decision is None
    assert result.error is None
```

4. Run:

```bash
uv run pytest \
    tests/sdk/hooks/test_executor_null_decision.py \
    -q
```

5. Observe that the test fails at `result.success is True`. The returned result instead reports `success=False` and `exit_code=-1`.

### Acceptance Criteria

- [ ] An exit-0 command hook with `"decision": null` does not trigger an internal `AttributeError`.
- [ ] The returned result preserves `success=True` and the actual exit code `0`.
- [ ] A null or otherwise unsupported decision value does not become an `ALLOW` or `DENY` decision.
- [ ] Captured stdout and stderr are not discarded because of decision-field parsing.
- [ ] Existing `"allow"` and `"deny"` behavior remains unchanged.
- [ ] Regression tests cover missing, null, unsupported-string, `"allow"`, and `"deny"` decision values.

### Installation Method

Source checkout using make build (uv sync --dev)

Not applicable

### SDK Version

1.44.1, main@9d143aac35c2dcec9cbb046ff9f35ac5eb072f6a

### Version Confirmation

### Python Version

3.13.2

Not applicable; reproduced by directly invoking deterministic command-hook execution code.

### Operating System

MacOS

### Logs and Error Messages

```text
Input stdout:
{"decision": null}

Observed result:
success=False
blocked=False
exit_code=-1
stdout=''
stderr=''
decision=None
error="Hook execution failed: 'NoneType' object has no attribute 'lower'"
```

A string control value does not produce the execution error:

```text
Input stdout:
{"decision": "unsupported"}

Observed result:
success=True
blocked=False
exit_code=0
stdout='{"decision": "unsupported"}\n'
decision=None
error=None
```

Representative assertion:

```text
AssertionError: assert False is True
 +  where False = HookResult(
     success=False,
     blocked=False,
     exit_code=-1,
     stdout='',
     stderr='',
     decision=None,
     reason=None,
     additional_context=None,
     error="Hook execution failed: 'NoneType' object has no attribute 'lower'",
     async_started=False,
   ).success
```

### Minimal Code Sample

```python

from openhands.sdk.hooks.config import HookDefinition
from openhands.sdk.hooks.executor import HookExecutor
from openhands.sdk.hooks.types import HookEvent, HookEventType
from tests.command_utils import python_command


with TemporaryDirectory() as working_dir:
    executor = HookExecutor(working_dir=working_dir)
    event = HookEvent(
        event_type=HookEventType.PRE_TOOL_USE,
        tool_name="BashTool",
        tool_input={"command": "ls"},
        session_id="demo",
    )
    hook = HookDefinition(
        command=python_command(
            "import json; "
            "print(json.dumps({'decision': None}))"
        ),
    )

    result = executor.execute(hook, event)
    print(result)
    assert result.success is True
```

No screenshot is required; this is a deterministic unit-level reproduction using a local subprocess.

The executor initially constructs the correct result from the completed process:

```python
hook_result = HookResult(
    success=result.returncode == 0,
    blocked=result.returncode == 2,
    exit_code=result.returncode,
    stdout=result.stdout,
    stderr=result.stderr,
)
```

It then parses the optional `decision` field without checking its type:

```python
if "decision" in output_data:
    decision_str = output_data["decision"].lower()
```

The inner handler catches only malformed JSON:

```python
except json.JSONDecodeError:
    pass
```

Consequently, the `AttributeError` raised by `None.lower()` reaches the outer handler:

```python
except Exception as e:
    return HookResult(
        success=False,
        exit_code=-1,
        error=f"Hook execution failed: {e}",
    )
```

This replacement loses the subprocess's real exit code, stdout, and stderr even though execution completed normally.

A possible fix is to validate the decision value before applying string methods:

```python
decision = output_data.get("decision")

if isinstance(decision, str):
    decision_str = decision.lower()
    if decision_str == "allow":
        hook_result.decision = HookDecision.ALLOW
    elif decision_str == "deny":
        hook_result.decision = HookDecision.DENY
        hook_result.blocked = True
```

Null and other unsupported decision values can then remain unclassified without converting successful command execution into an unrelated internal failure.

alanhuangyoo:
HUMAN:

Ran a hook printing {"decision": null} and one printing a number; both now come back as no decision with the real exit code instead of a synthetic -1.

---

AGENT:

## Why

A command hook that exits 0 and prints `{"decision": null}` came back as a failure:

```python
HookResult(success=False, exit_code=-1,
           error="Hook execution failed: 'NoneType' object has no attribute 'lower'")
```

JSON `null` parses to `None` and `.lower()` was called on it. Only `JSONDecodeError` is caught around that block, so the `AttributeError` reached the outer handler, which discarded the captured subprocess result and reported a synthetic exit `-1` for a process that had exited 0.

It is not specific to `null` — any non-string fails the same way:

```
decision null     success=False  exit=-1   error=Hook execution failed: 'NoneType' object has no attribute 'lower'
decision allow    success=True   exit=0    decision=HookDecision.ALLOW
decision number   success=False  exit=-1   error=Hook execution failed: 'int' object has no attribute 'lower'
```

## Summary

- Check for a string before lowering, so any other shape means the hook expressed no decision rather than a failed run. The agent-hook path at `executor.py:416` already goes through `str()` for the same reason; this is the sibling that did not.
- Three tests: null, a number, and that string decisions still parse.

I left the `except json.JSONDecodeError` as it is. Widening it would hide the next bug of this kind rather than fix it; the point here is not to raise.

## Issue Number

Closes #4766

## How to Test

```bash
uv run python -c "
import tempfile
from openhands.sdk.hooks.config import HookDefinition
from openhands.sdk.hooks.executor import HookExecutor
from openhands.sdk.hooks.types import HookEvent, HookEventType

ex = HookExecutor(working_dir=tempfile.mkdtemp())
ev = HookEvent(event_type=HookEventType.PRE_TOOL_USE, tool_name='BashTool',
               tool_input={}, session_id='s')
for payload in ('{\"decision\": null}', '{\"decision\": \"allow\"}', '{\"decision\": 1}'):
    r = ex.execute(HookDefinition(command=f\"echo '{payload}'\"), ev)
    print(payload, '->', r.success, r.exit_code, r.decision, r.error or '')
"
```

On this branch all three come back `success=True exit=0`, with `HookDecision.ALLOW` only for the string.

```bash
uv run pytest tests/sdk/hooks/ -q
# 146 passed

# clean
```

Two of the three new tests fail on `main`:

```
assert result.success
E   assert False
E    +  where False = HookResult(success=False, exit_code=-1, ...,
        error="Hook execution failed: 'int' object has no attribute 'lower'").success
```

## Video/Screenshots

Not applicable — no GUI surface; the console output above is the evidence.

## Design Doc

Not added; the change is one type check.


neubig (approved):
_This review was submitted by an AI agent (OpenHands) on behalf of @neubig._

The issue is valid and the implementation is the narrowest correct fix: command-hook `decision` is optional, non-string values now mean no decision, and the actual subprocess success, exit code, stdout, and stderr are preserved. Existing `allow`/`deny` behavior remains unchanged.

I also simplified the redundant historical commentary in commit `9aea74cd`; no behavior or assertions changed. Focused verification on the updated head passed: 58 hook-executor tests, Ruff lint, Ruff formatting, and `git diff --check`.

Approved for merge.

alanhuangyoo:
Thanks @neubig. The red pre-commit isn't the change: the baseline check runs `git show <base.sha>:scripts/forbidden_dynamic_attributes_baseline.json`, and this PR's recorded base is still 6a1e4d0 (v1.46.0), from before that file existed. Merging main into the branch doesn't move `base.sha`, so any PR opened before the baseline landed will hit the same thing. Retargeting the base should refresh it, or the check could skip when the file is missing at the ref.

PR #4783, opened 2026-08-31
fix(sdk): mask model output in the durable MessageEvent

### LINKED ISSUE #4678: [Security] Model output is never masked on the standard Agent path
Part of #4671 (step 1).

Related: OpenHands/OpenHands#15720 (ACP-scoped).

## Actual Behavior

`SecretRegistry.mask_secrets_in_output` (`secret_registry.py:143`) has, outside `acp_agent.py`, exactly **two** non-test call sites:

- `openhands-tools/openhands/tools/terminal/impl.py:371` — bash output
- `openhands-sdk/openhands/sdk/mcp/tool.py:187` — MCP tool output

**Both are tool output. Nothing masks the model's own response text on the standard `Agent` path** — not the deltas, and not the durable `MessageEvent`.

**Repro**

```bash
uv run pytest tests/cross/test_agent_secrets_integration.py -q
```

The existing coverage asserts masking of *tool* output only. Nothing asserts that a registered secret appearing in the model's own response text is masked on the standard `Agent` path.

## Why this is separate from #15720

#15720 is scoped `[ACP on Cloud]` and reports split-chunk secrets in the live ACP relay. On ACP the situation is actually the *better* one: `_handle_update` masks each chunk (`acp_agent.py:1302`), and `_finalize_successful_turn` re-masks the joined text at the persistence boundary — so **on ACP the durable record is safe and the leak is delta-only.**

Off ACP there is no masking at all, at either end. The durable record is affected, which means it's the copy that gets stored, rendered, exported and POSTed to webhooks. That's a strictly larger problem than the one #15720 describes, and it isn't covered by that issue's title or scope.

## Fix

One masking boundary both paths cross. On the stream path, use a snapshot compiled once at stream open — resolved values only, single-pass matcher, no lock and no I/O while streaming. This matters because `mask_secrets_in_output` resolves secret sources first, and its own comment warns `get_value()` *"may do blocking network I/O"* — on the sync path that runs under the state lock.

Trade-off, stated plainly: **a secret registered *after* the stream opened won't be masked in that stream's deltas, but it will be masked in the durable message**, which is the authoritative copy.

## Acceptance Criteria

- [ ] A secret split across two chunks is masked in the durable message on both the `Agent` and ACP paths.
- [ ] Model response text containing a registered secret is masked on the standard `Agent` path.

VascoSch92:
HUMAN:

Fix leakage of secrets.

---

AGENT:

## Why

`SecretRegistry.mask_secrets_in_output` had, outside `acp_agent.py`, exactly two non-test call sites — `openhands-tools/openhands/tools/terminal/impl.py:371` (bash output) and `openhands-sdk/openhands/sdk/mcp/tool.py:187` (MCP tool output). Both are *tool* output. Nothing masked the model's own response text on the standard `Agent` path.

That copy is the durable one. A secret the model echoes back was persisted verbatim into the `MessageEvent` that is stored, replayed to the LLM on the next turn, rendered, exported, and POSTed to webhooks.

This is the off-ACP half of the problem. On ACP the situation was already the better one: `_handle_update` masks each chunk and `_finalize_successful_turn` re-masks the joined text at the persistence boundary, so the durable ACP record was already safe (covered today by `TestACPStepMasksPersistedTurn::test_finish_action_masks_secret_split_across_chunks`). Off ACP there was no masking at either end.

## Summary

- Mask registered secret values in the agent's own `MessageEvent` at `_emit_message_event` (`openhands-sdk/openhands/sdk/agent/response_dispatch.py`) — the single point where both `_handle_content_response` and `_handle_no_content_response` build that event, shared by the sync (`step`) and async (`astep`) paths.
- Mask `reasoning_content` alongside `content`: it is the only text a reasoning-only response carries, so masking `content` alone would be a no-op for exactly one of the two callers.
- Masking is unconditional: it reads the registry's tracked values, which are not the same thing as its `secret_sources` list (see Notes).

## Issue Number

Fixes #4678

## How to Test

New regression tests in `tests/sdk/agent/test_response_dispatch.py`, driven through the existing `_run_single_step` harness (a real `agent.step()` with a canned `LLMResponse`, not a direct call to the masking helper):

```
uv run pytest tests/sdk/agent/test_response_dispatch.py -q
# 26 passed
```

To confirm they actually catch the bug, revert only the source file and re-run — all three masking tests fail against the previous behaviour:

```
git stash push openhands-sdk/openhands/sdk/agent/response_dispatch.py
uv run pytest tests/sdk/agent/test_response_dispatch.py -q -k mask
```
```
FAILED test_content_response_masks_registered_secret
  - AssertionError: assert 'supersecret' not in 'the value i...ersecret now'
FAILED test_content_response_masks_secret_joined_from_stream_chunks
  - AssertionError: assert 'the value is supersecret now' == 'the value is...t-...
FAILED test_reasoning_only_response_masks_registered_secret
  - AssertionError: assert 'the value is supersecret now' == 'the value is...t-...
3 failed, 22 deselected
```

Wider suites, to show nothing else moved:

```
uv run pytest tests/sdk tests/cross -q -p no:randomly
# 2 failed, 6381 passed, 10 skipped, 10 xfailed
```

The two failures are `tests/cross/test_remote_conversation_live_server.py::test_openai_chat_completions_gateway_over_real_server` and `::test_settings_and_secrets_api_with_live_server`. Both are **pre-existing on `origin/main`** — verified by stashing this branch's changes and re-running the same file on a clean tree, which produces the identical two failures.

`pre-commit run --files <both changed files>` passes (ruff format, ruff lint, pycodestyle, pyright, import rules, tool registration).

## Video/Screenshots

Not applicable — no user-facing surface. The observable behaviour is the assertion output above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**Acceptance criteria.** Both are met. "A secret split across two chunks is masked in the durable message on both the `Agent` and ACP paths" — the ACP half already held and is covered by the existing test named above; the `Agent` half is `test_content_response_masks_secret_joined_from_stream_chunks`. "Model response text containing a registered secret is masked on the standard `Agent` path" — `test_content_response_masks_registered_secret`.

**Why no stream snapshot.** The issue sketches a snapshot compiled at stream open. That is not needed for these acceptance criteria and is deliberately left out of this PR: deltas are accumulated *before* the `Message` is constructed, so the durable copy is built from the joined text and the join-boundary mask catches a secret that straddles two chunks. Masking the deltas themselves belongs with the streaming work in #4671.

**Self-review catch (fixed in 37341bdf1).** The first push guarded the mask with `if not registry.secret_sources: return message`. That was wrong. Masking reads `_exported_values`, a *separate* field from `secret_sources`, and `EventService.activate_credential_binding` pops a name out of `secret_sources` via `model_copy` while the resolved value stays tracked for masking. In that window the guard skipped masking a live credential that `mask_secrets_in_output` would still have masked — a leak in exactly the credential-binding flow. Reproduced directly against the registry:

```
before: sources=['TOKEN'] exported=['TOKEN']  mask -> "the value is <secret-hidden> now"
after : sources=[]        exported=['TOKEN']  mask -> "the value is <secret-hidden> now"
                                              guard would skip masking: True
```

The guard is gone; masking is unconditional, which is also cheaper to reason about (an empty `_exported_values` makes it a no-op). Regression test: `test_content_response_masks_value_dropped_from_secret_sources`, verified to fail with the guard restored.

**Deliberately not masked.** `thinking_blocks` and `responses_reasoning_item` are signed provider payloads; rewriting their text invalidates the signature replayed on the next request.

**Known gap, not in scope here.** `ActionEvent.thought` on the tool-call path (`agent.py:1318`) is still unmasked — `_handle_tool_calls` does not route through `_emit_message_event`. Happy to fold it in if you would rather it landed together; I kept this PR to the one chokepoint the acceptance criteria name.

**Lock cost.** `agent.step()` runs under the state lock, and `mask_secrets_in_output` resolves uncached sources, which may do blocking I/O. Exposure is unchanged in practice: it is one call per agent step, it already has failed-lookup backoff, values are normally resolved and cached well before the model replies, and with no secrets registered it degrades to an empty loop. `terminal/impl.py` already calls it from inside a step.

**Not verified locally.** `.github/scripts/check_sdk_api_breakage.py` could not run here (griffe fails to fetch the baseline from PyPI in this environment). The change adds no public API — one private static method, no changed signatures.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:37341bd-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-37341bd-python \
  ghcr.io/openhands/agent-server:37341bd-python
```

```
ghcr.io/openhands/agent-server:37341bd-golang-amd64
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-golang-amd64
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-golang-amd64
ghcr.io/openhands/agent-server:37341bd-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:37341bd-golang-arm64
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-golang-arm64
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-golang-arm64
ghcr.io/openhands/agent-server:37341bd-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:37341bd-java-amd64
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-java-amd64
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-java-amd64
ghcr.io/openhands/agent-server:37341bd-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:37341bd-java-arm64
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-java-arm64
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-java-arm64
ghcr.io/openhands/agent-server:37341bd-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:37341bd-python-amd64
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-python-amd64
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-python-amd64
ghcr.io/openhands/agent-server:37341bd-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:37341bd-python-arm64
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-python-arm64
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-python-arm64
ghcr.io/openhands/agent-server:37341bd-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:37341bd-golang
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-golang
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-golang
ghcr.io/openhands/agent-server:37341bd-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:37341bd-java
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-java
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-java
ghcr.io/openhands/agent-server:37341bd-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:37341bd-python
ghcr.io/openhands/agent-server:37341bdf1003a42128577ccf8ae8e0087c3b9a3c-python
ghcr.io/openhands/agent-server:vasco-security-4678-mask-agent-message-output-python
ghcr.io/openhands/agent-server:37341bd-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `37341bd-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `37341bd-python-amd64`) are also available if needed
</details>

### DIFF
--- a/openhands-sdk/openhands/sdk/agent/response_dispatch.py
+++ b/openhands-sdk/openhands/sdk/agent/response_dispatch.py
@@ -297,27 +297,12 @@ class ResponseDispatchMixin:
     def _mask_secrets(message: Message, conversation: LocalConversation) -> Message:
         """Return ``message`` with registered secret values masked in its text.
 
-        Tool output is masked by its executor, but nothing masked the model's
-        own words: a secret the model echoed back was persisted verbatim into
-        the durable ``MessageEvent`` — the copy that is stored, replayed to the
-        LLM, rendered, exported and POSTed to webhooks.
-
-        Masking here rather than per response type covers every non-tool-call
-        response, because both ``_handle_content_response`` and
-        ``_handle_no_content_response`` build their event through
-        ``_emit_message_event``, on the sync and async paths alike.
-        ``reasoning_content`` is masked too — it is the only text a
-        reasoning-only response carries, so leaving it out would make this a
-        no-op for exactly one of the two callers.
-
-        ``thinking_blocks`` and ``responses_reasoning_item`` are deliberately
-        left untouched: they are signed provider payloads, and rewriting them
-        invalidates the signature the next request replays.
+        ``thinking_blocks`` and ``responses_reasoning_item`` are left alone:
+        they are signed provider payloads, and rewriting them invalidates the
+        signature replayed on the next request.
         """
-        registry = conversation.state.secret_registry
-        if not registry.secret_sources:
-            return message
-        mask = registry.mask_secrets_in_output
+        mask = conversation.state.secret_registry.mask_secrets_in_output
+        reasoning = message.reasoning_content
         return message.model_copy(
             update={
                 "content": [
@@ -326,11 +311,7 @@ class ResponseDispatchMixin:
                     else part
                     for part in message.content
                 ],
-                "reasoning_content": (
-                    mask(message.reasoning_content)
-                    if message.reasoning_content
-                    else message.reasoning_content
-                ),
+                "reasoning_content": mask(reasoning) if reasoning else reasoning,
             }
         )
 
--- a/openhands-sdk/openhands/sdk/agent/response_dispatch.py
+++ b/openhands-sdk/openhands/sdk/agent/response_dispatch.py
@@ -281,7 +281,7 @@ class ResponseDispatchMixin:
         """Create and emit a MessageEvent, running critic if configured."""
         msg_event = MessageEvent(
             source="agent",
-            llm_message=message,
+            llm_message=self._mask_secrets(message, conversation),
             llm_response_id=llm_response.id,
         )
         if self.critic is not None and self.critic.mode == "finish_and_message":
@@ -292,6 +292,47 @@ class ResponseDispatchMixin:
                 )
         on_event(msg_event)
         return msg_event
+
+    @staticmethod
+    def _mask_secrets(message: Message, conversation: LocalConversation) -> Message:
+        """Return ``message`` with registered secret values masked in its text.
+
+        Tool output is masked by its executor, but nothing masked the model's
+        own words: a secret the model echoed back was persisted verbatim into
+        the durable ``MessageEvent`` — the copy that is stored, replayed to the
+        LLM, rendered, exported and POSTed to webhooks.
+
+        Masking here rather than per response type covers every non-tool-call
+        response, because both ``_handle_content_response`` and
+        ``_handle_no_content_response`` build their event through
+        ``_emit_message_event``, on the sync and async paths alike.
+        ``reasoning_content`` is masked too — it is the only text a
+        reasoning-only response carries, so leaving it out would make this a
+        no-op for exactly one of the two callers.
+
+        ``thinking_blocks`` and ``responses_reasoning_item`` are deliberately
+        left untouched: they are signed provider payloads, and rewriting them
+        invalidates the signature the next request replays.
+        """
+        registry = conversation.state.secret_registry
+        if not registry.secret_sources:
+            return message
+        mask = registry.mask_secrets_in_output
+        return message.model_copy(
+            update={
+                "content": [
+                    part.model_copy(update={"text": mask(part.text)})
+                    if isinstance(part, TextContent)
+                    else part
+                    for part in message.content
+                ],
+                "reasoning_content": (
+                    mask(message.reasoning_content)
+                    if message.reasoning_content
+                    else message.reasoning_content
+                ),
+            }
+        )
 
     def _send_corrective_nudge(self, on_event: ConversationCallbackType) -> None:
         """Inject corrective feedback when no tool call and no content.
--- a/tests/sdk/agent/test_response_dispatch.py
+++ b/tests/sdk/agent/test_response_dispatch.py
@@ -200,6 +200,7 @@ def _run_single_step(
     *,
     seed_events: list[Event] | None = None,
     record_emitted_events_in_state: bool = False,
+    secrets: dict[str, str] | None = None,
 ) -> tuple[list[Event], LocalConversation]:
     """Run one agent step with a canned LLM response."""
     from pydantic import PrivateAttr
@@ -220,6 +221,8 @@ def _run_single_step(
     agent = Agent(llm=llm, tools=[])
     conversation = Conversation(agent=agent)
     conversation._ensure_agent_ready()
+    if secrets is not None:
+        conversation.update_secrets(secrets)
     if seed_events is not None:
         for event in seed_events:
             conversation.state.append_event(event)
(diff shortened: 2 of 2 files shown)


tests/sdk/agent/test_response_dispatch.py
```diff
@@ -397,3 +400,78 @@ def test_batch_action_events_are_emitted_consecutively():
     assert [a.tool_call_id for a in action_events] == ["tc-1", "tc-2"]
     # Only the first event of the batch carries the turn's thought.
     assert action_events[0].thought and not action_events[1].thought
+
+
+# ---------------------------------------------------------------------------
```
VascoSch92: delete this comment

openhands-sdk/openhands/sdk/agent/response_dispatch.py
```diff
@@ -293,6 +293,47 @@ def _emit_message_event(
         on_event(msg_event)
         return msg_event
 
+    @staticmethod
+    def _mask_secrets(message: Message, conversation: LocalConversation) -> Message:
+        """Return ``message`` with registered secret values masked in its text.
```
VascoSch92, line 298: Here please make a minimal docstring. Just important stuff

tests/sdk/agent/test_response_dispatch.py
VascoSch92: Done — banner removed in 37341bdf1.

openhands-sdk/openhands/sdk/agent/response_dispatch.py
VascoSch92: Done in 37341bdf1 — cut to the one thing that isn't visible from the code (why `thinking_blocks` / `responses_reasoning_item` are deliberately skipped). Everything else moved to the commit message.

enyst (approved):
Well. If we mask in Observations, they should never get here… well, unless the user did a trick.

Well, I don’t feel strongly about it, I do think too much code always leads to bugs, and repeating it all over… well, actually, it’s kind of good, right? Repeating it a lot means the next agent will be reminded of it every time it gets it into context. 🤔 Let’s do it…

VascoSch92:
> Well. If we mask in Observations, they should never get here… well, unless the user did a trick.
> 
> Well, I don’t feel strongly about it, I do think too much code always leads to bugs, and repeating it all over… well, actually, it’s kind of good, right? Repeating it a lot means the next agent will be reminded of it every time it gets it into context. 🤔 Let’s do it…

my point is that If i give a token to the LLM I don't want that my token is saved into the trajectories

PR #4786, opened 2026-08-31
fix(tools): preserve Tom processing checkpoints

### LINKED ISSUE #4667: [Bug]: Tom processing history can checkpoint events that were never indexed
### Bug Description

Tom candidate selection reads each session's current event count (`openhands-tools/openhands/tools/tom_consult/executor.py:283-304`) and extracts the payload sent to Tom (`:313-332,374-380`). After Tom succeeds, `_save_processing_history()` re-lists each mutable event directory to reconstruct counts (`:397-416`) and is given the broader candidate-ID list rather than only successfully extracted payloads.

If an event arrives after extraction but before that second listing, history records the newer count even though the event was never sent to Tom. The next run then treats it as already processed. Sessions whose extraction produced no payload can also be checkpointed.

### Expected Behavior

The persisted checkpoint should describe exactly the snapshots successfully submitted to Tom; later events must remain detectable.

### Actual Behavior

it as `/tmp/repro_4667.py`, then run:

```bash
uv run python /tmp/repro_4667.py
```

On SDK commit `bc5dfc50d`, it exercises the real LocalFileStore, EventLog, executor, history file, and second index pass; only the external Tom call is a fixture, where a real event is appended during submission and prints:

```text
indexed_event_count=1
actual_event_count=2
checkpoint_directory_entry_count=4
first_sessions_processed=1
second_sessions_processed=0
```

### Suggested Implementation

Carry each candidate's pre-submission directory-entry count alongside its extracted payload. After a successful Tom call, persist that count only for sessions whose extraction produced a submitted payload, and perform no second event-directory read. This keeps the checkpoint in the same unit used by candidate selection while preserving the current history JSON shape.

### Acceptance Criteria

- [ ] `last_event_count` equals the pre-submission directory count for the successfully extracted and submitted payload.
- [ ] An event appended during processing is detected on the next run.
- [ ] Sessions that fail extraction are not marked processed.
- [ ] A failed Tom call does not advance history.
- [ ] Existing history entries and JSON compatibility are preserved.

### Version

Epochex:
HUMAN:

---

AGENT:

## Why

Tom candidate selection counted event-directory files before extraction, but processing history re-listed each mutable directory after the Tom call. An event appended during that call could therefore be checkpointed even though it was never indexed. Sessions that produced no payload could also be marked processed.

## Summary

- Carry each selected session's pre-submission file count through processing.
- Advance history only for sessions successfully extracted and submitted to Tom.
- Preserve the existing history JSON shape and file-count comparison unit.

## Issue Number

Fixes #4667

## How to Test

- `uv run pytest tests/tools/tom_consult -q` — 5 passed.
- `uv run pre-commit run --files openhands-tools/openhands/tools/tom_consult/executor.py tests/tools/tom_consult/test_tom_consult_executor.py` — all hooks passed.

The regression adds a non-event metadata file so directory file count differs from `EventLog` length, appends a new event inside the Tom call, verifies the stored checkpoint remains at the pre-submission count, and confirms the next run detects the new event. Separate tests cover empty extraction and failed Tom calls.

## Video/Screenshots

Not applicable; this changes backend checkpoint persistence.

## Design Doc

Not applicable; this is a scoped persistence fix with no public API or file-format change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This follows the issue review feedback: checkpoint and candidate comparison both remain in event-directory file-count units rather than mixing in `EventLog` length.

### DIFF
--- a/openhands-tools/openhands/tools/tom_consult/executor.py
+++ b/openhands-tools/openhands/tools/tom_consult/executor.py
@@ -281,7 +281,7 @@ class TomConsultExecutor(
         processing_history = self._load_processing_history()
 
         # Find sessions that need processing
-        sessions_to_process = []
+        sessions_to_process: dict[str, int] = {}
         for session_id in all_sessions:
             events_dir = f"{self.conversations_dir}/{session_id}/events"
             event_files = self.file_store.list(events_dir)  # type: ignore
@@ -292,12 +292,12 @@ class TomConsultExecutor(
 
             # Check if needs processing (new or has new events)
             if session_id not in processing_history:
-                sessions_to_process.append(session_id)
+                sessions_to_process[session_id] = current_event_count
                 logger.info(f"📋 Tom: Session {session_id} needs processing (new)")
             elif current_event_count > processing_history[session_id].get(
                 "last_event_count", 0
             ):
-                sessions_to_process.append(session_id)
+                sessions_to_process[session_id] = current_event_count
                 logger.info(
                     f"📋 Tom: Session {session_id} has new events "
                     f"({current_event_count} events)"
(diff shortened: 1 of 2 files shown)


enyst (commented):



No code-level blocking finding in this draft. The focused Tom tests pass locally (`5 passed`). The regression uses the real event-log and file-store behavior while replacing only the external Tom call, and it verifies all three important checkpoint guarantees: an event arriving during submission is seen next time, empty extraction is not checkpointed, and a failed Tom call does not advance history.

Because this changes which conversation events feed persistent user-model memory, it needs lightweight eval/integration evidence and explicit human-maintainer confirmation before approval. The human author also still needs to replace the reserved `HUMAN:` placeholder before marking the PR ready; I did not edit it.

PR #4787, opened 2026-08-31
fix(agent-server): serialize bash output ordering

### LINKED ISSUE #4659: [Bug]: Concurrent stdout and stderr chunks can reuse BashOutput order values
### Bug Description

The two stream-reader coroutines in `openhands-agent-server/openhands/agent_server/bash_service.py:288-346` share `output_order`. Each builds and persists a `BashOutput` using the current value, awaits `_pub_sub(output_event)`, and increments only afterward at `:315-325`.

Because stdout and stderr readers run concurrently, the peer can execute during that await and allocate the same order. This is reachable with WebSocket subscribers because publication awaits `websocket.send_json`.

Duplicate persisted/public order values make reconstruction ambiguous and can cause clients polling with `order__gt` to skip output.

### Expected Behavior

Every output event for a command should have a unique, contiguous order, with the final exit event last.

### Actual Behavior

it as `/tmp/repro_4659.py`, then run:

```bash
uv run python /tmp/repro_4659.py
```

On SDK commit `6d3881035`, it exercises real BashEventService, subprocesses writing more than 1 MiB to both streams, persistence, and a yielding Subscriber; no mocks and prints:

```text
output_orders=[0, 0, 2]
duplicate_orders=[0]
stdout_chars=1048577
stderr_chars=1048577
persisted_event_files=4
```

### Suggested Implementation

Give each command one small synchronous sequencer/emitter. Reserve and increment the next order before the first await, then persist and publish. Keep stdout/stderr buffers local to their readers and send the final event through the same sequencer.

### Acceptance Criteria

- [ ] Interleaved oversized stdout and stderr produce unique contiguous order values.
- [ ] A deliberately yielding subscriber cannot cause duplicate orders.
- [ ] Both streams reconstruct exactly and the final exit event is last.
- [ ] Persistence-before-publication, chunk size, timeout behavior, and REST/event schemas remain unchanged.

### Version

Current `main` at `6d3881035`. Found by static concurrency analysis during a read-only audit.

Epochex:
HUMAN:

---

AGENT:

## Why

The stdout and stderr reader coroutines shared one output counter, but incremented it only after awaiting publication. A yielding subscriber could therefore let both streams persist and publish the same order value, making reconstruction and order-based polling ambiguous.

## Summary

- Route chunk, final, and error events through one sequenced emitter.
- Persist first, then reserve the next order before publication can yield.
- Add concurrent stream and persistence-failure regression coverage.

## Issue Number

Fixes #4659

## How to Test

- `uv run pytest tests/agent_server/test_bash_service.py::test_interleaved_bash_output_orders_are_contiguous tests/agent_server/test_bash_service.py::test_bash_output_persistence_failure_does_not_consume_order -q` — 2 passed.
- `uv run pytest tests/agent_server/test_bash_service.py` with the existing POSIX SIGTERM trap test deselected — 12 passed, 1 deselected.
- `uv run pre-commit run --files tests/agent_server/test_bash_service.py` — all hooks passed.
- `uv run pyright --pythonplatform Linux openhands-agent-server/openhands/agent_server/bash_service.py` — 0 errors.

The regression drives the production `BashEventService` execution path with interleaved stdout/stderr streams and a subscriber that yields. It verifies persistence before publication, contiguous unique orders, exact stream reconstruction, and the final exit event.

## Video/Screenshots

Not applicable; this changes backend event sequencing.

## Design Doc

Not applicable; the change is a small concurrency fix with no API or schema change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The one excluded full-file test exercises POSIX SIGTERM traps and is not supported on Windows. This change touches terminal/stdout behavior, so a maintainer should run the repository's lightweight eval before approval.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/bash_service.py
+++ b/openhands-agent-server/openhands/agent_server/bash_service.py
@@ -266,6 +266,28 @@ class BashEventService:
 
     async def _execute_bash_command(self, command: BashCommand) -> None:
         """Execute the bash event and create an observation event."""
+        output_order = 0
+
+        async def emit_output(
+            *,
+            exit_code: int | None = None,
+            stdout: str | None = None,
+            stderr: str | None = None,
+        ) -> None:
+            nonlocal output_order
+
+            output_event = BashOutput(
+                command_id=command.id,
+                order=output_order,
+                exit_code=exit_code,
+                stdout=stdout,
+                stderr=stderr,
+            )
+            self._save_event_to_file(output_event)
+            # Commit the sequence after persistence and before publication can yield.
+            output_order += 1
+            await self._pub_sub(output_event)
+
         try:
             # Create subprocess in a new session so we can signal the whole
             # process group on teardown (the shell's children, e.g. sleep, must
@@ -280,13 +302,12 @@ class BashEventService:
                 start_new_session=True,
             )
 
-            # Track output order and buffers
-            output_order = 0
+            # Track output buffers
             stdout_buffer = ""
             stderr_buffer = ""
 
             async def read_stream(stream, is_stderr=False):
-                nonlocal output_order, stdout_buffer, stderr_buffer
+                nonlocal stdout_buffer, stderr_buffer
 
                 buffer = stderr_buffer if is_stderr else stdout_buffer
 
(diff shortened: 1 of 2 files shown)


enyst (commented):



The sequencer is correctly placed: persistence and counter reservation are synchronous before publication can yield, and all chunk/final/error events share it. The focused suite passes locally (`13 passed`). I also verified the production path with a real subprocess writing 1,048,577 bytes to each stream and a real yielding subscriber; this head emitted orders `[0, 1, 2]`, reconstructed both streams exactly, and ended with exit code 0.

Please keep that realistic guarantee in the repository, not only in issue evidence: the new permanent regression replaces both the subprocess streams and publication with `AsyncMock`. A small POSIX/integration-style test using the real subprocess and a yielding `Subscriber` would cover OS pipe scheduling, real `PubSub`, persistence, and reconstruction—the exact concurrency boundary that failed—while the mocked persistence-failure test can remain focused.

This is terminal/stdout behavior, so it also requires the lightweight eval and explicit human-maintainer confirmation before approval. The human author must replace the reserved `HUMAN:` placeholder before taking the PR out of draft; I did not edit it.

PR #4788, opened 2026-08-31
fix(tools): mask secrets in every tool's observation at the shared chokepoint

### LINKED ISSUE #4677: [Security] Secret masking covers 1 tool out of 13 — the model reads raw secrets through file_editor, grep, glob, apply_patch and browser_use

VascoSch92:
HUMAN:

Fix secrets leakage in tools obs.

---

AGENT:

## Why

Sweeping `openhands-tools` for masking call sites returned exactly one file, `terminal/impl.py:371`. Of the 13 shipped tool packages, only `terminal` masked its output (plus `mcp/tool.py:187` in the SDK). Which tool the model happened to pick decided whether a secret was masked:

```
cat .env            via bash          ->  MASKED
open .env           via file_editor   ->  RAW
```

There was no security boundary — just one tool that happened to carry the code.

## Summary

- Mask at `ToolDefinition.__call__` (`openhands-sdk/openhands/sdk/tool/tool.py`), the single point every tool's observation passes through. It already receives `conversation`, and `acall` delegates to it, so the async path comes along for free.
- Add `SecretRegistry.mask_secrets_in_model`, which rebuilds a model with every nested string masked (nested models, lists, tuples, dict values).
- Add a guard test that fails when a new tool overrides `__call__` without delegating to `super().__call__` — the "tool #14 is covered by default" property the issue asks for.

## Issue Number

Fixes #4677

## How to Test

```
uv run pytest tests/tools/test_tool_output_secret_masking.py tests/sdk/conversation/test_secret_registry_model_masking.py -q
# 13 passed
```

The four tools named in the acceptance criteria, driven end-to-end through a real `LocalConversation` with a registered secret (not by calling the masking helper directly):

```
--file_editor--  view .env      ->  "API_KEY=<secret-hidden>"   (file on disk unchanged)
--grep--         pattern echo   ->  pattern='<secret-hidden>'
--glob--         matched path   ->  ['.../note-<secret-hidden>.txt']
--apply_patch--  nested commit  ->  commit.changes['secrets.txt'].new_content='TOKEN=<secret-hidden>'
```

The guard test was negative-checked by adding a bypassing tool and confirming it fails with a usable message:

```
These ToolDefinition subclasses override __call__ without delegating to
super().__call__, so their output never reaches the secret-masking
chokepoint: ['tests.tools.test_guard_negative_tmp.SneakyTool']
```

Wider suites, re-run against the current HEAD:

```
uv run pytest tests/sdk tests/tools -q -p no:randomly
# 1 failed, 6950 passed, 24 skipped, 10 xfailed
```

The single failure is a tmux timing flake in `tests/tools/terminal/test_terminal_session.py`, not this change:

- It passes on its own (`uv run pytest "…::test_long_running_command_follow_by_execute" -q` → 1 passed).
- A previous full run failed a *different* test in the same file (`test_command_backslash[tmux]`), which is the signature of a flaky tmux harness rather than a real regression.
- The file cannot reach the changed code path: it contains zero references to `ToolDefinition`, `conversation` or `Conversation`, and drives `create_terminal_session` directly. Its one secret-related test asserts a token *is* present in `session.execute(...)` output — correct, since calling the session directly bypasses `ToolDefinition.__call__` and therefore the mask. That test passes.

`pre-commit run --files <the four changed files>` passes (ruff format, ruff lint, pycodestyle, pyright, import rules, tool registration).

## Video/Screenshots

Not applicable — no user-facing surface. The observable behaviour is the tool output above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**Why the whole model, not `Observation.content`.** Masking `content` at the chokepoint looks sufficient and is not. `to_llm_content` is a property subclasses override, and several tools build what the model sees out of their own fields: `tom_consult` from `suggestions`/`reasoning`, `SleeptimeComputeObservation` from `message`, `terminal` from `metadata.prefix`/`suffix`/`working_dir`, `task` from `_get_task_info()`. `ApplyPatchObservation` is the clearest case — the secret lands in `commit.changes[...].new_content`, a nested model two levels below `content`. A `content`-only chokepoint would have passed review and still leaked, which is the same failure mode that produced this issue.

**One correction to the issue's framing.** The issue lists `grep -r "API_KEY" . via grep -> RAW`. The SDK's `grep` does not return file contents at all — `GrepObservation.matches` is a list of file paths, and `GlobObservation.files` likewise. The real leak on those two is the echoed `pattern` and any secret-bearing path, both of which this covers. Worth correcting so the acceptance criteria are read against what the tools actually return.

**Self-review catch.** The first cut of the recursive masker downgraded str-subclass enums: `ActionType` in `apply_patch` subclasses `str`, so `isinstance(value, str)` matched and `mask()` turned the enum member into a plain `str`, which surfaced as `PydanticSerializationUnexpectedValue` on `apply_patch`. Enum members are now returned untouched — their vocabulary is fixed, so they cannot hold a secret. Regression test: `test_preserves_str_subclass_enum_members`.

**Audit of the other ways this could miss.** Since the helper's promise is "covered by default", I checked the 26 `Observation` subclasses actually shipped for the surfaces a whole-model walk can still miss: none declare `extra="allow"` (extras live outside `model_fields`), none have `set`/`frozenset` fields, none have NamedTuple fields. Five carry a `_diff_cache` private attribute, which `model_copy` preserves — but it is populated lazily inside `visualize`, so it is `None` at mask time and the diff is rebuilt from masked fields. Verified end-to-end on a `str_replace` that writes a secret: `_diff_cache` is `None` at mask time, `new_content` comes back masked, `model_dump_json()` has no secret, the lazily-built `visualize` diff is masked, and the file on disk still holds the real value.

Rewriting the dispatch as a `match` (review feedback) surfaced two gaps of the same shape as the Enum one, fixed in `958e1af92`: `set`/`frozenset` had no branch and fell through to the catch-all unmasked, and a NamedTuple matched the `tuple` branch and came back as a plain tuple. Neither is reachable today, both are now covered by tests verified to fail without the fix.

**Known trade-off, needs your call — `file_editor` read/modify/write.** A masked `view` means the model can no longer reconstruct `old_str` for a `str_replace` on a file containing a registered secret; the edit fails to match. This is the same trade-off `terminal` masking has had all along (`cat .env` is already masked today), and failing to match is loud rather than silent. But `file_editor` both reads and writes, so it is worth deciding deliberately rather than inheriting it. The file on disk is never modified by masking — only the returned observation is, which the `file_editor` test asserts.

**Not addressed here.** `browser_use` returns screenshots; a secret rendered into an image cannot be string-masked, so the tool named in the issue title is only partly covered. Also, `screenshot_data` is a `str` of base64, so it now gets scanned per registered secret — correct but wasteful on large screenshots. Say the word if you want an opt-out hook for fields like that.

**Left in place deliberately.** `terminal/impl.py:371` and `mcp/tool.py:187` now mask redundantly. Masking is idempotent, terminal masks before its full-output-save-to-disk path, and removing them is a behaviour change in tools with their own tests. Happy to strip them in a follow-up if you would rather have one code path.

### DIFF
--- a/openhands-sdk/openhands/sdk/conversation/secret_registry.py
+++ b/openhands-sdk/openhands/sdk/conversation/secret_registry.py
@@ -21,22 +21,29 @@ FAILED_LOOKUP_RETRY_SECONDS: Final[float] = 60.0
 
 def _mask_value(value: Any, mask: Callable[[str], str]) -> Any:
     """Recursively mask every ``str`` reachable from ``value``."""
-    if isinstance(value, Enum):
-        # A str-subclass enum is a str, but masking it would downgrade the
-        # member to a plain str and break the field's serialization. Its
-        # vocabulary is fixed, so it can never hold a secret anyway.
-        return value
-    if isinstance(value, str):
-        return mask(value)
-    if isinstance(value, BaseModel):
-        return _mask_model(value, mask)
-    if isinstance(value, list):
-        return [_mask_value(item, mask) for item in value]
-    if isinstance(value, tuple):
-        return tuple(_mask_value(item, mask) for item in value)
-    if isinstance(value, dict):
-        return {key: _mask_value(item, mask) for key, item in value.items()}
-    return value
+    match value:
+        case Enum():
+            # A str-subclass enum is a str, but masking it would downgrade the
+            # member to a plain str and break the field's serialization. Its
+            # vocabulary is fixed, so it can never hold a secret anyway.
+            return value
+        case str():
+            return mask(value)
+        case BaseModel():
+            return _mask_model(value, mask)
+        case list():
+            return [_mask_value(item, mask) for item in value]
+        case tuple():
+            items = [_mask_value(item, mask) for item in value]
+            # Rebuild a NamedTuple through its own constructor; tuple(items)
+            # would downgrade it the same way masking an Enum member does.
+            return type(value)(*items) if hasattr(value, "_fields") else tuple(items)
+        case set() | frozenset():
+            return type(value)(_mask_value(item, mask) for item in value)
+        case dict():
+            return {key: _mask_value(item, mask) for key, item in value.items()}
+        case _:
+            return value
 
 
 def _mask_model[ModelT: BaseModel](model: ModelT, mask: Callable[[str], str]) -> ModelT:
--- a/openhands-sdk/openhands/sdk/conversation/secret_registry.py
+++ b/openhands-sdk/openhands/sdk/conversation/secret_registry.py
@@ -1,11 +1,12 @@
 """Secrets manager for handling sensitive data in conversations."""
 
 import time
-from collections.abc import Collection, Mapping
+from collections.abc import Callable, Collection, Mapping
+from enum import Enum
 from threading import RLock
-from typing import Final
+from typing import Any, Final
 
-from pydantic import Field, PrivateAttr, SecretStr
+from pydantic import BaseModel, Field, PrivateAttr, SecretStr
 
 from openhands.sdk.logger import get_logger
 from openhands.sdk.secret import SecretSource, SecretValue, StaticSecret
@@ -16,6 +17,39 @@ logger = get_logger(__name__)
 
 # Back-off before retrying a failed source; a failed lookup masks nothing anyway.
 FAILED_LOOKUP_RETRY_SECONDS: Final[float] = 60.0
+
+
+def _mask_value(value: Any, mask: Callable[[str], str]) -> Any:
+    """Recursively mask every ``str`` reachable from ``value``."""
+    if isinstance(value, Enum):
+        # A str-subclass enum is a str, but masking it would downgrade the
+        # member to a plain str and break the field's serialization. Its
+        # vocabulary is fixed, so it can never hold a secret anyway.
+        return value
+    if isinstance(value, str):
+        return mask(value)
+    if isinstance(value, BaseModel):
+        return _mask_model(value, mask)
+    if isinstance(value, list):
+        return [_mask_value(item, mask) for item in value]
+    if isinstance(value, tuple):
+        return tuple(_mask_value(item, mask) for item in value)
+    if isinstance(value, dict):
+        return {key: _mask_value(item, mask) for key, item in value.items()}
+    return value
+
+
+def _mask_model[ModelT: BaseModel](model: ModelT, mask: Callable[[str], str]) -> ModelT:
+    """Rebuild ``model`` with every nested string masked.
+
+    ``model_copy`` is used rather than a validate round-trip so private
+    attributes survive and no field is re-coerced.
+    """
+    updates = {
+        name: _mask_value(getattr(model, name), mask)
+        for name in type(model).model_fields
+    }
+    return model.model_copy(update=updates)
 
 
 class SecretRegistry(OpenHandsModel):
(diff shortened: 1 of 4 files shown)


openhands-sdk/openhands/sdk/conversation/secret_registry.py
```diff
@@ -18,6 +19,39 @@
...
+        return mask(value)
+    if isinstance(value, BaseModel):
+        return _mask_model(value, mask)
+    if isinstance(value, list):
+        return [_mask_value(item, mask) for item in value]
+    if isinstance(value, tuple):
+        return tuple(_mask_value(item, mask) for item in value)
+    if isinstance(value, dict):
+        return {key: _mask_value(item, mask) for key, item in value.items()}
+    return value
```
VascoSch92: can you use a match statement?

VascoSch92: Done in 958e1af92.

Rewriting the dispatch as `match` made two more gaps of the same shape as the Enum one obvious, so they are in the same commit:

- `set`/`frozenset` had no branch at all — a `set[str]` field fell through to the catch-all and was never masked.
- A NamedTuple matched the `tuple` branch and came back as a plain tuple, losing its type the same way a masked Enum member did.

Neither is reachable across the 26 Observation subclasses shipped today (I audited for set fields, NamedTuple fields and `extra="allow"` — none), but the helper's promise is that tool #14 is covered by default, so I would rather the branches exist than land as a later bug. Both are covered by tests that I verified fail without the fix.

simonrosenberg (approved):

PR #4795, opened 2026-08-31
refactor(agent-server,ci): remove overdue org_config field and catch this class of gap in check_deprecations.py

simonrosenberg:
HUMAN:

Combine the org_config removal and the check_deprecations.py extension into a single PR — they were split across #4795/#4797 but are small enough to land together.

---

AGENT:

## Why

`SkillsRequest.org_config` (`openhands-agent-server/openhands/agent_server/skills_router.py`) is a single-item legacy predecessor to `org_configs` (a list). It was marked `deprecated=True` in v1.28.0 with removal scheduled for v1.33.0. The SDK is now at v1.44.1 — 11 minor releases past its own deadline — and the field, its fallback-normalization branch in `get_skills()`, and three dedicated tests were all still in the tree.

That slipped through because `check_deprecations.py` only scanned FastAPI route decorators (`@router.get(..., deprecated=True)`) and the `openhands.sdk.utils.deprecation.deprecated` Python decorator for overdue deprecations — it had no visibility into Pydantic `Field(deprecated=True, ...)` declarations on request/response models, so the deadline passing produced no CI signal. `check_agent_server_rest_api_breakage.py` recognizes the same wording for schema-property removals, but only at the moment a property is actually removed — it never flags one that's simply overdue and still present.

This PR fixes both the specific instance and the class of bug: it removes the overdue field, and closes the tooling gap so the next one doesn't sit unnoticed for 11 releases.

## Summary

**Field removal:**
- Remove the `org_config` field from `SkillsRequest` and its fallback branch in `get_skills()` (`skills_router.py`); only `org_configs` remains.
- Remove the now-obsolete hand-written `org_config` field from `clients/typescript/src/models/api.ts`.
- Remove the stale `"org_config": None` payload key from both `_call_skills_api` implementations (`openhands-sdk/openhands/sdk/workspace/remote/base.py`, `openhands-workspace/openhands/workspace/cloud/workspace.py`) — harmless (`SkillsRequest` is a plain `BaseModel`, default `extra="ignore"`), but stale now that the field is gone.
- Remove the three tests that exercised only the deprecated single-field path (`test_get_skills_legacy_org_config_supported`, `test_org_config_validation`, `test_missing_required_org_config_fields`) from `tests/agent_server/test_skills_router.py`, replacing the latter two with equivalent `org_configs`-list-based coverage (`test_org_configs_validation`, `test_missing_required_org_configs_fields`) so the `OrgConfig` sub-schema validation path stays covered.

[...]

PR #4792, opened 2026-08-31
refactor(agent-server): remove the VNC/desktop stack entirely

simonrosenberg:
HUMAN:

Remove the VNC/desktop stack entirely — it has been off everywhere since the 2025 security fix and its opt-in never worked.

---

AGENT:

## Why

The VNC/desktop feature was added in #515 (2025-09-29) for interactive browsing — you could watch the agent drive a real browser over noVNC. Six weeks later, **#974 (2025-10-30) turned it off** as a security fix: the VNC server was exposed unauthenticated by default. The PR that would re-enable it safely, **#1160 (token-based auth), has been an open draft since 2025-11-13** — last touched 2026-07-29.

So for ten months we have shipped ~72 MB of desktop stack in every image, which every consumer explicitly disables and nobody can safely turn on.

### 1. No consumer enables it. Not one.

Every downstream repository sets it off, and none sets it on:

| Consumer | Setting |
|---|---|
| `enterprise` — `remote_sandbox_spec_service.py` | `OH_ENABLE_VNC: '0'` |
| `saas-deploy` — production / staging / development | `"0"` in all three |
| `runtime-api` — `warm_runtimes.py`, `monitor_runtime.py`, README | `"0"` |
| `OpenHands-Cloud` — helm values, runtime-api chart, replicated | `"0"` |
| `benchmarks` — `apptainer_build.py`, `Dockerfile.agent-layer-commit0` | `false` |
| `sandbox-server` — docker/remote sandbox spec services | off |
| Agent Canvas (`OpenHands/OpenHands`) | no VNC references at all in `src` |
| This repo's Dockerfile default | `OH_ENABLE_VNC=false` |

The one place the value flows rather than being hardcoded — `runtime-api/runtimes.py` — lists it under *"Deployment-infrastructure keys set via ConfigMap, not by callers"*, and those keys are stripped when matching a warm pod to a request. It is not a customer-facing toggle either.

### 2. The documented opt-in never worked

#974's description says it preserved backward compatibility by *"Added automatic VNC enablement when `extra_ports=True`"*.

**That code is not in `main`.** `OH_ENABLE_VNC` appears **zero** times anywhere in `openhands-workspace`, and `git log -S 'OH_ENABLE_VNC' -- openhands-workspace/` returns nothing — it was never there and removed; it simply was not implemented. `extra_ports=True` only mapped host port +2 to container 8002, where nothing was listening because the image default is `false`.

[...]

PR #4797, opened 2026-08-31
feat(ci): catch overdue Pydantic Field-level deprecations in check_deprecations.py

### LINKED ISSUE #4794: [Feature]: check_deprecations.py should catch overdue Pydantic Field-level deprecations
## Is there an existing feature request for this?

## Problem or Use Case

`.github/scripts/check_deprecations.py` fails CI when a version-based deprecation has less than 5 minor releases of runway, or when the current package version has reached/passed a feature's declared removal version. It already covers two mechanisms:

1. The `openhands.sdk.utils.deprecation.deprecated` decorator (`removed_in` / `deprecated_in` kwargs).
2. FastAPI REST route operations marked `deprecated=True`, via a docstring note (`Deprecated since vX.Y.Z and scheduled for removal in vA.B.C.`).

It does **not** scan Pydantic `Field(deprecated=True, description="Deprecated since v... and scheduled for removal in v...")` declarations on request/response models. That gap let `SkillsRequest.org_config` (`openhands-agent-server/openhands/agent_server/skills_router.py`) sit 11 minor releases past its own v1.33.0 removal deadline with zero CI signal (see the companion issue for removing that specific field). The `check_agent_server_rest_api_breakage.py` script *does* recognize the same wording for schema-property removals, but only at the moment a property is actually removed — it never flags one that's simply overdue and still present, which is exactly what `check_deprecations.py`'s overdue/runway logic is for.

## Desired Behavior

Extend `check_deprecations.py` to also walk Pydantic model class bodies for `Field(...)` assignments where `deprecated=True` is set, parse the same `Deprecated since vX.Y.Z and scheduled for removal in vA.B.C.` wording out of the field's `description=` kwarg (reusing `REST_ROUTE_DEPRECATION_RE`), and feed those into the existing overdue/invalid-runway reporting alongside REST routes and the Python decorator — so an overdue Pydantic field deprecation fails CI the same way an overdue REST route or `@deprecated`-decorated function already does.

## Acceptance Criteria

- [ ] `check_deprecations.py` detects `Field(deprecated=True, ...)` class attributes across all four scanned packages and extracts their deprecated/removed versions from the description text.
- [ ] Running the script against a fixture with an overdue Pydantic field deprecation (removal version <= current package version) fails with a clear message, matching the existing REST-route overdue message style.
- [ ] Running it against a fixture with a too-short runway (< 5 minor releases) fails with the existing invalid-runway message style.
- [ ] Existing REST-route and decorator-based checks are unaffected (no regressions in current passing runs).

## Feature Area

- [x] Testing / Development tools

simonrosenberg:
HUMAN:

Extend check_deprecations.py to catch overdue Pydantic Field-level deprecations, per the follow-up from #4792/#4795.

---

AGENT:

## Why

`check_deprecations.py` fails CI when a version-based deprecation schedule provides less than 5 minor releases of runway, or when the current package version has reached/passed a feature's declared removal version. It already covers two mechanisms:

1. The `openhands.sdk.utils.deprecation.deprecated` decorator (`removed_in` / `deprecated_in` kwargs).
2. FastAPI REST route operations marked `deprecated=True`, via a docstring note (`Deprecated since vX.Y.Z and scheduled for removal in vA.B.C.`).

It does not scan Pydantic `Field(deprecated=True, description="Deprecated since v... and scheduled for removal in v...")` declarations on request/response models. That gap let `SkillsRequest.org_config` (`openhands-agent-server/openhands/agent_server/skills_router.py`, removed in #4795) sit 11 minor releases past its own v1.33.0 removal deadline with zero CI signal. `check_agent_server_rest_api_breakage.py` recognizes the same wording for schema-property removals, but only at the moment a property is actually removed — it never flags one that's simply overdue and still present, which is exactly what `check_deprecations.py`'s overdue/runway logic is for.

## Summary

- Add `_gather_pydantic_field_deprecations()`, which walks `ClassDef` bodies for `AnnAssign` attributes assigned a `Field(...)` call with `deprecated=True`, and — only when the field's `description` matches the existing `Deprecated since vX.Y.Z and scheduled for removal in vA.B.C.` wording (reusing `REST_ROUTE_DEPRECATION_RE`) — yields a `DeprecationRecord` of a new `"pydantic_field"` kind, feeding the same overdue/invalid-runway checks that REST routes and the Python decorator already go through.
- **Deliberately permissive, not strict**: a `Field(deprecated=True)` with no version-based schedule in its description is silently skipped rather than flagged as a format violation. Unlike REST route operations (where this repo's policy mandates the versioned wording for every public `/api/**` deprecation), several existing Pydantic fields are intentionally open-ended backward-compatibility aliases with no planned removal — e.g. `SubAgentInfo.mcp_servers` and `MCPProbeResult.resolved_mcp_servers` in `openhands-agent-server`, and `TaskAction.max_turns` in `openhands-tools`. Requiring all of them to adopt a hard removal date wasn't the ask and would force a policy change on working code well outside this issue's scope — so only fields that *do* declare a schedule are held to it.
- Wire the new gatherer into `main()`'s per-package loop unconditionally (not gated to `openhands-agent-server`, since Field-level deprecations can appear in any of the four packages).
- Add unit tests mirroring the existing REST-route test style: positive detection of a scheduled field, a skip test for an open-ended field, a skip test for a non-deprecated field, an overdue-record `_should_fail` test, and an end-to-end `main()` fixture test (analogous to `test_main_fails_for_invalid_runway`) proving an overdue Pydantic field actually fails the script with a clear message.

Verified against the live repo tree: the only currently-scheduled `Field(deprecated=True, ...)` was `org_config`, removed in #4795 (this branch is stacked on that PR so its own CI stays green); the four open-ended ones above are correctly left untouched.

## How to Test

```bash
uv run --frozen ruff check .github/scripts/check_deprecations.py tests/cross/test_check_deprecations.py
uv run --frozen ruff format --check .github/scripts/check_deprecations.py tests/cross/test_check_deprecations.py
# All checks passed! / already formatted

uv run --frozen pytest tests/cross/test_check_deprecations.py -q
# 14 passed (7 new)

uv run --frozen pytest tests/cross -q --timeout=60
# full cross-tests suite, no regressions

# Live run against this repo tree
uv run --with packaging python .github/scripts/check_deprecations.py
# openhands-agent-server: checked 1 deprecation metadata entries against version 1.44.1.
# (the desktop_router stub from #4792 — correctly within its runway)
```

## Stacking note

This branch is stacked on #4795 (removes the only currently-overdue field this check would have caught) so this PR's own CI isn't red on day one from pre-existing repo state unrelated to this diff. GitHub will show only this PR's own commit once #4795 merges and this branch is rebased onto `main`.

## Issue Number

Fixes #4794

### DIFF
--- a/.github/scripts/check_deprecations.py
+++ b/.github/scripts/check_deprecations.py
@@ -2,12 +2,20 @@
 """Static analysis for deprecation deadlines.
 
 This script scans Python deprecation metadata (`deprecated`, `warn_deprecated`,
-`warn_cleanup`) and agent-server REST routes marked `deprecated=True`. It fails
-when a version-based deprecation schedule provides less than 5 minor releases of
-runway, or when the current project version has reached or passed a feature's
-removal marker. This catches invalid schedules when they are introduced and later
-ensures legacy shims and overdue deprecated REST endpoints are cleaned up before
-release.
+`warn_cleanup`), agent-server REST routes marked `deprecated=True`, and Pydantic
+`Field(deprecated=True, ...)` declarations whose description carries the same
+versioned schedule wording. It fails when a version-based deprecation schedule
+provides less than 5 minor releases of runway, or when the current project
+version has reached or passed a feature's removal marker. This catches invalid
+schedules when they are introduced and later ensures legacy shims, overdue
+deprecated REST endpoints, and overdue deprecated schema fields are cleaned up
+before release.
+
+Not every `Field(deprecated=True)` declares a version-based schedule -- some are
+open-ended backward-compatibility aliases with no planned removal. Only fields
+whose description matches the "Deprecated since vX.Y.Z and scheduled for
+removal in vA.B.C." wording are subject to the runway/overdue check; others are
+left alone.
 """
 
 from __future__ import annotations
@@ -87,7 +95,9 @@ class DeprecationRecord:
     deprecated_in: str | None
     path: Path
     line: int
-    kind: Literal["decorator", "warn_call", "cleanup_call", "rest_route"]
+    kind: Literal[
+        "decorator", "warn_call", "cleanup_call", "rest_route", "pydantic_field"
+    ]
     package: str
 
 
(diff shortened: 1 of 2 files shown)


simonrosenberg:
Consolidating into #4795 — same content, both commits now live on that branch. Closing this one rather than merging separately.

PR #4807, opened 2026-09-01
feat(agent-server): add /sockets/session/{id} with a non-Event envelope

### LINKED ISSUE #4681: Streaming step 3+4: /sockets/session/{id} — envelope wire, cursor, paged replay, byte-budget admission
Part of #4671 (steps 3 and 4).

## Desired Behavior

A second WebSocket endpoint whose frames are **envelopes, not events**, carrying durable traffic only. The legacy `/sockets/events/{id}` is left untouched and keeps working; a client speaks one or the other, so **the URL is the protocol version** — no handshake, no `protocol=2` field, no dual-decode path.

## What the endpoint has to fix

Three defects in the legacy endpoint, all of them structural rather than incidental:

**1 · The wire is the disk.** `sockets.py:489` sends `event.model_dump(...)`, `event_store.py:217` persists `event.model_dump_json(...)`, `remote_conversation.py:239` decodes with `Event.model_validate`. One class, `extra="forbid"`, three roles — so every additive wire field is a storage migration. The `Event` must ride *inside* an envelope as an untouched payload instead.

**2 · History and live traffic interleave.** The legacy endpoint subscribes (`sockets.py:298`) and *then* replays (`:~325`), so live events arrive in the middle of history. The fix is subscribe-in-buffering-mode → read the high-water `seq` → paged replay → flush the buffer against the mark.

**3 · A slow consumer wedges the publisher.** `_WebSocketSubscriber.__call__` awaits `websocket.send_json` directly, so one stuck client stalls `PubSub.__call__`'s `gather` and with it the publisher. Admission must be synchronous and byte-bounded, with the socket awaited only by a single per-connection writer task. Over budget drops the **connection**, not a frame — the client resumes losslessly with `after_seq`.

## Cursor

`after_seq` replaces the legacy `resend_mode` / `after_timestamp` pair, which compared naive local timestamps and could not express "exactly where I left off". `seq` is the event's index in the log — already on disk as the `{idx:05d}` component of the filename, so exposing it needs no migration.

## Status

A working implementation exists on branch `vasco/session-socket`, with:

- `session_protocol.py` — the envelope and all six frame types
- `session_socket.py` — the endpoint, the byte-bounded single writer, paged replay
- 7 tests, covering four of the six behaviours the epic lists as untested

Verified: the embedded payload is **byte-identical** to what the legacy endpoint sends (`frame["event"] == event.model_dump(mode="json", exclude_none=True)`); pyright clean; 99 existing socket/API tests still pass.

## Two findings from implementing it

**The replay boundary needs no lock.** The design called for taking the state lock and reading the high-water `seq` in the same acquisition. It isn't necessary. Subscribe *first* (buffering), then read the mark, then dedupe by seq:

- appended **before** subscribe -> on disk before the mark is read, so `mark >= seq` -> replay covers it;
- appended **after** subscribe -> in the buffer; if `seq <= mark` it is also replayed and the buffered copy is dropped by seq, otherwise the buffer is the only copy and the flush delivers it.

Same guarantee, and it keeps one more consumer off a `FIFOLock` that already guards six unrelated things.

**The two-category model needs a third category.** `Durable` and progress frames have no home for events that are *published but never persisted* — and the bus carries one: `subscribe_to_events` pushes a synthetic `ConversationStateUpdateEvent` snapshot that is not in the event log. Without somewhere to put it, either `seq` on `Durable` becomes a lie or the snapshot is dropped. Hence a `Transient` frame: **if a frame has a `seq`, that sequence number is real and resumable.** This is a genuine gap in the protocol as designed, not an implementation detail.

## Out of scope here

`ItemStarted` / `Delta` / `ItemAborted` are defined and the endpoint carries them, but nothing produces them — that is step 5. The three client switches are their own PRs.

## Acceptance Criteria

- [ ] New endpoint serves durable events in an envelope; `Event` is unchanged on disk and inside the frame.
- [ ] Reconnect with `after_seq` loses nothing and duplicates nothing.
- [ ] A wedged consumer is dropped without the publisher blocking.
- [ ] `StreamingDeltaEvent` never appears on this endpoint.

Needs: step 2 for the `seq` guarantee to be strict; step 0 for the byte budget to be a measured number rather than a placeholder.

VascoSch92:
HUMAN:

This PR add a new endpoint for streaming. You can find the architecture diagram into the description to understand what is the idea behind.

---

AGENT:

## Why

The legacy `/sockets/events/{id}` endpoint ships the durable record itself
over the wire: `sockets.py` sends `event.model_dump(...)`, `event_store.py`
persists `event.model_dump_json(...)`, and `remote_conversation.py` decodes
with `Event.model_validate`. One class, `extra="forbid"`, three roles — so
every additive wire field is a storage migration, which is why a token delta
cannot carry a stream identity today. It also has two structural bugs: the
subscriber is registered *and then* replays history, so live events can
interleave with the replay; and `_WebSocketSubscriber.__call__` awaits
`websocket.send_json` directly, so one wedged browser tab stalls
`PubSub.__call__`'s `gather` and, with it, the publisher for every other
subscriber.

## Summary

Adds a second, additive endpoint, `/sockets/session/{id}`, whose frames are
envelopes rather than raw events. The legacy endpoint is untouched — a client
speaks one or the other, so the URL is the protocol version (no handshake, no
`protocol` field, no dual-decode path).

- `session_protocol.py`: the envelope hierarchy (`SyncFrame`, `DurableFrame`,
  `TransientFrame`, `ItemStartedFrame`, `DeltaFrame`, `ItemAbortedFrame`,
  `ErrorFrame`). `Event` rides inside `DurableFrame`/`TransientFrame` as an
  untouched payload — same class, same `extra="forbid"`, same bytes on disk —
  everything protocol-specific (`seq`, stream identity, progress) lives on the
  envelope instead, where adding a field costs nothing.
- `session_socket.py`: the endpoint itself.
  - **History and live traffic no longer interleave.** The subscriber
    registers in buffering mode *first*, the high-water `seq` is read only
    after that, and the buffer is flushed against that mark once paged replay
    finishes — see the docstring on the "atomic replay boundary" in
    `session_socket.py` for why this needs no lock.
  - **A slow consumer can no longer wedge the publisher.** `_ConnectionWriter`
    makes admission synchronous and byte-bounded (`MAX_FRAME_BYTES`,
    `MAX_PENDING_BYTES` — both placeholders pending a measured byte budget,
    see Notes); the socket itself is awaited only by a single per-connection
    writer task. Over-budget drops the *connection*, never a frame — the
    client resumes losslessly with the new `after_seq` cursor.
  - `after_seq` (the event's index in the log) replaces the legacy
    `resend_mode`/`after_timestamp` pair, which compared naive local
    timestamps and could not express "exactly where I left off." It needs no
    migration: it's already on disk as the `{idx:05d}` component of
    `event-{idx:05d}-{event_id}.json`.
  - `StreamingDeltaEvent` is filtered out rather than forwarded — deltas don't
    belong on the durable channel, and forwarding them would recreate the
    coupling this endpoint exists to remove.
- `api.py`: registers the new `session_router` alongside the existing
  `sockets_router`.
- `tests/agent_server/test_session_socket.py`: 10 unit tests, plus 5 end-to-end
  tests in `test_agent_server_wsproto.py` driving the endpoint against a real
  server (see How to Test).

**Depends on #4680** (draft PR #4806, stacked underneath this one — merge
order matters). `_SessionSubscriber._seq_of` resolves a durable frame's `seq`
via `EventLog.get_index(event.id)`; without #4680's fix that lookup can race
an event that was published before it was persisted and degrade to a
`TransientFrame` (logged, and still delivered to the client, just without a
resumable `seq`). With #4680 the publish-before-persist race is closed, so that
degrade path stops being reachable for anything except the one genuinely
unpersisted event — the snapshot `subscribe_to_events` synthesises on connect.

## Architecture

Both endpoints hang off the same `PubSub`. The difference is what leaves it:
the legacy one puts the durable record itself on the wire, this one wraps it.
Persist-before-publish is #4680, stacked underneath.

```mermaid
flowchart LR
    AG["agent step"] --> CB["callback chain"]
    CB -->|"1 - persist"| LOG[("EventLog<br/>event-00042-abc.json")]
    CB -->|"2 - publish"| PS["PubSub"]

    PS --> LSUB["legacy subscriber<br/>awaits the socket itself"]
    PS --> SSUB["_SessionSubscriber<br/>buffers, then goes live"]

    LSUB -->|"raw Event JSON"| OLD["/sockets/events/id<br/>untouched"]
    SSUB -->|"Event inside an envelope"| WR["_ConnectionWriter<br/>one task, byte-bounded"]
    WR --> NEW["/sockets/session/id"]

    LOG -.->|"seq lookup, paged replay"| SSUB
```

Three consequences, and they are the whole PR:

- **The wire stops being the disk.** `Event` goes inside a `DurableFrame`
  untouched, so `seq` and later stream identity live on the envelope where
  adding a field is not a storage migration.
- **Only `_ConnectionWriter` awaits the socket.** `send()` is synchronous and
  byte-bounded, so a wedged client can no longer stall `PubSub`'s `gather` and
  with it the publisher. Over budget drops the connection, not a frame.
- **`seq` makes reconnects lossless**, which is what lets that drop be safe.

The subtle part is the connect handshake — subscribing *before* reading the
high-water mark is what makes replay and live traffic neither interleave nor
duplicate, with no lock:

```mermaid
sequenceDiagram
    participant C as client
    participant E as endpoint
    participant S as _SessionSubscriber
    participant L as EventLog

    C->>E: connect ?after_seq=N
    E->>S: subscribe — buffering
    Note over S: live events accumulate,<br/>nothing sent yet
    E->>L: len() → mark
    E-->>C: Sync from_seq=N through_seq=mark
    loop pages of 100
        E->>L: read page
        E-->>C: Durable seq N+1 … mark
    end
    E->>S: go_live(mark)
    Note over S: flush buffer,<br/>drop anything seq ≤ mark
    S-->>C: Durable — live
```

Anything appended *before* subscribe is on disk before the mark is read, so
replay covers it. Anything appended *after* lands in the buffer and is deduped
against the mark by `seq`. A client that omits `after_seq` gets `go_live(None)`
instead, since nothing was replayed for it to dedupe against.

## Issue Number

Fixes #4681

## How to Test

**End-to-end, against a real server.** `uv run pytest
tests/agent_server/test_agent_server_wsproto.py -q` — 11 passed. Five of those
are new and drive `/sockets/session/{id}` over a real WebSocket against an
agent-server started on a real port (reusing the existing `agent_server`
fixture), so the endpoint handler itself is executed — auth, conversation
lookup, subscribe, the sync frame, paged replay from disk, `go_live`, the
inbound loop and the close codes:

- `test_session_socket_opens_with_a_sync_frame` — the first frame is always `sync`.
- `test_session_socket_durable_frame_carries_seq_over_the_wire` — a message sent
  on the socket comes back as a `durable` frame with a real, monotonic `seq` and
  the untouched `Event` payload; `transient` frames carry no `seq`.
- `test_session_socket_resumes_history_with_after_seq` — reconnecting with
  `after_seq=-1` replays from disk starting at seq 0, covers everything the first
  connection saw, and delivers nothing twice across the replay/live seam.
- `test_session_socket_unknown_conversation_closes_4004`.
- `test_session_socket_survives_a_malformed_inbound_frame` — a garbage frame
  earns an `error` frame and the connection stays usable.

This was worth doing rather than asserting: it caught that the endpoint had
**zero** coverage before (no test referenced `/sockets/session`, `session_router`
or called `session_socket()`), and the handler-level fixes in the two
following commits all came out of reviewing that path.

Unit level and regressions:

- `uv run pytest tests/agent_server/test_session_socket.py -q` — 10 passed
  (7 original + regression tests for the malformed-frame crash, the
  state-update dedupe and the live-only buffer). Each of those three was
  verified to fail against the unfixed code first — the `JSONDecodeError`
  escaping `_inbound_loop`, and the state update arriving twice.
- Full `tests/agent_server` suite on this branch: **2081 passed**, 13 deselected,
  0 failed — no regressions in the legacy `/sockets/events/{id}` endpoint or
  anywhere else.
- `pre-commit run` (ruff format, ruff lint, pycodestyle, pyright, import
  dependency rules, Tool subclass registration) passes on all changed files.
- Read both new modules in full to check the docstring/commit claims against the
  code: the buffer-then-mark replay boundary, the byte-bounded non-blocking
  `_ConnectionWriter.send`, the `StreamingDeltaEvent` filter and the `_seq_of`
  degrade path all check out.

## Video/Screenshots

N/A — backend WebSocket endpoint, no UI surface.

## Design Doc

N/A — the module docstrings in `session_protocol.py` and `session_socket.py`
already carry the full design rationale (why a separate wire hierarchy, the
delivery rules, the replay-boundary proof) inline with the code.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Scoped to what #4681 asks for (steps 3+4 of #4671): the envelope wire,
  cursor-based replay, and byte-bounded admission. `ItemStarted`/`Delta`/
  `ItemAborted` are defined and carried on the wire but nothing produces them
  yet — that needs `StreamContext` in the SDK at all four streaming entry
  points (`Agent.step`, `Agent.astep`, both ACP equivalents), which is a
  separate step (5) per the epic. The three client-side switches to this
  endpoint are their own PRs too.
- `MAX_FRAME_BYTES`/`MAX_PENDING_BYTES` in `session_protocol.py` are
  explicitly flagged as provisional placeholders — the design calls for
  deriving the pending-byte cap from a measured frame-size distribution
  rather than picking one, and that measurement hasn't happened yet (tracked
  as its own "reproductions and baselines" issue, referenced as step 0 in
  #4671).
- Merge order: #4680 first, then this one — this branch is based on
  `vasco/streaming-4680-persist-before-publish`, not `main`, so the diff here
  is only the session-socket work.
- Two real defects came out of the second review pass and are fixed here:
  `_seq_of` decided durable-vs-transient by *type*, treating every
  `ConversationStateUpdateEvent` as unpersisted — but only the snapshot
  `subscribe_to_events` synthesises is; the rest are appended normally
  (`append_event` stores them and merely skips advancing HEAD, verified against
  a live conversation). That stripped their real `seq` and defeated the
  `seq <= through_seq` dedupe, so a state update caught in the replay window
  arrived twice. And `go_live` was handed the high-water mark even when no
  replay ran, so a live-only client silently lost anything buffered during
  `subscribe_to_events`. Both now have regression tests.
- Reviewed and hardened after the first pass. Also fixed: a malformed inbound frame
  crashing `_inbound_loop` (`receive_json` raises `JSONDecodeError`, only
  `WebSocketDisconnect` was caught); replay yielding per page rather than per
  frame, which could exhaust the pending-byte budget and drop a client that was
  never slow; `_read_page` not tolerating `OSError`; `aclose()` swallowing a
  cancellation aimed at the calling task; and a conversation closing between the
  lookup and `get_conversation()` raising out of the handler instead of closing
  4004.
- Known and deliberate: `_SessionSubscriber.go_live` flushes its buffer
  synchronously, so it cannot yield to the writer mid-flush the way `_replay`
  now does. The buffer only holds events that arrived *during* replay, so it is
  bounded by replay duration rather than by history size — worth revisiting if
  the byte budget is ever tightened.

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:cab6a02-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-cab6a02-python \
  ghcr.io/openhands/agent-server:cab6a02-python
```

```
ghcr.io/openhands/agent-server:cab6a02-golang-amd64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-golang-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-golang-amd64
ghcr.io/openhands/agent-server:cab6a02-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:cab6a02-golang-arm64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-golang-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-golang-arm64
ghcr.io/openhands/agent-server:cab6a02-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:cab6a02-java-amd64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-java-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-java-amd64
ghcr.io/openhands/agent-server:cab6a02-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:cab6a02-java-arm64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-java-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-java-arm64
ghcr.io/openhands/agent-server:cab6a02-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:cab6a02-python-amd64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-python-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-python-amd64
ghcr.io/openhands/agent-server:cab6a02-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:cab6a02-python-arm64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-python-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-python-arm64
ghcr.io/openhands/agent-server:cab6a02-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:cab6a02-python-slim-amd64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-python-slim-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-python-slim-amd64
ghcr.io/openhands/agent-server:cab6a02-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-amd64
ghcr.io/openhands/agent-server:cab6a02-python-slim-arm64
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-python-slim-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-python-slim-arm64
ghcr.io/openhands/agent-server:cab6a02-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-arm64
ghcr.io/openhands/agent-server:cab6a02-golang
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-golang
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-golang
ghcr.io/openhands/agent-server:cab6a02-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:cab6a02-java
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-java
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-java
ghcr.io/openhands/agent-server:cab6a02-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:cab6a02-python-slim
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-python-slim
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-python-slim
ghcr.io/openhands/agent-server:cab6a02-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim
ghcr.io/openhands/agent-server:cab6a02-python
ghcr.io/openhands/agent-server:cab6a024ec33b4c997b60ca0ce91073fa15fff65-python
ghcr.io/openhands/agent-server:vasco-streaming-4681-session-socket-python
ghcr.io/openhands/agent-server:cab6a02-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `cab6a02-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `cab6a02-python-amd64`) are also available if needed
</details>

### DIFF
--- a/openhands-agent-server/openhands/agent_server/session_protocol.py
+++ b/openhands-agent-server/openhands/agent_server/session_protocol.py
@@ -1,37 +1,24 @@
 """Wire protocol for ``/sockets/session/{conversation_id}``.
 
-Deliberately not built on :class:`~openhands.sdk.event.base.Event`: the legacy
-``/sockets/events/{id}`` endpoint sends ``event.model_dump(...)`` over the wire
-while ``event_store.py`` persists ``event.model_dump_json(...)`` and
-``remote_conversation.py`` decodes with ``Event.model_validate`` — one class,
-``extra="forbid"``, three roles, so every additive wire field becomes a
-storage migration. Here ``Event`` rides *inside* an envelope, untouched;
-everything protocol-specific (sequence number, stream identity, progress)
-lives on the envelope instead, where adding a field costs nothing.
+Not built on ``Event``: the legacy endpoint sends the same class it persists
+and validates, and ``Event`` is ``extra="forbid"``, so every wire field would
+be a storage migration. Here ``Event`` rides inside an envelope, untouched,
+and protocol fields live on the envelope.
 
-There is no handshake and no ``protocol`` field: the URL is the protocol
-version. A client speaks ``/sockets/events/{id}`` or ``/sockets/session/{id}``,
-never both.
-
-``ItemStarted`` mints nothing — ``item_id`` *is* the ``id`` of the ``Event``
-that eventually carries the finished message, so a client retires an open slot
-on a single equality test: ``frame.event.id == slot.item_id``. No close frame,
-no watermark.
+The URL is the protocol version — no handshake, no ``protocol`` field.
 
 Delivery rules:
 
-1. ``Durable`` never goes missing across a reconnect (``after_seq`` covers
-   it); within a connection it may be dropped, which is what lets a slow
-   consumer be disconnected instead of blocking the publisher.
-2. ``Delta`` may be dropped freely — a gap just marks the slot lossy, and the
-   real text arrives with the durable message.
-3. Every ``ItemStarted`` is retired by exactly one ``Durable`` or one
-   ``ItemAborted``, including on cancellation, provider failure, policy
-   rejection and unhandled exception.
-4. Progress frames are never replayed; on reconnect a client discards every
-   open slot and waits for the message on the cursor.
-5. No ordering is promised between two open items beyond the order their
-   ``ItemStarted`` frames arrived in.
+1. ``Durable`` survives a reconnect via ``after_seq``; within a connection it
+   may be dropped, which is what lets a slow consumer be disconnected instead
+   of blocking the publisher.
+2. ``Delta`` may be dropped freely; a gap marks the slot lossy and the real
+   text arrives with the durable message.
+3. Every ``ItemStarted`` is retired by exactly one ``Durable`` or
+   ``ItemAborted``. ``item_id`` *is* the finished message's ``Event.id``, so a
+   client closes a slot on ``frame.event.id == slot.item_id``.
+4. Progress frames are never replayed.
+5. No ordering is promised between two open items.
 """
 
 from typing import Annotated, Final, Literal
@@ -41,43 +28,28 @@ from pydantic import BaseModel, ConfigDict, Field
 from openhands.sdk import Event
 
 
-# PROVISIONAL sizes. The design calls for the pending-byte cap to be *derived*
-# from the observed frame-size distribution (a multiple of the largest legal
-# frame) rather than picked. That distribution has not been measured yet, so
-# these are placeholders with the right shape and the wrong numbers. Do not
-# treat them as tuned; see the "reproductions and baselines" issue first.
+# PROVISIONAL. The cap is meant to be derived from a measured frame-size
+# distribution, which has not been done yet. Do not treat these as tuned.
 
 MAX_FRAME_BYTES: Final[int] = 4 * 1024 * 1024
-"""Largest single frame the server will emit. A frame above this drops the
-connection rather than the frame, because ``Durable`` must survive a reconnect
-and the cursor makes reconnecting lossless."""
+"""Largest frame the server will emit; a bigger one drops the connection."""
 
 MAX_PENDING_BYTES: Final[int] = 4 * MAX_FRAME_BYTES
-"""Per-connection write budget. Admission is synchronous: if a frame does not
-fit, the connection is dropped. It is never queued and the publisher never
-blocks."""
+"""Per-connection write budget. Over it, the connection is dropped."""
 
 
 class SessionFrameBase(BaseModel):
-    """Base for every frame on the session socket.
+    """Base for every frame.
 
-    Note the *absence* of ``extra="forbid"``. Unlike ``Event`` — where
-    forbidding extras is a correctness property of the durable record — the
-    envelope is meant to grow. Clients must ignore frame types and fields they
-    do not recognize, which is what lets the server add both without a version
-    bump.
+    Deliberately not ``extra="forbid"``: the envelope is meant to grow, and
+    clients must ignore what they do not recognize.
     """
 
     model_config = ConfigDict(frozen=True)
 
 
 class SyncFrame(SessionFrameBase):
-    """Sent once, before any replay, describing the range about to be sent.
-
-    A client uses ``through_seq`` to know when it has caught up with history
-    and can stop showing a loading state. ``through_seq`` is ``None`` for an
-    empty conversation.
-    """
+    """Sent once, before any replay, describing the range about to be sent."""
 
     type: Literal["sync"] = "sync"
     from_seq: int | None = Field(
@@ -90,8 +62,8 @@ class SyncFrame(SessionFrameBase):
     through_seq: int | None = Field(
         default=None,
         description=(
-            "Highest sequence number on disk at the moment the connection was "
-            "established. Everything above it arrives live."
+            "Highest sequence number on disk when the connection was "
+            "established; everything above it arrives live. None if empty."
         ),
     )
 
@@ -99,10 +71,8 @@ class SyncFrame(SessionFrameBase):
 class DurableFrame(SessionFrameBase):
     """One persisted event, after it is safely on disk.
 
-    ``event`` is the existing ``Event`` JSON, byte-for-byte what the legacy
-    endpoint sends and what the event log stores. ``seq`` is the event's index
-    in the log — already on disk as the ``{idx:05d}`` component of
-    ``event-{idx:05d}-{event_id}.json``, so exposing it needs no migration.
+    ``seq`` is its index in the log, already on disk as the ``{idx:05d}``
+    filename component, so exposing it needs no migration.
     """
 
     type: Literal["durable"] = "durable"
@@ -113,14 +83,8 @@ class DurableFrame(SessionFrameBase):
 class TransientFrame(SessionFrameBase):
     """An event that is published but never persisted.
 
-    The design's two-category model (durable vs. progress) has no home for
-    these, but the event bus genuinely carries them — ``subscribe_to_events``
-    pushes a synthetic ``ConversationStateUpdateEvent`` snapshot on connect
-    that is not in the event log. Giving them their own frame keeps ``seq`` on
-    ``DurableFrame`` honest: if a frame has a ``seq``, that sequence number is
-    real and resumable.
-
-    A client should render these but must never use them to advance its cursor.
+    Keeps ``seq`` honest: if a frame has one, it is real and resumable. A
+    client renders these but must not advance its cursor on them.
     """
 
     type: Literal["transient"] = "transient"
@@ -130,14 +94,9 @@ class TransientFrame(SessionFrameBase):
 class ItemStartedFrame(SessionFrameBase):
     """A stream is opening. Sent once per attempt, before the first token.
 
-    ``item_id`` is the ``Event.id`` the finished message will carry, minted at
-    stream open. A higher ``attempt`` for the same ``item_id`` supersedes a
-    lower one — that is how a retry announces itself, and it is the reason a
-    re-streamed retry no longer looks like new text.
-
-    ``anchor_seq`` is the sequence number the slot sits after in the
-    transcript. It exists for exactly one reason: to stop a user message that
-    lands mid-stream from splitting the bubble.
+    A higher ``attempt`` for the same ``item_id`` supersedes a lower one, so a
+    re-streamed retry does not read as new text. ``anchor_seq`` is the seq the
+    slot sits after, so a user message landing mid-stream cannot split it.
     """
 
     type: Literal["item_started"] = "item_started"
@@ -149,9 +108,8 @@ class ItemStartedFrame(SessionFrameBase):
 class DeltaFrame(SessionFrameBase):
     """One masked increment of a stream.
 
-    ``order`` is monotonic within (``item_id``, ``attempt``) and exists so a
-    client can *detect* a gap, not repair one — a gap marks the slot lossy and
-    the authoritative text arrives with the durable message.
+    ``order`` is monotonic within (``item_id``, ``attempt``) so a client can
+    detect a gap, not repair one.
     """
 
     type: Literal["delta"] = "delta"
@@ -165,9 +123,7 @@ class DeltaFrame(SessionFrameBase):
 class ItemAbortedFrame(SessionFrameBase):
     """A stream ended without producing a message.
 
-    This is what guarantees rule 3: every ``ItemStarted`` is retired. Without
-    it a cancelled or failed stream strands an open slot in the client
-    forever.
+    Without it, a cancelled or failed stream strands an open slot forever.
     """
 
     type: Literal["item_aborted"] = "item_aborted"
@@ -177,10 +133,10 @@ class ItemAbortedFrame(SessionFrameBase):
 
 
 class ErrorFrame(SessionFrameBase):
-    """A connection-level problem.
+    """A problem with this socket, never persisted.
 
-    Distinct from ``ConversationErrorEvent``, which is a durable fact about the
-    conversation. This is a fact about *this socket* and is never persisted.
+    Distinct from ``ConversationErrorEvent``, a durable fact about the
+    conversation.
     """
 
     type: Literal["error"] = "error"
--- a/openhands-agent-server/openhands/agent_server/session_protocol.py
+++ b/openhands-agent-server/openhands/agent_server/session_protocol.py
@@ -1,0 +1,224 @@
+"""Wire protocol for ``/sockets/session/{conversation_id}``.
+
+This module defines the *wire* format for the session socket. It is
+deliberately **not** built on :class:`~openhands.sdk.event.base.Event`.
+
+Why a separate hierarchy
+------------------------
+The legacy ``/sockets/events/{id}`` endpoint sends ``event.model_dump(...)``
+(``sockets.py``) while the event log persists ``event.model_dump_json(...)``
+(``event_store.py``) and the Python client decodes with ``Event.model_validate``
+(``remote_conversation.py``). One class, one serialization, three roles — and
+``Event`` is ``extra="forbid"``, so every additive wire field is a storage
+migration and every storage change is a wire break. That coupling is why a
+token delta cannot carry a stream identity today.
+
+Here the ``Event`` rides *inside* an envelope as a payload and is not modified
+in any way: the same class, the same ``extra="forbid"``, the same bytes on
+disk. Everything the protocol needs — sequence number, stream identity,
+progress — lives on the envelope, where adding a field costs nothing.
+
+Versioning
+----------
+There is no handshake and no ``protocol`` field: **the URL is the protocol
+version**. A client speaks ``/sockets/events/{id}`` or
+``/sockets/session/{id}``, never both, so there is no dual-decode path to
+maintain and no legacy client to negotiate with.
+
+Frames, and the one rule that makes them work
+---------------------------------------------
+``ItemStarted`` announces a stream and mints nothing — the ``item_id`` *is*
+the ``id`` of the ``Event`` that will eventually carry the finished message.
+So a client holding an open slot retires it on a single equality test::
+
+    frame.event.id == slot.item_id
+
+That is the entire close protocol. There is no close frame, no watermark and
+no text comparison. A stream that dies without producing a message is retired
+by ``ItemAborted`` instead; exactly one of the two always fires.
+
+Delivery rules
+--------------
+1. ``Durable`` never goes missing *across a reconnect*. Within a connection it
+   may be dropped, because the client resumes with ``after_seq`` and loses
+   nothing — that permission is what lets the server drop a slow consumer
+   instead of blocking the publisher on it.
+2. ``Delta`` may be dropped freely. A gap marks the slot lossy; the real text
+   arrives with the message.
+3. Every ``ItemStarted`` is retired by exactly one ``Durable`` or one
+   ``ItemAborted`` — including on cancellation, provider failure, policy
+   rejection and unhandled exception.
+4. Progress frames are **never replayed**. On reconnect a client throws every
+   open slot away and waits; the message is coming on the cursor regardless.
+5. No ordering is promised *between* two open items beyond the order their
+   ``ItemStarted`` frames arrived in.
+"""
+
+from typing import Annotated, Literal
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from openhands.sdk import Event
+
+
+# --------------------------------------------------------------------------
+# Size limits
+#
+# PROVISIONAL. The design calls for the pending-byte cap to be *derived* from
+# the observed frame-size distribution (a multiple of the largest legal frame)
+# rather than picked. That distribution has not been measured yet, so these are
+# placeholders with the right shape and the wrong numbers. Do not treat them as
+# tuned; see the "reproductions and baselines" issue before relying on them.
+# --------------------------------------------------------------------------
+
+MAX_FRAME_BYTES = 4 * 1024 * 1024
+"""Largest single frame the server will emit. A frame above this drops the
+connection rather than the frame, because ``Durable`` must survive a reconnect
+and the cursor makes reconnecting lossless."""
+
+MAX_PENDING_BYTES = 4 * MAX_FRAME_BYTES
+"""Per-connection write budget. Admission is synchronous: if a frame does not
+fit, the connection is dropped. It is never queued and the publisher never
+blocks."""
+
+
+class SessionFrameBase(BaseModel):
+    """Base for every frame on the session socket.
+
+    Note the *absence* of ``extra="forbid"``. Unlike ``Event`` — where
+    forbidding extras is a correctness property of the durable record — the
+    envelope is meant to grow. Clients must ignore frame types and fields they
+    do not recognize, which is what lets the server add both without a version
+    bump.
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+
+class SyncFrame(SessionFrameBase):
+    """Sent once, before any replay, describing the range about to be sent.
+
+    A client uses ``through_seq`` to know when it has caught up with history
+    and can stop showing a loading state. ``through_seq`` is ``None`` for an
+    empty conversation.
+    """
+
+    type: Literal["sync"] = "sync"
+    from_seq: int | None = Field(
+        default=None,
+        description=(
+            "Exclusive lower bound of the replay range, echoing the client's "
+            "`after_seq`. None means no history was requested."
+        ),
+    )
+    through_seq: int | None = Field(
+        default=None,
+        description=(
+            "Highest sequence number on disk at the moment the connection was "
+            "established. Everything above it arrives live."
+        ),
+    )
+
+
+class DurableFrame(SessionFrameBase):
+    """One persisted event, after it is safely on disk.
+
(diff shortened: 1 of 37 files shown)


openhands-agent-server/openhands/agent_server/session_protocol.py
```diff
@@ -0,0 +1,224 @@
+"""Wire protocol for ``/sockets/session/{conversation_id}``.
+
```
VascoSch92, line 2: make this docstring more coincise

openhands-agent-server/openhands/agent_server/session_protocol.py
```diff
@@ -0,0 +1,224 @@
...
+# placeholders with the right shape and the wrong numbers. Do not treat them as
+# tuned; see the "reproductions and baselines" issue before relying on them.
+# --------------------------------------------------------------------------
+
+MAX_FRAME_BYTES = 4 * 1024 * 1024
+"""Largest single frame the server will emit. A frame above this drops the
+connection rather than the frame, because ``Durable`` must survive a reconnect
+and the cursor makes reconnecting lossless."""
+
+MAX_PENDING_BYTES = 4 * MAX_FRAME_BYTES
```
VascoSch92: from typing import Final and use the correct type here.

openhands-agent-server/openhands/agent_server/session_protocol.py
```diff
@@ -0,0 +1,224 @@
...
+
+class SyncFrame(SessionFrameBase):
+    """Sent once, before any replay, describing the range about to be sent.
+
+    A client uses ``through_seq`` to know when it has caught up with history
+    and can stop showing a loading state. ``through_seq`` is ``None`` for an
+    empty conversation.
+    """
+
+    type: Literal["sync"] = "sync"
```
VascoSch92, line 54: do we need that?

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+from openhands.sdk import Event, Message
+from openhands.sdk.conversation.event_store import EventLog
+from openhands.sdk.event import StreamingDeltaEvent
+from openhands.sdk.event.conversation_state import ConversationStateUpdateEvent
+
+
+session_router = APIRouter(prefix="/sockets", tags=["WebSockets"])
+logger = logging.getLogger(__name__)
+
+REPLAY_PAGE_SIZE = 100
```
VascoSch92: from typing use Final[int]

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+"""Events read from disk per executor hop. The loop yields between pages so a
+long history cannot starve the event loop."""
+
+
+# --------------------------------------------------------------------------
+# One writer per connection, bounded in bytes
+# --------------------------------------------------------------------------
+
```
VascoSch92: delete that

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+
+@dataclass
```
VascoSch92: slots=True and frozen=True

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+        task = self._task
+        if task is not None and not task.done():
+            task.cancel()
+            try:
+                await task
+            except (asyncio.CancelledError, Exception):
+                pass
+
+
+# --------------------------------------------------------------------------
```
VascoSch92: delete

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+# Subscriber: buffers, then goes live
+# --------------------------------------------------------------------------
+
+
+@dataclass
```
VascoSch92: as before

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+        """Flush the buffer, dropping anything already replayed from disk."""
+        buffered, self._buffer = self._buffer, None
+        for event in buffered or ():
+            seq = self._seq_of(event)
+            if seq is not None and through_seq is not None and seq <= through_seq:
+                continue  # already sent from disk during replay
+            self._emit(event)
+
+
+# --------------------------------------------------------------------------
```
VascoSch92: delete

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+            None, _read_page, events, page_start, page_stop
+        )
+        for seq, event in page:
+            if not writer.send(DurableFrame(seq=seq, event=event)):
+                return False
+        await asyncio.sleep(0)
+    return True
+
+
+# --------------------------------------------------------------------------
```
VascoSch92: delete

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -0,0 +1,456 @@
...
+    #   * appended before subscribe -> on disk before the mark is read, so
+    #     mark >= seq and replay covers it;
+    #   * appended after subscribe  -> in the buffer; if seq <= mark it is also
+    #     replayed and the buffered copy is dropped by seq, otherwise the
+    #     buffer is the only copy and the flush delivers it.
+    #
+    # The design called for taking the state lock across both steps. It is not
+    # required, and skipping it keeps one more consumer off a FIFOLock that
+    # already guards six unrelated things.
+    try:
```
VascoSch92: Make this comment smaller

openhands-agent-server/openhands/agent_server/session_protocol.py
```diff
@@ -0,0 +1,204 @@
...
+
+# --------------------------------------------------------------------------
+# Size limits
+#
+# PROVISIONAL. The design calls for the pending-byte cap to be *derived* from
+# the observed frame-size distribution (a multiple of the largest legal frame)
+# rather than picked. That distribution has not been measured yet, so these are
+# placeholders with the right shape and the wrong numbers. Do not treat them as
+# tuned; see the "reproductions and baselines" issue before relying on them.
+# --------------------------------------------------------------------------
```
VascoSch92: Delete this

openhands-agent-server/openhands/agent_server/session_protocol.py
VascoSch92: Condensed: the module docstring went from 55 lines to ~30, keeping the "why a separate hierarchy" rationale and the numbered delivery rules (those are load-bearing spec, not narrative) and dropping the longer prose around them.

(Apologies — this reply originally landed on the wrong thread; it was meant for the `type: Literal["sync"]` question below.)

openhands-agent-server/openhands/agent_server/session_protocol.py
VascoSch92: Yes — it has to stay. `type` is the discriminator for the tagged union at the bottom of the file:

```python
SessionFrame = Annotated[SyncFrame | DurableFrame | ... , Field(discriminator="type")]
```

Pydantic needs a distinct literal on each member to pick the right variant when decoding, so dropping it (here or on any sibling frame) breaks client-side deserialization of the union.

The `= "sync"` default is load-bearing too: every construction site in `session_socket.py` relies on it rather than passing `type=` explicitly — `SyncFrame(from_seq=..., through_seq=...)`, `DurableFrame(seq=..., event=...)`, `TransientFrame(event=...)`, `ErrorFrame(code=..., detail=...)`.

Left as is.

openhands-agent-server/openhands/agent_server/session_protocol.py
VascoSch92: Done — both are now `Final[int]`:

```python
MAX_FRAME_BYTES: Final[int] = 4 * 1024 * 1024
MAX_PENDING_BYTES: Final[int] = 4 * MAX_FRAME_BYTES
```

Matches the existing convention in the package (e.g. `_PROBE_ATTEMPTS: Final[int] = 3` in `credential_binding.py`).

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Done — `REPLAY_PAGE_SIZE: Final[int] = 100`.

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Deleted (the `# One writer per connection, bounded in bytes` divider block).

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Deleted (the `# Subscriber: buffers, then goes live` divider block).

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Deleted (the `# Replay` divider block).

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Deleted (the `# The endpoint` divider block).

openhands-agent-server/openhands/agent_server/session_protocol.py
VascoSch92: Deleted the divider lines around it. Kept the note itself, since it records that the byte limits are provisional placeholders pending the step-0 measurement — it now reads as a plain comment above the constants.

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Applied: `@dataclass(frozen=True, slots=True)` (kwarg order matching the 15 other uses in the repo).

One caveat worth your call: `frozen=True` is not free here. `_ConnectionWriter` legitimately mutates `_task`, `drop_reason` and `_pending_bytes` after construction, so those four sites now go through `object.__setattr__` — the same escape hatch used elsewhere in the codebase (`agent/base.py`, `utils/cipher.py`, `event/conversation_error.py`).

A reviewer pass flagged that this buys little for a private, connection-scoped object that no external caller holds — it does not actually prevent mutation, it just routes it through a longer spelling. `slots=True` alone would give the memory win without the ceremony. Happy to drop `frozen` on this class if you agree; kept as requested for now.

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Same treatment applied to `_SessionSubscriber`: `@dataclass(frozen=True, slots=True)`.

Only one mutation site here (`go_live` clearing `_buffer`), so the `object.__setattr__` cost is much lower than on `_ConnectionWriter` — this one is a clean fit.

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Shortened from 14 lines to 4:

```python
    # Subscribe FIRST (buffering), then read the high-water mark: no lock
    # needed, since an event appended before subscribe is on disk before the
    # mark is read (replay covers it), and one appended after lands in the
    # buffer, deduped against the mark by seq once replay finishes.
```

Kept the two-case argument for why no lock is needed, dropped the restatement of the design it was contrasting with.

juanmichelini (commented):
are both frozen=True (line 59) and send() uses object.__setattr__ (line 121) needed?

VascoSch92:
> are both frozen=True (line 59) and send() uses object.**setattr** (line 121) needed?

You are right. It is useless to use `frozen` here. I removed it.

enyst (commented):
Does it work across, with agent-canvas?

VascoSch92:
> Does it work across, with agent-canvas?

@enyst No... we will need to do a migration to make it work in agent-canvas. (mostly deleting code). However, the old end-point is always present in agent-server.

juanmichelini (approved):
LGTM

enyst:
> > Does it work across, with agent-canvas?
> 
> @enyst No... we will need to do a migration to make it work in agent-canvas. (mostly deleting code). However, the old end-point is always present in agent-server.

Right, I just mean as a test of all PRs together. 🤔

VascoSch92:
> > > Does it work across, with agent-canvas?
> > 
> > 
> > @enyst No... we will need to do a migration to make it work in agent-canvas. (mostly deleting code). However, the old end-point is always present in agent-server.
> 
> Right, I just mean as a test of all PRs together. 🤔

yes.... however before releasing I would like to create integration tests for streaming because now we don't have them and this is also the reason why we do have so much bugs

PR #4808, opened 2026-09-01
feat(agent-server): explain Docker's absence in images built without it

simonrosenberg:
HUMAN:

Make a Docker-less image say so, instead of a bare "command not found".

---

AGENT:

## Why

`INSTALL_CAPABILITIES` (#4698) already lets an image be built without Docker Engine — 135.5 MB compressed, 354.2 MB on disk. Nothing publishes such an image yet, but the capability works today and the slim variant in #4805 makes it reachable.

The problem is what happens inside one. The agent runs `docker build ...` in bash and gets `docker: command not found`, which is indistinguishable from a broken PATH. So it retries, or fails the task without being able to tell the user why.

That matters more for Docker than for anything else we gate, because **Docker cannot be recovered at runtime**. Installing `docker-ce` later does not help: `dockerd` needs `--privileged` or the Sysbox runtime at *container start*, which a running container cannot grant itself. Verified:

```
failed to start daemon: Error initializing network controller:
error creating default "bridge" network: operation not permitted
```

So unlike ACP providers — which the npx fallback can fetch on demand — a message is the *entire* recovery path here.

**The primary reader is the agent, not a human.** The shim's output arrives as bash tool output, so it is written to let a model adapt ("Docker isn't available; verify the Dockerfile by inspection instead") and report the limitation accurately, rather than retry.

## Summary

When `INSTALL_CAPABILITIES` excludes `docker`, write a small `/usr/local/bin/docker` that exits 127 and explains three things: that Docker is deliberately absent, how to get an image with it, and that `--privileged` is *also* required. That last point is the wall users hit immediately after solving the first one, so pre-empting it is most of the value.

Written in the exclusion branch of the existing `case` statement, so it cannot shadow a real install.

## Verification

**Excluded build** — `--build-arg INSTALL_CAPABILITIES=vscode`:

```
$ docker ps
docker: Docker Engine is not installed in this image.
This image was built without the "docker" capability, so no Docker CLI or daemon is
present. Container builds and `docker compose` cannot work here, and Docker cannot be
installed at runtime: the daemon needs --privileged (or the Sysbox runtime) granted
when the container starts.
To use Docker: run an image built with "docker" in INSTALL_CAPABILITIES, and start the
container with --privileged.
exit=127
```

[...]

Issue #4643, opened 2026-08-26
[Feature]: Reorganise agent-server image build — selectable capabilities, parameterised provider set, and shared eval layers


simonrosenberg:
### Problem Statement

The agent-server image bundles every optional capability unconditionally. At 1.64 GB compressed, most of it is capability a given deployment never uses, and the provider set is growing. This issue makes image contents an explicit, selectable contract.

**Status: Steps 0–2 done, plus the desktop removal. Step 3 is next and unblocked apart from one product call (does the lean tag ship an editor?).**

| Step | What | Status |
|---|---|---|
| 0–1 | Provider payload built once, `COPY --from` into `base-image-minimal` so the layer is parent-independent | ✅ #4651 |
| 2 | `INSTALL_ACP_PROVIDERS` and `INSTALL_CAPABILITIES` build arguments, defaults unchanged | ✅ #4687, #4698 |
| 3 | Publish lean tags; default OSS and Agent Canvas to them | tag **published**; Canvas PR open — OpenHands/OpenHands#17079, ⛔ **blocked on an SDK release** |
| 4 | Make the lazy `npx` path solid (cache persistence, `--prefer-offline`, progress) | ✅ #4805 |
| 5 | Provider additions adopt the mechanism | ongoing |
| — | Desktop/VNC stack removed entirely (not just made optional) | ✅ #4792 |

Steps 0–2 changed nothing observable: same providers, same pins, same tag contents. Everything is now selectable; nothing published yet exercises the option. **Step 3 is the first step where anything moves.**

> [!NOTE]
> **First measurement of the published images (amd64), after #4805 merged.** Compressed layer totals from the ghcr manifests — what a cold pull transfers:
>
> | Tag | Compressed |
> |---|---:|
> | `agent-server:latest-python` | 1279.2 MB |
> | **`agent-server:latest-python-slim`** | **954.8 MB** |
> | saving | **324.4 MB (25.4%)** |
>
> Consistent with the arm64 A/B figure of 315.3 MB — Codex ships a larger `linux-x64` binary. Published slim contents verified: `/acp-node` empty, all three wrappers absent, `node v22.23.2` and `npx` present for the fallback, Chromium/Docker/OpenVSCode intact.
>
> **Still reaching no user.** Agent Canvas adoption is open as OpenHands/OpenHands#17079 (reviewed: *"well-scoped and correct, no substantive correctness or architecture issues"*), but it cannot merge — see **The one thing blocking Step 3** below.

[...]

PR #4809, opened 2026-09-01
feat(agent-server): drop Docker Engine from the python-slim variant

simonrosenberg:
HUMAN:

Drops Docker Engine from the `python-slim` variant only — 129.7 MB that provably cannot function in the launch we document for Canvas and OSS. Draft until its two dependencies land; the default `python` tag is untouched.

---

AGENT:

> [!WARNING]
> **Draft — stacked on #4805, and should not merge before #4808.**
> - Base is `acp-version-parity` (#4805), which introduces the `python-slim` variant this modifies. There is nothing to change on `main` yet.
> - **#4808 should land first.** It adds the `docker` shim that explains the absence. Without it a slim user gets a bare `docker: command not found` — confirmed on this branch, since #4808 is not in this stack. Merging this first would ship the regression without the mitigation.
>
> Retarget to `main` once #4805 merges.

## Why

`dockerd` cannot start in an unprivileged container:

```
failed to start daemon: Error initializing network controller:
error creating default "bridge" network: operation not permitted
```

And the documented Canvas/OSS launch (`docs/openhands/usage/agent-canvas/backend-setup/docker.mdx`) is a plain `docker run -p 8000:8000 -v … agent-canvas:latest` — no `--privileged`, no `runtime:`, no socket mount. Nothing in Canvas's own code uses the Docker CLI either: no `DockerWorkspace` usage, no server-side shelling out, no socket in its compose examples.

So for the audience the slim variant targets, this is 129.7 MB of payload that cannot be used.

**Be precise about what this is, though.** With `--privileged`, Docker *does* work — verified against published `latest-python`, daemon 29.7.2 starts normally. So this is **a real regression for an undocumented-but-working configuration**, not the removal of dead code. Unlike ACP providers there is no runtime fallback: privilege is granted at container start, so nothing can be installed later. That is why #4808's shim is a hard prerequisite rather than a nicety, and why this needs release notes.

## Summary

Adds a per-matrix `capability_flavor` that strips `docker` from `INSTALL_CAPABILITIES` for `python-slim` only, mirroring the existing `acp_provider_flavor` mechanism exactly — same matrix-key shape, same resolution in the `prep` step, same `steps.prep.outputs.*` plumbing into the build arg.

One workflow file, no Dockerfile change, no new build argument.

## Verification

**Resulting slim image** — built with `INSTALL_ACP_PROVIDERS=` and `INSTALL_CAPABILITIES=vscode,browser`:

```
docker binary:  ABSENT      <- bare "command not found" without #4808
dockerd:        ABSENT
chromium:       /usr/bin/chromium
openvscode:     PRESENT
/acp-node:      0 entries
npx:            /usr/local/bin/npx      <- the ACP fallback still works
```

**Size**, `--target base-image`, arm64, `--output type=oci`, summed compressed `layers[].size`:

| Build | Compressed |
|---|---:|
| today's default (ACP + all capabilities) | 1133.6 MB |
| slim, with Docker (#4805 as it stands) | 818.3 MB |
| **slim, without Docker (this PR)** | **688.6 MB** |

−129.7 MB against #4805, and **−39% against today's default**.

**The `docker`-stripping expression is position-independent** — the fragile part, so tested directly:

```
'vscode,browser,docker'  -> 'vscode,browser'
'docker,vscode,browser'  -> 'vscode,browser'
'vscode,docker,browser'  -> 'vscode,browser'
'docker'                 -> ''
'vscode,browser'         -> 'vscode,browser'   (no-op when absent)
''                       -> ''
```

**The default variant is unaffected by construction**: `capability_flavor` is set only on the two `python-slim` matrix entries, and the `prep` step falls through to `$INSTALL_CAPABILITIES` unchanged for every other variant. The `workflow_dispatch` override path is untouched.

## How to Test

```bash
# The slim variant as the workflow will now build it: no ACP, no Docker
docker buildx build -f openhands-agent-server/openhands/agent_server/docker/Dockerfile \
  --target base-image \
  --build-arg BASE_IMAGE=nikolaik/python-nodejs:python3.13-nodejs22-slim \
  --build-arg INSTALL_ACP_PROVIDERS= \
  --build-arg INSTALL_CAPABILITIES=vscode,browser \
  --load -t local/slim:nodocker .

docker run --rm --entrypoint /bin/sh local/slim:nodocker -c '
  echo "docker:     $(command -v docker || echo ABSENT)"
  echo "dockerd:    $(command -v dockerd || echo ABSENT)"
  echo "chromium:   $(command -v chromium || echo ABSENT)"
  echo "npx:        $(command -v npx || echo ABSENT)"'

# The default variant must be unaffected — Docker still present and working
docker buildx build -f openhands-agent-server/openhands/agent_server/docker/Dockerfile \
  --target base-image \
  --build-arg BASE_IMAGE=nikolaik/python-nodejs:python3.13-nodejs22-slim \
  --load -t local/slim:default .
docker run --rm --entrypoint /bin/sh local/slim:default -c 'docker --version'

# Compressed sizes, compared the same way
docker buildx build ... --output type=oci,dest=/tmp/slim.tar .
python3 -c "import json,tarfile; ..."   # sum layers[].size from the manifest

# The docker-stripping expression, in every position
strip() { echo "$1" | sed -E 's/(^|,)docker(,|$)/\1/; s/,$//'; }
for v in "vscode,browser,docker" "docker,vscode,browser" "vscode,docker,browser" \
         "docker" "vscode,browser" ""; do printf '%-24s -> %s\n' "'$v'" "'$(strip "$v")'"; done
```

Output from all of the above is quoted under **Verification**.

## Before merging

- [ ] #4805 merges, then retarget this to `main`
- [ ] #4808 merges, so absence is explained rather than a bare `command not found`
- [ ] Release notes: any OSS user running Canvas with `--privileged` for docker-in-sandbox loses it on the slim tag and must use the full tag
- [ ] Re-measure on amd64 — every figure here is arm64

## Deliberately out of scope

- **The default `python` tag.** Enterprise, Replicated, cloud and evaluation all consume it, and docker-in-sandbox is a marketed feature there: `saas-deploy` sets `RUNTIME_CLASS: "sysbox-runc"` in production, staging and development; Sysbox is installed via ArgoCD in every environment; Replicated exposes "Sandbox Isolation" as a customer-facing option defaulting to sysbox.
- Removing the Docker install block from the Dockerfile. It stays; only the slim variant declines it.
- Canvas adoption — different repository, and blocked on a published slim tag.

## Issue Number

Relates to #4643

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -441,10 +441,8 @@ jobs:
                       echo "install_acp_providers=$INSTALL_ACP_PROVIDERS" >> $GITHUB_OUTPUT
                   fi
 
-                  # Docker Engine is dropped from the slim variant: dockerd cannot
-                  # start without --privileged (or Sysbox) granted at container
-                  # start, which the documented Canvas/OSS launch does not pass.
-                  # Images built without it carry a `docker` shim that says so.
+                  # dockerd needs privilege granted at container start, which the
+                  # documented Canvas/OSS launch does not pass, so slim omits it.
                   if [ "${{ matrix.capability_flavor }}" = "no-docker" ]; then
                       echo "install_capabilities=$(echo "$INSTALL_CAPABILITIES" | sed -E 's/(^|,)docker(,|$)/\1/; s/,$//')" >> $GITHUB_OUTPUT
                   else
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -305,6 +305,7 @@ jobs:
                       custom_tags: python
                       image_flavor: slim
                       acp_provider_flavor: none
+                      capability_flavor: no-docker
                       arch: amd64
                       base_image: nikolaik/python-nodejs:python3.13-nodejs22-slim
                       runner: ubuntu-24.04
@@ -314,6 +315,7 @@ jobs:
                       custom_tags: python
                       image_flavor: slim
                       acp_provider_flavor: none
+                      capability_flavor: no-docker
                       arch: arm64
                       base_image: nikolaik/python-nodejs:python3.13-nodejs22-slim
                       runner: ubuntu-24.04-arm
@@ -439,6 +441,16 @@ jobs:
                       echo "install_acp_providers=$INSTALL_ACP_PROVIDERS" >> $GITHUB_OUTPUT
                   fi
 
+                  # Docker Engine is dropped from the slim variant: dockerd cannot
+                  # start without --privileged (or Sysbox) granted at container
+                  # start, which the documented Canvas/OSS launch does not pass.
+                  # Images built without it carry a `docker` shim that says so.
+                  if [ "${{ matrix.capability_flavor }}" = "no-docker" ]; then
+                      echo "install_capabilities=$(echo "$INSTALL_CAPABILITIES" | sed -E 's/(^|,)docker(,|$)/\1/; s/,$//')" >> $GITHUB_OUTPUT
+                  else
+                      echo "install_capabilities=$INSTALL_CAPABILITIES" >> $GITHUB_OUTPUT
+                  fi
+
                   # Verify outputs
                   echo "=== Build outputs ==="
                   echo "Build context: $(grep '^build_context=' $GITHUB_OUTPUT | cut -d= -f2-)"
@@ -464,7 +476,7 @@ jobs:
                       OPENHANDS_BUILD_GIT_SHA=${{ env.SDK_SHA }}
                       OPENHANDS_BUILD_GIT_REF=${{ env.SDK_REF }}
                       INSTALL_ACP_PROVIDERS=${{ steps.prep.outputs.install_acp_providers }}
-                      INSTALL_CAPABILITIES=${{ env.INSTALL_CAPABILITIES }}
+                      INSTALL_CAPABILITIES=${{ steps.prep.outputs.install_capabilities }}
 
             - name: Cleanup build context
               if: always()


VascoSch92 (approved):

simonrosenberg:
@enyst — before this merges, I researched whether we can keep the Docker capability at a fraction of the size instead of dropping it. Measured on arm64, compressed, per package subset:

| Config | Docker cost | Saving |
|---|---:|---:|
| full suite (today) | 116.3 MB | — |
| drop `compose` | 102.0 MB | 14.3 MB (12%) |
| **drop `compose` + `buildx`** | **80.2 MB** | **36.1 MB (31%)** |
| `docker-ce-cli` only | 18.6 MB | 97.6 MB (84%) |

**The 84% option isn't available.** CLI-only means no daemon, so you'd mount the host's `/var/run/docker.sock` — the agent then drives the *host* daemon and can escape the sandbox. `docs/enterprise/docker-in-sandbox.mdx` explicitly rejects both that and privileged DinD.

**No lighter runtime escapes the privilege requirement either.** Rootless BuildKit unprivileged: `rootlesskit: fork/exec /proc/self/exe: operation not permitted`. With `seccomp=unconfined --security-opt apparmor=unconfined` it gets further — daemon starts, base image pulls — then `RUN` fails: `error mounting "proc" to rootfs: operation not permitted`. Nested containerization needs `CAP_SYS_ADMIN` for mounts; that's what Sysbox provides at the *host* level and can't be shipped in an image.

**But the 31% trim keeps build-and-run, verified:**

```
daemon+cli+containerd (no buildx):  build: OK   run: OK-BUILT   buildx: absent
+ buildx:                            build: OK   run: OK-BUILT   buildx: present
```

Docker 29.7.2's daemon still serves `docker build` via the legacy builder when the buildx plugin is absent. (Both configs first failed identically on nested overlayfs — that was Docker Desktop's overlay-on-overlay limitation, not buildx; `--storage-driver=vfs` cleared it for both.)

**So there's an alternative to this PR worth considering:** trim the suite in the **default** image rather than dropping Docker from slim.

- `docker-ce-rootless-extras` (23.5 MB on disk) is free — we require `--privileged`, so rootless mode is unused. No capability loss.
- `compose` (14.3 MB) is a judgement call — is "bring up a compose stack" worth keeping?
- `buildx` (21.8 MB) I'd keep; legacy builds work but BuildKit is materially better for that price.

Smaller win than this PR (36 MB vs 116 MB) but **no regression, no shim dependency, no migration window** — and it also benefits enterprise, which pays the full 116.3 MB today and actually uses the feature.

Happy to go either way. This PR is still correct as written and CI-green; the question is whether losing docker-in-sandbox for `--privileged` Canvas users is worth the extra 80 MB. Context and the capability-loss detail: https://github.com/OpenHands/software-agent-sdk/issues/4643#removing-docker-from-canvas

simonrosenberg:
Closing — Docker-in-Docker stays in the slim image.

Product decision: being able to run the agent inside Docker *and* have it run Docker and spin up containers is a capability worth keeping, so trading it for ~116 MB is the wrong call.

The research that led here is in the comment above. Summary of what was established:

- **No lighter runtime avoids the privilege requirement.** Rootless dockerd with `seccomp=unconfined`, `apparmor=unconfined`, `--device /dev/fuse` and `--device /dev/net/tun` gets the daemon fully started ("Daemon has completed initialization", server 29.7.2) — but container execution then fails on `error mounting "proc" to rootfs: operation not permitted`, identically with BuildKit and the legacy builder, and with both `overlayfs` and `vfs`. `runc` needs `CAP_SYS_ADMIN` to mount `/proc`, which no narrower grant supplies. Same wall for rootless BuildKit.
- So `--privileged` (or Sysbox at host level) remains the only working path, and there is no safer middle option to document.
- **A trim is available instead**, without losing the capability: dropping `compose` + `buildx` saves 36.1 MB (31%) and `docker build` / `docker run` still work (verified — the daemon serves the legacy builder when buildx is absent). `docker-ce-rootless-extras` also looks like dead weight given the above.

Follow-ups worth doing, tracked in #4643:
- trim the Docker package set in the **default** image, which benefits enterprise too — it pays the full 116.3 MB today and actively uses the feature
- document `--privileged` for Canvas: the capability is currently undiscoverable for OSS users, since `docker-in-sandbox.mdx` lives only under `enterprise/`

PR #4813, opened 2026-09-01
fix(agent-server): own confirmation_policy, security_analyzer, secrets in ConversationState

### LINKED ISSUE #4810: [Bug]: confirmation_policy, security_analyzer, and secrets mutations not persisted to meta.json
### Bug Description

When `set_confirmation_policy`, `set_security_analyzer`, or `update_secrets` are called via the API, the change is written to `ConversationState`/`base_state.json` but `StoredConversation`/`meta.json` is never updated. On next startup, `event_service.py` lines 1090–1102 read all three fields from `stored` to initialize the conversation, so the changes are silently reverted after eviction or server restart.

### Expected Behavior

API mutations to `confirmation_policy`, `security_analyzer`, and `secrets` persist across conversation eviction and server restart.

### Actual Behavior

Changes are lost; the conversation reverts to the values set at creation time. This can be verified by running the existing test suite and adding a test that:

1. Creates a conversation
2. Calls the mutation endpoint (e.g. `set_confirmation_policy`)
3. Simulates eviction by removing the event service from `_event_services`
4. Reloads the conversation via `_get_or_load_event_service`
5. Asserts the updated value is still present

```bash
pytest tests/agent_server/ -k "confirmation_policy or security_analyzer or secrets"
```

The relevant code paths are:

- `EventService.set_confirmation_policy` (line 1668) — only calls `self._conversation.set_confirmation_policy(policy)`; no `self.stored` update, no `save_meta()`
- `EventService.set_security_analyzer` (line 1677) — same pattern
- `EventService.update_secrets` (line 1658) — only calls `self._conversation.update_secrets(secrets)`; no `self.stored` update, no `save_meta()`

On startup, `event_service.py` reads all three from `stored`:
```python
# line 1090
secrets=self.stored.secrets,
# line 1101
conversation.set_confirmation_policy(self.stored.confirmation_policy)
# line 1102
conversation.set_security_analyzer(self.stored.security_analyzer)
```

Compare to `EventService.apply_resume_secrets` (lines 352–355) which correctly updates `self.stored` and calls `await self.save_meta()` after mutating state.

### Acceptance Criteria

- [ ] `set_confirmation_policy` updates `self.stored.confirmation_policy` and calls `save_meta()` after mutating `ConversationState`
- [ ] `set_security_analyzer` updates `self.stored.security_analyzer` and calls `save_meta()` after mutating `ConversationState`
- [ ] `update_secrets` updates `self.stored.secrets` and calls `save_meta()` after mutating `ConversationState`
- [ ] Tests verify each change survives a simulated eviction/reload

### Additional Context

Discovered while reviewing [#4617](https://github.com/OpenHands/software-agent-sdk/pull/4617). `tags` is currently the only field where API mutations correctly write back to `stored` and persist to `meta.json`. Discussed in #proj-agent-server.

BSmick6:
HUMAN:
Moving `confirmation_policy`, `security_analyzer`, and `secrets` out of `StoredConversation` so that they're only in `ConversationState` which makes `base_state.json` the only place where the values are stored, eliminating the need to sync to `meta.json`.

One side effect is that we need to move these values down to `StartConversationRequest` as init values to be manually stored in the state and passed around in a few other places. These values now behave much like the agent specification values in this way. In fact, various comments were updated to include these "init-only fields" that need to be manually passed/set along with the agent spec values.

---

AGENT:

## Why

`set_confirmation_policy`, `set_security_analyzer`, and `update_secrets` mutations wrote only to `ConversationState` but never back to `StoredConversation`/`meta.json`. On eviction or server restart, `event_service.start()` re-initialised from `stored`, silently reverting all mutations. Fixes #4810.

## Summary

This PR follows the same structural pattern as #4440 (which removed `agent` from `ConversationConfig`): `confirmation_policy`, `security_analyzer`, `secrets`, and `secrets_encrypted` are moved out of `ConversationConfig` / `StoredConversation` entirely and added to `StartConversationRequest` and `EventService` as init-only fields. This makes it structurally impossible for `meta.json` to carry stale copies — there is no field to serialize.

Key changes:

- **`ConversationConfig`**: remove `confirmation_policy`, `security_analyzer`, `secrets`, `secrets_encrypted`. Docstring updated to explain the boundary between "stored" and "init-only" fields.
- **`StartConversationRequest`**: add those four fields with a doc comment explaining they are init-only (seeded into `ConversationState` on first start, never written to `meta.json`).
- **`StoredConversation`**: remove the `Field(exclude=True)` overrides (fields are gone entirely by construction).
- **`EventService`**: add `confirmation_policy`, `security_analyzer`, `secrets` as dataclass fields; replace `_without_stored_secret` (which returned a mutated `StoredConversation`) with inline `self.secrets.pop(secret_name, None)` calls.
- **`conversation_service._start_conversation`**: pop init-only fields from `request_data` before constructing `StoredConversation`; handle the `secrets_encrypted` path with `TypeAdapter`; pass all three to `_start_event_service`.
- **`conversation_service._start_event_service`**: accept `confirmation_policy`, `security_analyzer`, `secrets` params and forward to `EventService(...)`.
- **`conversation_service._get_or_load_event_service_locked`**: accept `secrets` param for the CODEX fallback-secret injection path, replacing the previous `stored` mutation/restore pattern.
- **`_build_telemetry_context`**: accept `confirmation_policy` param; caller reads it from the live `ConversationState` instead of stale `stored`.
- **`LocalConversation.fork()`**: copy `confirmation_policy` and `security_analyzer` from the source `ConversationState` to the fork's — these were not previously copied, so forks silently started with `NeverConfirm` regardless of the source conversation's policy.

## Issue Number

#4810

## How to Test

Run the agent-server and SDK test suites:

```
uv run pytest tests/agent_server/ tests/sdk/conversation/local/test_fork.py -v
```

All 2109 tests pass. The `TestMutationsOwnedByConversationState` class in `test_event_service.py` covers the acceptance criteria directly: meta.json exclusion, delegation to `ConversationState`, and the resume guard.

## Video/Screenshots

N/A — pure backend logic, covered by tests.

## Design Doc

N/A

## Type

- [x] Bug fix
- [x] Refactor

## Notes

The fork path seeds `EventService` with default values for `confirmation_policy`, `security_analyzer`, and `secrets` — which is correct because `LocalConversation.fork()` now copies those fields into the fork's `ConversationState` before writing `base_state.json`. The `if not base_state_exists:` guard in `EventService.start()` then correctly skips re-initialisation, using the values already in `base_state.json`.


enyst (commented):
Thank you so much for this, @BSmick6 ! I’m having the agent take an initial look to see if it can spot issues.

BSmick6:
sorry, i had to fix one more CI check. i figured i probably shouldn't have to request review from the bot again. everything should pass now.

alanhuangyoo:
I had #4859 open against the same issue and closed it — this is the better fix, taking the fields off `StoredConversation` instead of syncing two copies. @enyst pointed me here.

Since the reproduction was already set up, I checked two things this depends on rather than just agreeing. Both hold.

**Secrets actually reach `base_state.json` now.** On `main` they do not: the live registry has the value but the persisted one is empty, because `update_secrets` mutated the registry in place and nothing marked the state dirty.

```
main                                 this PR
live registry keys ['MY_TOKEN']      live registry keys ['MY_TOKEN']
persisted secret_sources {}          persisted secret_sources {"MY_TOKEN": {...}}
'MY_TOKEN' in base_state.json False  'MY_TOKEN' in base_state.json True
```

That is what the `model_copy` + reassignment under `with self._state:` buys, and it matters more than it might look: with the field gone from `meta.json`, an in-place mutation would have left API-set secrets with nowhere to persist at all. The stored value is the Fernet ciphertext, not plaintext — `'shh' in base_state.json` is `False`.

**The upgrade path keeps existing conversations.** The case I wanted to rule out is a conversation created before this change, whose three values live in `meta.json`, loaded after the field is dropped from the model. Created on `main` with all three set at creation time, then the same directory loaded on this branch:

```
main, meta.json      confirmation_policy {"kind":"AlwaysConfirm"}
                     security_analyzer   {"kind":"LLMSecurityAnalyzer"}
                     secrets             {"MY_TOKEN": {"value":"gAAAAABqm5-aM906...
this branch, loaded  confirmation_policy AlwaysConfirm
                     security_analyzer   LLMSecurityAnalyzer
                     secrets             ['MY_TOKEN']  -> 'shh'
```

Nothing is lost, because creation-time values already reach `base_state.json` through the conversation constructor. Worth knowing that this is what carries the migration rather than any explicit step — a conversation whose `base_state.json` were missing them would have nothing to fall back on once `meta.json` stops carrying them.

One note in passing, from having gone the other way first: `model_copy(update=...)` does not validate, so the old sync could put a plain `str` into `stored.secrets` (typed `dict[str, SecretSource]`) and only fail later at serialisation. That hazard goes away entirely with the field, which is another point for this approach over mine.

enyst (commented):
Thank you so much for this PR, @BSmick6 !

I took a look and seems great to me! My only two thoughts were:
- it might be that the scope is a bit large, or it would have been, in the Old Times of 2024 when we were writing code by hand 😅; my reflex was to wonder if we should separate one by one, but today… I think agents can do it all 🙏
- we have a ‘cross’ tests directory that we use for conversation restore tests, in more realistic scenarios than pure mocks; it would be nice to have some for this behavior.

Then I asked Astra to review, and it claims it found some missing behavior. I told it to make a PR with tests to your PR, let’s see what it comes up with.

AGENT:

I would fix these before merging. I reviewed commit a4e7a74 and found five reproducible issues:

1. P1 — New secrets can leak through ACP streaming.
    [update_secrets()](https://github.com/OpenHands/software-agent-sdk/blob/a4e7a74c0fb8f989a2c58ca1717c176f0aa1aaff/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py#L2753)⁠￼ replaces the registry, but ACP’s streaming masker remains bound to the original registry. After adding a dummy secret, an incoming chunk containing it reached the token callback in plaintext; the base commit masked it. Keep the masker connected to the current secrets while preserving autosave.
2. P2 — Server forks lose conversation secrets.
    [The fork path](https://github.com/OpenHands/software-agent-sdk/blob/a4e7a74c0fb8f989a2c58ca1717c176f0aa1aaff/openhands-agent-server/openhands/agent_server/conversation_service.py#L2055)⁠￼ assumes secrets are already in the fork’s base_state.json, but LocalConversation.fork() never copies the secret registry. A token supplied through StartConversationRequest.secrets exists in the parent and disappears in the fork. Previously, the copied metadata supplied it. Copy the parent’s current registry into the fork.
3. P2 — Forks discard guardrail API keys.
    [Copying the analyzer](https://github.com/OpenHands/software-agent-sdk/blob/a4e7a74c0fb8f989a2c58ca1717c176f0aa1aaff/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py#L892)⁠￼ triggers persistence, but the fork was constructed without the parent’s cipher. Its analyzer credentials are therefore redacted. Through the real server fork path, a ToolShield analyzer’s configured API key became None; the base commit preserved it. Pass the cipher before writing the fork’s state.
4. P2 — Encrypted analyzer credentials bypass decryption.
    [Conversation startup](https://github.com/OpenHands/software-agent-sdk/blob/a4e7a74c0fb8f989a2c58ca1717c176f0aa1aaff/openhands-agent-server/openhands/agent_server/conversation_service.py#L1787)⁠￼ forwards request.security_analyzer directly after removing it from the payload validated with the cipher. With secrets_encrypted=True, the actor’s key decrypts correctly, but ToolShield’s key remains ciphertext until restart. Validate the analyzer with the cipher too.
5. P2 — Direct SDK forks share mutable analyzer state.
    [The analyzer assignment](https://github.com/OpenHands/software-agent-sdk/blob/a4e7a74c0fb8f989a2c58ca1717c176f0aa1aaff/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py#L892)⁠￼ copies the object reference. ToolShield maintains action history, so subsequent fork actions contaminate the parent’s risk assessment. I reproduced a fork-only command appearing in the parent’s guardrail prompt. Give the fork an independent analyzer instance.

Validation: Separate probes confirmed the regressions and verified that the intended restart fix works for policy, analyzer, and secrets. These used fake credentials and stubbed provider execution; no live provider calls were made.

PR #4814, opened 2026-09-01
fix(sdk): remove title from LocalConversation.fork() — store it on StoredConversation instead

### LINKED ISSUE #4811: [Bug]: fork sets title by injecting into state.tags instead of a dedicated field
### Bug Description

In `local_conversation.py` (line 876–881), when forking a conversation with a title, the title is set by injecting it into `ConversationState.tags` as `tags["title"] = title`. This conflates a display/metadata field with the user-defined tag map and means a forked conversation's tags will always contain a `"title"` key, which leaks into tag-based queries, observability span attributes, and any downstream consumer that iterates over tags.

```python
# local_conversation.py:876
# Copy title via tags if provided
if title is not None:
    fork_conv._state.tags = {
        **fork_conv._state.tags,
        "title": title,
    }
```

This appears to be a workaround for `ConversationState` not having a dedicated `title` field — the agent server handles fork titles separately via `StoredConversation.title`.

### Expected Behavior

A forked conversation's title is stored in a dedicated field and does not appear as a tag in the conversation's tag map.

### Actual Behavior

Forking with a title injects `{"title": "<value>"}` into `state.tags`, polluting the tag map with a non-user-defined key. Reproduce with:

```python
python -c "
# fork a conversation with a title and inspect state.tags
# tags will contain {'title': 'My Fork'} alongside any user-defined tags
"
```

Or via the existing test suite:

```bash
pytest tests/ -k "fork" -v
```

### Steps to Reproduce

1. Fork a conversation with a title
2. Inspect the forked conversation's `state.tags`
3. Observe that `"title"` appears as a key in the tag map alongside user-defined tags

### Acceptance Criteria

- [ ] Forking with a title does not inject a `"title"` key into `state.tags`
- [ ] The fork title is stored in a dedicated field or passed through a mechanism that does not pollute the tag map
- [ ] Existing fork behavior in the agent server (`StoredConversation.title`) is unaffected

### Additional Context

Discovered while investigating the boundary between `ConversationState` and `StoredConversation` (see #4810 and discussion in #proj-agent-server). `ConversationState` has no `title` field, which is why this workaround exists. The fix may require deciding where title lives in the standalone SDK context.

BSmick6:
HUMAN:
Inspired by some digging I did for #4617. Turns out there are some duplicate values between `StoredConversation` and `ConversationState`. I discovered this because `tags` is one of them. So I had to worry about syncing, which was actually an already existing issue in `update_conversation`.

Anyway, while considering whether `tags` should be pulled from `ConversationState`, I discovered that, when forking with a title, the title is written into the forked conversation's `_state.tags`. On its own, that's worth fixing.

In this change, it may look like the title is being removed completely, but the agent explains below how the functionality is preserved via `fork_overrides`.

---

AGENT:

## Why

Forking a `LocalConversation` with a title injected `{"title": "<value>"}` into `state.tags`, polluting the user-defined tag map. This conflated a display/metadata field with user-defined tags and leaked into tag-based queries and any downstream consumer iterating over tags.

`ConversationState` has no `title` field (title belongs on `StoredConversation` in the agent-server layer). Rather than add a duplicate field, the `title` parameter on `LocalConversation.fork()` is deprecated: it is still accepted to avoid a breaking API change, but emits a `DeprecationWarning` and no longer writes into `state.tags`.

## Summary

- Deprecates `title` on `LocalConversation.fork()` with a `DeprecationWarning` — the parameter is kept for backwards compatibility but has no effect; `ConversationState` has no title field and adding one would duplicate `StoredConversation`
- Removes the `tags["title"]` injection workaround from the fork body (the fix for the actual bug)
- Drops the `title=title` passthrough in `conversation_service.py` — the agent server applies title via `fork_overrides` on `StoredConversation` independently, so passing it to `fork()` would only trigger the deprecation warning unnecessarily
- `RemoteConversation.fork(title=...)` is **unaffected** — it routes through the agent server which owns `StoredConversation`

## Issue Number

Fixes #4811

## How to Test

```bash
uv run pytest tests/sdk/conversation/local/test_fork.py tests/sdk/conversation/remote/test_remote_fork.py -v
```

All 23 tests pass. `test_fork_with_title_is_deprecated` verifies that `title=` emits a `DeprecationWarning` and that `"title"` does not appear in `state.tags`.

For the agent-server path, title on a forked conversation is set via `fork_overrides["title"]` in `conversation_service.py` — that path is unchanged.

## Video/Screenshots

N/A — pure logic change with no UI impact.

## Design Doc

N/A — follow-up discussion on whether the standalone SDK should eventually surface fork titles (e.g. a `LocalConversation.title` property backed by a proper field) is tracked on #4811.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The `title` parameter remains fully functional on `RemoteConversation.fork()` since the agent server handles it correctly via `StoredConversation`. An intentional eventual removal of the deprecated parameter from `LocalConversation.fork()` should follow the SDK's deprecation policy.

### DIFF
--- a/examples/01_standalone_sdk/48_conversation_fork.py
+++ b/examples/01_standalone_sdk/48_conversation_fork.py
@@ -47,13 +47,12 @@ print(f"Source events count    : {len(source.state.events)}")
 # =================================================================
 # 2. Fork and continue independently
 # =================================================================
-fork = source.fork(title="Follow-up fork")
+fork = source.fork()
 source_event_count = len(source.state.events)
 
 print("\n--- Fork created ---")
 print(f"Fork ID                : {fork.id}")
 print(f"Fork events (copied)   : {len(fork.state.events)}")
-print(f"Fork title             : {fork.state.tags.get('title')}")
 
 assert fork.id != source.id
 assert len(fork.state.events) == source_event_count
@@ -80,7 +79,6 @@ alt_agent = Agent(llm=alt_llm, tools=[Tool(name=TerminalTool.name)])
 
 fork_alt = source.fork(
     agent=alt_agent,
-    title="Tool-change experiment",
     tags={"purpose": "a/b-test"},
 )
 
@@ -104,7 +102,7 @@ cut_event_id = next(
     if isinstance(e, MessageEvent) and e.source == "user"
 )
 
-fork_slice = source.fork(from_event_id=cut_event_id, title="Branch from first turn")
+fork_slice = source.fork(from_event_id=cut_event_id)
 
 print("\n--- Branch-slice fork (from_event_id) ---")
 print(f"Cut event id          : {cut_event_id}")
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -1986,10 +1986,11 @@ class ConversationService:
         source_conversation = source_service.get_conversation()
 
         # fork() deep-copies events, state, and writes to a new persistence dir.
+        # Title is applied to StoredConversation via fork_overrides below — not
+        # passed to fork() since LocalConversation has no title field.
         fork_conv = await asyncio.to_thread(
             source_conversation.fork,
             conversation_id=fork_id,
-            title=title,
             tags=tags,
             reset_metrics=reset_metrics,
             from_event_id=from_event_id,
--- a/openhands-sdk/openhands/sdk/conversation/base.py
+++ b/openhands-sdk/openhands/sdk/conversation/base.py
@@ -429,7 +429,10 @@ class BaseConversation(ABC):
                 if ``None``).
             agent: Agent for the fork. Defaults to a deep-copy of the
                 source agent.
-            title: Optional title for the forked conversation.
+            title: Deprecated on ``LocalConversation`` since 1.44.1 — has no
+                effect there and emits a ``DeprecationWarning``. Still
+                honoured by ``RemoteConversation`` (passed to the agent
+                server which stores it on ``StoredConversation``).
             tags: Optional tags for the forked conversation.
             reset_metrics: If ``True`` (default), cost/token stats start
                 fresh on the fork.
--- a/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
+++ b/openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py
@@ -4,7 +4,6 @@ import contextlib
 import copy
 import json
 import uuid
-import warnings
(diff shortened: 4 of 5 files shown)


enyst (approved):
Thank you!

enyst:
@openhands why is the Python API check failing in CI? Is the workflow correct we should worry, or could it be more lenient?

enyst:
Hi — I’m **OpenHands-GPT-5.6, enyst’s AI agent**.

The check is failing for a real and very specific reason: Griffe sees `LocalConversation.fork(title)` disappear from the published `openhands-sdk` API. Removing that keyword means existing standalone-SDK code that calls `conversation.fork(title="...")` will now raise `TypeError`. The [job annotation](https://github.com/OpenHands/software-agent-sdk/actions/runs/33544984395/job/100030894998) is:

> `LocalConversation.fork(title)`: Parameter was removed  
> Breaking changes detected (1); require at least minor version bump from 1.44.x, but new is 1.44.1

So the **detection is correct**, and I would not make the detector ignore this case. The bug fix should stop writing the title into `state.tags`, but that does not require removing the accepted keyword immediately. The least disruptive fix for this PR is to retain `title: str | None = None` on `LocalConversation.fork()` and no longer write it into tags (optionally deprecating it/documenting that local conversations do not persist titles). That fixes the tag pollution without breaking callers. An intentional eventual removal should follow the SDK's deprecation policy rather than being classified as a patch-level bug fix.

There is, however, a **workflow-policy mismatch worth making more lenient**:

- [#2173](https://github.com/OpenHands/software-agent-sdk/pull/2173) originally made API failures non-blocking on ordinary PRs and blocking only on `rel-*` PRs.
- [#2432](https://github.com/OpenHands/software-agent-sdk/pull/2432) later made them visibly red on every PR, while explicitly saying they remain optional unless branch protection requires them.
- `gh pr checks 4814 --required` confirms **Python API is not a required check** here; only pre-commit is currently required and failing.
- Nevertheless, the review bot treats any red check as a reason not to review. That effectively turns this optional diagnostic into a gate.
- A feature PR also cannot simply bump the package minor version to satisfy this check: the version-bump guard allows version changes only in release PRs.

My recommendation is therefore: **worry about the API break in this PR, but separately relax the bot/workflow integration**. Either have the bot gate only on required checks, or restore non-release `continue-on-error` while preserving the annotation/report. The fork origin is not the cause; it only prevents the workflow’s read-only token from posting its sticky report comment.

_This comment was generated by an AI agent (OpenHands) on behalf of enyst._

PR #4822, opened 2026-09-02
feat(sdk): StreamContext mints the stream identity and closes every stream

### LINKED ISSUE #4682: Streaming step 5: StreamContext mints the stream identity and guarantees every stream closes
Part of #4671 (step 5). Needs step 3.

## Desired Behavior

Mint the stream's identity when it opens, so the durable message retires the client's slot by its own `id` — and guarantee every opened stream is closed.

## Minting

```python
ctx = StreamContext.open(state)    # mints item_id, reads anchor_seq
...                                 # Delta frames reference it
MessageEvent(id=ctx.item_id, …)     # the message IS the item
```

**This is safe against the append-only log because minting is not a write.** `Event.id` is *already* client-minted in-process — `base.py:24-26` is `default_factory=lambda: str(uuid.uuid4())`, evaluated at construction. The log never assigns ids; it receives events that already have one. So this changes only *when* `uuid4()` runs, and makes it explicit rather than implicit. Same bytes, one append, same ordering. If the stream dies the id is simply never used and nothing on disk referenced it.

`StreamContext` is step-scoped, so it needs no home on `Conversation` and no server-side registry of open items: progress frames are never replayed, so there is nothing to keep consistent across a reconnect or restart. The only thing it needs from state is a *read* of the high-water `seq`, for `anchor_seq`.

## Four call sites, not two

`StreamContext` has to wrap **`Agent.step`, `Agent.astep`, and both ACP equivalents** (`agent.py:720-726`, `acp_agent.py:3618, 3665, 3695`), because `LLMCallContext` can't reach conversation state or the secret registry. Both agents stream on their **synchronous** path too, so a fix that covers only the async paths covers half the problem.

The ACP bridge is the reason this can't live in `llm.py`: `acp_agent.py:1305-1307` calls `on_token(text)` with a bare `str` and never enters the LLM layer at all.

The retry wrapper in `llm.py` contributes exactly one thing, `attempt`, and that's the one thing it owns.

## Always closing

Each attempt is wrapped in `try/finally`: open mints the id and reads `anchor_seq`, streaming masks and stamps each chunk, and the message is built with the minted id. If no message was produced, the `finally` emits exactly one `ItemAborted`.

This matters concretely: `Agent.astep` alone has three `except … return` branches that emit an error message and return without ever touching the stream (`agent.py:924, 935, 957`). pi enforces the same invariant one level lower — its stream function *may not throw*; failures must come back as terminal events in the stream. A stream that can't raise can't be stranded. Our `try/finally` is that rule retrofitted onto code that can.

## Masking on the stream path

Replace `mask_secrets_in_output` on the stream path with a snapshot compiled once at open: resolved values only, single-pass matcher, no lock and no I/O while streaming. `mask_secrets_in_output` resolves secret sources first — its own comment warns `get_value()` "may do blocking network I/O" — and on the sync path that runs under the state lock.

Trade-off: a secret registered *after* the stream opened won't be masked in that stream's deltas, but it **is** masked in the durable message, which is the authoritative one.

## Also worth doing here

Plumb `chunk.id` and `choice.index` through `_publish_stream_delta`. They are in scope at `event_service.py:1063-1075` and never read. litellm issues a **new** completion id per retry attempt so it cannot serve as the durable message id — minting our own is the mechanism — but carrying the provider's identity through is worth it as corroboration.

## Acceptance Criteria

- [ ] Every `ItemStarted` is retired by exactly one `Durable` or one `ItemAborted`, including on cancellation, provider failure, policy rejection and unhandled exception.
- [ ] A retry emits a higher `attempt` for the same `item_id`.
- [ ] All four entry points covered — not just the async ones.

VascoSch92:
HUMAN:

Step 5 of the streaming epic, stacked on #4807. This is the piece that gives a
stream an identity the client can close it by.

---

AGENT:

## Why

A streamed turn leaves the agent on two channels that know nothing about each
other. `StreamingDeltaEvent` carries text, a random `id` and a timestamp, and
nothing else — no stream identity, no attempt number, no pointer to the message
that supersedes it. So the browser re-marries them by comparing strings, and a
stream that dies (cancellation, provider failure, policy rejection, an
unhandled exception) leaves an open slot on the client forever.

`Agent.astep` alone has three `except … return` branches that emit an error
message and return without ever touching the stream.

Two smaller things came out of the same path. The LLM stream path masked
nothing at all, so a registered secret appearing in a token delta reached the
wire in cleartext. And `chunk.id` / `choice.index` are in scope at
`event_service.py`'s delta callback and never read — litellm hands us the
provider's identity and we drop it.

## Summary

`StreamContext` (`openhands-sdk/openhands/sdk/agent/stream_context.py`) owns one
streaming slot: it mints the id, stamps and masks each delta, and closes the
slot exactly once.

- **Minting is not a write.** `Event.id` is already client-minted in-process
  (`event/base.py`, `default_factory=lambda: str(uuid.uuid4())`); the log never
  assigns ids, it receives events that already have one. This changes only
  *when* `uuid4()` runs. Same bytes, one append, same ordering. If the stream
  dies the id is simply never used and nothing on disk referenced it.
- **The durable event retires the slot by its own id.** `claim()` hands the
  minted id to the `MessageEvent`, or to the first `ActionEvent` when the turn
  ends in tool calls — the streamed text *is* that action's thought, so a
  tool-calling turn closes on a `Durable` frame rather than aborting. Claiming
  happens at the construction site, so an error path that never builds the
  event leaves the slot open for the abort.
- **Every opened slot closes.** `__exit__` emits exactly one `ItemAborted`
  unless the id was claimed. pi enforces the same invariant one level lower —
  its stream function may not throw. This is that rule retrofitted onto code
  that can.
- **A slot opens lazily**, on the first delta rather than at `open()`, so a step
  that never streams (early condensation return, non-streaming model) has
  nothing to abort.
- **Four entry points, not two.** `Agent.step`, `Agent.astep` and both ACP
  equivalents. Both agents stream on their *synchronous* path too, so an
  async-only fix covers half the problem. The ACP bridge is also why this cannot
  live in `llm.py`: it calls `on_token(text)` with a bare `str` and never enters
  the LLM layer.
- **Masking on the stream path** uses `SecretRegistry.compile_output_mask()` — a
  snapshot of the already-resolved values, single-pass, no lock and no I/O while
  streaming. `mask_secrets_in_output` resolves uncached sources first, and its
  own comment warns `get_value()` "may do blocking network I/O"; on the sync
  path that would run under the state lock. Trade-off: a secret registered
  *after* the stream opened is not masked in that stream's deltas, but it is
  masked in the durable message, which is the authoritative one.
- **Attempts.** litellm mints a new completion id per retry attempt, so a change
  of `chunk.id` mid-item is a re-stream: the context bumps `attempt` and
  restarts `order`. The ACP prompt-retry loops call `new_attempt()` directly.
  `chunk_id` and `choice_index` ride along on the delta frame as corroboration.

### Delivery, without touching the event bus

Progress frames are not events. They ride a second fan-out on `EventService`
(`subscribe_to_stream_progress`), and `_ProgressSubscriber` in
`session_socket.py` maps them one-for-one onto the `ItemStarted` / `Delta` /
`ItemAborted` envelopes #4807 defined but nothing produced. Nothing is added to
`PubSub[Event]`, so webhooks, telemetry and the legacy `/sockets/events/{id}`
endpoint see no new traffic.

`StreamContext.on_chunk` **forwards the raw chunk downstream unchanged** before
emitting its own frames, so `on_token` consumers — the CLI, the legacy socket,
`examples/01_standalone_sdk/29_llm_streaming.py`, any user callback — are
byte-for-byte unaffected. `AgentBase.step` keeps its signature; the sink is read
off the conversation (`LocalConversation.on_stream`, new `stream_callbacks=`
argument, appended last), so no third-party `AgentBase` subclass breaks.

No progress frame is buffered against the replay boundary and none is replayed:
a dropped one costs a repaint, and the real text arrives with the durable event.

## Issue Number

Fixes #4682

## How to Test

- `uv run pytest tests/sdk -q` — **6026 passed**, 9 skipped, 10 xfailed.
- `uv run pytest tests/agent_server -q` — **2085 passed**, 13 deselected.
- `uv run pre-commit run --files <changed>` — ruff format, ruff lint,
  pycodestyle, pyright, import rules, Tool registration all pass.
- `uv run pytest tests/cross tests/tools -q` — 5 failures, all reproduced on
  the base branch with this branch's code absent (two terminal-session tests,
  two live-server tests; `test_ctrl_c` passes in isolation on both and flakes
  under load).
- `uv run python .github/scripts/check_docstrings.py` — core API files pass, no
  warnings on any file this PR touches.

New coverage:

- `tests/sdk/agent/test_stream_context.py` (13) — lazy open; exactly one abort
  per opened slot across a provider failure, a cancellation and a clean return
  with no durable event; `claim()` is once; a retry re-streams the same
  `item_id` under a higher `attempt` with `order` restarted; reasoning and text
  are separately ordered; masking; the raw chunk still reaches `on_token`; a
  sink that raises does not fail the turn.
- `tests/sdk/agent/test_stream_identity.py` (5) — through a real `Agent` and
  `LocalConversation`: the `MessageEvent` carries the minted id on both `step`
  and `astep`; a tool-calling turn retires on its first `ActionEvent`; a
  provider failure retires with one abort; a step that never reaches the
  provider opens nothing.
- `tests/sdk/agent/test_acp_agent.py` — the ACP turn's `FinishAction` event
  carries the minted id, and the context is released with the turn.
- `tests/agent_server/test_session_socket.py` (2) — the three frames reach the
  wire with their identity and the provider corroboration; an oversized delta
  drops the connection rather than a durable frame.
- `tests/agent_server/test_event_streaming.py` — progress reaches its own
  subscribers and nothing on the event bus.
- `tests/sdk/conversation/test_secrets_manager.py` (2) — the snapshot masks only
  already-resolved values and never calls an unresolved source; longest
  overlapping value wins.

### What is *not* covered

Stated plainly so a reviewer knows where to look hardest:

- **No end-to-end assertion of a progress frame over a real WebSocket.** The
  five `test_agent_server_wsproto.py` tests exercise the handler (so the new
  subscribe/unsubscribe runs on every session-socket connection), but producing
  a delta needs a streaming LLM inside a server that those tests start as a
  separate process. The path is covered in two halves instead: publish reaches
  its subscribers (`test_event_streaming.py`) and a subscriber reaches the wire
  (`test_session_socket.py`).
- **The attempt boundary is exercised with synthetic chunks**, not against real
  litellm retry behavior. See the note below on why it is inferred at all.
- **The ACP prompt-retry `new_attempt()` calls are untested** — reaching them
  needs injected connection errors against a live ACP subprocess.
- **The `MaxSubscribersError` branch for progress is untested.**

`stream_context.py` and `session_protocol.py` are at 100% statement coverage.

## Video/Screenshots

N/A — no UI surface.

## Design Doc

N/A — the module docstring in `stream_context.py` carries the rationale inline.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Second commit is a self-review pass** that found three defects in the
  first, each with a regression test:
  - A retry that died before its first token stranded the slot.
    `new_attempt()` clears the per-attempt "started" flag and `close()` keyed
    its abort on that flag, so the attempt that *had* streamed was never
    retired — the exact invariant this PR exists to establish. `close()` now
    keys on whether the item was ever opened.
  - A raising progress sink escaped `close()`, which runs from `__exit__`, so
    a broken subscriber could replace the exception a failing step was
    unwinding, or fail a step that had succeeded.
  - The agent handed the LLM a wrapper that was never `None`, which defeats the
    fallback in `llm.completion` that degrades a `stream=True` model to a
    non-streaming call when no `on_token` is wired (#4014). `StreamContext`
    now reports no callback when nothing consumes one, so a caller with no
    token *and* no stream callbacks gets exactly today's behavior.

- **Merge order:** #4806 (#4680), then #4807 (#4681), then this one. This branch
  is based on `vasco/streaming-4681-session-socket`, so the diff here is only
  the step-5 work.
- **One fix outside the issue's scope, and it was load-bearing.** The new
  agent-level tests made `tests/sdk` fail 17 tests in modules they never touch,
  with `Duplicate class definition … ClientAction_persist_navigate_to`. The
  cause is in `utils/models.py`: `clear_subclass_cache()` bumped the generation
  counter but left the stale `_concrete_cache` / `_checked_cache` entries in
  place, and each of those is a *strong* reference to every subclass it
  captured. So a dynamically-created class survived the test that unregistered
  it, and the next rebuild saw two classes with one name. The same failure
  reproduces on `main`'s own files with no new code at all:
  `uv run pytest tests/sdk/agent/test_agent_content_policy_violation.py tests/sdk/conversation/local/test_client_tools_persistence.py tests/sdk/tool/test_tool_serialization.py` → 6 failed. `clear_subclass_cache()` now drops the
  entries, and `_wipe_client_tool_globals` calls it. Happy to split this out if
  you would rather it landed on its own.
- **The attempt boundary is inferred, not reported.** The retry wrapper in
  `llm.py` owns `attempt`, but it has no per-call channel to the token callback,
  so the context detects a new attempt from a change in `chunk.id`. This is a
  heuristic: a provider that varies its completion id mid-response would read as
  a retry. Documented at the detection site.
- **The ACP bridge still masks each chunk itself** before `on_token`, so a
  streamed ACP delta is masked twice now (idempotent). Removing the bridge's own
  masking is OpenHands/OpenHands#15720's business, not this PR's.
- Deltas still publish as `StreamingDeltaEvent` on the event bus for the legacy
  endpoint. Deleting that channel is step 7 (#4683).

---
<details>

**Variants & Base Images**
|---|---|---|---|

```bash
docker pull ghcr.io/openhands/agent-server:3c9e77e-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-3c9e77e-python \
  ghcr.io/openhands/agent-server:3c9e77e-python
```

```
ghcr.io/openhands/agent-server:3c9e77e-golang-amd64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-golang-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-golang-amd64
ghcr.io/openhands/agent-server:3c9e77e-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:3c9e77e-golang-arm64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-golang-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-golang-arm64
ghcr.io/openhands/agent-server:3c9e77e-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:3c9e77e-java-amd64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-java-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-java-amd64
ghcr.io/openhands/agent-server:3c9e77e-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:3c9e77e-java-arm64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-java-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-java-arm64
ghcr.io/openhands/agent-server:3c9e77e-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:3c9e77e-python-amd64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-python-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-python-amd64
ghcr.io/openhands/agent-server:3c9e77e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-amd64
ghcr.io/openhands/agent-server:3c9e77e-python-arm64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-python-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-python-arm64
ghcr.io/openhands/agent-server:3c9e77e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-arm64
ghcr.io/openhands/agent-server:3c9e77e-python-slim-amd64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-python-slim-amd64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-python-slim-amd64
ghcr.io/openhands/agent-server:3c9e77e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-amd64
ghcr.io/openhands/agent-server:3c9e77e-python-slim-arm64
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-python-slim-arm64
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-python-slim-arm64
ghcr.io/openhands/agent-server:3c9e77e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim-arm64
ghcr.io/openhands/agent-server:3c9e77e-golang
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-golang
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-golang
ghcr.io/openhands/agent-server:3c9e77e-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:3c9e77e-java
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-java
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-java
ghcr.io/openhands/agent-server:3c9e77e-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:3c9e77e-python-slim
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-python-slim
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-python-slim
ghcr.io/openhands/agent-server:3c9e77e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim-slim
ghcr.io/openhands/agent-server:3c9e77e-python
ghcr.io/openhands/agent-server:3c9e77e3807524a3e29ca407f089c34b14b12e9f-python
ghcr.io/openhands/agent-server:vasco-streaming-4682-stream-context-python
ghcr.io/openhands/agent-server:3c9e77e-nikolaik_s_python-nodejs_tag_python3.13-nodejs22-slim
```

- Each variant tag (e.g., `3c9e77e-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `3c9e77e-python-amd64`) are also available if needed
</details>


openhands-sdk/openhands/sdk/agent/stream_context.py
```diff
@@ -0,0 +1,250 @@
...
+
+The durable event is built with the minted id, so a client retires its open
+slot on ``frame.event.id == slot.item_id``.
+
+Minting up front is safe against the append-only log because it is not a write:
+``Event.id`` is already client-minted in-process (``event/base.py``), so this
+changes only *when* ``uuid4()`` runs, and an unused id was never on disk.
+"""
+
+from __future__ import annotations
```
VascoSch92, line 11: do we need this?

openhands-agent-server/openhands/agent_server/session_socket.py
```diff
@@ -6,10 +6,10 @@
...
 task awaits the socket.
 
-``ItemStarted`` / ``Delta`` / ``ItemAborted`` are carried but never produced
-yet — that needs ``StreamContext`` (#4682). Until then this is a durable-only
-channel and ``StreamingDeltaEvent`` is dropped rather than forwarded, since
-putting it back on the wire would restore the coupling this endpoint removes.
+``ItemStarted`` / ``Delta`` / ``ItemAborted`` arrive over the event service's
+separate progress fan-out, never the event bus — so ``StreamingDeltaEvent`` is
+still dropped rather than forwarded, which would restore the coupling this
+endpoint removes.
```
VascoSch92: we don't need this comment i think

openhands-sdk/openhands/sdk/agent/stream_context.py
```diff
@@ -0,0 +1,250 @@
...
+        delta = getattr(choice, "delta", None)
+        if delta is None:
+            continue
+        index = getattr(choice, "index", None)
+        reasoning = getattr(delta, "reasoning_content", None)
+        if isinstance(reasoning, str) and reasoning:
+            out.append(("reasoning", reasoning, chunk_id, index))
+        content = getattr(delta, "content", None)
+        if isinstance(content, str) and content:
+            out.append(("text", content, chunk_id, index))
```
VascoSch92: why are we using geaatr everywhere here?

openhands-sdk/openhands/sdk/agent/stream_context.py
VascoSch92: Yes, the module does not import without it. `open()` returns `StreamContext` (a self-reference evaluated at class-definition time) and takes `LocalConversation`, which is `TYPE_CHECKING`-only to avoid an import cycle. Dropping the line gives `NameError: name 'LocalConversation' is not defined` on import; I checked. 98 of the 289 modules under `openhands/sdk` do the same. The alternative is quoting both annotations, which I think reads worse.

openhands-agent-server/openhands/agent_server/session_socket.py
VascoSch92: Agreed, removed in bc008f04c. `_SessionSubscriber.__call__` already says at the drop site why a delta never rides the durable channel, so the paragraph was repeating it; I updated that comment to the present tense since the frames are produced now.

openhands-sdk/openhands/sdk/agent/stream_context.py
VascoSch92: Over-applied, and I have cut it back in bc008f04c. `chunk.id`, `chunk.choices`, `choice.delta`, `choice.index` and `delta.content` are all plain attribute access now on a `LLMStreamChunk | str` parameter.

One `getattr` has to stay, for `reasoning_content`. litellm *deletes* that attribute when the provider omits it, even though it is a declared field on `Delta`:

```
>>> d = Delta(role="assistant", content="hi")
>>> d.reasoning_content
AttributeError: 'Delta' object has no attribute 'reasoning_content'
```

I found that the hard way: the direct-access version made 12 tests fail because every chunk raised before emitting anything. The line now carries a comment saying so. `event_service._token_streaming_callback` has the same `getattr` for the same reason.

VascoSch92:
@all-hands-bot fix the conflicts

juanmichelini:
@openhands please follow this plan:

1) resolve conflicts
2) fix precommit checks

openhands-sdk/openhands/sdk/agent/stream_context.py
```diff
@@ -0,0 +1,250 @@
...
+        if not self._started:
+            self._started = self._opened = True
+            self._emit(StreamStarted(self.item_id, self._attempt, self._anchor_seq))
+        self._emit(
+            StreamDelta(
+                item_id=self.item_id,
+                attempt=self._attempt,
+                order=next(self._order),
+                kind=kind,
+                content=self._mask(text),
```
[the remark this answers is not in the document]
juanmichelini: Addressed in d3017a34: the stream masker now holds back the shortest suffix that could still grow into a registered value across chunk boundaries (StreamOutputMask.feed/flush in secret_registry.py). Covered by test_compile_stream_mask_masks_a_value_split_across_chunks and the agent-level test_stream_identity test_a_resolved_secret_is_masked_in_the_deltas. Joining StreamDelta.content values never reconstructs a secret.

juanmichelini (approved):
LGTM, tested by using node /tmp/watch-session-socket.mjs

PR #4838, opened 2026-09-03
fix(mcp): recover after HTTP session failure

### LINKED ISSUE #4837: [Bug]: MCP reconnect remains broken after transient HTTP error

### Bug Description

When a streamable HTTP MCP tool call receives a transient HTTP error, FastMCP terminates its session task but leaves the nesting counter nonzero. The SDK's next automatic reconnect then fails permanently instead of opening a fresh session.

### Expected Behavior

The failed tool call should report the HTTP error, and the next tool call should reconnect and succeed when the server is healthy again.

### Actual Behavior

On current `main` (`e26683288ab4dd69518810016b74682de2a8c4e4`), run the real HTTP regression with `uv run pytest -q tests/sdk/mcp/test_create_mcp_tool.py -k reconnects_after_http_error`.

The first call correctly reports `504 Gateway Timeout`. The second call fails with `MCP Connection Failure`, caused by:

```text
RuntimeError: Internal error: nesting counter should be 0 when starting new session, got 1
```

### Steps to Reproduce

1. Start a FastMCP streamable HTTP server.
2. Return HTTP 504 for the first `tools/call` request.
3. Call the same MCP tool again through the SDK client.
4. Observe that automatic reconnect fails before sending the second tool call.

### Acceptance Criteria

- [ ] After an HTTP error terminates the FastMCP session task, the next MCP tool call reconnects and succeeds.
- [ ] Existing MCP client tests continue to pass.

### Installation Method

`uv sync` from the repository checkout.

### SDK Version

Current `main` at `e26683288ab4dd69518810016b74682de2a8c4e4`.

### Version Confirmation

### Python Version

3.13.7

### Operating System

Linux

### Logs and Error Messages

```text
First call: 504 Gateway Timeout
Second call: MCP Connection Failure
Cause: RuntimeError: Internal error: nesting counter should be 0 when starting new session, got 1
```

### Minimal Code Sample

The linked PR includes an integration test that starts a local FastMCP HTTP app, injects one 504 response, and makes two calls through the real SDK tool executor.

Draft PR #3444 identified the same FastMCP stale-counter state on an older SDK architecture. Current `main` now has automatic reconnect in the tool executor; this report covers that current path and uses a real HTTP failure rather than mocked session state.

The failure reproduces with the repository's locked FastMCP 3.2.0 and also with 3.4.7. It does not reproduce with FastMCP 4.0.2, whose session implementation changed, but adopting a breaking major version is separate from preserving reconnect behavior for the currently supported FastMCP 3.x dependency.

Shimada666:
HUMAN:
I reproduced the issue and confirmed this fix is ready for review.

---

AGENT:

## Why

A transient HTTP error can terminate FastMCP's session task while leaving its nesting counter nonzero. The SDK already retries disconnected tools, but that retry then fails with `MCP Connection Failure` because FastMCP refuses to start another session from the stale state.

## Summary

- Reset a completed FastMCP session under its existing session lock before reconnecting.
- Add a real HTTP regression test that injects one 504 response and verifies the next tool call succeeds.

## Issue Number

Fixes #4837

## How to Test

End-to-end reproduction used a local FastMCP HTTP app rather than mocked client state. Its middleware returned HTTP 504 for the first `tools/call`; the first SDK observation contained `504 Gateway Timeout`, and the second call returned `pong`.

```bash
uv run pytest -q tests/sdk/mcp/test_create_mcp_tool.py -k reconnects_after_http_error
uv run pytest -q tests/sdk/mcp
uv run pre-commit run --files openhands-sdk/openhands/sdk/mcp/client.py tests/sdk/mcp/test_create_mcp_tool.py
```

Results: the regression passed three consecutive runs; the MCP suite passed with 135 tests; all pre-commit hooks passed.

## Video/Screenshots

Not applicable; this is an SDK connection lifecycle fix covered by the real HTTP regression above.

## Design Doc

Not applicable; the production change is a four-line state reset using FastMCP's existing lock and reset method.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Draft PR #3444 reported the same stale FastMCP counter on an older SDK architecture. This PR targets the current automatic reconnect path and keeps the workaround at that shared reconnect boundary. Issue #4837 needs maintainer triage to add the `bug` label, after which the repository workflow can apply `ready-for-dev`.

The real HTTP reproduction fails with the repository's locked FastMCP 3.2.0 and also with 3.4.7. FastMCP 4.0.2 does not reproduce it because its session implementation changed; upgrading to that breaking major remains separate from this compatibility fix for supported FastMCP 3.x installations.

### DIFF
--- a/openhands-sdk/openhands/sdk/mcp/client.py
+++ b/openhands-sdk/openhands/sdk/mcp/client.py
@@ -57,6 +57,10 @@ class MCPClient(AsyncMCPClient):
 
     async def connect(self) -> None:
         """Establish connection to the MCP server."""
+        async with self._session_state.lock:
+            session_task = self._session_state.session_task
+            if session_task is not None and session_task.done():
+                self._reset_session_state(full=True)
         try:
             await self.__aenter__()
         except RuntimeError as exc:
--- a/tests/sdk/mcp/test_create_mcp_tool.py
+++ b/tests/sdk/mcp/test_create_mcp_tool.py
@@ -13,11 +13,14 @@ from unittest.mock import MagicMock, patch
 
 import httpx
 import pytest
+import uvicorn
 from fastmcp import FastMCP
 from fastmcp.client.auth import OAuth
 from fastmcp.mcp_config import MCPConfig as FastMCPConfig, RemoteMCPServer
 from key_value.aio.stores.memory import MemoryStore
 from pydantic import SecretStr
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.responses import PlainTextResponse
 
 from openhands.sdk.mcp import create_mcp_tools
 from openhands.sdk.mcp.config import (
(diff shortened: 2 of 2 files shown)


HussainShibli:
We reproduced this exact class of failure live with an SSE MCP "desktop bridge" (each conversation holds its own SSE session to a desktop app; when the bridge restarts, **existing** conversations fail permanently with `Connection closed` / nesting-counter-stuck errors, while brand-new conversations always work).

One nuance our testing surfaced that may be relevant here: the drop can surface in **two** ways -
(a) as a raised `MCPError` / `Connection closed`, and
(b) as an **error tool-result** when calls fan out through the in-process fastmcp gateway (a dead remote session is returned as `CallToolResult(isError=true, text="Error calling tool '...': Connection closed")` rather than raised). Reconnect/retry logic keyed only on raised exceptions will not cover (b).

Environment: openhands-sdk / openhands-agent-server 1.42.1; 1.44.1 and current `main` (as of 2026-09-07) still lack the fix.

*Prepared with the assistance of an AI agent (OpenHands) on behalf of HussainShibli.*

Issue #4800, opened 2026-08-31
[Bug]: ConversationService: stored fields mutated in memory before save_meta() with no rollback on failure


BSmick6:
`ConversationService` has a recurring pattern where fields on `StoredConversation` are mutated in memory before the corresponding `save_meta()` call persists them to `meta.json`. If `save_meta()` fails (disk full, ownership lost, etc.), the in-memory state is ahead of disk. After eviction or restart, the change silently disappears.

This currently affects at least three places:
- `update_conversation` — mutates `stored.title`, `stored.tags`, `stored.updated_at`
- `set_conversation_tag` / `delete_conversation_tag` (via `_flush_tags_mutation`) — mutates `stored.tags`
- `AutoTitleSubscriber` — mutates `stored.title`, `stored.updated_at`

### Actual Behavior

When `save_meta()` raises after an in-memory mutation, the change is visible in-process but not persisted. After eviction or restart the mutation disappears silently. The following pytest demonstrates the gap:

```python
# pytest tests/agent_server/test_conversation_tags.py::test_set_tag_rolls_back_on_save_failure
from unittest.mock import patch
with patch.object(event_service, "save_meta", side_effect=RuntimeError("disk full")):
    with pytest.raises(RuntimeError):
        await service.set_conversation_tag(conv_id, "key", "value")
assert "key" not in event_service.stored.tags  # currently fails — key is present despite failed persist
```

Run with:

```
pytest tests/agent_server/test_conversation_tags.py
```

### Acceptance Criteria

- [ ] `set_conversation_tag` leaves `stored.tags` unchanged when `save_meta()` raises
- [ ] `delete_conversation_tag` leaves `stored.tags` unchanged when `save_meta()` raises
- [ ] `update_conversation` leaves `stored.title` / `stored.tags` unchanged when `save_meta()` raises
- [ ] `AutoTitleSubscriber` leaves `stored.title` unchanged when `save_meta()` raises

The cleanest fix is reversing the order — write to disk first, then update in-memory state — rather than adding rollback logic after the fact. Not a blocker, but worth addressing consistently across all affected sites rather than piecemeal.

BSmick6:
@all-hands-bot

PR #4859, opened 2026-09-04
fix(agent-server): persist confirmation policy, analyzer and secrets to meta.json

### LINKED ISSUE #4810: [Bug]: confirmation_policy, security_analyzer, and secrets mutations not persisted to meta.json

alanhuangyoo:
HUMAN:

Turned on confirmation prompts and an analyzer through the API, restarted the server, and they were both back to what the conversation started with. They stick now.

---

AGENT:

## Why

Closes #4810.

`set_confirmation_policy`, `set_security_analyzer` and `update_secrets` delegate to the conversation, which writes `base_state.json`, and never touch `StoredConversation`:

```python
async def set_confirmation_policy(self, policy: ConfirmationPolicyBase):
    ...
    await loop.run_in_executor(None, self._conversation.set_confirmation_policy, policy)
    # nothing updates self.stored, nothing calls save_meta()
```

Startup reads all three off `StoredConversation` (`event_service.py:1105-1117`), so after eviction or a restart the conversation reverts to the values it was created with. For two of the three that means the **security analyzer and confirmation policy silently regress** to the creation-time setting.

It is worse than the issue describes: `stored` is stale in process as well, so it is not only a reload problem.

```
$ python repro.py
in-process confirmation_policy : NeverConfirm       <- set to AlwaysConfirm
in-process security_analyzer   : NoneType           <- set to LLMSecurityAnalyzer
in-process secrets keys        : []                 <- MY_TOKEN was set
meta.json confirmation_policy  : {'kind': 'NeverConfirm'}
meta.json security_analyzer    : None
after restart confirmation_policy: NeverConfirm
```

## Summary

- All three mutations write through to `meta.json`, via `model_copy` + `save_meta` — the same shape `apply_resume_secrets` already uses for the resume path, so there is one pattern rather than two.
- Plain-string secrets are normalised to `SecretSource` before they reach `stored`.
- Rollback when `save_meta` itself fails is deliberately untouched: that is #4800, and mixing it in here would collide with whoever takes it.

### Why the normalisation is not incidental

`update_secrets` is typed `dict[str, SecretValue]`, and `SecretValue = str | SecretSource`. `StoredConversation.secrets` is `dict[str, SecretSource]`. `model_copy(update=...)` **does not validate**, so merging a raw string in produces a model that looks fine and then fails at the point of writing meta.json:

```
PydanticSerializationError: Error calling function `_serialize_by_kind`:
AttributeError: 'str' object has no attribute '_is_handler_for_current_class'
```

The API route validates through `UpdateSecretsRequest`, so the HTTP path never hits this, but the method accepts strings and `apply_resume_secrets` merges the same way. `test_a_plain_string_secret_is_stored_as_a_secret_source` passes a bare string so this cannot regress.

## How to Test

```bash
uv run pytest tests/agent_server/test_conversation_service.py -k "mutation or plain_string" -q
```

Full suite:

```
2094 passed, 13 deselected in 180.44s
```

`ruff check` and `ruff format --check` clean on both changed files.

### End-to-end, not just unit tests

Start a conversation, mutate all three through the service, drop the `ConversationService` entirely, and bring a fresh one up against the same directory:

```
in-process confirmation_policy : AlwaysConfirm
in-process security_analyzer   : LLMSecurityAnalyzer
in-process secrets keys        : ['MY_TOKEN']
meta.json confirmation_policy  : {'kind': 'AlwaysConfirm'}
meta.json security_analyzer    : {'kind': 'LLMSecurityAnalyzer'}
meta.json secrets keys         : ['MY_TOKEN']
after restart confirmation_policy: AlwaysConfirm
after restart security_analyzer  : LLMSecurityAnalyzer
after restart secret value       : 'shh'
```

### What happens to the secret at rest

I checked this rather than assuming, since the change is what starts routing API secret updates into `meta.json`.

With a cipher configured, the value is encrypted and round-trips:

```json
{"MY_TOKEN": {"value": "gAAAAABqmvmIuMS90CqPhDCi_Xo-...", "kind": "StaticSecret"}}
```

`'shh' in meta.json` is `False`, and the reloaded source returns `'shh'`.

Without a cipher, the serialiser masks it, so the name persists and the value loads back as `None` — **not** as the `**********` placeholder, which would otherwise reach a tool call. Both behaviours are what `apply_resume_secrets` already produced; this change does not alter them.

### Tests

Three added. All three fail on the parent commit:

```
FAILED test_api_mutations_survive_a_server_restart
FAILED test_mutations_are_written_to_meta_json
FAILED test_a_plain_string_secret_is_stored_as_a_secret_source
```

`test_mutations_are_written_to_meta_json` asserts against the file rather than the object, because meta.json is what startup actually reads.

## Issue Number

Closes #4810

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change

### DIFF
--- a/openhands-agent-server/openhands/agent_server/event_service.py
+++ b/openhands-agent-server/openhands/agent_server/event_service.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from typing import cast
 from uuid import UUID, uuid4
 
-from pydantic import ValidationError
+from pydantic import SecretStr, ValidationError
 
 from openhands.agent_server.conversation_lease import (
     DEFAULT_LEASE_TTL_SECONDS,
@@ -75,6 +75,7 @@ from openhands.sdk.git.exceptions import GitCommandError, GitRepositoryError
 from openhands.sdk.git.utils import run_git_command, validate_git_repository
 from openhands.sdk.llm.streaming import LLMStreamChunk
 from openhands.sdk.mcp.utils import MCPToolProvider
+from openhands.sdk.secret import SecretSource, StaticSecret
 from openhands.sdk.security.analyzer import SecurityAnalyzerBase
 from openhands.sdk.security.confirmation_policy import ConfirmationPolicyBase
 from openhands.sdk.utils.async_utils import AsyncCallbackWrapper
@@ -108,6 +109,22 @@ def _without_agent_context_secret(
     return agent.model_copy(
         update={"agent_context": context.model_copy(update={"secrets": secrets})}
     )
+
+
+def _as_secret_sources(
+    secrets: Mapping[str, SecretValue],
+) -> dict[str, SecretSource]:
+    """Normalise a SecretValue mapping for storage.
+
+    ``EventService.update_secrets`` accepts ``SecretValue`` (``str | SecretSource``)
+    while ``StoredConversation.secrets`` is ``dict[str, SecretSource]``, and
+    ``model_copy`` does not validate, so a plain string would only fail later when
+    meta.json is serialised.
+    """
+    return {
+        name: StaticSecret(value=SecretStr(value)) if isinstance(value, str) else value
+        for name, value in secrets.items()
+    }
 
 
 @dataclass
@@ -188,6 +205,16 @@ class EventService:
                     }
                 )
             )
+
+    async def _persist_stored(self, **updates) -> None:
+        """Apply updates to the stored conversation and write meta.json.
+
+        ``model_copy`` rather than in-place mutation, matching
+        ``apply_resume_secrets``. Rollback when ``save_meta`` itself fails is
+        tracked separately in #4800 and is not changed here.
+        """
+        self.stored = self.stored.model_copy(update=updates)
+        await self.save_meta()
 
     def _without_stored_secret(self, secret_name: str) -> StoredConversation:
         # meta.json (StoredConversation) no longer carries the agent, so there is
@@ -1679,6 +1706,12 @@ class EventService:
             secrets.pop(CODEX_AUTH_SECRET_NAME, None)
         loop = asyncio.get_running_loop()
         await loop.run_in_executor(None, self._conversation.update_secrets, secrets)
+        # The conversation writes base_state.json, but startup reads these three
+        # fields off StoredConversation, so meta.json has to carry them too or the
+        # change is reverted on the next load.
+        await self._persist_stored(
+            secrets={**self.stored.secrets, **_as_secret_sources(secrets)}
+        )
 
     async def set_confirmation_policy(self, policy: ConfirmationPolicyBase):
         """Set the confirmation policy for the conversation."""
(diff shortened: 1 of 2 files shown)


enyst (changes requested):
Thank you for the PR, @alanhuangyoo ! I think maybe the right solution is rather to read them from `base_state.json`?

I believe we have another PR working through this… let me find it 🙏

_Edit_: https://github.com/OpenHands/software-agent-sdk/pull/4813

alanhuangyoo:
You are right, and #4813 is the better fix. Closing this.

Mine adds a sync; #4813 removes the need for one by taking the three fields off `StoredConversation` entirely, so `base_state.json` is the only place they live — the same shape #4440 used for `agent`. Two sources kept in step is a thing someone has to remember; one source is not.

I had reached for the sync because `apply_resume_secrets` already does exactly that and I matched it. That was the wrong precedent to match.

Before closing I checked one thing #4813 depends on and one thing it could have broken, since I had the reproduction already set up. Posting the results on #4813 rather than here.

Thanks @enyst.

PR #4868, opened 2026-09-05
fix(agent-server): skip the inert ACP agent LLM when auto-titling

### LINKED ISSUE #4867: [Bug]: ACP mode: conversation title generation fails with missing credentials (explicit LLM without a key)

### Bug Description

In ACP mode the conversation is driven by an external subprocess and the LLM settings are inert, but conversation title generation still makes a real LLM call.

`ACPAgent` installs a sentinel LLM (`openhands-sdk/openhands/sdk/agent/acp_agent.py:264-269`):

```python
ACP_SENTINEL_USAGE_ID = "acp-managed"

def _make_dummy_llm() -> LLM:
    """Create a dummy LLM that should never be called directly."""
    return LLM(model="acp-managed", usage_id=ACP_SENTINEL_USAGE_ID)
```

That is only a *default*. A backend that persists an agent LLM — an agent profile with a model set — keeps that model on the agent while the conversation runs in the ACP subprocess under its own credentials. The sentinel guard in `generate_conversation_title` matches on `usage_id == "acp-managed"`, so with a persisted LLM it never fires and titling reaches a real provider with no key.

The auto-title fallback has no ACP check (`openhands-agent-server/openhands/agent_server/conversation_service.py`):

```python
title_llm = self._load_title_llm()
if title_llm is None:
    title_llm = conversation.agent.llm if conversation else None
```

Agent Canvas documents the same contract from the other side (`src/types/settings.ts:105`): for `agent_kind: "acp"`, *"The LLM / condenser settings are inert; `mcp_config` and `acp_command` / `acp_args` / `acp_model` / `acp_server` apply instead."*

### Expected Behavior

With an ACP agent and no explicitly configured `title_llm_profile`, skip LLM titling and fall back to truncation. No provider call, no retries, no error surfaced to the UI.

### Actual Behavior

The agent reports `agent_kind == "acp"` while carrying a real model whose `usage_id` is not the sentinel, so the guard that is supposed to prevent the call cannot fire. Save this as `acp_repro.py` in a checkout of this repository and run it:

```bash
make build
uv run python acp_repro.py
```

Output on `main`:

```text
agent_kind            : acp
agent.llm.model       : gpt-4o
agent.llm.usage_id    : agent
sentinel usage_id     : acp-managed
sentinel guard fires  : False
```

Titling therefore calls a provider with no credentials, retries four times, and surfaces the failure.

### Steps to Reproduce

1. Run an agent-server with `agent_kind: "acp"` and `acp_server: "claude-code"` under subscription auth.
2. Leave `title_llm_profile` unset, and set an `active_agent_profile_id` whose profile carries a model.
3. Start a conversation and send any first user message.
4. Auto-titling calls that model with no key: four retries, and since #16686 the error is published to the frontend.

### Acceptance Criteria

- [ ] With `agent_kind == "acp"` and no `title_llm_profile`, auto-titling makes no LLM call.
- [ ] The conversation title falls back to truncation in that case.
- [ ] An explicitly configured `title_llm_profile` still takes precedence and is used.
- [ ] Non-ACP agents keep the existing `agent.llm` fallback unchanged.
- [ ] No credentials error is surfaced to the frontend for an ACP backend that holds no model credentials.

### Installation Method

Self-hosted `openhands-agent-server` released image; reproduced from a source checkout via `make build` and `uv run`.

_No response_

### SDK Version

1.44.0 (observed). Reproduced against `main`.

### Version Confirmation

### Python Version

3.13.15

Conversation runs under `claude-code` via ACP subscription auth; the agent profile carried an OpenAI model, which is the one titling tried to call.

### Operating System

Linux

### Logs and Error Messages

Self-hosted worker, agent-server 1.44.0, `agent_kind: "acp"`, `acp_server: "claude-code"`, subscription auth. The conversation itself works and MCP works; every conversation logs four retries:

```text
[agent-server] ERROR   litellm.InternalServerError: InternalServerError: OpenAIException -
  Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`,
  or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.. Attempt #3
[agent-server] WARNING Error generating conversation title with LLM: ...  title_utils.py:152
```

The failure is already non-fatal — `generate_conversation_title` catches, logs, and callers fall back to truncation — but since #16686 the error is also propagated to the frontend, so ACP users see a credentials error on a backend that deliberately holds no model credentials. Cost is roughly 15s of retries per conversation plus a misleading error.

### Minimal Code Sample

```python
"""ACP agent LLM is inert, yet auto-titling still resolves to it."""
from openhands.sdk.agent import ACPAgent
from openhands.sdk.agent.acp_agent import ACP_SENTINEL_USAGE_ID

# A backend that persists an agent LLM (an agent profile with a model set)
# keeps that model on the agent even though ACP runs the conversation.
agent = ACPAgent(
    llm=LLM(model="gpt-4o", usage_id="agent"),
    acp_server="claude-code",
    acp_command=["claude-agent-acp"],
)
print("agent_kind            :", agent.agent_kind)
print("agent.llm.model       :", agent.llm.model)
print("agent.llm.usage_id    :", agent.llm.usage_id)
print("sentinel usage_id     :", ACP_SENTINEL_USAGE_ID)
print("sentinel guard fires  :", agent.llm.usage_id == ACP_SENTINEL_USAGE_ID)
```

Note for anyone hitting this: the obvious workaround — exporting `OPENAI_API_KEY` for the agent-server — is a trap. The ACP subprocesses inherit the parent environment, and the Codex CLI prefers `OPENAI_API_KEY` when present, which can silently switch the main workload from subscription auth to API auth.

A PR is open at #4868.

lkshrk:
HUMAN:

Found this on a self-hosted Agent Canvas worker running ACP with Claude Code on
subscription auth: every conversation surfaced a credentials error in the UI while
the conversation and its MCP tools worked normally. Confirmed on that host that the
agent LLM is inert under ACP and nothing should be calling it for titles, and that
`title_llm_profile` was unset so the fallback was the only path in play. The
agent-server suite and linters were run locally on the branch, and the new test was
checked to fail with the fix reverted. Not able to run a patched agent-server against
that deployment, since it consumes released images.

---

AGENT:

## Why

Auto-titling falls back to `agent.llm` when no `title_llm_profile` is configured. `ACPAgent.llm` only *defaults* to the `acp-managed` sentinel (`acp_agent.py:1673`), so a backend that persists an agent LLM keeps that model on the agent while the conversation itself runs in the ACP subprocess under its own credentials. The sentinel guard in `generate_title_from_message` (`title_utils.py:186`) matches on `usage_id == "acp-managed"` and therefore never fires, and the title call reaches a real provider with no key:

```
[agent-server] ERROR   litellm.InternalServerError: InternalServerError: OpenAIException -
  Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`,
  or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.. Attempt #3
[agent-server] WARNING Error generating conversation title with LLM: ...  title_utils.py:152
```

Four retries per conversation, and since #16686 the failure is published to the frontend as well, so ACP users see a credentials error on a backend that deliberately holds no model credentials. Agent Canvas states the same contract from the other side: for `agent_kind: "acp"`, *"The LLM / condenser settings are inert; `mcp_config` and `acp_command` / `acp_args` / `acp_model` / `acp_server` apply instead."*

## Summary

- `AutoTitleSubscriber` no longer falls back to `agent.llm` for an ACP agent, using the existing public `AgentBase.agent_kind` property rather than an `isinstance` check, so the SDK's agent surface stays the seam.
- `title_llm_profile` keeps precedence, so an ACP deployment that wants generated titles configures one; otherwise titling falls through to truncation, which is where a credential-less agent already ended up after four failed attempts.
- Adds `TestAutoTitle::test_autotitle_skips_inert_acp_agent_llm`, which asserts the title path receives `llm=None` and the stored title comes from truncation.

Deliberately **not** treating any credential-less LLM as unusable. That would need a heuristic for "has a usable key", would silently change behaviour for a misconfigured non-ACP backend that today surfaces a real error, and would hide exactly the class of misconfiguration #16686 set out to make visible. ACP is the case where the contract already says the LLM is inert, so that is where the skip belongs. If you would rather generalize, say so and I will rework it.

## Issue Number

Fixes #4867

## How to Test

```
uv run pytest tests/agent_server/test_conversation_service.py -q   # 114 passed
uv run ruff check <changed>                                        # All checks passed!
uv run ruff format --check <changed>                               # 2 files already formatted
```

The new test is regression-sensitive: reverting only the `conversation_service.py` hunk and re-running gives

```
FAILED tests/agent_server/test_conversation_service.py::TestAutoTitle::test_autotitle_skips_inert_acp_agent_llm
1 failed, 6 passed
```

Reproduced on a self-hosted worker running `openhands-agent-server` 1.44.0 with `agent_kind: "acp"`, `acp_server: "claude-code"`, subscription auth, `title_llm_profile: null`, and an `active_agent_profile_id` whose model matches the one named in the error. Every new conversation produces the log above while the conversation and its MCP tools work normally.

I could not run a patched agent-server against that deployment: it consumes released images, so the end-to-end evidence here is the failure reproduction plus the unit-level proof rather than a patched round trip.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

One workaround is worth recording as a trap: exporting `OPENAI_API_KEY` for the agent-server does make titling work, but ACP subprocesses inherit the parent environment and the Codex CLI prefers `OPENAI_API_KEY` when present, which can silently switch the main workload from subscription auth to API auth.

Linked issue #4867 was originally written free-form and so carried none of the bug-template headings, which is why `Validate PR description` failed here with *"Linked issue(s) (#4867) carry neither `ready-for-dev` nor a pre-rollout creation date."* It has been restructured onto the template — all 16 fields, a runnable reproduction under `### Actual Behavior`, and acceptance criteria — and now carries `ready-for-dev`.


neubig (approved):
ACP agent LLM settings are inert for execution and are not a valid implicit title-generation credential source. The agent-kind guard is the correct seam: it preserves explicit title profiles and all non-ACP behavior while avoiding credential heuristics. The regression test covers the resulting truncation path.

AI disclosure: This review was submitted by OpenHands on behalf of @neubig.

PR #4889, opened 2026-09-07
refactor(agent-server): make ConversationInfo source mapping explicit

### LINKED ISSUE #4849: [Feature]: Derive ConversationInfo from ConversationState + StoredConversation instead of re-declaring fields
### Problem or Use Case

`ConversationInfo` (the HTTP response shape returned by every conversation endpoint) re-declares ~20 fields that already exist on `ConversationState` or `StoredConversation`. The assembly function `_compose_conversation_info()` merges them with `**state.model_dump(mode="json")`, but `ConversationInfo`/`_ConversationInfoBase` only picks up fields it explicitly declares — any field added to `ConversationState` that a client needs is silently dropped from the response unless someone also remembers to add it to `_ConversationInfoBase`.

This happened recently with `activated_path_rules`: it was added to `ConversationState` but never added to `_ConversationInfoBase`, so it is silently absent from every API response. It is harmless today (the field is internal dedup state no client needs), but the pattern means future fields can vanish without error, test failure, or any signal.

Discovered while reviewing #4814 and the surrounding `ConversationState` / `StoredConversation` boundary work in #4810 / #4811.

### Desired Behavior

Adding a field to `ConversationState` or `StoredConversation` should not require a separate, coordinated change to `_ConversationInfoBase` for the field to appear in API responses. The mapping from source model to wire shape should be explicit and type-checked rather than relying on `model_dump` + silent Pydantic ignore.

A minimal step would be a typed `classmethod` constructor on `ConversationInfo` that takes `(state: ConversationState, stored: StoredConversation, ...)` and maps fields explicitly — making any missing field a type error rather than a silent drop. A more structural step would be replacing the re-declared fields with typed references to the source models, though that would change the wire shape and require client updates.

### Acceptance Criteria

- [ ] Adding a new field to `ConversationState` that should appear in API responses does not require a separate change to `_ConversationInfoBase`
- [ ] The mapping from `ConversationState` / `StoredConversation` fields to `ConversationInfo` fields is explicit and surfaced at compile/lint time, not discovered at runtime

### Alternatives Considered

Keep the current approach but add a test that asserts every public field on `ConversationState` either appears in `ConversationInfo` or is explicitly excluded. This is lower effort but still manual maintenance.

The four fields in `ConversationInfo` that come from neither source model (`current_model_id`, `available_models`, `supports_runtime_model_switch`, `sub_conversation_ids`) are computed at assembly time from ACP agent runtime state and the server catalog — any solution needs to handle these as explicit extra parameters to the constructor rather than derived from the two source models.

The existing `_compose_conversation_info(stored, state, ...)` function is already the right shape; the issue is that `ConversationInfo` is a flat re-declaration rather than being structurally tied to its sources.

KevinAndrewDong:
HUMAN:

ConversationInfo has some duplicated fields today. This makes it easy for the API response to get out of sync with ConversationState. This change gives those shared fields one source of truth and makes the mapping clearer.

---

AGENT:

## Why

`ConversationInfo` is assembled from multiple source models. Passing an unrestricted `ConversationState.model_dump()` into Pydantic can silently discard a newly added state field that is not declared on `ConversationInfo`, allowing the public response to drift without a test failure.

## Summary

- Keep the existing `ConversationState(OpenHandsModel)` hierarchy unchanged.
- Make `ConversationInfo.from_sources()` the explicit assembly boundary for state, stored metadata, and runtime-derived fields.
- Classify every `ConversationState` field as either public or intentionally internal; the contract test fails when a new field is left unclassified.
- Reuse `ConversationInfo.STORED_METADATA_FIELDS` for both response assembly and cache invalidation.
- Preserve JSON-mode serialization for nested secret-bearing values and retain the existing wire shape.

## Issue Number

Fixes #4849

## How to Test

- `uv run pytest tests/agent_server/test_conversation_info_model.py -q` (22 passed)
- `uv run pytest tests/agent_server/test_conversation_info_model.py tests/agent_server/test_conversation_service.py tests/agent_server/test_openapi_contract.py tests/agent_server/test_openapi_discriminator.py -q` (151 passed)
- `uv run pre-commit run --files openhands-agent-server/openhands/agent_server/models.py openhands-agent-server/openhands/agent_server/conversation_service.py tests/agent_server/test_conversation_info_model.py` (all hooks passed)

## Video/Screenshots

Not applicable; this changes response assembly without changing the response fields.

## Design Doc

Not included; the change is limited to explicit source mapping and contract coverage.

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This revision intentionally avoids changing the SDK state model or its inheritance hierarchy.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -458,10 +458,6 @@ def _compose_conversation_info(
     state: ConversationState,
     sub_conversation_ids: list[UUID] | None = None,
 ) -> ConversationInfo:
-    # Use mode='json' so SecretStr in nested structures (e.g. LookupSecret.headers,
-    # agent.agent_context.secrets) serialize to strings. Without it, validation
-    # fails because ConversationInfo expects dict[str, str] but receives SecretStr.
-    #
     # ACP model state is lifted onto top-level ConversationInfo fields because
     # the agent holds it in PrivateAttrs (ACPAgent is frozen) which don't survive
     # ``model_dump``. ``getattr`` keeps non-ACP agents a no-op. We read the live
@@ -514,21 +510,13 @@ def _compose_conversation_info(
     supports_runtime_model_switch = bool(
         agent_state.get("acp_supports_runtime_model_switch", False)
     )
-    return ConversationInfo(
-        **state.model_dump(mode="json"),
-        title=stored.title,
-        metrics=stored.metrics,
-        created_at=stored.created_at,
-        updated_at=stored.updated_at,
-        forked_from_conversation_id=stored.forked_from_conversation_id,
-        forked_from_event_id=stored.forked_from_event_id,
-        parent_conversation_id=stored.parent_conversation_id,
-        sub_conversation_ids=sub_conversation_ids or [],
+    return ConversationInfo.from_sources(
+        state,
+        stored,
         current_model_id=current_model_id,
         available_models=available_models,
         supports_runtime_model_switch=supports_runtime_model_switch,
-        client_tools=stored.client_tools,
-        launched_agent_profile=stored.launched_agent_profile,
+        sub_conversation_ids=sub_conversation_ids or [],
     )
 
 
--- a/openhands-agent-server/openhands/agent_server/conversation_service.py
+++ b/openhands-agent-server/openhands/agent_server/conversation_service.py
@@ -601,23 +601,12 @@ def _stored_metadata_signature(stored: StoredConversation) -> int:
 
     ``cached_info`` is keyed by ``base_state.json``, but also embeds
     ``StoredConversation`` metadata that can change independently via
-    ``meta.json`` (notably auto-title). Fingerprint exactly the fields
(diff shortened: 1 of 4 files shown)


enyst (changes requested):
Thank you for the PR @KevinAndrewDong !

I read it twice and frankly this is a possible solution… though it changes the `ConversationState` inheritance which is a deep change, and I’m not sure why isn’t an alternative possible, such as maybe reading `ConversationInfo` from the other objects… Please give me a moment and let’s think how this mess could be solved 🤔

KevinAndrewDong:
Thanks for the review — I reworked the PR in 7723ba9 to avoid changing the SDK model hierarchy. `ConversationState` is now identical to `main` again, with no `ConversationPublicState` base. `ConversationInfo.from_sources()` explicitly projects classified state fields and shared stored-metadata fields, while a contract test requires every `ConversationState` field to be classified as public or intentionally internal. The JSON wire regression, including secret redaction, remains in place, and the branch is rebased onto current `main` with `runtime_info` accounted for. Focused tests (22) and the related Agent Server/OpenAPI set (151) pass, as does pre-commit.

Issue #4454, opened 2026-08-10
MCP server connection failure at startup aborts the whole conversation (should degrade, not 500)


smolpaws:
### Summary

If an MCP server is unreachable when a conversation starts, the **entire conversation creation fails** with a 500 (`MCPError: MCP Connection Failure`). A single bad/unreachable MCP server takes down the whole conversation instead of the agent starting normally without that server's tools.

MCP servers are optional, best-effort tool sources. A failure to connect to one at startup should be **reported (logged / surfaced as a non-fatal warning) and skipped**, letting the conversation start with the tools that *did* load — not abort the conversation.

### Observed behavior

Starting a new conversation whose agent has an MCP server that can't be reached returns HTTP 500:

```
POST /api/conversations -> 500
{"detail":"Internal Server Error","exception":"MCP Connection Failure","error_id":"…",
 "traceback":"… openhands/sdk/mcp/client.py, line 105, in sync_close … "}
```

Reproduced in Agent Canvas: a new conversation with an agent profile whose skills/config pull in an MCP server that isn't reachable fails at creation. (Notably, ACP agents forward MCP to the subprocess and are more tolerant, so the same MCP being down only breaks the SDK-native agent path — the conversation never starts.)

### Root cause (code)

`create_mcp_tools()` builds a single `MCPClient` for **all** configured servers and connects eagerly during agent initialization. Any connection error propagates and aborts creation:

- `openhands-sdk/openhands/sdk/mcp/client.py` — `MCPClient.connect()` wraps a failed `__aenter__` and raises:
  ```python
  async def connect(self) -> None:
      try:
          await self.__aenter__()
      except RuntimeError as exc:
          raise MCPError("MCP Connection Failure") from exc
  ```
- `openhands-sdk/openhands/sdk/mcp/utils.py` — `create_mcp_tools()` runs `_connect_and_list_tools` and re-raises on timeout (`MCPTimeoutError`) or any `BaseException` after closing the client. There is no per-server isolation: one failing server fails the whole call.
- This propagates up through the agent's `_initialize()` (`openhands-sdk/openhands/sdk/agent/base.py`) into conversation start, so the API returns 500 and no conversation is created.

The `sync_close`/anyio frames people see in the traceback are the cleanup after the initial connect already failed — not the root cause.

### Expected behavior

A startup MCP connection/timeout failure should be **non-fatal**:

- Log a clear warning naming the server(s) that failed and why.
[...]

PR #4895, opened 2026-09-08
Tolerate unavailable MCP servers during agent initialization

### LINKED ISSUE #4891: MCP connection failure aborts send_message before the LLM is called
### Summary

When a configured MCP server is unreachable, `create_mcp_tools()` raises `MCPError` during agent initialization, which propagates all the way out of `send_message()`. The user's message never reaches the LLM and the conversation is dead. An optional tool source being unavailable should degrade, not abort.

### Version

`openhands-agent-server` 1.26.0-python, run as a Docker sandbox behind the legacy Local GUI app server.

### Call path

```
sockets.py:313 events_socket
  event_service.py:455 send_message
    local_conversation.py:888 send_message
      local_conversation.py:703 _ensure_agent_ready
        agent.py:367 init_state
          base.py:504 init_state
            base.py:531 _initialize
              mcp/utils.py:61 create_mcp_tools
                mcp/utils.py:36 _connect_and_list_tools
                  mcp/client.py:55 connect
openhands.sdk.mcp.exceptions.MCPError: MCP Connection Failure
```

Underlying cause: `httpx.ConnectError: All connection attempts failed` (the MCP base URL pointed at a host:port the sandbox container cannot reach).

### Impact

- Every prompt in the conversation fails. The UI shows only "MCP Connection Failure" or a generic "An error occurred".
- The agent is otherwise fully healthy: WebSocket connected, 61 skills loaded, LLM profile saved, git tools working.
- The failure is silent about *which* server failed and *what URL* it tried, which makes it hard to diagnose.

### Secondary issue

In the traceback, `mcp/client.py:97 sync_close` raises during cleanup of the failed connection, so the teardown exception is chained onto the original. That makes the root cause harder to read than it needs to be.

### Suggested behaviour

1. Treat an unreachable MCP server as non-fatal by default: log a warning naming the server and the URL, drop its tools, and continue agent initialization.
2. Optionally keep strict behaviour behind a flag for setups that genuinely require MCP.
3. Include the server name and attempted URL in the `MCPError` message.
4. Suppress or chain the `sync_close` exception so the original `ConnectError` stays the visible cause.

### Actual Behavior

Reproducible directly against the SDK:

```bash
python -c "
from openhands.sdk.mcp.utils import create_mcp_tools
create_mcp_tools({'mcpServers': {'broken': {'url': 'http://127.0.0.1:59999/mcp/mcp'}}})
"
```

Raises `openhands.sdk.mcp.exceptions.MCPError: MCP Connection Failure` out of `mcp/client.py:55 connect`, caused by `httpx.ConnectError: All connection attempts failed`. The message names neither the server nor the URL.

In a real conversation the same exception propagates through `base.py:531 _initialize` → `agent.py:367 init_state` → `local_conversation.py:703 _ensure_agent_ready` → `local_conversation.py:888 send_message`, so the user's message never reaches the LLM and every subsequent prompt fails.

Note the asymmetry inside `create_mcp_tools()` itself: the `except TimeoutError` branch already lists the configured server names and prints troubleshooting guidance before raising `MCPTimeoutError`. Connection failures get no equivalent handling and propagate raw.

Relevant tests:

```bash
pytest tests/ -k mcp
```

### Acceptance Criteria

- [ ] An unreachable MCP server logs a warning naming the server and the attempted URL instead of raising by default
- [ ] `send_message()` completes and the LLM is called when a configured MCP server is unreachable
- [ ] Tools from reachable MCP servers are still registered when another server in the same config fails
- [ ] Connection failures get the same server-name and guidance treatment as the existing `TimeoutError` branch in `create_mcp_tools()`
- [ ] Strict fail-fast behaviour remains available behind an opt-in flag
- [ ] The `sync_close()` teardown exception no longer masks the original `ConnectError` in the traceback
- [ ] A test covers agent initialization with one reachable and one unreachable MCP server

Happy to open a PR if the maintainers agree on the approach.

jstar0:
HUMAN:

Testing evidence: the local MCP regression, the mixed reachable and unreachable server case, and a real LocalConversation through TestLLM were run; the results are recorded below.

AGENT:

This pull request was prepared by an AI agent (OpenHands) on behalf of jstar0.

## Why

An unreachable optional MCP server currently raises `MCPError` while the agent is being initialized. That prevents `send_message()` from reaching the LLM, and the failure message does not identify the server or endpoint. Cleanup can also raise while handling the connection failure.

## Summary

- Treat MCP connection failures and timeouts as non-fatal by default. Keep the `/api/mcp/test` probe in strict mode so failed connection checks remain visible, and retain opt-in `strict=True` fail-fast behavior for callers that require every MCP server.
- Include server names, redacted remote URLs or stdio commands, the original connection detail, and actionable guidance in connection-failure diagnostics.
- Make cleanup tolerate `BaseException` values and add regressions for optional fallback, mixed-server discovery, `LocalConversation` startup, strict mode, and cleanup preservation.

## Issue Number

Fixes #4891

## How to Test

From the repository root:

```bash
uv run pytest -q tests/sdk/mcp/test_create_mcp_tool.py
uv run pytest -q tests/sdk/mcp tests/sdk/conversation/test_local_conversation_mcp.py tests/sdk/conversation/test_local_conversation_plugins.py tests/cross/test_agent_loading.py
uv run pre-commit run --files openhands-sdk/openhands/sdk/mcp/client.py openhands-sdk/openhands/sdk/mcp/utils.py tests/sdk/mcp/test_create_mcp_tool.py
```

Results:

- MCP creation regressions: `34 passed`
- MCP, conversation wiring, plugin, and agent-loading tests: `190 passed`
- Ruff format, Ruff lint, pycodestyle, Pyright, dependency rules, and Tool registration: passed
- Follow-up checks for the review fixes: `uv run --frozen pytest tests/sdk/mcp/test_create_mcp_tool.py tests/agent_server/test_mcp_router.py` (`59 passed`); applicable pre-commit hooks passed for the four changed files.

Before this change, the issue reproduction against the parent revision raised `MCPError: MCP Connection Failure` during agent initialization, before the LLM was called. After this change, the same configuration returns an empty `MCPClient`, emits a warning naming `broken` and the attempted endpoint, and a real `LocalConversation` completes with `TestLLM.call_count == 1`. A reachable server in the same configuration still contributes its tools.

## Video/Screenshots

Not applicable: this is an SDK connection and lifecycle behavior change. The `LocalConversation` regression exercises the end-to-end agent path.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The default behavior is intentionally best-effort for optional MCP sources. Callers that require every configured MCP server can pass `strict=True` to `create_mcp_tools()`. URL query parameters are redacted in diagnostics.

### DIFF
--- a/openhands-sdk/openhands/sdk/mcp/client.py
+++ b/openhands-sdk/openhands/sdk/mcp/client.py
@@ -103,7 +103,7 @@ class MCPClient(AsyncMCPClient):
         if hasattr(self, "close") and inspect.iscoroutinefunction(self.close):
             try:
                 self._executor.run_async(self.close, timeout=10.0)
-            except Exception:
+            except BaseException:  # noqa: BLE001 - cleanup must never mask the original error
                 pass  # Ignore close errors during cleanup
 
         # Always cleanup the executor
--- a/openhands-sdk/openhands/sdk/mcp/utils.py
+++ b/openhands-sdk/openhands/sdk/mcp/utils.py
@@ -22,8 +22,9 @@ from openhands.sdk.mcp.config import (
     enabled_mcp_servers,
     to_fastmcp_mcp_config,
 )
-from openhands.sdk.mcp.exceptions import MCPTimeoutError
+from openhands.sdk.mcp.exceptions import MCPError, MCPTimeoutError
 from openhands.sdk.mcp.tool import MCPToolDefinition
+from openhands.sdk.utils.redact import redact_url_params
 
 
 logger = get_logger(__name__)
@@ -46,6 +47,7 @@ class MCPToolProvider(Protocol):
         mcp_config: dict[str, MCPServer],
         timeout: float = 30.0,
         *,
+        strict: bool = False,
         on_tools_changed: ToolsChangedCallback | None = None,
         on_tools_reconciled: ToolsReconciledCallback | None = None,
     ) -> MCPClient: ...
@@ -59,12 +61,14 @@ class DefaultMCPToolProvider:
         mcp_config: dict[str, MCPServer],
         timeout: float = 30.0,
         *,
+        strict: bool = False,
         on_tools_changed: ToolsChangedCallback | None = None,
         on_tools_reconciled: ToolsReconciledCallback | None = None,
     ) -> MCPClient:
         return create_mcp_tools(
             mcp_config,
             timeout,
+            strict=strict,
             on_tools_changed=on_tools_changed,
             on_tools_reconciled=on_tools_reconciled,
         )
@@ -312,10 +316,43 @@ class _ToolListChangedHandler(MessageHandler):
             )
 
 
+def _server_target(name: str, server: MCPServer) -> str:
+    if server.url is not None:
+        return f"{name!r} at {redact_url_params(server.url)!r}"
+    if server.command is not None:
+        return f"{name!r} using command {server.command!r}"
+    return repr(name)
+
+
+def _connection_failure_message(
+    mcp_config: Mapping[str, MCPServer], error: BaseException
+) -> str:
+    targets = ", ".join(
+        _server_target(name, server) for name, server in mcp_config.items()
+    )
+    detail = str(error.__cause__ or error)
+    return (
+        f"Failed to connect to MCP server(s): {targets}. {detail}\n"
+        "Possible solutions:\n"
+        "  1. Check if the MCP server is running and responding\n"
+        "  2. Verify network connectivity to the MCP server"
+    )
+
+
+def _close_client_quietly(client: MCPClient) -> None:
+    try:
+        client.sync_close()
+    except BaseException as close_error:  # noqa: BLE001 - cleanup must not mask the original error
+        logger.debug(
+            "Failed to close MCP client during error cleanup", exc_info=close_error
+        )
+
+
 def create_mcp_tools(
     mcp_config: dict[str, MCPServer],
     timeout: float = 30.0,
     *,
+    strict: bool = False,
     on_tools_changed: ToolsChangedCallback | None = None,
     on_tools_reconciled: ToolsReconciledCallback | None = None,
     mcp_oauth_token_storage: AsyncKeyValue | None = None,
@@ -329,6 +366,10 @@ def create_mcp_tools(
             for tool in client.tools:
                 # use tool
         # Connection automatically closed
+
+    By default, an unavailable server is logged and skipped so an optional
+    MCP server cannot prevent the agent from starting. Pass ``strict=True``
+    to preserve fail-fast behavior for configurations that require MCP.
 
     The client subscribes to ``notifications/tools/list_changed`` and
     reconciles its tool list whenever the server signals a change. When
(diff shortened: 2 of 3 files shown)


smolpaws:
🐾 Hi, smolpaws here (small OpenHands agent). I reviewed this — it cleanly fixes #4891 and also my earlier #4454. Nice work.

**What's good**
- Tolerant-by-default with an opt-in `strict=True` — exactly the right shape for an optional tool source.
- Diagnostics now name the server + redacted endpoint (`redact_url_params`), and the test asserts the secret doesn't leak into the warning. 👍
- `_close_client_quietly` + the `except BaseException` → `BaseException` change stop cleanup from masking the original `ConnectError`. The `sync_close` cancellation-style test is a good touch.
- Tests are thorough: skip-with-warning, strict fail-fast, healthy+broken mix, a real `LocalConversation` reaching the LLM, and cleanup-doesn't-mask.

**One thing worth a look**
- The `except TimeoutError` branch still raises `MCPTimeoutError` unconditionally — it isn't gated by `strict`. So an unreachable server that *hangs* (slow timeout) still aborts init, while one that *refuses* (ConnectError) degrades gracefully. That's an inconsistent user experience for the same "server unavailable" situation. Consider routing the timeout path through the same tolerant/`strict` logic so both failure modes degrade by default.

Small nit: `_connection_failure_message` builds `targets` from the whole `mcp_config`, so if several servers are configured and only one is down, the message lists them all as failed. Minor, since it's a warning.

Not a blocker either way — the timeout point is the only one I'd genuinely consider before merge. 🐾

jstar0:
Follow-up: pushed 1d63668485956c145e31eec3b8632553b854c852 to address the hosted pre-commit Pyright failure. The injected MCPToolProvider protocol is intentionally kept minimal: strict remains available on create_mcp_tools() and DefaultMCPToolProvider, while LocalConversation only relies on the callback parameters. This preserves structural compatibility for existing custom providers without changing strict-mode behavior. Validation on the updated head: uv run pre-commit run --all-files --show-diff-on-failure passed; MCP creation regressions 34 passed; MCP, conversation wiring, plugin, and agent-loading tests 190 passed. The PR description and HUMAN section were not changed.

smolpaws:
🐾 One follow-up thought: does `strict` earn its keep?

Both #4891 and #4454 only *floated* a strict mode — neither had a real need for it. "Continue but inform" is the right default for an optional tool source, and a single behavior is easier to reason about than two. A caller who genuinely requires server X can just check the returned client for that tool and fail on their side — that's their policy, not the library's.

If `strict` goes, the one thing to keep is a **structured** "what failed" (server + reason: refused / timeout / auth) on the client, so that caller-side check is easy — not just a log line.

Nice side effect: it also dissolves the `TimeoutError`-vs-`ConnectError` inconsistency I mentioned, since both would just degrade-and-inform.

Your call, of course — just a simplification worth weighing. 🐾

jstar0:
Thanks for the review. Addressed both observations in f41600bba:

- `_probe_mcp_server` now calls `create_mcp_tools(..., strict=True)`, preserving structured failures for `/api/mcp/test`.
- Optional MCP timeouts now log a diagnostic warning and continue by default; `strict=True` still raises `MCPTimeoutError`.

Validation: `uv run --frozen pytest tests/sdk/mcp/test_create_mcp_tool.py tests/agent_server/test_mcp_router.py` passed (59 tests), and the applicable pre-commit hooks passed for the four changed files.

PR #4839, opened 2026-09-03
fix(extensions): compose local source with repo_path

VascoSch92:
HUMAN:

BE needed to install extensions from local path
Edit: this will apply the paths to any client applications 

---

AGENT:

## Why

Installing a canvas extension by pointing Source at a repository and Path at the
extension's subdirectory always fails. `fetch_with_resolution()` rejects
`repo_path` outright for `SourceType.LOCAL` before the subpath is ever applied,
so the composition that already works for git and GitHub sources is unreachable
for a local checkout.

The failure is also unreadable. The install route catches `ExtensionFetchError`
and replaces it with one fixed sentence about the source, so a typo in Path is
reported as a bad Source. On `origin/main` six of seven distinct mistakes return
the exact same message.

Reported as OSS-10405 and as OpenHands/OpenHands#17048.

## Summary

- `fetch.py`: local sources now compose with `repo_path` the same way git sources
  do, via `_apply_subpath()`, instead of raising.
- `_apply_subpath()` gained containment. A subpath that climbs out of the base
  through `..` or a symlink is rejected rather than resolved.
- `canvas_extensions_router.py`: the install route forwards the specific fetch
  reason (credential-redacted) and answers a missing manifest with its own 422
  telling the user that Source plus Path must name the extension's own directory.
- That 400 is worded "Could not read", not "Failed to fetch". The canvas client
  treats any message containing "failed to fetch" as a network outage and
  replaces it with a "Disconnected, check your network" toast, which hid the
  reason. A test asserts the phrase stays out.

## Issue Number

OpenHands/OpenHands#17048, Linear OSS-10405.

## How to Test

The ticket's own reproduction, run end to end against the real installer
(`install_canvas_extension`), with a fixture repository holding
`canvas-extensions/canvas-pulse/canvas-extension.json`:

```
$ uv run --frozen python evidence.py
```

On `origin/main` (e26683288):

```
  FAIL  Source=/path/canvas-extensions/  Path=/canvas-pulse
          ExtensionFetchError: repo_path is not supported for local extension sources...
  FAIL  Source=/path/canvas-extensions/  Path=canvas-pulse
          ExtensionFetchError: repo_path is not supported for local extension sources...
  FAIL  Source=/path/canvas-extensions   Path=/canvas-pulse
          ExtensionFetchError: repo_path is not supported for local extension sources...
  FAIL  Source=/path/canvas-extensions   Path=canvas-pulse
          ExtensionFetchError: repo_path is not supported for local extension sources...
[...]

PR #4836, opened 2026-09-02
fix(llm): PLTF-3531 register litellm_proxy alias pricing so spans aren't silently $0

juanmichelini:
HUMAN:

- [x] this was tested by a human, Ai.

---

AGENT:

## Why

A custom `litellm_proxy/*` model whose public name is not a litellm-known model id (e.g. `litellm_proxy/prod/claude-sonnet-4-5-bedrock`) was silently priced `$0` on its LLM span in two independent places, so Laminar cost dashboards underreported real spend (#4816). The two defects:

1. **Alias not priceable.** litellm strips the `litellm_proxy/` prefix and parses the first path segment (`prod`) as the **provider**, so both `get_llm_provider` and `cost_per_token` raise and the SDK never registered the alias's pricing into litellm's global `model_cost` map.
2. **Cost never reached the span.** Even once priceable, lmnr's LiteLLM instrumentation ends the `litellm.completion` span **inside** the completion call (ahead of `Telemetry.on_response`), so `gen_ai.usage.cost` was never attached.

## Summary

- **Alias pricing registration (`model_info.py`).** When resolving model info for a `litellm_proxy/*` deployment, the SDK now also `litellm.register_model({<alias>: {...}})` the alias into the global `model_cost` map, with pricing derived from the deployment's underlying model plus any proxy-side overrides. Already-priceable aliases (direct `model_cost` keys or provider-prefixed real models) are skipped so native pricing is never clobbered; when no pricing can be derived the alias is left unregistered with a debug log rather than registering a silent `$0` entry.
- **Provider label normalization (`model_info.py`, ai-vong's field-test fix).** The proxy's `/v1/model/info` reports `litellm_provider: "bedrock_converse"`, which `get_llm_provider` cannot resolve for an unknown alias id (only `bedrock` is accepted). We normalize `bedrock_converse` → `bedrock` before `register_model` so `cost_per_token` resolves. Pricing overrides still win — only the provider label changes.
[...]

PR #4911, opened 2026-09-08
fix(tests): walk through the llm.<model> span in the auto-title trace assertion

simonrosenberg:
HUMAN:

Verified the span tree myself and confirmed the tracing is correct — it's the assertion that went stale, not the instrumentation.

---

AGENT:

## Why

`main` is red. `agent-server-tests` fails on every PR that touches an agent-server path:

```
tests/agent_server/test_auto_title_span_metadata.py::test_auto_title_llm_span_joins_the_conversation_trace
E   AssertionError: assert 'llm.gpt-4o' == 'conversation.generate_title'
```

#4836 gave `Telemetry.on_request` its own `llm.<model>` span, which stays current while the call runs so the authoritative cost can be attached before lmnr ends its `litellm.completion` span. That span now sits between the operation span and the LLM span:

```
conversation.generate_title
  └── llm.gpt-4o            ← added by #4836
        └── litellm.completion
```

**The tracing is correct.** The title LLM span is still in the conversation trace and still under `conversation.generate_title`; only the direct-parent assertion was stale.

`tests/sdk/conversation/test_utility_llm_span_metadata.py` was updated for exactly this in the same PR — it grew an `operation_parent()` helper that skips the `llm.` wrapper, plus a separate `immediate_parent`. The agent-server counterpart was missed.

### How it reached main past a required check

Worth recording, because it is not obvious and it is why #4836's author never saw this test fail.

`agent-server-tests` is path-gated — every step is `if: steps.changed.outputs.any_changed == 'true'` over `openhands-agent-server/**`, `tests/agent_server/**`, `pyproject.toml`, `uv.lock`, `.github/workflows/tests.yml`. **When nothing matches, the job executes nothing and still reports `success`.**

| step | `fc9646ad8` (#4836) | `56b016e7f` (#4839) |
|---|---|---|
| Install deps | skipped | success |
| Run Agent Server tests | **skipped** | **failure** |
| job conclusion | **success** | failure |

#4836's diff is `openhands-sdk/**` + `tests/sdk/**` only, so the job was green having run nothing. #4839 touched agent-server paths, opening the gate, and so became the first commit to actually run this test since #4836 landed. **#4836 is the cause; #4839 is only the detector.**

The filter gap itself is broader than this PR fixes and is tracked separately in #4912.

## Summary

- `parent` now walks through the synthetic `llm.<model>` wrapper to the operation the call belongs to, matching the `operation_parent()` helper the sibling SDK test already uses rather than inventing a second convention.
- `immediate_parent` is added and asserted, so the test still fails if the LLM span detaches or the wrapper disappears silently.
- The `trace_id` and `title_generation` operation-attribute assertions are unchanged. No production code changes.

## Issue Number

Related to #4912 — this PR fixes the test breakage reported there; the CI path-filter gap in that issue is a separate change.

## How to Test

```bash
uv run pytest tests/agent_server/test_auto_title_span_metadata.py
```

Fails on `main`, passes on this branch. Verified it is not a local artifact by reproducing on a clean detached worktree at `fc9646ad8` with none of this branch's changes present — deterministic, 3/3 runs.

Full agent-server suite on this branch:

```
2101 passed, 13 deselected in 216.60s (0:03:36)
```

The span tree the probe actually emits, which is what the new assertions encode:

```
litellm.completion            parent=llm.gpt-4o                    same_trace=True
llm.gpt-4o                    parent=conversation.generate_title   same_trace=True
conversation.generate_title   parent=None                          same_trace=True
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Test-only change; no production code touched. Unblocks every PR whose diff opens the `agent-server-tests` gate, including #4864.

### DIFF
--- a/tests/agent_server/test_auto_title_span_metadata.py
+++ b/tests/agent_server/test_auto_title_span_metadata.py
@@ -109,13 +109,36 @@ def _probe_auto_title_spans() -> dict[str, Any]:
     names_by_id = {
         span.context.span_id: span.name for span in spans if span.context is not None
     }
+    parent_id_by_id = {
+        span.context.span_id: (span.parent.span_id if span.parent else None)
+        for span in spans
+        if span.context is not None
+    }
+
+    def operation_parent(span: Any) -> str | None:
+        """Nearest ancestor that isn't the SDK's synthetic ``llm.<model>`` span.
+
+        Telemetry.on_request opens an ``llm.<model>`` span that stays current
+        while the call runs (so the authoritative cost can be attached before
+        lmnr ends its ``litellm.completion`` span). That span sits between the
+        operation span and ``litellm.completion``; skip it to recover the
+        operation the LLM call belongs to.
+        """
+        pid = span.parent.span_id if span.parent else None
+        while pid is not None and names_by_id.get(pid, "").startswith("llm."):
+            pid = parent_id_by_id.get(pid)
+        return names_by_id.get(pid) if pid is not None else None
+
     return {
         "title": stored.title,
         "conversation_trace_id": root_span.span.get_span_context().trace_id,
         "spans": [
             {
                 "name": span.name,
-                "parent": names_by_id.get(span.parent.span_id) if span.parent else None,
+                "parent": operation_parent(span),
+                "immediate_parent": (
+                    names_by_id.get(span.parent.span_id) if span.parent else None
+                ),
                 "trace_id": span.context.trace_id if span.context else None,
                 "attributes": dict(span.attributes or {}),
             }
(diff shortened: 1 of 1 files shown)


simonrosenberg:
Closing as redundant — this was fixed independently in #4908 (Release v1.46.0), which landed on main as `6a1e4d0f0` while this was waiting for review.

Their fix asserts the two-level chain explicitly:

```python
sdk_llm_spans = [span for span in probe["spans"] if span["name"] == "llm.gpt-4o"]
assert sdk_llm["parent"] == "conversation.generate_title"
assert title_llm["parent"] == "llm.gpt-4o"
```

Verified main is green: `uv run pytest tests/agent_server/test_auto_title_span_metadata.py` passes at `6a1e4d0f0`. Same conclusion this PR reached — the tracing was correct and only the assertion was stale — so there is nothing left to fix and no reason to churn the file again.

One difference worth noting for whoever touches this next, not worth a PR on its own: their version hard-codes the span name `llm.gpt-4o`, so it will need editing if the probe's model fixture ever changes, and it diverges from the `operation_parent()` helper that `tests/sdk/conversation/test_utility_llm_span_metadata.py` already uses for the same wrapper. Fine as-is; just two conventions for one thing now.

**#4912 stays open and is unaffected** — the path-filter gap that let this reach main with every required check green is still there. That is the part with lasting value here.

PR #4938, opened 2026-09-10
llm: skip retries for hard quota errors, fall back immediately

### LINKED ISSUE #4936: Hard quota exhaustion delays fallback LLM behind retry backoff
### Bug Description

When the primary LLM returns a deterministic account or usage allowance error such as `usage_limit_reached` or `insufficient_quota`, the SDK retries the same model through its full exponential backoff before consulting a configured `FallbackStrategy`. These errors cannot recover until the quota resets or is raised, so retries delay fallback by minutes without any chance of success.

### Actual Behavior

Configure an `LLM` with `num_retries=3` and a `FallbackStrategy`, then make the LiteLLM completion transport raise a `RateLimitError` whose provider payload contains `usage_limit_reached`. Run the reproduction from the repository environment with `uv run python /tmp/repro_quota_fallback.py`. The primary transport is called for every configured retry before the fallback transport is attempted, even though repeating the same hard-quota request cannot succeed.

The production incident motivating this issue returned `usage_limit_reached` for a team-plan model and retried four times before the review automation errored instead of immediately using its configured fallback.

### Expected Behavior

Hard quota-exhaustion errors should leave the retry loop immediately and reach the configured fallback model on the first failure. Ordinary transient rate limits should continue to use the existing retry policy.

### Acceptance Criteria

- [ ] Hard quota-exhaustion signals skip retries.
- [ ] A configured fallback is attempted immediately.
- [ ] Transient 429/rate-limit errors retain existing retry behavior.
- [ ] Classifier and fallback behavior have regression tests.

This issue was created by an AI agent (OpenHands) on behalf of @enyst while addressing PR #4917.

Christian-Sidak:
HUMAN:

Verified the classifier logic manually with Python assertions covering all four hard-quota patterns and the two negative cases (transient 429 and a non-RateLimitError carrying the same text). Syntax-checked all modified files. The retry predicate change ensures hard-quota errors are not retried while transient ones still go through the full backoff loop.

---

AGENT:

## Why

A deterministic hard quota-exhaustion error (e.g. `usage_limit_reached`, `insufficient_quota`) cannot recover within the SDK's retry window. Burning through all exponential-backoff attempts delays an otherwise available fallback model by minutes for no benefit.

## Summary

- Adds `is_hard_quota_error()` classifier that detects hard quota signals inside `RateLimitError` payloads.
- Threads the new predicate into `RetryMixin.retry_decorator` so those errors skip all retry attempts and propagate immediately to `_handle_error`, which then consults the configured `FallbackStrategy` on the first failure.
- Transient per-minute 429 rate limits (no quota keyword) retain their existing retry behaviour unchanged.

## Issue Number

Fixes #4936

## How to Test

Run the focused regression suite:

    uv run pytest tests/sdk/llm/test_llm_fallback.py -k "quota" -v

The tests drive the public `LLM.completion()` path with `num_retries=3`: a `usage_limit_reached` primary error must produce exactly one primary attempt followed by one successful fallback attempt. A paired transient-rate-limit case confirms recoverable 429s still retry before fallback.

## Video/Screenshots

Not applicable; this is backend retry behavior with no UI change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


enyst:
Thank you for the PR @Christian-Sidak , I think we already had a PR and merged it, so the issue is closed. Could you tell what does your PR add or fix, or is it redundant?

enyst (changes requested):
I appreciate the PR but it seems to duplicate existing functionality?

Christian-Sidak:
Thanks for checking, @enyst. You are right that #4917 covers the same ground (immediate fallback on hard quota exhaustion). Closing this as redundant.

PR #4962, opened 2026-09-11
fix(terminal): make Windows PowerShell completion and interrupt recovery robust

### LINKED ISSUE #4963: [Bug]: Windows PowerShell terminal is permanently poisoned after a terminating error or interrupt

### Bug Description

On native Windows, a persistent PowerShell terminal session (`WindowsTerminal`)
can become permanently poisoned by a single command that ends with a
terminating error or is interrupted. After that, every subsequent command is
rejected with "the previous command is still running", so an agent cannot make
progress. This corresponds to the downstream report OpenHands/OpenHands#17198
("[Command finished execution with no output]").

Root cause is Windows-only and two-fold:

1. The command-completion sentinel (`###PS1END###` metadata block) is appended
   to the same PowerShell statement as the user command, so a terminating
   error (`throw`, `-ErrorAction Stop`, `$ErrorActionPreference='Stop'`) aborts
   the line before the sentinel runs and completion is never detected.
2. The shell is spawned with `CREATE_NEW_PROCESS_GROUP`, which disables Ctrl+C
   for the shell and its children, so interrupts never reach the running
   command and child process trees survive.

### Expected Behavior

After a command exits, fails, times out, or is interrupted, the terminal
recovers and the next command runs. One bad command must not poison the session.

### Actual Behavior

Reproducible with a plain SDK script (no LLM), on native Windows, using
`python`:

```
python -c "from openhands.tools.terminal.definition import TerminalAction; from openhands.tools.terminal.terminal import create_terminal_session as c; s=c(work_dir='.', terminal_type='powershell', no_change_timeout_seconds=1); s.initialize(); s.execute(TerminalAction(command='Start-Sleep 120')); s.execute(TerminalAction(command='C-c', is_input=True)); print(s.execute(TerminalAction(command='Get-ChildItem')).metadata.suffix)"
```

On stock 1.46.0 the final line prints "...the previous command is still
running...": the session is poisoned and every later command is rejected.
`throw` and `-ErrorAction Stop` reproduce the same poison without any interrupt.

### Steps to Reproduce

1. On native Windows, run the `python` snippet above against
   `openhands-tools` 1.46.0 (or run `uv run pytest tests/tools/terminal` with a
   Windows-only recovery test).
2. Observe the "previous command is still running" rejection on the command
   after the interrupt.

### OpenHands SDK Version

1.46.0 (also present on 1.47.0 / current main; the terminal files are
unchanged between them).

### Installation Method

Agent Server via `uvx` / Agent Canvas 1.17.0 on Windows 11, PowerShell 5.1.

### Acceptance Criteria

- [ ] After an interrupted or terminating-error command on the Windows
      PowerShell backend, the next command executes normally (no "previous
      command is still running").
- [ ] Ctrl+C stops the running command and its child process tree.
- [ ] A shell that exits or stops responding is recreated automatically before
      the next command.
- [ ] Linux/tmux/subprocess behaviour is unchanged and existing
      `tests/tools/terminal` tests still pass.

RedCat888:
HUMAN:

Installed OpenHands/Agent Canvas on bare Windows via npm.

When using OpenHands native agent on windows, (I tried both GLM 5.3-flash via open router and chatgpt sol) constant agent errors occured related to the terminal, and I even had a agent get stuck in a loop of attempting to exit terminal/process via Ctr+C repeatedly, with no real output nor a timeout cancelling the tool actions.

-----

AGENT:

## Why

On native Windows, a persistent PowerShell terminal session (`WindowsTerminal`)
can become **permanently poisoned** by a single command that ends with a
terminating error or is interrupted. Afterwards every command is rejected with
*"the previous command is still running"*, and the agent burns its remaining
turns retrying. This reproduced identically with two different models, so it is a
harness/runtime issue, not model behaviour. It matches the downstream report
OpenHands/OpenHands#17198 ("[Command finished execution with no output]", then
the terminal being considered occupied and repeated Ctrl+C/reset failing).

Root cause (all Windows-only):

1. **Completion sentinel was appended to the command line.** The PS1 metadata
   block was sent as extra statements on the *same* line as the user command, so
   any command that did not run to completion — `throw`, `-ErrorAction Stop`,
   `$ErrorActionPreference = 'Stop'` — aborted the line before the sentinel ran.
   `###PS1END###` never appeared and the session believed the command was still
   running forever.
2. **The shell was spawned with `CREATE_NEW_PROCESS_GROUP`**, which disables
   Ctrl+C for the shell and its children. `CTRL_BREAK_EVENT` to a
   `CREATE_NO_WINDOW` process is dropped and writing `\x03` to stdin does
   nothing, so interrupts never reached the running command and child trees were
   left alive.
3. **No recovery path.** `prev_status` stayed `NO_CHANGE_TIMEOUT`/`HARD_TIMEOUT`
   after the process had exited, and a wedged shell could only be escaped by a
   full `reset` that discarded cwd/env.

## Summary

- Emit the completion sentinel from PowerShell's `prompt` function (like the bash
  backends use `PS1`) so it reappears whenever the shell is ready, however the
  previous command ended.
- Start the shell with `CREATE_NO_WINDOW` only and re-enable Ctrl+C inside it via
  `SetConsoleCtrlHandler`; deliver interrupts with a short-lived helper that
  `AttachConsole()`s and calls `GenerateConsoleCtrlEvent(CTRL_C_EVENT)` (Windows
  `killpg(SIGINT)` equivalent), escalating to job-object process-tree
  termination, all with bounded waits. New `windows_console.py` isolates the
  Win32 `ctypes` plumbing (job objects + Toolhelp32).
- Track command state with a monotonic prompt counter instead of screen-tail
  matching; add `TerminalInterface.is_alive()` so a shell that exited or stopped
  responding is recreated before the next command. `is_alive()` defaults to
  `initialized and not closed` and is only overridden on Windows, so
  tmux/subprocess/Linux behaviour is unchanged.

## Issue Number

Fixes #4963

Also addresses the downstream report OpenHands/OpenHands#17198 (cross-repository,
so no auto-closing keyword there).

## How to Test

Deterministic, no LLM — on native Windows:

```python
from openhands.tools.terminal.definition import TerminalAction
from openhands.tools.terminal.terminal import create_terminal_session

s = create_terminal_session(work_dir=".", terminal_type="powershell",
                            no_change_timeout_seconds=1)
s.initialize()
s.execute(TerminalAction(command="Start-Sleep 120"))     # soft timeout -> exit -1
s.execute(TerminalAction(command="C-c", is_input=True))  # interrupt
print(s.execute(TerminalAction(command="Get-ChildItem")).metadata.suffix)
# stock 1.46.0: "...the previous command is still running..."  (poisoned)
# patched:       normal listing, exit code 0
```

`throw`, `Get-Item <missing> -ErrorAction Stop`, and
`$ErrorActionPreference='Stop'` poison the stock session without any interrupt,
because their sentinel never ran.

Automated checks (run on Windows 11, PowerShell 5.1):

```
uv run pytest tests/tools/terminal        # 197 passed, 154 skipped (Windows-only cases run)
uv run pre-commit run --files <changed>   # ruff-format, ruff, pycodestyle, pyright,
                                          # import-rules, tool-registration: all pass
```

End-to-end evidence (beyond unit tests): the Agent Server was run from this
checkout via `OH_AGENT_SERVER_LOCAL_PATH` behind a real Agent Canvas stack on
native Windows, and two native models each completed a multi-step workload
containing terminating errors, a non-zero exit, an interrupted `Start-Sleep`, and
a top-level `exit` that killed the shell — with **zero** "no output" symptoms,
"still running" wedges, or manual resets; the `exit` case self-healed into a
fresh session. A targeted reproduction suite went from **7/20** (stock 1.46.0) to
**21/21**; a 300-cycle interrupt/recovery stress test showed 0 failures, 0 leaked
processes (powershell/conhost/python/cmd/ping/node), and flat ~0.5 s recovery
latency.

Windows-specific tests are guarded with
`skipif(platform.system() != "Windows")`, so they run only on the Windows CI job;
the send-keys/init-script tests are platform-independent.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Behavioural change is Windows-only; Linux/tmux/subprocess paths are unchanged
(the new `is_alive()` guard never triggers there, and using
`terminal.is_running()` instead of a screen-tail `endswith(PS1_END)` check is
equivalent for those backends). No public API is removed or changed. This touches
terminal execution, which the review guide flags as eval-relevant — noting that
the effect is confined to the native-Windows backend.

### DIFF
--- a/openhands-tools/openhands/tools/terminal/descriptions.py
+++ b/openhands-tools/openhands/tools/terminal/descriptions.py
@@ -88,5 +88,8 @@ WINDOWS_TOOL_DESCRIPTION = "\n".join(
         "  one becomes unresponsive.",
         "* Resetting the terminal clears loaded modules, environment variables,",
         "  working directory changes, and running processes.",
+        "* If the shell exits or stops responding (for example after `C-c` could",
+        "  not stop a command), it is recreated automatically before your next",
+        "  command and the output tells you so.",
     ]
 )
--- a/openhands-tools/openhands/tools/terminal/impl.py
+++ b/openhands-tools/openhands/tools/terminal/impl.py
@@ -437,7 +437,15 @@ class TerminalExecutor(ToolExecutor[TerminalAction, TerminalObservation]):
         conversation: "LocalConversation | None" = None,
     ) -> TerminalObservation:
         """Execute *action* in single-session (non-pool) mode."""
-        if action.reset or self.session._closed:
+        # Self-heal: a session whose shell exited, was killed, or was declared
+        # unresponsive after a failed interrupt is replaced before the command
+        # is accepted, so the model never has to discover that itself.
+        if action.reset or self.session._closed or not self.session.is_alive():
+            if not action.reset and not self.session._closed:
+                logger.warning(
+                    "Terminal session is no longer alive; recreating it before "
+                    "running the next command"
+                )
             reset_result = self._reset_single_session()
 
             if action.command.strip():
--- a/openhands-tools/openhands/tools/terminal/terminal/interface.py
+++ b/openhands-tools/openhands/tools/terminal/terminal/interface.py
@@ -158,6 +158,16 @@ class TerminalInterface(ABC):
             True if a command is running, False otherwise.
         """
 
+    def is_alive(self) -> bool:
+        """Check whether the backend can still accept commands.
+
+        Backends that own a shell process should return ``False`` once that
+        process has exited or stopped responding, so the session controller
+        can recreate the terminal instead of waiting on it.  The default
+        only reflects the initialized/closed flags.
+        """
+        return self._initialized and not self._closed
+
     @property
     def initialized(self) -> bool:
         """Check if the terminal is initialized."""
--- a/openhands-tools/openhands/tools/terminal/terminal/terminal_session.py
+++ b/openhands-tools/openhands/tools/terminal/terminal/terminal_session.py
@@ -48,13 +48,35 @@ def _remove_command_prefix(command_output: str, command: str) -> str:
 
 
 def _remove_powershell_echo(command_output: str, command: str) -> str:
+    """Strip PowerShell's echo of the submitted command from its output.
+
+    When PowerShell reads commands from a pipe it echoes each input line; a
+    multiline command is echoed as its first line followed by one ``>> ``
+    continuation line per extra line (plus the blank line that submits it).
+    """
     command_output = command_output.lstrip()
     command = command.lstrip()
-    first_line = command_output.splitlines()[0] if command_output else ""
-    if command and command in first_line:
-        _, separator, rest = command_output.partition("\n")
-        command_output = rest if separator else ""
+    command_lines = command.splitlines() or [""]
+    lines = command_output.split("\n")
+    if command_lines[0] and command_lines[0] in lines[0]:
+        lines = lines[1:]
+        # Continuation echo: at most one ">> " line per extra command line,
+        # plus the blank line that terminates the multiline input.
+        budget = len(command_lines)
+        while budget and lines and lines[0].startswith(">> "):
+            lines = lines[1:]
+            budget -= 1
+        command_output = "\n".join(lines)
     return re.sub(r"(?:\r?\n)?PS [^\r\n]*>\s*$", "", command_output).lstrip()
+
+
+_UNRESPONSIVE_TERMINAL_SUFFIX = (
+    "\n[The terminal process exited or stopped responding while this command "
+    "was running, so its result is unknown. A new terminal session will be "
+    "created automatically for your next command; environment variables, "
+    "loaded modules and working-directory changes from the old session are "
+    "gone. Avoid top-level `exit` in terminal commands.]"
+)
 
 
 class TerminalSession(TerminalSessionBase):
@@ -136,6 +158,10 @@ class TerminalSession(TerminalSessionBase):
             TerminalCommandStatus.NO_CHANGE_TIMEOUT,
             TerminalCommandStatus.HARD_TIMEOUT,
         }
+
+    def is_alive(self) -> bool:
+        """Check whether the underlying terminal can still accept commands."""
+        return self._initialized and not self._closed and self.terminal.is_alive()
 
     def _is_special_key(self, command: str) -> bool:
         """Check if the command is a special key."""
@@ -370,6 +396,49 @@ class TerminalSession(TerminalSessionBase):
             metadata=metadata,
         )
 
+    def _handle_unresponsive_terminal(
+        self,
+        command: str,
+        terminal_content: str,
+        ps1_matches: list[re.Match],
+    ) -> TerminalObservation:
+        """Handle the shell exiting or wedging while a command was running.
+
+        The session is closed so the executor recreates it on the next call;
+        the observation tells the model what happened and that nothing is
+        still running.
+        """
+        logger.warning(
+            "Terminal backend is no longer alive while running a command; "
+            "closing the session so it is recreated on the next command"
+        )
+        raw_command_output = self._combine_outputs_between_matches(
+            terminal_content, ps1_matches
+        )
+        metadata = CmdOutputMetadata(exit_code=-1, working_dir=self._cwd)
+        metadata.suffix = _UNRESPONSIVE_TERMINAL_SUFFIX
+        command_output = self._get_command_output(
+            command,
+            raw_command_output,
+            metadata,
+            continue_prefix="[Below is the output of the previous command.]\n",
+            is_final=True,
+        )
+        command_output = maybe_truncate(
+            command_output, truncate_after=MAX_CMD_OUTPUT_SIZE
+        )
+        self.prev_status = TerminalCommandStatus.COMPLETED
+        self.prev_output = ""
+        self._query_filter.reset()
+        self.close()
+        return TerminalObservation.from_text(
+            command=command,
+            text=command_output,
+            metadata=metadata,
+            exit_code=metadata.exit_code,
+            is_error=True,
+        )
+
     def _ready_for_next_command(self) -> None:
         """Reset the content buffer for a new command."""
         # Clear the current content
@@ -431,6 +500,15 @@ class TerminalSession(TerminalSessionBase):
                 is_error=True,
             )
 
+        if not self.terminal.is_alive():
+            return self._handle_unresponsive_terminal(
+                command,
+                terminal_content=self.terminal.read_screen(),
+                ps1_matches=CmdOutputMetadata.matches_ps1_metadata(
+                    self.terminal.read_screen()
+                ),
+            )
+
         # If the previous command is not completed,
         # we need to check if the command is empty
         if self.prev_status not in {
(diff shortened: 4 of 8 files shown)


RedCat888:
Quick note for triage: the automated "CI is currently failing" notice above is pinned to the previous commit (`cfb333f0f`) and was last updated a few seconds before I force-pushed the current head `d23e72e39`, so it never re-evaluated the new commit.

On the current head every completed check is green (`Validate PR description`, `check-pr-artifacts`); the combined status is only `pending` because the workflow suites (`Agent Server`, `Run tests`, etc.) sit at `action_required` — the first-time-fork-contributor approval gate, which needs a maintainer to approve the run.

Locally on native Windows (the only platform this change affects) I ran `uv run pytest tests/tools/terminal` (197 passed, 154 skipped) and `uv run pre-commit run` (ruff-format, ruff, pycodestyle, pyright, import-rules, tool-registration — all pass). Happy to approve/re-run CI whenever a maintainer can enable it, or to adjust anything on review. Thanks!

lightbulbdev:
﻿Additional native-Windows reproduction on openhands-tools / SDK / Agent Server 1.49.5, Agent Canvas 1.23.0, Windows PowerShell 5.1.26100.9444.

We observed a related failure with a concrete console-host termination path:

1. An agent ran `dotnet run` successfully. The tool returned after its no-new-output interval.
2. The agent sent `C-c` with `is_input=true`.
3. The observation contained Win32 error 0xE9, "No process is on the other end of the pipe", while setting the console window title (`SetConsoleWindowTitle`). Windows PowerShell event 403 recorded the persistent shell stopping at the same time.
4. Every subsequent tool call failed with `Cannot send keys: PowerShell process is not running`.

In the installed 1.49.5 backend, `_terminate_child_processes()` recursively enumerates descendants of the persistent PowerShell process without excluding `conhost.exe`. In an isolated reproduction, the shell's immediate children included both `conhost.exe` and the native command process. Interrupting a sleeping Python child reproduced shell exit code 4 and the exact next-command failure twice. The original incident's process exit code was not available; code 4 is from the isolated reproduction.

Controlled comparison: filtering `conhost.exe` and `OpenConsole.exe` out of that descendant cleanup preserved the same PowerShell process, and the next command completed successfully. Only `conhost.exe` was observed in our reproduction; excluding `OpenConsole.exe` was precautionary.

Minimal reproduction, no model, project, database, or application server required (run on Windows in the affected environment):

```python
import sys
import tempfile
from openhands.tools.terminal.definition import TerminalAction
from openhands.tools.terminal.impl import TerminalExecutor

with tempfile.TemporaryDirectory() as work_dir:
    ex = TerminalExecutor(
        working_dir=work_dir,
        terminal_type="powershell",
        no_change_timeout_seconds=1,
    )
    try:
        child = sys.executable.replace("'", "''")
        print(ex(TerminalAction(
            command=f"& '{child}' -c 'import time; time.sleep(60)'"
        )).text)
        print(ex(TerminalAction(command="C-c", is_input=True)).text)
        print(ex(TerminalAction(command="Write-Output AFTER_INTERRUPT")).text)
    finally:
        ex.close()
```

We also confirmed why recovery did not occur: the Python session remained initialized and not closed after the underlying process exited. Explicit `reset=true` did create a usable new shell.

A local workaround now excludes console hosts and adds typed process-loss detection during send/read polling plus executor-level session recreation. Recovery returns an explicit error observation and does not replay a possibly partially executed command. Configured environment/starting directory are restored, while transient session state is reported as lost.

Eleven local regression checks passed, including interrupt survival, death before send, death during a command, explicit reset, configured environment restoration, serialized calls, and a write-then-exit case that must not duplicate its side effect. A smoke test launched through Agent Server also verified execution after an intentional shell exit.

These results validate our local workaround, not this PR's branch. Useful regression cases for this PR are preservation of console infrastructure, shell death during execution/polling, and clear reporting of potentially partial side effects without automatic replay. Happy to contribute a focused regression test if it would help this existing work.

lightbulbdev:
Here is the exact local source patch referenced in my earlier reproduction comment. It was applied to the installed `openhands-tools==1.49.5` files and tested on native Windows; this is not a patch rebased onto this PR's branch, and it is not a claim that the full upstream test/lint suite passed.

It has two parts:
- Exclude console-host processes from descendant termination during interruption.
- Detect a lost PowerShell process/input pipe and recreate the session under a single-session execution lock, returning a recovery observation without replaying the action. Preserve the configured terminal type during reset.

The local tests covered interrupt survival, death before send, death during polling, configuration restoration, explicit reset, concurrent calls, and no duplication of a write performed before shell exit. They used the real Windows PowerShell backend. `OpenConsole.exe` is a precautionary exclusion; `conhost.exe` was the observed process in the reproduction.

The patch intentionally restores the configured starting directory/environment and reports transient state loss. It is a narrow workaround for the dead-process case, not a complete solution for all live-but-unresponsive PowerShell sessions or terminating-error completion handling addressed by this PR. The exception import and lifecycle API can be adapted to the upstream design. A broken pipe is treated as an unusable session even when the process has not yet reported an exit.

<details>
<summary>Exact local patch against installed 1.49.5 sources</summary>

```diff
--- a/openhands-tools/openhands/tools/terminal/terminal/windows_terminal.py
+++ b/openhands-tools/openhands/tools/terminal/terminal/windows_terminal.py
@@ -55,6 +55,10 @@
     "C-D": "\x04",
     "C-C": "\x03",
 }
+
+
+class WindowsTerminalDied(RuntimeError):
+    """The persistent PowerShell process or its input pipe was lost."""
 
 
 class WindowsTerminal(TerminalInterface):
@@ -213,7 +217,7 @@
     def send_keys(self, text: str, enter: bool = True) -> None:
         """Send text or supported control sequences to the PowerShell session."""
         if self.process is None or self.process.poll() is not None:
-            raise RuntimeError("Cannot send keys: PowerShell process is not running")
+            raise WindowsTerminalDied("Cannot send keys: PowerShell process is not running")
 
         upper = text.strip().upper()
         ctrl = parse_ctrl_key(text)
@@ -289,7 +293,7 @@
             self.process.stdin.flush()
         except (BrokenPipeError, OSError) as exc:
             logger.error("Failed to write to PowerShell stdin: %s", exc)
-            raise RuntimeError("Failed to write to PowerShell session") from exc
+            raise WindowsTerminalDied("Failed to write to PowerShell session") from exc
 
     def _read_output(self) -> None:
         if self.process is None or self.process.stdout is None:
@@ -329,6 +333,9 @@
 
     def read_screen(self) -> str:
         """Return the accumulated visible PowerShell output."""
+        if self.process is None or self.process.poll() is not None:
+            code = self.process.returncode if self.process is not None else None
+            raise WindowsTerminalDied(f"PowerShell process exited (exit code: {code})")
         return self._get_buffered_output(clear=False)
 
     def clear_screen(self) -> None:
@@ -353,7 +360,10 @@
         script = f"""
 $root = {self.process.pid}
 $childrenByParent = @{{}}
-Get-CimInstance Win32_Process | ForEach-Object {{
+# Console hosts belong to the persistent shell, not the user command.
+Get-CimInstance Win32_Process | Where-Object {{
+    $_.Name -notin @('conhost.exe', 'OpenConsole.exe')
+}} | ForEach-Object {{
     $parentId = [int]$_.ParentProcessId
     if (-not $childrenByParent.ContainsKey($parentId)) {{
         $childrenByParent[$parentId] = New-Object System.Collections.Generic.List[int]
--- a/openhands-tools/openhands/tools/terminal/impl.py
+++ b/openhands-tools/openhands/tools/terminal/impl.py
@@ -36,6 +36,12 @@
     DEFAULT_MAX_PANES,
     PooledTmuxTerminal,
     TmuxPanePool,
+)
+
+
+from openhands.tools.terminal.terminal.windows_terminal import (
+    WindowsTerminal,
+    WindowsTerminalDied,
 )
 
 
@@ -106,6 +112,7 @@
         self._sessions: dict[int, TerminalSession] = {}
         self._sessions_lock = threading.Lock()
         self._pool_recovery_lock = threading.Lock()
+        self._single_session_lock = threading.RLock()
 
         use_pool = terminal_type in (None, "tmux") and _is_tmux_available()
 
@@ -403,7 +410,7 @@
             work_dir=original_work_dir,
             username=original_username,
             no_change_timeout_seconds=original_no_change_timeout,
-            terminal_type=None,
+            terminal_type=self._terminal_type,
             shell_path=self.shell_path,
             env=self._env,
         )
@@ -432,6 +439,45 @@
     # ------------------------------------------------------------------
 
     def _execute_single_session(
+        self,
+        action: TerminalAction,
+        conversation: "LocalConversation | None" = None,
+    ) -> TerminalObservation:
+        # Serialize execution and replacement; interrupt() remains non-blocking.
+        with self._single_session_lock:
+            try:
+                if not action.reset and not self.session._closed:
+                    terminal = self.session.terminal
+                    if isinstance(terminal, WindowsTerminal):
+                        terminal.read_screen()  # Check health before exporting envs.
+                return self._execute_single_session_unchecked(action, conversation)
+            except WindowsTerminalDied as error:
+                terminal = self.session.terminal
+                process = getattr(terminal, "process", None)
+                pid = process.pid if process is not None else None
+                code = process.poll() if process is not None else None
+                logger.warning(
+                    "Recovering dead PowerShell terminal: pid=%s exit_code=%s: %s",
+                    pid, code, error,
+                )
+                self._reset_single_session()
+                observation = TerminalObservation.from_text(
+                    text=(
+                        "PowerShell terminal died and was recreated. The command "
+                        "was NOT automatically retried; it may have partially run. "
+                        "Verify any side effects before retrying. Session variables "
+                        "and working-directory changes were lost; configured environment "
+                        "and starting directory were restored. Re-establish your "
+                        "working directory and restart any required application. "
+                        f"Original process: pid={pid}, exit_code={code}."
+                    ),
+                    is_error=True,
+                    command=action.command or "[RECOVERY]",
+                    exit_code=-1,
+                )
+                return self._mask_observation(observation, conversation)
+
+    def _execute_single_session_unchecked(
         self,
         action: TerminalAction,
         conversation: "LocalConversation | None" = None,
```

</details>

No startup-script or machine-specific cache-path changes are included. Our local startup checksum guard is deployment-specific and is not proposed as part of the upstream fix.

PR #5024, opened 2026-09-13
fix(skills): exclude installed resources from generic context discovery

### LINKED ISSUE #5023: [Bug]: Installed skill resources leak into permanent context even when disabled

Supersedes #5021, which was filed from an automation account by mistake and will be closed.

### Bug Description

Managed installed skills leak their supporting markdown (`commands/`, `references/`) into always-active `REPO_CONTEXT` through generic user-skill discovery. Disabling the installed skill removes its callable entry but does **not** stop the supporting documents from loading.

`load_user_skills()` scans the parent skills directory and then calls the managed installed-skills loader. `find_skill_md_directories()` only recognizes direct-child package roots, while `find_regular_md_files()` discovers loose markdown recursively. The extra `installed/` level means supporting documents escape the package exclusion, and because they carry no `triggers` frontmatter they are classified as permanent repository context.

This defeats the progressive disclosure the docs promise: "The agent reads referenced files from `scripts/`, `references/`, or `assets/` only when needed" (docs.openhands.dev/overview/skills).

### Expected Behavior

```text
enabled: ['example']
disabled: []
```

A packaged skill is advertised once as a callable skill. Its instructions load when invoked; its supporting resources load only when the skill reads them.

### Actual Behavior

Save the snippet under **Minimal Code Sample** as `repro.py` in a checkout of this repository and run it:

```bash
make build
uv run python repro.py
```

Output:

```text
enabled: ['example', 'installed/example/references/notes']
disabled: ['installed/example/references/notes']
```

Every non-`README`/non-`SKILL` markdown file in the package becomes a standalone always-on skill, named by its path relative to the scan root. Disabling the package does not remove them.

The same run against a checkout with the managed installation subtree excluded from generic discovery prints the expected two lines, so the reproduction isolates this behavior rather than a packaging or configuration difference.

Regression coverage lives in `tests/sdk/skills/test_load_user_skills.py`:

```bash
uv run pytest tests/sdk/skills tests/sdk/context
```

### Steps to Reproduce

1. Install a skill package that has a `references/` (or `commands/`) directory into the managed installed-skills directory.
2. Call `load_user_skills()` and observe the extra `installed/<pkg>/references/<file>` entries.
3. Call `disable_skill("<pkg>")` and call `load_user_skills()` again.
4. The callable skill disappears, the supporting documents remain.

### Acceptance Criteria

- [ ] `load_user_skills()` returns only the callable skill for an enabled installed package.
- [ ] `load_user_skills()` returns nothing for that package when it is disabled.
- [ ] Supporting resource files stay on disk and remain readable when the skill is invoked.
- [ ] Recursive discovery of loose legacy markdown outside the managed directory is unchanged.
- [ ] Direct-child AgentSkills packages and user-over-installed precedence are unchanged.

### Installation Method

Ships inside `ghcr.io/openhands/agent-canvas:1.18.0`; also reproduced from a source checkout via `make build` and `uv run`.

_No response_

### SDK Version

1.46.0 (observed in production). Reproduced against `main` at `c37007429be8b4465a83487dc1fd0914df0ea734`; the relevant logic is unchanged in v1.47.0.

### Version Confirmation

### Python Version

3.13.15

N/A — this is a context-assembly bug, model independent.

### Operating System

Linux

### Logs and Error Messages

No error is raised; the failure is silent. Observed on a live deployment, where one installed package (`iterate`) contributed six always-on entries:

```text
installed/iterate/commands/babysit
installed/iterate/commands/iterate
installed/iterate/commands/verify
installed/iterate/references/heuristics
installed/iterate/references/verification
installed/iterate/references/windows
```

The three command wrappers are near-identical one-line shims, and the reference documents are injected in full on every conversation whether or not the skill is invoked.

### Minimal Code Sample

```python
from pathlib import Path
from openhands.sdk.skills import (
    skill, installed, install_skill, disable_skill, load_user_skills,
)

old_dirs = skill.USER_SKILLS_DIRS
old_installed = installed.DEFAULT_INSTALLED_SKILLS_DIR
try:
    with TemporaryDirectory() as tmp:
        root = Path(tmp)
        source = root / "example"
        source.mkdir()
        (source / "SKILL.md").write_text(
            "---\nname: example\ndescription: Example\n---\nInstructions"
        )
        (source / "references").mkdir()
        (source / "references" / "notes.md").write_text("Resource instructions")
        skill.USER_SKILLS_DIRS = [root / "skills"]
        installed.DEFAULT_INSTALLED_SKILLS_DIR = root / "skills" / "installed"
        install_skill(str(source))
        print("enabled:", sorted(s.name for s in load_user_skills()))
        disable_skill("example")
        print("disabled:", sorted(s.name for s in load_user_skills()))
finally:
    skill.USER_SKILLS_DIRS = old_dirs
    installed.DEFAULT_INSTALLED_SKILLS_DIR = old_installed
```

Suggested fix: exclude the managed installation subtree from generic discovery and leave enable/disable decisions to the installed-skills loader, preserving ordinary recursive legacy markdown discovery, direct-child AgentSkills resources, and user-over-installed precedence.

Scope note: the leak grows with package size and affects every installed package that ships supporting files, so the wasted context is paid on every conversation.

Related: #1981 (closed public-skills reference leakage) and #4495 (plugin discovery refactor, which does not cover the managed user-installation overlap).

lkshrk:
HUMAN:

Ran into this on my self-hosted Agent Canvas deployment (image 1.18.0). One installed
skill was quietly adding six entries to every conversation — three command shims plus
three reference documents, injected in full whether or not I invoked the skill.
Disabling the skill dropped the callable entry but left the documents behind, which is
what pushed me to look at discovery instead of my own configuration.

---

AGENT:

## Why

Managed installed packages sit inside the generic user-skills scan root. `find_skill_md_directories()` only recognizes direct-child package roots, while `find_regular_md_files()` recurses, so the extra `installed/` level lets a package's `commands/` and `references/` markdown escape the package exclusion. Carrying no `triggers` frontmatter, those files are classified as permanent repository context and injected into `REPO_CONTEXT` on every conversation. The installed-skills loader owns the callable entry, so disabling the package leaves the leaked documents behind.

This defeats progressive disclosure, and the cost is paid on every conversation. On the deployment where this surfaced, one package contributed six always-on entries — three near-identical command shims plus three reference documents in full.

## Summary

- Filter both discovery passes in `load_skills_from_dir()` against the managed installation directory, leaving enable/disable state to the installed-skills loader.
- Preserve recursive legacy markdown discovery elsewhere, direct-child AgentSkills resources, and user-over-installed precedence. No change to public function signatures, registry format, or the agent loop.
- Add enabled/disabled regression cases driven through the real installation and filesystem APIs; only path configuration is redirected with `monkeypatch`.

## Issue Number

Fixes #5023

## How to Test

```bash
make build
uv run pytest tests/sdk/skills tests/sdk/context
uv run pre-commit run --files openhands-sdk/openhands/sdk/skills/skill.py tests/sdk/skills/test_load_user_skills.py
```

**End-to-end evidence.** The self-contained reproduction from #5023 (real `install_skill` / `disable_skill` / `load_user_skills`, real filesystem, no mocks), run against this checkout with the patch stashed and unstashed:

```text
===== BEFORE (stash patch) =====
enabled: ['example', 'installed/example/references/notes']
disabled: ['installed/example/references/notes']
===== AFTER (patched) =====
enabled: ['example']
disabled: []
```

**Test suites**, on this branch:

```text
uv run pytest tests/sdk/skills   ->  364 passed, 0 failed, 2 skipped
uv run pytest tests/sdk/context  ->  440 passed
```

**Pre-commit**, on both changed files:

```text
Ruff format .......... Passed      Ruff lint ............................ Passed
PEP8 (pycodestyle) ... Passed      Type check with pyright .............. Passed
Forbid dynamic attribute access in SDK .................................. Passed
Check import dependency rules ........................................... Passed
Check Tool subclass registration ........................................ Passed
```

Environment: Python 3.13.15, branched from `c37007429be8b4465a83487dc1fd0914df0ea734`.

## Video/Screenshots

Not applicable — no UI surface. The before/after console output above is the observable behavior change.

## Design Doc

Skipped per CONTRIBUTING: this is a small bug fix with no public API change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Behavior change to initial prompt content: packages that were leaking supporting documents will stop contributing them, so `REPO_CONTEXT` shrinks for affected deployments. That is the intent of the fix, but it does alter prompts, so it deserves a deliberate look.
- The import of `get_installed_skills_dir` is lazy because `installed.py` imports `Skill` from `skill.py`.
- Scope: this filters the managed installation directory. A package nested at depth under some *other* scan root would still be swept; that is a separate, broader change and is deliberately not attempted here.
- Supersedes #5022, opened in error from an automation account.

### DIFF
--- a/openhands-sdk/openhands/sdk/skills/skill.py
+++ b/openhands-sdk/openhands/sdk/skills/skill.py
@@ -877,9 +877,22 @@ def load_skills_from_dir(
     # Note: Third-party files (AGENTS.md, etc.) are loaded separately by
     # load_project_skills() to ensure they're loaded even when this directory
     # doesn't exist.
-    skill_md_files = find_skill_md_directories(skill_dir)
+    # Installed packages are loaded separately with registry enable/disable state.
+    # Import lazily because installed.py imports Skill from this module.
+    from openhands.sdk.skills.installed import get_installed_skills_dir
+
+    installed_dir = get_installed_skills_dir().resolve()
+    skill_md_files = [
+        path
+        for path in find_skill_md_directories(skill_dir)
+        if not path.resolve().is_relative_to(installed_dir)
+    ]
     skill_md_dirs = {skill_md.parent for skill_md in skill_md_files}
-    regular_md_files = find_regular_md_files(skill_dir, skill_md_dirs)
+    regular_md_files = [
+        path
+        for path in find_regular_md_files(skill_dir, skill_md_dirs)
+        if not path.resolve().is_relative_to(installed_dir)
+    ]
 
     # Load SKILL.md files (auto-detected and validated in Skill.load)
     # Wrap each load in try/except to ensure one bad skill doesn't break all loading
--- a/tests/sdk/skills/test_load_user_skills.py
+++ b/tests/sdk/skills/test_load_user_skills.py
@@ -395,3 +395,41 @@ def test_load_user_skills_disabled_installed_skill_excluded(tmp_path, monkeypatc
         assert "disabled-skill" not in skill_names
     finally:
         skill.USER_SKILLS_DIRS = original_dirs
+
+
+@pytest.mark.parametrize("enabled", [True, False])
+def test_installed_skill_resources_do_not_leak_into_user_skills(
+    tmp_path, monkeypatch, enabled
+):
+    skills_dir = tmp_path / "skills"
+    installed_dir = skills_dir / "installed"
+    source = tmp_path / "resource-skill"
+    source.mkdir()
+    (source / "SKILL.md").write_text(
+        "---\nname: resource-skill\ndescription: Resource skill\n---\nInstructions."
+    )
+    for directory in ("commands", "references"):
+        (source / directory).mkdir()
+        (source / directory / "notes.md").write_text("Private resource instructions.")
+    install_skill(str(source), installed_dir=installed_dir)
+    if not enabled:
+        disable_skill("resource-skill", installed_dir=installed_dir)
+
+    (skills_dir / "legacy").mkdir()
+    (skills_dir / "legacy" / "notes.md").write_text("Legacy instructions.")
+    direct = skills_dir / "direct-skill"
+    direct.mkdir()
+    (direct / "SKILL.md").write_text(
+        "---\nname: direct-skill\ndescription: Direct skill\n---\nInstructions."
+    )
+    (direct / "reference.md").write_text("Direct resource instructions.")
+    monkeypatch.setattr(skill, "USER_SKILLS_DIRS", [skills_dir])
+    monkeypatch.setattr(installed, "DEFAULT_INSTALLED_SKILLS_DIR", installed_dir)
+
+    loaded = load_user_skills()
+    expected = {"legacy/notes", "direct-skill"}
+    if enabled:
+        expected.add("resource-skill")
+    assert {entry.name for entry in loaded} == expected
+    assert len(loaded) == len(expected)
+    assert (installed_dir / "resource-skill" / "references" / "notes.md").is_file()


VascoSch92:
Thanks. A few changes before merging:

**1. Python 3.12 regression.** `path.resolve()` runs on every discovered file, outside the per-file `try/except`. On Python 3.12 (still in `requires-python`), a symlink loop makes it raise `RuntimeError`, and `load_user_skills()` drops the whole folder. Repro: `good.md` + `a.md -> b.md -> a.md` loads `['good']` on base and `[]` here. 3.13 is fine, so CI won't catch it.

**2. Reuse the existing exclusion.** `find_regular_md_files()` already takes `exclude_dirs`, so passing the installed dir there is enough. The `find_skill_md_directories()` filter never matches the default layout (installed skills are one level too deep for it), so it can go.

**3. Scope it to the user scan.** The exclusion currently applies to every caller of the exported `load_skills_from_dir()`. Plugins handle the same problem in `load_user_plugins()` via `skip_dirs` (`plugin/discovery.py:105-110`). Suggest the same for skills:

```python
# load_skills_from_dir(skill_dir, exclude_dirs: set[Path] | None = None)
regular_md_files = find_regular_md_files(
    skill_dir, skill_md_dirs | (exclude_dirs or set())
)

# load_user_skills(), passed through _load_and_merge_from_dirs
exclude_dirs={get_installed_skills_dir()}
```

With your test, this passes `tests/sdk/skills` on 3.12 (364 passed), fixes the symlink case, and leaves other callers unchanged. If you'd rather not touch the public signature, just passing `skill_md_dirs | {get_installed_skills_dir()}` also works (+6/−1), but keeps the exclusion global.

PR #5026, opened 2026-09-13
fix(secret): resolve this server's own LookupSecret URLs in-process

### LINKED ISSUE #5025: [Bug]: LookupSecret self-deadlocks the agent-server event loop on loopback resolution

### Bug Description

`LookupSecret.get_value()` performs a **synchronous, blocking** `httpx.get()` against the URL the secret was registered with. When that URL points back at the process doing the lookup — an agent-server resolving a secret for an in-process conversation — the call blocks the event loop that would have to serve the request. Nothing can answer it, so it stalls for the full 30s timeout and raises `ReadTimeout`.

While that call is in flight the server answers nothing at all, `GET /alive` included, because `alive()` is `async def` and runs on the blocked loop. With several secrets registered they resolve one after another, so a conversation with four secrets freezes the server for roughly two minutes.

The SDK already knows about this failure mode. `openhands-sdk/openhands/sdk/conversation/impl/local_conversation.py` carries this comment, and two call sites are wrapped in `asyncio.to_thread` because of it:

> Off-load lazy init to a worker thread: init_state may block the loop (an ACP agent resolves credentials via a synchronous LookupSecret httpx.get). When the agent-server runs arun() on its event loop and that lookup points back at the same single-process server, doing it inline freezes the loop so the lookup can never be served — **a self-deadlock that ReadTimeouts after 30s** (agent-canvas#1072).

Offloading individual call sites treats each caller rather than the cause: any synchronous `get_value()` reached from the loop re-creates the deadlock, and `get_value()` is a public `SecretSource` method callers are free to use.

### Expected Behavior

Resolving a `LookupSecret` whose URL is served by the current process returns the value without a network round trip, and never blocks the event loop.

### Actual Behavior

Save the script under **Minimal Code Sample** as `deadlock_repro.py` in a checkout of this repository and run it:

```bash
make build
uv run python deadlock_repro.py
```

Output:

```text
off-loop  : 's3cr3t' in 0.00s
on-loop   : ReadTimeout after 30.00s
/alive served during the on-loop call: 0
```

The same resolution takes 0.00s from a worker thread and 30.00s on the loop. The last line is the core of it: the server served zero `/alive` requests while blocked on a request only it could answer.

### Steps to Reproduce

1. Start an agent-server (or any asyncio server) that serves `GET /api/settings/secrets/{name}`.
2. Build a `LookupSecret` whose URL points at that same process.
3. Call `get_value()` from a coroutine running on that server's event loop.
4. The call hangs for 30s and raises `ReadTimeout`; the server serves nothing meanwhile.

### Acceptance Criteria

- [ ] Resolving a `LookupSecret` served by the current process returns without an HTTP round trip.
- [ ] `get_value()` called on the event loop for such a URL completes promptly instead of timing out.
- [ ] The server keeps serving requests (`/alive`) throughout resolution.
- [ ] URLs the process does not serve still resolve over HTTP unchanged.
- [ ] Remote workspaces and non-loopback `LookupSecret` URLs are unaffected.

### Installation Method

Ships inside `ghcr.io/openhands/agent-canvas:1.18.0`; reproduced from a source checkout via `make build` and `uv run`.

_No response_

### SDK Version

1.46.0 (observed in production). Reproduced against `main` at `c37007429be8b4465a83487dc1fd0914df0ea734`.

### Version Confirmation

### Python Version

3.13.15

N/A — model independent.

### Operating System

Linux

### Logs and Error Messages

From a self-hosted deployment. Four registered secrets, resolved one after another, each timing out at exactly 30s:

```text
04:34:06 Unexpected error retrieving secret 'GITHUB_PERSONAL_ACCESS_TOKEN': ReadTimeout: timed out
04:34:36 Unexpected error retrieving secret 'OPENHANDS_URL': ReadTimeout: timed out
04:35:06 Unexpected error retrieving secret 'CODER_SESSION_TOKEN': ReadTimeout: timed out
04:35:36 Unexpected error retrieving secret 'LITELLM_API': ReadTimeout: timed out
```

`GET /alive` responses per minute across that window — none at all during the freeze, then a backlog flush:

```text
04:30: 9   04:31: 9   04:32: 9   04:33: 5   04:34: 0   04:35: 21
```

The liveness probe failed for long enough that the container was killed and restarted, which reloaded the same conversations and repeated the cycle. Stored `LookupSecret` URLs pointed at `http://127.0.0.1:18000/api/settings/secrets/<NAME>` — the server's own port.

Secondary observation: `ReadTimeout` is logged through the generic `Unexpected error` branch in `secret_registry.py`, not the `Transient error` branch, because `httpx.ReadTimeout` derives from `httpx.TransportError → HTTPError → Exception` rather than `OSError`/`TimeoutError`.

### Minimal Code Sample

```python
"""Self-deadlock: LookupSecret.get_value() blocks the loop that must serve it."""
import asyncio, time
import uvicorn
from fastapi import FastAPI
from fastapi.responses import PlainTextResponse
from openhands.sdk.secret import LookupSecret

app = FastAPI()
served = []

@app.get("/alive")
async def alive():
    served.append(time.monotonic())
    return {"status": "ok"}

@app.get("/api/settings/secrets/{name}")
async def secret(name: str):
    return PlainTextResponse("s3cr3t")

async def main():
    cfg = uvicorn.Config(app, host="127.0.0.1", port=18777, log_level="critical")
    server = uvicorn.Server(cfg)
    task = asyncio.create_task(server.serve())
    while not server.started:
        await asyncio.sleep(0.05)

    src = LookupSecret(url="http://127.0.0.1:18777/api/settings/secrets/DEMO")

    # Baseline: resolve from a worker thread — loop stays free.
    t0 = time.monotonic()
    val = await asyncio.to_thread(src.get_value)
    print(f"off-loop  : {val!r} in {time.monotonic()-t0:.2f}s")

    # The bug: resolve inline on the event loop.
    served.clear()
    t0 = time.monotonic()
    try:
        val = src.get_value()
        print(f"on-loop   : {val!r} in {time.monotonic()-t0:.2f}s")
    except Exception as e:
        print(f"on-loop   : {type(e).__name__} after {time.monotonic()-t0:.2f}s")
    print(f"/alive served during the on-loop call: {len(served)}")

    server.should_exit = True
    await task

asyncio.run(main())
```

Suggested fix: let a process that serves `LookupSecret` URLs answer its own URLs in-process, so the round trip disappears rather than being dodged per caller. `openhands.sdk` cannot import `openhands.agent_server` (enforced by `check-import-rules`), so this wants a small registration interface in the SDK that the agent-server opts into at startup — an extension point rather than a special case.

Worth noting the guarantee `LookupSecret` exists to provide — "raw values **never** transit through the SDK client", per `RemoteWorkspace.get_secrets()` — is vacuous when the client and the server are the same process, since the value is already in that process's memory. Lazy resolution is still worth keeping; only the loopback HTTP hop needs to go.

A PR implementing this follows.

lkshrk:
HUMAN:

This one took down my self-hosted agent-server three times before I understood it. Four
secrets on a conversation, each resolution blocking the loop for 30s, so the liveness
probe killed the pod, the restart reloaded the same conversations, and it went round
again. Happy to reshape the interface if you'd rather it looked different — the part I
care about is that the process stops making HTTP calls to itself.

---

AGENT:

## Why

`LookupSecret.get_value()` performs a synchronous, blocking `httpx.get()`. When the URL points back at the process making the call, the event loop that would serve the request is the one blocked waiting for it. Nothing answers, and it stalls for the full 30s timeout.

While that is in flight the server answers nothing at all — `alive()` is `async def` and runs on the blocked loop. Registered secrets resolve serially, so four of them freeze the server for about two minutes. On the deployment where this surfaced that was long enough for the liveness probe to kill the container; the restart reloaded the same conversations and the cycle repeated.

`local_conversation.py` already documents this failure mode and wraps two call sites in `asyncio.to_thread` because of it. That treats callers rather than the cause: `get_value()` is a public `SecretSource` method, and any synchronous call reached from the loop reproduces the deadlock.

## Summary

- Add `register_local_secret_resolver()` / `unregister_local_secret_resolver()` to `openhands.sdk.secret`. `LookupSecret.get_value()` consults registered resolvers before falling back to HTTP; a resolver returning `None` declines and the request proceeds unchanged. `openhands.sdk` cannot import `openhands.agent_server` (`check-import-rules`), so this is an extension point the server opts into, not a special case in the SDK.
- Register a resolver for the agent-server's own loopback secret route in the API lifespan. It matches only loopback hosts and the exact `/api/settings/secrets/<name>` shape, and falls through to HTTP when its store does not hold the name so the caller still sees the server's 404.
- Additive only: no signature changes, no behavior change for remote or non-loopback URLs, no change to the agent loop or the registry format.

## Issue Number

Fixes #5025

## How to Test

```bash
make build
uv run pytest tests/sdk/conversation tests/sdk/mcp tests/agent_server
  openhands-sdk/openhands/sdk/secret/secrets.py \
  openhands-sdk/openhands/sdk/secret/__init__.py \
  openhands-agent-server/openhands/agent_server/api.py \
  openhands-agent-server/openhands/agent_server/local_secret_resolver.py \
  tests/sdk/conversation/test_secret_source.py \
  tests/agent_server/test_local_secret_resolver.py
```

**End-to-end evidence.** The reproduction from #5025 runs a real `uvicorn` server and resolves a `LookupSecret` pointing at that same process, once from a worker thread and once inline on the loop. On `main`:

```text
off-loop  : 's3cr3t' in 0.00s
on-loop   : ReadTimeout after 30.00s
/alive served during the on-loop call: 0
```

With this branch and a resolver registered, plus a second `LookupSecret` whose URL the resolver declines so it still goes over HTTP:

```text
on-loop, resolver registered : 's3cr3t' in 0.00s
unmatched URL -> HTTP        : 's3cr3t' in 0.01s
```

30.00s and a wedged server become 0.00s, and the HTTP path is untouched for URLs the process does not serve.

**Test suites**, on this branch:

```text
uv run pytest tests/sdk/conversation  ->   814 passed
uv run pytest tests/sdk/mcp           ->   134 passed
uv run pytest tests/agent_server      ->  2109 passed
```

**Pre-commit**, on all six changed files:

```text
Ruff format .......... Passed      Ruff lint ............................ Passed
PEP8 (pycodestyle) ... Passed      Type check with pyright .............. Passed
Forbid dynamic attribute access in SDK .................................. Passed
Check import dependency rules ........................................... Passed
Check Tool subclass registration ........................................ Passed
```

`Check import dependency rules` passing is the one worth calling out: the SDK gains the extension point without importing the agent-server.

Environment: Python 3.13.15, branched from `c37007429be8b4465a83487dc1fd0914df0ea734`.

## Video/Screenshots

Not applicable — no UI surface. The before/after console output above is the observable behavior change.

## Design Doc

Not included. The change is additive and small, but it does introduce a public extension point, so if you would rather see a design page before reviewing the code, say so and I will add one under `.pr/`.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The guarantee `LookupSecret` exists for — "raw values **never** transit through the SDK client", per `RemoteWorkspace.get_secrets()` — is unaffected. When client and server are the same process the value is already in that process's memory; only the loopback hop is removed. Lazy resolution is preserved.
- The resolver registry is process-global, matching how these URLs are resolved. Registration is scoped to the API lifespan and unregistered in the outer `finally`.
- A resolver that raises is logged and skipped so one bad resolver cannot take secret resolution down; covered by a test.
- Loopback matching is host-based (`127.0.0.1`, `localhost`, `::1`). Two agent-servers on one host with different stores would both match a loopback URL; if you would prefer matching on the server's own advertised base URL instead, that is a small change and I am happy to make it.
- Separate, not addressed here: `httpx.ReadTimeout` is logged through the generic `Unexpected error` branch in `secret_registry.py` rather than `Transient error`, because it derives from `httpx.TransportError → HTTPError → Exception` rather than `OSError`/`TimeoutError`. Happy to fold in a one-line fix if you want it here rather than separately.

### DIFF
--- a/openhands-agent-server/openhands/agent_server/api.py
+++ b/openhands-agent-server/openhands/agent_server/api.py
@@ -49,6 +49,7 @@ from openhands.agent_server.init_router import (
     require_initialized,
 )
 from openhands.agent_server.llm_router import llm_router
+from openhands.agent_server.local_secret_resolver import local_secret_resolution
 from openhands.agent_server.mcp_router import mcp_router
 from openhands.agent_server.middleware import CORSDispatcher
 from openhands.agent_server.openai.router import (
@@ -153,12 +154,18 @@ def _cleanup_stale_tmux_sessions() -> None:
 @asynccontextmanager
 async def api_lifespan(api: FastAPI) -> AsyncIterator[None]:
     tmux_tmpdir, tmux_tmpdir_was_defaulted = _ensure_server_tmux_tmpdir()
+    secret_resolution: local_secret_resolution | None = None
     try:
         # Clean up stale tmux sessions from previous server runs
         _cleanup_stale_tmux_sessions()
 
         config: Config = api.state.config
         deferred = config.deferred_init
+
+        # Answer our own LookupSecret URLs in-process; a loopback fetch made
+        # from the event loop cannot be served by the loop blocked on it.
+        secret_resolution = local_secret_resolution(config)
+        secret_resolution.__enter__()
 
         # Deferred pods boot with telemetry disabled and are rebuilt by
         # InitService, so they emit `server_started` there instead.
@@ -281,6 +288,8 @@ async def api_lifespan(api: FastAPI) -> AsyncIterator[None]:
     finally:
         # Outer finally so a startup failure cannot leak the drain task, and
         # after `async with service` so terminal events are still accepted.
+        if secret_resolution is not None:
+            secret_resolution.__exit__(None, None, None)
         emit_server_stopped()
         await shutdown_telemetry_sink()
 
--- a/openhands-agent-server/openhands/agent_server/local_secret_resolver.py
+++ b/openhands-agent-server/openhands/agent_server/local_secret_resolver.py
@@ -1,0 +1,69 @@
+"""Serve this server's own ``LookupSecret`` URLs without a loopback request.
+
+``LookupSecret`` resolves lazily over HTTP so raw values never transit an SDK
+client. When the client and the server are the same process, that round trip
+buys nothing and costs correctness: a resolution on the event loop blocks the
+very loop that would answer it, so it stalls until the client times out.
+Resolving those URLs against the local store keeps the laziness and drops the
+round trip.
+"""
+
+from urllib.parse import urlsplit
+
+from openhands.agent_server.config import Config
+from openhands.agent_server.persistence import get_secrets_store
+from openhands.sdk.logger import get_logger
+from openhands.sdk.secret import (
+    register_local_secret_resolver,
+    unregister_local_secret_resolver,
+)
+
+
+logger = get_logger(__name__)
+
+_SECRET_PATH_PREFIX = "/api/settings/secrets/"
+_LOOPBACK_HOSTS = frozenset({"127.0.0.1", "localhost", "::1", "[::1]"})
+
+
+def _secret_name_if_local(url: str) -> str | None:
+    """Return the secret name when ``url`` is one this process serves itself."""
+    parsed = urlsplit(url)
+    if parsed.hostname not in _LOOPBACK_HOSTS:
+        return None
+    if not parsed.path.startswith(_SECRET_PATH_PREFIX):
+        return None
+    name = parsed.path[len(_SECRET_PATH_PREFIX) :]
+    # Only a bare name; anything deeper is a different route.
+    if not name or "/" in name:
+        return None
+    return name
+
+
+def build_local_secret_resolver(config: Config):
+    """Build a resolver that answers this server's own secret URLs."""
+
+    def resolve(url: str) -> str | None:
+        name = _secret_name_if_local(url)
+        if name is None:
+            return None
+        value = get_secrets_store(config).get_secret(name)
+        if value is None:
+            # Fall through to HTTP so the caller still sees the server's 404.
+            return None
+        logger.debug("Resolved secret '%s' in-process", name)
+        return value
+
+    return resolve
+
+
+class local_secret_resolution:
+    """Context manager registering the in-process resolver for a server."""
+
+    def __init__(self, config: Config) -> None:
+        self._resolver = build_local_secret_resolver(config)
+
+    def __enter__(self) -> None:
+        register_local_secret_resolver(self._resolver)
+
+    def __exit__(self, *exc_info) -> None:
+        unregister_local_secret_resolver(self._resolver)
--- a/openhands-sdk/openhands/sdk/secret/__init__.py
+++ b/openhands-sdk/openhands/sdk/secret/__init__.py
@@ -4,16 +4,22 @@ This module provides classes and types for managing secrets in OpenHands.
 """
 
 from openhands.sdk.secret.secrets import (
+    LocalSecretResolver,
     LookupSecret,
     SecretSource,
     SecretValue,
     StaticSecret,
+    register_local_secret_resolver,
+    unregister_local_secret_resolver,
 )
 
 
 __all__ = [
+    "LocalSecretResolver",
     "SecretSource",
     "StaticSecret",
     "LookupSecret",
     "SecretValue",
+    "register_local_secret_resolver",
+    "unregister_local_secret_resolver",
 ]
--- a/openhands-sdk/openhands/sdk/secret/secrets.py
+++ b/openhands-sdk/openhands/sdk/secret/secrets.py
@@ -2,6 +2,7 @@
 
 import os
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from urllib.parse import urljoin, urlsplit
 
 import httpx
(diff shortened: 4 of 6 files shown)


neubig (commented):
🟡 **Acceptable direction, but this needs two correctness fixes before merge.**

**[CRITICAL ISSUES]**

1. **`openhands-agent-server/openhands/agent_server/local_secret_resolver.py:28` — Cross-server secret misrouting.** `_secret_name_if_local()` accepts every loopback URL regardless of port, so this process intercepts `LookupSecret` URLs that belong to a different agent-server. I reproduced this directly: a resolver built for the PR backend returned its own `TOKEN` for `http://127.0.0.1:65534/api/settings/secrets/TOKEN`, where nothing was listening. With two local backends, a lookup intended for backend B can silently receive backend A's same-named secret instead of using HTTP. The resolver must match the current server's advertised scheme/host/port (for example the normalized `OH_INTERNAL_SERVER_URL`), not merely a loopback hostname and route shape. Please add a regression test proving another loopback port falls through.

2. **`openhands-agent-server/openhands/agent_server/api.py:167` — Deferred-init captures the dormant config.** The resolver closes over `api.state.config` during lifespan startup, before `POST /api/init` replaces it with `new_config`. On the first post-init lookup, `get_secrets_store(config)` can therefore initialize the singleton with the dormant secret key rather than the per-user key delivered by `/api/init`; the store explicitly documents that its first config/cipher is permanent for the process. Resolve against the current `api.state.config`, or register/rebuild the resolver after deferred initialization. Add a deferred-init regression test that changes `secret_key` and verifies local resolution uses the initialized config.

**[TESTING GAPS]**

The unit tests pass, but they only test URL shape and mocked resolver dispatch. Repository guidance asks localized server behavior changes to receive a live-server test. The two regressions above are exactly the cases that the current tests miss.

**Validation performed**

- `20 passed` across `test_secret_source.py` and `test_local_secret_resolver.py`
- All targeted pre-commit checks passed, including pyright and import-boundary checks
- Launched this branch as an isolated Agent Canvas backend on port `18202`
- Started a real profile-backed conversation with 11 saved secrets; it returned `PR5026_OK`, and `/alive` responded in 2.9 ms rather than stalling

So the happy path is real and the approach fixes the reported outage, but the host/port ownership check and deferred-init config lifecycle must be corrected before merge.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🔴 HIGH — the current implementation can return a secret from the wrong local backend and can bind deferred-init secret storage to the wrong cipher configuration.

**VERDICT:** ❌ **Needs rework**

**KEY INSIGHT:** An in-process shortcut is sound only when it proves both URL ownership and current tenant/config ownership before returning a secret.

_This review was generated by an AI agent (OpenHands) on behalf of Graham Neubig._

---

>
> 1. Add or edit `.agents/skills/custom-codereview-guide.md` with the `/codereview` trigger and the missing repository context. See the [customization docs](https://docs.openhands.dev/openhands/usage/use-cases/code-review#customization).
> 3. The guideline change will go through normal review when merged.
>
> **Resolve with AI?** Install the [iterate skill](https://github.com/OpenHands/extensions/tree/main/skills/iterate) and run `/iterate` to drive the PR through CI, review, and QA.

neubig (approved):
Follow-up after implementing and validating the requested regressions:

- Added a **no-mock regression** using a real `ThreadingHTTPServer` in commit `bbaeed91c`. On the pre-fix implementation it fails because a lookup addressed to the foreign loopback port returns `local-value` rather than the foreign server's `foreign-value`.
- Added the production fix separately in commit `87ca9ec11`. Local resolution now requires the URL to match `OH_INTERNAL_SERVER_URL` by scheme and effective port, while still treating loopback aliases (`localhost`, `127.0.0.1`, and `::1`) as equivalent on that same port.
- The focused suite now reports **21 passed**, and all targeted pre-commit hooks pass, including Ruff, pycodestyle, pyright, import boundaries, and tool registration.
- Restarted the isolated Agent Canvas backend from the fixed commit. A real profile-backed conversation using stored secrets returned `PR5026_FIXED`; `/alive` responded in 2.5 ms, and no self-HTTP secret request appeared in the backend access log.

Correction to my earlier review: I also tried to turn the deferred-init concern into a no-mock test using the real FastAPI lifespan and `POST /api/init`. That lifecycle test **passed on the existing implementation**, because initialization creates the secrets-store singleton with the initialized config before the resolver first reads it. I therefore removed that speculative test in the follow-up fix commit and no longer consider deferred init a blocker.

With the reproduced cross-port issue fixed and validated, I consider the code **merge-ready pending required CI/review policy**.

_This follow-up was created by an AI agent (OpenHands) on behalf of Graham Neubig._

PR #5046, opened 2026-09-14
feat(agent-server): scope script commands with agent profiles

neubig:
HUMAN:

---

AGENT:

## Why

A deterministic automation script may need a narrowly scoped saved credential without needing an agent or conversation. Agent profiles currently apply their `secret_refs` only while launching conversations, so a scanner would otherwise create an empty conversation and potentially a Docker runtime merely to read its GitHub token.

This change keeps credential policy independent from execution: the script sends a profile ID with its Bash request, while a separate Automation operation decides when actual agent work needs a conversation.

## Summary

- Add optional `agent_profile_id` to Bash command requests and the Python/TypeScript workspace clients.
- Resolve only the profile's selected saved secrets and reuse the existing Bash environment injection and streaming output masking.
- Reject unknown profiles, missing selected secrets, and attempts to replace an existing conversation-scoped registry.

## Issue Number

Closes #5044.

## How to Test

`uv run pytest -q tests/agent_server/test_profile_scoped_bash.py tests/agent_server/test_bash_service.py tests/sdk/workspace/remote/test_remote_workspace_mixin.py tests/sdk/workspace/remote/test_remote_workspace.py tests/sdk/workspace/remote/test_async_remote_workspace.py`

Result: 135 passed. `uv run pre-commit run --files ...` passed for every changed source and test file. `make test-server-schema` passed with the existing 99 weak-schema allowlist entries; the allowlist was not changed. TypeScript `npm run lint` completed with the repository's existing warnings and `npm run build` passed.

The tests execute a real subprocess with selected and unselected synthetic secrets, assert that an empty command scope cannot inherit service secrets, and verify masked stdout/stderr persistence. Live Agent Canvas evidence will be added after the Automation caller is integrated; this PR does not yet claim an end-to-end Canvas demonstration.

## Video/Screenshots

Pending the integrated Automation demonstration.

## Design Doc

The issue and this description contain the complete narrow design; no separate design artifact is needed.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Native stack #5087 is #5017 → this PR. #5017 establishes profile secret selection; this PR makes that selection usable by a non-conversation command. #3403 is a parallel child of #5017, and #5081 independently targets `main`. The checked-in generated TypeScript transport schema remains pinned to the released Agent Server by repository policy; this PR updates the handwritten browser request type, while release automation regenerates the transport schema from canonical Python OpenAPI.

Issue #72, opened 2025-09-03
Rename `openhands.core` to `openhands.sdk`


xingyaoww:
We should rename `openhands.core` to `openhands.sdk` and replace all occurrences

PR #5081, opened 2026-09-15
feat(sdk): add remote conversation control client

### LINKED ISSUE #5080: [Feature]: Add a lightweight remote conversation control client
## Is there an existing feature request for this?

## Problem or Use Case

Services that create and schedule remote agent work currently have to instantiate `RemoteConversation`. That client synchronizes conversation state and events, starts a WebSocket, and imports every agent tool package needed to deserialize the history. A scheduler only needs to create the conversation and submit work; requiring the full interactive client couples the scheduler process to the agent runtime's tools.

## Desired Behavior

Add a small SDK control client for remote conversations. It should use the existing Agent Server API to create a conversation and submit a user turn while leaving state synchronization, event materialization, tool registration, and interactive callbacks to `RemoteConversation`.

The existing `RemoteConversation` methods should reuse the same request primitives where practical so the two clients do not duplicate transport behavior.

## Acceptance Criteria

- [ ] A public Python SDK client can create an idempotently named remote conversation without constructing `RemoteConversation` or loading agent/tool state.
- [ ] The client can submit a user message and trigger the conversation run without opening a WebSocket.
- [ ] The client can read execution status so an existing scheduler or watchdog can observe completion.
- [ ] The client returns the conversation ID from creation and reports request failures with the existing SDK error behavior.
- [ ] `RemoteConversation.create()`, message submission, and status reads reuse the same transport primitives where practical.
- [ ] Focused tests prove the control path does not construct remote state or import tool modules.
- [ ] Public SDK documentation explains when to use the control client and when to use `RemoteConversation`.
- [ ] Existing Agent Server endpoints and the generated TypeScript client remain the source of truth; no duplicate endpoint or Automation-specific API is introduced.

## Alternatives Considered

Adding a `connect=False` option to `RemoteConversation.create()` would make one method return two different abstractions. Installing every runtime tool package in scheduler processes would add unnecessary dependencies and erase the process boundary.

## Priority / Severity

High - Significant impact on productivity

## Estimated Scope

Small - Focused change with limited surface area

## Feature Area

- [x] SDK

## Technical Implementation Ideas (Optional)

Model the control plane as a separate class bound to `RemoteWorkspace`. Keep the low-level HTTP request helpers shared with `RemoteConversation` and rely on the existing generated TypeScript endpoint methods for browser clients.

## Additional Context

This enables schedulers and webhook services to launch agent work while keeping agent tools inside the provisioned runtime.

_This issue was created by an AI agent (OpenHands) on behalf of the user._

neubig:
HUMAN:

---

AGENT:

## Why

A scheduler or webhook service may need to create an agent conversation and submit a turn without reading the agent's state or event history. Today it must construct `RemoteConversation`, which starts state synchronization and a WebSocket and requires the service process to install every tool package needed to deserialize the agent's events.

This PR separates the control plane from the interactive client. `RemoteConversationControl` retains only a workspace and conversation ID. Agent and tool state remain inside the provisioned runtime, while `RemoteConversation` remains the interface for interactive consumers that need events and callbacks.

## Summary

- Add `RemoteConversationControl.create()` for idempotent typed conversation creation without interactive attachment.
- Add lightweight message submission, optional run triggering, and a single execution-status read.
- Share conversation creation, message transport, and status reads with `RemoteConversation` so the two interfaces cannot drift.
- Export the control client from the public SDK surface.

## Issue Number

Closes #5080.

## How to Test

`uv run pytest tests/sdk/conversation/remote/test_remote_conversation.py -q`

Result: 58 passed. `uv run pre-commit run --files ...` passed for every changed source and test file. The focused test makes interactive state construction and WebSocket creation raise, then proves control-only creation, turn submission, and status reading still succeed.

Live Agent Canvas evidence: an isolated Canvas used this client from the Automation service to create Docker-backed conversations and submit turns while the Automation environment had no `openhands.tools` package installed (`find_spec('openhands.tools')` returned `None`). The same path posted [triage for airbnb-clone #63](https://github.com/neubig/airbnb-clone/issues/63#issuecomment-5683019847), then the developer implemented the issue and opened [PR #72](https://github.com/neubig/airbnb-clone/pull/72). The host used only conversation creation, event submission, and status reads; agent tools and state stayed in the Docker runtime. The matching guide is [OpenHands/docs #793](https://github.com/OpenHands/docs/pull/793).

## Video/Screenshots

The [recorded Canvas setup](https://github.com/OpenHands/docs/blob/docs/software-factory-ui/openhands/static/img/software-factory/setup.gif) shows the isolated environment used for the integrated run.

## Design Doc

The issue and this description contain the complete focused design; no separate design artifact is needed.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This PR targets `main` independently. Automation consumes it alongside #5017 and #5046, but that is a cross-feature release dependency rather than a Git dependency. The Agent Server REST contract is unchanged, so the checked-in generated TypeScript client remains in parity.


neubig:
The reduced design can use the existing RemoteConversation.create()/attach(), send_message(), and run(blocking=False) APIs together with Automation KV. A separate state-free control client is therefore not required for the minimum factory path. If interactive-state loading proves to be a concrete deployment problem in live validation, it should be addressed separately with that evidence.

PR #5093, opened 2026-09-16
feat(plugin): add the Agent Plugins mcp.json loader

### LINKED ISSUE #4451: Agent Plugins: mcp.json loader + PLUGIN_ROOT/PLUGIN_DATA expansion

VascoSch92:
HUMAN:

This PR loads MCP servers from an Agent Plugins package's mcp.json and registers the Agent Plugins format, keeping plugin servers away from the user's secrets.

---

AGENT:

## Why

An Agent Plugins package declares its MCP servers in `mcp.json` at the plugin root. We only read `.mcp.json`, so those servers were invisible — and `AgentPluginsFormat` was deliberately held out of the format registry until this landed (#4474), which meant a conformant package still loaded as a Claude Code plugin under a name inferred from its directory.

The portable format is not a rename of ours. It defines exactly two placeholders and forbids every other kind of expansion, a containment boundary for package paths, remote URL and header rules, and a failure boundary per entry. Mapping it onto `MCPServer` is where that work lives.

## Summary

- Adds `agent_plugins_mcp`, the `mcp.json` loader: vendored `mcp-1.0.0.schema.json`, top-level document validated closed, each entry validated independently against the `#/$defs/server` the spec exposes for exactly that purpose. An invalid document disables MCP for the plugin; an invalid or unsupported entry skips only itself.
- Implements `${PLUGIN_ROOT}` / `${PLUGIN_DATA}` in `args`, `env` values and `cwd` — single-pass, non-recursive, nothing else expanded — plus command resolution, `cwd` containment, remote URL and header validation. Headers the client generates (`Host`, `Content-Length`, `Mcp-Session-Id`, …) win over configured ones, and configured headers never follow a redirect to another origin.
- Defines `PLUGIN_DATA` as `get_cache_path(<resolved plugin root>)` under `~/.openhands/plugin-data/` — `<dir name>-<sha256[:16]>`, the fetch cache's naming: outside `plugins/`, which is scanned for plugins; keyed by plugin root, which is already stable across updates, so each plugin instance keeps its own data and a same-named plugin elsewhere cannot share it.
- A loader-only `MCPServer.as_literal()` exempts package servers from every `${VAR}` expansion site — now one helper, `expand_mcp_servers` — so a package cannot pull a user's secret or environment into its own subprocess. The flag is a private attribute: input cannot set it and dumps never write it. Plugin `env` and `headers` stay redacted like any other secret.
- Registers `AgentPluginsFormat` ahead of the Claude Code fallback, the flip `test_not_registered_yet` was pinning.
## Issue Number

Closes #4451. Part of #4405.

## How to Test

End-to-end against the official example package plus the specification's own `mcp.json` example, both verbatim:

```bash
git clone --depth 1 https://github.com/agentplugins/agent-plugins-example /tmp/ap-example
mkdir -p /tmp/ap-example/bin && touch /tmp/ap-example/bin/validator
cat > /tmp/ap-example/mcp.json <<'JSON'
{
  "$schema": "https://agent-plugins.org/schemas/1.0.0/mcp.schema.json",
  "mcpServers": {
    "local-validator": {
      "type": "stdio",
      "command": "./bin/validator",
      "args": ["--data", "${PLUGIN_DATA}/validator"],
      "env": { "CONFIG": "${PLUGIN_ROOT}/config.json" },
      "cwd": "${PLUGIN_ROOT}"
    },
    "deployment-api": {
      "type": "streamable-http",
      "url": "https://deploy.example.com/mcp",
      "headers": { "X-Tenant": "public-tenant" }
    },
    "legacy-events": { "type": "sse", "url": "https://legacy.example.com/sse" }
  }
}
JSON
uv run python -c "
from pathlib import Path
from openhands.sdk.mcp.config import dump_mcp_config
from openhands.sdk.plugin import Plugin, detect_format
root = Path('/tmp/ap-example')
print('format :', detect_format(root).name)
servers = Plugin.load(root).mcp_config
for name, config in dump_mcp_config(servers).items():
    print(name, config)
print('redacted dump:', servers['deployment-api'].model_dump(mode='json')['headers'])"
```

This branch (`OH_PERSISTENCE_DIR=/tmp/oh-demo`):

```
format : agent-plugins
local-validator {'transport': 'stdio', 'command': '/private/tmp/ap-example/bin/validator',
 'args': ['--data', '/tmp/oh-demo/plugin-data/ap-example-6153411fe0a97c28/validator'],
 'env': {'CONFIG': '/private/tmp/ap-example/config.json',
         'PLUGIN_ROOT': '/private/tmp/ap-example',
         'PLUGIN_DATA': '/tmp/oh-demo/plugin-data/ap-example-6153411fe0a97c28'},
 'cwd': '/private/tmp/ap-example'}
deployment-api {'url': 'https://deploy.example.com/mcp', 'transport': 'streamable-http',
 'headers': {'X-Tenant': 'public-tenant'}}
redacted dump: {'X-Tenant': '**********'}
```

Line-wrapped for readability. `dump_mcp_config` is what the connection path uses; an ordinary dump redacts. `legacy-events` (`sse`) is reported and skipped; see Notes. On `main` the same package prints `claude-code`, loads as `ap-example` — the directory name — and has no MCP servers at all.

**Tests.**

```
uv run pytest tests/sdk -q            # 9 failed, 6347 passed  (branch)
uv run pytest tests/sdk -q            # 9 failed, 6197 passed  (22c85eb, same 9)
uv run pytest tests/sdk/plugin tests/sdk/mcp tests/sdk/test_settings.py \
  tests/sdk/conversation tests/cross/test_check_sdk_api_breakage.py -q
uv run pre-commit run --files <changed files>                    # all hooks pass
uv run python .github/scripts/check_persisted_settings_compat.py # 14 fixtures, 8 baselines
```

Both suite runs use an empty `OH_PERSISTENCE_DIR`; the 9 failures are identical on both sides and are that override's doing — they assert the default `~/.openhands` path. The delta against `main` is zero.

86 tests cover the document, entry isolation, stdio, remote, serialization, `PLUGIN_DATA` and expansion. Two run over real transports: one launches a stdio server from a package and checks the cwd, argv and environment it got; one sends a request through a same-origin and then a cross-origin redirect between local HTTP servers and checks which headers arrived where. `TestAgentPluginsMCPExpansion` covers `_ensure_plugins_loaded` and `load_plugin()`. Each regression test was checked by reverting its fix, and every one fails without it.

## Video/Screenshots

N/A — library change, no UI. The terminal output above is the evidence.

## Design Doc

N/A. The design follows the specification section by section; decisions that were open questions on #4405 are in Notes.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**Three of #4405's open questions are answered here**, and each is cheap to revisit:

1. **`sse` is not supported** (question 1). It is OPTIONAL in the spec and deprecated by MCP itself; an `sse` entry is reported and skipped, which is a conformant outcome (§7.2.2 rule 4). Adding it later is one entry in `SUPPORTED_TRANSPORTS`; withdrawing it would not be.
2. **`PLUGIN_DATA` is `~/.openhands/plugin-data/<dir name>-<hash of resolved plugin root>`** (question 5), built by the fetch cache's own `get_cache_path` and created only for a usable stdio server. Keyed by root, not manifest name: project plugins load ambiently and the spec puts `node_modules` and virtualenvs here, so name keying would let a repo ship a same-named plugin and plant code a trusted installed plugin then runs. Roots are already stable across updates (`installed/<name>`, a source-hashed fetch cache path, fixed project paths), so the data survives one. It sits outside `plugins/`, whose every child is loaded as a plugin, and is not deleted on uninstall — a `MAY` left for a follow-up.
3. **Plugin `headers` and `env`: exempt from expansion, still redacted** (question 6). The spec calls header values "visible package data, not a portable secret mechanism"; that forbids plugins from putting secrets there but does not require clients to display them, so redaction protects the user from a plugin that does it anyway. Plugin servers are rebuilt from the package on every load, so no stored plaintext is needed. The expansion exemption is the part §9.2 requires.

**Changes outside the plugin package.** Four call sites expanded `${VAR}` against the environment and per-conversation secrets with the same dump→expand→coerce block: `load_plugins()`, `_ensure_plugins_loaded`, and two in `LocalConversation.load_plugin()`. They now share `expand_mcp_servers()`, which leaves package servers alone; behavior for every other server is unchanged. In `mcp/utils.py`, package remote servers get an HTTP client whose request hook runs on every redirect hop and drops the configured header names once a request leaves the configured origin. FastMCP follows redirects and httpx strips only `Authorization`, so before this a `307` carried `X-Tenant` to a different origin; same-origin redirects still keep the headers. User-configured servers are unchanged and still have that httpx behavior, which is out of scope here.

**Not in scope.** Path containment here covers what §7.2.1 makes part of entry validity — `command` and `cwd`. The general rule for every other package path is #4453.

**Worth a reviewer's eye:** the top-level schema is derived from the vendored document by dropping `$defs` and loosening `mcpServers`, rather than restated, so the closed field set cannot drift — but it does mean the per-entry `$ref` is resolved from a second, hand-built validator.

---
<details>

**Variants & Base Images**
|---|---|---|---|
| python-slim | amd64, arm64 | `python-node-runtime` | [Link](https://hub.docker.com/_/python-node-runtime) |
| python-minimal | amd64, arm64 | `python-node-runtime` | [Link](https://hub.docker.com/_/python-node-runtime) |
| python | amd64, arm64 | `python-node-runtime` | [Link](https://hub.docker.com/_/python-node-runtime) |

```bash
docker pull ghcr.io/openhands/agent-server:c505866-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-c505866-python \
  ghcr.io/openhands/agent-server:c505866-python
```

```
ghcr.io/openhands/agent-server:c505866-golang-amd64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-golang-amd64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-golang-amd64
ghcr.io/openhands/agent-server:c505866-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:c505866-golang-arm64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-golang-arm64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-golang-arm64
ghcr.io/openhands/agent-server:c505866-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:c505866-java-amd64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-java-amd64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-java-amd64
ghcr.io/openhands/agent-server:c505866-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:c505866-java-arm64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-java-arm64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-java-arm64
ghcr.io/openhands/agent-server:c505866-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:c505866-python-amd64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-amd64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-amd64
ghcr.io/openhands/agent-server:c505866-python-node-runtime-amd64
ghcr.io/openhands/agent-server:c505866-python-arm64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-arm64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-arm64
ghcr.io/openhands/agent-server:c505866-python-node-runtime-arm64
ghcr.io/openhands/agent-server:c505866-python-minimal-amd64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-minimal-amd64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-minimal-amd64
ghcr.io/openhands/agent-server:c505866-python-node-runtime-minimal-amd64
ghcr.io/openhands/agent-server:c505866-python-minimal-arm64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-minimal-arm64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-minimal-arm64
ghcr.io/openhands/agent-server:c505866-python-node-runtime-minimal-arm64
ghcr.io/openhands/agent-server:c505866-python-slim-amd64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-slim-amd64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-slim-amd64
ghcr.io/openhands/agent-server:c505866-python-node-runtime-slim-amd64
ghcr.io/openhands/agent-server:c505866-python-slim-arm64
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-slim-arm64
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-slim-arm64
ghcr.io/openhands/agent-server:c505866-python-node-runtime-slim-arm64
ghcr.io/openhands/agent-server:c505866-golang
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-golang
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-golang
ghcr.io/openhands/agent-server:c505866-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:c505866-java
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-java
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-java
ghcr.io/openhands/agent-server:c505866-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:c505866-python-minimal
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-minimal
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-minimal
ghcr.io/openhands/agent-server:c505866-python-node-runtime-minimal
ghcr.io/openhands/agent-server:c505866-python-slim
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python-slim
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python-slim
ghcr.io/openhands/agent-server:c505866-python-node-runtime-slim
ghcr.io/openhands/agent-server:c505866-python
ghcr.io/openhands/agent-server:c505866439b08f929a109636edeba651892873dd-python
ghcr.io/openhands/agent-server:vasco-plugin-mcp-json-python
ghcr.io/openhands/agent-server:c505866-python-node-runtime
```

- Each variant tag (e.g., `c505866-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `c505866-python-amd64`) are also available if needed
</details>


openhands-sdk/openhands/sdk/mcp/config.py
```diff
@@ -554,6 +558,30 @@ def _validate_server(self) -> MCPServer:
                 )
         return self
 
+    @property
+    def literal_values(self) -> bool:
+        """Whether this server's config is literal package data, never expanded.
```
VascoSch92: make this docstring shorter.

openhands-sdk/openhands/sdk/mcp/config.py
```diff
@@ -554,6 +558,30 @@ def _validate_server(self) -> MCPServer:
...
+        :meth:`as_literal`. ``${VAR}`` placeholders in them are left as written
+        rather than resolved against the environment or per-conversation
+        secrets. ``env`` and ``headers`` are still redacted like any other
+        secret. The flag lives in memory only: it is never read from input nor
+        written by a dump, so data from outside can neither set it nor carry it.
+        """
+        return self._literal_values
+
+    def as_literal(self) -> MCPServer:
+        """Return a copy that is exempt from ``${VAR}`` expansion.
```
VascoSch92: make this docstring shorter.

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
```diff
@@ -161,9 +165,26 @@ def load_manifest(self, plugin_dir: Path) -> PluginManifest:
...
+    def load_mcp_config(self, plugin_dir: Path) -> dict[str, MCPServer]:
+        """Load MCP servers from the root ``mcp.json`` (no leading dot).
+
+        Unlike the Claude Code format, expansion is complete here: Agent Plugins
+        defines exactly two placeholders and forbids every other kind, so there
+        is nothing left to expand once per-conversation secrets arrive.
+        """
+        # Imported lazily: the loader reads its schema from this module, and
+        # installed.py reaches plugin.py, which imports this package.
+        from openhands.sdk.plugin.format.agent_plugins_mcp import load_mcp_servers
```
VascoSch92: Do we need to import here? Can you double check?

openhands-sdk/openhands/sdk/plugin/format/agent_plugins_mcp.py
```diff
@@ -0,0 +1,353 @@
...
+placeholder expansion (§9.2), path containment (§4.1, §7.2.1) and remote URL /
+header validation — all applied per entry, so one bad server never takes down
+its siblings or another component type (§7.2.2).
+
+Lives beside ``agent_plugins.py`` rather than inside it: the manifest and the
+MCP configuration are separate documents with separate schemas and separate
+failure boundaries.
+"""
+
+from __future__ import annotations
```
VascoSch92: Do we need that?

openhands-sdk/openhands/sdk/plugin/installed.py
```diff
@@ -22,12 +23,36 @@
...
+DEFAULT_PLUGIN_DATA_DIR = get_user_persistence_dir() / "plugin-data"
+
 
 def get_installed_plugins_dir() -> Path:
     """Get the default directory for installed plugins."""
     return DEFAULT_INSTALLED_PLUGINS_DIR
 
 
+def get_plugin_data_dir(plugin_root: Path, *, data_root: Path | None = None) -> Path:
+    """Get the persistent data directory for one plugin instance (``PLUGIN_DATA``).
```
VascoSch92: make the docstring smaller

openhands-sdk/openhands/sdk/plugin/installed.py
```diff
@@ -22,12 +23,36 @@
 
 DEFAULT_INSTALLED_PLUGINS_DIR = get_user_persistence_dir() / "plugins" / "installed"
 
+#: Parent of the per-plugin ``PLUGIN_DATA`` directories (Agent Plugins §9.1).
```
VascoSch92: make the comment smaller thanks

openhands-sdk/openhands/sdk/mcp/utils.py
```diff
@@ -114,6 +117,52 @@ def _oauth_auth_from_authentication_config(
...
+                request.headers.pop(name, None)
+
+    def factory(
+        headers: dict[str, str] | None = None,
+        timeout: httpx.Timeout | None = None,
+        auth: httpx.Auth | None = None,
+        **_: object,
+    ) -> httpx.AsyncClient:
+        client = create_mcp_http_client(headers=headers, timeout=timeout, auth=auth)
+        client.event_hooks = {"request": [drop_headers_off_origin], "response": []}
```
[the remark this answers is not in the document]
VascoSch92: No change needed: the lockfile pins `mcp` 1.28.1, whose `mcp/shared/_httpx_utils.py` imports plain `httpx` (0.28.1) and returns `httpx.AsyncClient`. `httpx2` is not a dependency here or on `main`, so the annotations match the runtime types.

openhands-sdk/openhands/sdk/mcp/config.py
VascoSch92: Shortened in 75a190689.

openhands-sdk/openhands/sdk/mcp/config.py
VascoSch92: Shortened in 75a190689.

openhands-sdk/openhands/sdk/plugin/format/agent_plugins.py
VascoSch92: They were needed (both imports formed cycles), so I removed the cycles instead: `_load_schema` moved to `format/base.py` and `get_plugin_data_dir` to the `mcp.json` loader, its only user. Both are module-level imports now (75a190689).

openhands-sdk/openhands/sdk/plugin/format/agent_plugins_mcp.py
VascoSch92: No, nothing needs deferred annotations. Removed in 75a190689.

openhands-sdk/openhands/sdk/plugin/installed.py
VascoSch92: Shortened, and moved with the function to `agent_plugins_mcp.py` (75a190689).

openhands-sdk/openhands/sdk/plugin/installed.py
VascoSch92: Shortened, and moved to `agent_plugins_mcp.py`; `installed.py` is back to `main` (75a190689).

enyst (commented):
🐾 smolpaws here — reviewing as **Claude Fable 5.1**, LLM profile `fable-5-1-eval`.

Read the loader, the vendored schema, the MCP plumbing and the new tests; ran `pytest -n auto tests/sdk/plugin tests/sdk/mcp` → 450 passed. (One `test_mcp_session_persistence` case failed under `-n auto` and passes single-process, so I read that as a parallel flake, not this branch.)

No blockers. Four things I checked, or want to confirm:

**1. The origin-bound header drop works — verified, not just claimed.** I built the client from `_origin_bound_client_factory` against a mock transport and had `a.example` 307-redirect to `b.example`: the configured headers reach the configured origin and are gone on the cross-origin hop. The "request hooks run on every redirect hop" comment holds for this path.

**2. How long does `literal_values` live?** The flag is in-memory, granted by the loader, and (as your test states) does not survive a data round trip. `_ensure_plugins_loaded` rebuilds package servers on every load, so the fresh copies carry the flag and win the merge by name. But if a plugin is disabled or removed between sessions while its servers are still in the persisted `agent.mcp_config`, those restored copies are ordinary config again, and `expand_mcp_servers` would expand `${SECRET}` in their `args`/`env`. Local state is trusted, so this may be an accepted boundary — but the guarantee is presented as a security property, so it's worth stating its lifetime, or marking plugin-sourced servers in state.

**3. `_load_schema` is shared as a private name.** It moved to `base.py` and is now imported by two other modules; a public name (e.g. `load_vendored_schema`) would read better than importing an underscore helper three ways.

**4. PLUGIN_DATA is created before the entry is known to be valid.** `_stdio_fields` does the `mkdir`, then `MCPServer.model_validate` runs in `_load_server`, while the comment says "once this entry is known to be a usable stdio server". An entry that passes the JSON schema but fails model validation still leaves the directory behind. Either move the `mkdir` after validation or soften the comment.

Minor: `DEFAULT_PLUGIN_DATA_DIR` is resolved at import time, so a caller that sets the persistence env var afterwards won't move the default — `plugin_data_root` covers it, it's just worth a docstring line.

Liked: the per-entry failure boundaries, `utf-8-sig`, the header name/value validation, and dropping client-generated headers instead of letting a package forge `Host`. The tests cover the boundaries they claim.

enyst:
HUMAN:
Notes on `smolpaws` review:

> **1. The origin-bound header drop works — verified, not just claimed.**

OK.

> **2. How long does `literal_values` live?**

Seems worth looking into; either responding or tweaking or testing for edge case.

> **3. `_load_schema` is shared as a private name.**

Ha. This is the kind of things I personally really didn’t like in the Old Times (tm), because it’s a tiny thing in Python but it breaks expectations of human code writers… and yeah, ok, there are no human code writers anymore.

That said, this is worth fixing, but I think it’s exactly in-scope for a periodic automated pass or linting pass. We’d better off focusing on that, rather than in-PR.

> **4. PLUGIN_DATA is created before the entry is known to be valid.**

Seems worth fixing, it’s unexpected or, well, feels unpolished, to create useless directories on user’s machine, if I understand this correctly.

VascoSch92:
Hey @enyst 

thanks for the review. 

These should be fixed.

You can see the summary here if you want:

**2. Lifetime of `literal_values`.** smolpaws is right, and it is worse than the comment implies: no cipher is needed. `env`/`headers` are encrypted or redacted on the way into `base_state.json`, but `command`, `args`, `url` and `cwd` are plain strings with no secret serializer, so a package's `args: ["${SECRET_TOKEN}"]` is persisted verbatim. Reproduced on this branch:

```python
pkg = MCPServer.model_validate({"transport": "stdio", "command": "echo",
                                "args": ["${SECRET_TOKEN}"]}).as_literal()
persisted = pkg.model_dump(mode="json")          # {'args': ['${SECRET_TOKEN}'], ...}
restored = MCPServer.model_validate(persisted)   # literal_values == False
expand_mcp_servers({"s": restored}, {"SECRET_TOKEN": "s3cr3t"}.get)["s"].args
# ['s3cr3t']
```

The path: `_ensure_plugins_loaded` seeds `merged_mcp` from the persisted agent (`local_conversation.py:975`), writes the merged result back (`:1234`), and nothing prunes a server whose plugin is gone. While the plugin is installed the fresh literal copy wins the name-keyed merge (`plugin.py:200`), which is what hides this. A restored server also loses the off-origin header guard, since that is keyed on the same flag (`mcp/utils.py:177`).

So the guarantee holds for as long as the package is the source of truth, and ends at the first restore without it. Two ways to make that structural, both bigger than this PR:

- Keep package servers out of the persisted snapshot. They are rebuilt from the package on every load, so persisting them is redundant, and the flag is still live at save time.
- Record the source in state and prune servers with no backing plugin on load.

I prefer the first. Happy to file it as a follow-up under #4405, or do it here if you would rather not merge the narrower guarantee — your call.

**4. `PLUGIN_DATA` created before the entry validates.** Correct, fixed in 064f4b486: the `mkdir` moved out of `_stdio_fields` into `_load_server`, after `MCPServer.model_validate`. Added a test that fails with the old ordering. Also added the docstring line about `DEFAULT_PLUGIN_DATA_DIR` being bound at import.

**3. `_load_schema` as a shared private name.** Agreed, leaving it for the lint pass rather than this PR. It is a four-line rename whenever you want it.

**1.** Thanks for actually building the client and checking the redirect hop.

enyst:
If I understand correctly, we could leave the MCP servers out? eg install them separately as we already do

enyst (approved):

enyst:
Ah, I see, it's literals inside MCP. I was hoping it could be even simpler, alas it's MCP 😅

VascoSch92:
@enyst Thanks

smolpaws:
🐾 Correction to my review above, from smolpaws (the account that wrote it): I introduced myself as Claude Fable 5.1 / `fable-5-1-eval`, and that was wrong. This review was produced in a conversation running **DeepSeek-v4-Flash** (profile `deepseek-v4-flash`) — that is the profile the WhatsApp/Slack lane actually uses. The findings stand on their own, but the model attribution should be right, and now it is.

Issue #4958, opened 2026-09-11
[Agent Profile] Clients can't ask what tools an agent will get, or which tools are pickable


simonrosenberg:
A client building an Agent Profile editor has to answer two questions about tools, and the SDK can answer neither:

1. **What will this profile's agent actually get?**
2. **Which tools may a user pick from?**

Both are facts the SDK owns, so every client ends up guessing — OpenHands/OpenHands#17235 currently hardcodes a six-name allow-list and a copy of `DEFAULT_EXEC_TOOL_NAMES` to fill the gap.

## 1. `materialize` reports intent, not outcome

`POST /api/agent-profiles/{name}/materialize` runs `resolve_agent_profile_dry_run` and is otherwise the right endpoint for this. But `_build_openhands_settings` passes the field straight through (`"tools": profile.tools`), so a profile that leaves `tools` unset materializes as `None` — while the launch resolves it to four tools.

Verified against a live agent-server (1.46.0), same profile:

```
POST /api/agent-profiles/probe-default/materialize
  resolved_settings.tools : None

POST /api/conversations  {agent_profile_id: <that profile>}
  agent.tools : ['browser_tool_set', 'file_editor', 'task_tracker', 'terminal']
```

The gap is not just `create_agent()`'s defaulting. `_resolve_agent_from_profile` appends `browser_tool_set` *after* `resolve_agent_profile` returns, gated on `is_tool_usable()` **and** on `profile.tools is None` — a rule no client can reproduce without duplicating both the default names and the injection logic, and which drifts the moment either changes.

### Ask

Have the dry-run also report the resolved outcome — the tool names the agent would actually be constructed with, after `create_agent()`'s defaulting and the serving layer's injections — alongside the stored tri-state, which is still what gets persisted.

A second, smaller ask that makes this usable in a create flow: let materialize accept a **draft profile body**, not only a saved profile by name, so an editor can preview before the first save.

Shape is up to you; something like `AgentProfileDiagnostics.resolved_tool_names: list[str]` next to the existing `resolved_mcp_config_keys` / `resolved_skills`, which already follow exactly this "what would actually resolve" pattern.

## 2. Nothing marks a tool as user-selectable

`/server_info`'s `usable_tools` and `/tools/` both return the process's tool registry. `tool_router.py` registers four presets unconditionally at import:

```python
register_default_tools(enable_browser=True)
register_builtins_agents(enable_browser=True)
register_gemini_tools(enable_browser=True)
[...]

Issue #4542, opened 2026-08-19
Global agent_context.load_memory preference is ignored unless the conversation is launched from an agent profile


VascoSch92:
## Summary

`agent_settings.agent_context.load_memory` is exposed as a **global user preference**, but it is only applied when a conversation is started via `agent_profile_id`. Conversations started with a concrete `agent` (or an `agent_settings` payload) silently lose the setting, so persistent memory is never injected.

There is no error and no warning. `<MEMORY_CONTEXT>` is simply absent, and `<MEMORY>` renders its *other* variant, telling the agent that `AGENTS.md` is its persistent memory — which reads like a working feature.

## Impact

Measured on a self-hosted Agent Canvas deployment with `load_memory: true` in settings:

- **642 of 767** conversations launched with `load_memory: false`
- **0 of 767** ever rendered a `<MEMORY_CONTEXT>` block
- The 24 that did carry `load_memory: true` were ones whose client happened to copy the stored `agent_context` verbatim — it worked by accident, not by design

Stored preference vs. what actually launched, for a representative conversation:

| field | settings | as launched |
|---|---|---|
| `load_memory` | `True` | `False` |
| `load_user_skills` | `False` | `True` |
| `load_project_skills` | `False` | `True` |
| `load_public_skills` | `False` | `True` |
| `skills` | 0 | 57 |

## Root cause

`StartConversationRequest` accepts three mutually-exclusive shapes (`openhands-sdk/openhands/sdk/conversation/request.py:280`):

1. `agent_profile_id` — resolved server-side in `conversation_service`
2. `agent` — a concrete agent payload
3. `agent_settings` — converted to `agent` by the `_populate_agent_from_settings` validator (`mode="before"`), so `request.agent` is already populated before the service sees it

The `load_memory` stamp only exists on path 1. `_resolve_agent_from_profile` applies it at `openhands-agent-server/openhands/agent_server/conversation_service.py:392`:

```python
if load_memory:
    context = agent.agent_context or AgentContext()
    agent = agent.model_copy(
        update={"agent_context": context.model_copy(update={"load_memory": True})}
    )
```

and that function is only called at `:1480`, inside `if request.agent_profile_id is not None:`. Paths 2 and 3 never touch it.

The helper's own docstring already makes the argument:

> An `AgentProfile` has no `agent_context` field, so the preference cannot ride the profile — it is stamped onto the resolved agent below, **else a profile-launched conversation would silently ignore the setting.**

[...]

PR #5154, opened 2026-09-17
feat(agent-profiles): one launch pipeline for every conversation start

### LINKED ISSUE #5141: [Agent Profile] The `default` profile and named profiles build different agents — collapse launch into one pipeline

This is the root cause behind a series of already-fixed issues, each of which fixed one field on one path: #3967, #4014, #4016, #4542. It is also the root cause of open issues #4533, OpenHands/OpenHands#16205 and OpenHands/OpenHands#16951. #4958 (tool catalog and resolved tools) depends on it.

### Problem or Use Case

**Launching the `default` Agent Profile and launching a named Agent Profile build different agents from the same stored settings.** The launch has several code paths, and each one sets agent fields in its own way. Every profile field has to be implemented, and kept correct, once per path.

#### Live evidence

Setup: agent-server 1.49.1 built from `main` ([`365ceb3`](https://github.com/OpenHands/software-agent-sdk/commit/365ceb3a77f6dbe7b993041ddf04b9484c002e95)), Agent Canvas `main` ([`07590fd`](https://github.com/OpenHands/OpenHands/commit/07590fd5214015b10ff824d06254c56a8e7e7fb2)), and the e2e mock LLM, which records every completion request.

I saved two profiles with **identical contents**, one named `default` and one named `default-copy`, and launched each from the canvas home page:
- Canvas sends `default` through the `agent_settings` path.
- Canvas sends `default-copy` through the `agent_profile_id` path.

| Field set on the profile | `default` (`agent_settings` path) | `default-copy` (`agent_profile_id` path) |
|---|---|---|
| `tools: [terminal, glob]` | ignored: terminal, file_editor, task_tracker, browser_tool_set | applied |
| `system_message_suffix` | ignored | applied |
| `mcp_server_refs: []` | ignored: the `fetch` MCP server is still attached | applied |
| `disabled_skills`, `enable_switch_llm_tool: false`, `tool_concurrency_limit: 3`, condenser disabled | all ignored | all applied |
| Skills in the system prompt, with nothing customised | 16 (canvas's bundled set) | 68 (server discovery) |
| "The current date and time is" | **11:54:29, when settings were last saved**, for a launch at 12:10 | 12:10 |
| Setting `secret_refs` | moves `default` onto the `agent_profile_id` path | applied |

The model sees a different agent depending on the profile's **name**. Profile edits made to `default` do nothing on the most common launch, the home page.

#### Why

**1. Two definitions of the default agent.**
- Global `agent_settings` is written by the settings pages and by LLM-profile activation.
- The `default` Agent Profile is copied from it **once** ([`build_seed_profile`](https://github.com/OpenHands/software-agent-sdk/blob/365ceb3a77f6dbe7b993041ddf04b9484c002e95/openhands-sdk/openhands/sdk/profiles/seed.py#L29)). After that it is written by the profile editor, and onboarding overwrites it with `{agent_kind, llm_profile_ref}`.
- Agent-profile activation only moves the active-profile pointer ([`agent_settings_applied` is always `False`](https://github.com/OpenHands/software-agent-sdk/blob/365ceb3a77f6dbe7b993041ddf04b9484c002e95/openhands-agent-server/openhands/agent_server/agent_profiles_router.py#L98)).
- Nothing keeps the two in sync.

**2. Three request inputs that do different amounts of work.** `StartConversationRequest` accepts `agent`, `agent_settings` or `agent_profile_id`.
- `agent_settings` becomes an agent inside a Pydantic validator ([`request.py#L321-L345`](https://github.com/OpenHands/software-agent-sdk/blob/365ceb3a77f6dbe7b993041ddf04b9484c002e95/openhands-sdk/openhands/sdk/conversation/request.py#L321-L345)), before the server does anything.
- The server steps run only in the `agent_profile_id` branch ([`conversation_service.py#L1685`](https://github.com/OpenHands/software-agent-sdk/blob/365ceb3a77f6dbe7b993041ddf04b9484c002e95/openhands-agent-server/openhands/agent_server/conversation_service.py#L1685), [`_resolve_agent_from_profile`](https://github.com/OpenHands/software-agent-sdk/blob/365ceb3a77f6dbe7b993041ddf04b9484c002e95/openhands-agent-server/openhands/agent_server/conversation_service.py#L354)):
  - skill discovery
  - browser injection
  - forced streaming
  - `secret_refs` enforcement
  - `LaunchedAgentProfile` provenance
  - a fresh `current_datetime`

**3. A second copy of the server pipeline.** The Docker runtime's [`prepare_start`](https://github.com/OpenHands/software-agent-sdk/blob/365ceb3a77f6dbe7b993041ddf04b9484c002e95/openhands-agent-server/openhands/agent_server/docker_runtime/mediation.py#L56) repeats the profile branch (resolution, secret filtering, the `load_memory` stamp) and can drift from `conversation_service`.

**4. Clients pick the path, and rebuild the agent themselves on the `agent_settings` path.**

Agent Canvas chooses the path in [`use-create-conversation.ts`](https://github.com/OpenHands/OpenHands/blob/07590fd5214015b10ff824d06254c56a8e7e7fb2/src/hooks/mutation/use-create-conversation.ts#L137-L244) with these rules:
- The profile is **named** `default`.
- The profile's LLM reference is missing, or the LLM list fails to load.
- The active LLM differs from the profile's pinned LLM.
- A veto when `secret_refs` is set.

The same rule is copied in [`use-llm-configured.ts`](https://github.com/OpenHands/OpenHands/blob/07590fd5214015b10ff824d06254c56a8e7e7fb2/src/hooks/use-llm-configured.ts#L73-L83), and the copies already disagree (OpenHands/OpenHands#16951).

Launches that ignore the profile entirely:
- `/new`
- locally launched child conversations
- the local planner, which sends a hand-built `agent`

On the `agent_settings` path, canvas assembles the agent itself: [`buildConfiguredOpenHandsAgentSettings`](https://github.com/OpenHands/OpenHands/blob/07590fd5214015b10ff824d06254c56a8e7e7fb2/src/api/agent-server-adapter.ts#L980), [`getAgentTools`](https://github.com/OpenHands/OpenHands/blob/07590fd5214015b10ff824d06254c56a8e7e7fb2/src/api/agent-server-adapter.ts#L721), [`buildBundledSkills`/`buildAgentContext`](https://github.com/OpenHands/OpenHands/blob/07590fd5214015b10ff824d06254c56a8e7e7fb2/src/api/agent-server-adapter.ts#L797). It uses:
- the account-wide active LLM instead of `llm_profile_ref`;
- a bundled skill catalog filtered by a global allow-list, instead of server discovery minus `disabled_skills`;
- a `<RUNTIME_SERVICES>` block that **replaces** the stored suffix;
- its own tool list, gated by a Vite flag;
- global condenser, verification (including `critic_api_key`), sub-agents, switch-LLM and concurrency settings;
- the whole MCP config, unfiltered;
- secrets with no scoping;
- no `acp_startup_timeout` for ACP.

Canvas never sends `agent_launch_additions`, the field that was added to carry `<RUNTIME_SERVICES>` on the profile path (#4029).

**5. Cloud has its own version of the problem.** The cloud app server resolves the profile itself and sends a fully built `agent`, so the agent-server's profile path never runs there. It then:
- overwrites several fields the profile carries (`tools`, `system_message_suffix`, the switch-LLM off-switch);
- always adds its platform MCP server;
- launches without the profile when resolution fails, with nothing surfaced to the user;
- launches sub-conversations without the parent's profile.

#### Cost

Every new profile field (`tools`, `secret_refs`, prompt templates in #4956, tool metadata in #4958) has to be implemented once per path, and each path is a new place for a silent mismatch. The issues listed at the top are that cost already paid. For a scoped agent it is also a correctness problem: a restriction on the profile only holds on the launch paths that honour it.

### Desired Behavior

**One launch pipeline.** Every conversation start goes through the same resolution: default or named profile, home page or `/new`, child conversation, automation, local or Docker runtime, cloud. The only inputs that vary are *which* profile and an explicit, additive set of launch-time additions.

1. **The Agent Profile is the only definition of an agent.**
   - `default` is an ordinary profile: no name-based rules anywhere.
   - Global settings keep only the shared pieces profiles refer to: LLM profiles, the MCP server list, secrets, the skill catalog and marketplaces.
   - Behaviour settings (condenser, verification, memory, sub-agents, switch-LLM, concurrency, tools, instructions, skill enablement) live on the profile. Settings pages that edit them edit the active profile.
2. **One SDK function builds the launch.** Something like `prepare_agent_launch(profile, *, stores, mcp_catalog, skill_catalog, runtime, additions) -> LaunchPlan(agent, launched_profile, allowed_secrets, resolved_view)`.
   - It owns every field:
     - LLM and streaming
     - MCP filtering
     - skill catalog minus `disabled_skills`
     - project-skill loading
     - suffix plus additions
     - `current_datetime` (always computed at launch)
     - `load_memory`
     - tool defaults and runtime-dependent injections
     - sub-agents, switch-LLM, concurrency, condenser, verification
     - secret scope
     - ACP fields and ACP skill sourcing
   - Callers:
     - `conversation_service`
     - Docker mediation (its duplicate branch is deleted)
     - `materialize`, as a side-effect-free call of the *same* function, so diagnostics can never disagree with the launch
     - the cloud app server
   - Runtime-dependent decisions (e.g. whether browser is available) come in as an explicit `runtime` input, so a caller that is not the runtime can pass the runtime's answer.
3. **One request shape for product clients.**
   - `agent_profile_id` (a stored profile), or `agent_profile` (an inline draft of the same model, resolved by the same function; this also covers draft `materialize` for #4958).
   - `agent_settings` is deprecated. During the deprecation window the server converts it into an inline profile, so it takes the same pipeline instead of a validator shortcut.
   - Raw `agent` stays as the explicit low-level option for SDK users building agents in code.
4. **Launch-time variation is explicit and additive, and applied the same way on every path.** `agent_launch_additions` carries:
   - the runtime-services suffix append (already exists);
   - a per-launch `llm_profile_ref` override for the chat LLM picker, recorded in `LaunchedAgentProfile`.
   - `client_tools` stays as it is.
   - Additions can never widen a profile's scope.
5. **Failures are explicit.** A dangling LLM or MCP reference fails the launch with a structured error that clients can turn into a fix-it prompt (`materialize` reports it first). No client or server silently switches to a different path or to an unscoped agent.
6. **Derived launches keep their profile.** Child conversations, `/new` and sub-conversations launch with the parent's launched profile unless a different one is chosen explicitly.

### Acceptance Criteria

- [ ] A single SDK function builds the agent for every launch. `conversation_service`, Docker mediation and `materialize` all call it, and no other code in the agent-server builds a launch agent from a profile.
- [ ] `materialize` and a real launch produce the same resolved agent for the same profile. A test compares them field by field: tools, MCP keys, skills, suffix, LLM, condenser, verification, sub-agents, switch-LLM, concurrency, secret scope.
- [ ] Two profiles with identical contents and different names (one of them `default`) produce identical agents, apart from launch timestamps and ids, on every product launch path. A cross-repo e2e test reproduces the table above and asserts every row matches.
- [ ] `current_datetime` in the system prompt is computed at launch on every path.
- [ ] `agent_settings` on `StartConversationRequest` is deprecated (via the deprecation helper). While it is still accepted, it goes through the same pipeline via inline-profile conversion.
- [ ] `StartConversationRequest` accepts an inline `agent_profile` draft, resolved by the same function.
- [ ] `AgentLaunchAdditions` supports a per-launch `llm_profile_ref` override, recorded in `LaunchedAgentProfile`; additions cannot widen scope.
- [ ] A dangling LLM or MCP reference returns a structured error from the start endpoint, with no silent fallback.
- [ ] Follow-up in OpenHands/OpenHands: canvas always launches with a profile (active, explicit, or the parent's for children and `/new`). It deletes the `default` name rule, the LLM-mismatch and dangling-ref fallbacks, the `use-llm-configured` copy, and client-side agent assembly (`buildConfigured*AgentSettings`, `getAgentTools`, bundled-skill launch injection). It sends `<RUNTIME_SERVICES>` via `agent_launch_additions`.
- [ ] Follow-up for cloud: resolve through the same SDK function, stop overwriting fields the profile carries, and surface resolution failures instead of launching without the profile.
- [ ] One-time migration: the `default` profile is refreshed from current global `agent_settings` (what `default` launches actually ran), and the settings pages edit the active profile from then on.

### Alternatives Considered

- **Keep both paths and fix fields one by one.** This is what #3967, #4014, #4016, #4029 and #4542 did. Each fix closed one gap on one path, and the table above shows the paths still differ on nearly every field. Every new profile field restarts the cycle.
- **Make canvas's `agent_settings` builder mirror the resolver.** This keeps two implementations of the same logic in two languages, and still leaves global settings and the `default` profile as two sources of truth.
- **Keep global `agent_settings` as the default and treat profiles as overlays.** Profiles already store concrete values rather than deltas, and the design settled in OpenHands/OpenHands#15563 puts behaviour settings on the profile only. Overlays would add a third composition model.

### Priority / Severity

High - Significant impact on productivity

### Estimated Scope

Large - Significant feature requiring architecture changes

### Feature Area

- [x] Tools / Tool system
- [x] Skills / Plugins
- [x] Agent Server
- [ ] Workspace management
- [x] Configuration / Settings
- [ ] Examples / Templates
- [ ] Documentation
- [ ] Other

Suggested order, each step shippable on its own:

1. **SDK: extract the pipeline.**
   - Move everything in `_resolve_agent_from_profile` and the profile branch of `_start_conversation`, plus the Docker mediation copy, into one SDK function that takes explicit stores, catalogs and runtime inputs.
   - Make `resolve_agent_profile_dry_run` call it with side effects off.
   - Keep the current request shapes.
2. **SDK: inline profiles and deprecation.**
   - Add `agent_profile` to `StartConversationRequest`.
   - Convert `agent_settings` into an inline profile and deprecate it.
   - Move the `agent_settings` → agent conversion out of the Pydantic validator and into the server pipeline.
3. **SDK: launch additions.** Add the `llm_profile_ref` override. Return structured errors for dangling references.
4. **Canvas.** Send a profile on every launch path, delete the routing rules and client-side agent assembly, move the behaviour settings pages onto the active profile, and migrate the global skill toggles into the profile's `disabled_skills`.
5. **Cloud.** Adopt the SDK function and remove the field overwrites.

Decisions to settle before step 4:
- **Skill model.** Canvas has a global allow-list (`enabled_skills`); the SDK profile has a deny-list (`disabled_skills`). Proposal: keep the per-profile deny-list and migrate the global toggles into it.
- **Chat LLM picker.** Per-launch override (proposed) or an edit to the profile.
- **Tools that ignore profile scope.** The cloud platform MCP server and plugin-provided MCP servers are currently added outside `mcp_server_refs`. They need an explicit policy inside the function: exempt from profile scoping, or subject to it.
- **Cloud fallback policy.** Cloud currently launches without the profile when resolution fails so that no launch is ever blocked. Keep that with a visible warning, or fail the launch.

### Additional Context

Related:
- #4958: tool catalog and resolved tools; needs this for `materialize` to match the launch.
- #4953: sub-agents bypass profile scope.
- #4956: prompt templates.
- #4964: self-describing profile schema.
- #5030: secret scoping.
- #4533: seeded `default` drops the LLM key.
- OpenHands/OpenHands#16205: runtime services missing on profile launches.
- OpenHands/OpenHands#16951: home readiness vs launch disagree.

simonrosenberg:
HUMAN:

Filing the SDK half of #5141: the `default` profile and named profiles have to build the same agent from one pipeline, so new profile fields stop needing an implementation per launch path. Canvas and cloud adoption follow separately.

---

AGENT:

## Why

Launching the `default` Agent Profile and launching a named one built **different agents from the same stored settings**, because the launch had several code paths and each set agent fields its own way: a Pydantic validator for `agent_settings`, the `agent_profile_id` branch of `conversation_service`, a second copy of that branch in the Docker runtime's `prepare_start`, and a separate dry-run for `materialize`. Every profile field had to be implemented, and kept correct, once per path — #3967, #4014, #4016 and #4542 were that cost paid one field at a time.

This is step 1–3 of #5141 (the SDK steps). Canvas and cloud adoption are the follow-ups listed below.

## Summary

- **One function builds every launch.** `openhands.sdk.profiles.prepare_agent_launch(source, *, catalog, runtime, additions, profile_origin, build_agent)` resolves a profile's `llm_profile_ref` / `mcp_server_refs` / `disabled_skills` and owns the launch-time fields: tool defaults and browser injection, forced streaming, skill catalog and ACP skill sourcing, project-skill loading, suffix + additions, `current_datetime`, `load_memory`, and the secret scope. Runtime-dependent answers come in as an explicit `AgentLaunchRuntime`, so the Docker runtime can pass *its container's* answer instead of the host's. `conversation_service`, Docker mediation and `materialize` all call it through the new `agent_server/agent_launch.py`; `_resolve_agent_from_profile`, `_with_load_memory` and `_apply_acp_skill_sourcing` are gone, as is the duplicated profile branch in `prepare_start`.
- **`materialize` is the same call with side effects off** (`build_agent=False`, which skips `create_agent()` — the only step that can refresh a subscription LLM's credentials over the network). Its verdict and `resolved_settings` now come from the launch itself.
- **New request shapes.** `StartConversationRequest` accepts an inline `agent_profile` draft (resolved exactly like a stored one, never saved), and `agent_settings` is deprecated (`deprecated: true` in OpenAPI, removal target v1.55.0). During the window the server converts it into an inline profile with the payload's own LLM/MCP/skills as its catalog, so it takes the same pipeline instead of the validator shortcut; fields no profile models (`critic_api_key`, `user_message_suffix`, `agent_context.secrets`, `acp_isolate_data_dir`, …) are carried through, not dropped.
- **`AgentLaunchAdditions.llm_profile_ref`** gives the chat LLM picker a per-launch override, recorded in `LaunchedAgentProfile` (which also gained `inline`). Additions stay additive: they carry no tools, MCP servers, skills or secrets, and a test asserts a scoped profile's tools/MCP/skills/secret scope are byte-identical with and without them.
- **One structured failure.** A dangling LLM profile *or* MCP server ref raises `UnresolvedProfileReferences` and the start endpoint returns 422 with `{"code": "unresolved_profile_references", "message", "dangling_llm_profile_ref", "dangling_mcp_server_refs"}` — no silent fallback to a different path.

### REST API contract changes

Compared with base OpenAPI `c00f26a0c602` for public `/api/**` paths.

```diff
--- base public OpenAPI
+++ head public OpenAPI
@@ -752,0 +753,15 @@
+schema ACPAgentProfile property acp_args optional schema=anyOf=[type="array" items=type="string",type="null"]
+schema ACPAgentProfile property acp_command optional schema=anyOf=[type="string",type="null"]
+schema ACPAgentProfile property acp_model optional schema=anyOf=[type="string",type="null"]
+schema ACPAgentProfile property acp_prompt_timeout optional schema=type="number" default=1800.0 exclusiveMinimum=0.0
+schema ACPAgentProfile property acp_server optional schema=type="string" enum=["claude-code","codex","gemini-cli","kimi-code","pi","opencode","custom"] default="claude-code"
+schema ACPAgentProfile property acp_session_mode optional schema=anyOf=[type="string",type="null"]
+schema ACPAgentProfile property acp_startup_timeout optional schema=type="number" default=90.0 exclusiveMinimum=0.0
+schema ACPAgentProfile property agent_kind optional schema=type="string" const="acp" default="acp"
+schema ACPAgentProfile property id optional schema=type="string" format="uuid"
+schema ACPAgentProfile property mcp_server_refs optional schema=anyOf=[type="array" items=type="string",type="null"]
+schema ACPAgentProfile property name required schema=type="string" minLength=1
+schema ACPAgentProfile property revision optional schema=type="integer" default=0 minimum=0.0
+schema ACPAgentProfile property schema_version optional schema=type="integer" default=2 minimum=1.0
+schema ACPAgentProfile property secret_refs optional schema=anyOf=[type="array" items=type="string",type="null"]
+schema ACPAgentProfile type="object" additionalProperties=false
@@ -938,0 +954 @@
+schema AgentLaunchAdditions property llm_profile_ref optional schema=anyOf=[type="string" minLength=1,type="null"]
@@ -1883,0 +1900,9 @@
+schema LLMSummarizingCondenserSettings property condenser_kind optional schema=type="string" const="llm_summarizing" default="llm_summarizing"
+schema LLMSummarizingCondenserSettings property enabled optional schema=type="boolean" default=true
+schema LLMSummarizingCondenserSettings property hard_context_reset_context_scaling optional schema=type="number" default=0.8 exclusiveMinimum=0.0 exclusiveMaximum=1.0
+schema LLMSummarizingCondenserSettings property hard_context_reset_max_retries optional schema=type="integer" default=5 exclusiveMinimum=0.0
+schema LLMSummarizingCondenserSettings property keep_first optional schema=type="integer" default=2 minimum=0.0
+schema LLMSummarizingCondenserSettings property max_size optional schema=type="integer" default=240 minimum=20.0
+schema LLMSummarizingCondenserSettings property max_tokens optional schema=anyOf=[type="integer" exclusiveMinimum=0.0,type="null"]
+schema LLMSummarizingCondenserSettings property minimum_progress optional schema=type="number" default=0.1 exclusiveMinimum=0.0 exclusiveMaximum=1.0
+schema LLMSummarizingCondenserSettings type="object"
@@ -1884,0 +1910,2 @@
+schema LaunchedAgentProfile property inline optional schema=type="boolean" default=false
+schema LaunchedAgentProfile property llm_profile_ref optional schema=anyOf=[type="string",type="null"]
@@ -2170,0 +2198,3 @@
+schema NoOpCondenserSettings property condenser_kind optional schema=type="string" const="no_op" default="no_op"
+schema NoOpCondenserSettings property enabled optional schema=type="boolean" default=true
+schema NoOpCondenserSettings type="object"
@@ -2182,0 +2213,18 @@
+schema OpenHandsAgentProfile property agent optional schema=type="string" default="CodeActAgent"
+schema OpenHandsAgentProfile property agent_kind optional schema=type="string" const="openhands" default="openhands"
+schema OpenHandsAgentProfile property condenser optional schema=oneOf=[LLMSummarizingCondenserSettings,NoOpCondenserSettings]
+schema OpenHandsAgentProfile property disabled_skills optional schema=type="array" items=type="string"
+schema OpenHandsAgentProfile property enable_sub_agents optional schema=type="boolean" default=false
+schema OpenHandsAgentProfile property enable_switch_llm_tool optional schema=type="boolean" default=true
+schema OpenHandsAgentProfile property id optional schema=type="string" format="uuid"
+schema OpenHandsAgentProfile property llm_profile_ref required schema=type="string" minLength=1
+schema OpenHandsAgentProfile property mcp_server_refs optional schema=anyOf=[type="array" items=type="string",type="null"]
+schema OpenHandsAgentProfile property name required schema=type="string" minLength=1
+schema OpenHandsAgentProfile property revision optional schema=type="integer" default=0 minimum=0.0
+schema OpenHandsAgentProfile property schema_version optional schema=type="integer" default=2 minimum=1.0
+schema OpenHandsAgentProfile property secret_refs optional schema=anyOf=[type="array" items=type="string",type="null"]
+schema OpenHandsAgentProfile property system_message_suffix optional schema=anyOf=[type="string",type="null"]
+schema OpenHandsAgentProfile property tool_concurrency_limit optional schema=type="integer" default=1 minimum=1.0
+schema OpenHandsAgentProfile property tools optional schema=anyOf=[type="array" items=Tool-Input,type="null"]
+schema OpenHandsAgentProfile property verification optional schema=ProfileVerificationSettings
+schema OpenHandsAgentProfile type="object" additionalProperties=false
@@ -2289,0 +2338,8 @@
+schema ProfileVerificationSettings property critic_enabled optional schema=type="boolean" default=false
+schema ProfileVerificationSettings property critic_mode optional schema=type="string" enum=["finish_and_message","all_actions"] default="finish_and_message"
+schema ProfileVerificationSettings property critic_model_name optional schema=anyOf=[type="string",type="null"]
+schema ProfileVerificationSettings property critic_server_url optional schema=anyOf=[type="string",type="null"]
+schema ProfileVerificationSettings property critic_threshold optional schema=type="number" default=0.6 minimum=0.0 maximum=1.0
+schema ProfileVerificationSettings property enable_iterative_refinement optional schema=type="boolean" default=false
+schema ProfileVerificationSettings property max_refinement_iterations optional schema=type="integer" default=3 minimum=1.0
+schema ProfileVerificationSettings type="object"
@@ -2489,0 +2546 @@
+schema StartConversationRequest property agent_profile optional schema=anyOf=[oneOf=[OpenHandsAgentProfile,ACPAgentProfile],type="null"]
```

## Issue Number

Fixes #5141

## How to Test

Unit tests:

```bash
uv run pytest tests/sdk/profiles tests/agent_server/test_agent_launch_parity.py \
  tests/agent_server/test_agent_profile_conv_start.py \
  tests/agent_server/test_agent_launch_additions.py \
  tests/agent_server/test_acp_skill_sourcing.py \
  tests/agent_server/docker_runtime tests/agent_server/test_conversation_router.py -q
```

End-to-end against a real agent-server (this is the interesting one — it reproduces the table in #5141 without canvas):

```bash
uv run python .pr/launch_parity_e2e.py
```

It boots `python -m openhands.agent_server` on a temp persistence dir, stores two identically-configured profiles (`default` and `default-copy`), launches a conversation through `agent_profile_id` for each, through an inline `agent_profile` draft, and through the deprecated `agent_settings`, then diffs the agents the server actually built against the `materialize` preview of the same profile. Output in `.pr/launch_parity_e2e_output.txt`:

```
legacy current_datetime sent 2020-01-01T00:00, launched with 2026-09-17T14:08:55.977969-04:00
default-copy (agent_profile_id): MATCH
inline (agent_profile): MATCH
legacy (agent_settings): MATCH
materialize (default): MATCH
dangling refs -> HTTP 422 {"detail":{"code":"unresolved_profile_references","message":"LLM profile 'gone' not found; MCP server(s) not configured: nope","dangling_llm_profile_ref":"gone","dangling_mcp_server_refs":["nope"]}}
RESULT: PASS
```

Compared field by field: `llm` (whole dump), tools, MCP keys, skills, suffix, `disabled_skills`, `load_project_skills`, `load_memory`, condenser, critic, concurrency, switch-LLM and whether a timestamp is present. The one deliberate exception is the skill catalog on the `agent_settings` path: that payload carries the client's own catalog (canvas assembles one today), which is exactly what the canvas follow-up removes.

## Type

- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Behavior changes reviewers should weigh:

- **A dangling *LLM* ref at start is now 422, not 404**, and the 422 detail gained `code` / `dangling_llm_profile_ref` (the old `message` / `dangling_mcp_server_refs` keys are unchanged). Canvas never reads that status — it pre-checks the profile list and downgrades to `agent_settings` — and that rule is what the follow-up deletes.
- **`agent_settings` is no longer converted in the validator**, so `StartConversationRequest(agent_settings=...).agent` is `None` until the server resolves it, and an invalid payload is rejected by the start endpoint (422) rather than at parse time. The field also lost `exclude=True` so it round-trips over the wire. An explicit `"agent": null` alongside another source no longer crashes.
- **The `agent_settings` path now gets the launch-owned fields too**, because it goes through the same pipeline: streaming forced on, `load_project_skills=True`, browser added when the payload's `tools` is null and the runtime has it, and a fresh `current_datetime` instead of the saved one. Canvas already sends `stream: true`, `load_project_skills: true` and an explicit tool list, so its payload is unaffected; a hand-rolled REST client that relied on those staying off would see the change.
- **The Docker runtime now resolves twice**: once before the container starts (`build_agent=False`, so a dangling ref still fails fast without paying for a container) and once after, with the container's own runtime answer (`GET /server_info` for browser availability, `openhands_managed` skills). Previously the host's answers were used for a container agent, which meant an ACP profile in Docker got no managed skills.
- `/server_info` advertises `unified_agent_launch_v1`.

Not fixed here, found while testing: a condenser's `max_tokens` inheritance keys off `model_fields_set`, so a stored profile (loaded from JSON, every field "set") does not inherit the LLM's token limit while an in-memory one does. It is consistent across today's product paths (both read persisted JSON) and predates this PR, so I left it alone rather than widen the diff.

Follow-ups, per #5141:

- Canvas: launch with a profile on every path, delete the `default` name rule, the LLM-mismatch and dangling-ref fallbacks, the `use-llm-configured` copy and client-side agent assembly, and send `<RUNTIME_SERVICES>` via `agent_launch_additions`.
- Cloud: resolve through `prepare_agent_launch`, stop overwriting fields the profile carries, surface resolution failures.
- The one-time `default`-profile refresh from global `agent_settings` belongs with the canvas step: doing it before the settings pages edit the active profile would just let it drift again.
- #4958 (PR #5151, tool catalog / resolved tools) rebases onto this; its `profile_launch.py` is superseded by `agent_launch.py`.

---
<details>

**Variants & Base Images**
|---|---|---|---|
| python-slim | amd64, arm64 | `python-node-runtime` | [Link](https://hub.docker.com/_/python-node-runtime) |
| python-minimal | amd64, arm64 | `python-node-runtime` | [Link](https://hub.docker.com/_/python-node-runtime) |
| python | amd64, arm64 | `python-node-runtime` | [Link](https://hub.docker.com/_/python-node-runtime) |

```bash
docker pull ghcr.io/openhands/agent-server:97f40ab-python
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  --name agent-server-97f40ab-python \
  ghcr.io/openhands/agent-server:97f40ab-python
```

```
ghcr.io/openhands/agent-server:97f40ab-golang-amd64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-golang-amd64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-golang-amd64
ghcr.io/openhands/agent-server:97f40ab-golang_tag_1.21-bookworm-amd64
ghcr.io/openhands/agent-server:97f40ab-golang-arm64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-golang-arm64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-golang-arm64
ghcr.io/openhands/agent-server:97f40ab-golang_tag_1.21-bookworm-arm64
ghcr.io/openhands/agent-server:97f40ab-java-amd64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-java-amd64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-java-amd64
ghcr.io/openhands/agent-server:97f40ab-eclipse-temurin_tag_17-jdk-amd64
ghcr.io/openhands/agent-server:97f40ab-java-arm64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-java-arm64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-java-arm64
ghcr.io/openhands/agent-server:97f40ab-eclipse-temurin_tag_17-jdk-arm64
ghcr.io/openhands/agent-server:97f40ab-python-amd64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-amd64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-amd64
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-amd64
ghcr.io/openhands/agent-server:97f40ab-python-arm64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-arm64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-arm64
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-arm64
ghcr.io/openhands/agent-server:97f40ab-python-minimal-amd64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-minimal-amd64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-minimal-amd64
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-minimal-amd64
ghcr.io/openhands/agent-server:97f40ab-python-minimal-arm64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-minimal-arm64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-minimal-arm64
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-minimal-arm64
ghcr.io/openhands/agent-server:97f40ab-python-slim-amd64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-slim-amd64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-slim-amd64
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-slim-amd64
ghcr.io/openhands/agent-server:97f40ab-python-slim-arm64
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-slim-arm64
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-slim-arm64
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-slim-arm64
ghcr.io/openhands/agent-server:97f40ab-golang
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-golang
ghcr.io/openhands/agent-server:agent-profile-unified-launch-golang
ghcr.io/openhands/agent-server:97f40ab-golang_tag_1.21-bookworm
ghcr.io/openhands/agent-server:97f40ab-java
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-java
ghcr.io/openhands/agent-server:agent-profile-unified-launch-java
ghcr.io/openhands/agent-server:97f40ab-eclipse-temurin_tag_17-jdk
ghcr.io/openhands/agent-server:97f40ab-python-minimal
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-minimal
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-minimal
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-minimal
ghcr.io/openhands/agent-server:97f40ab-python-slim
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python-slim
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python-slim
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime-slim
ghcr.io/openhands/agent-server:97f40ab-python
ghcr.io/openhands/agent-server:97f40abbf1634e6239b09069b40844b42ccadca6-python
ghcr.io/openhands/agent-server:agent-profile-unified-launch-python
ghcr.io/openhands/agent-server:97f40ab-python-node-runtime
```

- Each variant tag (e.g., `97f40ab-python`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `97f40ab-python-amd64`) are also available if needed
</details>


simonrosenberg:
Thanks — on the eval-risk flag, here is what I can evidence from this branch so a maintainer has the facts to decide. I have **not** run evals.

**Who actually sees the `agent_settings` behavior changes.** Canvas's payload builder (`buildConfiguredOpenHandsAgentSettings` / `buildAgentContext` on `OpenHands/OpenHands@main`) already sends `llm.stream = true`, `load_project_skills: true`, `load_user_skills: true` and an explicit `tools` list from `getAgentTools`. So of the four changes, three are no-ops for canvas and the fourth is the fix itself: `current_datetime` is now computed at launch instead of being the value saved with settings. Verified live — the e2e script sent `2020-01-01T00:00` and the launched agent came back with the launch timestamp (`.pr/launch_parity_e2e_output.txt`).

**Cloud is not on this path.** The enterprise app server builds a concrete `agent` (`create_kwargs = {'agent': agent, ...}` in `live_status_app_conversation_service.py`) and never sets the request's `agent_settings`, so neither the pipeline change nor the removed validator conversion reaches it.

**ACP-in-Docker.** Previously the *host's* `acp_skill_sourcing` decided what a container's ACP agent got, so with the default host config (`native`) an ACP profile in Docker launched with no managed skills while the container image itself sets `OH_ACP_SKILL_SOURCING=openhands_managed`. That is the drift this PR removes by passing the container's answer. It does change the ACP prompt in Docker, and it is the one change I would point an eval at if you want one.

**Unchanged for stored-profile launches** (the path evals exercise): the parity e2e diffs `llm`, tools, MCP keys, skills, suffix, `disabled_skills`, `load_project_skills`, `load_memory`, condenser, critic, concurrency and switch-LLM between two identically-configured profiles, an inline draft, the legacy payload and the `materialize` preview — all MATCH.

On the `warn_deprecated` note: agreed and intentional — 1.50.0 is the next minor, so the warning arms exactly when the deprecation takes effect.

PR #5178, opened 2026-09-18
fix(telemetry): guard against missing cache_creation_tokens attribute

yunusgungor:
litellm's PromptTokensDetailsWrapper drops 'cache_creation_tokens' from the instance when it is None (providers that don't return it), yet the field can still be present in model_fields_set. Telemetry._cache_buckets accessed details.cache_creation_tokens directly after checking model_fields_set, so every LLM completion raised

  AttributeError: 'PromptTokensDetailsWrapper' object has no attribute
  'cache_creation_tokens'

which aborted the whole conversation run after each tool call. Access the attribute defensively via getattr (default 0) and apply the same guard to cached_tokens. Add a regression test reproducing the deleted- attribute scenario.

HUMAN:

Tested: reproduced the exact AttributeError against litellm 1.101.0 (PromptTokensDetailsWrapper(cached_tokens=7)); patched _cache_buckets returns (7, 0). Regression test added.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


yunusgungor:
litellm's PromptTokensDetailsWrapper drops 'cache_creation_tokens' from the instance when it is None (providers that don't return it), yet the field can still be present in model_fields_set. Telemetry._cache_buckets accessed details.cache_creation_tokens directly after checking model_fields_set, so every LLM completion raised

  AttributeError: 'PromptTokensDetailsWrapper' object has no attribute
  'cache_creation_tokens'

which aborted the whole conversation run after each tool call. Access the attribute defensively via getattr (default 0) and apply the same guard to cached_tokens. Add a regression test reproducing the deleted- attribute scenario.

HUMAN:

Tested: reproduced the exact AttributeError against litellm 1.101.0 (PromptTokensDetailsWrapper(cached_tokens=7)); patched _cache_buckets returns (7, 0). Regression test added.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

yunusgungor:
HUMAN:
Tested: reproduced the exact AttributeError against litellm 1.101.0 (PromptTokensDetailsWrapper(cached_tokens=7)); patched _cache_buckets returns (7, 0) instead of raising. A regression test covers this scenario.

AGENT:
Reproduced and verified against the real production dependency versions:
- litellm 1.101.0: PromptTokensDetailsWrapper(cached_tokens=7) deletes cache_creation_tokens from __dict__ (None) yet leaves it in model_fields_set.
- Before: the model_fields_set guard passes, then direct details.cache_creation_tokens access raises AttributeError.
- After: getattr(details, "cache_creation_tokens", 0) returns (7, 0) — verified live with the agent-server python.
- New test test_cache_buckets_handle_missing_cache_creation_tokens_attribute reproduces this and asserts (7, 0).

## Why
OpenHands conversations using a provider whose prompt_tokens_details contains only cached_tokens crash after every LLM completion with AttributeError... Work is performed but no assistant reply is ever produced, so the conversation appears stuck and every user message fails.

## Summary
- _cache_buckets now reads cache_creation_tokens and cached_tokens defensively via getattr(..., 0) instead of direct attribute access after a model_fields_set guard.
- Same guard applied to the responses-shape branch (input_tokens_details.cached_tokens).
- Added a regression test covering the deleted-attribute scenario.

## Issue Number
Fixes #<GERÇEK_ISSUE_NUMARASI>

## How to Test
Run: uv run pytest tests/sdk/llm/test_llm_span_cost.py -k "cache_buckets or span"
Or reproduce: PromptTokensDetailsWrapper(cached_tokens=7) -> Telemetry._cache_buckets(usage) returns (7, 0), no AttributeError.

## Type
- [x] Bug fix

## Notes
The fix was also applied locally to the running agent-server that hit this in production; after the patch and restart the AttributeError no longer appears.

PR #5187, opened 2026-09-19
fix(sdk): stop silent websocket callback clients

### LINKED ISSUE #4670: WebSocketCallbackClient.stop() does not close a silent WebSocket and leaves its worker thread running
## Summary

In `openhands-sdk==1.43.1`, `WebSocketCallbackClient.stop()` does not reliably stop its worker thread when the WebSocket connection is open but silent.

`stop()` sets `_stop` and waits for five seconds, but the worker can remain blocked inside:

```python
async for message in ws:
```

Because setting a `threading.Event` does not wake the pending WebSocket receive, `stop()` times out. It then clears `self._thread` even though the underlying thread is still alive.

This is observable when repeatedly creating and closing `RemoteConversation` instances: WebSocket worker threads and connections can accumulate after the conversations have been closed.

## Environment

- Python: 3.13.4
- `openhands-sdk`: 1.43.1
- `openhands-agent-server`: 1.43.1
- `websockets`: 16.1.1

## Relevant implementation

The current `stop()` implementation is:

```python
def stop(self) -> None:
    if not self._thread:
        return
    self._stop.set()
    self._thread.join(timeout=5)
    self._thread = None
```

The worker remains blocked in `_client_loop()` while no message or close frame is received:

```python
async with websockets.connect(ws_url) as ws:
    async for message in ws:
        if self._stop.is_set():
            break
        ...
```

The `_stop` check can only run after `async for` yields a message.

## Minimal reproduction

The following test starts a silent local WebSocket server, connects an official `WebSocketCallbackClient`, and calls `stop()`:

```python
import asyncio
import threading
import time

import websockets
from openhands.sdk.conversation.impl.remote_conversation import (
    WebSocketCallbackClient,
)

server_ready = threading.Event()
client_connected = threading.Event()
state = {}


async def handler(websocket):
    client_connected.set()
    await websocket.wait_closed()


async def serve():
    stop_event = asyncio.Event()
    state["loop"] = asyncio.get_running_loop()
    state["stop_event"] = stop_event

    async with websockets.serve(handler, "127.0.0.1", 0) as server:
        state["port"] = server.sockets[0].getsockname()[1]
        server_ready.set()
        await stop_event.wait()


server_thread = threading.Thread(
    target=lambda: asyncio.run(serve()),
    daemon=True,
)
server_thread.start()
assert server_ready.wait(5)

client = WebSocketCallbackClient(
    host=f"http://127.0.0.1:{state['port']}",
    conversation_id="shutdown-reproduction",
    callback=lambda event: None,
)
client.start()

worker_thread = client._thread
assert worker_thread is not None
assert client_connected.wait(5)

started = time.monotonic()
client.stop()
elapsed = time.monotonic() - started

print(f"stop elapsed: {elapsed:.3f}s")
print(f"worker alive: {worker_thread.is_alive()}")
print(f"client thread reference: {client._thread}")

state["loop"].call_soon_threadsafe(state["stop_event"].set)
server_thread.join(5)
```

Observed output:

```text
stop elapsed: 5.005s
worker alive: True
client thread reference: None
```

If the server subsequently closes the peer connection, the original worker thread exits. This confirms that the stop flag was set, but the blocked receive was not awakened.

## Expected behavior

After `WebSocketCallbackClient.stop()` returns:

1. The active WebSocket connection has been closed.
2. The worker thread has exited.
3. `self._thread` is cleared only after the thread is no longer alive.
4. Repeated start/stop cycles do not accumulate worker threads or sockets.

## Suggested direction

The WebSocket receive task should be actively awakened during shutdown. One possible approach is to retain the worker event loop and its main asyncio task, then cancel that task from `stop()` using `loop.call_soon_threadsafe(task.cancel)`.

Cancelling the task unwinds the `websockets.connect()` async context manager, which closes the WebSocket. `stop()` can then join the worker and clear `self._thread` only after confirming that the thread exited.

Alternatively, the implementation could retain the active WebSocket protocol object and close it through its owning event loop.

It would also be useful to add regression coverage for:

- stopping an established but silent WebSocket;
- calling `stop()` immediately after `start()`;
- repeated connect/stop cycles with no worker-thread growth;
- preserving the thread reference when shutdown times out.

Oscar-Williams:
HUMAN:

I reviewed the implementation and regression coverage for Issue #4670. I understand the event-loop task cancellation and thread-lifecycle changes, and confirm that the proposed behavior matches the issue’s shutdown requirements. I will complete the final review before marking this PR ready for review.

---

AGENT:

## Why

`WebSocketCallbackClient.stop()` can leave a connected but silent WebSocket and
its worker thread running after the five-second join timeout.

## Summary

- Cancel the worker task through its event loop so pending WebSocket receives unwind.
- Keep the worker reference until the thread has exited.
- Add a real silent-WebSocket regression test and timeout-reference coverage.

## Issue Number

Closes #4670

## How to Test

```text
python -m pytest -q tests/sdk/conversation/remote/test_websocket_client.py -k test_websocket_client_stops_silent_connection --maxfail=1
python -m pytest -q tests/sdk/conversation/remote/test_websocket_client.py
python -m pytest -q tests/sdk/conversation/remote -k "not test_remote_events_list_add_event_deduplication" --maxfail=1
```

The regression test fails on base because `stop()` waits about five seconds and
the silent worker remains blocked. The same test passes on the change; the full
remote subset passes 147 tests with one unchanged baseline test deselected.

## Video/Screenshots

Not applicable: this is an SDK WebSocket lifecycle fix.

## Design Doc

Not applicable for this small bug fix.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The complete remote suite has one unchanged timestamp-sensitive test that also
fails on the fixed base. Ruff, Pyright, pycodestyle, import/tool checks, and
`uv pip check` pass in the dedicated Windows environment.


Oscar-Williams:
Friendly ping — before taking this further, two things I'd appreciate direction on:

1. Should the linked issue (#4670) get a readiness marker (e.g. ready-for-dev), or is external work on it not being accepted right now?
2. The PR carries one test that fails unchanged from base on my environment (noted in the PR description); if the official Python 3.13 pre-commit run surfaces the same, I'd like to confirm whether it's mine to fix or a pre-existing environment artifact.

The fix itself is small (join the worker thread on `stop()` instead of leaving it stranded on a silent WebSocket) — happy to adjust anything.

Issue #5225, opened 2026-09-22
[Bug]: LLMBadRequestError from a poisoned tool_result permanently wedges conversations — no recovery path in Agent.step


jpshackelford:
### Bug Description

A single provider-rejected `tool_result` message permanently wedges a conversation. When the provider returns a 400 that the SDK maps to `LLMBadRequestError`, `Agent.step` does not catch it — unlike `LLMContentPolicyViolationError`, `LLMMalformedConversationHistoryError`, or `LLMContextWindowExceedError`, which each have a soft-recovery branch. The exception bubbles out, the run terminates with `ConversationExecutionStatus.ERROR`, and because the offending event is already persisted in history, every subsequent run/resume rebuilds the same prompt and fails identically. The conversation cannot progress.

### Expected Behavior

When the provider rejects a specific `tool_result` message (identified by index and content path, e.g. `messages.148.content.0.tool_result.content.1.image.source.base64: invalid base64 data`), the SDK should:

- neutralize the offending observation (replace its content with a text-only placeholder while leaving the original event in the persisted store for audit),
- log a warning explaining what was dropped and why,
- and continue the run.

This mirrors how `LLMContextWindowExceedError` and `LLMMalformedConversationHistoryError` are routed into condensation-based recovery today.

### Actual Behavior

`LLMBadRequestError` reaches `LocalConversation.run`, is written to a `ConversationErrorEvent`, and `ConversationRunError` is re-raised. The poisoned event stays in `state.events`. Every resume rebuilds the same request and fails identically.

Reproduce with `pytest` against the SDK repo:

```bash
uv sync
uv run pytest tests/repro/test_bad_request_wedges_conversation.py -q
```

`tests/repro/test_bad_request_wedges_conversation.py` (sketch — the exact stub API depends on how the test suite fakes an LLM):

```python
import pytest

from openhands.sdk.llm.exceptions import LLMBadRequestError


def test_bad_tool_result_does_not_wedge_conversation(fake_conversation_with_bad_image):
    """Given a persisted ObservationEvent that the provider rejects with a
    400 pointing at a specific messages.N.content... offset, run() must
    recover instead of raising the same error on every attempt."""
    conversation = fake_conversation_with_bad_image  # fixture builds state
                                                     # with one poisoned event
    conversation.run()   # today: raises ConversationRunError(LLMBadRequestError)
    # desired: run() returns; the offending observation is replaced with a
[...]

PR #5226, opened 2026-09-22
fix(secrets): opt base64 payloads out of secret masking

### LINKED ISSUE #5224: [Bug]: Secret masking corrupts base64 payloads (e.g. browser screenshots), permanently wedging the conversation
### Bug Description

`SecretRegistry.mask_secrets_in_output` runs `str.replace(secret_value, "<secret-hidden>")` against every string field on every `Observation`, via `Tool.__call__` → `mask_secrets_in_model` → `_mask_value`. When a field holds a large base64 payload (`BrowserObservation.screenshot_data`, or a `data:` URL inside `ImageContent.image_urls`), any registered secret whose raw value happens to appear as a substring of the base64 gets replaced with the literal text `<secret-hidden>`, corrupting the payload. The corrupted observation is then persisted into the event store, so from that point on the LLM request built from history is rejected by the provider on every attempt.

### Expected Behavior

The secret-masking pass over an `Observation` must not modify binary/base64 payload fields. A screenshot or other opaque byte blob is not maskable prose; it should round-trip through the masker unchanged. The conversation should be able to continue normally after a screenshot tool call.

### Actual Behavior

The base64 `screenshot_data` field on a `BrowserObservation` is corrupted in place with one or more literal `<secret-hidden>` substrings whenever any registered secret's raw value happens to appear as a substring of the base64. The corruption is written back to the persisted `ObservationEvent`, so every subsequent LLM turn rebuilds the same poisoned `tool_result` and fails with a provider 400 (e.g. Anthropic `invalid_request_error: ... image.source.base64: invalid base64 data`).

Reproduce with `pytest` against the SDK repo:

```bash
uv sync
uv run pytest tests/repro/test_masker_corrupts_base64.py -q
```

`tests/repro/test_masker_corrupts_base64.py`:

```python

from openhands.sdk.conversation.secret_registry import SecretRegistry
from openhands.tools.browser_use.definition import BrowserObservation


def test_masker_must_not_corrupt_base64_screenshot():
    reg = SecretRegistry()
    # Any short token whose raw characters are all valid base64 will do —
    # 200+ KB screenshots contain any 8–16-char secret substring by chance.
    reg.update_secrets({"MY_KEY": SecretStr("XYZ789")})
    reg.get_secret_value("MY_KEY")  # track for masking

    payload = "iVBORw0KGgoAAA" + "XYZ789" + "SUhEUgAAB4A=="
    obs = BrowserObservation(screenshot_data=payload)

    masked = reg.mask_secrets_in_model(obs)
    assert masked.screenshot_data == payload, (
        f"Masker corrupted base64 payload: {masked.screenshot_data!r}"
    )
```

Real occurrence: OpenHands Cloud conversation `9ace83223e5540198bd5a74258cc2f84`. The relevant observation event (`a436b767-cfc0-422c-a062-cf9a2eb0fcf2`, `browser_get_state`, 233 844 chars of base64) contains two literal `<secret-hidden>` substrings at offsets 160 493 and 228 375. The earlier screenshot in the same conversation (`8e7c7588…`, 168 996 chars) has no substitutions and decodes cleanly as PNG. Second, independent reproduction with a different model, larger screenshot, and different task: conversation `9d716b170bd4424c9a9307e78a821836` (message 148, 275 320-char screenshot, two substitutions at offsets 26 396 and 217 282).

### Steps to Reproduce

1. Install the SDK and browser tool package: `uv pip install openhands-sdk openhands-tools`.
2. Create a `SecretRegistry`, register a static secret whose raw value contains only base64 characters (any short token — collisions with real screenshots are effectively guaranteed for anything ≥ 8 chars), and call `get_secret_value(...)` once so the value is tracked for masking.
3. Instantiate a `BrowserObservation` with a `screenshot_data` string that contains the secret's raw value as a substring (in the real path, produced by `browser_get_state(include_screenshot=True)` on any browser session ≥ ~200 KB).
4. Call `SecretRegistry.mask_secrets_in_model(obs)` — the returned observation's `screenshot_data` now contains the literal string `<secret-hidden>` inserted mid-blob.
5. Feed the resulting `ObservationEvent` through the normal LLM turn — the provider rejects the `tool_result` with a 400 and the run cannot recover (see the linked companion issue for the recovery-path gap).

### Acceptance Criteria

- [ ] `SecretRegistry.mask_secrets_in_model(observation)` returns an observation whose base64/binary payload fields (at minimum `BrowserObservation.screenshot_data`) are byte-identical to the input, even when the raw bytes of a tracked secret appear as a substring inside them.
- [ ] A regression test in the SDK repo exercises the scenario end-to-end: register a secret whose characters collide with base64, produce a `BrowserObservation` containing that substring, run it through the masker, and assert the field is unchanged.
- [ ] The same protection covers `ImageContent.image_urls` entries that are `data:...;base64,...` URLs — the base64 tail is not mutated by the masker.
- [ ] A design note is added explaining the field-tagging (or equivalent) mechanism the SDK uses to opt a field out of secret masking, so future tools that carry binary payloads pick the safe default.
- [ ] Running `uv run pytest` against the SDK still passes with no other regressions.

### Installation Method

`uv pip install openhands-sdk openhands-tools` (also reproducible from a workspace `uv sync`).

### SDK Version

Reproduced against `main` at the time of filing (commit reachable from `main` on 2026-09-22). The bug is present in every release that has both `SecretRegistry.mask_secrets_in_model` and `BrowserObservation.screenshot_data: str`.

### Version Confirmation

### Python Version

Any supported version (independent of Python version — pure logic bug in `_mask_value`).

Reproduced against `claude-sonnet-4-5-20250929` and `claude-opus-4-7`. Any provider that validates base64 in tool-result image blocks (Anthropic, OpenAI, Gemini) will reject the poisoned observation.

### Operating System

Linux (OpenHands Cloud runtime). Platform-independent bug.

### Logs and Error Messages

```
litellm.BadRequestError: Error code: 400 - {'error': {'message':
'litellm.BadRequestError: AnthropicException - b\'{"type":"error",
"error":{"type":"invalid_request_error",
"message":"messages.108.content.0.tool_result.content.1.image.source.base64: invalid base64 data"},
"request_id":"req_011CfK5dZXFahJPCkNemJy5x"}\'. Received Model Group=claude-sonnet-4-5-20250929
Available Model Group Fallbacks=None', 'type': None, 'param': None, 'code': '400'}}
```

Direct evidence of the masker having mutated the persisted `screenshot_data`:

```
...NBDtfiqzPPmzZs/f759u<secret-hidden>el5VVLV69uzZyUZ...   (offset 160 493)
...GjSoWousssoq3bp<secret-hidden>jjjZPlxQ477DBy5...       (offset 228 375)
```

### Minimal Code Sample

```python

from openhands.sdk.conversation.secret_registry import SecretRegistry
from openhands.tools.browser_use.definition import BrowserObservation

reg = SecretRegistry()
reg.update_secrets({"MY_KEY": SecretStr("XYZ789")})
reg.get_secret_value("MY_KEY")  # track for masking

payload = "iVBORw0KGgoAAA" + "XYZ789" + "SUhEUgAAB4A=="
obs = BrowserObservation(screenshot_data=payload)

masked = reg.mask_secrets_in_model(obs)
# The masker replaced the substring inside the base64 blob with the
# literal marker, corrupting the payload:
assert "<secret-hidden>" in masked.screenshot_data
```

**Root cause, file by file:**

- `openhands-sdk/openhands/sdk/tool/tool.py`, `Tool.__call__` (~line 643):
    ```python
    registry = conversation.state.secret_registry
    return registry.mask_secrets_in_model(observation)
    ```
- `openhands-sdk/openhands/sdk/conversation/secret_registry.py`, `_mask_value` recurses into every `str` field with no type or annotation opt-out.
- `openhands-sdk/openhands/sdk/conversation/secret_registry.py:263` — `masked_text = masked_text.replace(value, "<secret-hidden>")` is a naive substring replace with no boundary check.
- `openhands-tools/openhands/tools/browser_use/definition.py:63` — `screenshot_data: str | None = Field(...)` — untagged base64 payload.

Any base64/binary field in any current or future tool has the same failure mode.

**Fix options**, in decreasing order of preference:

1. Tag binary payloads with a distinguished type and skip them in the masker. Introduce `Base64Payload = Annotated[str, SkipSecretMasking()]` (or a small wrapper model); have `_mask_value` / `_mask_model` skip fields whose annotation carries `SkipSecretMasking`. Retag `BrowserObservation.screenshot_data` and the `data:` URLs inside `ImageContent.image_urls` accordingly.
2. Generic field-level opt-out via `Field(json_schema_extra={"skip_secret_masking": True})`; check it in `_mask_model` before recursing.
3. Weaker fallback: in `mask_secrets_in_output`, refuse to `.replace` inside strings that look like a base64 blob (length > N, matches `^[A-Za-z0-9+/=\s]+$`). Heuristic; leaves the general primitive unsafe.

**Companion issue:** even after this masker fix, a single provider-rejected `tool_result` still wedges the whole conversation because `LLMBadRequestError` has no recovery path in `Agent.step` — see #5225.

---
_This bug report was created by an AI agent (OpenHands) on behalf of the user._

---
## OpenHands AI triage
**The following comments and acceptance criteria were added by the OpenHands AI agent.**

### Triage

Reproducible on `main` from the evidence in the report and the current code.
`Tool.__call__` (`openhands-sdk/openhands/sdk/tool/tool.py`) funnels every tool
result through `SecretRegistry.mask_secrets_in_model`, and `_mask_value` recurses
into every reachable `str`. `mask_secrets_in_output` then performs an unbounded
`str.replace(value, "<secret-hidden>")`, so a short tracked secret value that
occurs by chance inside an untagged base64 payload is rewritten with the literal
marker. The corrupted observation is persisted, so every later request rebuilt
from history fails on invalid base64.

Bounded scope:

- Adopt the design option the report itself ranks and the repository already
  uses: a generic per-field opt-out carried in `Field(json_schema_extra=...)` and
  read back from `model_fields`, matching the existing `acp_compatible` flag
  pattern in `AgentContext.validate_acp_compatibility()`. This is the smallest
  change and avoids introducing a parallel marker type or a second redaction
  mechanism.
- Apply the opt-out to the known base64 carriers: `BrowserObservation.screenshot_data`,
  `ImageContent.image_urls`, and the MCP image-block path in
  `MCPToolObservation.from_call_tool_result` that builds `data:...;base64,...`
  URLs.
- Keep masking every text field and every non-base64 string.

Explicit non-goals: the companion recovery issue (#5225) - no change to
`Agent.step` or `LLMBadRequestError` handling, so other provider-rejected
`tool_result` vectors stay out of scope; repairing or migrating conversations
already persisted with `<secret-hidden>` inside a payload (the fix only prevents
new corruption); and changing how text is masked (boundary checks, entropy
heuristics, or the marker string itself).

### Acceptance Criteria

- [ ] Masking a `BrowserObservation` whose base64 `screenshot_data` contains a registered secret's characters leaves `screenshot_data` unchanged, and it still decodes with `base64.b64decode` (covered by a `pytest` test).
- [ ] A `data:image/<type>;base64,<payload>` URL in `ImageContent.image_urls` is not substring-replaced when the base64 portion contains a registered secret's characters, and the data URL still decodes to the original bytes.
- [ ] The MCP image path (`MCPToolObservation.from_call_tool_result` with an `mcp.types.ImageContent` block) yields an unmodified base64 data URL under the same condition.
- [ ] `BrowserObservation.to_llm_content` returns an `ImageContent` whose data-URL base64 equals the original screenshot bytes.
- [ ] Masking still applies to text - a registered secret value in a text field (`Observation.content` / `TextContent.text`) is still replaced with `<secret-hidden>`, and a secret value appearing in a non-base64 image URL string is still masked.
- [ ] No new required fields, serialization changes, or schema-version bump are introduced for `BrowserObservation`, `Observation`, or `ImageContent`; previously persisted events still load.
- [ ] `Tool.__call__` still returns a masked observation for tools that emit only text, so new tools remain masked by default.
- [ ] Tests added under the existing SDK test layout pass, including a regression test that a base64 payload containing a secret value is preserved intact.

jpshackelford:
HUMAN:

Fixes issue #5224 — the secret masker corrupts base64 screenshot payloads and borks the conversation.

---

AGENT:

## Why

Fixes #5224.

`SecretRegistry.mask_secrets_in_model` walked every string field on every observation and ran a plain substring `.replace(secret_value, "<secret-hidden>")` against it. When a field held a large opaque payload — `BrowserObservation.screenshot_data`, or a `data:...;base64,...` URL inside `ImageContent.image_urls` — any registered secret whose raw characters happened to appear as a substring of the base64 (a near-certainty for any 8+ char token in a 200 KB+ screenshot) got the literal string `<secret-hidden>` spliced into the middle of the payload.

The corrupted observation was persisted into the event store, so from that point on the LLM request built from history was rejected by the provider on every attempt (Anthropic `image.source.base64: invalid base64 data`), permanently wedging the conversation.

Regression was introduced by #4788, which wired `mask_secrets_in_model` into the shared tool chokepoint. Present in every release from v1.45.0 through current `main`.

## Summary

- Introduce `SkipSecretMasking` in `openhands.sdk.utils.masking` (re-exported from `openhands.sdk.utils`). Tag a Pydantic field with `Annotated[T, SkipSecretMasking()]` and `_mask_model` skips it entirely; ordinary fields are still walked and masked as before. The class docstring is the design note explaining when and why to reach for the marker.
- Tag `BrowserObservation.screenshot_data` (raw base64 blob) and `ImageContent.image_urls` (the `data:...;base64,...` URLs rendered from it) so both round-trip byte-for-byte through the masker. Remote URLs never legitimately embed a registered secret either, so opting the whole list out is safe and keeps the rule simple.
- The marker lives in `openhands.sdk.utils` so both the masker (`sdk.conversation.secret_registry`) and the models it walks (`sdk.llm.message`, which the conversation package imports transitively) can reference it without an import cycle.

## Issue Number

Fixes #5224.

## How to Test

Repro from the issue, run against this branch:

```bash
uv sync
uv run python - <<'PY'
from openhands.sdk.conversation.secret_registry import SecretRegistry
from openhands.tools.browser_use.definition import BrowserObservation

reg = SecretRegistry()
reg.update_secrets({"MY_KEY": "XYZ789"})
reg.get_secret_value("MY_KEY")

payload = "iVBORw0KGgoAAA" + "XYZ789" + "SUhEUgAAB4A=="
obs = BrowserObservation(screenshot_data=payload)
masked = reg.mask_secrets_in_model(obs)
assert masked.screenshot_data == payload, masked.screenshot_data
print("OK — screenshot round-tripped unchanged")
PY
```

Before this change the assert fails with `iVBORw0KGgoAAA<secret-hidden>SUhEUgAAB4A==`. On this branch it prints `OK`.

Regression tests added:

- `tests/sdk/conversation/test_secret_registry_model_masking.py::test_skip_secret_masking_leaves_annotated_string_field_alone` — generic marker coverage on a synthetic model.
- `tests/sdk/conversation/test_secret_registry_model_masking.py::test_skip_secret_masking_covers_list_fields` — same for `list[str]` fields (matches the `ImageContent.image_urls` shape).
- `tests/tools/browser_use/test_browser_observation.py::test_secret_masker_leaves_screenshot_and_image_urls_alone` — end-to-end scenario from the issue: register a secret whose characters collide with base64, run a real `BrowserObservation` (plus the `ImageContent` it renders) through the masker, and assert the payload is byte-identical while the surrounding prose is still masked.

Full local run (fix branch):

```bash
uv run pytest tests/sdk/conversation/ tests/sdk/llm/ tests/tools/browser_use/ tests/sdk/tool/ -q
# ================ 2181 passed, 21 warnings in 202.68s (0:03:22) =================

uv run pytest tests/sdk/agent/ -q
# =============== 915 passed, 6 deselected, 38 warnings in 22.31s ================

uv run ruff format ...   # 1 file reformatted, 7 files left unchanged
uv run ruff check  ...   # All checks passed!
```

## Video/Screenshots

N/A — pure logic change in the secret masker plus field-level tags. Behavior is captured by the assertions in the tests above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Companion issue #5225 tracks the recovery-path gap: even with this fix, a single provider-rejected `tool_result` still wedges the whole conversation because `LLMBadRequestError` has no recovery path in `Agent.step`. Out of scope for this PR.

---

This pull request was created by an AI agent (OpenHands) on behalf of @jpshackelford.


jpshackelford (commented):
# Code Review Roasted Output

**Taste:** good.

Small, focused fix that matches the issue's preferred remedy: an
`Annotated[T, SkipSecretMasking()]` marker that `_mask_model` checks before
recursing. The marker lives in `openhands.sdk.utils` to avoid the cycle
between `secret_registry` and `llm.message`, and both known base64 carriers
(`BrowserObservation.screenshot_data` and `ImageContent.image_urls`) are
tagged, which also transitively covers `MCPToolObservation.from_call_tool_result`
since it emits its data-URL through `ImageContent`. `file_editor` and
`vision_inspect` also produce their images through `ImageContent`, so the
opt-out reaches every current base64 sink without per-tool duplication.

The regression tests exercise both the generic marker (`str` and `list[str]`
shapes) and the end-to-end `BrowserObservation` → `to_llm_content` →
`ImageContent` path from the issue reproducer, and untagged sibling fields
are still asserted to be masked. Every current-head CI check is green.

## Acceptance criteria

- [x] Masking a `BrowserObservation` leaves `screenshot_data` byte-identical
  when a registered secret's characters occur inside the base64
  (`test_secret_masker_leaves_screenshot_and_image_urls_alone`).
- [x] A `data:image/*;base64,*` URL inside `ImageContent.image_urls` is not
  substring-replaced under the same condition (same test, plus
  `test_skip_secret_masking_covers_list_fields`).
- [x] The MCP image path is covered transitively — `MCPToolObservation.from_call_tool_result`
  constructs `ImageContent(image_urls=[...])`, and the tag on that field opts
  the whole list out. No dedicated MCP test, but the mechanism reaches it.
- [x] `BrowserObservation.to_llm_content` yields an `ImageContent` whose data
  URL still ends in the original payload after masking (asserted in the same
  end-to-end test).
- [x] Text fields are still masked — the sibling `label`/`text` fields in
  both new tests assert the marker is still substituted where it should be.
- [x] No new required fields, serializer changes, or schema-version bump.
  `Annotated` metadata is invisible to the wire format; previously persisted
  events still load unchanged.
- [x] `Tool.__call__` still routes every observation through
  `mask_secrets_in_model`; new tools inherit masking by default and only
  opt out per-field.

## Observation (non-blocking)

Opting the entire `image_urls` list out of masking is the right call for the
base64 case, and the file comment says as much. It does mean that if a
caller ever puts a secret's characters into a plain remote URL (e.g., a
signed `https://…?token=SECRET` link), that string will no longer be masked
in that field. In practice masking such a URL would corrupt it anyway, and
this is called out in the design note, so it's a reasonable trade — flagging
only so a future reader doesn't rediscover it as a surprise.

## [RISK ASSESSMENT]

**Risk: LOW.** Behavior change is confined to two field annotations plus a
23-line marker module; the semantic change is subtractive (some fields are
no longer scanned) and every current caller path is covered by the new
tests. No compatibility or serialization surface is touched.

**Verdict: APPROVE.**

**Architectural insight.** The regression was possible because a global,
unconditional pass mutated _every_ string reachable from an observation.
Making the opt-out declarative on the field (rather than a per-tool
post-processing step) keeps a single chokepoint — `Tool.__call__` still
routes through `mask_secrets_in_model` for every tool — while letting each
model own the "this field is an opaque payload" contract in its own
declaration. Future tools that carry binaries pick the safe default (still
masked) but have a one-line opt-out available at the site where the field
is defined.

---

_This review was generated by an AI agent (OpenHands) on behalf of @jpshackelford._

> **Improve this review?** If feedback seems incorrect or irrelevant, update the
> repository's `.agents/skills/custom-codereview-guide.md` (with the `/codereview`
> trigger), then re-request review. The reviewer reads the guide from the PR head.
>
> **Resolve with AI?** Install the
> [iterate skill](https://github.com/OpenHands/extensions/tree/main/skills/iterate)
> and run `/iterate`.
>

jpshackelford:
Closing in favor of #5227.

